PR #3, opened 2018-02-01
Remove empty class

wjsl:
it's a sad lil guy


warehouse/query-core/src/main/java/datawave/query/iterator/FunctionIterator.java
```diff
@@ -1,5 +0,0 @@
-package datawave.query.iterator;
-
-public class FunctionIterator {
```
ivakegg: I wonder who created this class?  Thank you.

PR #24, opened 2018-02-05
Closes #9: Updated geowave functions descriptor to generate tighter q…

### LINKED ISSUE #9: Add ability to decompose multipolygons into smaller ranges when computing geo query bounds.
Instead of generating a bounding envelope which contains the entire multipolygon, try to break the multipolygon down into smaller bounding areas so that we generate tighter, more accurate scan ranges.

jwomeara:
…uery bounds for WKT geometries which have 4 or less internal geometries.


warehouse/query-core/src/main/java/datawave/query/jexl/functions/GeoWaveFunctionsDescriptor.java
```diff
@@ -53,7 +68,9 @@ public GeoWaveJexlArgumentDescriptor(ASTFunctionNode node, String name, List<Jex
         public JexlNode getIndexQuery(ShardQueryConfiguration config, MetadataHelper helper, DateIndexHelper dateIndexHelper, Set<String> datatypeFilter) {
```
jwomeara: @ivakegg I'm going to update this tomorrow morning to pull the max expansion/envelopes from the configuration here!  Stay tuned!

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GeoWaveFunctionsDescriptor.java
```diff
@@ -219,4 +238,61 @@ public JexlNode apply(ByteArrayRange input) {
...
+            // find all geometries that intersect with this geometry, and as this combined set grows, we will need to
+            // reevaluate all the remaining geometries until no intersections are found, or until we run out of
+            // geometries
+            boolean geomAdded = true;
+            while (geomAdded) {
+                geomAdded = false;
+                // for each of the remaining geometries
+                for (Geometry srcGeom : geometries) {
+                    // check to see if this srcGeom intersects with any of the overlapped geometries
+                    for (Geometry ovlp : overlapped) {
```
ivakegg: Some of this work is redundant.  Once an item is added to the overlapped set, it will be checked against all of the geometries on every subsequent pass.  Perhaps we can do some bookkeeping to avoid the extract intersection checks.

jwomeara: I just pushed an updated implementation which should be better.  Let me know what you think.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GeoWaveFunctionsDescriptor.java
jwomeara: @ivakegg Added config params to the shard query configuration...

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GeoWaveFunctionsDescriptor.java
jwomeara: @ivakegg This should be much better now with my second revision.

PR #29, opened 2018-02-12
Closes #27, Closes #28: Updated the GeometryNormalizer to support dyn…

### LINKED ISSUE #27: Add GeoWave JexlQueryFunctions to enable Lucene syntax for geo queries

### LINKED ISSUE #28: Update GeometryNormalizer to support normalization of multiple string geometry formats.
At the moment, we only support WKT.  We should update the geometry normalizer so that it can handle additional geometry representations besides WKT.

jwomeara:
…amic geometry string parsing.

Also, added GeoWave JexlQueryFunctions in order to enable support for Lucene formatted GeoWave queries.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/attributes/Geometry.java
+++ b/warehouse/query-core/src/main/java/datawave/query/attributes/Geometry.java
@@ -7,6 +7,7 @@ import com.vividsolutions.jts.io.ParseException;
 import com.vividsolutions.jts.io.WKBReader;
 import com.vividsolutions.jts.io.WKBWriter;
 import datawave.data.normalizer.GeometryNormalizer;
+import datawave.data.normalizer.Normalizer;
 import datawave.query.jexl.DatawaveJexlContext;
 import datawave.query.collections.FunctionalSet;
 import datawave.webservice.query.data.ObjectSizeOf;
@@ -24,17 +25,15 @@ import java.util.Collections;
 public class Geometry extends Attribute<Geometry> implements Serializable {
     private static final long serialVersionUID = 1L;
     
-    private static final GeometryNormalizer NORMALIZER = new GeometryNormalizer();
-    
     private com.vividsolutions.jts.geom.Geometry geometry;
     
     protected Geometry() {
         super(null, true);
     }
     
-    public Geometry(String wellKnownText, Key docKey, boolean toKeep) {
+    public Geometry(String geoString, Key docKey, boolean toKeep) {
         super(docKey, toKeep);
-        setGeometryFromWkt(wellKnownText);
+        setGeometryFromGeoString(geoString);
         validate();
     }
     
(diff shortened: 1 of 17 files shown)


warehouse/query-core/src/main/java/datawave/query/attributes/Geometry.java
```diff
@@ -34,7 +33,7 @@ protected Geometry() {
     
     public Geometry(String wellKnownText, Key docKey, boolean toKeep) {
         super(docKey, toKeep);
-        setGeometryFromWkt(wellKnownText);
+        setGeometryFromGeoString(wellKnownText);
```
ivakegg: change variable name

warehouse/query-core/src/main/java/datawave/query/language/functions/jexl/Geowave.java
```diff
@@ -0,0 +1,264 @@
...
+import datawave.webservice.query.exception.BadRequestQueryException;
+import datawave.webservice.query.exception.DatawaveErrorCode;
+
+import java.text.MessageFormat;
+import java.util.ArrayList;
+import java.util.List;
+
+public interface Geowave {
+    
+    class Contains extends JexlQueryFunction {
```
ivakegg: Lets consolidate some of this code: e.g. make an abstract class in the middle that gives us the validate and toString method implementations that are duplicated in each one of these.

web-services/client/src/main/java/datawave/data/normalizer/GeometryNormalizer.java
```diff
@@ -2,29 +2,22 @@
...
-import mil.nga.giat.geowave.core.store.index.BasicIndexModel;
-import mil.nga.giat.geowave.core.store.index.CustomIdIndex;
-import mil.nga.giat.geowave.core.store.index.PrimaryIndex;
 import org.apache.commons.codec.binary.Hex;
-import org.apache.commons.lang.StringUtils;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.List;
+import java.util.*;
```
ivakegg: By convention we have been expanding these.  We should enforce this somewhere of course so this is on us.  However you can change your IDE to force expansion.

warehouse/query-core/src/main/java/datawave/query/language/functions/jexl/Geowave.java
tomnelson: Can you add interface default methods for validate and toString so we don't need another class?

warehouse/query-core/src/main/java/datawave/query/attributes/Geometry.java
jwomeara: Fixed!

warehouse/query-core/src/main/java/datawave/query/language/functions/jexl/Geowave.java
jwomeara: Fixed!

web-services/client/src/main/java/datawave/data/normalizer/GeometryNormalizer.java
jwomeara: D'oh!  Fixed!

PR #39, opened 2018-02-27
Closes 38 Remove duplicate data type handlers from the configured lis…

friedlou:
…t of data type handlers.

### DIFF
--- a/warehouse/core/src/main/java/datawave/util/StringUtils.java
+++ b/warehouse/core/src/main/java/datawave/util/StringUtils.java
@@ -1,10 +1,6 @@
 package datawave.util;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Iterator;
-import java.util.List;
-import java.util.NoSuchElementException;
+import java.util.*;
 
 /**
  * A class for general String utilities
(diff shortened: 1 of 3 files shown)


warehouse/core/src/main/java/datawave/util/StringUtils.java
```diff
@@ -1,10 +1,6 @@
 package datawave.util;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Iterator;
-import java.util.List;
-import java.util.NoSuchElementException;
+import java.util.*;
```
brianloss: Don't import with wildcards, please.

friedlou: Funny. I did not make that change deliberately. Most probably Intelij did it. I will fix it.

warehouse/core/src/main/java/datawave/util/StringUtils.java
```diff
@@ -429,4 +431,22 @@ static boolean isEscapableLiteral(char c) {
             return values;
         }
     }
+    
+    /**
+     * Remove the strings that are empty
```
brianloss: Documentation of this method is incorrect.

brianloss:
Your commit comment should indicate it "closes" or "fixes" 38, so that the associated issue will automatically be closed when the PR is merged.

friedlou:
done

PR #40, opened 2018-02-28
Update quickstart readme with certificate import steps for chrome and…

medined:
… firefox

Step-by-step instructions to import the certificate into firefox and chrome.


medined:
That makes sense. I was going through the process and my first instinct is to write down steps.

PR #44, opened 2018-03-14
Add support for scores, skips, and same position matching…

pcgrenier:
… for content queries(adjacent, within, phrase)


warehouse/query-core/src/main/java/datawave/query/function/JexlEvaluation.java
```diff
@@ -66,17 +75,54 @@ public boolean isMatched(Object o) {
...
         Object o = script.execute(input.third());
+        Document document = input.second();
         
         if (log.isTraceEnabled()) {
             log.trace("Evaluation of " + query + " against " + input.third() + " returned " + o);
         }
         
         boolean matched = isMatched(o);
         
+        // If content functions have set a term position and term timestamps exist. Filter the term timestamps to matched positions
```
jwomeara: Is there a test out there that validates this logic?

pcgrenier:
will be merged at a later time

PR #51, opened 2018-04-04
Update ingest scripts to use pssh instead of pdsh commands.

ejrgilbert:
Listed are the alternate options I used and why during the refactor:

| pssh | pdsh | Use |
|--------|--------|-------|
| `-h <host_file>` | `-w ^<host_file>` (must be prepended with `^` if reading from file) | Specify list of hosts to connect to |
| `-v -i` | `-d` | Output error messages |
| `-p <concurrent_sessions>` | `-f <simultaneous_commands>` | Number of concurrent commands/sessions to run |
| `-o <filename>` | `1>> <filename>` | Redirect stdout to file |
| `-e <filename>` | `2>> <filename>` | Redirect stderr to file |


ivakegg:
So, can I presume that the title of this pull request is exactly backwards from what these changes actually do?

ejrgilbert:
@ivakegg Yes, you're right. I changed the title, sorry about the confusion.

PR #58, opened 2018-04-16
fixes #61: Updated the QueryPropertyMarker.instanceof() method to work correctly…

### LINKED ISSUE #61: QueryPropertyMarker.instance of is not identifying marked nodes correctly
The instanceof method in QueryPropertyMarker is used to identify nodes that have been marked with a QueryPropertyMarker (like ASTDelayedPredicate), but at the moment, it does not correctly identify marked nodes after they have gone through serialization/deserialization.  At times, it also prematurely identifies the parent of a marked node as the marked node, which can cause problems.

jwomeara:
… in cases where a QueryPropertyMarker node is serialized and deserialized.


jwomeara:
@ivakegg @tomnelson If there are no objections, I plan on merging this in today.

PR #68, opened 2018-04-27
Add source file to event if different from ORIG_FILE

hlgp:


warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/EventMapper.java
```diff
@@ -758,6 +768,16 @@ public FieldNormalizationError(String message, Throwable cause) {
             }
         }
         
+        if (createRawFileName && !value.getRawFileName().isEmpty() && !value.getRawFileName().equals(seqFileName)) {
+            String rawFileName = value.getRawFileName();
+            int index = rawFileName.lastIndexOf('/');
+            if (index >= 0) {
+                rawFileName = rawFileName.substring(index + 1);
```
ivakegg: I feel the setting up of the raw filename to include only the last portion should be done where the rawfilename is set in the event.  However the sequence file name probably should be stripped above as that name is derived here.

PR #90, opened 2018-05-23
Allow for null dates in queries

ejrgilbert:
It'll be easier to see my changes in the test file if you just view the file as a whole rather than as in-line changes. I refactored a bit to make things less copy-and-pasted by moving the creation of test queries/parameters to helper methods.

## NOTE
I have tried unsuccessfully to get the project to build with tests via the `mvn clean install` command for testing before this PR. Every time I try, it hangs after the `datawave.ingest.mapreduce.partition.BalancedShardPartitionerTest`. However, I have successfully run all tests in the module these changes were made in `datawave-ws-query`. This suggests to me that the issue is actually due to configuration rather than from any changes made.

If anyone has any suggestions for how to fix my VM configurations, that would be appreciated! In the meantime, I wanted to get this PR up so the review process could begin.

### DIFF
--- a/web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
+++ b/web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
@@ -423,8 +423,8 @@ public class QueryExecutorBean implements QueryExecutor {
             throwBadRequest(DatawaveErrorCode.INVALID_EXPIRATION_DATE, response);
         }
         
-        // Ensure begin date does not occur after the end date
-        if (qp.getBeginDate().after(qp.getEndDate())) {
+        // Ensure begin date does not occur after the end date (if dates are not null)
+        if ((qp.getBeginDate() != null && qp.getEndDate() != null) && qp.getBeginDate().after(qp.getEndDate())) {
             GenericResponse<String> response = new GenericResponse<>();
             throwBadRequest(DatawaveErrorCode.BEGIN_DATE_AFTER_END_DATE, response);
         }
--- a/web-services/query/src/test/java/datawave/webservice/query/runner/QueryExecutorBeanTest.java
+++ b/web-services/query/src/test/java/datawave/webservice/query/runner/QueryExecutorBeanTest.java
@@ -24,7 +24,6 @@ import java.util.concurrent.atomic.AtomicBoolean;
 
 import javax.ejb.EJBContext;
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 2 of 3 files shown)


web-services/query/src/test/java/datawave/webservice/query/runner/QueryExecutorBeanTest.java
```diff
@@ -435,26 +416,42 @@ public void testListWithPost() {
...
+        final MultivaluedMap<String,String> queryParameters = createNewQueryParameterMap();
+        queryParameters.remove(QueryParameters.QUERY_BEGIN);
+        queryParameters.remove(QueryParameters.QUERY_END);
+        queryParameters.putSingle(QueryParameters.QUERY_BEGIN, null);
+        queryParameters.putSingle(QueryParameters.QUERY_END, null);
+
+        final Thread createQuery = new Thread(() -> {
+            try {
+                bean.createQuery("EventQueryLogic", queryParameters);
+                fail(); // If doesn't throw exception, should fail
```
ivakegg: Will failing here actually fail the test case since it is in a separate thread?

web-services/query/src/test/java/datawave/webservice/query/runner/QueryExecutorBeanTest.java
```diff
@@ -435,26 +416,42 @@ public void testListWithPost() {
...
+            } catch (BadRequestException e) {
+                assertEquals(DatawaveErrorCode.BEGIN_DATE_AFTER_END_DATE.toString(), e.getCause().getMessage());
```
ivakegg: If the dates are null then this should not happen....

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -423,8 +423,8 @@ private QueryData validateQuery(String queryLogicName, MultivaluedMap<String,Str
             throwBadRequest(DatawaveErrorCode.INVALID_EXPIRATION_DATE, response);
         }
         
-        // Ensure begin date does not occur after the end date
-        if (qp.getBeginDate().after(qp.getEndDate())) {
+        // Ensure begin date does not occur after the end date (if dates are not null)
+        if ((qp.getBeginDate() != null && qp.getEndDate() != null) && qp.getBeginDate().after(qp.getEndDate())) {
```
tomnelson: I think you can leave out the extra parens around the 2 null tests

PR #99, opened 2018-06-07
Updates to allow age off configs to inherit from common file

bbux-atg:
Updated FileRuleWatcher to allow the inheritance and merging of rules

Updated TokenizingFilterBase to include a merge mode that allows ttl values
later in the config to overwrite the ones earlier one, while maintaining the
priority

Added FileRuleLoadContentsMergeFiltersTest to fully test merge of configs and
the application of the rules against data.

### DIFF
--- a/warehouse/core/src/test/resources/test-root-rules.xml
+++ b/warehouse/core/src/test/resources/test-root-rules.xml
@@ -1,0 +1,16 @@
+<ageoffConfiguration>
+	<rules>
+		<rule>
+			<filterClass>datawave.ingest.util.cache.watch.TestTrieFilter
+			</filterClass>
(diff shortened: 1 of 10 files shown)


bbux-atg:
@ivakegg Would you be able to look at this, or ask someone else to?

warehouse/core/src/test/resources/test-root-rules.xml
```diff
@@ -0,0 +1,26 @@
+<ageoffConfiguration>
```
bbux-atg: @brianloss @ivakegg  This is an example of a root config that would contain all of the core configs and rules, test-customized-rules.xml is the example of the child that overrides some of these

PR #105, opened 2018-06-15
Fixes #103 - Updates for ShardTableConfigHelper: mitigate type safety…

### LINKED ISSUE #103: ShardTableConfigHelper hardcodes ShardedTableTabletBalancer as the TabletBalancer
The ShardTableConfigHelper class hardcodes ShardedTableTabletBalancer as the TabletBalancer implementation, via static final string to reference the FQCN.
```
 protected static final String SHARDED_TABLET_BALANCER_CLASS = "datawave.ingest.table.balancer.ShardedTableTabletBalancer";
```
This is problematic for several reasons. E.g.,....

1) There's no compile-time validation of the String-referenced class name, so a future change to the actual class name will result in a late, runtime error within Accumulo, if the update to this string happens to be overlooked.

2) There's currently no way to override the TabletBalancer implementation for a given table until after the table has already been created, i.e., *AFTER* ingest has been run and data has likely already been loaded. In some cases, ShardedTableTabletBalancer can actually be observed to yield poor distribution across available tservers, particularly when the user's data tends to be highly concentrated within a small number of shard dates and/or when pre-computing table splits is impractical for the given use case. In these cases, Accumulo's default balancer can result in better overall performance

Solution:

1) Use Class<ShardedTableTabletBalancer> reference instead to get the FQCN at runtime.

2) Keep ShardedTableTabletBalancer as the default, but allow it to be overridden via shard-ingest-config.xml, if needed

keith-ratcliffe:
… concerns; add config option for shard table balancer; improve code readability


warehouse/ingest-core/pom.xml
```diff
@@ -18,6 +18,11 @@
             <groupId>com.timgroup</groupId>
             <artifactId>java-statsd-client</artifactId>
         </dependency>
+        <dependency>
+            <groupId>datawave</groupId>
```
ivakegg: The group id may be different now (gov.nsa.datawave?)

ivakegg: resolve conflicts please

keith-ratcliffe: Conflict resolved

keith-ratcliffe:
@ivakegg , @brianloss, it's been a while since I made these changes, but looking back now I'm wondering if it wouldn't be better to relocate the ShardedTableTabletBalancer.java from the balancers jar to another jar as part of this PR. That could alleviate the need for the updates here to both ingest-core/pom.xml and ingest-libs.sh

E.g., the balancers module only contains this one TabletBalancer impl and associated tests (along with an unused ExtentParser interface).

How would you guys feel about me moving everything in the balancers module to datawave-core and then just removing balancers altogether?

ivakegg:
I think we should leave the balancers in a separate package for now.  The changes required in the pom and ingest-libs.sh is not enough for me to thing it should be merged back in.

PR #111, opened 2018-06-21
Added updates to help validate shard splits

bbux-atg:
I am not a fan of the ShardedTableMapFile class.  It is overly complicated and should be refactored and simplified.  In spite of that, these changes should help mitigate problems when shards have not be created or balanced before the ingest jobs launch.

Note: Not sure if unbalanced shard splits is an easily recoverable situation.  It may be necessary to split up the existence and balance checks for the shard splits.

Credit to @smonaem and @matthpetterson for reference implementations that were borrowed from

Updated ShardedTableMapFile.setupFile method to take an additional boolean
argument specifying whether validation of the generated shard to tablet mappings
should be validated for a given date range

Validation is turned off by default in IngestJob unless the:
- shardedMap.validation.enabled
property is set to true

Default is to validate the last two days of shard table splits.  The range can be
overridden using the:
- shards.balanced.days.to.verify
property, where 1 would mean only validate today, 2 yeterday and today, etc.

Validation is dependent on the ShardIdFactory class to inform us of the number
of shards that should exist for a given day.  Any deviation from this number,
when validation is turned on, will result in an IllegalStateException

Validation that the shards are balanced, i.e. have been spread out to unique
tservers via custom balancer, is also performed when validation is on. An
unbalanced mapping of shard splits will also result in an IllegalStateException.

### DIFF
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
@@ -787,8 +787,10 @@ public class IngestJob implements Tool {
             conf.set("split.work.dir", conf.get("ingest.work.dir.qualified"));
         }
         conf.setInt("splits.num.reduce", this.reduceTasks);
+        // want validation turned off by default
+        boolean validateShardTableMapFiles = conf.getBoolean(ShardedTableMapFile.SHARD_VALIDATION_ENABLED, false);
         // used by the output formatter and the sharded partitioner
-        ShardedTableMapFile.setupFile(conf);
+        ShardedTableMapFile.setupFile(conf, validateShardTableMapFiles);
         
         conf.setInt(MultiRFileOutputFormatter.EVENT_PARTITION_COUNT, this.reduceTasks * 2);
         configureMultiRFileOutputFormatter(conf, compressionType, compressionTableBlackList, maxRFileEntries, maxRFileSize, generateMapFileRowKeys);
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
@@ -3,7 +3,9 @@ package datawave.ingest.mapreduce.job;
 import datawave.ingest.data.config.ConfigurationHelper;
 import datawave.ingest.data.config.ingest.AccumuloHelper;
 import datawave.ingest.mapreduce.handler.shard.ShardedDataTypeHandler;
+import datawave.ingest.mapreduce.handler.shard.ShardIdFactory;
 import datawave.util.StringUtils;
+import datawave.util.time.DateHelper;
 import org.apache.accumulo.core.client.*;
 import org.apache.accumulo.core.client.impl.ClientContext;
 import org.apache.accumulo.core.conf.AccumuloConfiguration;
@@ -11,6 +13,7 @@ import org.apache.accumulo.core.data.impl.KeyExtent;
 import org.apache.accumulo.core.metadata.MetadataServicer;
 import org.apache.accumulo.core.client.impl.Credentials;
 import org.apache.accumulo.fate.util.UtilWaitThread;
+import org.apache.commons.lang.time.DateUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.*;
 import org.apache.hadoop.io.*;
@@ -19,22 +22,28 @@ import org.apache.log4j.Logger;
 import java.io.IOException;
 import java.net.URISyntaxException;
 import java.util.*;
+import java.util.Map.Entry;
 
 /**
  * Extracted from IngestJob
  */
 public class ShardedTableMapFile {
+    public static final String SHARDS_BALANCED_DAYS_TO_VERIFY = "shards.balanced.days.to.verify";
     private static final String PREFIX = ShardedTableMapFile.class.getName();
+    private static final int MAX_RETRY_ATTEMPTS = 10;
+    
+    private static final Logger log = Logger.getLogger(ShardedTableMapFile.class);
+    
     public static final String TABLE_NAMES = "job.table.names";
     public static final String SHARD_TSERVER_MAP_FILE = PREFIX + ".shardTServerMapFile";
     public static final String SPLIT_WORK_DIR = "split.work.dir";
     
     public static final String CONFIGURED_SHARDED_TABLE_NAMES = ShardedDataTypeHandler.SHARDED_TNAMES + ".configured";
-    private static final Logger log = Logger.getLogger(ShardedTableMapFile.class);
     public static final String SHARDED_MAP_FILE_PATHS_RAW = "shardedMap.file.paths.raw";
+    public static final String SHARD_VALIDATION_ENABLED = "shardedMap.validation.enabled";
     
-    public static void setupFile(Configuration conf) throws IOException, URISyntaxException, AccumuloSecurityException, AccumuloException {
-        Map<String,Path> map = loadMap(conf);
+    public static void setupFile(Configuration conf, boolean doValidation) throws IOException, URISyntaxException, AccumuloSecurityException, AccumuloException {
+        Map<String,Path> map = loadMap(conf, doValidation);
         if (null == map) {
             log.fatal("Receieved a null mapping of sharded tables to split files, exiting...");
             throw new RuntimeException("Receieved a null mapping of sharded tables to split files, exiting...");
@@ -58,6 +67,7 @@ public class ShardedTableMapFile {
         
         Text shardID = new Text();
         Text location = new Text();
+        
         try {
             while (reader.next(shardID, location)) {
                 locations.put(new Text(shardID), location.toString());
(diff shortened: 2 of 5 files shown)


warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
```diff
@@ -787,8 +787,10 @@ protected void configureBulkPartitionerAndOutputFormatter(Job job, AccumuloHelpe
             conf.set("split.work.dir", conf.get("ingest.work.dir.qualified"));
         }
         conf.setInt("splits.num.reduce", this.reduceTasks);
+        // want validation turned off by default
+        boolean validateShardTableMapFiles = conf.getBoolean(ShardedTableMapFile.SHARD_VALIDATION_ENABLED, false);
```
matthpeterson: It's just a matter of preference, but I think I would pull the boolean out of the conf in ShardedTableMapFile because the conf is already provided.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
```diff
@@ -204,14 +319,26 @@ public static Path createShardedMapFile(Logger log, Configuration conf, Path wor
...
-                MetadataServicer servicer = MetadataServicer.forTableName(
-                                new ClientContext(instance, credentials, AccumuloConfiguration.getDefaultConfiguration()), shardedTableName);
-                servicer.getTabletLocations(locations);
-                break;
+                ClientContext context = new ClientContext(instance, credentials, AccumuloConfiguration.getDefaultConfiguration());
+                // if table does not exist don't want to catch the errors and end up in infinite loop
+                Map<String,String> tableIdMap = context.getConnector().tableOperations().tableIdMap();
+                boolean tableExists = tableIdMap.containsKey(shardedTableName);
+                if (!tableExists) {
+                    log.error("Table " + shardedTableName + " not found, skipping split locations for missing table");
```
matthpeterson: nice improvement

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
```diff
@@ -204,14 +319,26 @@ public static Path createShardedMapFile(Logger log, Configuration conf, Path wor
...
+                } else {
+                    MetadataServicer servicer = MetadataServicer.forTableName(context, shardedTableName);
+                    servicer.getTabletLocations(locations);
+                }
+                // made it here, no errors so break out
+                keepRetrying = false;
+                ;
```
matthpeterson: stray line

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
```diff
@@ -204,14 +319,26 @@ public static Path createShardedMapFile(Logger log, Configuration conf, Path wor
...
+                Map<String,String> tableIdMap = context.getConnector().tableOperations().tableIdMap();
+                boolean tableExists = tableIdMap.containsKey(shardedTableName);
+                if (!tableExists) {
+                    log.error("Table " + shardedTableName + " not found, skipping split locations for missing table");
+                } else {
+                    MetadataServicer servicer = MetadataServicer.forTableName(context, shardedTableName);
+                    servicer.getTabletLocations(locations);
+                }
+                // made it here, no errors so break out
+                keepRetrying = false;
```
matthpeterson: if the table doesn't exist, do we want to proceed without failing?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
bbux-atg: That makes sense, much cleaner

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
```diff
@@ -68,6 +78,102 @@ public static void setupFile(Configuration conf) throws IOException, URISyntaxEx
...
+            boolean shardsAreBalanced = shardsAreBalanced(shardIdToLocation, datePrefix);
+            if (!shardsAreBalanced) {
+                log.warn("Shards for " + datePrefix + " for table " + tableName + " are not balanced!");
+                isValid = false;
+            }
+        }
+        if (!isValid) {
+            throw new IllegalStateException("All of today's shards have not been created.  Run bin/ingest/create_shards_since.sh "
+                            + "yyyymmdd, substitute yyyymmdd with today's date.  Ensure that the normal mechanism "
+                            + "for creating shards is running - e.g, crontab");
```
matthpeterson: It may also be the case that the shards are just not balanced.  I believe this will also run for all sharded tables, like errorShard, so it might be useful to have the table names that had problems in the message.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
```diff
@@ -68,6 +78,102 @@ public static void setupFile(Configuration conf) throws IOException, URISyntaxEx
...
+            return false;
+        }
+        for (int i = 0; i < prefixBytes.length; i++) {
+            if (prefixBytes[i] != keyBytes[i]) {
+                return false;
+            }
+        }
+        // at this point didn't fail match, so should be good
+        return true;
+    }
```
matthpeterson: good work

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
bbux-atg: Makes sense, should we fail on first bad table or aggregate all the failures and log everything that is bad?

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/job/ShardedTableMapFileTest.java
```diff
@@ -188,5 +226,109 @@ public void testWriteSplitsFileExistingPathMultipleKeyExtents() throws Exception
...
+    }
+    
+    @Test(expected = IllegalStateException.class)
+    public void testMissingAllOfTodaysSplits() throws Exception {
+        String tableName = "missingTodaysSplits";
+        SortedMap<Text,String> splits = simulateMissingSplitsForDay(0, tableName);
+        createSplitsFile(splits, conf, splits.size(), tableName);
+        Map<Text,String> locations = ShardedTableMapFile.getShardIdToLocations(conf, tableName);
+        // three days of splits, today should be invalid, which makes the rest bad too
+        ShardedTableMapFile.validateShardIdLocations(conf, tableName, 0, locations);
```
matthpeterson: if ,0, doesn't fail but ,1, does then this test still passes

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/job/ShardedTableMapFileTest.java
```diff
@@ -188,5 +226,109 @@ public void testWriteSplitsFileExistingPathMultipleKeyExtents() throws Exception
...
+            String day = DateHelper.format(now - (daysAgo * DateUtils.MILLIS_PER_DAY));
+            for (int currShard = 0; currShard < SHARDS_PER_DAY; currShard++) {
+                locations.put(new Text(day + "_" + currShard), Integer.toString(tserverId++));
+            }
+        }
+        return locations;
+    }
+    
+    private static String formatDay(int daysBack) {
+        return DateHelper.format(System.currentTimeMillis() - (daysBack * DateUtils.MILLIS_PER_DAY));
```
matthpeterson: Nice tests, bro.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
matthpeterson: right, good point.  the validation will be enough.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
matthpeterson: I am fine with the former but I like how you logged each that was wrong and think it would be helpful to include that in the exception message too.  It might seem like overkill, and maybe it is, given that you're writing to the log, but I'm hoping to lower the bar on identifying the solution for folks that aren't as familiar with the system.

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/job/ShardedTableMapFileTest.java
bbux-atg: yep, adding a fail() call after each of these

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
```diff
@@ -68,6 +93,102 @@ public static void setupFile(Configuration conf) throws IOException, URISyntaxEx
...
+            boolean shardsAreBalanced = shardsAreBalanced(shardIdToLocation, datePrefix);
+            if (!shardsAreBalanced) {
+                log.warn("Shards for " + datePrefix + " for table " + tableName + " are not balanced!");
+                isValid = false;
+            }
+        }
+        if (!isValid) {
+            throw new IllegalStateException("All of today's shards have not been created or are not balanced for " + tableName +"."
+                            + "Run bin/ingest/create_shards_since.sh yyyymmdd, substitute yyyymmdd with today's date. "
+                            + "Ensure that the normal mechanism for creating shards is running - e.g, crontab");
```
keith-ratcliffe: Couldn't the IllegalStateException message here be misleading, since technically the exception might be thrown based on the non-existent and/or unbalanced shards from some day other than today's? That is, depending on 'daysToVerify'.

And maintenance-wise, instead of referencing a specific script name/parameters/crontab/etc (all of which are environment-specific and subject to change), it may be better just to say something like "Shard validation failed for *tableName*. Please ensure that shards have been generated. Check log for details about the dates in question". Or something similar. And maybe also change the log messages above to ERROR rather than WARN, since we're saying that this represents an invalid system state

All in all though, everything else LGTM. The validation is a cool feature. I built, deployed, and tested your branch locally with no issues.

bbux-atg: That makes sense, updated with recommended changes

keith-ratcliffe: @ivakegg, @brianloss, this looks good to me. Do you guys see any reason not to go ahead and merge this?

PR #119, opened 2018-06-28
Make default date type configurable

hlgp:


warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -165,6 +165,8 @@
      */
     public static final String DATATYPE_FILTER_SET = "datatype.filter.set";
     
+    public static final String DEFAULT_DATE_TYPE_NAME = "default.date.type.name";
```
ivakegg: What is the point of this constant?

hlgp: leftover from previous experiment.  removing.

PR #121, opened 2018-06-29
Fixing missing else in RangeSplit

roshanp:
I believe there should be an `else` here


ivakegg:
Good catch. Thank you!

PR #125, opened 2018-07-05
Fixes #124 - Updated QueryExecutorBean to ensure that duplicate query…

### LINKED ISSUE #124: Query reset fails with error: Required parameter auditUserDN only accepts one value
It looks like the query reset logic in QueryExecutorBean is creating a multivalued map of query parameters from the settings in the running query (which already includes userDN), and then manually adds userDN to the multivaluedmap.  Then, during audit parameter validation, it complains that there are multiple values for auditUserDN.

jwomeara:
… parameters are not created during a reset.


web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -1028,10 +1028,18 @@ public VoidResponse reset(@Required("id") @PathParam("id") String id) {
                 AuditType auditType = query.getLogic().getAuditType(query.getSettings());
                 
                 MultivaluedMap<String,String> queryParameters = query.getSettings().toMap();
-                queryParameters.add("logicClass", query.getLogic().getLogicName());
-                queryParameters.add(AuditParameters.USER_DN, query.getSettings().getUserDN());
                 queryParameters.add(AuditParameters.QUERY_AUDIT_TYPE, auditType.name());
```
ivakegg: Lets replace the value in the map if it exists for QUERY_AUDIT_TYPE

PR #131, opened 2018-07-12
Added PasswordConverter to allow passwords from environment

bbux-atg:
Created PasswordConverter class to allow users to specify passwords
using an env: prefix.  This will indicate that an ENVIRONMENT variable
should be used to obtain the password.

This is useful in order to prevent passwords from showing up in the process list

Updated some core classes to make use of this convention.

### DIFF
--- a/warehouse/core/src/test/java/datawave/util/cli/PasswordConverterTest.java
+++ b/warehouse/core/src/test/java/datawave/util/cli/PasswordConverterTest.java
@@ -5,8 +5,6 @@ import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.nullValue;
 import static org.junit.Assert.assertThat;
 
-import java.util.Random;
-
 import org.junit.Before;
 import org.junit.Test;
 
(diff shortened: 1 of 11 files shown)


warehouse/core/src/test/java/datawave/util/cli/PasswordConverterTest.java
```diff
@@ -0,0 +1,65 @@
...
+        // prefix with no var defined should return null
+        argv[1] = "env:";
+        new JCommander(password).parse(argv);
+        assertThat(password.password, is(nullValue()));
+    }
+    
+    @Test
+    public void testConvertEnvPrefixedUndefinedEnvironmentVar() {
+        // prefix with env var not defined should return null
+        argv[1] = "env:HighlyUnlikelyThisWillBeDefined" + new Random().nextInt();
```
brianloss: You could assert that HighlyUnlikelyThisWillBeDefined actually doesn't exist in the environment.

warehouse/ingest-core/src/main/java/datawave/ingest/util/GenerateMultipleNumShardsCacheFile.java
```diff
@@ -24,6 +24,7 @@
     public static final String CONFIG_DIRECTORY_LOCATION_OVERRIDE = "cd";
     public static final String CONFIG_SUFFIEX_OVERRIDE = "cs";
     
+    @SuppressWarnings("static-access")
```
brianloss: This caused me to look at why you were adding this and didn't fix the underlying static access problem. This code is dangerous. It appears OptionBuilder sets static variables. I think we're safe since we're always setting the same flags on the option, but if we weren't, we'd have a bug since we're not calling OptionBuilder.reset between each use. Not suggesting you need to change anything for this PR, but just noticed that.

warehouse/core/src/test/java/datawave/util/cli/PasswordConverterTest.java
bbux-atg: makes sense, adding change

bbux-atg:
@brianloss any more changes needed, could we get this merged in?

brianloss:
No, looks good. Sorry, I had received an e-mail that indicated @ivakegg merged it. Weird.

PR #139, opened 2018-07-18
use builder pattern for IndexIterator and Ivarator classes

tomnelson:
Changes to be committed:
	modified:   warehouse/query-core/src/main/java/datawave/query/ancestor/AncestorIndexIterator.java
	modified:   warehouse/query-core/src/main/java/datawave/query/ancestor/AncestorIndexIteratorBuilder.java
	modified:   warehouse/query-core/src/main/java/datawave/query/iterator/builder/CardinalityIteratorBuilder.java
	modified:   warehouse/query-core/src/main/java/datawave/query/iterator/builder/IndexIteratorBuilder.java
	modified:   warehouse/query-core/src/main/java/datawave/query/iterator/logic/IndexIterator.java
	modified:   warehouse/query-core/src/main/java/datawave/query/tld/TLDIndexIterator.java
	modified:   warehouse/query-core/src/main/java/datawave/query/tld/TLDIndexIteratorBuilder.java

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
+++ b/warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
@@ -59,6 +59,121 @@ import java.util.concurrent.atomic.AtomicLong;
  * 
  */
 public abstract class DatawaveFieldIndexCachingIteratorJexl extends WrappingIterator implements CompositePredicateFilterer {
+    
+    public static abstract class Builder<B extends Builder<B>> {
+        private Text fieldName;
+        protected Text fieldValue;
+        private Predicate<Key> datatypeFilter;
+        private TimeFilter timeFilter;
+        private boolean negated;
+        private PartialKey returnKeyType = DEFAULT_RETURN_KEY_TYPE;
+        private int maxRangeSplit = 11;
+        private FileSystem fs;
+        private Path uniqueDir;
+        private QueryLock queryLock;
+        private boolean allowDirReuse;
+        private long scanThreshold = 10000;
+        private int hdfsBackedSetBufferSize = 10000;
+        private int maxOpenFiles = 100;
+        private boolean sortedUIDs = true;
+        protected QuerySpanCollector querySpanCollector = null;
+        protected volatile boolean collectTimingDetails = false;
+        private volatile long scanTimeout = 1000L * 60 * 60;
+        private Map<String,Map<String,CompositePredicateFilter>> compositePredicateFilters;
+        
+        @SuppressWarnings("unchecked")
+        protected B self() {
+            return (B) this;
+        }
+        
+        public B withFieldName(Text fieldName) {
+            this.fieldName = fieldName;
+            return self();
+        }
+        
+        public B withFieldValue(Text fieldValue) {
+            this.fieldValue = fieldValue;
+            return self();
+        }
+        
+        public B withTimeFilter(TimeFilter timeFilter) {
+            this.timeFilter = timeFilter;
+            return self();
+        }
+        
+        public B withDatatypeFilter(Predicate datatypeFilter) {
+            this.datatypeFilter = datatypeFilter;
+            return self();
+        }
+        
+        public B negated(boolean negated) {
+            this.negated = negated;
+            return self();
+        }
+        
+        public B withScanThreshold(long scanThreshold) {
+            this.scanThreshold = scanThreshold;
+            return self();
+        }
+        
+        public B withScanTimeout(long scanTimeout) {
+            this.scanTimeout = scanTimeout;
+            return self();
+        }
+        
+        public B withHdfsBackedSetBufferSize(int hdfsBackedSetBufferSize) {
+            this.hdfsBackedSetBufferSize = hdfsBackedSetBufferSize;
+            return self();
+        }
+        
+        public B withMaxRangeSplit(int maxRangeSplit) {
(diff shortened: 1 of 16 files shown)


warehouse/query-core/src/main/java/datawave/query/ancestor/AncestorIndexIterator.java
```diff
@@ -1,32 +1,32 @@
...
+        protected Builder(Text field, Text value, SortedKeyValueIterator<Key,Value> source) {
+            super(field, value, source);
+        }
+        
+        public T build() {
+            return (T) new AncestorIndexIterator(this);
+        }
+    }
+    
+    public static Builder<?,?> builder(Text field, Text value, SortedKeyValueIterator<Key,Value> source) {
```
brianloss: You know things about the builder you are returning here. Can't you make the signature of this be something like the following?
```
public static <T extends AncestorIndexIterator,B extends Builder<T,B>> Builder<T,B> builder(...
```
If you return something that is qualified, the compiler can make assumptions about the return type, and I think you might be able to avoid having to do a cast everywhere you use this function.

tomnelson: If I put this in IndexIterator:

```
    public static <T extends IndexIterator,B extends Builder<T,B>> Builder<T,B> builder(Text field, Text value, SortedKeyValueIterator<Key,Value> source) {
        return new Builder(field, value, source);
    }
```
then doing this in (the subclass) AncestorIndexIterator 

```
    public static <T extends AncestorIndexIterator,B extends Builder<T,B>> Builder<T,B> builder(Text field, Text value, SortedKeyValueIterator<Key,Value> source) {
        return new Builder(field, value, source);
    }

```


clashes and causes a compiler error:

`
[ERROR] /home/tom/projects/datawave/warehouse/query-core/src/main/java/datawave/query/ancestor/AncestorIndexIterator.java:[24,89] name clash: <T,B>builder(org.apache.hadoop.io.Text,org.apache.hadoop.io.Text,org.apache.accumulo.core.iterators.SortedKeyValueIterator<org.apache.accumulo.core.data.Key,org.apache.accumulo.core.data.Value>) in datawave.query.ancestor.AncestorIndexIterator and <T,B>builder(org.apache.hadoop.io.Text,org.apache.hadoop.io.Text,org.apache.accumulo.core.iterators.SortedKeyValueIterator<org.apache.accumulo.core.data.Key,org.apache.accumulo.core.data.Value>) in datawave.query.iterator.logic.IndexIterator have the same erasure, yet neither hides the other
`

The way they are now, there is no casting required when they are used.... you may have seen this 

```
        // @TODO: remove redundant type cast to check builder is building the subtype
        return (AncestorIndexIterator) AncestorIndexIterator.builder(field, value, source).withTimeFilter(timeFilter).withTypeMetadata(typeMetadata)
                        .shouldBuildDocument(buildDocument).withDatatypeFilter(datatypeFilter).withAggregation(aggregator).build();

```



and not realized that the typecast is unnecessary and is there temporarily only to make sure I don't break it while trying out various things. I plan to remove the typecasts (and instead make a unit test to confirm that the builder syntax is solid.

brianloss: Yeah, it seems the self-referential generics really complicate things. The static builder method adds confusion for someone trying to read the code and buys only a small readability improvement. You might consider removing it (and use new Foo.Builder()... instead), though I don't really have a strong opinion about it. Another couple nits... change if you want:
- The build methods could return the concrete type rather than a B
- You could add a "self()" method that does the cast of this to a B and avoid having to suppress or live with warnings on every "with*" method.

tomnelson:
I updated this to include the Ivarator classes and to isolate and reduce the number of SuppressWarnings annotations (using a self() method)

jwomeara:
Looks good to me!

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -59,6 +59,121 @@
...
+        private int hdfsBackedSetBufferSize = 10000;
+        private int maxOpenFiles = 100;
+        private boolean sortedUIDs = true;
+        protected QuerySpanCollector querySpanCollector = null;
+        protected volatile boolean collectTimingDetails = false;
+        private volatile long scanTimeout = 1000L * 60 * 60;
+        private Map<String,Map<String,CompositePredicateFilter>> compositePredicateFilters;
+        
+        @SuppressWarnings("unchecked")
+        B self() {
```
brianloss: This is package protected. Is there a reason we'd want to prevent extension into a new package?

tomnelson: good catch. i think it should be marked 'protected' so only the Builder subclasses see it

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexListIteratorJexl.java
```diff
@@ -45,67 +39,50 @@
...
+            this.fst = fst;
+            return self();
+        }
+        
+        public B withValues(Collection<String> values) {
+            this.values = new ArrayList<>(values);
+            return self();
+        }
+        
+        Builder() {
```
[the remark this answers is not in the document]
tomnelson: I'll remove those unneeded CTORs

PR #149, opened 2018-07-26
add function to carry queryOptions into the ShardQueryConfiguration

tomnelson:
Changes to be committed:
	modified:   warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
	modified:   warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctionsDescriptor.java
	new file:   warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryOptionsFromQueryVisitor.java
	modified:   warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
	new file:   warehouse/query-core/src/main/java/datawave/query/planner/QueryOptionsSwitch.java
	modified:   warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
	modified:   warehouse/query-core/src/test/java/datawave/query/CompositeFunctionsTest.java
	modified:   warehouse/query-core/src/test/java/datawave/query/function/HitsAreAlwaysIncludedTest.java

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
@@ -1185,6 +1185,18 @@ public class EvaluationPhaseFilterFunctions {
     }
     
     /**
+     * This is the query function to pass query option pairs to the ShardQueryConfiguration. It always evaluates to 'true' so that it does not need to be
+     * removed from the query after the args are parsed to the ShardQueryConfiguration
+     * 
+     * @param args
+     *            ignored here. used to pass key/value pairs to the query options
+     * @return true
+     */
+    public static boolean options(String[] args) {
+        return true;
+    }
+    
+    /**
      * A special format denoting the time since epoch format
      */
     public static final String TIME_SINCE_EPOCH_FORMAT = "e";
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
@@ -1185,18 +1185,6 @@ public class EvaluationPhaseFilterFunctions {
     }
     
     /**
-     * This is the query function to pass query option pairs to the ShardQueryConfiguration. It always evaluates to 'true' so that it does not need to be
-     * removed from the query after the args are parsed to the ShardQueryConfiguration
-     * 
-     * @param args
-     *            ignored here. used to pass key/value pairs to the query options
-     * @return true
-     */
-    public static boolean options(String[] args) {
-        return true;
-    }
-    
-    /**
      * A special format denoting the time since epoch format
      */
     public static final String TIME_SINCE_EPOCH_FORMAT = "e";
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctionsDescriptor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctionsDescriptor.java
@@ -41,7 +41,7 @@ public class EvaluationPhaseFilterFunctionsDescriptor implements JexlFunctionArg
         private static final ImmutableSet<String> dateBetweenFunctions = ImmutableSet.of("betweenDates", "betweenLoadDates");
         private static final String MATCHCOUNTOF = "matchesAtLeastCountOf";
         private static final String TIMEFUNCTION = "timeFunction";
-        
+        private static final String OPTIONS_FUNCTION = "options";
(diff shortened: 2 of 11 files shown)


warehouse/query-core/src/main/java/datawave/query/planner/QueryOptionsSwitch.java
```diff
@@ -0,0 +1,37 @@
...
+                    String[] lf = StringUtils.split(value, Constants.PARAM_VALUE_SEP);
+                    config.setLimitFields(Sets.newHashSet(lf));
+                    break;
+                case QueryOptions.TYPE_METADATA_IN_HDFS:
+                    config.setTypeMetadataInHdfs(Boolean.parseBoolean(value));
+                    break;
+                case QueryOptions.DISABLE_EVALUATION:
+                    config.setDisableEvaluation(Boolean.parseBoolean(value));
+                    break;
+            }
```
tomnelson: far from an exhaustive list... please comment if you have a favorite that should be added

warehouse/query-core/src/main/java/datawave/query/planner/QueryOptionsSwitch.java
```diff
@@ -0,0 +1,41 @@
...
+                    config.setIncludeGroupingContext(Boolean.parseBoolean(value));
+                    break;
+                case QueryOptions.HIT_LIST:
+                    config.setHitList(Boolean.parseBoolean(value));
+                    break;
+                case QueryOptions.LIMIT_FIELDS:
+                    String[] lf = StringUtils.split(value, Constants.PARAM_VALUE_SEP);
+                    config.setLimitFields(Sets.newHashSet(lf));
+                    break;
+                case QueryOptions.TYPE_METADATA_IN_HDFS:
```
ivakegg: users should not be able to override TYPE_METADATA_IN_HDFS or DISABLE_EVALUATION.  Perhaps if the user is an admin.....

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctionsDescriptor.java
```diff
@@ -41,7 +41,7 @@
         private static final ImmutableSet<String> dateBetweenFunctions = ImmutableSet.of("betweenDates", "betweenLoadDates");
         private static final String MATCHCOUNTOF = "matchesAtLeastCountOf";
         private static final String TIMEFUNCTION = "timeFunction";
-        
+        private static final String OPTIONS_FUNCTION = "options";
```
ivakegg: Need to update getFields functions to return an empty list for the options function

tomnelson: The arguments for filter:options are all of StringLiteral type so the call to getIdentifiers (called when you get fields) correctly returns an empty collection.

warehouse/query-core/src/main/java/datawave/query/planner/QueryOptionsSwitch.java
tomnelson: I agree about DISABLE_EVALUATION because it could cause incorrect results.
I believe there is a requirement (perhaps not a permanent requirement) to allow users to select the type.metadata.in.hdfs optimization be turned on or off. Query results are expected to be the same either way.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
```diff
@@ -1184,6 +1184,18 @@ public static boolean excludeRegex(Iterable<?> values, String regex) {
...
+    /**
+     * This is the query function to pass query option pairs to the ShardQueryConfiguration. It always evaluates to 'true' so that it does not need to be
+     * removed from the query after the args are parsed to the ShardQueryConfiguration
+     * 
+     * @param args
+     *            ignored here. used to pass key/value pairs to the query options
+     * @return true
+     */
+    public static boolean options(String[] args) {
+        return true;
```
ivakegg: This seems dangerous to me in that if this is embedded with an OR, then we would return everything that gets evaluated.  Perhaps we need to actually remove this function from the jexl tree during the planning stage.

warehouse/query-core/src/main/java/datawave/query/planner/QueryOptionsSwitch.java
ivakegg: ok

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
tomnelson: changed visitor to return 'null' for this function, which lets the rebuilding visitor remove it from the tree

PR #162, opened 2018-08-04
UNIQUE function

ivakegg:
This is a reworking of the DocumentTransformer to use DocumentTransform classes to enable an implementation of a Unique functionality.  Unique in this context means for a set of fields, returning only one document for each set of unique values.  This also required reworking the Group functionality for use with the DocumentTransform mechanism.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/tables/DeduppingIterator.java
+++ b/warehouse/query-core/src/main/java/datawave/query/tables/DeduppingIterator.java
@@ -15,14 +15,14 @@ import org.apache.accumulo.core.data.Value;
 /**
  * Created on 9/13/16.
  */
-class DedupingIterator implements Iterator<Entry<Key,Value>> {
+class DeduppingIterator implements Iterator<Entry<Key,Value>> {
     private Iterator<Entry<Key,Value>> delegate;
     private Entry<Key,Value> next;
     private BloomFilter<byte[]> bloom = null;
     private HashSet<ByteSequence> seen;
     private final boolean DEBUG = true;
     
-    public DedupingIterator(Iterator<Entry<Key,Value>> iterator) {
+    public DeduppingIterator(Iterator<Entry<Key,Value>> iterator) {
         this.delegate = iterator;
         this.bloom = BloomFilter.create(new ByteFunnel(), 500000, 1e-15);
         if (DEBUG) {
--- a/warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
+++ b/warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
@@ -41,7 +41,6 @@ import datawave.query.scheduler.SequentialScheduler;
 import datawave.query.tables.stats.ScanSessionStats;
 import datawave.query.transformer.DocumentTransformer;
 import datawave.query.transformer.EventQueryDataDecoratorTransformer;
-import datawave.query.transformer.GroupingDocumentTransformer;
 import datawave.query.transformer.GroupingTransform;
 import datawave.query.transformer.UniqueTransform;
 import datawave.query.util.DateIndexHelper;
@@ -830,7 +829,7 @@ public class ShardQueryLogic extends BaseQueryLogic<Entry<Key,Value>> {
         this.iterator = this.scheduler.iterator();
         
         if (!config.isSortedUIDs()) {
-            this.iterator = new DeduppingIterator(this.iterator);
+            this.iterator = new DedupingIterator(this.iterator);
         }
         
         stopwatch.stop();
--- a/warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
+++ b/warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
@@ -42,6 +42,8 @@ import datawave.query.tables.stats.ScanSessionStats;
 import datawave.query.transformer.DocumentTransformer;
 import datawave.query.transformer.EventQueryDataDecoratorTransformer;
 import datawave.query.transformer.GroupingDocumentTransformer;
+import datawave.query.transformer.GroupingTransform;
+import datawave.query.transformer.UniqueTransform;
 import datawave.query.util.DateIndexHelper;
 import datawave.query.util.DateIndexHelperFactory;
 import datawave.query.util.MetadataHelper;
@@ -160,23 +162,6 @@ public class ShardQueryLogic extends BaseQueryLogic<Entry<Key,Value>> {
     
     protected static final Logger log = ThreadConfigurableLogger.getLogger(ShardQueryLogic.class);
     
-    /**
-     * Set of datatypes to limit the query to.
-     */
-    public static final String DATATYPE_FILTER_SET = "datatype.filter.set";
-    
-    public static final String LIMIT_FIELDS = "limit.fields";
-    
-    public static final String GROUP_FIELDS = "group.fields";
-    
-    public static final String HIT_LIST = "hit.list";
-    
-    public static final String TYPE_METADATA_IN_HDFS = "type.metadata.in.hdfs";
-    
-    public static final String METADATA_TABLE_NAME = "model.table.name";
-    
-    public static final String RAW_TYPES = "raw.types";
-    
     public static final String NULL_BYTE = "\0";
     
     private String dateIndexTableName;
@@ -246,6 +231,7 @@ public class ShardQueryLogic extends BaseQueryLogic<Entry<Key,Value>> {
     private boolean limitFieldsPreQueryEvaluation = false;
     private String limitFieldsField = null;
     private Set<String> groupFields = new HashSet<>(0);
+    private Set<String> uniqueFields = new HashSet<>(0);
     private boolean compressServerSideResults = false;
     private boolean indexOnlyFilterFunctionsEnabled = false;
     /**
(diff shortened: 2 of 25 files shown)


warehouse/query-core/src/main/java/datawave/query/transformer/DocumentTransformer.java
```diff
@@ -165,80 +165,90 @@ static protected Key correctKey(Key origKey) {
...
-            dataType = colf.substring(0, index);
-            uid = colf.substring(index + 1);
-            
-            // We don't have to consult the Document to rebuild the Visibility, the key
-            // should have the correct top-level visibility
-            ColumnVisibility eventCV = new ColumnVisibility(documentKey.getColumnVisibility());
-            
-            try {
-                
-                for (String contentFieldName : this.contentFieldNames) {
```
ivakegg: This loop replace with ContentTransform

warehouse/query-core/src/main/java/datawave/query/transformer/DocumentTransformer.java
```diff
@@ -165,80 +165,90 @@ static protected Key correctKey(Key origKey) {
...
-                    if (document.containsKey(contentFieldName)) {
-                        Attribute<?> contentField = document.remove(contentFieldName);
-                        if (contentField.getData().toString().equalsIgnoreCase("true")) {
-                            Content c = new Content(uid, contentField.getMetadata(), document.isToKeep());
-                            document.put(contentFieldName, c, false, this.reducedResponse);
-                        }
-                    }
-                }
-                
-                for (String primaryField : this.primaryToSecondaryFieldMap.keySet()) {
```
ivakegg: This loop replaced with FieldMappingTransform

warehouse/query-core/src/main/java/datawave/query/transformer/DocumentTransformer.java
```diff
@@ -165,80 +165,90 @@ static protected Key correctKey(Key origKey) {
...
+    
+    @Override
+    public Object transform(Object input) throws EmptyObjectException {
         if (input instanceof Entry<?,?>) {
             @SuppressWarnings("unchecked")
             Entry<Key,org.apache.accumulo.core.data.Value> entry = (Entry<Key,org.apache.accumulo.core.data.Value>) input;
             
             Entry<Key,Document> documentEntry = deserializer.apply(entry);
-            
-            documentKey = correctKey(documentEntry.getKey());
```
ivakegg: This section moved down to _transform method

warehouse/query-core/src/main/java/datawave/query/tables/DeduppingIterator.java
```diff
@@ -15,14 +15,14 @@
 /**
  * Created on 9/13/16.
  */
-class DedupingIterator implements Iterator<Entry<Key,Value>> {
+class DeduppingIterator implements Iterator<Entry<Key,Value>> {
     private Iterator<Entry<Key,Value>> delegate;
```
tomnelson: dedupping?

warehouse/query-core/src/test/java/datawave/query/transformer/UniqueTransformTest.java
```diff
@@ -0,0 +1,153 @@
...
+            if (Character.isLetterOrDigit(c)) {
+                builder.append(c);
+            }
+        }
+        return builder.toString();
+    }
+    
+    private String createVisibility(Random random) {
+        int count = random.nextInt(5) + 2;
+        int size = random.nextInt(7) + 3;
```
tomnelson: size is not used

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -904,6 +868,12 @@ public QueryLogicTransformer getTransformer(Query settings) {
             transformer.setProjectFields(config.getProjectFields());
             transformer.setBlacklistedFields(config.getBlacklistedFields());
         }
+        if (settings.findParameter(QueryParameters.UNIQUE_FIELDS).getParameterValue().length() > 0) {
+            transformer.addTransform(new UniqueTransform(this, config.getUniqueFields()));
+        }
+        if (settings.findParameter(QueryParameters.GROUP_FIELDS).getParameterValue().length() > 0) {
+            transformer.addTransform(new GroupingTransform(this, config.getGroupFields()));
+        }
```
FineAndDandy: does the order here matter? Is it possible that unique shouldn't be applied until after the group transform?

warehouse/query-core/src/main/java/datawave/query/transformer/ContentTransform.java
```diff
@@ -0,0 +1,44 @@
...
+import datawave.query.attributes.Attribute;
+import datawave.query.attributes.Content;
+import datawave.query.attributes.Document;
+import org.apache.accumulo.core.data.Key;
+
+import javax.annotation.Nullable;
+import java.util.List;
+import java.util.Map;
+
+public class ContentTransform extends DocumentTransform.DefaultDocumentTransform {
```
FineAndDandy: javadoc please

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+    
+    private List<Attribute<?>> getFields(Document document, String field) {
+        List<Attribute<?>> attrs = new ArrayList<>();
+        List<String> fields = new ArrayList<>();
+        if (modelMapping != null) {
+            fields.addAll(modelMapping.get(field));
+        }
+        if (fields.isEmpty()) {
+            fields.add(field);
```
FineAndDandy: shouldn't we add the original field as long as it isn't already in the list regardless of it if is in the model?

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
jwomeara: I think the order does matter.  the GroupingTransform always returns null on calls to apply.  In the DocumentTransformer, it stops applying the transformers to the document entry if a null is returned by the apply method.  So, if the grouping transform runs first, the unique transform won't run, right?

warehouse/query-core/src/main/java/datawave/query/transformer/DocumentTransformer.java
```diff
@@ -165,80 +165,90 @@ static protected Key correctKey(Key origKey) {
...
-                        for (String secondaryField : this.primaryToSecondaryFieldMap.get(primaryField)) {
-                            if (document.containsKey(secondaryField)) {
-                                document.put(primaryField, document.get(secondaryField), false, this.reducedResponse);
-                                break;
-                            }
-                        }
-                    }
+            for (DocumentTransform transform : transforms) {
+                if (documentEntry != null) {
+                    documentEntry = transform.apply(documentEntry);
```
jwomeara: If the grouping transformer runs first here, and it returns null (which it always does), the unique transformer will never run.

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
```diff
@@ -0,0 +1,224 @@
...
+
+import javax.annotation.Nullable;
+import java.math.BigDecimal;
+import java.util.Collection;
+import java.util.LinkedList;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+
+public class GroupingTransform extends DocumentTransform.DefaultDocumentTransform {
```
jwomeara: This logic would be easier to follow with a test.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
jwomeara: I don't think you want to add the original field if the list is non-empty.  That implies that the original field was a model field, which wouldn't be present in your document.

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
jwomeara: Whoops!  Didn't see that you have a test.  Running through it now.  Disregard!

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
FineAndDandy: that assumes that the model entry always returns itself otherwise it would drop the original field. If we are going to do that it should be well documented.

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
jwomeara: However, it probably doesn't make sense to ever run the grouping followed by unique.  The way it is now, assuming they are running against the same fields, group by will return a count of 1 for every entry returned by unique.  If the grouping is run on a completely different set of fields, I would expect nothing to be returned.  If it's run on a subset of the fields used for unique, I would expect it to group and count accordingly.

If you group first, and then run unique, the unique transform will do nothing if the fields are the same between the transforms.  If the fields are completely different for unique, I would expect nothing to be returned.  If it's run on a subset of the fields used for unique, I guess it would return all of the unique counts?  Is that useful?

jwomeara: It would be nice to test all the different permutations of unique and group by (i.e. unique on set of fields then group by subset of fields, or unique on set of fields then group by a totally different field)

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
jwomeara: Ok, you're right...  Whatever...

warehouse/query-core/src/main/java/datawave/query/transformer/DocumentTransformer.java
jwomeara: Ok, I get it.  This isn't a problem right now, but if any transforms are added in the future, they'll have to be added prior to this one.

PR #164, opened 2018-08-07
fixes #163 - Updated Intersection constructor to mark a node as ABSEN…

### LINKED ISSUE #163: RangeStream fails to create Intersection for EXCEEDED_VALUE_THRESHOLD nodes when querying outside the date range cap
Either update the date range cap to always fail if we are outside the valid range, or update the Intersection logic to handle this properly.

jwomeara:
…T if it contains an EXCEEDED_VALUE_THRESHOLD IndexStream with no ranges to scan.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/index/lookup/Intersection.java
+++ b/warehouse/query-core/src/main/java/datawave/query/index/lookup/Intersection.java
@@ -70,8 +70,6 @@ public class Intersection implements IndexStream {
         isVariable = false;
         if (childrenItr.hasNext()) {
             boolean absent = false;
-            boolean exceededTermThreshold = false;
-            boolean exceededValueThreshold = false;
             boolean delayedField = false;
             while (childrenItr.hasNext()) {
                 IndexStream stream = childrenItr.next();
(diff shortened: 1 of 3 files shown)


warehouse/query-core/src/main/java/datawave/query/index/lookup/Intersection.java
```diff
@@ -70,8 +70,6 @@ public Intersection(Iterable<? extends IndexStream> children, UidIntersector uid
         isVariable = false;
         if (childrenItr.hasNext()) {
             boolean absent = false;
-            boolean exceededTermThreshold = false;
```
jwomeara: These fields were unused, so I removed them.

warehouse/query-core/src/main/java/datawave/query/index/lookup/Intersection.java
```diff
@@ -83,9 +81,6 @@ public Intersection(Iterable<? extends IndexStream> children, UidIntersector uid
                     continue;
                 
                 if (stream.hasNext()) {
-                    if (StreamContext.EXCEEDED_VALUE_THRESHOLD == stream.context()) {
```
jwomeara: This was essentially a noop, so I removed it.

warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
```diff
@@ -871,12 +871,6 @@ public static boolean isNormalized(String field, Set<String> ctx) {
         Calendar end = Calendar.getInstance();
         end.setTime(config.getEndDate());
         
-        List<String> datatypes = new ArrayList<>();
```
jwomeara: This wasn't being used for anything, so I removed it.

PR #181, opened 2018-08-17
File rule watcher load path fix

bbux-atg:
Fix now works with hdfs:/// and classpath based URLs.  Verified unit tests are passing once again.

### DIFF
--- a/warehouse/core/src/main/java/datawave/ingest/util/cache/watch/FileRuleWatcher.java
+++ b/warehouse/core/src/main/java/datawave/ingest/util/cache/watch/FileRuleWatcher.java
@@ -252,11 +252,10 @@ public class FileRuleWatcher extends FileSystemWatcher<Collection<FilterRule>> {
     private Collection<? extends RuleConfig> loadParentRuleConfigs(Node parent) throws IOException {
         Collection<RuleConfig> rules = new ArrayList<>();
         String parentPathStr = parent.getTextContent();
-        URL resource = this.getClass().getResource(parentPathStr);
-        if (resource == null) {
-            throw new IllegalArgumentException("Invalid parent config path specified, resource " + parentPathStr + " not found!");
+        if (null == parentPathStr || parentPathStr.isEmpty()) {
+            throw new IllegalArgumentException("Invalid parent config path, none specified!");
         }
-        Path parentPath = new Path(resource.toString());
+        Path parentPath = new Path(parentPathStr);
         if (!fs.exists(parentPath)) {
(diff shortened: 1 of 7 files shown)


bbux-atg:
@tomnelson any objections to these changes?

warehouse/core/src/main/java/datawave/ingest/util/cache/watch/FileRuleWatcher.java
```diff
@@ -252,11 +252,18 @@ private RuleConfig getRuleConfigForNode(NodeList rules, int index) {
     private Collection<? extends RuleConfig> loadParentRuleConfigs(Node parent) throws IOException {
         Collection<RuleConfig> rules = new ArrayList<>();
         String parentPathStr = parent.getTextContent();
+        
+        if (null == parentPathStr || parentPathStr.isEmpty()) {
+            throw new IllegalArgumentException("Invalid parent config path, none specified!");
+        }
+        // look for file on classpath first
         URL resource = this.getClass().getResource(parentPathStr);
-        if (resource == null) {
```
tomnelson: Does looking for the file on the classpath make sense only because of the way we currently unit test this class? Would the file ever be in the classpath in an operational scenerio (or would it typically be in hdfs)?
If the answer to the first question is yes, then I'm not happy with having operational code make a special accomodation for the unit test situation. In that case, it might be better to have the unit test code read the file from the classpath, write the file to a /tmp location, set the property to file:///whatever then let this class use that file path (file or hdfs).

bbux-atg: The difficulty is that the path is coming from one xml config file that is pointing to another xml config file as its parent.  With both files in hdfs that is not an issues because it is easy to use fully qualified names.  With the xml file in src/test/resources setting a fully qualified path to the parent config file would be very fragile and prone to error.  Not sure how to unit test to child parent reference with out the classpath check.  Open to ideas?

tomnelson: .. then I would write both files into /tmp, and also change the parent name to /tmp/filename in the xml (since that is a file that exists only for testing in /test/resources).
I think the better solution is to go with your original fix to the FileRuleWatcher and then fix up the unit tests so that they do a better job of testing with an actual file instead of a classpath resource.
I have some ideas I can show you next week.

tomnelson:
Looks good!

PR #184, opened 2018-08-22
Add pdsh redirects

ejrgilbert:
Added back in `pdsh` redirects to `/tmp/pdsh_logs/<script_name>/stdout` and `/tmp/pdsh_logs/<script_name>/stderr`. This will help for debugging scripts and systems that fail as certain actions are performed on the system. The output directories are nested in this way so that all pdsh log files are lumped together and if multiple scripts are running at once, they aren't fighting for the same output file. The only time this would conflict is if multiple instances of the same script are running at the same time. [Originally](https://github.com/ejrgilbert/datawave/blob/2f028f3f3f69e423506c1b8873183fdecac09f5e/warehouse/ingest-scripts/src/main/resources/bin/system/start-ingest.sh), the redirect was to `/tmp/stderr` which would have been troublesome commonly if someone were to need to debug a script failure as another script ran and overwrote that log.


ejrgilbert:
@drewfarris The only way I can think of how to do this is the following command:
```bash
$ pdsh -f 25 -w <host> "ls /dne" < /dev/null 1> >(dshbak -f -d /tmp/pdsh_log/testing_cmd/stdout) 2> >(dshbak -f -d /tmp/pdsh_log/testing_cmd/stderr); cat /tmp/pdsh_log/testing_cmd/stderr/pdsh\@<localhost> 2> /dev/null
<localhost>: ssh exited with exit code 2
```

This will only show exit statuses if any hosts exited unsuccessfully. Otherwise, if there were no errors on any of the hosts, the file `/tmp/pdsh_log/testing_cmd/stderr/pdsh\@<localhost>` would not exist and not be output. This way the person who executes the command will see that an error occurred and will be able to investigate further in the `/tmp/pdsh_log/testing_cmd/stderr` directory.

I will also need to do a `rm -rf /tmp/pdsh_log/testing_cmd` at the beginning of the script to make sure we don't have leftover errors from other executions (but this should probably be in there anyway).

Does this sound desirable?

drewfarris:
Hi @ejrgilbert - I like this approach. You originally pointed out having the script write to same file every time could be a little problematic if - for example - there are multiple instances of that script running simultaneously. Your point about adding the `rm -rf` at the beginning of the script sorta reinforces this.

I wonder if it would be possible to capture the process ID of the script that is executing via the `$$` built-in variable and then use that to generate directory or file names like  `/tmp/pdsh_log/testing_cmd/<PID>/...`.

ejrgilbert:
@drewfarris Good call on the PID addition. I'll add that in and go ahead and do the `rm -rf /tmp/pdsh_log/testing_cmd/<PID>` anyway. It's extremely unlikely it'll ever actually run, but I figure it's a good "just in case" since it's a quick operation.

ejrgilbert:
[Created a PR](https://github.com/NationalSecurityAgency/datawave/pull/274) to include a utility that cleans up the logs generated by the `pdsh` commands.

warehouse/ingest-scripts/src/main/resources/bin/system/start-ingest.sh
```diff
@@ -29,7 +31,12 @@ else
   trap 'rm -f "$ingestHost"; exit $?' INT TERM EXIT
   echo $INGEST_HOST > $ingestHost
 
-  pdsh -f 25 -w ^${ingestHost} "$INGEST_BIN/ingest/start-ingesters.sh $FORCE" < /dev/null
+  localhost=$(hostname -s)
+  rm -rf /tmp/pdsh_log/${script_name}/$$
+  pdsh -f 25 -w ^${ingestHost} "$INGEST_BIN/ingest/start-ingesters.sh $FORCE" < /dev/null \
```
brianloss: It seems this pattern is repeated a lot. Maybe it's worth pulling it into a bash function and call it from each script?

warehouse/ingest-scripts/src/main/resources/bin/system/start-ingest.sh
```diff
@@ -29,7 +31,12 @@ else
   trap 'rm -f "$ingestHost"; exit $?' INT TERM EXIT
   echo $INGEST_HOST > $ingestHost
 
-  pdsh -f 25 -w ^${ingestHost} "$INGEST_BIN/ingest/start-ingesters.sh $FORCE" < /dev/null
+  localhost=$(hostname -s)
+  rm -rf /tmp/pdsh_log/${script_name}/$$
```
brianloss: /tmp is often on a small partition. Also, it gets cleaned up on reboot and at other times. Should we put these logs somewhere else? Maybe at least have a variable that defines the location so it can easily be customized for all scripts.

ejrgilbert: I just realized I hadn't pushed my latest commit. Sorry about that! In the latest commit, I made the log location configurable with the default being to `${LOG_DIR}/pdsh_logs` defined in `dev.properties`.

Will push my commit soon.

warehouse/ingest-scripts/src/main/resources/bin/system/start-ingest.sh
ejrgilbert: Good idea, done in next commit.

warehouse/ingest-scripts/src/main/resources/bin/system/stop-metrics-ingest.sh
```diff
@@ -9,8 +9,12 @@ else
 fi
 THIS_DIR="${THIS_SCRIPT%/*}"
 cd $THIS_DIR
+
 #stop scripts do not require force despite lock files
 . ../ingest/ingest-env.sh -force
+. ../util/logging_pdsh.sh
```
brianloss: logging_pdsh sources ingest-env, however without the -force arg. I'm wondering if there's a potential problem here. Maybe it's better not to source ingest-env in logging_pdsh, and instead require that anyone who sources it defines the things it needs (usually by having sourced ingest-env first).

brianloss: Sounds good. It looks like only PDSH_LOG_DIR is assumed to be in the environment anyway.

PR #194, opened 2018-08-31
pushes grouping to tservers

tomnelson:
Changes to be committed:
	modified:   warehouse/query-core/src/test/java/datawave/query/GroupingTest.java


warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -366,6 +369,35 @@ else if (documentRange != null && (!this.isContainsIndexOnlyTerms() && this.getT
             
             this.serializedDocuments = pipelineIter;
             
+            // apply the grouping transform if requested
+            GroupingTransform groupify = getGroupingTransform();
+            if (groupify != null) {
+                
+                // transform Entries from Key,Value to Key,Document
+                Iterator<Entry<Key,Document>> docIterator = Iterators.transform(this.serializedDocuments, new Function<Entry<Key,Value>,Entry<Key,Document>>() {
```
phrocker: is there a better way to inline this to avoid serde in the same memory space?

tomnelson: When group.fields is turned on,  return.fields gets set to the same grouped fields so the documents are smaller and i am only serializing/deserializing the parts that i use. 
I am still thinking about better (more efficient) ways to do this.

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
```diff
@@ -69,13 +75,45 @@ private Key createAttrKey(String fieldName) {
...
+        return new Iterator<Entry<Key,Document>>() {
+            
+            Entry<Key,Document> next;
+            
+            @Override
+            public boolean hasNext() {
+                while (in.hasNext()) {
+                    GroupingTransform.this.apply(in.next());
+                }
+                next = GroupingTransform.this.flush();
```
tomnelson: The while loop aggregates all of the Documents that are fetchedby the tserver. 
It's also possible to modify this code set an arbitrary limit on the loop so that smaller 'batches' of documents are returned. The benefit would be to reduce wait time for the webserver and to prevent potential OOM errors on the tserver.

PR #195, opened 2018-09-04
Batch grouping

tomnelson:
adds parameter to control grouping batch size on tserver

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/QueryParameters.java
+++ b/warehouse/query-core/src/main/java/datawave/query/QueryParameters.java
@@ -144,6 +144,7 @@ public class QueryParameters {
     public static final String LIMIT_FIELDS = "limit.fields";
     
     public static final String GROUP_FIELDS = "group.fields";
+    public static final String GROUP_FIELDS_BATCH_SIZE = "group.fields.batch.size";
     public static final String UNIQUE_FIELDS = "unique.fields";
     /**
      * Used to cause Documents to contain a list of selectors that hit;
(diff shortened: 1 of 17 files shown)


warehouse/query-core/src/main/java/datawave/query/QueryParameters.java
```diff
@@ -144,6 +144,7 @@
     public static final String LIMIT_FIELDS = "limit.fields";
     
     public static final String GROUP_FIELDS = "group.fields";
+    public static final String GROUP_FIELDS_BATCH_SIZE = "group.fields.batch.size";
     public static final String UNIQUE_FIELDS = "unique.fields";
```
tomnelson: controls the 'batch size' on the tservers. Purpose is to prevent OOM and to reduce wait time for the webserver. If value is zero, no grouping is done on the tservers and it is done only on the webserver

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -364,7 +367,67 @@ else if (documentRange != null && (!this.isContainsIndexOnlyTerms() && this.getT
...
+                    }
+                });
+            }
+            
+            // now apply the unique transform if requested
+            UniqueTransform uniquify = getUniqueTransform();
+            if (uniquify != null) {
+                pipelineDocuments = Iterators.filter(pipelineDocuments, uniquify.getUniquePredicate());
+            }
+
```
tomnelson: UniqueTransform was moved out of the pipeline to here

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -364,7 +367,67 @@ else if (documentRange != null && (!this.isContainsIndexOnlyTerms() && this.getT
...
+                        public boolean apply(@Nullable Entry<Key,Document> keyDocumentEntry) {
+                            log.trace("after grouping, keyDocumentEntry:" + keyDocumentEntry);
+                            return true;
+                        }
+                    });
+                }
+                
+            }
+            
+            if (this.getReturnType() == ReturnType.kryo) {
```
tomnelson: serialization moved out of the pipeline to here (after unique or grouping is done)

PR #204, opened 2018-09-10
Fixes #191 - Lots of build config cleanup, property consolidation, etc...

### LINKED ISSUE #191: Hardcoding of table names, redundant table name properties, and other similar issues....
While attempting to override/customize table names in my environment (for case/formatting consistency and for use with Accumulo namespaces, etc), I came across a number of issues:

(1) Some tables, e.g., DatawaveMetadata, are hardcoded in configs rather than being interpolated from Maven props as required

(2) Some tables are no longer in use within the codebase, e.g., knowledge* tables, ${PROTOBUF_ID_EDGE_TABLE_NAME}, etc

(3) For historical reasons which are no longer valid, there are redundant config properties that should now be deduped/consolidated in order to make configuring ingest and query components easier. E.g.,..
```
...
table.name.metadata=DatawaveMetadata
table.name.date.index=dateIndex
table.name.shardIndex=shardIndex
...
METADATA_TABLE_NAME=DatawaveMetadata
DATE_INDEX_TABLE_NAME=dateIndex
SHARD_INDEX_TABLE_NAME=shardIndex
....
num.shards=241
NUM_SHARDS=10
...
```

keith-ratcliffe:
Probably worth noting up front that no Java code was harmed in the making of this PR.
Only configs and a few scripts were modified here ;)

Many redundant/duplicate properties were consolidated, and many unused props were removed.

However, the main goal here was just to make it so that table names (and their namespaces) can be managed easily/robustly from build props, as originally intended.


keith-ratcliffe:
Since there are no java code changes and thus no side effects possible (afaik), I'll probably go ahead and merge this in later today, unless there are objections

keith-ratcliffe:
Okay, here goes... :)

PR #215, opened 2018-09-25
Added some logging for the GenerateSplitsFile code

matthpeterson:


matthpeterson:
I can't add reviewers myself.  @ivakegg @brianloss @tomnelson @bbux-atg @alerman @apmoriarty  please take a look.

apmoriarty:
Added requested reviewers, @alerman did not show up as an option.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/MetadataTableSplitsCacheStatus.java
```diff
@@ -24,10 +24,14 @@ public static boolean isCacheValid(Configuration conf) throws IOException {
         try {
             fileStatus = FileSystem.get(conf).getFileStatus(splitsPath);
         } catch (IOException ex) {
-            log.warn("Could not get the FileStatus of the splits file");
+            log.warn("Could not get the FileStatus of the splits file " + splitsPath.toString());
         }
-        return null != fileStatus
-                        && fileStatus.getModificationTime() >= System.currentTimeMillis() - conf.getLong(SPLITS_CACHE_TIMEOUT_MS, DEFAULT_CACHE_TIMEOUT);
+        long expirationTime = System.currentTimeMillis() - conf.getLong(SPLITS_CACHE_TIMEOUT_MS, DEFAULT_CACHE_TIMEOUT);
+        boolean isFileAgeValid = fileStatus.getModificationTime() >= expirationTime;
```
alerman: +1

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/MetadataTableSplits.java
```diff
@@ -184,6 +187,7 @@ public void update() {
                 log.info("updating splits file");
```
alerman: Do you think we should update this log message to include the path?

matthpeterson: I didn't mess with it because the path will get logged within the call to createTempFile on 181.
`             Path tmpSplitsFile = createTempFile(fs);`

matthpeterson:
@apmoriarty, thanks for adding reviewers.  Now that two have approved, could you please merge this for me too?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/MetadataTableSplits.java
```diff
@@ -130,7 +130,10 @@ private Path createTempFile(FileSystem fs) throws IOException {
         Path tmpSplitsFile = null;
         try {
             do {
-                tmpSplitsFile = new Path(this.splitsPath.getParent(), SPLITS_CACHE_FILE + "." + count);
+                Path parentDirectory = this.splitsPath.getParent();
+                String fileName = SPLITS_CACHE_FILE + "." + count;
+                log.info("Attempting to create " + fileName + " under " + parentDirectory.toString());
```
tomnelson: is INFO the right level for the new logging calls? They look more like DEBUG to me.

matthpeterson: Most of the info log statements would get written during calls to update.  The current design has that happen with the generateSplitsCache.sh script which currently outputs a lot but has very little useful output.  I believe the total number of info logs we're adding to its output is under 10 log entries per run.  There would be one additional info log message per ingest job.

tomnelson: I just talked with the author. If what you want is to be able to go back to see 'what happened' after a failure (without running things again with a new log level), I'm fine with the INFO log level for now.

PR #220, opened 2018-09-27
MapFileLoaderShutdownBug

4a4f484e:
exit the loader when shutdown port already taken and make running flag observable between threads


matthpeterson:
@apmoriarty, can you merge this?

PR #223, opened 2018-10-02
Fix OptionalQueryParameters & RequiredQueryParameters

apmoriarty:
Updating required & optional QueryParameters to reflect the code that validates query parameters.


web-services/client/src/main/java/datawave/webservice/query/QueryParametersImpl.java
```diff
@@ -89,11 +110,21 @@ public void validate(Map<String,List<String>> parameters) throws IllegalArgument
...
+            Preconditions.checkNotNull(this.auths, "QueryParameter 'auths' cannot be null");
+            Preconditions.checkNotNull(this.pagesize, "QueryParameter 'pagesize' cannot be null");
+            Preconditions.checkNotNull(this.pageTimeout, "QueryParameter 'pageTimeout' cannot be null");
+            Preconditions.checkNotNull(this.expirationDate, "QueryParameter 'expiration' date cannot be null");
+            Preconditions.checkNotNull(this.beginDate, "QueryParameter 'begin' cannot be null");
+            Preconditions.checkNotNull(this.endDate, "QueryParameter 'end' cannot be null");
+            Preconditions.checkNotNull(this.logicName, "QueryParameter 'logicName' cannot be null");
+        } catch (Exception e) {
+            throw new IllegalArgumentException(e.getMessage());
+        }
```
apmoriarty: Based on the validate method here and the QueryExecutorBean it appears that all these parameters need to be present. I think it makes more sense to throw an IllegalArgumentException instead of letting the query process a little later and throwing a DatawaveWebApplicationException. If someone has another opinion please share here.

web-services/client/src/main/java/datawave/webservice/query/QueryParametersImpl.java
```diff
@@ -89,11 +110,21 @@ public void validate(Map<String,List<String>> parameters) throws IllegalArgument
...
+            Preconditions.checkNotNull(this.queryName, "QueryParameter 'queryName' cannot be null");
+            Preconditions.checkNotNull(this.persistenceMode, "QueryParameter 'persistence' mode cannot be null");
+            Preconditions.checkNotNull(this.auths, "QueryParameter 'auths' cannot be null");
+            Preconditions.checkNotNull(this.pagesize, "QueryParameter 'pagesize' cannot be null");
+            Preconditions.checkNotNull(this.pageTimeout, "QueryParameter 'pageTimeout' cannot be null");
+            Preconditions.checkNotNull(this.expirationDate, "QueryParameter 'expiration' date cannot be null");
+            Preconditions.checkNotNull(this.beginDate, "QueryParameter 'begin' cannot be null");
+            Preconditions.checkNotNull(this.endDate, "QueryParameter 'end' cannot be null");
+            Preconditions.checkNotNull(this.logicName, "QueryParameter 'logicName' cannot be null");
+        } catch (Exception e) {
```
bbux-atg: I think the only type of exception that can be thrown here is NPE.  Might be worth catching that specifically.

web-services/client/src/main/java/datawave/webservice/query/QueryParametersImpl.java
```diff
@@ -89,11 +110,21 @@ public void validate(Map<String,List<String>> parameters) throws IllegalArgument
...
+            throw new IllegalArgumentException(e.getMessage());
```
jschmidt10: You might keep the "caused by" exception here. I know the message will tell us the variable but it doesn't hurt to have the original line number of the error as well: `throw new IllegalArgumentException("Missing required parameters", e)`

apmoriarty: Makes a lot of sense, I'll add that in

web-services/client/src/main/java/datawave/webservice/query/QueryParametersImpl.java
apmoriarty: You're right, the only exception that can be thrown is a NPE. I'll update the catch to reflect that.

warehouse/query-core/src/main/java/datawave/query/tables/edge/EdgeQueryLogic.java
```diff
@@ -699,14 +699,12 @@ public void setQueryThreads(int queryThreads) {
     
     @Override
     public Set<String> getOptionalQueryParameters() {
-        Set<String> params = new TreeSet<>();
-        params.add(datawave.webservice.query.QueryParameters.QUERY_BEGIN);
```
jschmidt10: It looks like begin and end date are optional for EdgeQuery (I'm looking at EdgeQuery.getDateFilter).

web-services/client/src/main/java/datawave/webservice/query/QueryParametersImpl.java
```diff
@@ -89,11 +110,21 @@ public void validate(Map<String,List<String>> parameters) throws IllegalArgument
...
+        
+        try {
+            Preconditions.checkNotNull(this.query, "QueryParameter 'query' cannot be null");
+            Preconditions.checkNotNull(this.queryName, "QueryParameter 'queryName' cannot be null");
+            Preconditions.checkNotNull(this.persistenceMode, "QueryParameter 'persistence' mode cannot be null");
+            Preconditions.checkNotNull(this.auths, "QueryParameter 'auths' cannot be null");
+            Preconditions.checkNotNull(this.pagesize, "QueryParameter 'pagesize' cannot be null");
+            Preconditions.checkNotNull(this.pageTimeout, "QueryParameter 'pageTimeout' cannot be null");
+            Preconditions.checkNotNull(this.expirationDate, "QueryParameter 'expiration' date cannot be null");
+            Preconditions.checkNotNull(this.beginDate, "QueryParameter 'begin' cannot be null");
```
jschmidt10: Since it looks like EdgeQuery has optional dates, and the parsing above this block actually sets beginDate and endDate explicitly to null in some cases, I'd drop these checkNotNull's.

apmoriarty: I'll update this soon, thanks for pointing that out.

warehouse/query-core/src/main/java/datawave/query/tables/edge/EdgeQueryLogic.java
apmoriarty: Good catch

web-services/client/src/main/java/datawave/webservice/query/QueryParametersImpl.java
keith-ratcliffe: Btw, I noticed that web service responses no longer seem to indicate the missing required parameter for the user the way that they used to, but I checked and that's also the case with the master branch behavior, so nothing that has to be addressed here. I'll create a separate ticket and see if I can figure out when/how that changed

contrib/datawave-quickstart/bin/services/datawave/test-web/tests/EventQuery400Error.test
```diff
@@ -4,4 +4,4 @@ configureTest \
          "Creates an error for EventQuery by omitting all required query params" \
          "--header 'Content-Type: application/x-www-form-urlencoded' -X POST ${URI_ROOT}/Query/EventQuery/create" \
          application/xml \
-         500
+         400
```
FineAndDandy: if this is changing from a 500 to a 400 the test should probably be renamed

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -1805,36 +1805,41 @@ public void setIndexOnlyFilterFunctionsEnabled(boolean enabled) {
...
-        params.add(QueryOptions.HIT_LIST);
-        params.add(QueryOptions.TYPE_METADATA_IN_HDFS);
-        params.add(QueryOptions.DATE_INDEX_TIME_TRAVEL);
-        params.add(QueryParameters.LIMIT_FIELDS);
-        params.add(QueryParameters.GROUP_FIELDS);
-        params.add(QueryParameters.UNIQUE_FIELDS);
-        params.add(QueryOptions.LOG_TIMING_DETAILS);
-        return params;
+        Set<String> optionalParams = new TreeSet<>();
+        optionalParams.add(QueryParameters.QUERY_SYNTAX);
```
FineAndDandy: syntax required? Or can this be inferred from the logic?

warehouse/query-core/src/test/java/datawave/query/jexl/functions/ContentFunctionQueryTest.java
```diff
@@ -249,6 +250,7 @@ public void withinTestWithAlternateDate() throws Exception {
         MultivaluedMap<String,String> params = new MultivaluedMapImpl<>();
         params.putSingle(QUERY_STRING, queryStr);
         params.putSingle(QUERY_NAME, "contentQuery");
+        params.putSingle(QUERY_LOGIC_NAME, "EventQueryLogic");
```
FineAndDandy: This implies the API is changing

web-services/client/src/main/java/datawave/webservice/query/QueryParametersImpl.java
```diff
@@ -41,14 +41,39 @@ public QueryParametersImpl() {
         clear();
     }
     
+    /**
+     * Configure internal variables vi the incoming parameter map, performing validation of values.
```
FineAndDandy: typo

web-services/client/src/test/java/datawave/webservice/query/TestQueryParameters.java
```diff
@@ -21,14 +21,17 @@
...
-    }
-    
-    @Test
-    public void testNullExpirationDate() {
         parameters.putSingle(QueryParameters.QUERY_AUTHORIZATIONS, "ALL");
         parameters.putSingle(QueryParameters.QUERY_NAME, "Test");
         parameters.putSingle(QueryParameters.QUERY_PERSISTENCE, "TRANSIENT");
         parameters.putSingle(QueryParameters.QUERY_STRING, "FOO == BAR");
+        parameters.putSingle(QueryParameters.QUERY_BEGIN, "20180928");
+        parameters.putSingle(QueryParameters.QUERY_END, "20180929");
```
FineAndDandy: implies API changes

web-services/client/src/main/java/datawave/webservice/query/QueryParametersImpl.java
apmoriarty: Fixed in next commit

web-services/client/src/test/java/datawave/webservice/query/TestQueryParameters.java
apmoriarty: Removed in next commit

warehouse/query-core/src/test/java/datawave/query/jexl/functions/ContentFunctionQueryTest.java
apmoriarty: We're enforcing the API in the QueryExecutorBean which requires a logic name to define and create queries.

warehouse/query-core/src/main/java/datawave/query/tables/facets/FacetedQueryLogic.java
```diff
@@ -173,14 +173,13 @@ public boolean apply(Entry<Key,Value> input) {
     
     @Override
     public Set<String> getOptionalQueryParameters() {
-        Set<String> params = new TreeSet<>();
-        params.add(FacetedConfiguration.MINIMUM_COUNT);
-        params.add(FacetedConfiguration.FACETED_SEARCH_TYPE);
-        params.add(FacetedConfiguration.FACETED_FIELDS);
-        params.add(FacetedConfiguration.MAXIMUM_GROUP_COUNT);
-        params.add(FacetedConfiguration.STREAMING_ENABLED);
-        params.add(QueryOptions.LIMIT_FIELDS);
```
jschmidt10: Was LIMIT_FIELDS dropped intentionally? It appears optional from a quick peek at the code.

jschmidt10: Nevermind, I overlooked the super call. Cool.

contrib/datawave-quickstart/bin/services/datawave/test-web/tests/EventQuery400Error.test
```diff
@@ -4,4 +4,4 @@ configureTest \
          "Creates an error for EventQuery by omitting all required query params" \
          "--header 'Content-Type: application/x-www-form-urlencoded' -X POST ${URI_ROOT}/Query/EventQuery/create" \
          "application/xml;charset=UTF-8" \
-         500
+         400
```
FineAndDandy: if its now a 400 error test can the file be renamed to reflect that?

Issue #224, opened 2018-10-10
EventDataQueryExpressionVisitor should test normalized field values against normalized patterns


FineAndDandy:
EventDataQueryExpressionVisitor.apply() is using the normalized field values to accept/reject Keys, as a result the fieldValues, fieldRanges, and fieldPatterns must be normalized as well otherwise matches may be missed

drewfarris:
After some discussion @FineAndDandy and I realize this should not only do the normalized case, but also handle the non-normalized case (so that potentially both versions are included in the context that gets sent to Jexl for evalation). This will handle the case where we have a regular expression that would match the non-normalized version of the field value, yet does not match the normalized form. Currently, if we encounter that case - we fail to properly include the value in the context for evaluation, possibly ignoring a field value that would otherwise match the expression.

PR #225, opened 2018-10-10
normalize EventDataQueryExpressionVisitor values,patterns and ranges

FineAndDandy:
see #224


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EventDataQueryExpressionVisitor.java
```diff
@@ -141,17 +136,59 @@ public boolean apply(Key key) {
...
+                        } catch (Exception e) {
+                            // can't normalize this pattern, add the original matcher
+                            normalizedPatternMatchers.add(fieldPatterns.get(fieldPattern));
+                        }
+                    }
+                    
+                    // normalize all values
+                    for (String fieldValue : fieldValues) {
+                        try {
+                            String normalizedValue = type.normalize(fieldValue);
```
ivakegg: To be safe, check for a null return from type.normalize

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EventDataQueryExpressionVisitor.java
```diff
@@ -141,17 +136,59 @@ public boolean apply(Key key) {
...
-                                keyFieldValue, key);
+                final List<Type> types = EventDataQueryExpressionVisitor.extractTypes(attributeFactory, keyFieldName, keyFieldValue, key);
+                final List<Matcher> normalizedPatternMatchers = new ArrayList<>();
+                final List<String> normalizedFieldValues = new ArrayList<>();
+                final List<LiteralRange> normalizedRanges = new ArrayList<>();
+                for (Type type : types) {
+                    // normalize all patterns
+                    for (Pattern fieldPattern : fieldPatterns.keySet()) {
+                        try {
+                            String normalizedPattern = type.normalizeRegex(fieldPattern.toString());
```
ivakegg: To be safe, check for a null return from type.normalizeRegex

ivakegg:
I nudged Drew to take a look at this....

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EventDataQueryExpressionVisitor.java
```diff
@@ -141,17 +136,77 @@ public boolean apply(Key key) {
                 }
                 
                 final String keyFieldValue = datawaveKey.getFieldValue();
-                final Set<String> normalizedFieldValues = EventDataQueryExpressionVisitor.extractNormalizedAttributes(attributeFactory, keyFieldName,
-                                keyFieldValue, key);
+                final List<Type> types = EventDataQueryExpressionVisitor.extractTypes(attributeFactory, keyFieldName, keyFieldValue, key);
+                final List<Matcher> normalizedPatternMatchers = new ArrayList<>();
+                final List<String> normalizedFieldValues = new ArrayList<>();
+                final List<LiteralRange> normalizedRanges = new ArrayList<>();
```
drewfarris: Should `normalizedPatternMatchers`, `normalizedFieldValues` and `normalizedRanges` be `Set<T>` instead of a List? e.g: Do we care about duplicate values? Is deduplication handled elsewhere?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EventDataQueryExpressionVisitor.java
```diff
@@ -347,8 +402,8 @@ private static String print(JexlASTHelper.IdentifierOpLiteral iol) {
         return iol.getIdentifier() + " " + iol.getOp() + " " + iol.getLiteral();
     }
     
-    public static Set<String> extractNormalizedAttributes(AttributeFactory attrFactory, String fieldName, String fieldValue, Key key) {
-        final Set<String> normalizedAttributes = new HashSet<>();
+    public static List<Type> extractTypes(AttributeFactory attrFactory, String fieldName, String fieldValue, Key key) {
+        final List<Type> types = new ArrayList<>();
```
drewfarris: Could we end up putting duplicate types into this List unintentionally (e.g: if one type may appear in multiple TypeAttributes or if a type might appear in both an AttributeBag and a TypeAtribute - we would wind up with duplicates in this list. Perhaps instead of unique types we should consider building the list of unique normalizers?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EventDataQueryExpressionVisitor.java
FineAndDandy: agree

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EventDataQueryExpressionVisitor.java
FineAndDandy: yep, fixed

FineAndDandy:
can this be merged?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EventDataQueryExpressionVisitor.java
```diff
@@ -151,17 +153,77 @@ public boolean apply(Key key) {
...
+                        } catch (Exception e) {
+                            // can't normalize the range values, add the original
+                            normalizedRanges.add(range);
+                        }
+                    }
+                }
                 
-                for (String normalizedFieldValue : normalizedFieldValues) {
-                    if (fieldValues.contains(normalizedFieldValue)) {
+                Set<String> fieldValuesToEvaluate = EventDataQueryExpressionVisitor.extractNormalizedValues(types);
```
drewfarris: We likely need to add the non-normalized value to this set as well, see the comment in #246

PR #231, opened 2018-10-11
Jira-5355

AJDubya:
Add timers to InjestJob.java to improve efficiency

Placing timers in InjestJob.java file to analyze how long methods/processes take at the beginning of ingest jobs. If any process takes to long may consider removing those methods/processes and have them executed outside of ingest.

### DIFF
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
@@ -260,11 +260,21 @@ public class IngestJob implements Tool {
         log.setLevel(Level.INFO);
         
         // Initialize the markings file helper so we get the right markings file
+        long start = System.currentTimeMillis();
         MarkingFunctions.Factory.createMarkingFunctions();
+        long stop = System.currentTimeMillis();
+        log.trace("Time to initialize MarkingsFunctions=" + (stop - start));
+        
+        start = System.currentTimeMillis();
         TypeRegistry.reset();
+        stop = System.currentTimeMillis();
+        log.trace("Time for TypeRegistry=" + (stop - start));
         
         // Parse the job arguments
+        start = System.currentTimeMillis();
         Configuration conf = parseArguments(args, this.getConf());
+        stop = System.currentTimeMillis();
+        log.trace("Time for parseArguments=" + (stop - start));
         
         if (conf == null) {
             printUsage();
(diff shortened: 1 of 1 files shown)


warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
```diff
@@ -260,11 +260,21 @@ public int run(String[] args) throws Exception {
         log.setLevel(Level.INFO);
         
         // Initialize the markings file helper so we get the right markings file
+        long start = System.currentTimeMillis();
```
alerman: This would be a great place to look at the org.slf4j.profiler.StopWatch class. I would much prefer that (one stopwatch with named timers) to the repetitive checking of current time.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
```diff
@@ -347,7 +383,11 @@ public int run(String[] args) throws Exception {
             job.addFileToClassPath(dependency);
         
         configureInputFormat(job, cbHelper, conf);
+        stop = System.currentTimeMillis();
+        log.trace("Time for setJarByClass=" + (stop - start));
```
jschmidt10: In addition to timing setJarByClass, this is adding files to the classpath and setting up the input format.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
```diff
@@ -289,11 +300,17 @@ public int run(String[] args) throws Exception {
...
         }
+        stop = System.currentTimeMillis();
+        log.trace("Time for TypeRegistry.getInstance=" + (stop - start));
         
+        start = System.currentTimeMillis();
         if (!registerTableNames(conf)) {
             return -1;
         }
+        stop = System.currentTimeMillis();
+        log.trace("Time for !registerTableNames=" + (stop - start));
```
jschmidt10: This is the time to read all the tables from all the ingest components and then register them under job.table.names. I would either drop the not (!) here. The thing you are timing i sjust the registerTableNames() function. The not is used to make sure that it succeeded.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
jschmidt10: I know we've used these timings-in-code in other places before, so there's a precedent but do you think we might have more success just profiling a few jobs instead of adding timing code? I feel like it adds clutter to the code and makes it harder to follow.

ivakegg: I will have to concur with @jschmidt10 here in that this is really cluttering up the code.  Using the StopWatch class may make this clean enough to accept.

ivakegg:
Closed due to inactivity

PR #232, opened 2018-10-12
Added RabbitMQ Health Monitoring to the audit service.

jwomeara:


services/audit-service/src/main/java/datawave/microservice/audit/health/rabbit/RabbitHealthChecker.java
```diff
@@ -0,0 +1,705 @@
...
+            queuesHealthCheck();
+            
+            // check the bindings
+            bindingsHealthCheck();
+            
+            boolean isHealthy = isHealthy();
+            
+            if (wasHealthy) {
+                if (isHealthy) {
+                    log.info("RabbitMQ is still healthy.");
```
brianloss: Health check might run very frequently. Do we want to log every successful health check to info?

services/audit-service/src/main/java/datawave/microservice/audit/health/rabbit/RabbitHealthChecker.java
```diff
@@ -0,0 +1,705 @@
...
+                }
+            } else {
+                if (isHealthy) {
+                    lastSuccessfulHealthCheck = new Date();
+                    if (currentOutageStats != null) {
+                        currentOutageStats.setStopDate(lastSuccessfulHealthCheck);
+                        currentOutageStats = null;
+                        log.info("RabbitMQ has recovered.");
+                    } else
+                        log.info("RabbitMQ is healthy");
```
brianloss: Nit: if you have braces on the if, then they should be on the else as well.

services/audit-service/src/main/java/datawave/microservice/audit/health/rabbit/RabbitHealthChecker.java
```diff
@@ -0,0 +1,705 @@
...
+     * This method is synchronized with the isHealthy method to ensure that the health check is given time to complete before deciding whether to accept an
+     * audit request via the {@link AuditController}
+     * <p>
+     * RabbitMQ is considered to be healthy if the cluster, exchanges, queues, and bindings are configured as specified in the {@link RabbitHealthProperties}.
+     * If we are not able to determine the health of the cluster, exchanges, queues, and bindings (perhaps due to the management API being unavailable, or the
+     * cluster being down), we will consider RabbitMQ to be unhealthy.
+     */
+    @Override
+    public void runHealthCheck() {
+        initRabbitClient();
```
brianloss: This can leave rabbitClient and rabbitManagementTemplate as null. I expect that's ok since you could have the case where this health check runs before RabbitMQ comes up. However, I would think if that's the case, we could fail fast here and report everything as unhealthy. Thoughts?

brianloss: Doing that could also remove a bunch of null checks in the code called from the rest of this method.

services/audit-service/src/main/java/datawave/microservice/audit/health/rabbit/RabbitHealthChecker.java
```diff
@@ -0,0 +1,705 @@
...
+        synchronized (this) {
+            return clusterHealthy && exchangesHealthy && queuesHealthy && bindingsHealthy;
+        }
+    }
+    
+    private void updateOutageStats(RabbitOutageStats outageStats) {
+        if (numNodesMissing > 0)
+            outageStats.setNumNodesMissing(Math.max(outageStats.getNumNodesMissing(), numNodesMissing));
+        
+        outageStats.getMissingExchanges().addAll(missingExchanges.stream().map(Exchange::getName).collect(Collectors.toList()));
```
brianloss: Not that this gets called a lot, so it's not a big deal, but you could avoid creating an extra list by doing something like:
```
missingExchanges.stream().map(Exchange::getName).forEach(e -> outageStats.getMissingExchanges().add(e));
```

services/audit-service/src/main/java/datawave/microservice/audit/health/rabbit/config/RabbitHealthProperties.java
```diff
@@ -0,0 +1,366 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+@ConfigurationProperties(prefix = "audit.health.rabbit")
+public class RabbitHealthProperties {
+    private boolean enabled = false;
+    
+    // default - 5 minutes
+    private long healthyPollIntervalMillis = 300000L;
```
brianloss: It might be more readable with `TimeUnit.MINUTES.toMillis(5)`.

services/audit-service/src/main/java/datawave/microservice/audit/health/rabbit/config/RabbitHealthProperties.java
```diff
@@ -0,0 +1,366 @@
...
+    
+    // default - 5 seconds
+    private long unhealthyPollIntervalMillis = 5000L;
```
brianloss: It might be more readable with `TimeUnit.SECONDS.toMillis(5)`.

services/audit-service/src/main/java/datawave/microservice/audit/controller/AuditController.java
```diff
@@ -17,34 +27,131 @@
...
+ * Also, if configured, the audit controller will verify that the message passing infrastructure is healthy before returning successfully to the user. If the
+ * message passing infrastructure is unhealthy, or if we can't verify that the message was successfully passed to our messaging infrastructure, a 500 Internal
+ * Server Error will be returned to the caller.
+ */
 @RestController
 @RequestMapping(path = "/v1", produces = MediaType.APPLICATION_JSON_VALUE)
 public class AuditController {
     private final Logger log = LoggerFactory.getLogger(this.getClass());
     
+    private AuditProperties auditProperties;
```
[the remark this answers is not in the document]
jwomeara: Updated!

services/audit-service/src/main/java/datawave/microservice/audit/controller/AuditController.java
```diff
@@ -17,34 +27,131 @@
...
+     *            The audit parameters to be sent
+     */
     private void sendMessage(AuditParameters parameters) {
-        if (!messageChannel.send(MessageBuilder.withPayload(AuditMessage.fromParams(parameters)).build())) {
-            throw new RuntimeException("Failed to send audit: " + parameters);
+        String auditId = parameters.getAuditId();
+        
+        CountDownLatch latch = null;
+        if (auditProperties.isConfirmAckEnabled()) {
+            new CountDownLatch(1);
```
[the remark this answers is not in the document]
jwomeara: D'oh!  Fixed!

services/audit-service/src/main/java/datawave/microservice/audit/health/rabbit/RabbitHealthChecker.java
```diff
@@ -0,0 +1,705 @@
...
+    public RabbitHealthChecker(RabbitHealthProperties rabbitHealthProperties, CachingConnectionFactory rabbitConnectionFactory) throws Exception {
+        this(rabbitHealthProperties, rabbitConnectionFactory.getHost(), rabbitConnectionFactory.getUsername(),
+                        rabbitConnectionFactory.getRabbitConnectionFactory().getPassword());
+    }
+    
+    public RabbitHealthChecker(RabbitHealthProperties rabbitHealthProperties, String host, String username, String password) {
+        this.rabbitHealthProperties = rabbitHealthProperties;
+        
+        ManagementProperties mgmtProps = (rabbitHealthProperties != null) ? rabbitHealthProperties.getManagement() : null;
+        this.host = (mgmtProps != null && mgmtProps.getHost() != null && !mgmtProps.getHost().isEmpty()) ? mgmtProps.getHost() : host;
```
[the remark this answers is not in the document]
jwomeara: I'm not sure, but I updated the RabbitHealthProperties to have non-null values for every property.  After doing that, this becomes a bit cleaner.

services/audit-service/src/main/java/datawave/microservice/audit/health/rabbit/RabbitHealthChecker.java
```diff
@@ -0,0 +1,705 @@
...
+        
+        init();
+    }
+    
+    private void init() {
+        initRabbitClient();
+        
+        List<ExchangeProperties> exchProps = rabbitHealthProperties.getExchanges();
+        if (exchProps != null && !exchProps.isEmpty())
+            exchanges = exchProps.stream().map(RabbitHealthChecker::createExchange).collect(Collectors.toList());
```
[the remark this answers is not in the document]
jwomeara: Now that all of the properties in RabbitHealthChecker have non-null values, I was able to simplify this logic quite a bit.

services/audit-service/src/main/java/datawave/microservice/audit/health/rabbit/RabbitHealthChecker.java
jwomeara: I was conflicted on this.  Looking at it again, maybe try this...  Use INFO for transitions from unhealthy to healthy.  Use WARN for transitions from healthy to unhealthy.  Everything else is trace.  Sound good?

services/audit-service/src/main/java/datawave/microservice/audit/health/rabbit/RabbitHealthChecker.java
jwomeara: Updated!

services/audit-service/src/main/java/datawave/microservice/audit/health/rabbit/RabbitHealthChecker.java
jwomeara: OK, this is fair.  I updated runHealthCheck to skip the health checks if the client & template are null.  I also updated the isHealthy method to ensure that rabbitClient, and rabbitManagementTemplate are non-null.  This should account for any failure to connect to rabbit at initialization.  The redundant null checks should have been removed as well.  

I also updated the recover method to perform a similar check on the client and management template.  So, that should be cleaner now as well.

services/audit-service/src/main/java/datawave/microservice/audit/health/rabbit/RabbitHealthChecker.java
jwomeara: Good point.  Updated!

services/audit-service/src/main/java/datawave/microservice/audit/health/rabbit/config/RabbitHealthProperties.java
jwomeara: Sure, updated!

services/audit-service/src/main/java/datawave/microservice/audit/health/rabbit/config/RabbitHealthProperties.java
jwomeara: Same as above.

services/audit-service/src/main/java/datawave/microservice/audit/controller/AuditController.java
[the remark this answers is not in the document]
brianloss: restAuditParams and messageChannel too, no?

services/audit-service/src/main/java/datawave/microservice/audit/health/rabbit/RabbitHealthChecker.java
brianloss: Yup, that works. I'd say leave this log, but move it to debug level.

services/audit-service/src/main/java/datawave/microservice/audit/controller/AuditController.java
```diff
@@ -17,34 +27,131 @@
...
+     * 
+     * @param message
+     *            the confirmation ack message
+     */
+    @ConditionalOnProperty(value = "audit.confirmAckEnabled", havingValue = "true", matchIfMissing = true)
+    @ServiceActivator(inputChannel = CONFIRM_ACK_CHANNEL)
+    public void processConfirmAck(Message<?> message) {
+        String correlationId = null;
+        if (message.getHeaders().containsKey(IntegrationMessageHeaderAccessor.CORRELATION_ID)) {
+            Object headerObj = message.getHeaders().get(IntegrationMessageHeaderAccessor.CORRELATION_ID);
```
brianloss: Is getHeaders() not a normal map? Since you're null-checking the value, you don't need to test if it's in the map. Just get it and you'll either get the value or null.

services/audit-service/src/main/java/datawave/microservice/audit/controller/AuditController.java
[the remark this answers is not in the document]
jwomeara: Double updated!

jwomeara:
@brianloss Did I address all of your concerns?

brianloss:
I think so. In the future, I'd suggest not rebasing every time you change something. That makes it hard to review changes on a PR that happen after the initial review. We can squash or rebase at merge time.

PR #235, opened 2018-10-23
Flatten portion of visit(EQNode) method in the RangeStream class

apmoriarty:
Flatten portion of visit(EQNode) method in the RangeStream class for readability.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
+++ b/warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
@@ -442,112 +442,109 @@ public class RangeStream extends BaseVisitor implements CloseableIterable<QueryP
             return ScannerStream.unindexed(node);
         }
         
-        if (isIndexed(fieldName, config.getIndexedFields())) {
-            
-            log.debug("\"" + fieldName + "\" is indexed. for " + literal);
-            
-            try {
-                
-                // two scenarios
-                Iterator<Tuple2<String,IndexInfo>> itr = null;
-                int stackStart = config.getBaseIteratorPriority();
-                
-                if (limitScanners) {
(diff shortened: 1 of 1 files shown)


warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
```diff
@@ -442,112 +442,109 @@ public ScannerStream visit(ASTEQNode node, Object data) {
...
+                uidSetting.addOption(CreateUidsIterator.COLLAPSE_UIDS, Boolean.valueOf(config.getCollapseUids()).toString());
+                scanner.addScanIterator(uidSetting);
+                
+                itr = Iterators.transform(scanner.iterator(), new EntryParser(node, fieldName, literal, indexOnlyFields));
             }
-            log.debug("{\"" + fieldName + "\": \"" + literal.toString() + "\"} is not an observed field.");
-            return ScannerStream.unknownField(node);
+            
+            /**
+             * Create a scanner in the initialized state so that we can
```
alerman: so that we can...

I know you didn't change this, but i feel like we should complete the sentence or delete this comment

bbux-atg:
I think I can see what changed.  Do the unit tests already cover this case?  If not would it be helpful to create one?

apmoriarty:
> 
> 
> I think I can see what changed. Do the unit tests already cover this case? If not would it be helpful to create one?

The unit tests in warehouse/query-core cover the visit(EQNode) method well. A specific test for the RangeStream would be a good idea but take some time due to the class complexity.

warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
ivakegg: Just remove the comment....it is useless

PR #238, opened 2018-10-25
Re #234: Lucene dependency upgraded from 4.x to 7.x

keith-ratcliffe:
First pass at upgrading Lucene, for SME review

Summary of things that were broken in DW as a result of the upgrade:
* Tokenization of acronyms and words containing apostrophes changed significantly due to deprecation of **StandardFilter**
* Query parsing WRT wildcard characters, for translating Lucene queries to JEXL changed significantly due to changes in Lucene's WildcardQueryNodeProcessor
* Misc API changes, which broke the build

All resolved (see below), pending more testing/vetting of course


keith-ratcliffe:
@drewfarris see [12e5f37](https://github.com/NationalSecurityAgency/datawave/pull/238/commits/12e5f370e20b15156a7e6cd99bed1890e58d78b7) regarding the observed _TestLuceneToJexlQueryParser_ failures. The commit resolves  those. Mainly, the newer version of WildcardQueryNodeProcessor was to blame b/c it adds lowercase normalization and other new behaviors, which broke the tests

Also, in LuceneToJexlQueryParser, note that I swapped out Lucene's StandardAnalyzer for our own custom one, so that we don't pick up any unwanted Analyzer-specific normalizations from other NodeProcessors in the future. That was part of the problem with WildcardQueryNodeProcessor#postProcessNode, as it applies normalization behaviors provided by the configured Analyzer

keith-ratcliffe:
A few parting thoughts for this PR, fwiw...

I kinda like that we're now using _datawave.ingest.data.tokenize.StandardAnalyzer_ in both the ingest and query pipelines, because we should be able to leverage that fact to insulate ourselves from future Lucene changes and to ensure consistent tokenization behavior across both tiers. 

It also got me thinking about a couple more things we might want to consider down the road, to enhance our Lucene integration:

* Provide a simple API that would allow users to more easily plug their own custom analyzers and tokenizers into ingest and query, as needed.
* Maybe consolidate all of our Lucene-related code, from ingest-core and query-core, into a single, distinct project module, to simplify maintenance

drewfarris:
> Provide a simple API that would allow users to more easily plug their own custom analyzers and tokenizers into ingest and query, as needed.
Maybe consolidate all of our Lucene-related code, from ingest-core and query-core, into a single, distinct project module, to simplify maintenance

Yes, to both of this - specifically the first. I would like to be able to do per-field or per-type tokenizer configurations.

Good point about consolidating everything in to a single module, it would be interesting to see if this just ends up being a bunch of facade classes though, which would seem less valuable.

drewfarris:
There are some tweaks I'd like to make to this around `StandardAnalyzer` / `StandardTokenizer` that make them a little more like the current Lucene implementations than they are currently. @keith-ratcliffe if you give me write access to your PR branch, I'll go ahead and add what I have.

keith-ratcliffe:
> There are some tweaks I'd like to make to this around `StandardAnalyzer` / `StandardTokenizer` that make them a little more like the current Lucene implementations than they are currently. @keith-ratcliffe if you give me write access to your PR branch, I'll go ahead and add what I have.

@drewfarris the branch (lucene-upgrade) is on this repo, so you should be g2g

drewfarris:
Replaced by #302

PR #244, opened 2018-11-05
Re #234 Tweaks to lucene upgrade PR

drewfarris:
@keith-ratcliffe  here are some changes to #238 that bring the `StandardTokenizer` / `StandardAnalyzer` in line with some of the changes to the equivalent classes in the Lucene codebase.


keith-ratcliffe:
@drewfarris looks great. definitely makes more sense relative to 7.5.0 analyzer/tokenizer changes

PR #247, opened 2018-11-08
Let QueryLogic determine its response class

ejrgilbert:
The way getting a `QueryLogic`'s response class works right now makes the assumption that the `QueryLogic` doesn't need to be initialized in order to have a successful call on `getTransformer`. Since the call fails on several logics, the `listQueryLogic` page has several "unknown" response classes.

This fix enables `QueryLogic`s to Override the default implementation of the new method `getResponseClass`. This way the logics that work with this original implementation are not affected and new logics that would need to implement an override to this method could do so and avoid always having an "unknown" response class.

### DIFF
--- a/web-services/query/src/main/java/datawave/webservice/query/logic/QueryLogic.java
+++ b/web-services/query/src/main/java/datawave/webservice/query/logic/QueryLogic.java
@@ -1,6 +1,7 @@
 package datawave.webservice.query.logic;
 
 import java.security.Principal;
+import java.util.Date;
 import java.util.List;
 import java.util.Set;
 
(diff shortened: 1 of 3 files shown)


web-services/query/src/main/java/datawave/webservice/query/logic/QueryLogic.java
```diff
@@ -65,6 +71,23 @@
      */
     public QueryLogicTransformer getTransformer(Query settings);
     
+    default public String getResponseClass() throws QueryException {
+        Query q = new QueryImpl();
```
tomnelson: Would it be better to pass a Query in as a parameter to getResponseClass? There are no usages of getResponseClass in this PR that do not have a Query instance available for use, and after I changed it, all the tests still pass.

ejrgilbert: Pushed this change. Good idea on this, now `QueryLogic`s have the option of returning different response classes based on the query passed.

tomnelson: After discussing with you, it seems that the better argument to pass is the QueryLogicTransformer, because that is what creates the response object.

PR #252, opened 2018-11-13
Datawave cleanup 1

cjmctague:
- Anonymous types replaced with lambda
- Removed unnecessary interface modifiers
- Removed unnecessary calls to toString()

I left the individual commits to make review a little easier. `mvn clean verify` passed


cjmctague:
I've pushed a squashed and rebased version for merge

PR #262, opened 2018-11-26
Datawave cleanup part 3

cjmctague:
d179738 - Replaced explicit type with <> expression and Simplify JUnit asserts
ea1906c - removed a redundant dependency found in the pom
b04e60f - cleans the import statements throughout the project. (Remove unused imports, removes unnecessary imports from the same package, removes all use of star Imports (except static), and removes duplicates)

I'd like to implement [impsort-maven-plugin](https://github.com/revelc/impsort-maven-plugin) so the imports can have some consistency but that will be a follow on issue.


cjmctague:
Rebased and squashed

ivakegg:
I am so sorry we did not merge this sooner.  Would you please resolve the conflicts and then we can pull this in.  Thanks.

PR #264, opened 2018-11-28
Reworked the composite logic to be much faster.

jwomeara:
This includes a bunch of changes/updates/refactoring to the composite query logic based on testing against real data.  Composite queries are now at least as fast as fully-indexed queries, woohoo!

I also updated our support for the NYCTLC dataset.  That is all working correctly now.

There are a few specific things called out in the PR that I would like feedback on, but any feedback is appreciated.


warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -1331,7 +1335,13 @@ public static void logDebug(List<String> output, String message) {
     }
     
     public static void logQuery(ASTJexlScript queryTree, String message) {
-        logDebug(PrintingVisitor.formattedQueryStringList(queryTree), message);
```
jwomeara: @ivakegg Thoughts?  This would print the simple query string instead of the formatted query tree if the simple query string is over 1000000 characters (which is common for geo queries).

web-services/deploy/configuration/src/main/resources/datawave/query/QueryLogicFactory.xml
```diff
@@ -235,7 +235,7 @@
         </property>
         <property name="queryPlanner" ref="DefaultQueryPlanner" />
         <property name="sendTimingToStatsd" value="false" />
-	<property name="collectQueryMetrics" value="false" />
```
jwomeara: @keith-ratcliffe @ivakegg Any feelings against having this set to true by default?

warehouse/ingest-configuration/src/main/resources/config/ingest-config.xml
```diff
@@ -68,7 +68,7 @@
 
 <property>
   <name>ingest.value.dedup.aggregation</name>
-  <value>${table.name.shard},${table.name.shardIndex}},${table.name.shardReverseIndex},${table.name.metadata},${table.name.dateIndex}</value>
```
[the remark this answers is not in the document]
keith-ratcliffe: Good catch! Sorry about that

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
ivakegg: I would like to improve the side of the formattedQueryString.    When we have over N children at a node, then print the first 5, ... N-10, last 5 instead of the entire set.

ivakegg: or instead of printing ... (N-10), print ... (n-10 of X, Y, Z) where X, Y, Z are the unique set of fields found in the skipped children.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -782,9 +759,7 @@ protected ASTJexlScript updateQueryTree(ScannerFactory scannerFactory, MetadataH
         }
         
         // Print the nice log message
-        if (log.isDebugEnabled()) {
-            logQuery(queryTree, "Computed that the query " + (containsComposites ? " contains " : " does not contain any ") + " composite field(s)");
-        }
+        logQuery(queryTree, "Computed that the query " + (containsComposites ? " contains " : " does not contain any ") + " composite field(s)");
```
ivakegg: wrap with if (log.isDebugEnabled())

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -823,10 +796,8 @@ protected ASTJexlScript updateQueryTree(ScannerFactory scannerFactory, MetadataH
             queryTfFields = SetMembershipVisitor.getMembers(termFrequencyFields, config, metadataHelper, dateIndexHelper, queryTree);
             
             // Print the nice log message
-            if (log.isDebugEnabled()) {
-                logQuery(queryTree, "Computed that the query " + (queryTfFields.isEmpty() ? " does not contain any " : "contains " + queryTfFields + " as ")
-                                + " term frequency field(s)");
-            }
+            logQuery(queryTree, "Computed that the query " + (queryTfFields.isEmpty() ? " does not contain any " : "contains " + queryTfFields + " as ")
```
ivakegg: wrap with if (log.isDebugEnabled()).  In general do this wherever we have computations in the arguments

Issue #265, opened 2018-12-04
Apply the distributive property to fix !isExecutable(queryTree)


FineAndDandy:
Write a new visitor to expand a query when it is not executable due to an Or that is not executable nested inside an AND. If an AndNode exists which is a parent, that AND can be distributed to the ORNode to allow the query to be EXECUTABLE. The query can be rewritten using the distributive property to rewrite A && (B || C) to (A && B) || (A && C). In some cases the newly rewritten query will be executable when the original was not.

PR #267, opened 2018-12-06
DATAWAVE-265: Add ExecutableExpansionVisitor

FineAndDandy:
see #265

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExecutableExpansionVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExecutableExpansionVisitor.java
@@ -1,0 +1,437 @@
+package datawave.query.jexl.visitors;
+
+import datawave.query.config.ShardQueryConfiguration;
+import datawave.query.util.MetadataHelper;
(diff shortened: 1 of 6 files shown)


jwomeara:
Are there any key changes/additions I should look out for?  If not, then it LGTM.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExecutableExpansionVisitor.java
```diff
@@ -0,0 +1,437 @@
...
+                            + " aborting distribute");
+        }
+        
+        // fetch the minimal expansion terms instead of expanding everything
+        List<JexlNode> orTerms = getOrTerms(orNode, andNodeChild);
+        
+        // everything goes together, abort
+        if (orTerms.size() == 1) {
+            return null;
+        }
```
FineAndDandy: This is the only change besides unit tests

jwomeara: Okey dokey.  Thanks for pointing that out.

PR #273, opened 2018-12-10
Fix issue in javadoc causing build to fail

ejrgilbert:
When building datawave using the scripts, Javadoc was failing due to invalid HTML formatting. This change fixes that issue by replacing `&&` with `&amp;&amp;` and now allows the scripts to successfully build datawave.


brianloss:
Looks like this is OBE due to someone else making changes in master (see commit 0c540411bfcd53ab9f414837e3e2f7f6f0fa846e).

PR #274, opened 2018-12-10
Add utility to clean up logs produced by pdsh dshbak logging

ejrgilbert:
[PR 184](https://github.com/NationalSecurityAgency/datawave/pull/184) will generate logs via `pdsh` commands that need to be cleaned up on clusters. This added utility will remove `pdsh` logs older than 7 days in order to keep the system clean and ensure there is enough time to investigate `pdsh` command errors.

### DIFF
--- a/warehouse/ingest-scripts/src/main/resources/bin/system/cleanup_old_pdsh_logs.sh
+++ b/warehouse/ingest-scripts/src/main/resources/bin/system/cleanup_old_pdsh_logs.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+#
+# Cleanup old pdsh logs on the cluster
+
+if [[ `uname` == "Darwin" ]]; then
+    THIS_SCRIPT=`python -c 'import os,sys;print os.path.realpath(sys.argv[1])' $0`
+else
+    THIS_SCRIPT=`readlink -f $0`
+fi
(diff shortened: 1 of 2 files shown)


warehouse/ingest-scripts/src/main/resources/bin/system/cleanup_old_pdsh_logs.sh
```diff
@@ -0,0 +1,53 @@
...
+
+if [[ `uname` == "Darwin" ]]; then
+    THIS_SCRIPT=`python -c 'import os,sys;print os.path.realpath(sys.argv[1])' $0`
+else
+    THIS_SCRIPT=`readlink -f $0`
+fi
+THIS_DIR="${THIS_SCRIPT%/*}"
+cd ${THIS_DIR}
+
+. ../ingest/ingest-env.sh
```
brianloss: cleanup_old_files_util.sh already sources ingest-env, so you can avoid sourcing it twice.

ejrgilbert: Fixed in next commit.

bmwmaestoso:
Is there any any effort being done to complete this pull request?

ejrgilbert:
I'm just waiting on it to be merged in, it's still a valid PR.

Issue #276, opened 2018-12-11
TLDFieldIndexAggregator should set Attribute keep flag based on filter.apply


FineAndDandy:
When determining if an Attribute should be kept, if it is a fieldToAggregate and the filter.keep() is true is not sufficient for child documents. In order to include these Attributes in the document, change from filter.keep() to filter.apply()

FineAndDandy:
Type | Keep
----- | -----------
Event | true
Index only | true
Non event | false
Child event | true
Child index only | true
Child non-event | true

Issue #226, opened 2018-10-10
TLDIteratorBuildingVisitor uses TermFrequencyAggregator


FineAndDandy:
The TermFrequencyAggregator doesn't account for TLD documents and will produce separate documents for each child. The TLDIteratorBuildingVisitor expects documents to be generated at the TLD level

PR #277, opened 2018-12-11
DATAWAVE-276: TLDFieldIndexAggregator use apply() for toKeep flag

FineAndDandy:
see #276


drewfarris:
Nice change. `keep()` is used for determining which fields get displayed to the user while `apply()` determines which fields get loaded into the context for evaluation. Seems pretty clear we need to use the latter here.

ivakegg:
I am sorry, but I do not get why we should use the apply if the attr filter to determine whether we are "toKeep" the attribute.  If that is required, then I surmise that the toKeep method on the attr filter is not implemented correctly.

ivakegg:
I think I still need an answer to my query here...

FineAndDandy:
merged into #226 since both required similar updates to the EventQueryDataFilters

Issue #279, opened 2018-12-12
Refactor AuthorizationsUtil class


keith-ratcliffe:
AuthorizationsUtil's "downgrade" functionality is needed by all query-related components that are in the process of being transitioned to microservices. 

Therefore, we'll relocate those bits from AuthorizationsUtil.java to a new home under the services parent, and we'll refactor the existing class as needed

keith-ratcliffe:
@brianloss , @tomnelson

What do you guys think about the following, for revising the UserAuthFunctions api from the last PR

First, define the following (somewhere)...
```
@FunctionalInterface
interface AuthSupplier {
    Collection<String> getAuths();
}
```
...and then just have DatawaveUser implement that interface. Alternatively, we could forego the new FunctionalInterface above, and just have ...
```
class DatawaveUser implements Supplier<Collection<String>>, ...
...
Collection<String> get() {
   return this.getAuths();
}
``` 
...but that seems rather confusing

Anyway, in AuthorizationsUtil, we could still just pass the DatawaveUser instance as before. That could eliminate the authorization-api dependency from accumulo-utils and still allow custom impl's of UserAuthFunctions to access our DatawaveUser instance(s) via explicit casting, if needed.

So, the interface methods could become....
```
Authorizations getRequestedAuthorizations(String requestedAuths, AuthSupplier authSupplier);
Set<Authorizations> mergeAuthorizations(Authorizations primaryUserAuths, Collection<AuthSupplier> proxyChain);
```

Thoughts?

brianloss:
I don't think adding the FunctionalInterface buys anything, does it? Since we'd be passing the object and not a method reference, any old interface would work. That's certainly an option.

keith-ratcliffe:
Yep, buys us nothing. Just used to demo comparison with the ```implements Supplier<>``` option, since ```Supplier``` itself is annotated as such

Really, just a matter of where something like AuthSupplier should live. *Could* be defined in accumulo-utils, I suppose, which then forces authorization-api to pick up accumulo-utils as a dependency. But perhaps that's preferable to the current relationship(?) Not sure. Or define yet another new module, where only interfaces and the like will reside(?) I dunno

SethSmucker:
AuthorizationsUtil and the downgrade logic were pulled out to core/utils/common-utils and are shared by the microservices now, closing.

PR #280, opened 2018-12-12
Re #279 - user auth downgrade functionality refactored

keith-ratcliffe:
AuthorizationsUtil refactored to depend on new UserAuthFunctions implementation.

Additionally, UserAuthFunctions default impl may be overridden in AuthorizationsUtil if needed via  system property. E.g.,
```-Ddatawave.user.auth.functions.class=org.foo.MyCustomUserAuthFunctionsImpl```

### CODE REVIEW

File: pom.xml [hunk 1]
```diff
@@ -77,6 +77,7 @@
         <version.log4j>1.2.16</version.log4j>
         <version.log4j-extras>1.0</version.log4j-extras>
         <version.lucene>7.5.0</version.lucene>
+        <version.microservice.accumulo-utils>1.0-SNAPSHOT</version.microservice.accumulo-utils>
```
brianloss: We need to figure out what to do for things like this. The services folder isn't built by default (though you can do it by adding -Dservices) so a person just building datawave wouldn't have this snapshot dependency. The idea is that the artifacts would be pulled from maven central (or for now our hackish git branch), much like they would for any third party library. However, that would necessitate releasing the dependent artifact. Maybe an interim solution is add module dependencies just for the snapshot artifacts? (Or release a "beta" version of the artifact is another option.) Thoughts?

services/accumulo-utils/pom.xml
```diff
@@ -0,0 +1,100 @@
...
+            <artifactId>accumulo-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-jcl</artifactId>
```
brianloss: Didn't notice anything in the test class that would use this. Did I miss it, or can this be removed?

services/accumulo-utils/src/main/java/datawave/accumulo/util/security/UserAuthFunctions.java
```diff
@@ -0,0 +1,146 @@
...
+         *
+         * @return IFF every element in {@code requestedAuths} also exists in {@link DatawaveUser#getAuths()} then {@code requestedAuths} is translated to
+         *         {@link Authorizations} and returned. If either of {@code requestedAuths} or {@link DatawaveUser#getAuths()} is empty/null, then
+         *         {@link Authorizations#EMPTY} is returned. If any requested auths are missing in {@link DatawaveUser#getAuths()}, then an exception is thrown
+         *
+         * @throws IllegalArgumentException
+         *             if {@code requestedAuths} contains elements that do not exist in {@link DatawaveUser#getAuths()}
+         */
+        @Override
+        public Authorizations getRequestedAuthorizations(String requestedAuths, DatawaveUser user) {
```
brianloss: The only reason DatawaveUser (and therefore authorization api) needs to be a dependency here is to call getAuths. I'm not sure if this makes sense or not, but one thought is to do the api in terms of java 8 functions or suppliers. For example:
```java
public Authorizations getRequestedAuthorizations(String requestedAuths, Supplier<Collection<String>> authSupplier) {
   Collection<String> auth = authSupplier.get();
  // do stuff with the auths
}
```
then you would call with
```java
DatawaveUser user;
getRequestedAuthorizations("foo, bar", user::getAuths)
```
That won't work for nulls, though, so you'd have to do something like:
```java
public <T> Authorizations getRequestedAuthorizations(String requestedAuths, T authSupplier, Function<T,Collection<String>> authFunction) {
  if (authSupplier != null) {
    Collection<String> auth = authFunction.apply(authSupplier);
  }
```
and then you'd call with
```java
DatawaveUser user;
getRequestedAuthorizations("foo, bar", user, DatawaveUser::getAuths)
```
Just a thought. Like I said, not sure if it's worth it or not.

pom.xml
keith-ratcliffe: Yeah, I wasn't sure how we wanted to do this. Adding module dep temporarily seems ok I guess, as long as it goes away before a DW release is cut. But maybe beta releases are the cleaner option, even if a little more of a hassle, since that'll maintain the boundary between services/* artifacts and everything else?

brianloss: Well, remember the release of the underlying dependency can be completely independent of the datawave release. The problem with having a snapshot version is that there's no place for someone who does a git clone on the repo and builds it to get the snapshot (without the special knowledge to build the services modules). I'm thinking beta release. Maybe once the PR is approved, merge it in and immediately make the beta release of the library and update the dependency in the top level pom?

services/accumulo-utils/pom.xml
keith-ratcliffe: It was just to work around the missing runtime dependency on ```org.apache.commons.logging.LogFactory```

E.g., without that on the classpath you get the following during the test...
```
Caused by: java.lang.NoClassDefFoundError: org/apache/commons/logging/LogFactory
        at org.apache.hadoop.io.FastByteComparisons.<clinit>(FastByteComparisons.java:39)
        at org.apache.hadoop.io.WritableComparator.compareBytes(WritableComparator.java:187)
        at org.apache.accumulo.core.data.ByteSequence.compareTo(ByteSequence.java:116)
        at org.apache.accumulo.core.data.ByteSequence.compareTo(ByteSequence.java:24)
        at java.util.TreeMap.compare(TreeMap.java:1294)
        at java.util.TreeMap.put(TreeMap.java:538)
        at java.util.TreeSet.add(TreeSet.java:255)
        at java.util.AbstractCollection.addAll(AbstractCollection.java:344)
        at java.util.TreeSet.addAll(TreeSet.java:312)
        at java.util.TreeSet.<init>(TreeSet.java:160)
        at org.apache.accumulo.core.security.Authorizations.checkAuths(Authorizations.java:91)
```

services/accumulo-utils/src/main/java/datawave/accumulo/util/security/UserAuthFunctions.java
keith-ratcliffe: I like the approach. And on one hand, I agree that it's kinda tempting to remove the DatawaveUser dependency in this particular case. I feel like the main argument for keeping it though is that it leaves the door open for implementations to make their own user-dependent transformation decisions, which may not depend on the auths collection alone.

How do you feel about specifying both options in the interface? Then our "Default" impl could have the DatawaveUser version of getRequestedAuthorizations simply delegate to the "T authSupplier" version of the method (?)

pom.xml
keith-ratcliffe: That should work

services/accumulo-utils/src/main/java/datawave/accumulo/util/security/UserAuthFunctions.java
keith-ratcliffe: Or perhaps provide the "T authSupplier" version as a public static method on the interface? To include a ```..., boolean throwOnMissingAuths)``` arg to allow user to toggle between those two use cases as desired? That is, ```false``` would return the intersection, no matter what was in requestedAuths, while ```true``` would yield our default behavior

brianloss: Yeah, it's a toss up. I'm not sure there's much value to having the generic variants if we're going to keep the authorization-api dependency for DatawaveUser. @tomnelson, any thoughts?

services/accumulo-utils/pom.xml
keith-ratcliffe: Yeah, our spring-boot:2.0.4.RELEASE depends on spring-core:5.0.8.RELEASE, so it seems like adding spring-jcl:5.0.8.RELEASE with runtime scope *should* be safe.

I've definitely experienced microservice start up failures in my own environment due to transitive dependency on older log4j artifacts, which aren't compatible with spring-boot 2.x. Wasn't very fun to troubleshoot

services/accumulo-utils/src/main/java/datawave/accumulo/util/security/UserAuthFunctions.java
keith-ratcliffe: Yeah, I'm really ok either way, as long as users have some way to easily override our default behavior, if needed

services/accumulo-utils/pom.xml
keith-ratcliffe: Made a few pom updates, including adding runtime scope to spring-jcl. I had also noticed that I was getting commons-lang 2.x transitively from acccumulo-core, and that we were using commons-lang3 elsewhere. So, I updated for that as well

services/accumulo-utils/src/main/java/datawave/accumulo/util/security/UserAuthFunctions.java
keith-ratcliffe: @brianloss  Made a few tweaks based on your feedback, while leaving the DatawaveUser in place for now. Added...
```
static Authorizations getRequestedAuthorizations(String requestedAuths, Supplier<Collection<String>> authSupplier, boolean throwOnMissingAuths)
```

services/accumulo-utils/pom.xml
```diff
@@ -0,0 +1,117 @@
...
+            <groupId>gov.nsa.datawave.microservice</groupId>
+            <artifactId>authorization-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.accumulo</groupId>
+            <artifactId>accumulo-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
```
brianloss: I think this dependency isn't necessary. We already include Guava through Accumulo and/or the authorization API. It has `Strings.isNullOrEmpty`, which does the same thing as `StringUtils.isEmpty`, which is the only reason for including commons-lang.

keith-ratcliffe: Ah, good catch. Thanks!

pom.xml
brianloss: Ok, given that I have a PR that will be merged into this same module, and am planning additional work that will go there. Maybe the best solution for now is to add `<module>services/accumulo-utils</module>` to the modules section of the top level pom. Then when we're ready to make the first release, we'll tag the accumulo-utils module, update the version here to 1.0, and remove the module specification. This way, someone downloading and building datawave while it's still a snapshot version won't have problems building.  If you don't mind making/testing that change, then we can merge this in.

keith-ratcliffe: Done

keith-ratcliffe:
Sorry @ivakegg , my squash+force-push "dismissed [your] stale review"

brianloss:
Will merge since we had previous second approval--the only post-review change was the pom addition.

PR #245, opened 2018-11-07
refactor and extract scanner-api and typemetadata-api

tomnelson:
Changes to be committed:
	modified:   pom.xml
	modified:   services/build-parent/pom.xml
	new file:   services/scanner-api/pom.xml
	new file:   services/scanner-api/src/main/java/datawave/security/iterator/ConfigurableVisibilityFilter.java
	new file:   services/scanner-api/src/main/java/datawave/security/util/Minimizer.java
	new file:   services/scanner-api/src/main/java/datawave/security/util/ScannerHelper.java
	new file:   services/scanner-api/src/main/java/datawave/webservice/common/connection/ScannerBaseDelegate.java
	new file:   services/scanner-api/src/main/java/datawave/webservice/common/connection/ScannerDelegate.java
	new file:   services/typemetadata-api/pom.xml
	renamed:    warehouse/query-core/src/main/java/datawave/query/util/TypeMetadata.java -> services/typemetadata-api/src/main/java/datawave/query/util/TypeMetadata.java
	renamed:    warehouse/query-core/src/main/java/datawave/query/util/TypeMetadataHelper.java -> services/typemetadata-api/src/main/java/datawave/query/util/TypeMetadataHelper.java
	modified:   warehouse/query-core/pom.xml
	modified:   warehouse/query-core/src/main/java/datawave/query/iterator/FieldIndexOnlyQueryIterator.java
	modified:   warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
	modified:   warehouse/query-core/src/main/java/datawave/query/predicate/AncestorEventDataFilter.java
	modified:   warehouse/query-core/src/main/java/datawave/query/predicate/TLDEventDataFilter.java
	modified:   warehouse/query-core/src/main/java/datawave/query/tables/facets/FacetedVisitor.java
	modified:   warehouse/query-core/src/test/java/datawave/query/CompositeFunctionsTest.java
	modified:   warehouse/query-core/src/test/java/datawave/query/jexl/visitors/EventDataQueryExpressionVisitorTest.java
	modified:   warehouse/query-core/src/test/java/datawave/query/predicate/ValueToAttributesTest.java
	modified:   warehouse/query-core/src/test/java/datawave/query/util/MockMetadataHelper.java
	modified:   web-services/common-util/pom.xml
	modified:   web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java

PR #281, opened 2018-12-13
Refactor ScannerHelper and dependent classes to utils module.

brianloss:
This is a little bit of an overlap with what @tomnelson did in pr #245. It moves the ScannerHelper and its dependent classes to a separate module that can be used by microservices as well as the rest of datawave. Note that the root pom depends on 1.0-SNAPSHOT, but my intent would be upon approval of this pull request to release a 1.0 version and update the root pom to that. This code has been stable for some time, so it seems that the risk is low of releasing a 1.0 version and then having to immediately release a bugfix version.

I made sure to add the new module to the assembly descriptor so that the jar will get uploaded to HDFS and will be available on tservers (since there is an iterator that is included in this util package). Also, I checked usages and the minimize2 method from AuthorizationUtils wasn't used anywhere other than in the minimize method so I removed it from AuthorizationUtils entirely and made it private in the delegate class AuthorizationsMinimizer.

I have also run all of this through the tests in the quickstart folder. They execute queries, and therefore make use of this code path.


brianloss:
Ha, I hadn't noticed that I chose the same name as @keith-ratcliffe did for pr #280. It looks like the dependencies added by his change aren't _too_ bad (jackson and jwt libs to support the authorization-api project). I'm trying to be conscious of external dependencies here since this library will be used on tservers. We need to do one of:

1. Accept the extra dependencies and merge this with PR #280.
2. Rename the module.
3. Rename the module added in PR #280 to something like accumulo-web-utils or something else to indicate it's not intended for use on tservers.

I'm leaning towards the first choice. What do other people think?

keith-ratcliffe:
lol! In light of your point about potential deployment to tservers, I think maybe #3 makes the most sense.  Just let me know, and I'll update my PR

brianloss:
> Although, perhaps any risks associated with option 1 are easily mitigated. Wouldn't authorization-api have to be explicitly whitelisted in the assembly descriptor to ever make it into tserver context? And no reason we couldn't set it as `<scope>provided</scope>` or `<optional>true</optional>`, right?

Yeah, that's right. It looks like jackson dependencies have already made it to tservers anyway, so I think the only dependency we'd be adding is the JWT library.

`<optional>true</optional>` will exclude it from the transitive dependency list, which is an option, although I think I'm fine with just leaving the dependencies as-is.

services/accumulo-utils/src/main/java/datawave/security/util/AuthorizationsMinimizer.java
```diff
@@ -0,0 +1,67 @@
...
+            Collection<? extends Collection<String>> minimized = minimize2(allAuths);
+            authorizations = new LinkedHashSet<>(minimized.size());
+            for (Collection<String> a : minimized)
+                authorizations.add(new Authorizations(a.toArray(new String[a.size()])));
+        }
+        return authorizations;
+    }
+    
+    protected static Collection<? extends Collection<String>> minimize2(Collection<? extends Collection<String>> authorizations) {
+        if (authorizations.size() > 1) {
```
tomnelson: 'protected static' feels a little weird to me

brianloss: Fair enough. I can remove `protected` and just have it be package level access. Or, make it `private static`.

ivakegg:
Merging this into the same module as Keith's PR makes sense to me.

services/accumulo-utils/src/main/java/datawave/security/util/AuthorizationsMinimizer.java
```diff
@@ -0,0 +1,67 @@
...
+
+import org.apache.accumulo.core.security.Authorizations;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.TreeSet;
+
+public class AuthorizationsMinimizer {
```
ivakegg: document class

services/accumulo-utils/src/main/java/datawave/security/util/AuthorizationsMinimizer.java
```diff
@@ -0,0 +1,67 @@
...
+            
+            Collection<? extends Collection<String>> minimized = minimize2(allAuths);
+            authorizations = new LinkedHashSet<>(minimized.size());
+            for (Collection<String> a : minimized)
+                authorizations.add(new Authorizations(a.toArray(new String[a.size()])));
+        }
+        return authorizations;
+    }
+    
+    private static Collection<? extends Collection<String>> minimize2(Collection<? extends Collection<String>> authorizations) {
```
ivakegg: maybe instead of minimize and minize2 we use apply and applyToStrings.  This can be done in a separate PR.

services/accumulo-utils/src/main/java/datawave/security/util/AuthorizationsMinimizer.java
```diff
@@ -0,0 +1,68 @@
...
+    public static Collection<Authorizations> minimize(Collection<Authorizations> authorizations) {
+        if (authorizations.size() > 1) {
+            // Convert collection of Authorizations into a collection of String sets (the individual authorizations).
+            // Since we are adding to a LinkedHashSet, this will de-dupe any duplicate authorization sets.
+            final LinkedHashSet<TreeSet<String>> allAuths = authorizations.stream()
+                            .map(a -> a.getAuthorizations().stream().map(String::new).collect(Collectors.toCollection(TreeSet::new)))
+                            .collect(Collectors.toCollection(LinkedHashSet::new));
+            
+            // Go through the authorizations sets and remove any that are supersets of any other.
+            for (Iterator<TreeSet<String>> it = allAuths.iterator(); it.hasNext(); /* empty */) {
```
ivakegg: Do we want to check for a threshold wherein it is no longer worth doing an n^2 operation?

ivakegg: Check size before doing containsall

ivakegg: HashSet vs TreeSet?

brianloss: I'm not sure on the threshold. I'm kind of thinking that a threshold big enough to worry about the n^2 setup would mean we'll have so many authorization sets that running each entry through that many visibility filters would be so slow as to be unusable. I feel that that's one of those "Doctor, it hurts when I do that." scenarios.
Will make the other changes.

Issue #228, opened 2018-10-10
IteratorBuildingVisitor.buildLiteralRange generates invalid range


FineAndDandy:
IteratorBuildingVisitor.buildLiteralRange() should produce a range of \u0000 to \uFFFF when there is no leading literal since a leading wildcard effectively produces a full table scan for the field index

Issue #227, opened 2018-10-10
TermFrequencyIndexIterator generates incorrect scanRange


FineAndDandy:
the endKey should be generated using he stopKeyParser rather than the startKeyParser. The endKey should also be limited by the fi key start and end fields

PR #290, opened 2018-12-20
DATAWAVE-226: Use TLDTermFrequencyAggregator

FineAndDandy:
* DATAWAVE-227: Generate correct TLDTermFrequency scan range
 * DATAWAVE-228: Generate correct range for leading wildcard TF

see #226 #227 #228 

Since these tickets all impact correctness of TF based queries, fixed all together for inclusion of correct unit tests


warehouse/query-core/src/main/java/datawave/query/jexl/functions/TermFrequencyAggregatorFactory.java
```diff
@@ -0,0 +1,9 @@
+package datawave.query.jexl.functions;
+
+import datawave.query.predicate.EventDataQueryFilter;
+
+import java.util.Set;
+
+public interface TermFrequencyAggregatorFactory {
+    TermFrequencyAggregator create(Set<String> fieldsToAggregate, EventDataQueryFilter attrFilter, int maxNextCount);
```
jwomeara: Instead of passing the Iterator Building Visitor to the iterator builders, and calling create within the iterator builders, why not create the the TermFrequencyAggregator in the iterator building visitor and pass that into the iterator builder.  It feels odd to be passing the iterator building visitor around to the iterator builders.  In that case, do we even need the interface?

FineAndDandy:
I found a bug in this code, expect another revision

warehouse/query-core/src/main/java/datawave/query/jexl/functions/TermFrequencyAggregatorFactory.java
FineAndDandy: I can refactor buildExceededFromTermFrequency() to call some new method in the IteratorBuildingVisitor which can be overridden in the TLD version to generate the correct TermFrequencyAggregator.

jwomeara: That sounds good to me!

FineAndDandy:
this is now ready for review again

FineAndDandy:
> I found a bug in this code, expect another revision

This is now ready for review

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyIndexIterator.java
```diff
@@ -194,6 +205,21 @@ public void next() throws IOException {
                 continue;
             }
             
+            // only inspect the values specified in the range since a broad row or uid range will potentially go cross document
+            if (scanRange.isStartKeyInclusive() && key.getFieldValue().compareTo(startKeyParser.getFieldValue()) < 0) {
```
ivakegg: Do we need to check the field name or are we guaranteed the range is reduced to those of interest?

warehouse/query-core/src/main/java/datawave/query/jexl/functions/TermFrequencyAggregator.java
```diff
@@ -60,4 +68,45 @@ protected Key getSeekStartKey(Key current, ByteSequence pointer) {
...
+    /**
+     * Limit keep fields to those that are index only or if there is no filter specified
+     * 
+     * @param topKey
+     * @param fieldNameValue
+     * @return
+     */
+    @Override
+    protected boolean toKeep(Key topKey, Tuple2<String,String> fieldNameValue) {
+        return fieldsToKeep == null || filter == null || fieldsToKeep.contains(JexlASTHelper.removeGroupingContext(fieldNameValue.first()));
```
ivakegg: Since the filter is only being used for the addToDoc concept, is it even to be considered here?  As you have it written here, If we have fieldsToKeep specified but no filter then we keep all fields?

warehouse/query-core/src/main/java/datawave/query/jexl/functions/TermFrequencyAggregator.java
```diff
@@ -60,4 +68,45 @@ protected Key getSeekStartKey(Key current, ByteSequence pointer) {
...
+     * Only aggregate tf fields that are part of the returned document or necessary for query evaluation, not all TF
+     * 
+     * @param topKey
+     * @param fieldNameValue
+     * @param toKeep
+     * @return
+     */
+    @Override
+    protected boolean addToDoc(Key topKey, Tuple2<String,String> fieldNameValue, boolean toKeep) {
+        return toKeep && (filter == null || filter.apply(Maps.immutableEntry(topKey, StringUtils.EMPTY)));
```
ivakegg: should this be toKeep || (filter....) ?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
```diff
@@ -1147,6 +1160,29 @@ public void ivarateFilter(JexlNode source, Object data) throws IOException {
...
+        
+        return buildTermFrequencyAggregator(fieldsToKeep, wrapped, maxNextCount);
+    }
+    
+    protected TermFrequencyAggregator buildTermFrequencyAggregator(Set<String> fieldsToKeep, EventDataQueryFilter filter, int maxNextCount) {
+        return new TermFrequencyAggregator(fieldsToKeep, filter, maxNextCount);
+    }
+    
+    protected EventDataQueryFilter createWrappedTermFrequencyFilter(JexlNode node, EventDataQueryFilter existing) {
+        EventDataQueryFilter expressoinFilter = new EventDataQueryExpressionFilter(node, typeMetadata);
```
ivakegg: Does this filter express "oin" often?

warehouse/query-core/src/main/java/datawave/query/predicate/ChainableEventDataQueryFilter.java
```diff
@@ -0,0 +1,231 @@
...
+        
+        while (iterator.hasNext() && result) {
+            result = iterator.next().keep(k);
+        }
+        
+        return result;
+    }
+    
+    /**
+     * Get the maximum start key from all filters
```
ivakegg: hmm, are we sure we want the maximum?

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyIndexIterator.java
FineAndDandy: line 176 checks that the cq ends with the field name, technically we should test \0<fieldname> for 100% confidence

warehouse/query-core/src/main/java/datawave/query/predicate/ChainableEventDataQueryFilter.java
FineAndDandy: I assumed we'd want the most restrictive of the filter stack for all methods. Do you think this is the wrong assumption?

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyIndexIterator.java
FineAndDandy: I believe the IteratorBuildingVisitor which ultimately creates the range will prevent creating a range outside the target field, but we can update the TermFrequencyIndexIterator to protect against it anyway

warehouse/query-core/src/main/java/datawave/query/jexl/functions/TermFrequencyAggregator.java
FineAndDandy: This is definitely not a superset, I will revise

warehouse/query-core/src/main/java/datawave/query/jexl/functions/TermFrequencyAggregator.java
FineAndDandy: will remove filter == null as a condition

FineAndDandy:
this should be ready for final review

jwomeara:
Looks good to me.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/IdentityAggregator.java
```diff
@@ -102,11 +104,12 @@ public Key apply(SortedKeyValueIterator<Key,Value> itr, Document doc, AttributeF
             
             Attribute<?> attr = attrs.create(fieldNameValue.first(), fieldNameValue.second(), topKey, true);
             // only keep fields that are index only and pass the attribute filter
-            boolean toKeep = (fieldsToKeep == null || fieldsToKeep.contains(JexlASTHelper.removeGroupingContext(fieldNameValue.first())))
-                            && (filter == null || filter.keep(topKey));
+            boolean toKeep = toKeep(topKey, fieldNameValue);
             attr.setToKeep(toKeep);
-            doc.put(fieldNameValue.first(), attr);
-            key = nextKey;
+            // only add to the document if the filter passes or we are keeping it
```
ivakegg: I think this comment is wrong: // only add to the document if the filter passes
the apply is the only gateway to adding to the document.  After adding, the keep flag is set based on the toKeep method.

ivakegg: We may consider removing the addToDoc method and simply use the filter here to avoid confusion.

warehouse/query-core/src/main/java/datawave/query/predicate/EventDataQueryFieldFilter.java
```diff
@@ -57,7 +62,10 @@ public void startNewDocument(Key document) {
      */
     @Override
     public boolean keep(Key k) {
-        return true;
+        DatawaveKey key = new DatawaveKey(k);
+        
+        // keep as long as it is part of the event or it is part of the query
```
ivakegg: So apply means add to document, keep means send back to user (but only those actually added to the document).
This says we are sending back to the user anything we add to the document.  Is that what we want?

warehouse/query-core/src/main/java/datawave/query/predicate/EventDataQueryExpressionFilter.java
```diff
@@ -56,7 +70,10 @@ public void startNewDocument(Key document) {
     
     @Override
     public boolean keep(Key k) {
-        return true;
+        DatawaveKey key = new DatawaveKey(k);
+        
+        // keep as long as it is part of the event or it is part of the query
+        return !nonEventFields.contains(key.getFieldName()) || apply(new AbstractMap.SimpleEntry<>(k, null));
```
ivakegg: So apply means add to document, keep means send back to user (but only those actually added to the document).
This says we are sending back to the user anything we add to the document. Is that what we want?

warehouse/query-core/src/main/java/datawave/query/predicate/ChainableEventDataQueryFilter.java
```diff
@@ -0,0 +1,231 @@
...
+     * 
+     * @param k
+     * @return
+     */
+    @Override
+    public boolean keep(Key k) {
+        Iterator<EventDataQueryFilter> iterator = filters.iterator();
+        boolean result = true;
+        
+        while (iterator.hasNext() && result) {
```
ivakegg: excellent...this makes sense to me

warehouse/query-core/src/main/java/datawave/query/predicate/ChainableEventDataQueryFilter.java
```diff
@@ -0,0 +1,231 @@
...
+     * @param entry
+     * @return
+     */
+    @Override
+    public boolean apply(@Nullable Map.Entry<Key,String> entry) {
+        Iterator<EventDataQueryFilter> iterator = filters.iterator();
+        boolean result = true;
+        
+        while (iterator.hasNext() && result) {
+            result = iterator.next().apply(entry);
```
ivakegg: excellent

warehouse/query-core/src/main/java/datawave/query/jexl/functions/TermFrequencyAggregator.java
```diff
@@ -60,4 +69,30 @@ protected Key getSeekStartKey(Key current, ByteSequence pointer) {
...
+        // test that the returned document has values required for evaluation
+        for (Attribute attr : doc.getDictionary().values()) {
+            if (attr.isToKeep()) {
+                necessary = true;
+                break;
+            }
+        }
+        
+        // only return a key if something was added to the document necessary for evaluating the query
+        if (!necessary) {
```
ivakegg: why is attr.isToKeep() the determination that it is necessary for evaluation.  The fact that it is in the document means it is necessary in the first place if the underlying filters did the correct thing.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/IdentityAggregator.java
```diff
@@ -104,8 +107,11 @@ public Key apply(SortedKeyValueIterator<Key,Value> itr, Document doc, AttributeF
             boolean toKeep = (fieldsToKeep == null || fieldsToKeep.contains(JexlASTHelper.removeGroupingContext(fieldNameValue.first())))
                             && (filter == null || filter.keep(topKey));
             attr.setToKeep(toKeep);
-            doc.put(fieldNameValue.first(), attr);
-            key = nextKey;
+            
+            // Anything that is being kept has to be added to the doc to be returned, if we aren't keeping only add to the doc if necessary for evaluation
+            if (toKeep || (filter == null || filter.apply(new AbstractMap.SimpleEntry<>(topKey, null)))) {
+                doc.put(fieldNameValue.first(), attr);
```
ivakegg: I am still confused by this.  the apply method means we need it in the document (either for evaluation or to send back to the user).  Keep is only the determination as to whether it is actually sent back to the user.  I would think filter.apply is enough here.

warehouse/query-core/src/main/java/datawave/query/predicate/EventDataQueryExpressionFilter.java
ivakegg: If this is actually dead code, then return true.

warehouse/query-core/src/main/java/datawave/query/predicate/EventDataQueryFieldFilter.java
ivakegg: If this is dead code then just return true.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/TermFrequencyAggregator.java
FineAndDandy: yep, this slipped through. I will remove it

warehouse/query-core/src/main/java/datawave/query/jexl/functions/IdentityAggregator.java
FineAndDandy: The apply method only means we need it in the document, not that it needs to be sent back to the user. keep and apply are potentially mutually exclusive. This check will ensure we don't add something that is not necessary for either accidentally. One case I need to revisit is if its possible to have an index fully satisfied query, and how this would behave in that situation

FineAndDandy:
The PR now captures the following rules when determining if a field/value should be kept or rejected.

Type | Keep
----- | -----------
Event | true
Index only | true
Non event | false
Child event | true
Child index only | true
Child non-event | true

warehouse/query-core/src/main/java/datawave/query/predicate/EventDataQueryFieldFilter.java
FineAndDandy: set in next commit

warehouse/query-core/src/main/java/datawave/query/predicate/EventDataQueryExpressionFilter.java
FineAndDandy: next commit will have true

warehouse/query-core/src/main/java/datawave/query/jexl/functions/IdentityAggregator.java
FineAndDandy: are you good with this?

FineAndDandy:
@ivakegg please take another look at this when you have some time. I believe I have addressed all comments and achieved functional correctness.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/IdentityAggregator.java
ivakegg: I am good with this

warehouse/query-core/src/main/java/datawave/query/jexl/functions/TermFrequencyAggregator.java
```diff
@@ -60,4 +65,20 @@ protected Key getSeekStartKey(Key current, ByteSequence pointer) {
...
         return new Key(current.getRow(), current.getColumnFamily(), new Text(pointer + Constants.NULL_BYTE_STRING + Constants.MAX_UNICODE_STRING));
     }
+    
+    @Override
+    public Key apply(SortedKeyValueIterator<Key,Value> itr, Document doc, AttributeFactory attrs) throws IOException {
+        Key key = super.apply(itr, doc, attrs);
+        
+        // only return a key if something was added to the document, documents that only contain Document.DOCKEY_FIELD_NAME as they found nothing of value to
+        // aggregate
+        if (doc.size() == 1 && doc.get(Document.DOCKEY_FIELD_NAME) != null) {
```
ivakegg: I am pretty sure we have something that drops empty documents in the end.  Are you sure we need this?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
```diff
@@ -376,8 +378,18 @@ else if (source instanceof ASTNRNode)
...
             
             LiteralRange<String> range = new LiteralRange<>(JexlASTHelper.getIdentifier(node), NodeOperand.AND);
-            range.updateLower(analyzer.getLeadingOrTrailingLiteral(), true);
-            range.updateUpper(analyzer.getLeadingOrTrailingLiteral() + Constants.MAX_UNICODE_STRING, true);
+            if (!analyzer.isLeadingLiteral()) {
+                // if the range is a leading wildcard we have to seek over the whole range since it's forward indexed only
+                range.updateLower(Constants.NULL_BYTE_STRING, true);
+                range.updateUpper(Constants.MAX_UNICODE_STRING, true);
+            } else {
+                range.updateLower(analyzer.getLeadingOrTrailingLiteral(), true);
```
ivakegg: If we know we should be using the leading literal here, use getLeadingLiteral().

FineAndDandy: will update

warehouse/query-core/src/main/java/datawave/query/jexl/functions/TermFrequencyAggregator.java
FineAndDandy: This is explicitly necessary for the negated leading wildcard case when dealing with a document specific range. This generates a scan over all keys for that field on that document. Without this check if the aggregator runs, but finds no specific matches there will still be a docKey generated and this will be returned back up through the chain resulting in a missed hit. 

Consider: '!(FIELD =~ '.*z)' 

If a document d has FIELD=b, this will in the TermFrequencyIndexIterator must scan against the entire range \0 to MAX for this document. This will cause a document to be generated for our docKey b even though it doesn't match the entire regex (verified in the filter within the aggregation). If the document were going all the way to evaluation this wouldn't matter, but it will be skipped when popping off the nestedIterators due to the way NOT is short circuited. Since in this case the Range must be overly broad to prevent missing a match, we have to rely on the aggregators which employ the filter logic to protect us. 

This check ensures that post aggregation (and application of the filter) there was actually something to get. Since the field is negated, the presence of the docKey will cause the document to be excluded, even though it satisfies the query. 

An alternative to adding this check in the TermFrequencyAggregator would be to add this check to the TermFrequencyIndexIterator following the aggregation.

ivakegg: I think I understand.  This looks like a great explanation to perhaps put in the code here.

PR #291, opened 2018-12-21
extraction of zookeeper cache code modeled after accumulo-utils module.

tomnelson:
Changes to be committed:
	modified:   pom.xml
	modified:   services/build-parent/pom.xml
	new file:   services/zookeeper-utils/pom.xml
	renamed:    web-services/common/src/main/java/datawave/webservice/common/cache/SharedBoolean.java -> services/zookeeper-utils/src/main/java/datawave/zookeeper/cache/util/SharedBoolean.java
	renamed:    web-services/common/src/main/java/datawave/webservice/common/cache/SharedBooleanListener.java -> services/zookeeper-utils/src/main/java/datawave/zookeeper/cache/util/SharedBooleanListener.java
	renamed:    web-services/common/src/main/java/datawave/webservice/common/cache/SharedBooleanReader.java -> services/zookeeper-utils/src/main/java/datawave/zookeeper/cache/util/SharedBooleanReader.java
	renamed:    web-services/common/src/main/java/datawave/webservice/common/cache/SharedCacheCoordinator.java -> services/zookeeper-utils/src/main/java/datawave/zookeeper/cache/util/SharedCacheCoordinator.java
	renamed:    web-services/common/src/main/java/datawave/webservice/common/cache/SharedTriState.java -> services/zookeeper-utils/src/main/java/datawave/zookeeper/cache/util/SharedTriState.java
	renamed:    web-services/common/src/main/java/datawave/webservice/common/cache/SharedTriStateListener.java -> services/zookeeper-utils/src/main/java/datawave/zookeeper/cache/util/SharedTriStateListener.java
	renamed:    web-services/common/src/main/java/datawave/webservice/common/cache/SharedTriStateReader.java -> services/zookeeper-utils/src/main/java/datawave/zookeeper/cache/util/SharedTriStateReader.java
	modified:   warehouse/query-core/src/main/java/datawave/query/util/MetadataHelperCacheManagementListener.java
	modified:   warehouse/query-core/src/main/java/datawave/query/util/MetadataHelperUpdateHdfsListener.java
	modified:   web-services/common/pom.xml
	modified:   web-services/common/src/main/java/datawave/webservice/common/cache/AccumuloTableCache.java
	modified:   web-services/common/src/main/java/datawave/webservice/common/cache/BaseTableCache.java
	modified:   web-services/common/src/main/java/datawave/webservice/common/cache/TableCache.java
	modified:   web-services/common/src/test/java/datawave/webservice/common/cache/SharedCacheCoordinatorTest.java
	modified:   web-services/common/src/test/java/datawave/webservice/common/curator/TestSharedCacheCoordinator.java


brianloss:
I'm not sure about this one. It pulls in deltaspike, which is a CDI library. We haven't set up to use any kind of CDI in any services implementation (not all micro services have to use Spring Boot, but things are easier if they do), so I wonder how useful this module would be outside of Wildfly. Thoughts?

tomnelson:
Yeah, that's not good.
I just need to make the SharedTriState* classes visible to a microservice
for the type metadata in hdfs.

brianloss:
I thought about this a little bit, and it looks like deltaspike is only used for passing the constructor parameters to SharedCacheCoordinator. It's a little bit of extra work, but one solution is to make an `@Produces` that lives in the old location for this code (so it will only be used by Wildfly -- see MetricsCdiConfiguration for an example of the idea). The producer class can create the injectable instance of SharedCacheCoordinator and receive the parameters that are annotated with `@ConfigProperty`. Then, you could create a new properties class that is annotated with `@ConfigurationProperties` and supplies the same properties. This class would be used by spring boot and values can be supplied by the configuration service.

services/zookeeper-utils/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>datawave-microservice-parent</artifactId>
+        <groupId>gov.nsa.datawave.microservice</groupId>
+        <version>1.3</version>
+    </parent>
+    <artifactId>zookeeper-utils</artifactId>
+    <version>1.0-SNAPSHOT</version>
+    <properties>
+        <version.curator>2.11.0</version.curator>
```
brianloss: Looks like this version of curator is way out of date. If we're moving things around (and re-testing anyway), it might be worth moving to a more recent version at the same time. I think the only reason that might not be possible is zookeeper version issues. We need to be compatible with whatever zookeeper is running along with Accumulo 1.9.2 and on the primary user systems.

PR #293, opened 2018-12-22
Fixes #292 GroupingRequiredFilterFunctionsDescriptor doesn't properly define addFilters

### LINKED ISSUE #292: GroupingRequiredFilterFunctionsDescriptor doesn't properly define addFilters
Most query functions expand the query to add clauses outside of the function that identify the fields and values/ranges/regexs required for query evaluation. In cases where the function does not do this, it is necessary for the `JexlArgumentDescriptor` implementation corresponding to the function to properly implement `addFilter`. This allows the `EventDataQueryExpressionVisitor` to determine the fields/values required by that function for query evaluation and ensure the proper keys are loaded into the jexl context. 

The `GroupingRequiredFilterFunctionsDescriptor` does not properly implement `addFilter`. As a result queries that use the `EventDataQueryExpressionVisitor` and include one or more of the 'groups' functions will not be evaluated properly - the keys required to properly evaluate the function won't be loaded by the `EventDataQueryExpressionVisitor`.

drewfarris:


warehouse/query-core/src/main/java/datawave/query/jexl/functions/GroupingRequiredFilterFunctionsDescriptor.java
```diff
@@ -50,7 +54,36 @@ public JexlNode getIndexQuery(ShardQueryConfiguration config, MetadataHelper hel
         
         @Override
         public void addFilters(AttributeFactory attributeFactory, Map<String,EventDataQueryExpressionVisitor.ExpressionFilter> filterMap) {
-            // noop, covered by getIndexQuery (see comments on interface)
+            FunctionJexlNodeVisitor functionMetadata = new FunctionJexlNodeVisitor();
+            node.jjtAccept(functionMetadata, null);
+            Map<String, String> fieldValues = new HashMap<>();
+
+            if (functionMetadata.name().equals("atomValuesMatch")) {
+                // special case
```
tomnelson: I'd like to update atomValuesMatch to take a varargs array after this PR is merged (and then update this code to get all the fields)

PR #294, opened 2018-12-31
Encapsulate variable for ease of extensibility.

apmoriarty:


jschmidt10:
I know there is precedent for this but I think it's not ideal to expose class members, even to subclasses. This allows references to the member to leak across the code base. If there are constraints on what the value can hold, you rely on other classes to maintain those constraints. I'd suggest adding a method to add filters.

apmoriarty:
> 
> 
> I know there is precedent for this but I think it's not ideal to expose class members, even to subclasses. This allows references to the member to leak across the code base. If there are constraints on what the value can hold, you rely on other classes to maintain those constraints. I'd suggest adding a method to add filters.

I only need access to the filters for some additional setup work not covered in the DataTypeOverrideHelper.setup() method call. Adding a getter/setter should suffice, negating the need for protected variables.

jschmidt10:
Sounds good! Thanks!

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/DataTypeOverrideHelper.java
```diff
@@ -208,4 +208,12 @@ public void closeFilters() {
             this.filters = null;
         }
     }
+
+    public List<? extends RecordFilter> getFilters() {
+        return filters;
+    }
```
tomnelson: Do you want callers to be allowed to mutate the list of filters?
If not, you could return an unmodifiable copy

apmoriarty: I don't believe so, I'll make the update

Issue #287, opened 2018-12-20
TLD query logic can produce keys outside the original range


ivakegg:
The TLD query logic when used with the FinalDocumentTrackingIterator can produce keys outside of the initial range which will subsequently cause errors when rebuilding the TLDQueryIterator and subsequently attempting to reseek the iterator (range with start key after end key).

PR #297, opened 2019-01-04
re #287: Rebuild errors

ivakegg:
Created rebuild mechanism for test cases and fixed various rebuild bugs discovered as a result
Added some TLD queries test cases that demostrated the original bug which is now fixed


contrib/in-memory-accumulo/src/main/java/datawave/accumulo/inmemory/InMemoryBatchScanner.java
```diff
@@ -69,6 +88,27 @@ public void setRanges(Collection<Range> ranges) {
...
+
+    @Override
+    public Iterator<Entry<Key,Value>> rebuild(Key lastKey) {
+        // rebuild the ranges
+        List<Range> newRanges = new ArrayList<>();
+        boolean found = false;
+        for (Range range : ranges) {
+            if (found) {
+                newRanges.add(range);
+            } else if (range.contains(lastKey)) {
```
brianloss: Won't this potentially remove ranges? If there were, say 3 ranges. The first does not contain lastKey, the second did, and the third did not. For the first range, found would be false and also range.contains(lastKey) is false, so the range wouldn't be added to the newRanges set.

bmwmaestoso: It is not clear to me what the purpose of the rebuild as the interface does not have any docs. As the code exists it includes all ranges starting with the lastKey (inclusive). I believe this is the intended result based upon the brief discussion we had concerning the expected  behavior.

contrib/in-memory-accumulo/src/main/java/datawave/accumulo/inmemory/ScannerRebuilder.java
```diff
@@ -0,0 +1,11 @@
+package datawave.accumulo.inmemory;
+
+import org.apache.accumulo.core.data.Key;
+import org.apache.accumulo.core.data.Value;
+
+import java.util.Iterator;
+import java.util.Map;
+
+public interface ScannerRebuilder {
+    public Iterator<Map.Entry<Key, Value>> rebuild(Key lastKey);
```
bmwmaestoso: Perhaps a javadoc for this for clarity.

contrib/in-memory-accumulo/src/main/java/datawave/accumulo/inmemory/InMemoryBatchScanner.java
ivakegg: In the IteratorChain used in the previous method, the ranges are processed in order.  So the idea is to subset the ranges starting with the first range that contains the lastKey (which was the last key returned by the iterator).  All ranges up to the range containing the lastKey need to be dropped, and the the range containing the key is reset to start with lastKey, and then all subsequent ranges are readded to the list.  I will add documentation to state the intention here.

ivakegg: I will add documentation as the code is not as obvious as I had thought.

warehouse/query-core/src/test/java/datawave/query/RebuildingScannerTestHelper.java
```diff
@@ -0,0 +1,565 @@
...
+        return new RebuildingConnector((InMemoryConnector) (i.getConnector(principal, token)), teardown);
+    }
+    
+    public interface TeardownListener {
+        boolean teardown();
+        
+        boolean checkConsistency();
+    }
+    
+    public static class RebuildingIterator implements Iterator<Map.Entry<Key,Value>> {
```
ivakegg: Should I split out the inner classes here?

contrib/in-memory-accumulo/src/main/java/datawave/accumulo/inmemory/InMemoryBatchScanner.java
brianloss: It needs to be more than processing the ranges in order, doesn't it? The ranges have to be sorted (and not overlapping at all). Do we have that guarantee?

ivakegg: That is a very astute observation.  However sorting is not required, only non-overlapping.  Since the ranges are processed in the same order in the list, I only need to find the first one containing the key returned.  In terms of the non-overlapping assumption, the cost of not taking that into account would be duplicate keys being returned.  For the sake of the test cases, I do not believe this needs to be taken into account.  However to be thorough, I probably should.  This will take me a little longer to do.

brianloss: Good point. I think I was thinking of threading going on. Anyway, never mind about the sorting. For overlapping ranges, if the real batch scanner merges overlapping ranges then I think we should do one of the following:
1. Merge overlapping ranges here too.
2. Detect that there are overlapping ranges and throw an exception so the test case will at least fail
If merging looks like it's going to be a lot of work, then I would do the second since that should be trivial. If we're actually sending overlapping ranges in any test case, I'd rather detect it and have test cases fail then get duplicate keys back.

warehouse/query-core/src/main/java/datawave/core/iterators/GlobalIndexTermMatchingIterator.java
```diff
@@ -87,6 +87,17 @@ public void seek(Range range, Collection<ByteSequence> columnFamilies, boolean i
         this.scanCFs = columnFamilies;
         this.scanInclusive = inclusive;
         getSource().seek(range, columnFamilies, inclusive);
+        
+        // if we have been reseeked after being torn down, and we are only returning unique terms, then advance to the next unique row/cf
+        if (!range.isStartKeyInclusive() && uniqeTermsOnly && getSource().hasTop()) {
```
FineAndDandy: why are uniqueTerms a condition of a rebuild?

warehouse/query-core/src/main/java/datawave/query/tld/TLDQueryIterator.java
```diff
@@ -171,17 +171,11 @@ public void seek(Range range, Collection<ByteSequence> columnFamilies, boolean i
             if (!startKey.equals(range.getStartKey())) {
                 Key endKey = range.getEndKey();
                 boolean endKeyInclusive = range.isEndKeyInclusive();
-                // if the start key is outside of the range, then reset the end key to the next key
-                if (range.afterEndKey(startKey)) {
-                    endKey = startKey.followingKey(PartialKey.ROW_COLFAM);
-                    endKeyInclusive = false;
-                }
                 range = new Range(startKey, false, endKey, endKeyInclusive);
```
FineAndDandy: wont this restart the current document?

contrib/in-memory-accumulo/src/main/java/datawave/accumulo/inmemory/ScannerRebuilder.java
ivakegg: done

contrib/in-memory-accumulo/src/main/java/datawave/accumulo/inmemory/InMemoryBatchScanner.java
ivakegg: done

ivakegg: I ensure non-overlapping in the setRanges method above.  Done.

warehouse/query-core/src/main/java/datawave/core/iterators/GlobalIndexTermMatchingIterator.java
ivakegg: If we are not dealing with uniqueTerms, then we can simple continue scanning per the supplied range to produce the correct results.  Only when dealing with uniqueTerms will we need to advance forward on a rebuild.

warehouse/query-core/src/main/java/datawave/query/tld/TLDQueryIterator.java
ivakegg: nope.  The TLD.getNextParentKey ensures that.  I am just removing the nasty logic that would advance the end key resulting in a range mutually exclusive with the original range.  That is where we got into trouble because the "FinalDocument" key would return a key using the start key of this range which would result in the iterator returning a key outside the original range.  I would rather fail here quickly if we attempt to create a start key past the original end key.  Setting the "isTld" flag correctly results in the TupleToRange class creating a range that is large enough that this will not happen.

warehouse/query-core/src/test/java/datawave/query/testframework/AccumuloSetupHelper.java
```diff
@@ -89,14 +90,28 @@ public AccumuloSetupHelper(final Collection<DataTypeHadoopConfig> types, FileTyp
      */
     public Connector loadTables(final Logger parentLog) throws AccumuloException, AccumuloSecurityException, IOException, InterruptedException,
                     TableExistsException, TableNotFoundException, URISyntaxException {
+        return loadTables(parentLog, RebuildingScannerTestHelper.TEARDOWN.EVERY_OTHER);
```
bmwmaestoso: All of the exiting tests will call this loadTables method. Perhaps I am missing but I am not sure why the teardown would be EVERY_OTHER. I am assuming this is the current behavior.

warehouse/query-core/src/test/java/datawave/query/testframework/QueryLogicTestHarness.java
```diff
@@ -86,9 +91,16 @@ public void assertLogicResults(BaseQueryLogic<Map.Entry<Key,Value>> logic, Colle
                         }
                     }
                     Assert.assertEquals(AttributeFactory.getKeepers(types), types);
+                } else {
+                    count++;
                 }
             }
             
+            // ignore empty documents (possible when only passing FinalDocument back)
+            if (count == 0) {
```
bmwmaestoso: I assume something from the updates in this pull request will return empty documents.

warehouse/query-core/src/test/java/datawave/query/tld/TLDNormalFlattenQueryTest.java
```diff
@@ -0,0 +1,39 @@
...
+    
+    public TLDNormalFlattenQueryTest() {}
+    
+    @BeforeClass
+    public static void filterSetup() throws Exception {
+        Collection<DataTypeHadoopConfig> dataTypes = new ArrayList<>();
+        dataTypes.add(flatten);
+        
+        final AccumuloSetupHelper helper = new AccumuloSetupHelper(dataTypes, FileLoaderFactory.FileType.JSON);
+        connector = helper.loadTables(log, RebuildingScannerTestHelper.TEARDOWN.EVERY_OTHER);
```
bmwmaestoso: How is this different from the base class? Isn't NormalFlattenQueryTest going to also use the EVERY_OTHER teardown?

warehouse/query-core/src/test/java/datawave/query/testframework/AccumuloSetupHelper.java
ivakegg: I am using EVERY_OTHER instead of the RANDOM one to ensure we have deterministic test cases.  I could to ALWAYS, but I need to test both the teardown cases and the non-teardown cases.  Hence tearing down the iterator stack every other time seemed like a good alternative.  In real life, the tear downs would happen at seemingly random times.

PR #298, opened 2019-01-07
Move StringUtils and DateHelper to common-utils module.

brianloss:
Also switch uses of joda time to built-in java 8 time libs.

**Note:** Since this is intended to be a limited library and I don't forsee many other classes going in here (certainly not in the near term, anyway), I was planning to release version 1.0 as soon as this PR is approved.


services/common-utils/README.md
```diff
@@ -0,0 +1,8 @@
+# Common-Utils Library
+
+This library holds low-level common utility classes that are used by iterators
+on tablet servers, microservices, and the legacy Wildfly web service.
+
+**WARNING:** Think very carefully about adding any dependencies to this
+library. Only low-level additions with little or no external dependencies
+should be added.
```
ivakegg: extend this comment to include the reasoning: namely that it is expected many other services will be dependent on this one, and as such and additional dependencies or changes will potentially affect many other microservices etc.

warehouse/metrics-core/src/main/java/datawave/metrics/mapreduce/MetricsIngester.java
```diff
@@ -258,8 +259,7 @@ protected int launchErrorsJob(Job job, Configuration conf) throws IOException, I
             cf = iterKey.getColumnFamily().toString();
             try {
                 dateObj = DateHelper.parseTimeExactToSeconds(cf);
-                Date dateObjNext = (Date) dateObj.clone();
-                dateObjNext.setHours(dateObj.getHours() + 1);
+                Date dateObjNext = DateHelper.addDays(dateObj, 1);
```
ivakegg: addHours here

PR #303, opened 2019-01-10
fix server side grouping when iterators are violently torn down and recreated

tomnelson:
Changes to be committed:
	modified:   warehouse/query-core/src/test/java/datawave/query/GroupingTest.java


warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
```diff
@@ -153,6 +200,76 @@ public boolean hasNext() {
         return null;
     }
     
+    // @formatter:off
```
brianloss: Note that this only keeps the javadoc source from wrapping. The generated javadoc will still be wrapped, so you likely want to add the appropriate HTML markers to it to format it in a way that will be readable.

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
```diff
@@ -153,6 +200,76 @@ public boolean hasNext() {
...
+     */
+    // @formatter:on
+    private void flatten(List<Document> documents) {
+        if (log.isTraceEnabled()) {
+            log.trace("flatten documents:" + documents);
+        }
+        Document theDocument = new Document(documents.get(0).getMetadata(), true);
+        int context = 0;
+        for (Document document : documents) {
+            log.info("document:" + document);
```
brianloss: Either switch to slf4j and change this to `log.info("document: {}",document);` or guard with a check on the log level so we don't have to perform a toString on document unless it's going to be printed. Also, do you really want this to be info level?

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
```diff
@@ -153,6 +200,76 @@ public boolean hasNext() {
...
+     *
+     *
+     * @param documents
+     */
+    // @formatter:on
+    private void flatten(List<Document> documents) {
+        if (log.isTraceEnabled()) {
+            log.trace("flatten documents:" + documents);
+        }
+        Document theDocument = new Document(documents.get(0).getMetadata(), true);
```
brianloss: Are we guaranteed to never have an empty documents list?

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
```diff
@@ -214,22 +331,30 @@ private void getListKeyCounts(Entry<Key,Document> entry) {
...
         Set<String> expandedGroupFieldsList = new LinkedHashSet<>();
         // if the incoming Documents have been aggregated on the tserver, they will have a COUNT field.
         // use the value in the COUNT field as a loop max when the fields are put into the multiset
         // During the flush operation, a new COUNT field will be created based on the number of unique
         // field sets in the multiset
-        if (entry.getValue().getDictionary().containsKey("COUNT")) {
-            TypeAttribute countTypeAttribute = ((TypeAttribute) entry.getValue().getDictionary().get("COUNT"));
-            count = ((BigDecimal) countTypeAttribute.getType().getDelegate()).intValue();
+        Map<String,Attribute<? extends Comparable<?>>> dictionary = entry.getValue().getDictionary();
+        Set<String> countKeys = dictionary.keySet().stream().filter(s -> s.startsWith("COUNT")).collect(Collectors.toSet());
```
brianloss: You're making a set here only for the purpose of iterating over it 2 lines down, yes? This could be combined in to something like:
```java
Map<String,Integer> countKeyMap = new HashMap<>();
dictionary.keySet().stream()
    .filter(key -> key.startsWith("COUNT"))
    .filter(countKey -> entry.getValue().getDictionary().containsKey(countKey))
    .forEach(countKey -> {
        TypeAttribute countTypeAttribute = ((TypeAttribute) entry.getValue().getDictionary().get(key));
        int count = ((BigDecimal) countTypeAttribute.getType().getDelegate()).intValue();
        countKeyMap.put(countKey, count);
    });
```

tomnelson:
closing until accumulo-utils PR is merged

PR #304, opened 2019-01-15
Added configurable retry logic to the audit service.

jwomeara:
Assuming this is approved/merged, I can create and tag a new audit-service release.

### DIFF
--- a/services/audit-service/src/main/java/datawave/microservice/audit/config/AuditProperties.java
+++ b/services/audit-service/src/main/java/datawave/microservice/audit/config/AuditProperties.java
@@ -34,8 +34,9 @@ public class AuditProperties {
     }
     
     public void setRetry(Retry retry) {
-        if (retry != null)
-            this.retry = retry;
+        if (retry == null)
+            throw new NullPointerException("Audit Retry properties must not be null.");
+        this.retry = retry;
     }
     
     public static class Retry {
--- a/services/audit-service/src/main/java/datawave/microservice/audit/config/AuditProperties.java
+++ b/services/audit-service/src/main/java/datawave/microservice/audit/config/AuditProperties.java
@@ -3,11 +3,15 @@ package datawave.microservice.audit.config;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 
+import java.util.concurrent.TimeUnit;
+
 @EnableConfigurationProperties(AuditProperties.class)
 @ConfigurationProperties(prefix = "audit")
 public class AuditProperties {
     private boolean confirmAckEnabled = true;
     private long confirmAckTimeoutMillis = 500L;
+    
+    private Retry retry = new Retry();
     
     public boolean isConfirmAckEnabled() {
         return confirmAckEnabled;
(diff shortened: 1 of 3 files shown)


services/audit-service/src/main/java/datawave/microservice/audit/controller/AuditController.java
```diff
@@ -105,25 +105,32 @@ public void processConfirmAck(Message<?> message) {
...
+            
+            boolean success = messageChannel.send(MessageBuilder.withPayload(AuditMessage.fromParams(parameters)).setCorrelationId(auditId).build());
+            
+            if (success && auditProperties.isConfirmAckEnabled()) {
+                try {
+                    success = latch.await(auditProperties.getConfirmAckTimeoutMillis(), TimeUnit.MILLISECONDS);
+                } catch (InterruptedException e) {
+                    success = false;
+                } finally {
+                    correlationLatchMap.remove(auditId);
```
brianloss: If success is false above, then you'll never remove auditId from the map.

brianloss: Also, not sure if this is a problem or not, but it seems like there might be a race condition where a slow confirm from one audit call to count down the latch of a previous timed out call. That would be a duplicate anyway, so that's why I'm thinking it might not be a problem.

jwomeara: That's a good catch.  I'll throw out a fix tomorrow morning. 

W.R.T. duplicates, that's definitely possible if the confirm ever takes longer than the confirmAckTimeout.  We definitely open ourselves up to that by allowing for retry logic.  I'm not sure how to better handle that case...

jwomeara: @brianloss I think I addressed your concerns with the latest commit.

services/audit-service/src/main/java/datawave/microservice/audit/controller/AuditController.java
```diff
@@ -141,13 +148,44 @@ private void sendMessage(AuditParameters parameters) {
...
+        
+        restAuditParams.clear();
+        restAuditParams.validate(parameters);
+        
+        boolean success;
+        final long auditStartTime = System.currentTimeMillis();
+        long currentTime;
+        int attempts = 0;
+        
+        AuditProperties.Retry retry = auditProperties.getRetry();
```
brianloss: I don't think it will happen using normal spring boot property mapping, but AuditProperties has a setter for retry that allows a null retry to be set. Therefore, you must guard for that here, or throw an error in the setter if null is passed.

services/audit-service/src/main/java/datawave/microservice/audit/config/AuditProperties.java
```diff
@@ -24,4 +28,42 @@ public long getConfirmAckTimeoutMillis() {
...
+    public Retry getRetry() {
+        return retry;
+    }
+    
+    public void setRetry(Retry retry) {
+        this.retry = retry;
+    }
+    
+    public static class Retry {
+        private int maxAttempts = Integer.MAX_VALUE;
```
FineAndDandy: should this be a smaller number to rule out a crazy number of interrupted exceptions that causes a constant spam?

jwomeara: Sure.  Done!

services/audit-service/src/main/java/datawave/microservice/audit/controller/AuditController.java
jwomeara: Updated the setter to disallow null assignment.

services/audit-service/src/main/java/datawave/microservice/audit/config/AuditProperties.java
```diff
@@ -34,13 +34,14 @@ public Retry getRetry() {
     }
     
     public void setRetry(Retry retry) {
-        this.retry = retry;
+        if (retry != null)
+            this.retry = retry;
```
brianloss: Wouldn't it be better to throw an exception rather than silently doing nothing? If this ever occurred, it would be very difficult to debug if it just silently ignores the set call.

Issue #282, opened 2018-12-17
Migrate Accumulo "LookupBean" EJB to Spring Boot Microservice


keith-ratcliffe:
Accumulo Lookup implementation will serve as a precursor for migration of our core query services to the new microservice architecture, as it will necessitate refactoring of shared REST API code within the current web-services/ module, such as web service response handling, query request handling, web exception handling, etc

PR #312, opened 2019-01-24
re #282 - accumulo microservice wip

keith-ratcliffe:
@brianloss , just wanted to go ahead and push this up for preview of LookupBean-related refactoring so far

Still a work in progress, pending....

* merge of [#311](https://github.com/NationalSecurityAgency/datawave/pull/311), which will provide Type-related dependencies required here
* refactor of remaining EJB's as we discussed (e.g., UpdateBean, TableAdminBean, StatsBean, etc)


services/accumulo-service/src/main/java/datawave/microservice/accumulo/config/AccumuloProperties.java
```diff
@@ -0,0 +1,44 @@
+package datawave.microservice.accumulo.config;
```
brianloss: I wonder if this should go in a different module (accumulo-utils? or the service starter). It seems a number of services might need access to an accumulo connection and it would be nice to not re-create the same class over and over (it's already in audit for the accumulo auditor).

services/accumulo-service/src/main/java/datawave/microservice/accumulo/lookup/LookupController.java
```diff
@@ -0,0 +1,59 @@
...
+ * REST controller for Accumulo lookup service
+ */
+@RestController
+@ConditionalOnProperty(name = "accumulo.lookup.enabled", havingValue = "true")
+@RolesAllowed({"AuthorizedUser", "AuthorizedServer", "InternalUser", "Administrator"})
+@RequestMapping(path = "/v1", produces = {MediaType.APPLICATION_JSON_VALUE, MediaType.APPLICATION_XML_VALUE})
+public class LookupController {
+    
+    @Autowired
+    private LookupService lookupService;
```
brianloss: It should be possible to make this final and set in the constructor with autowiring.

brianloss: > It should be possible to make this final and set in the constructor with autowiring.

Like [this](https://github.com/NationalSecurityAgency/datawave/blob/febd10fa76daa3d6ec7bdf485681973e3263ab0f/services/authorization-service/src/main/java/datawave/microservice/authorization/AuthorizationOperations.java#L37).

services/accumulo-service/src/main/java/datawave/microservice/accumulo/lookup/LookupService.java
```diff
@@ -0,0 +1,400 @@
...
+        String CF = "colFam";
+        String CF_ENCODING = "cfEnc";
+        String CQ = "colQual";
+        String CQ_ENCODING = "cqEnc";
+        String BEGIN_ENTRY = "beginEntry";
+        String USE_AUTHS = "useAuthorizations";
+        String END_ENTRY = "endEntry";
+    }
+    
+    @Resource(name = "auditLookupSecurityMarking")
```
brianloss: Shouldn't this be Autowired (with an `@Qualifier`) if you need a name. Same comments about finals an injecting in the constructor (at least IntelliJ tells me that's preferred).

services/accumulo-service/src/main/java/datawave/microservice/audit/AuditClient.java
```diff
@@ -0,0 +1,196 @@
+package datawave.microservice.audit;
```
brianloss: This seems like another candidate for a shared library of some kind. When we get to any query type of service, we'll need to send audits from them too.

services/accumulo-service/src/main/java/datawave/microservice/audit/AuditClient.java
```diff
@@ -0,0 +1,196 @@
...
+    public void submit(Request request) {
+        submit(request, DEFAULT_REQUEST_PATH);
+    }
+    
+    public void submit(Request request, String requestPath) {
+        
+        Preconditions.checkNotNull(request, "request cannot be null");
+        Preconditions.checkNotNull(requestPath, "requestPath cannot be null");
+        
+        if (log.isDebugEnabled()) {
```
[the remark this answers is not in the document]
brianloss: Same for all the log statements in this class.

services/accumulo-service/src/main/java/datawave/microservice/audit/config/AuditServiceConfiguration.java
```diff
@@ -0,0 +1,22 @@
...
+import datawave.microservice.audit.config.discovery.AuditDiscoveryConfiguration;
+import org.springframework.boot.autoconfigure.AutoConfigureAfter;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ConditionalOnProperty(name = "audit.enabled", havingValue = "true")
```
brianloss: Maybe add matchIfMissing to ensure that audit defaults to true?

brianloss:
I noticed you added CA and a cert. Is there a reason you can't use the ones in the spring-boot-starter-datawave project?

services/accumulo-service/src/main/java/datawave/microservice/accumulo/config/AccumuloProperties.java
keith-ratcliffe: Yep, had been thinking the same thing, just didn't want to stir the pot too much until I knew this was closer to being fully baked ;) Same with the AuditClient...I packaged that stuff separately, knowing this shouldn't really be it's final home

services/accumulo-service/src/main/java/datawave/microservice/audit/AuditClient.java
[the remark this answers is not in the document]
keith-ratcliffe: But slf4j doesn't prevent the first argument (string literal) from getting thrown onto the heap, right?

services/accumulo-service/src/main/java/datawave/microservice/audit/config/AuditServiceConfiguration.java
keith-ratcliffe: Yeah, good call. Will do

keith-ratcliffe:
> I noticed you added CA and a cert. Is there a reason you can't use the ones in the spring-boot-starter-datawave project?

Good point. Pretty sure I had just blatantly plagiarized the audit-service's test resources setup. Probably is best to consolidate them though. I'll take a stab

services/accumulo-service/src/main/java/datawave/microservice/accumulo/lookup/LookupService.java
brianloss: Nah, no big benefit I guess. [here](https://www.baeldung.com/spring-annotations-resource-inject-autowire) is a pretty decent overview. I had just been avoiding using the JSR-250 annotations, but I suppose if the code isn't going to get included in Wildfly there's no conflict.

services/accumulo-service/src/main/java/datawave/microservice/audit/AuditClient.java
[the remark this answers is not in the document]
brianloss: slf4j won't convert the given parameter to a string unless the log message is going to get sent (e.g., the log level is debug in this case). If you do:
```
log.debug("Received audit request: {}", request.getAuditParameters().toString());
```
then there's no benefit, but if you do:
```
log.debug("Received audit request: {}", request.getAuditParameters());
```
then unless `getAuditParameters()` is expensive somehow, this will be just as effective and more readable than wrapping in `if (log.isDebugEnabled())`.

PR #314, opened 2019-01-29
make the javadoc generation faster

tomnelson:
Changes to be committed:
	modified:   docs/pom.xml


docs/pom.xml
```diff
@@ -297,25 +297,68 @@
...
                         </goals>
                         <phase>package</phase>
-                        <configuration combine.children="append">
-                            <includeDependencySources>true</includeDependencySources>
-                            <includeTransitiveDependencySources>true</includeTransitiveDependencySources>
-                            <dependencySourceIncludes>
-                                <dependencySourceInclude>gov.nsa.datawave:*</dependencySourceInclude>
-                                <dependencySourceInclude>gov.nsa.datawave.webservices:*</dependencySourceInclude>
-                            </dependencySourceIncludes>
+                        <configuration>
```
brianloss: This can't be inherited from the parent? Previously, it was using `combine.children="append"` (and the same execution ID as in the parent (`module-apidocs`)) to add configuration to that specified in the parent. Now the configuration is copied from the parent.

docs/pom.xml
```diff
@@ -310,12 +323,12 @@
                         </goals>
                         <phase>package</phase>
                         <configuration combine.children="append">
-                            <includeDependencySources>true</includeDependencySources>
-                            <includeTransitiveDependencySources>true</includeTransitiveDependencySources>
-                            <dependencySourceIncludes>
-                                <dependencySourceInclude>gov.nsa.datawave:*</dependencySourceInclude>
-                                <dependencySourceInclude>gov.nsa.datawave.webservices:*</dependencySourceInclude>
-                            </dependencySourceIncludes>
+                            <javadocVersion>1.8.0</javadocVersion>
```
brianloss: The version is defined in the parent, and that seems like something we should keep.

docs/pom.xml
```diff
@@ -310,12 +323,12 @@
...
+                            <sourcepath>${project.build.directory}/dependency</sourcepath>
+                            <outputDirectory>${project.build.directory}/core-api</outputDirectory>
+                            <finalName>core-api</finalName>
+                            <attach>false</attach>
```
brianloss: Why the attach of false?

docs/pom.xml
```diff
@@ -297,11 +297,24 @@
...
+                        <configuration>
+                            <classifier>sources</classifier>
+                            <includeGroupIds>gov.nsa.datawave, gov.nsa.datawave.webservices</includeGroupIds>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
                 <artifactId>maven-javadoc-plugin</artifactId>
+                <version>3.0.1</version>
```
brianloss: Just curious why this version isn’t coming from the parent pom?

tomnelson: removed

PR #315, opened 2019-01-29
Improve metadata-utils

brianloss:
* Remove initialize methods, require initialize method args in constructor, make parameters final
  It used to be that a prototype metadata helper would be injected, and then initialized with specific params. However, initialize
  could never be called more than once, and really this should have been part of construction. However, the metadata helper needs
  to be proxied by Spring in order to perform caching, since at each injection point, the connector, metadata table name, and auths
  could all be different, there was no way to do this. However, it turns out this can be done by instantiating the bean directly with
  the bean factory and passing args. The MetadataHelperFactory was updated to require the args, and use the bean factory to
  construct a MetadataHelper with the passed parameters.
* Cache eviction (and the trace-level cache printing) was moved to an external class. The CacheEvict annotations require all cache
  names to be passed, and there are many different cache names used by the MetadataHelper. Not all were cleared by the cache
  evict method (and one class chained to the parent, which doesn't work with a proxied class -- the proxy won't be called). This
  also removed the need for MetadataHelper to implement ApplicationContextAware.
* Annotate helpers with prototype scope
* Annotate factory methods with @Lookup
* Make helpers take constructor parameters (and pass from XML configs)
* Fix issue with specifying auths in metadata helper XML file.

These changes will help when using these classes via spring boot. I set metadata-utils to release 1.1, since the intent is to release it as soon as the PR is merged.

### DIFF
--- a/services/metadata-utils/src/main/java/datawave/query/util/AllFieldMetadataHelper.java
+++ b/services/metadata-utils/src/main/java/datawave/query/util/AllFieldMetadataHelper.java
@@ -29,7 +29,6 @@ import org.apache.hadoop.io.Text;
 import org.apache.hadoop.io.WritableUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.cache.annotation.CacheEvict;
 import org.springframework.cache.annotation.Cacheable;
 import org.springframework.cache.annotation.EnableCaching;
 import org.springframework.context.annotation.Scope;
@@ -1004,17 +1003,6 @@ public class AllFieldMetadataHelper {
     @Override
     public String toString() {
         return getKey(this);
-    }
-    
-    /**
-     * Invalidates all elements in all internal caches
-     */
-    @CacheEvict(value = {"loadAllFields", "isIndexed", "getAllDatatypes", "getCompositeToFieldMap", "getFieldsToDatatypes", "getFieldsForDatatype",
-            "getIndexOnlyFields", "loadTermFrequencyFields", "loadIndexedFields", "loadExpansionFields", "loadContentFields", "loadDatatypes"},
-                    allEntries = true, cacheManager = "metadataHelperCacheManager")
-    public void evictCaches() {
-        log.debug("evictCaches");
-        typeMetadataHelper.evictCaches();
     }
     
 }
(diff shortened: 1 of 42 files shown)


services/metadata-utils/src/main/java/datawave/query/util/AllFieldMetadataHelper.java
```diff
@@ -52,18 +53,24 @@
...
+        Preconditions.checkNotNull(typeMetadataHelper);
+        this.typeMetadataHelper = typeMetadataHelper;
+        
+        Preconditions.checkNotNull(compositeMetadataHelper);
+        this.compositeMetadataHelper = compositeMetadataHelper;
+    }
+    
     protected String getDatatype(Key k) {
         String datatype = k.getColumnQualifier().toString();
         int index = datatype.indexOf('\0');
```
keith-ratcliffe: Not related to your changes, but I notice we define a NULL_BYTE constant, yet we don't use it here

brianloss: > Not related to your changes, but I notice we define a NULL_BYTE constant, yet we don't use it here

The NULL_BYTE constant is a string, whereas we're searching for a char here. I'm not sure if there's a performance impact to doing indexOf on a single character string vs a character, but that is one difference.

brianloss:
Based on an offline discussion, I further refactored to remove the initialize methods from the metadata helpers. Instead, they require the connector, metadata table name, and auths set as constructor parameters. The metadatahelperfactory createMetadataHelper method is responsible for passing the required constructor parameters along to the metadata helper.

brianloss:
Added more updates -- the cache eviction is moved to an external class. Previously this was accomplished using spring CacheEvict annotations. However, due to a chained call to the parent class in one case (MetadataHelperWithDescriptions), and likely just missing cache names, not all caches were actually evicted when the metadata table cache was refreshed. Moving to an external cache eviction class allows:
1. All caches in the metadata helper cache manager to be cleared
2. The cache print method to be moved
3. MetadataHelper no longer needs to implement ApplicationContextAware.

brianloss:
@tomnelson, @keith-ratcliffe, @ivakegg Any issues with this PR? I'd like to get it merged in soon.

PR #321, opened 2019-02-01
Replace some code in warehouse with Accumulo API

milleruntime:
* Migrate some code to use Accumulo API compatible with 1.9 and 2.0 in
ShardedTableMapFile, AccumuloHelper and ConfigurableAgeOffFilter


warehouse/core/src/main/java/datawave/iterators/filter/ConfigurableAgeOffFilter.java
```diff
@@ -346,7 +347,7 @@ private void initFilterRules() throws IllegalArgumentException, IOException {
                                     .expireAfterAccess(EXPIRATION_INTERVAL_MS, TimeUnit.MILLISECONDS).build(new ReloadableCacheBuilder());
                     // this will schedule a check to see if the update or expiration intervals have changed
                     // if so the ruleCache will be rebuilt with these new intervals
-                    SimpleTimer.getInstance(AccumuloConfiguration.getDefaultConfiguration()).schedule(
+                    SimpleTimer.getInstance(1).schedule(
```
milleruntime: Since the default is "1" for the thread pool size, I made this call the other SipmleTimer method which just takes an integer for the thread pool size.  SimpleTimer hasn't changed but AccumuloConfiguration has changed.

keith-ratcliffe:
Awesome, thanks a lot for taking a first stab at this. Not sure how close we are to a 2.5.0 release here, so instead of merging into master I'm thinking we might want to start a new branch for the Accumulo 2.0 push in general. That way we can work through any conflicts with out worrying about pending releases. Although, if these changes are compatible with 1.9.2 and all our integration tests pass, probably not a big deal to go ahead and merge w/master.

milleruntime:
> Although, if these changes are compatible with 1.9.2 and all our integration tests pass, probably not a big deal to go ahead and merge w/master.

Yeah these changes should work with 1.9.2 and 2.0

milleruntime:
Tests are passing that were run with "mvn -Pdev -Ddeploy -Dtar clean install"

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
```diff
@@ -322,33 +317,30 @@ public static Path createShardedMapFile(Logger log, Configuration conf, Path wor
...
      *            logger for reporting errors
-     * @param instance
-     *            Accumulo instance to query shard locations
-     * @param credentials
-     *            Accumulo access credentials for querying shard locations
+     * @param accumuloHelper
+     *            Accumulo helper to query shard locations
      * @param shardedTableName
      *            name of the shard table--the table whose locations we are querying
      * @return a map of KeyExtent to the location of those key extents in accumulo
```
matthpeterson: The @return description is no longer accurate.

milleruntime:
> I'm unconvinced though that it's time to migrate datawave to Accumulo 2.0 on the master branch. I'll let others weigh in. Accumulo 2.0 is only in early alpha testing and the Accumulo team recently advised us not to migrate immediately to 2.0 given the number of Accumulo changes.

I hear you on that.  These changes were more general improvements that will work with 1.9 and only help with future upgrades.  One of the changes I made uses TableOperations.locate(), which was added in 1.8.0.  So that will work with 1.8.x, 1.9.x and 2.0. 

I think the more we can get Datawave using Accumulo API instead of internals, the better.  Looking at how Datawave is using Accumulo also helps us better prepare the alpha for the 2.0 release candidate.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
milleruntime: Fixed in 669fc27

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
```diff
@@ -54,7 +52,7 @@
     public static final String SHARDED_MAP_FILE_PATHS_RAW = "shardedMap.file.paths.raw";
     public static final String SHARD_VALIDATION_ENABLED = "shardedMap.validation.enabled";
     public static final String MAX_SHARDS_PER_TSERVER = "shardedMap.max.shards.per.tserver";
-    
+
```
brianloss: These whitespace changes suggest this code has not been compiled, since the formatter that runs during compilation is what adds the blank space at the beginning of the line. (I confirmed by checking out and building this PR--I'm left with an unclean tree.) Please make sure you've built the code and check in any auto-formatter changes.

milleruntime: It should be fixed now with bed8dae7604e15c67ce5ec106928842629b3eb1d.  FYI I used "mvn -Pdev -Ddeploy -Dtar -DskipTests clean install"

brianloss: It's a good idea to build with `-Ddist` before submitting a PR too, so doc is generated and you'll discover if there are any doc errors.

milleruntime: Build is also passing with -Ddist

PR #322, opened 2019-02-01
fixes #320: IngestFieldFilter to handle values

### LINKED ISSUE #320: The IngestFieldFilter needs to be able to handle comparing values, and multiple field combinations
The IngestFieldFilter currently supports dropping one field if another one exists.  This needs to be extended to be able to only drop a field if its value matches another specified field.  In addition we need to be able to handle tuples of fields and optionally values to other tuples of fields and values.
The current configuration is as follows:

```
configuration = <expression>{,<expression>}
expression = <field>:<field>
```

where the field on the left of the colon is dropped if field on the right exists.  Perhaps an extended configuration may be:

```
configuration = <expression>{,<expression>}
expression = <fields>:<fields>
fields = <field>{|<field>}
```

where the field(s) on the left of the operand are dropped if the fields on the right exist.  Then the same configuration format could be used in a separate property that denotes that the values must match as well.  Both forms of functionality could be implemented in the same field filter but does not have to be.

ivakegg:
* Updated the IngestFieldFilter to compare values as configured
  * Updated the IngestFieldFilter to handle groups of fields to keep or drop
  * Kept the functionality backward compatible with the original code
  * Added test cases for all additional functionality

### DIFF
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/IngestFieldFilter.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/IngestFieldFilter.java
@@ -1,18 +1,29 @@
 package datawave.ingest.data.config.ingest;
 
+import com.google.common.collect.Lists;
 import com.google.common.collect.Multimap;
-import com.google.common.collect.TreeMultimap;
 import datawave.ingest.data.Type;
-import org.apache.commons.lang.StringUtils;
+import datawave.ingest.data.config.GroupedNormalizedContentInterface;
+import datawave.ingest.data.config.NormalizedContentInterface;
+import org.apache.commons.collections.Unmodifiable;
+import org.apache.commons.collections.keyvalue.AbstractMapEntry;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.log4j.Logger;
 
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
-import java.util.Map;
+import java.util.Collections;
+import java.util.List;
 
 /**
  * The field filter can drop fields based on the presence of other preferred fields. This is configuration driven and expects colon delimited pairs, A:B, where
- * if A is present in the fields, then any occurrences of B will be removed.
+ * if A is present in the fields, then any occurrences of B will be removed. A and/or B can represent a set of fields delimited by the ampersand sign. In the
+ * case of the data.fieldvalue.filter, the values must match as well which implies that the number of KEEP fields must be the same as the number of DROP fields
+ * which is not the case for the fieldName filter. For backward compatibility datatype1.data.field.filter can be used in place of
+ * datatype1.data.fieldname.filter however that is deprecated and will be removed in a future release.
+ *
  * <p>
  * Example:
  * 
@@ -20,9 +31,14 @@ import java.util.Map;
  * {@code
  * 
  *     <property>
- *         <name>datatype1.data.field.filter</name>
- *         <value>KEEP1:DROP1,KEEP2:DROP2</value>
+ *         <name>datatype1.data.fieldname.filter</name>
+ *         <value>KEEP1:DROP1,KEEP2&KEEP3:DROP2</value>
  *     </property>
+ *     <property>
+ *         <name>datatype1.data.fieldvalue.filter</name>
+ *         <value>KEEP1:DROP1,KEEP2&KEEP3:DROP2&DROP3</value>
+ *     </property>
+ * 
  * }
  * </pre>
  */
@@ -30,12 +46,18 @@ public class IngestFieldFilter {
     
     private static final Logger logger = Logger.getLogger(IngestFieldFilter.class);
     
+    @Deprecated
     public static final String FILTER_FIELD_SUFFIX = ".data.field.filter";
-    public static final String PAIR_DELIM = ",";
-    public static final String VALUE_DELIM = ":";
+    
+    public static final String FILTER_FIELD_NAME_SUFFIX = ".data.fieldname.filter";
+    public static final String FILTER_FIELD_VALUE_SUFFIX = ".data.fieldvalue.filter";
+    public static final char PAIR_DELIM = ',';
+    public static final char VALUE_DELIM = ':';
+    public static final char FIELD_DELIM = '&';
     
     private final Type dataType;
-    private Map<String,Collection<String>> filters;
+    private FieldConfiguration fieldNameFilters;
+    private FieldConfiguration fieldValueFilters;
     
     public IngestFieldFilter(Type dataType) {
         this.dataType = dataType;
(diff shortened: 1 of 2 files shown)


warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/IngestFieldFilter.java
```diff
@@ -1,42 +1,66 @@
...
+import java.util.List;
 
 /**
- * The field filter can drop fields based on the presence of other preferred fields. This is configuration driven and expects colon delimited pairs, A:B, where
+ * The field filter can drop fields based on the presence of other preferred fields.
+ * This is configuration driven and expects colon delimited pairs, A:B, where
  * if A is present in the fields, then any occurrences of B will be removed.
+ * In the case of the data.fieldvalue.filter, the values must match as well which
+ * implies that the number of KEEP fields must be the same as the number of DROP
+ * fields which is not the case for the fieldName filter.  For backward compatibility
```
hlgp: An explicit blurb about the & option would be nice.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/IngestFieldFilter.java
```diff
@@ -1,42 +1,66 @@
...
+ * datatype1.data.field.filter can be used in place of datatype1.data.fieldname.filter.
```
hlgp: Can we add a note that data.field.filter will be deprecated in the configs and release notes, then remove it after the next major version (or two) to trim down some config bloat?

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/IngestFieldFilter.java
ivakegg: ok, will do

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/IngestFieldFilter.java
ivakegg: sure.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/IngestFieldFilter.java
```diff
@@ -47,42 +71,269 @@ public IngestFieldFilter(Type dataType) {
...
      */
     public void apply(Multimap<String,?> fields) {
-        for (Map.Entry<String,Collection<String>> entry : filters.entrySet()) {
-            if (fields.containsKey(entry.getKey())) {
-                for (String toRemove : entry.getValue()) {
-                    fields.removeAll(toRemove);
+        // for the field name filters, remove the drop fields if the keep fields exist
+        for (FieldFilter filter : fieldNameFilters) {
+            if (fields.keySet().containsAll(filter.getKeepFields())) {
+                for (String fieldToRemove : filter.getDropFields()) {
```
hlgp: No need to iterate if you're using removeAll.

ivakegg: presuming you mean fields.getKeySet().removeAll(filter.getDropFields()).  The removeAll on a Multimap only support a single key.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/IngestFieldFilter.java
```diff
@@ -1,41 +1,63 @@
...
     private static final Logger logger = Logger.getLogger(IngestFieldFilter.class);
     
+    @Deprecated
     public static final String FILTER_FIELD_SUFFIX = ".data.field.filter";
+    
+    public static final String FILTER_FIELD_NAME_SUFFIX = ".data.fieldname.filter";
+    public static final String FILTER_FIELD_VALUE_SUFFIX = ".data.fieldvalue.filter";
     public static final String PAIR_DELIM = ",";
     public static final String VALUE_DELIM = ":";
+    public static final String FIELD_DELIM = "&";
```
alerman: Since these delimiters are single characters, why not use chars?

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/IngestFieldFilter.java
```diff
@@ -47,42 +69,284 @@ public IngestFieldFilter(Type dataType) {
...
+     * Gather sets of values for the fields specified. This is basically the dot product of the values for each field to gather, with the complexity that we
+     * need to keep groups together. So if one of the fields' values contains a group (or the group is specified) then only those other fields' values with the
+     * same group should be considered in the dot product.
+     * 
+     * @param fieldValues
+     *            The field values
+     * @param fieldsToGather
+     *            The fields to gather (this ordering is maintained in the return value)
+     * @param group
+     *            The group the to use or null if not specified
```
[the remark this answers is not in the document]
ivakegg: ok

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/IngestFieldFilter.java
```diff
@@ -47,42 +69,284 @@ public IngestFieldFilter(Type dataType) {
...
+     * @return a collection of value lists which maintain the order as specified in the fieldsToGather
+     */
+    private Collection<List<FieldValue>> gatherValueSets(Multimap<String,?> fieldValues, List<String> fieldsToGather, int index, String group) {
+        List<List<FieldValue>> valueSets = Lists.newArrayList();
```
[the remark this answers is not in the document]
ivakegg: agreed

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/IngestFieldFilter.java
```diff
@@ -47,42 +69,284 @@ public IngestFieldFilter(Type dataType) {
...
+        }
+        String field = fieldsToGather.get(index);
+        
+        // for each of the values for the current field
+        for (FieldValue value : getValues(fieldValues, field, group)) {
+            // if this is the last field, then simply create a value set per value
+            if (index == 0) {
+                // initialize with an initial capacity equal to the number of fields
+                List<FieldValue> valueSet = new ArrayList<>(fieldsToGather.size());
+                valueSet.add(value);
```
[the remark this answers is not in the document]
ivakegg: ok

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/IngestFieldFilter.java
```diff
@@ -47,42 +69,284 @@ public IngestFieldFilter(Type dataType) {
...
+                valueSets.add(valueSet);
+            }
+            // else this add this value to each of the value sets found recursively for the remaining fields
```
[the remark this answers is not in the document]
ivakegg: You must be bored...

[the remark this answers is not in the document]
alerman: Not much else to do here while I was waiting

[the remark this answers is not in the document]
ivakegg: Thank you for choosing code reviews instead of playing solitaire ;)

alerman:
I like to feel a but useful...now and then

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/IngestFieldFilter.java
```diff
@@ -47,42 +69,284 @@ public IngestFieldFilter(Type dataType) {
...
+        }
+        
+        return valueSets;
     }
     
     /**
-     * Applies the configured filter rules to the given fields.
-     *
-     * @param fields
+     * Get the values for a given field and optional group
```
jschmidt10: This javadoc block duplicates the one below it.

PR #325, opened 2019-02-07
Replace use of Accumulo internals with 1.8 API

milleruntime:
* Replace IterInfo with IteratorSetting in IngestJob
* Use the Accumulo API IteratorSetting instead of internal thrift
IterInfo class
* Replace old accumulo code in AggregatingReducer
* Replace Constants.NO_AUTHS with Authorizations.EMPTY
* Replace Tables class with Accumulo API

Since my other request (#321) was already approved I didn't want to push more changes to that branch. These changes are similar in that they should be safe and will work with Accumulo versions 1.8, 1.9 and 2.0.


milleruntime:
@ivakegg I found a similar change to error handling in the AccumuloRecordWriter so I pushed up to this branch.  This was the same thing I did with UpdateBean but less code.

PR #344, opened 2019-02-26
Prefer default table names over local magic strings.

apmoriarty:
Consolidating the table names used in unit tests and some production code into default values.


warehouse/core/src/main/java/datawave/util/TableNames.java
```diff
@@ -0,0 +1,21 @@
+package datawave.util;
+
+/**
+ * Class that contains default table names
+ */
+public class TableNames {
```
tomnelson: interface instead of class, and you won't have to declare public static final on the members (since they already will be)

apmoriarty: Good call, updated in next commit.

warehouse/core/src/main/java/datawave/util/TableName.java
```diff
@@ -0,0 +1,23 @@
+package datawave.util;
+
+/**
+ * Class that contains default table names
+ */
+public interface TableName {
```
alerman: @tomnelson I disagree with making this an interface. While yes, you can remove the access modified keywords, it really doesn't make any sense. What would it mean to implement TableName?

tomnelson: So you think it has to be a (final) class just to prevent some misguided future programmer from extending/implementing it, and you're willing to let it be more cluttered with extraneous declarations just to outwit that future troublemaker?
There's no rule that says an interface must be implemented.

alerman: Honestly yes I would prefer that. Using an interface for this goes directly against the definition of an interface. 

On top of that, in the future should some developer see this and choose to implement the interface, the values could be silently overridden by members of the implementing class. I understand that this pr does not Implement the class, but the minor amount of clutter caused by having the modifiers is worth protecting from other issues in the future.

apmoriarty: Good discussion, I'm leaning towards making it a final class and will have an update shortly.

tomnelson: i'm fine with leaving it up to the author

warehouse/core/src/main/java/datawave/util/TableName.java
```diff
@@ -0,0 +1,23 @@
+package datawave.util;
+
+/**
+ * Class that contains default table names
+ */
+public final class TableName {
+
```
tomnelson: you need a private CTOR:

private TableName() {}

warehouse/ingest-csv/src/test/java/datawave/ingest/csv/mr/handler/ColumnBasedHandlerTestUtil.java
```diff
@@ -40,9 +41,9 @@
  */
 public class ColumnBasedHandlerTestUtil {
     
-    public static final Text shardTableName = new Text("shard");
-    public static final Text shardIndexTableName = new Text("shardIndex");
-    public static final Text shardReverseIndexTableName = new Text("shardReverseIndex");
+    public static final Text shardTableName = new Text(TableName.SHARD);
+    public static final Text shardIndexTableName = new Text(TableName.SHARD_INDEX);
+    public static final Text shardReverseIndexTableName = new Text(TableName.SHARD_RINDEX);
     public static final Text edgeTableName = new Text("edge");
```
keith-ratcliffe: No edge table constant?

apmoriarty: I'll add EDGE as a table name in the next commit.

warehouse/core/src/main/java/datawave/util/TableName.java
apmoriarty: Added in next commit

warehouse/core/src/main/java/datawave/util/TableName.java
phrocker: A table name as a representation of the code has always seem short sighted -- a table name should be a representation of its utility. As a result it should be pulled from configuration or translated dynamically . The virtue of data tables being sharded should be a feature and not part of its nomenclature. Intrinsically you have a data store of information and its indices. As such I think the project would be better suited by not making constants from table names but instead defining a mechanism to translate ( either by config or dynamically ) their names and references. 

The academic argument above, not withstanding, the string location has been condensed, but a re-compilation change would need to be made. One may argue, "that wasn't the original intent," but the same level of effort would be required for this and any other true improvement.

warehouse/ingest-csv/src/test/java/datawave/ingest/csv/mr/handler/ColumnBasedHandlerTestUtil.java
keith-ratcliffe: Btw, I created #346 for the cleanup work. If you want to just commit as-is that's fine. Or if you want to go ahead and remove it and the references that you've refactored as part of this PR, that's cool to. Doesn't matter to me

warehouse/core/src/main/java/datawave/util/TableName.java
tomnelson: Also, it looks clumsy to me now.
Using a class for this goes directly against the definition of a class (to provide a template for the creation of Objects)

warehouse/ingest-csv/src/test/java/datawave/ingest/csv/mr/handler/ColumnBasedHandlerTestUtil.java
apmoriarty: This isn't an everything-or-nothing PR, I'm fine leaving that work for a separate issue.

warehouse/core/src/main/java/datawave/util/TableName.java
apmoriarty: Sounds like a good improvement, you should write up an issue for it!

brianloss: If it helps the discussion, Josh Bloch says using an interface for constants is an anti-pattern (see https://stackoverflow.com/questions/2659593/what-is-the-use-of-interface-constants).
Seems like an enum could be well suited for use here.

tomnelson: I have seen that. My interpretation is that **implementing** an interface for its constants is the anti-pattern, not the mere existence of a structure to hold those constants. I prefer a clean structure with sensible use over a more cluttered class that is really not a 'class'. ymmv.

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/handler/edge/ProtubufEdgeDeleteModeTest.java
```diff
@@ -462,7 +460,7 @@ public static void processEvent(Multimap<String,NormalizedContentInterface> even
                         log.info(keyString.trim());
                     }
                 }
-                Assert.fail(String.format("Expected: %s edge keys.\nFound: %s", expectedEdgeKeys, countMap.get(shardTableName), countMap.get(edgeTableName)));
+                Assert.fail(String.format("Expected: %s edge keys.\nFound: %s", expectedEdgeKeys, countMap.get(TableName.SHARD), countMap.get(edgeTableName)));
```
ejrgilbert: Will this work to pass a String to a `Map<Text,Integer>`?

warehouse/core/src/main/java/datawave/util/TableName.java
```diff
@@ -0,0 +1,27 @@
+package datawave.util;
+
+/**
+ * Class that contains default table names
+ */
+public final class TableName {
+    
+    public static final String DATE_INDEX = "dateIndex";
```
ejrgilbert: I've seen places in the code that are "DateIndex", should the first char be capitalized?

warehouse/query-core/src/test/java/datawave/query/jexl/functions/ContentFunctionQueryTest.java
```diff
@@ -190,7 +191,7 @@ public static void setupConfiguration(Configuration conf) {
         conf.set(DATA_TYPE_NAME + DataTypeHelper.Properties.INGEST_POLICY_ENFORCER_CLASS, IngestPolicyEnforcer.NoOpIngestPolicyEnforcer.class.getName());
         conf.set(DataTypeHelper.Properties.DATA_NAME, DATA_TYPE_NAME);
         conf.set(DATA_TYPE_NAME + ".data.category.index", "ID, BODY");
-        // conf.set(DATA_TYPE_NAME + ".model.table.name", METADATA_TABLE_NAME);
+        // conf.set(DATA_TYPE_NAME + ".model.table.name", METADATA);
```
ejrgilbert: Even though this is commented out, it should be: `METADATA_TABLE_NAME`

warehouse/core/src/main/java/datawave/util/TableName.java
apmoriarty: From looking at all the instances of table names in the code, it appears that 'dateIndex' would be preferred over 'DateIndex'. I'm in favor of picking one standard convention

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/handler/edge/ProtubufEdgeDeleteModeTest.java
apmoriarty: Fixed in next commit.

apmoriarty:
Resolved merge conflicts.

PR #349, opened 2019-03-05
Fix for bypassAccumulo in certain cases

apmoriarty:
Fix for LookupUUIDTune to persist value of bypassAccumulo via the ShardQueryConfiguration.

### DIFF
--- a/web-services/query/src/main/java/datawave/webservice/query/configuration/GenericQueryConfiguration.java
+++ b/web-services/query/src/main/java/datawave/webservice/query/configuration/GenericQueryConfiguration.java
@@ -34,7 +34,7 @@ public abstract class GenericQueryConfiguration {
     private Date beginDate = null;
     private Date endDate = null;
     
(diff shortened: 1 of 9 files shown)


web-services/query/src/main/java/datawave/webservice/query/configuration/GenericQueryConfiguration.java
```diff
@@ -137,15 +149,19 @@ public Long getMaxQueryResults() {
     }
     
     public void setMaxQueryResults(Long maxQueryResults) {
-        this.maxQueryResults = maxQueryResults <= 0l ? Long.MAX_VALUE : maxQueryResults;
```
brianloss: What's wrong with the ternary operator?

apmoriarty: Nothing, it was a personal preference. This line won't be changed once I rebase to pick up the maxResults changes.

Issue #176, opened 2018-08-15
Build/test DW on Hadoop 3.x


keith-ratcliffe:
This ticket tracks progress on DW updates required for Hadoop 3.x. Currently, the hadoop3x branch successfully builds/tests/deploys DW on (Apache) **Hadoop 3.1.1** and **Accumulo 1.9.2**

More details below...

keith-ratcliffe:
Currently, ingest examples in the quickstart seem to be working fine, but web service deploy/test is failing due to errors such as the following:

```
# From $WILDFLY_HOME/standalone/server.log
....
MSC000001: Failed to start service jboss.deployment.subunit."datawave-ws-deploy-application-2.3.0-SNAPSHOT-dev.ear"."datawave-ws-modification-2.3.0-SNAPSHOT.jar".INSTALL: org.jboss.msc.service.StartException in service jboss.deployment.subunit."datawave-ws-deploy-application-2.3.0-SNAPSHOT-dev.ear"."datawave-ws-modification-2.3.0-SNAPSHOT.jar".INSTALL: WFLYSRV0153: Failed to process phase INSTALL of subdeployment "datawave-ws-modification-2.3.0-SNAPSHOT.jar" of deployment "datawave-ws-deploy-application-2.3.0-SNAPSHOT-dev.ear"
        at org.jboss.as.server.deployment.DeploymentUnitPhaseService.start(DeploymentUnitPhaseService.java:154)
        at org.jboss.msc.service.ServiceControllerImpl$StartTask.startService(ServiceControllerImpl.java:1948)
        at org.jboss.msc.service.ServiceControllerImpl$StartTask.run(ServiceControllerImpl.java:1881)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.NoClassDefFoundError: com/ctc/wstx/io/InputBootstrapper
        at org.apache.hadoop.io.WritableComparator.<init>(WritableComparator.java:128)
        at org.apache.hadoop.io.WritableComparator.<init>(WritableComparator.java:116)
        at org.apache.hadoop.io.Text$Comparator.<init>(Text.java:360)
        at org.apache.hadoop.io.Text.<clinit>(Text.java:374)
[...]

PR #357, opened 2019-03-06
re: #176 - Updates required to build/test/deploy DataWave on Hadoop 3…

keith-ratcliffe:
….x and Accumulo 1.9.2

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,11 +34,12 @@
         <version.arquillian-weld-ee-embedded>1.0.0.Final</version.arquillian-weld-ee-embedded>
         <version.avro>1.7.4</version.avro>
         <version.caffeine>2.6.2</version.caffeine>
-        <version.commons-beanutils>1.7.0</version.commons-beanutils>
+        <version.commons-beanutils>1.9.3</version.commons-beanutils>
         <version.commons-cli>1.4</version.commons-cli>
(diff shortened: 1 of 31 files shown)


pom.xml
```diff
@@ -149,6 +153,11 @@
                 <artifactId>jackson-module-jaxb-annotations</artifactId>
                 <version>${version.jackson}</version>
             </dependency>
+            <dependency>
```
brianloss: What is it that needs woodstox?

keith-ratcliffe: It's a compile dependency for hadoop Configuration class via hadoop-common.

keith-ratcliffe: I took the standard approach to updating hadoop dependencies for this, but it's worth noting that H3x does also provide shaded jars for its client api and transitive dependencies, which would ostensibly eliminate the need for this. What do you guys think about moving toward usage of the shaded jars instead?

brianloss: I think it's great that Hadoop has moved to shaded jars. For us, I'm not sure we're a dependency of anything else, so there's probably no need. The only place I can think of where we might benefit is the guava library when used within code that is on the iterator stack. Then we're tied to Accumulo's version. If we shade and relocate guava, we could use a newer version on tservers, although really the reverse should be true -- Accumulo should be doing that.

PR #358, opened 2019-03-07
DATAWAVE-351: Stop using ConfigurableEventDataQueryFilter

FineAndDandy:
This is a dependency for #290


properties/default.properties
```diff
@@ -174,9 +174,6 @@ disable.index.only.documents=false
 enable.index.only.filter.functions=false
 query.tld.collapse.uids=false
 # Indicates whether tuples loaded for an event should be filtered based on field and value (true) or just based on field (false)
```
alerman: Should this comment also be removed?

apmoriarty: Remove comment as well

PR #364, opened 2019-03-12
when yield happens during grouping aggregation, undo the yield and ca…

tomnelson:
…rry on

 Changes to be committed:


warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
```diff
@@ -157,17 +158,34 @@ private Key createAttrKey(String fieldName) {
...
-                    Entry<Key,Document> lastEntry = in.next();
-                    GroupingTransform.this.apply(lastEntry);
+                for (int i = 0; i < groupFieldsBatchSize; i++) {
+                    if (in.hasNext()) {
+                        Entry<Key,Document> lastEntry = in.next();
+                        if (lastEntry != null) {
+                            GroupingTransform.this.apply(lastEntry);
+                        } else if (yieldCallback != null && yieldCallback.hasYielded()) {
+                            log.trace("lastEntry is null and yield was called");
+                            yieldCallback.yield(null); // undo the yield
```
ivakegg: Don't we need to break here?  Otherwise we will continue looping and calling in.hasNext() again which is wasteful.

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
```diff
@@ -157,17 +158,34 @@ private Key createAttrKey(String fieldName) {
...
+                        } else {
+                            // lastEntry was null and there was no yield
+                            break;
+                        }
+                    } else if (yieldCallback != null && yieldCallback.hasYielded()) {
+                        // hasNext is false and yield has been called. Undo the yield.
+                        log.trace("hasNext is false because yield was called");
+                        yieldCallback.yield(null);
```
ivakegg: Don't we want to break here?

tomnelson: By calling yieldCallback.yield(null), the yield is disabled and hasNext() will likely return true when we hit it again at the top of the loop. Glad to see your initial instinct was the same as mine: my initial instinct was to break here, and that is why my early tests today failed to gather and aggregate all of the entries after a yield. Once I allowed the loop to continue here, the same grouping was done whether there were yields or not.

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
tomnelson: This is similar to the comment below, however we may not ever have a case where in.hasNext() was true, but in.next() returns null. In any event, this was intended to assume that if something happened because of a yield, the right thing to do is to turn off the yield and see if there is another record. I believe we would only waste time here if in.hasNext() returns true over and over while in.next() returns null.

ivakegg: Apparently I was not clear in my explanation so my apologies.
The reason we yielded in the first place was that this scan thread needs to return control to the Tablet.  There are two ways to return control: 1) yield, 2) return a key.  What we need to happen here is since we detected a yield, we need to cancel the yield and return a key instead.  The "last key processed" in this case needs to be the key that was registered in the yield callback before setting it to null.  Return that key along with your grouping information gathered thus far.

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
```diff
@@ -157,17 +158,36 @@ private Key createAttrKey(String fieldName) {
...
             
             @Override
             public boolean hasNext() {
-                for (int i = 0; i < max && in.hasNext(); i++) {
-                    Entry<Key,Document> lastEntry = in.next();
-                    GroupingTransform.this.apply(lastEntry);
+                for (int i = 0; i < groupFieldsBatchSize; i++) {
+                    if (in.hasNext()) {
+                        GroupingTransform.this.apply(in.next());
+                    } else if (yieldCallback != null && yieldCallback.hasYielded()) {
```
ivakegg: If you do not have any grouping results to return, then perhaps we should let the yield stand as is.

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
```diff
@@ -157,17 +158,36 @@ private Key createAttrKey(String fieldName) {
...
+                        // hasNext is false and yield has been called. Save the key and undo the yield.
+                        log.trace("hasNext is false because yield was called");
+                        if (countingMap == null || countingMap.isEmpty()) {
+                            log.trace("There is no countingMap to flatten and i have yielded. Return false");
+                            return false;
+                        }
+                        Key positionAtYield = yieldCallback.getPositionAndReset();
+                        if (positionAtYield == null) {
+                            log.error("yielded with null key when lastEntry was null");
```
ivakegg: This sounds like an exceptional state of affairs....

tomnelson:
sorry. i cleaned it up but rebased in master. i'll fix it some other time.

PR #366, opened 2019-03-12
Fixes #365.  No longer assuming that typed fields = indexed fields

### LINKED ISSUE #365: Typed !=> Indexed
Long, long ago in a release far, far away, an assumption was made that all typed fields were indexed.  This is not the case and this assumption causes results to be missed if one of the query terms is indexed but not typed and vice versa.

hlgp:


warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -1672,6 +1684,10 @@ public static String compressOption(final String data, final Charset characterSe
         return new String(Base64.encodeBase64(byteStream.toByteArray()));
     }
     
+    /*
+     * TODO: The following two methods are more appropriately generified and put into utils.There is nothing indexOnly field specific about them at all. They
```
ivakegg: then just rename the method (e.g. buildStringFromFieldList)

ivakegg: I confirmed that PF does not use this method

hlgp: done in next commit

PR #368, opened 2019-03-13
Added a file auditor and audit replay service

jwomeara:
1) This PR adds the ability to save audit messages as ~~URL-encoded~~, JSON-formatted lines to either a local file, or a file in HDFS in the event that we are unable to successfully publish the message to RabbitMQ.  

2) This PR adds audit replay endpoints and functionality, which can be used to either read lines from a local file, or a file from HDFS, and then parse the ~~URL-encoded~~, json-formatted audit messages contained within that file, and replay them via the audit controller.


jwomeara:
@brianloss @ivakegg @FineAndDandy I think this is ready for you guys to pick apart.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
```diff
@@ -0,0 +1,726 @@
...
+import static datawave.microservice.audit.replay.status.Status.ReplayState.RUNNING;
+import static datawave.microservice.audit.replay.status.Status.ReplayState.STOPPED;
+
+/**
+ * The ReplayController presents the REST endpoints for audit replay.
+ *
+ */
+@RestController
+@RequestMapping(path = "/v1/replay", produces = MediaType.APPLICATION_JSON_VALUE)
+@ConditionalOnProperty(name = "audit.replay.enabled", havingValue = "true")
```
brianloss: Is there a reason we don't want this enabled by default? (matchIfMissing = true)? Also, most things in spring boot that do a conditionalonproperty don't do havingProperty of true, since that would exclude "True" for example. The default is for the value to not equal "false", which works well for something that is enabled by default and the user override is to disable.

services/audit-service/src/main/java/datawave/microservice/audit/AuditController.java
```diff
@@ -171,22 +184,33 @@ public String audit(@RequestParam MultiValueMap<String,String> parameters) {
...
-            log.info("[" + restAuditParams.getAuditId() + "] Audit successful. { attempts = " + attempts + ", elapsedMillis = " + (currentTime - auditStartTime)
-                            + "}");
-            
-            return restAuditParams.getAuditId();
+        // last ditch effort to write the audit message to fileSystem for subsequent processing
+        if (!success && fileAuditor != null) {
+            success = true;
+            try {
+                if (log.isDebugEnabled())
+                    log.debug("[" + auditParameters.getAuditId() + "] Attempting to log audit to the filesystem");
```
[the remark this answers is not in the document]
brianloss: Same goes for basically all of the log messages you've added in this pr.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
```diff
@@ -0,0 +1,726 @@
...
+                    resp = "Cannot start audit replay with id " + id;
+                    response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+                }
+            } else {
+                resp = "Cannot start audit replay with state " + status.getState();
+                response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+            }
+        } else {
+            resp = "No audit replay found with id " + id;
+            response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
```
brianloss: Usually when a resource is not found, a 404 (SC_NOT_FOUND) would be returned. Is there a reason for the bad request response?

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
```diff
@@ -0,0 +1,726 @@
...
+        // if the state is 'created', we can run the replay
+        if (status != null) {
+            if (status.getState() == CREATED) {
+                RunningReplay runningReplay = start(status);
+                if (runningReplay != null) {
+                    runningReplays.put(id, runningReplay);
+                    resp = "Started audit replay with id " + id;
+                } else {
+                    resp = "Cannot start audit replay with id " + id;
+                    response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
```
brianloss: Bad request is documented "The server cannot or will not process the request due to an apparent client error (e.g., malformed request syntax, size too large, invalid request message framing, or deceptive request routing)." Using this as a response code doesn't seem to fit here.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
```diff
@@ -0,0 +1,726 @@
...
+                }
+            } else {
+                resp = "Cannot start audit replay with state " + status.getState();
+                response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
```
brianloss: Same as above.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
```diff
@@ -0,0 +1,726 @@
...
+    @RolesAllowed({"AuthorizedUser", "AuthorizedServer", "InternalUser", "Administrator", "JBossAdministrator"})
+    @RequestMapping(path = "/{id}/status", method = org.springframework.web.bind.annotation.RequestMethod.GET)
+    public Status status(@ApiParam("The audit replay id") @PathVariable("id") String id, HttpServletResponse response) throws IOException {
+        
+        log.info("Getting status for audit replay with id " + id);
+        
+        Status status = status(id, replayProperties.isPublishEvents());
+        
+        if (status == null)
+            response.sendError(HttpServletResponse.SC_BAD_REQUEST, "No audit replay found with id " + id);
```
brianloss: Same comment as previous about sending a bad request here instead of a not found.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
```diff
@@ -0,0 +1,726 @@
...
+    @ApiOperation(value = "Lists the status for all audit replays.")
+    @RolesAllowed({"AuthorizedUser", "AuthorizedServer", "InternalUser", "Administrator", "JBossAdministrator"})
+    @RequestMapping(path = "/statusAll", method = org.springframework.web.bind.annotation.RequestMethod.GET)
+    public List<Status> statusAll() {
+        
+        log.info("Getting status for all audit replays");
+        
+        List<Status> statuses = statusCache.retrieveAll();
+        if (statuses != null)
+            statuses = statuses.stream().map(status -> idleCheck(status, replayProperties.isPublishEvents())).collect(Collectors.toList());
```
brianloss: It looks like idleCheck can never actually return a different status--it only has side effects. If that's the case, you could avoid duplicate list creation (and have more concise code) with:
```
if (statuses != null)
    statuses.forEach(status -> idleCheck(status, replayProperties.isPublishEvents());

File: services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java [hunk 8]
```diff
@@ -0,0 +1,726 @@
...
+        
+        // only update if the send rate is valid
+        if (sendRate >= 0) {
+            // pull the replay status from cache to ensure it exists
+            Status status = status(id, replayProperties.isPublishEvents());
+            if (status != null) {
+                update(status, sendRate, replayProperties.isPublishEvents());
+                resp = "Updated audit replay with id " + id;
+            } else {
+                response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
```
brianloss: not found instead of bad request

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
```diff
@@ -0,0 +1,726 @@
...
+    private int updateAll(long sendRate, boolean publishEnabled, boolean onlyUpdateRunning) {
+        int replaysUpdated = 0;
+        
+        if (publishEnabled)
+            appCtx.publishEvent(new AuditReplayRemoteRequestEvent(this, busProperties.getId(), Request.updateAll(sendRate)));
+        
+        // update all of our replays.
+        List<Status> statuses = statusAll();
+        if (onlyUpdateRunning)
+            statuses = statuses.stream().filter(status -> status.getState() == RUNNING).collect(Collectors.toList());
```
brianloss: I believe you an avoid the duplicate list creation with:
```
statuses.removeIf(status -> status.getState() != RUNNING)
```

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
```diff
@@ -0,0 +1,726 @@
...
+        log.info("Stopping audit replay with id " + id);
+        
+        String resp;
+        
+        Status status = status(id, replayProperties.isPublishEvents());
+        if (status != null) {
+            if (stop(status, replayProperties.isPublishEvents())) {
+                resp = "Stopped audit replay with id " + id;
+            } else {
+                response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
```
brianloss: same comments as before about the response type

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
```diff
@@ -0,0 +1,726 @@
...
+            resp = "No audit replay found with id " + id;
+        }
+        
+        log.info(resp);
+        return resp;
+    }
+    
+    private boolean stop(Status status, boolean publishEvent) {
+        // is the replay running?
+        if (status.getState() == RUNNING) {
```
brianloss: This is a lot to parse, but I'm wondering if these status object can be updated by multiple threads? If so, then you need to synchronize here, yes?

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
```diff
@@ -0,0 +1,726 @@
...
+        
+        log.info("Resuming audit replay with id " + id);
+        
+        String resp;
+        Status status = status(id, replayProperties.isPublishEvents());
+        if (status != null) {
+            if (resume(status)) {
+                resp = "Resumed audit replay with id " + id;
+            } else {
+                response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
```
brianloss: response code

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
```diff
@@ -0,0 +1,726 @@
...
+                resp = "Cannot resume audit replay with id " + id;
+            }
+        } else {
+            response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+            resp = "No audit replay found with id " + id;
```
brianloss: not found

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
```diff
@@ -0,0 +1,726 @@
...
+            if (status.getState() != RUNNING) {
+                statusCache.delete(status.getId());
+                resp = "Deleted audit replay with id " + id;
+            } else {
+                response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+                resp = "Cannot delete an audit replay with state " + status.getState();
+            }
+        } else {
+            response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+            resp = "No audit replay found with id " + id;
```
brianloss: not found

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
```diff
@@ -0,0 +1,726 @@
...
+        
+        String resp;
+        Status status = status(id, replayProperties.isPublishEvents());
+        
+        if (status != null) {
+            if (status.getState() != RUNNING) {
+                statusCache.delete(status.getId());
+                resp = "Deleted audit replay with id " + id;
+            } else {
+                response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
```
brianloss: response code

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
```diff
@@ -0,0 +1,726 @@
...
+    /**
+     * Deletes an audit replay
+     *
+     * @param id
+     *            The audit replay id
+     * @return status, indicating whether the audit replay was successfully deleted
+     */
+    @ApiOperation(value = "Deletes an audit replay.")
+    @RolesAllowed({"Administrator", "JBossAdministrator"})
+    @RequestMapping(path = "/{id}/delete", method = org.springframework.web.bind.annotation.RequestMethod.POST)
```
brianloss: Usually, delete methods would use the DELETE verb. Is there a reason you aren't?

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
```diff
@@ -0,0 +1,726 @@
...
+    /**
+     * Resumes an audit replay
+     *
+     * @param id
+     *            The audit replay id
+     * @return status, indicating whether the audit replay was successfully resumed
+     */
+    @ApiOperation(value = "Resumes an audit replay.")
+    @RolesAllowed({"Administrator", "JBossAdministrator"})
+    @RequestMapping(path = "/{id}/resume", method = org.springframework.web.bind.annotation.RequestMethod.POST)
```
brianloss: The "PUT" verb is usually used for update/modification of a resource.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
```diff
@@ -0,0 +1,726 @@
...
+    /**
+     * Stops an audit replay
+     *
+     * @param id
+     *            The audit replay id
+     * @return status, indicating whether the audit replay was successfully stopped
+     */
+    @ApiOperation(value = "Stops an audit replay.")
+    @RolesAllowed({"Administrator", "JBossAdministrator"})
+    @RequestMapping(path = "/{id}/stop", method = org.springframework.web.bind.annotation.RequestMethod.POST)
```
brianloss: The "PUT" verb is usually used for update/modification of a resource.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
```diff
@@ -0,0 +1,726 @@
...
+    /**
+     * Updates all audit replays
+     *
+     * @param sendRate
+     *            The number of messages to send per second
+     * @return status, indicating the number of audit replays which were successfully updated
+     */
+    @ApiOperation(value = "Updates all audit replays.")
+    @RolesAllowed({"Administrator", "JBossAdministrator"})
+    @RequestMapping(path = "/updateAll", method = org.springframework.web.bind.annotation.RequestMethod.POST)
```
brianloss: The "PUT" verb is usually used for update/modification of a resource.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
```diff
@@ -0,0 +1,726 @@
...
+     *
+     * @param id
+     *            The audit replay id
+     * @param sendRate
+     *            The number of messages to send per second
+     * @return status, indicating whether the update was successful
+     */
+    @ApiOperation(value = "Updates an audit replay.")
+    @RolesAllowed({"Administrator", "JBossAdministrator"})
+    @RequestMapping(path = "/{id}/update", method = org.springframework.web.bind.annotation.RequestMethod.POST)
```
brianloss: The "PUT" verb is usually used for update/modification of a resource.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
```diff
@@ -0,0 +1,726 @@
...
+    }
+    
+    /**
+     * Deletes all audit replays
+     *
+     * @return status, indicating the number of audit replays which were successfully deleted
+     */
+    @ApiOperation(value = "Deletes all audit replays.")
+    @RolesAllowed({"Administrator", "JBossAdministrator"})
+    @RequestMapping(path = "/deleteAll", method = org.springframework.web.bind.annotation.RequestMethod.POST)
```
brianloss: The "DELETE" verb would normally be used for deleting resources.

services/audit-service/src/main/java/datawave/microservice/audit/replay/config/ReplayConfig.java
```diff
@@ -0,0 +1,39 @@
...
+import org.springframework.cache.annotation.EnableCaching;
+import org.springframework.cache.concurrent.ConcurrentMapCacheManager;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
+
+@Configuration
+@EnableCaching
+@EnableConfigurationProperties(ReplayProperties.class)
+@ConditionalOnProperty(name = "audit.replay.enabled", havingValue = "true")
```
brianloss: Same comments as before about the conditionalonproperty.

services/audit-service/src/main/java/datawave/microservice/audit/replay/config/ReplayProperties.java
```diff
@@ -0,0 +1,103 @@
+package datawave.microservice.audit.replay.config;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+import java.util.concurrent.TimeUnit;
+
+@ConfigurationProperties(prefix = "audit.replay")
+public class ReplayProperties {
```
brianloss: Maybe add validation (using validation annotations) to this? The millis values, I would guess at least all need to be positive, for example.

services/audit-service/src/main/java/datawave/microservice/audit/replay/config/ReplayConfig.java
```diff
@@ -0,0 +1,39 @@
...
+public class ReplayConfig {
+    
+    @Bean
+    public ThreadPoolTaskExecutor auditReplayExecutor(ReplayProperties replayProperties) {
```
brianloss: If the context is refreshed, this may cause thread leaks when this bean is re-created. May need to look into that.

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/FileAuditor.java
```diff
@@ -0,0 +1,220 @@
...
+        config.set("fs.hdfs.impl", org.apache.hadoop.hdfs.DistributedFileSystem.class.getName());
+        config.set("fs.file.impl", org.apache.hadoop.fs.LocalFileSystem.class.getName());
+        
+        if (builder.configResources != null) {
+            for (String resource : builder.configResources)
+                config.addResource(new Path(resource));
+        }
+        
+        if (builder.fileUri != null)
+            fileSystem = FileSystem.get(new URI(builder.fileUri), config);
```
ivakegg: Lets use the basePath here instead.  Also suggest configuring basePath with the "fileUri" all in one property (e.g. file:///some/dir or hdfs://warehouse-a/some/dir)

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/FileAuditor.java
```diff
@@ -0,0 +1,220 @@
...
+    
+    protected Path currentFile = null;
+    protected Date creationDate = null;
+    
+    protected FileAuditor(Builder<?> builder) throws URISyntaxException, IOException {
+        this.maxFileLenBytes = builder.maxFileLenBytes;
+        this.maxFileAgeMillis = builder.maxFileAgeMillis;
+        
+        Configuration config = new Configuration();
+        config.set("fs.hdfs.impl", org.apache.hadoop.hdfs.DistributedFileSystem.class.getName());
```
ivakegg: This seems odd.  If we have the hadoop configuration mounted to the proper place, then "new Configuration()" will already load everything we need.

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/FileAuditor.java
```diff
@@ -0,0 +1,220 @@
...
+        
+        if (builder.configResources != null) {
+            for (String resource : builder.configResources)
+                config.addResource(new Path(resource));
+        }
+        
+        if (builder.fileUri != null)
+            fileSystem = FileSystem.get(new URI(builder.fileUri), config);
+        else
+            fileSystem = FileSystem.get(config);
```
ivakegg: this should never be allowed.  We always need a basePath (with protocol specified).  We need to fail if it is not specified.

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/FileAuditor.java
```diff
@@ -0,0 +1,220 @@
...
+    
+    protected static final ObjectMapper mapper = new ObjectMapper();
+    
+    protected final SimpleDateFormat sdf;
+    protected final ReentrantLock writeLock = new ReentrantLock(true);
+    
+    protected long maxFileLenBytes;
+    protected long maxFileAgeMillis;
+    
+    protected FileSystem fileSystem;
```
ivakegg: Note that this is a Closeable object.  If this will exist for the lifetime of the jvm, then no problem.  Otherwise we need a close/cleanup mechanism.  A configuration refresh may be a path of concern here.

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/FileAuditor.java
```diff
@@ -0,0 +1,220 @@
...
+            if (currentDate != null) {
+                try {
+                    Thread.sleep((long) (Math.random() * 10));
+                } catch (InterruptedException e) {
+                    // getting interrupted is ok
+                }
+            }
+            
+            currentDate = new Date();
+            currentFile = new Path(basePath, sdf.format(currentDate));
```
ivakegg: Can we use podname or hostname instead to avoid the possibility of conflicts?  The only reason I say this is hdfs can be troublesome as a locking mechanism.

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/FileAuditor.java
```diff
@@ -0,0 +1,220 @@
...
+        if (builder.fileUri != null)
+            fileSystem = FileSystem.get(new URI(builder.fileUri), config);
+        else
+            fileSystem = FileSystem.get(config);
+        
+        basePath = new Path(builder.path);
+        if (!fileSystem.exists(basePath))
+            fileSystem.mkdirs(basePath);
+        
+        this.sdf = new SimpleDateFormat("'" + builder.prefix + "'-yyyyMMdd_HHmmss.SSS'.json'");
```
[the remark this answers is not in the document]
jwomeara: The sdf created here is only used in the createNewFile method.  In the current implementation, the createNewFile method is only ever called if the calling thread is able to acquire a lock on the writeLock object.  I think this should be enough to prevent multiple threads from accessing the sdf.

services/audit-service/src/main/java/datawave/microservice/audit/AuditController.java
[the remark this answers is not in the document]
jwomeara: Sure, this will be addressed in the next commit.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
jwomeara: This is mostly just a personal preference of mine to have non-critical features disabled unless they are explicitly enabled.  I hadn't thought about the 'true' vs. 'True' bit...  This is definitely not the only place I'm using ConditionalOnProperty this way.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
jwomeara: This will be addressed in the next commit.  I revamped the status codes across the board.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
jwomeara: The next commit will have this changed to SC_INTERNAL_SERVER_ERROR.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
jwomeara: If the current state of the replay is not 'CREATED', then start is an invalid request.  Is there a better status code which reflects that?

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
jwomeara: Good catch.  Updated in the next commit.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
jwomeara: I didn't know about that.  Updated in the next commit.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
brianloss: I'm sure there's a suitable one in https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
jwomeara: If the audit replay is not running, and you try to stop it, is that not a bad request?  Should that be SC_FORBIDDEN or something like that instead?

brianloss: Maybe a 409 fits better. All I know is the syntax of the request is fine, so 400 isn't the right answer.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
jwomeara: Updated in the next commit.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
jwomeara: Updated in the next commit.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
jwomeara: Updated in the next commit.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
jwomeara: Updated in the next commit.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
jwomeara: Updated in the next commit.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
jwomeara: Updated in the next commit.

services/audit-service/src/main/java/datawave/microservice/audit/replay/config/ReplayConfig.java
jwomeara: I'm working on a test for the next commit.

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/FileAuditor.java
jwomeara: Working on this for the next commit.

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/FileAuditor.java
jwomeara: Working on this for the next commit.

services/audit-service/pom.xml
```diff
@@ -22,33 +24,91 @@
...
+                        <artifactId>*</artifactId>
+                        <groupId>log4j</groupId>
+                    </exclusion>
+                    <exclusion>
+                        <artifactId>*</artifactId>
+                        <groupId>commons-logging</groupId>
+                    </exclusion>
+                </exclusions>
+            </dependency>
+            <dependency>
```
brianloss: spring-boot-starter-cache and datawave-in-memory-accumulo (test scope) are now included via spring-boot-starter-datawave so they can be removed from here.

services/audit-service/src/main/java/datawave/microservice/audit/replay/util/ConcurrentMapCacheInspector.java
```diff
@@ -0,0 +1,61 @@
+package datawave.microservice.audit.replay.util;
```
brianloss: hazelcast-common now includes a ConcurrentMapCacheInspector, so you might be able to drop this class.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
jwomeara: I thought about this a bit...  A status could be updated by multiple threads, or even by multiple audit services depending upon which pod you get when you hit the audit endpoint.  

To protect against race conditions I'll need some kind of distributed lock.  From the looks of it, Hazelcast has locking functionality built in.  I suppose I could either update the existing HazelcastCacheInspector, or extend it and add some methods which expose the locking functionality.  Thoughts?

jwomeara: I'll add that given how I envision the replay endpoints being used, I doubt we would run encounter race condition issues with the code as-is, but I don't think it will be hard to utilize the Hazelcast locking functionality, and that would solve the problem.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
jwomeara: I am leaning towards 422 - UNPROCESSABLE_ENTITY as this implies that the server understands the request, it is syntactically correct, but was unable to process the instructions.  The error message I am sending along with this status code should make it clear what went wrong.

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
jwomeara: Ok, so the next commit will update the restcontroller to acquire a lock on the Hazelcast map before updating the status of a replay.  For running replays, updates and stop commands are forwarded to the running replay object, which will be responsible for updating the cache.

services/audit-service/src/main/java/datawave/microservice/audit/replay/config/ReplayProperties.java
jwomeara: added in next commit.

services/audit-service/src/main/java/datawave/microservice/audit/replay/config/ReplayConfig.java
jwomeara: I tested this.  As it is now, nothing happens on a refresh event.  However, if I mark the ThreadPoolTaskExecutor as @RefreshScope, any running replays are killed.  That's ok though, because eventually they will be marked as IDLE, at which point they can be resumed.  

For now, I am going to leave the RefreshScope annotation out.  What I will do instead is create a separate ticket to go over the entire audit service and try to get it configured to handle refresh appropriately.

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/FileAuditor.java
jwomeara: It will live the lifetime of the container, so I'll leave it alone for now.

services/audit-service/pom.xml
jwomeara: fixed in next commit

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/FileAuditor.java
jwomeara: fixed in next commit

jwomeara:
I rebased the branch against master and added a commit which should resolve all of the feedback that was given.  Let me know what you guys think when you get a chance.  @ivakegg @brianloss

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/FileAuditor.java
```diff
@@ -0,0 +1,212 @@
...
+/**
+ * An implementation for {@link Auditor}, which writes URL encoded, JSON formatted audit messages to a file.
+ */
+public class FileAuditor implements Auditor {
+    
+    private final Logger log = LoggerFactory.getLogger(this.getClass());
+    
+    protected static final ObjectMapper mapper = new ObjectMapper();
+    
+    protected final SimpleDateFormat sdf;
```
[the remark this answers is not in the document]
FineAndDandy: since the create file call is wrapped inside the writeLock this is probably fine

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/config/FileAuditConfig.java
```diff
@@ -0,0 +1,51 @@
...
+import javax.validation.Valid;
+import java.util.List;
+
+/**
+ * Configures an FileAuditor to process messages received by the audit service in the case that our messaging infrastructure has failed. This configuration is
+ * activated via the 'audit.auditors.file.enabled' property.
+ *
+ */
+@Configuration
+@ConditionalOnProperty(name = "audit.auditors.file.enabled", havingValue = "true")
```
FineAndDandy: how is this different than the dump auditor?

services/audit-service/src/main/java/datawave/microservice/audit/replay/ReplayController.java
FineAndDandy: this looks good

services/audit-service/src/main/java/datawave/microservice/audit/replay/config/ReplayProperties.java
```diff
@@ -0,0 +1,146 @@
...
+    private long stopGracePeriodMillis = 500L;
+    
+    @PositiveOrZero
+    private long statusUpdateIntervalMillis = TimeUnit.SECONDS.toMillis(1);
+    
+    @PositiveOrZero
+    private long lockWaitTimeMillis = TimeUnit.SECONDS.toMillis(5);
+    
+    @PositiveOrZero
+    private long lockLeaseTimeMillis = TimeUnit.SECONDS.toMillis(5);
```
FineAndDandy: Can you better describe the difference between lease/lock timeouts? I'm not clear on when I would use one over the other and the impacts

services/audit-service/pom.xml
```diff
@@ -56,6 +125,13 @@
         </dependency>
     </dependencies>
     <repositories>
+        <repository>
```
brianloss: I hate having 3rd party repos in here. Do we really need to depend on the Cloudera Hadoop version for what we're doing, or would it make more sense to depend on stock? Could see an argument both ways...

services/audit-service/src/main/java/datawave/microservice/audit/replay/util/LockableCacheInspector.java
```diff
@@ -0,0 +1,22 @@
+package datawave.microservice.audit.replay.util;
```
brianloss: This doesn't seem audit-specific to me. Why isn't it in hazelcast-client?

services/audit-service/src/main/java/datawave/microservice/audit/replay/util/LockableHazelcastCacheInspector.java
```diff
@@ -0,0 +1,91 @@
+package datawave.microservice.audit.replay.util;
```
brianloss: This doesn't seem audit-specific to me. Why isn't it in hazelcast-client?

services/audit-service/src/main/java/datawave/microservice/audit/replay/util/UniversalLockableCacheInspector.java
```diff
@@ -0,0 +1,83 @@
+package datawave.microservice.audit.replay.util;
```
brianloss: This doesn't seem audit-specific to me. Why isn't it in hazelcast-client?

services/sample_configuration/audit-dev.yml.example
```diff
@@ -4,6 +4,11 @@
 server:
   port: 8743
 
+# Don't use a Hazelcast client, but rather use an embedded Hazelcast member. This is
+# only useful for development, and in production, this property shouldnot be set (or
```
brianloss: "should not" instead of "shouldnot"

services/audit-service/src/main/java/datawave/microservice/audit/AuditController.java
```diff
@@ -171,22 +184,33 @@ public String audit(@RequestParam MultiValueMap<String,String> parameters) {
...
-        } else {
-            log.info("[" + restAuditParams.getAuditId() + "] Audit successful. { attempts = " + attempts + ", elapsedMillis = " + (currentTime - auditStartTime)
-                            + "}");
-            
-            return restAuditParams.getAuditId();
+        // last ditch effort to write the audit message to fileSystem for subsequent processing
+        if (!success && fileAuditor != null) {
+            success = true;
+            try {
+                if (log.isDebugEnabled())
```
brianloss: Unless the method call `auditParameters.getAuditId()` is expensive, there's no need to guard this statement with `if (log.isDebugEnabled())`

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/FileAuditor.java
```diff
@@ -0,0 +1,212 @@
...
+            if (currentFile == null || !fileSystem.exists(currentFile) || isFileTooOld() || isFileTooBig())
+                createNewFile();
+            
+            writeAudit(jsonAuditParams);
+        } finally {
+            writeLock.unlock();
+        }
+    }
+    
+    protected String urlEncodeString(String value) {
```
brianloss: Why the need for this? If the ObjectMapper won't do the URL encoding for you, then what's the purpose of catching the exception and returning the original value? If that case happens, then the object mapper will just fail on it. If the object mapper does take care of the encoding, then there's no need for this method.

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/FileAuditor.java
```diff
@@ -0,0 +1,212 @@
...
+            return URLEncoder.encode(value, "UTF8");
+        } catch (UnsupportedEncodingException e) {
+            log.error("Unable to URL encode value: {}", value);
+        }
+        return value;
+    }
+    
+    protected void writeAudit(String jsonAuditParams) throws Exception {
+        OutputStream appendStream = (fileSystem instanceof LocalFileSystem) ? new FileOutputStream(new File(currentFile.toUri()), true)
+                        : fileSystem.append(currentFile);
```
brianloss: Just curious... Why the complexity here on testing the file system instance? Does LocalFileSystem not support the append method?

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/config/FileAuditProperties.java
```diff
@@ -0,0 +1,83 @@
...
+import org.springframework.validation.annotation.Validated;
+
+import javax.validation.Valid;
+import javax.validation.constraints.DecimalMin;
+import javax.validation.constraints.NotNull;
+
+@Validated
+public class FileAuditProperties {
+    
+    @NotNull
```
brianloss: `@NotBlank`, or is an empty string allowed for the pathUri?

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/FileAuditor.java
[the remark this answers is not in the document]
jwomeara: That was my thought as well.

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/config/FileAuditConfig.java
jwomeara: The file auditor our last ditch attempt at auditing, used when all other attempts to audit have failed (at the audit rest endpoint).  

The dump auditor is listening to the audit.dump queue, and writes any messages moved there to disk.  So, if you have some stubborn messages in your DLQ that you need to analyze, you can move them to the audit.dump queue, and then work with the file on disk.

services/audit-service/src/main/java/datawave/microservice/audit/replay/config/ReplayProperties.java
jwomeara: lockWaitTimeMillis is the amount of time we will wait to acquire a lock.

lockLeaseTimeMillis is the amount of time we are allowed to hold the lock before it is automatically released.  This is meant to safeguard against cases where an audit service instance goes down while processing a locking call.  If we didn't have this lease timeout, and our task died mid-call, the lock would never be released.

services/audit-service/src/main/java/datawave/microservice/audit/replay/util/LockableCacheInspector.java
jwomeara: I have no objection to that.  Do you think it should still be a separate interface, or should I just add the methods to the CacheInspector interface?

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/FileAuditor.java
jwomeara: It does not, unfortunately.  :(

services/audit-service/src/main/java/datawave/microservice/audit/replay/util/LockableCacheInspector.java
brianloss: I'd keep it separate. We don't want to require that every cache type supports locking functions.

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/FileAuditor.java
brianloss: Hmm, weird. It implements the method. Oh well...

services/audit-service/pom.xml
jwomeara: I don't have a dog in this fight.  I was just trying to carry forward the precedent set in the warehouse pom.

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/FileAuditor.java
jwomeara: I'm just thinking ahead to the moment when I will be reading these json-formatted files in spark.  Spark natively supports reading json-formatted data, where each json object is written on a single line.

Multi-line json objects are also supported, but you have to write all of the json objects as an array, which isn't as convenient when writing in append mode (basically, it would be nice to avoid having to write beginning & closing square brackets).  

To get around this, I was planning on just URL-encoding the parameter values to ensure that no newlines would be present in a given json object.  This is easy to undo in spark when it comes time to process the data.  

To my knowledge, you don't just get URL encoding for free by using an object mapper, so that's why I have the extra step.  Also, while I do have logic for catching an unsupported encoding exception, you can see that it's hard coded to UTF8 (which is a supported character encoding), so I don't expect that exception to ever be thrown.  

Alternatively, I suppose I could write Base64 encoded bytes instead and URL encoding altogether, but i thought this would be good enough.  

What are your thoughts?

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/config/FileAuditProperties.java
jwomeara: updated.

services/audit-service/src/main/java/datawave/microservice/audit/AuditController.java
jwomeara: updated.

services/sample_configuration/audit-dev.yml.example
jwomeara: updated in multiple places.

services/audit-service/src/main/java/datawave/microservice/audit/replay/util/UniversalLockableCacheInspector.java
jwomeara: moved.

services/audit-service/src/main/java/datawave/microservice/audit/replay/util/LockableHazelcastCacheInspector.java
jwomeara: moved.

services/audit-service/src/main/java/datawave/microservice/audit/replay/config/ReplayProperties.java
jwomeara: Added a description in the latest commit.

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/FileAuditor.java
jwomeara: During testing, it ultimately boiled down to this:
https://github.com/apache/hadoop/blob/release-2.6.0/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/ChecksumFileSystem.java#L352

services/audit-service/src/main/java/datawave/microservice/audit/replay/util/LockableCacheInspector.java
jwomeara: ok, cool.  that's how it is in the latest commit.

services/audit-service/src/main/java/datawave/microservice/audit/auditors/file/FileAuditor.java
ivakegg: The URL encoding sounds a like hoky to me.  In fact any nested encoding makes the file difficult to parse in other ways.  I would rather create an appropriate json output that can be readable by more than the spark processing.

ivakegg: Sounds like some sort of escaping is required in any case, but using a standard JSON writer would avoid having to do anything like this.  I understand your Spark concerns, but I would rather have standardized json output that Spark can handle instead of having to do a urlDecode wherever we have to read these files.

jwomeara: I had assumed that I would need to URL encode to ensure that the json stays as a single line (to play nicely with Spark), but after testing out the object mapper without doing any URL encoding, I think that will work just fine.  I updated the pr.

PR #377, opened 2019-03-19
fixes #371: limit source passed to Aggregator from IndexIterator

### LINKED ISSUE #371: Aggregators used from the IndexIterator's may need more specific ranges
The range produced from the IndexIterator when no document specific range exists will be from:
row to row\0 with a columnFamiliy specifying the fi\0field. This means that the aggregator will have an unbound source that will iterate until the end of the document (samePointer returns false). In this case it would be better to either restrict the source or to pass the relevant information to the aggregator so that only the desired keys are looped over.

FineAndDandy:


warehouse/query-core/src/main/java/datawave/query/iterator/logic/IndexIterator.java
```diff
@@ -95,6 +96,7 @@ public IndexIterator build() {
     public static final String INDEX_FILTERING_CLASSES = "indexfiltering.classes";
     
     protected SortedKeyValueIterator<Key,Value> source;
+    protected LimitedSortedKeyValueIterator limitedSource;
```
drewfarris: Would it be possible to do this in a way that we only interact with `limitedSource` within this class, and remove `source` as a field -or- simply assign the `LimitedSortedKeyValueIterator` to the `source` field?

FineAndDandy: we don't actually want to limit the source generally, just for the aggregator. This is to avoid re-seeking the source between aggregator calls

drewfarris: got it. thanks.

warehouse/query-core/src/test/java/datawave/query/iterator/QueryIteratorIT.java
```diff
@@ -0,0 +1,831 @@
...
+import static datawave.query.iterator.QueryOptions.QUERY_ID;
+import static datawave.query.iterator.QueryOptions.SERIAL_EVALUATION_PIPELINE;
+import static datawave.query.iterator.QueryOptions.START_TIME;
+import static datawave.query.iterator.QueryOptions.TERM_FREQUENCIES_REQUIRED;
+
+/**
+ * Integration tests for the QueryIterator
+ *
+ */
+public class QueryIteratorIT extends EasyMockSupport {
```
ivakegg: awesome!  Thank You!

PR #379, opened 2019-03-21
fixes #372: Forward vs Reverse index

### LINKED ISSUE #372: Indexed vs Reverse Indexed
It appears that if a term is indexed but not reverse indexed, then searching for terms with leading wildcards will simply return 0 results.  Instead the code should take this into account and delay those terms if possible or return an error.

ivakegg:
* Push down terms that cannot be expanded against the index
* Ensure we use the correct index based on the regex and which indexes are filled

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ShardIndexQueryTableStaticMethods.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ShardIndexQueryTableStaticMethods.java
@@ -9,6 +9,7 @@ import datawave.core.iterators.filter.GlobalIndexTermMatchingFilter;
 import datawave.data.type.Type;
 import datawave.query.Constants;
 import datawave.query.config.ShardQueryConfiguration;
+import datawave.query.exceptions.DatawaveFatalQueryException;
 import datawave.query.exceptions.DoNotPerformOptimizedQueryException;
 import datawave.query.exceptions.IllegalRangeArgumentException;
 import datawave.query.jexl.JexlASTHelper;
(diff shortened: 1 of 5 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ShardIndexQueryTableStaticMethods.java
```diff
@@ -587,44 +599,60 @@ public static RefactoredRangeDescription getRegexRange(Map.Entry<String,String>
...
+     * 
+     * @param analyzer
+     * @param fieldName
+     * @param metadataHelper
+     * @param config
+     * @return
+     * @throws TableNotFoundException
+     * @throws ExecutionException
+     */
     public static boolean shouldUseReverseIndex(JavaRegexAnalyzer analyzer, String fieldName, MetadataHelper metadataHelper, ShardQueryConfiguration config)
```
FineAndDandy: there is similar logic in the IteratorBuildingVisitor, should these maybe be combined and extracted out?

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ShardIndexQueryTableStaticMethods.java
```diff
@@ -587,44 +599,60 @@ public static RefactoredRangeDescription getRegexRange(Map.Entry<String,String>
...
-            boolean isForwardIndexed = (null != fieldName) ? indexedInDatatype(fieldName, datatypeFilter, metadataHelper) : true;
-            
-            // If we're only reverse indexed, gotta use the reverse
-            if (!isForwardIndexed && isReverseIndexed) {
+        }
+        // we have neither leading or trailing, use foward index
+        if (trailingLiteral == null && leadingLiteral == null) {
+            return false;
+        }
+        // we have both leading and trailing. use the biggest one (sans the realm)
```
FineAndDandy: we might want to default to forward index because the FI is forward indexed only

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ShardIndexQueryTableStaticMethods.java
ivakegg: good call, I will take a look.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ShardIndexQueryTableStaticMethods.java
FineAndDandy: Disregard this, not relevant for global index

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ShardIndexQueryTableStaticMethods.java
ivakegg: After contemplating this one, I do not believe we can share code:  the IteratorBuildingVisitor which is for running against the field index does not have any reverse index to deal with whereas we do with the global indexes.

ivakegg: After contemplating this one, I do not believe we can share code: the IteratorBuildingVisitor which is for running against the field index does not have any reverse index to deal with whereas we do with the global indexes.

PR #381, opened 2019-03-22
Update to Wildfly 16.0.0.Final fixes #380

### LINKED ISSUE #380: Update Wildfly
Our version of Wildfly is very far behind, and is beginning to cause problems. Recently, the dependent version of Jackson was updated due to CVEs in the included version. However, this change actually breaks the web service. That is due to the fact that Wildfly provides Jackson as a runtime dependency and the version provided by Wildfly 10 is much older than the version referenced in the POM. Therefore, when certain actions happen (such as attempting to encode a JWT for invoking the audit service), we get a NoSuchMethodError due to the compile and runtime discrepancy in Jackson versions.

Given this, it makes sense to update to a newer Wildfly that has dependencies that are not so far out of date.

brianloss:
Update to wildfly 16.0.0.
Update dependent versions of libs included by Wildfly.
Update certain specifications for CDI, EJB to the latest version.


warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/EventMapperTest.java
```diff
@@ -15,30 +15,38 @@
...
 public class EventMapperTest {
     
+    @Rule
+    public EasyMockRule easyMockRule = new EasyMockRule(this);
+    
+    @Mock
+    private Mapper.Context mapContext;
+    
     private Configuration conf;
-    private MapDriver<LongWritable,RawRecordContainer,BulkIngestKey,Value> driver;
```
keith-ratcliffe: Wasn't expecting to see changes here of all places. Funny...the hoops we jump thru sometimes to beat dependencies into submission :) All in all though, was expecting a lot more pain to be inflicted by this upgrade. This is great

brianloss: Yeah, it seems mrunit was retired years ago. I forget exactly what tripped updating it-I think I got there through Arquillian updates. Overall, the end changes aren’t too bad. We are, however, completely ignoring wildfly’s new security system and using the old one—took a while to find that as the best option. But as long as it’s still there, we can keep upgrading...

contrib/datawave-quickstart/bin/services/datawave/bootstrap-web.sh
```diff
@@ -1,19 +1,20 @@
...
 
-export WILDFLY_HOME="${DW_CLOUD_HOME}/${DW_WILDFLY_SYMLINK}"
+export WILDFLY_HOME="$(readlink -f ${DW_CLOUD_HOME}/${DW_WILDFLY_SYMLINK})"
 export JBOSS_HOME="${WILDFLY_HOME}"
 export PATH="${WILDFLY_HOME}/bin:${PATH}"
+export DW_DATAWAVE_WEB_JAVA_OPTS="-Duser.timezone=GMT -Dfile.encoding=UTF-8"
 
-DW_DATAWAVE_WEB_CMD_START="( cd "${WILDFLY_HOME}/bin" && nohup ./standalone.sh -c standalone-full.xml & )"
-DW_DATAWAVE_WEB_CMD_START_DEBUG="( cd "${WILDFLY_HOME}/bin" && nohup ./standalone.sh --debug -c standalone-full.xml & )"
+DW_DATAWAVE_WEB_CMD_START="( cd "${WILDFLY_HOME}/bin" && JAVA_OPTS=\"$DW_DATAWAVE_WEB_JAVA_OPTS\" nohup ./standalone.sh -c standalone-full.xml & )"
```
ivakegg: So the DW_DATAWAVE_WEB_JAVA_OPTS no longer exist or are they used directly in the standalone.sh script?

ivakegg: never mind... you added it, not removed it....DOH

brianloss:
@keith-ratcliffe I added a commit to this, which dismissed your review. When you get a second, could you review the new commit and comment or approve? Thanks!

PR #385, opened 2019-03-28
fixes #384: allow multiple close requests

### LINKED ISSUE #384: A close request should not necessarily fail
If a query is closed on a webserver (either manually or automatically), a subsequent close request for that query should not fail.

ivakegg:
Added a cache of closed ids to allow subsequent close requests to succeed


web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -1956,9 +1960,12 @@ private VoidResponse close(String id, Principal principal) {
                     RunningQuery query = getQueryById(id, principal);
                     close(query);
```
matthpeterson: maybe it would be worth checking the cache before doing any additional work, but it seems either way will work.

ivakegg: I debated that, however there are several steps that occur and I wanted at least an administrator to attempt the close process all the way through in case something did not take.  Also there are some subtle differences between the cancel, close, and remove that in fact I am a little confused about.  This seemed the safest solution.

PR #388, opened 2019-04-01
Added unit test for ValidPatternVisitor.

apmoriarty:


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidPatternVisitor.java
```diff
@@ -25,13 +25,21 @@ public ValidPatternVisitor() {
     
     public static void check(JexlNode node) {
         ValidPatternVisitor visitor = new ValidPatternVisitor();
-        
         node.jjtAccept(visitor, null);
     }
     
+    /**
+     * Visit a Regex Equals node, catches the situation where a users might enter FIELD1 =~ FIELD2.
```
FineAndDandy: FIELD1 =~ VALUE1?

FineAndDandy: this generally should be fixed in all the comments

apmoriarty: Fixed in latest commit

PR #395, opened 2019-04-03
re #384: Made the closed query cache a singleton, and added better lo…

ivakegg:
…gging

### DIFF
--- a/web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
+++ b/web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
@@ -237,7 +237,7 @@ public class QueryExecutorBean implements QueryExecutor {
     
     private Multimap<String,PatternWrapper> traceInfos;
     private CacheListener traceCacheListener;
-
+    
     @Inject
     private ClosedQueryCache closedQueryCache;
     
@@ -1876,11 +1876,13 @@ public class QueryExecutorBean implements QueryExecutor {
             }
             
             QueryException qe = new QueryException(DatawaveErrorCode.QUERY_NEXT_ERROR, e, MessageFormat.format("query id: {0}", id));
-            log.error(qe, e);
-            response.addException(qe.getBottomQueryException());
             if (e.getCause() instanceof NoResultsException) {
+                log.debug("Got a nested NoResultsException", e);
                 close(id);
                 closedQueryCache.add(id); // remember that we auto-closed this query
+            } else {
+                log.error(qe, e);
+                response.addException(qe.getBottomQueryException());
             }
             int statusCode = qe.getBottomQueryException().getStatusCode();
             throw new DatawaveWebApplicationException(qe, response, statusCode);
@@ -1961,10 +1963,10 @@ public class QueryExecutorBean implements QueryExecutor {
                     RunningQuery query = getQueryById(id, principal);
                     close(query);
                 } catch (Exception e) {
-                    log.warn("Failed to close " + id + ", checking if closed previously");
+                    log.debug("Failed to close " + id + ", checking if closed previously");
                     // if this is a query that is in the closed query cache, then we have already successfully closed this query so ignore
                     if (!closedQueryCache.exists(id)) {
-                        log.warn("Failed to close " + id + ", checking if connection request was canceled");
+                        log.debug("Failed to close " + id + ", checking if connection request was canceled");
                         // if connection request was canceled, then the call was successful even if a RunningQuery was not found
                         if (!connectionRequestCanceled) {
                             log.error("Failed to close " + id, e);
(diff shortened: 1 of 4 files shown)


web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -1875,10 +1876,11 @@ private BaseQueryResponse next(final String id, boolean checkForContentLookup) {
             }
             
             QueryException qe = new QueryException(DatawaveErrorCode.QUERY_NEXT_ERROR, e, MessageFormat.format("query id: {0}", id));
-            log.error(qe);
+            log.error(qe, e);
```
brianloss: I'm not sure we want to log the full stack trace to the error channel if this is a wrapped NoResultsException. (We may also not want to add the exception to the response in that case either.)

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -1959,10 +1961,13 @@ private VoidResponse close(String id, Principal principal) {
                     RunningQuery query = getQueryById(id, principal);
                     close(query);
                 } catch (Exception e) {
-                    // if this is a query that is in the old query cache, then we have already successfully closed this query so ignore
-                    if (oldQueryCache.getIfPresent(id) == null) {
+                    log.warn("Failed to close " + id + ", checking if closed previously");
```
brianloss: It seems this is a case that we're expecting to happen, so we probably want to log at info or debug.

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -2019,10 +2024,13 @@ public VoidResponse adminClose(@Required("id") @PathParam("id") String id) {
                     RunningQuery query = adminGetQueryById(id);
                     close(query);
                 } catch (Exception e) {
-                    // if this is a query that is in the old query cache, then we have already successfully closed this query so ignore
-                    if (oldQueryCache.getIfPresent(id) == null) {
+                    log.warn("Failed to adminClose " + id + ", checking if closed previously");
```
brianloss: adminClose is for administrators closing a user's query. I would think if the administrator submits that request, and the user query happened to complete/return a 204 immediately before, we would still want this request to fail so the administrator can look into what happened.

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -2130,10 +2135,13 @@ public VoidResponse cancel(@Required("id") @PathParam("id") String id) {
                     query.cancel();
                     close(query);
                 } catch (Exception e) {
-                    // if this is a query that is in the old query cache, then we have already successfully closed this query so ignore
-                    if (oldQueryCache.getIfPresent(id) == null) {
+                    log.warn("Failed to cancel " + id + ", checking if closed previously");
```
brianloss: Log at info or debug, since this is an expected case?

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -2191,10 +2199,13 @@ public VoidResponse adminCancel(@Required("id") @PathParam("id") String id) {
                     query.cancel();
                     close(query);
                 } catch (Exception e) {
-                    // if this is a query that is in the old query cache, then we have already successfully closed this query so ignore
-                    if (oldQueryCache.getIfPresent(id) == null) {
+                    log.warn("Failed to adminCancel " + id + ", checking if closed previously");
```
brianloss: Same as for adminClose... I'm not sure we want to "hide" the error if the admin cancels a query that has already completed.

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -2394,25 +2405,28 @@ public VoidResponse remove(@Required("id") @PathParam("id") String id) {
                 close(query);
                 persister.remove(query.getSettings());
             } catch (Exception e) {
-                // if this is a query that is in the old query cache, then we have already successfully closed this query so ignore
-                if (oldQueryCache.getIfPresent(id) == null) {
+                log.warn("Failed to remove " + id + ", checking if closed previously");
```
brianloss: Same comment about log level.

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
ivakegg: that is reasonable, I will change the pattern

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
ivakegg: agreed...will do

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
ivakegg: agreed

Issue #373, opened 2019-03-15
[SECURITY] Releases are built/executed/released in the context of insecure/untrusted code


JLLeitschuh:
[CWE-829: Inclusion of Functionality from Untrusted Control Sphere](https://cwe.mitre.org/data/definitions/829.html)
[CWE-494: Download of Code Without Integrity Check](https://cwe.mitre.org/data/definitions/494.html)

The build files indicate that this project is resolving dependencies over HTTP instead of HTTPS. Any of these artifacts could have been MITM to maliciously compromise them and infect the build artifacts that were produced. Additionally, if any of these JARs or other dependencies were compromised, any developers using these could continue to be infected past updating to fix this.

This vulnerability has a CVSS v3.0 Base Score of 8.1/10
https://nvd.nist.gov/vuln-metrics/cvss/v3-calculator?vector=AV:N/AC:H/PR:N/UI:N/S:U/C:H/I:H/A:H

### This isn't just theoretical
POC code has existed since 2014 to maliciously compromise a JAR file inflight.
See:
* https://max.computer/blog/how-to-take-over-the-computer-of-any-java-or-clojure-or-scala-developer/
* https://github.com/mveytsman/dilettante

### MITM Attacks Increasingly Common
See:
* https://serverfault.com/a/153065
* https://security.stackexchange.com/a/12050
* [Comcast continues to inject its own code into websites you visit](https://thenextweb.com/insights/2017/12/11/comcast-continues-to-inject-its-own-code-into-websites-you-visit/#) (over HTTP)
* You're the NSA, you tell me...

### Source Locations

https://github.com/NationalSecurityAgency/datawave/blob/cff28531f1a101d3411b80412bcba5db6ad31741/pom.xml#L1129-L1164

JLLeitschuh:
There's still one remaining vulnerability here:

https://github.com/NationalSecurityAgency/datawave/blob/9bd04645ce24f1f62ec3272fce36f10c473fdba4/pom.xml#L1102-L1108

Should a CVE be issued after this is resolved to indicate that previous releases may have been compromised by a MITM?

brianloss:
As the commit comments, indicate, the protocol was updated where possible. The GeoWave repo isn't available via https, so it couldn't be updated.

SethSmucker:
Fixed by #400. All repo/pluginRepo URL in the poms are https now, closing.

PR #400, opened 2019-04-04
Clean-up: update dependencies, fix repo protocol

brianloss:
Update dependencies of 3rd party libs (to avoid versions with known CVEs)
Update mvn repositories to https (where possible) re #373
Update docs to newer version of jQuery/foundation.


brianloss:
@jwomeara can you take a look at this, especially for the changes to the geo-related dependencies. All unit and quickstart query tests do pass with these changes.

warehouse/pom.xml
```diff
@@ -63,17 +63,12 @@
...
                 <artifactId>kryo</artifactId>
-                <version>2.20</version>
+                <version>${version.kryo}</version>
             </dependency>
             <dependency>
                 <groupId>com.esotericsoftware.kryonet</groupId>
                 <artifactId>kryonet</artifactId>
-                <version>2.20</version>
-            </dependency>
-            <dependency>
```
brianloss: This, and the other removals below are either because the same dependencies are declared in the top-level pom's dependencyManagement section, or the dependency was never referenced in any child pom.

web-services/common/src/main/java/datawave/webservice/common/cache/SharedCacheCoordinator.java
```diff
@@ -332,19 +333,34 @@ public InterProcessLock getMutex(String path) {
...
         ArgumentChecker.notNull(counterName, listener);
-        Preconditions.checkArgument(!sharedCounters.containsKey(counterName), "Counter " + counterName + " has already been registered!");
         
         SharedCount count = new SharedCount(curatorClient, ZKPaths.makePath("/counters", counterName), seedValue);
         count.start();
         sharedCounters.put(counterName, count);
         sharedCountListeners.put(counterName, listener);
         localCounters.put(counterName, count.getCount());
         
+        count.addListener(new SharedCountListener() {
```
brianloss: I was seeing a failure in the SharedCacheCoordinator test after updating the curator version. I believe we were incorrectly using the API. When a shared object (counter, boolean, etc) loses the connection to zookeeper, the listeners must be re-registered upon re-connection.

web-services/deploy/docs/docs/index.html
```diff
@@ -5,32 +5,34 @@
...
+                    <ul class="menu vertical">
                         <li><a href="javadocs/index.html">Javadoc</a></li>
                         <li><a href="apidocs/index.html">API Docs</a></li>
                     </ul>
                 </li>
-
                 <li>
                     <a href="#">DataWave</a>
-                    <ul class="menu">
+                    <ul class="menu veritcal">
```
jschmidt10: Typo: Should be 'vertical'

brianloss:
Manually rebased and merged via e54339c08917f1a858959159d3828ed4f714e442

PR #404, opened 2019-04-10
fixes #402

### LINKED ISSUE #402: ShardQueryLogic/BaseQueryLogic inconsistent states
BaseQueryLogic uses GenericQueryConfiguration
ShardQueryLogic uses ShardQueryConfiguration
ShardQueryConfiguration extends GenericQueryConfiguration

In the ShardQueryLogic both a GenericQueryConfiguration and ShardQueryConfiguration exist, and are not synchronized.

FineAndDandy:
This normalizes all the config options to use getConfig() which is overridden on the way up through the query logics.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
+++ b/warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
@@ -167,7 +167,7 @@ public class ShardQueryLogic extends BaseQueryLogic<Entry<Key,Value>> {
     protected ScannerFactory scannerFactory = null;
     protected Scheduler scheduler = null;
     protected EventQueryDataDecoratorTransformer eventQueryDataDecoratorTransformer = null;
-    protected ShardQueryConfiguration config;
+    private ShardQueryConfiguration config;
     protected MetadataHelperFactory metadataHelperFactory = null;
     protected DateIndexHelperFactory dateIndexHelperFactory = null;
     protected Function<String,String> queryMacroFunction;
@@ -187,8 +187,6 @@ public class ShardQueryLogic extends BaseQueryLogic<Entry<Key,Value>> {
      */
     public ShardQueryLogic() {
         super();
-        // super(ShardQueryConfiguration.create());
-        // config = (ShardQueryConfiguration) super.getConfig();
         if (log.isTraceEnabled())
             log.trace("Creating ShardQueryLogic: " + System.identityHashCode(this));
     }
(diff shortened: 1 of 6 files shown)


warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -1079,8 +1079,13 @@ public void close() {
         
     }
     
+    @Override
```
FineAndDandy: @ivakegg we are

ivakegg: Ah, this file was collapsed in my view so I missed it....sorry.

web-services/query/src/main/java/datawave/webservice/query/logic/BaseQueryLogic.java
```diff
@@ -77,11 +72,13 @@ public BaseQueryLogic(BaseQueryLogic<T> other) {
         setConnPoolName(other.getConnPoolName());
         setPrincipal(other.getPrincipal());
         setRoleManager(other.getRoleManager());
-        setMarkingFunctions(other.getMarkingFunctions());
-        setResponseObjectFactory(other.getResponseObjectFactory());
         setSelectorExtractor(other.getSelectorExtractor());
     }
     
+    public GenericQueryConfiguration getConfig() {
+        return baseConfig;
```
ivakegg: Perhaps we should lazily create the baseConfig so that we at least get a NPE if somebody tries to access it directly but it is not the primary configuration.

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -167,7 +167,7 @@
     protected ScannerFactory scannerFactory = null;
     protected Scheduler scheduler = null;
     protected EventQueryDataDecoratorTransformer eventQueryDataDecoratorTransformer = null;
-    protected ShardQueryConfiguration config = ShardQueryConfiguration.create();
+    protected ShardQueryConfiguration config;
```
ivakegg: should we make this private like we did for baseConfig?

FineAndDandy: sure

web-services/query/src/main/java/datawave/webservice/query/logic/BaseQueryLogic.java
FineAndDandy: another option here is to remove baseConfig entirely, and make getConfig abstract. I think I like that better

PR #407, opened 2019-04-10
Update metadata table with normalized data type for tokenized fields.

friedlou:


warehouse/ingest-core/src/main/java/datawave/ingest/metadata/EventMetadata.java
```diff
@@ -310,15 +310,26 @@ protected void addTokenizedContent(IngestHelperInterface helper, RawRecordContai
...
+                    contentIndexed = true;
                 }
                 
                 if (h.isReverseContentIndexField(field)) {
                     updateForReverseIndexedField(helper, event, fields, countDelta, loadDate, tokenDesignator, field);
+                    contentIndexed = true;
+                }
+                
+                if (contentIndexed) {
+                    log.debug(field + tokenDesignator + " has a data type");
```
alerman: This should either be wrapped in log.isDebugEnabled, or use a parameterized log to avoid wasting time on doing the string concatenation when not logging.

friedlou: Modified in the next commit

friedlou: Applied parameterized logging for all the debug logging.

PR #408, opened 2019-04-11
Service status 405 - modify status functions to return more information

bmwmaestoso:
Updates all of the [SERVICE]Status functions to display the status for each process that should be running. The current method would only display the pids and on occasion one or more processes would not be running.


ivakegg:
Please move to release/version2.6.  Thanks.

contrib/datawave-quickstart/bin/services/accumulo/bootstrap.sh
```diff
@@ -117,7 +117,61 @@ function accumuloStop() {
...
+    test -z "${_gc}" && error "accumulo gc is not running"
+    test -z "${_master}" && error "accumulo master is not running"
+    test -z "${_monitor}" && error "accumulo monitor is not running"
+    test -z "${_tracer}" && error "accumulo tracer is not running"
+    test -z "${_tserver}" && error "accumulo tserver is not running"
+
+    unset _pid
+    local -r _zoo=$(jps -l | grep org.apache.zookeeper)
+    test -n "${_zoo}" && {
+        set -- ${_zoo}
```
brianloss: Overwriting the function args could lead to bugs down the road if someone wants to add to this function and puts something at the end assuming they can work with the args. Everything in lines 163-168 could be replaced with
```bash
_pid=$(jps -l | grep org.apache.zookeeper | cut -d' ' -f 1)
```

bmwmaestoso: There are typically several alternative solutions. This may be a style issue but I don't agree with your assessment that this could lead to bugs. I prefer to use internal bash feature whenever possible rather than creating a new process.

brianloss: > I don't agree with your assessment that this could lead to bugs

`set -- ${_zoo}` replaces the function arguments array `$@` with the value of the `_zoo` variable. If someone doesn't understand that and went to use the args later on in some addition to this function, they'd get something they didn't expect. Or worse, if they tested without zookeeper running, then everything would work as they expect (since the args wouldn't be overwritten) and you wouldn't find out until later that the change was bad. Now since it doesn't seem any args are passed, it probably doesn't matter in this case.

Personally, I find:
```bash
_pid=$(jps -l | grep org.apache.zookeeper | cut -d' ' -f 1)
```
a lot easier to read and understand vs
```bash
    unset _pid
    local -r _zoo=$(jps -l | grep org.apache.zookeeper)
    test -n "${_zoo}" && {
        set -- ${_zoo}
        _pid=$1
    }
```
at the expense of an extra pipe through `cut` (you've already created the sub-shell and two processes).

Anyway, since args aren't passed to this function, I'll retract my objection unless someone else also has an issue with this.

alerman: Personally I prefer the readability of what @brianloss posted. Also, there isn't a large enough gain to justify the difference in readability, but that's just my opinion.

keith-ratcliffe:
I've been mulling this over, and while I really do like the idea of having the extra status information, I also worry that the maintenance cost of all the extra code could outweigh the benefit in the long run.

Having maintained the quickstart code for a while now, I've already experienced the pain of breakage caused by `pgrep` misfire, due to subtle changes in process naming/identification over time...from various iterations of our own code and also from updates to our external service dependencies.

Similar breakage in the future will now inflict more pain, because internally the *IsRunning functionality still relies on the existing pgrep's. Diagnosing and fixing *Status- and *IsRunning-related bugs for a given process would require distinct debug efforts and code updates. When *IsRunning and *Status don't agree for whatever reason, at best it'll be really confusing for users

Having said all that, the existing *Status output is definitely lame on its own. I've always used it more as a quick sanity-check, and only as a supplement to my own usage of `jps` / `ps` / `pgrep` / etc, if I need to troubleshoot further.

It would be great if we could find a better middle ground, between the admittedly sad/lame *Status utility of today and the *Status utility you've created here, but which minimizes the maintenance hassle as much as humanly possible

Thoughts?

keith-ratcliffe:
One way to resolve the potential issue of `{service}IsRunning` and `{service}Status` being in disagreement would be for each *Status method to leverage the associated *IsRunning method, and then use the internally-maintained `${DW_*_PID_LIST}`, as your primary input for determining which processes to display more detailed status on and to make a determination about which services are MIA. That would address a lot of my maintenance concerns

bmwmaestoso:
Keith and I discussed consolidating the {service}Status methods to use the pids that are returned from {service}IsRunning to method. This should keep these two methods in sync.

contrib/datawave-quickstart/bin/services/accumulo/bootstrap.sh
```diff
@@ -117,7 +117,61 @@ function accumuloStop() {
...
+                address) _opt=addrValue;;
+                addrValue) _opt=pid;;
+            esac
+        done
+    }
+
+    test -z "${_gc}" && error "accumulo gc is not running"
+    test -z "${_master}" && error "accumulo master is not running"
+    test -z "${_monitor}" && error "accumulo monitor is not running"
+    test -z "${_tracer}" && error "accumulo tracer is not running"
```
ivakegg: Having a tracer running is not a requirement for a usable accumulo system.  Neither is the monitor for that matter.  I believe all you need is a master, gc, and 1 or more tservers for the system to be up and usable.
Event then, if the gc is not running or the master is not running one can use the system.  Just saying....

ivakegg: At a minimum specifying which pieces of the system are not running like you do for the datawave status below would be useful.

bmwmaestoso: will make these info instead of error

PR #415, opened 2019-04-24
generation of shard statistics

bmwmaestoso:
Initial code drop for generating cardinality and selectivity for the shard field index entries. This uses the HyperLogLogPlus class to estimate the selectivity. Upon installation of the datawave-ingest, the shell script to generate shard statistics will run successfully.

Other modifications include:

- Updates to the Hadoop settings for mapred-site.xml. Some of the settings were not valid and others had to be modified for the stats map/reduce job to run properly.
- An options was added to the BulkIngestMapFileLoader to disable generation of metrics.
- The Quickstart environment was modified to create tables, creates splits for the shard table for the current day, create splits for all of the index tables, and generate the splits file. This is performed before any of the example data is ingested.


warehouse/index-stats/src/main/java/datawave/mapreduce/shardStats/HyperLogFieldSummary.java
```diff
@@ -0,0 +1,70 @@
...
+    @Override
+    public void add(Value value) throws IOException {
+        StatsHyperLogSummary stats = new StatsHyperLogSummary(value);
+        this.count += stats.getCount();
+        HyperLogLogPlus hllpAdd = stats.getHyperLogPlus();
+        try {
+            this.logPlus.addAll(hllpAdd);
+        } catch (CardinalityMergeException e) {
+            // addAll throws an out of scope exception
+            throw new IOException();
```
jschmidt10: Can we propagate the CardinalityMergeException to preserve the full stacktrace?

warehouse/index-stats/src/main/java/datawave/mapreduce/shardStats/StatsHyperLogMapper.java
```diff
@@ -0,0 +1,274 @@
...
+        }
+    }
+    
+    @Override
+    protected void map(Key key, Value value, Context context) throws IOException, InterruptedException {
+        if (log.isTraceEnabled()) {
+            log.trace("map(" + key + ", " + value + ")");
+        }
+        
+        if (!setPartitionerContext) {
```
jschmidt10: Can this block of code be moved to setup?

warehouse/index-stats/src/main/java/datawave/mapreduce/shardStats/StatsHyperLogMapper.java
```diff
@@ -0,0 +1,274 @@
...
+            if (null == typeInfo) {
+                // add new datatype entry
+                typeInfo = new DataTypeInfo(dwKey, this.sumUniqueCounts, this.normalPrecision, this.sparsePrecision);
+                this.dataTypeMapping.put(dataType, typeInfo);
+            }
+            typeInfo.add(dwKey.getFieldValue());
+        }
+        
+        context.progress();
+        ;
```
jschmidt10: Stray semicolon

warehouse/index-stats/src/main/java/datawave/mapreduce/shardStats/StatsHyperLogMapper.java
```diff
@@ -0,0 +1,274 @@
...
+            } else {
+                if (!this.currentFieldName.equals(fieldName)) {
+                    flushFieldValues(context);
+                    this.currentFieldName = fieldName;
+                }
+            }
+            
+            // add value to proper data type
+            String dataType = dwKey.getDataType();
+            DataTypeInfo typeInfo = this.dataTypeMapping.get(dataType);
```
jschmidt10: This is purely stylistic but in Java 8+, there is a `computeIfAbsent(key, key -> // compute initial value)` method on java.util.Map that can simplify the initialization of a Map entry.

warehouse/index-stats/src/main/java/datawave/mapreduce/shardStats/StatsHyperLogMapper.java
```diff
@@ -0,0 +1,274 @@
...
+                this.outputTotal++;
+                BulkIngestKey bulkKey = new BulkIngestKey(this.outputTable, key);
+                StatsHyperLogSummary sum = entry.getStatsSummary();
+                if (log.isDebugEnabled()) {
+                    log.debug("output: key(" + bulkKey + ")");
+                    log.debug("stats(" + sum.statsString() + ")");
+                }
+                
+                Value val = new Value(sum.toByteArray());
+                context.write(bulkKey, val);
```
jschmidt10: There is a lot of object creation in here and in generateOutputKey. The `context.write()` will serialize out the key/value pairs so the objects can actually be reused. Some candidates for reuse would be: outRow, outFam, outQual, val. Depending on how large these jobs are, the object creation could add significant processing time.

warehouse/index-stats/src/main/java/datawave/mapreduce/shardStats/HyperLogFieldSummary.java
bmwmaestoso: Yes, I meant to add that as the cause.

warehouse/index-stats/src/main/java/datawave/mapreduce/shardStats/StatsHyperLogMapper.java
bmwmaestoso: Will remove.

ivakegg:
Please move this up to release/version2.6.  Thank you.

warehouse/index-stats/src/main/java/datawave/mapreduce/shardStats/StatsHyperLogMapper.java
bmwmaestoso: I will have to look into this. This was taken from the original code.

warehouse/index-stats/src/main/java/datawave/mapreduce/shardStats/StatsHyperLogMapper.java
bmwmaestoso: If the object is serialized immediately then there could be some reuse. I haven't stepped into this code but I assume what you have noted is true as the BulkContextWriter.flush() method calls the same write method.

Will investigate.

warehouse/index-stats/src/main/java/datawave/mapreduce/shardStats/StatsHyperLogReducer.java
```diff
@@ -0,0 +1,117 @@
...
+        this.totalKeys++;
+        HyperLogLogPlus hllp = new HyperLogLogPlus(this.normalPrecision, this.sparsePrecision);
+        HyperLogFieldSummary stats = new HyperLogFieldSummary(hllp);
+        int valueCount = 0;
+        for (Value val : values) {
+            stats.add(val);
+            valueCount++;
+            if (0 == (valueCount % this.valueInterval) || this.countsOnly) {
+                if (this.countsOnly) {
+                    StatsHyperLogSummary addStats = new StatsHyperLogSummary(val);
```
jschmidt10: It looks like the intent of this block is for debugging. I'd suggest we include `&& log.isDebugEnabled()` to the if condition above. No point in generating these objects if we aren't debug logging.

warehouse/ingest-scripts/src/main/resources/bin/util/buildIndexStats.sh
```diff
@@ -0,0 +1,365 @@
...
+    if [[ -n "${_GCLog}" || -n "${_GCMapperLog}" ]]; then
+        local -r _logFile="-Xloggc:${_GCLogDir}/gcReduce-$$-${_dateStr}.log"
+        local -r _gcLog="${_GCLogOpts} ${_logFile}"
+    fi
+
+    _ChildReduceOpts="\
+-Xmx${_ReducerMemoryMB}m \
+${_DefaultJVMArgs} \
+${CHILD_INGEST_OPTS} \
+${_gcLog}"
```
jschmidt10: We missed the `_HeapDumpOpts` here.

warehouse/index-stats/src/main/java/datawave/mapreduce/shardStats/StatsHyperLogReducer.java
bmwmaestoso: The variable countsOnly is a debug option that provides this functionality. It can be used to examine the data without creating an rfile. However, it does require the log level set to debug. The log level should be info, which is really the intent.

warehouse/index-stats/src/main/java/datawave/mapreduce/shardStats/StatsCounters.java
```diff
@@ -0,0 +1,135 @@
...
+        return selectivity.get();
+    }
+    
+    public Value getValue() throws IOException {
+        return new Value(toByteArray());
+    }
+    
+    public byte[] toByteArray() throws IOException {
+        try (final OutputStream baos = new ByteArrayOutputStream()) {
+            DataOutputStream dataOutput = new DataOutputStream(baos);
```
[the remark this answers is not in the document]
bmwmaestoso: Easy enough to do.

warehouse/index-stats/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+        <dependency>
+            <groupId>org.apache.accumulo</groupId>
+            <artifactId>accumulo-minicluster</artifactId>
+            <version>${version.accumulo}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.mrunit</groupId>
+            <artifactId>mrunit</artifactId>
+            <version>1.1.0</version>
```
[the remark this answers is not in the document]
bmwmaestoso: Yes, I will need to replace this as I just rebased to release/version2.6.

warehouse/index-stats/src/main/java/datawave/mapreduce/shardStats/StatsHyperLogMapper.java
bmwmaestoso: The context.write() does serialize the object. Just for fun, I wrote a simple test case to see how much additional overhead was required for the new objects versus reuse. I had to make sure that the JVM reached a steady state before these two versions would provide accurate information. I created 400000 entries for the loop and found that the overhead was roughly 2.5% (reuse ~390ms, new object ~400ms), which is not much. In the overall picture it is probably less than .01%. It is a simple change to make, so it will be done.

warehouse/ingest-scripts/src/main/resources/bin/util/buildIndexStats.sh
bmwmaestoso: will add

warehouse/index-stats/src/main/java/datawave/mapreduce/shardStats/StatsHyperLogMapper.java
jschmidt10: Thanks for looking into it!

warehouse/index-stats/src/main/java/datawave/mapreduce/shardStats/StatsHyperLogMapper.java
bmwmaestoso: Looking at the MultiTableRangePartitioner class, this should be in the setup method.

warehouse/index-stats/src/main/java/datawave/mapreduce/shardStats/StatsHyperLogMapper.java
bmwmaestoso: Yes, that is quite handy.

properties/dev.properties
```diff
@@ -32,15 +32,15 @@ MAPRED_INGEST_OPTS=-useInlineCombiner
...
+CHILD_INGEST_OPTS=
 
 BULK_CHILD_MAP_MAX_MEMORY_MB=2048
 LIVE_CHILD_MAP_MAX_MEMORY_MB=1024
 BULK_CHILD_REDUCE_MAX_MEMORY_MB=2048
 LIVE_CHILD_REDUCE_MAX_MEMORY_MB=1024
 
-BULK_INGEST_DATA_TYPES=mycsv
-LIVE_INGEST_DATA_TYPES=wikipedia
+BULK_INGEST_DATA_TYPES=shardStats
```
ivakegg: I think these were intended to stay as mycsv and wikipedia for the quickstart stuff.  Please ask Keith.

warehouse/ingest-scripts/src/main/resources/bin/util/buildIndexStats.sh
```diff
@@ -0,0 +1,365 @@
...
+    if [[ -n "${_UniqueCounts}" ]]; then
+        # memory setting may need to be adjusted based upon input data
+        # check max container size
+        _MapperMemoryMB=1792
+    else
+        _MapperMemoryMB=1024
+    fi
+
+    # allocate 1/3 memory to mapper io sort buffer but muat be less than 2048
+    ((_MapInMemSortBufferMBSize = _MapperMemoryMB / 3))
```
keith-ratcliffe: Guessing that `_MapInMemSortBufferMBSize` was intended to be `_MapInMemorySortBufferMBSize` instead(?) 

`_MapInMemorySortBufferMBSize` declared below but never used

warehouse/ingest-scripts/src/main/resources/bin/util/buildIndexStats.sh
```diff
@@ -0,0 +1,365 @@
...
+        exit 1
+    }
+    touch "${_lockFile}"
+    trap "cleanup 1" 1 2 3 15
+    trap "cleanup 0" 0
+}
+
+#==================================
+# Sets JVM parameters for Mapper
+function setMapperJVMParameters() {
```
keith-ratcliffe: Why are the memory values hardcoded to these specific values?

warehouse/ingest-scripts/src/main/resources/bin/util/buildIndexStats.sh
```diff
@@ -0,0 +1,365 @@
...
+#    SCRIPT 20180909 3 => generate data for shards 20180909_0, ... _2 (shards 0 1 2)
+#    SCRIPT 20180909_1 => generate data for shard 20180909_1
+#    SCRIPT 20180909_1,20180909_3 => generate data for shard 20180909_1 and 20180909_3
+#
+#++++++++++++++++++++++++++++++++++++++++++++++++
+# HyperLogLogPlus
+# The HyperlogLogPlus algorithm is utilized to provide an estimate of the cardinality of
+# a dataset. There are two parameters that affect the memory and precision of the
+# algorithm, a sparse and normal precision. These parameters determine how many buckets
+# are allocated to represent the data, which cooresponds to the amount of memory
```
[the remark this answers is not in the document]
bmwmaestoso: Ha! You missed the other words, signficat, chose, analyis, and deault. If I only had spell checker for emacs!

warehouse/ingest-scripts/src/main/resources/bin/util/buildIndexStats.sh
bmwmaestoso: I had fixed this before. Somehow it got back in. fixed again

warehouse/ingest-scripts/src/main/resources/bin/util/buildIndexStats.sh
bmwmaestoso: A different script was used as the basis for this script, which had hardcoded values. I guess I could use the default but I still need to set the io sort buffer size.

warehouse/ingest-scripts/src/main/resources/bin/util/buildIndexStats.sh
```diff
@@ -0,0 +1,365 @@
...
+    local _MapperMemoryMB
+    if [[ -n "${_UniqueCounts}" ]]; then
+        # memory setting may need to be adjusted based upon input data
+        # check max container size
+        _MapperMemoryMB=1792
+    else
+        _MapperMemoryMB=1024
+    fi
+
+    # allocate 1/3 memory to mapper io sort buffer but muat be less than 2048
```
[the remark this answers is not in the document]
bmwmaestoso: I like to be like George W and create my own words.
fixed

warehouse/ingest-scripts/src/main/resources/bin/util/buildIndexStats.sh
bmwmaestoso: The memory value will be allow to be a set by an environmental variable.

warehouse/ingest-scripts/src/main/resources/bin/util/buildIndexStats.sh
```diff
@@ -0,0 +1,364 @@
...
+    if [[ -n "${_NumShards}" ]]; then
+        ((_num = _NumShards / _divisor))
+    else
+        IFS=","
+        set -- ${_argDate}
+        unset IFS
+        if [[ $# -gt 1 ]]; then
+            ((_num = _cnt / _divisor))
+        else
+            ((_num = NUM_SHARDS / _divisor))
```
ivakegg: NUM_SHARDS is not necessarily the number of shards (see the multi-numshard cached mechanism).  This is probably fine for now but I may create a separate ticket to account for this.,

PR #416, opened 2019-04-24
Updates to simplify DW setup for use with authorization microservice

keith-ratcliffe:
Hey @brianloss , nothing much to see here, mainly just wanted to make sure you didn't have any objections to my changes to the authorization-mock.yml config or anything else before I pushed

In a nutshell....
- Default PKI materials for microservices and the quickstart are now the same
- And, if desired, users can deploy the quickstart for use with the auth microservice by overriding a variable in their env. E.g.,...
```
export DW_ENABLE_REMOTE_AUTH_SERVICE=true
```


contrib/datawave-quickstart/bin/query.sh
```diff
@@ -450,7 +450,10 @@ function configureUserIdentity() {
     DW_PKI_TMP_DIR="${DW_DATAWAVE_DATA_DIR}"/pki-temp
 
     DW_PKCS12_CLIENT_CERT=${DW_PKCS12_CLIENT_CERT:-"${DW_DATAWAVE_SOURCE_DIR}"/web-services/deploy/application/src/main/wildfly/overlay/standalone/configuration/certificates/testUser.p12}
-    DW_CLIENT_CERT_PASS=${DW_CLIENT_CERT_PASS:-secret}
+    DW_CLIENT_CERT_PASS=${DW_CLIENT_CERT_PASS:-ChangeIt}
+
+    DW_PKCS12_CA_CERT=${DW_PKCS12_CA_CERT:-"${DW_DATAWAVE_SOURCE_DIR}"/web-services/deploy/application/src/main/wildfly/overlay/standalone/configuration/certificates/testUser.p12}
```
brianloss: Shouldn't this be a CA cert, given the name of the variable?

keith-ratcliffe: Thanks for spotting. Removed in last commit. The existing code will already export any cacerts present in the DW_PKCS12_CLIENT_CERT keystore for use with curl. I had briefly considered that a user may have a distinct p12 keystore containing the CA certs. But alas, that's an unlikely edge case and misleading to have the variable be defaulted to testUser.p12, which doesn't contain the full cert chain.

Not to mention, I never made use of the variables here ;)

PR #421, opened 2019-04-29
QueryIterator tracks yield counts in addition to next and seek counts.

apmoriarty:


warehouse/query-core/src/main/java/datawave/query/iterator/profile/QuerySpan.java
```diff
@@ -28,6 +28,8 @@
     
     protected long seek = 0;
     
+    protected long yield = 0;
```
ivakegg: This should probably be a boolean.  Either we yielded or we didn't

warehouse/query-core/src/main/java/datawave/query/iterator/profile/QuerySpanCollector.java
```diff
@@ -9,6 +9,7 @@
 public class QuerySpanCollector {
     private AtomicLong seekCount = new AtomicLong();
     private AtomicLong nextCount = new AtomicLong();
+    private AtomicLong yieldCount = new AtomicLong();
```
ivakegg: AtomicBoolean?

warehouse/query-core/src/main/java/datawave/query/iterator/profile/QuerySpan.java
apmoriarty: Updated

warehouse/query-core/src/main/java/datawave/query/iterator/profile/QuerySpanCollector.java
apmoriarty: Changed to AtomicBoolean

warehouse/query-core/src/main/java/datawave/query/iterator/profile/QuerySpan.java
```diff
@@ -94,10 +96,22 @@ public long getSeekCount() {
...
+    public boolean getYield() {
+        if (yield) {
+            return true;
+        }
+        for (QuerySpan subSpan : sources) {
+            if (subSpan.getYield()) {
+                return true;
+            }
+        }
+        return true;
```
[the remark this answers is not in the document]
apmoriarty: Yes it should be. I'll update this shortly

ivakegg:
OK, so what you have looks good, but you are missing the part where the metrics are combined on the webserver side and subsequently put into the query metrics.  See DocumentTransformerSupport.extractMetrics() and writeQueryMetrics(metric).

warehouse/query-core/src/main/java/datawave/query/attributes/TimingMetadata.java
```diff
@@ -39,32 +40,34 @@ public long getSeekCount() {
...
     
-    public long getNextCount() {
-        Numeric numericValue = (Numeric) get(NEXT_COUNT);
+    public long getYieldCount() {
+        Numeric numericValue = (Numeric) get(YIELD_COUNT);
         if (numericValue != null) {
             return ((Number) numericValue.getData()).longValue();
         } else {
-            return 0;
+            return 0L;
```
alerman: why 0L here, and not change the above to also use L

PR #423, opened 2019-04-30
re #282 - accumulo microservice refactoring

keith-ratcliffe:


brianloss:
Please re-target this to release/version2.6.

services/spring-boot-starter-datawave-audit/src/main/java/datawave/microservice/audit/AuditClient.java
```diff
@@ -0,0 +1,208 @@
...
+                log.debug("Audit request with AuditType == {} was suppressed", AuditType.NONE);
+                return;
+            }
+        }
+        
+        //@formatter:off
+        ServiceInstance auditService = serviceProvider.getServiceInstance();
+        UriComponents uri = UriComponentsBuilder.fromUri(auditService.getUri())
+            .path(auditService.getServiceId() + requestPath)
+            .queryParams(request.getAuditParametersAsMap())
```
brianloss: The audit request is a POST request, and the params should be passed in the body rather than as query parameters.

keith-ratcliffe:
Thanks for the feedback. Updated to target release/version2.6. Will address merge conflicts and other items tomorrow, with a fresh pair of eyes ;)

pom.xml
```diff
@@ -84,7 +88,7 @@
         <version.microservice.dictionary-api>1.0</version.microservice.dictionary-api>
         <version.microservice.metadata-utils>1.3</version.microservice.metadata-utils>
         <version.microservice.metrics-reporter>1.2</version.microservice.metrics-reporter>
-        <version.microservice.type-utils>1.0</version.microservice.type-utils>
+        <version.microservice.type-utils>1.1-SNAPSHOT</version.microservice.type-utils>
```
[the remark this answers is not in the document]
keith-ratcliffe: Added temp module inclusion for now, in latest commit

services/accumulo-api/pom.xml
```diff
@@ -0,0 +1,145 @@
...
+        <groupId>gov.nsa.datawave.microservice</groupId>
+        <version>1.3</version>
+    </parent>
+    <artifactId>accumulo-api</artifactId>
+    <version>1.0-SNAPSHOT</version>
+    <properties>
+        <version.accumulo>1.9.2</version.accumulo>
+        <version.base-rest-responses>1.1</version.base-rest-responses>
+        <version.microservice.accumulo-utils>1.3</version.microservice.accumulo-utils>
+        <version.microservice.type-utils>1.1-SNAPSHOT</version.microservice.type-utils>
```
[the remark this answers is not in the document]
keith-ratcliffe: Added temp module inclusion for now, in latest commit

services/accumulo-api/pom.xml
```diff
@@ -0,0 +1,145 @@
...
+            </dependency>
+            <dependency>
+                <groupId>xerces</groupId>
+                <artifactId>xercesImpl</artifactId>
+                <version>2.11.0.SP4</version>
+            </dependency>
+            <dependency>
+                <groupId>com.fasterxml.jackson.core</groupId>
+                <artifactId>jackson-annotations</artifactId>
+                <version>2.9.0</version>
```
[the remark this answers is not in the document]
keith-ratcliffe: Removed in latest commit

services/accumulo-service/pom.xml
```diff
@@ -0,0 +1,278 @@
...
+        <version.accumulo>1.9.2</version.accumulo>
+        <version.microservice.accumulo-api>1.0-SNAPSHOT</version.microservice.accumulo-api>
+        <version.microservice.accumulo-utils>1.3</version.microservice.accumulo-utils>
+        <version.microservice.authorization-api>1.3</version.microservice.authorization-api>
+        <version.microservice.starter-datawave-audit>1.0-SNAPSHOT</version.microservice.starter-datawave-audit>
+    </properties>
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>commons-codec</groupId>
```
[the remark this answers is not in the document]
keith-ratcliffe: Fixed in latest commit

services/accumulo-service/pom.xml
```diff
@@ -0,0 +1,278 @@
...
+                <artifactId>commons-codec</artifactId>
+                <version>1.10</version>
+            </dependency>
+            <dependency>
+                <groupId>commons-lang</groupId>
```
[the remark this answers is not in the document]
keith-ratcliffe: Fixed in latest commit

services/accumulo-service/pom.xml
```diff
@@ -0,0 +1,278 @@
...
+                <artifactId>commons-lang</artifactId>
+                <version>2.6</version>
+            </dependency>
+            <dependency>
+                <groupId>gov.nsa.datawave.contrib</groupId>
+                <artifactId>datawave-in-memory-accumulo</artifactId>
```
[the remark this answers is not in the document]
keith-ratcliffe: Fixed in latest commit

services/accumulo-service/pom.xml
```diff
@@ -0,0 +1,278 @@
...
+            <artifactId>curator-client</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.curator</groupId>
+            <artifactId>curator-framework</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-auth</artifactId>
+            <version>2.6.4</version>
```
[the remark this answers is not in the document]
keith-ratcliffe: Fixed in latest commit

services/spring-boot-starter-datawave-audit/src/main/java/datawave/microservice/audit/AuditClient.java
keith-ratcliffe: Still working this one. Wasn't difficult to fix here, but it affects a bunch of tests since it changes the uri assertions in all my audit-related junits

services/accumulo-service/src/main/java/datawave/microservice/accumulo/lookup/config/LookupAuditProperties.java
```diff
@@ -0,0 +1,179 @@
...
+            // if colFam == null, we still want it to match against a colFamRegex of .*
+            if (colFam == null) {
+                colFam = "";
+            }
+            // if colQual == null, we still want it to match against a colQualRegex of .*
+            if (colQual == null) {
+                colQual = "";
+            }
+            
+            if (this.tableRegex != null && this.rowRegex != null && this.colFamRegex != null && this.colQualRegex != null) {
```
brianloss: You may need to check for empty values as well. You have a default of ".*" set for a number of these, but then a value could be set via property binding. I don't know of a way to set null (since it will turn into the string "null"), which means the empty string is the only way to override.
Not a likely case, I know, but I'd say either handle it or change the validation to not allow null/blank.

services/accumulo-service/src/main/java/datawave/microservice/accumulo/mock/MockAccumuloDataService.java
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * When the <strong>mock</strong> profile is activated, this bean will automatically ingest a tiny amount of data into an in-memory accumulo instance. The
+ * tables and data here are designed mainly for this application's unit tests, but the service may also be useful for integration testing
+ * <p>
+ * TODO: Replace with a better mock data provider (i.e., a more generic service that supports DataWave's query use cases and underlying table schema)
+ */
+@Service
+@Profile("mock")
+public class MockAccumuloDataService {
```
brianloss: I wonder if this could be a CommandLineRunner instead. Seems like you're only after the side effect of creating the bean created, and that's what CommandLineRunner is suited for.

services/accumulo-service/src/test/java/datawave/microservice/accumulo/TestHelper.java
```diff
@@ -0,0 +1,99 @@
...
+    public static ProxiedUserDetails userDetails(Collection<String> assignedRoles, Collection<String> assignedAuths) {
+        return userDetails(DEFAULT_USER_DN, assignedRoles, assignedAuths);
+    }
+    
+    public static ProxiedUserDetails userDetails(SubjectIssuerDNPair dn, Collection<String> assignedRoles, Collection<String> assignedAuths) {
+        DatawaveUser dwUser = new DatawaveUser(dn, USER, assignedAuths, assignedRoles, null, System.currentTimeMillis());
+        return new ProxiedUserDetails(Collections.singleton(dwUser), dwUser.getCreationTime());
+    }
+    
+    public static String queryString(String... params) {
```
brianloss: `String.join("&",params)` as this method. Well, without the extra "&" at the end.

services/accumulo-service/src/test/java/datawave/microservice/accumulo/lookup/LookupServiceAuditDisabledTest.java
```diff
@@ -0,0 +1,330 @@
...
+            "useAuthorizations=A,B,C,D,E,F,G,H,I",
+            "columnVisibility=foo",
+            LookupService.Parameter.CF + "=cf2",
+            LookupService.Parameter.BEGIN_ENTRY + "=7",
+            LookupService.Parameter.END_ENTRY + "=5");
+        //@formatter:on
+        
+        try {
+            doLookup(defaultUserDetails, path(testTableName + "/row2"), queryString);
+            fail("This code should never be reached");
```
brianloss: FYI, you can use an ExpectedException @Rule to cover this case. See DatawaveServerPropertiesTest for an example.

services/sample_configuration/application.yml
```diff
@@ -60,9 +60,6 @@ spring:
   cloud:
     consul:
       enabled: false
-  # Configure jackson mappers so that null properties are not included in the serialized response.
-  jackson:
-    default-property-include: non_null
```
brianloss: Why is this removed? It was there to support compatibility with prior web service responses.

keith-ratcliffe: Yikes. No idea. Lol. Inadvertent cut I guess. Adding back

brianloss: It's not critical in the sample configuration, of course. I brought it up since I wanted to make sure there won't be a conflict down the road with actual deployments. Sounds like maybe that's not the case.

web-services/accumulo/src/main/java/datawave/webservice/operations/remote/RemoteStatsService.java
```diff
@@ -0,0 +1,41 @@
...
+    private ObjectReader statsReader;
+    
+    @Override
+    @PostConstruct
+    public void init() {
+        super.init();
+        statsMapper = new ObjectMapper();
+        statsMapper.registerModule(new JaxbAnnotationModule());
+        statsMapper.setAnnotationIntrospector(new JaxbAnnotationIntrospector(TypeFactory.defaultInstance()));
+        statsReader = statsMapper.readerFor(StatsResponse.class);
```
brianloss: RemoteHttpService, which this extends via RemoteAccumuloService includes an objectMapper member that is configured to handle JAX-B annotations, Guava objects (e.g., Multimap), and also a property that configures the jackson mapper to use XmlElementWrapper names as the property name. You need the last one for complete compatibility with previous responses. Is there a reason you aren't using this objectMapper member here and in your other RemoteXXXService classes?

keith-ratcliffe: In the first iteration, I did make use of that mapper, but when the 2.8.x vs. 2.7.x Jackson incompatibilities with Wildfly 10 surfaced in my testing, my main concern was isolating RemoteHttpService from all my trial/error, particularly since I needed additional mapper configuration, e.g., to map the abstract KeyBase to DefaultKey for LookupResponse. I'll revisit.

Although, it occurs to me now that I could simplify the code quite a bit, and also protect against similar Jackson versioning issues in the future, by just requesting application/xml explicitly from the microservice and then just using simple unmarshal via built-in JAXBContext, no Jackson required. That might require additional hoop to jump thru, which is that the XML responses from the microservice won't be XMLNS-qualified, as would be required for deserialization on the Wildfly side. But that's easily solved with a custom XMLFilter to manage the SAX parse, to inject the expected namespace, which would only be a few additional lines of code. The existing StatsBean code does something similar for the Accumulo monitor's response, which is why I thought of it

Thoughts?

services/spring-boot-starter-datawave-audit/src/main/java/datawave/microservice/audit/AuditClient.java
keith-ratcliffe: Fixed in latest commit

services/accumulo-service/src/main/java/datawave/microservice/accumulo/lookup/config/LookupAuditProperties.java
keith-ratcliffe: Fixed in latest commit

services/accumulo-service/src/test/java/datawave/microservice/accumulo/TestHelper.java
keith-ratcliffe: Fixed in latest commit

services/accumulo-service/src/test/java/datawave/microservice/accumulo/lookup/LookupServiceAuditDisabledTest.java
keith-ratcliffe: Fixed in latest commit

services/sample_configuration/application.yml
keith-ratcliffe: Restored in latest commit

web-services/accumulo/src/main/java/datawave/webservice/operations/remote/RemoteStatsService.java
keith-ratcliffe: Updated in latest commit to use RemoteHttpService's objectmapper, which is now configured via CDI bean

services/accumulo-service/src/main/java/datawave/microservice/accumulo/mock/MockAccumuloDataService.java
keith-ratcliffe: Not sure I follow what you had in mind exactly. The intent is just auto-init an in-memory Accumulo with some dummy data by simply activating the mock profile, so that you can run the microservice without an actual Accumulo instance. I also inject this bean into my junits and make use of its public methods for various tests. If those use cases can be supported more cleanly with CommandLineRunner, then I'm open to it.

keith-ratcliffe: FWIW, I tried modifying the class to implement `CommandLineRunner`, moved the existing `setupMockWarehouseTables` call from the constructor to the `run` method, and verified that everything still works as intended. Did you have something different in mind?

brianloss: Sorry, missed this... Not a big deal since it's covered by the mock profile, but I was just thinking having a CommandLineRunner might be more appropriate than an actual Service annotation. I didn't realize at the time I commented, that you were injecting this in tests.

PR #424, opened 2019-04-30
Release/version2.5

rqualis-altamiracorp:
Need to use for Quickstart


brianloss:
> Not really sure that all of these changes in 2.5 should be merged into master at this point in time. Will let others weigh in though. If you are mainly interested in the quickstart-related changes, would it not suffice to `git cherry-pick` the specific commit(s) into your own fork for now(?)

We definitely do not want to merge all of this in to 2.5. In fact, we'd prefer not to make any more changes to it at all, though I suppose one could argue that quickstart changes are ok.

keith-ratcliffe:
> We definitely do not want to merge all of this in to 2.5.
Just clarifying that this is merging from release/version2.5 into master, in case that changes your view of the PR

brianloss:
> > We definitely do not want to merge all of this in to 2.5.
> 
> Just clarifying that this is merging from release/version2.5 into master, in case that changes your view of the PR

Oh, good call--I read it backwards. It seems the source is incorrect -- should be from master then, rather than release/version2.5. The fact that extra commits show up indicate that people have not been merging up the tree. I'll merge release/version2.5 -> release/version2.6 -> master.

PR #431, opened 2019-05-02
fixes #422: Created a regextrie

### LINKED ISSUE #422: Create long-or list to regex conversion
A long list of items can be turned into a regex.  This would not be more efficient than the long list ivarator unless the long list needs to be delayed until evaluation time.  In that case, compressing the data into a regex may help us out a little.

ivakegg:
- can be used as a simple trie
- will create an equivalent regex from the trie

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/util/regex/RegexTrie.java
+++ b/warehouse/query-core/src/main/java/datawave/query/util/regex/RegexTrie.java
@@ -1,0 +1,159 @@
+package datawave.query.util.regex;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * This class can be used to turn a large list of strings into a regex.
+ * This may be used to handle queries that arrive with lists so large that the term limit is reached.
+ */
+public class RegexTrie {
+    final Node root = new Node();
+    class Node {
+        final HashMap<Character, Node> next = new HashMap<>();
+        boolean terminator = false;
+    }
+
+    public RegexTrie() {
+    }
+
+    public RegexTrie(List<String> strings) {
+        addAll(strings);
+    }
+
+    /**
+     * Add a set of strings into the trie
+     * @param strings A list of strings. May contain the empty string.
+     */
+    public void addAll(List<String> strings) {
+        for (String string : strings) {
+            add(string);
+        }
+    }
+
+    /**
+     * Add a string into the trie
+     * @param string A string.  May be the empty string.
+     */
+    public void add(String string) {
+        // starting with the root node
+        Node current = root;
+        final int length = string.length();
+
+        // for each character in the string
+        for (int i = 0; i < length; i++) {
+            // get the node for this character
+            Node next = current.next.get(string.charAt(i));
+
+            // if we have a node, then simply move to it
+            if (next != null) {
+                current = next;
+            } else {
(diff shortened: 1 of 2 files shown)


warehouse/query-core/src/main/java/datawave/query/util/regex/RegexTrie.java
```diff
@@ -0,0 +1,159 @@
...
+        for (String string : strings) {
+            add(string);
+        }
+    }
+
+    /**
+     * Add a string into the trie
+     * @param string A string.  May be the empty string.
+     */
+    public void add(String string) {
```
brianloss: Just a stylistic comment, but using `Map.computeIfAbsent`, I think you could simplify the add method to:
```java
public void add(String string) {
    Node current = root;
    for (int i = 0; i < string.length(); i++) {
        current = current.next.computeIfAbsent(string.charAt(i), n -> new Node());
    }
    current.terminator = true;
}
```

warehouse/query-core/src/main/java/datawave/query/util/regex/RegexTrie.java
```diff
@@ -0,0 +1,159 @@
...
+                toRegex(entry.getValue(), regex);
+            }
+            // if we had multiple choices then end the group
+            if (multipleChoices) {
+                regex.append(')');
+            }
+        }
+    }
+
+    char[] CHARS_TO_ESCAPE = new char[] {'\\','.','[',']','{','}','(',')','<','>','*','+','-','=','!','?','^','$','|'};
```
brianloss: private static final?

warehouse/query-core/src/main/java/datawave/query/util/regex/RegexTrie.java
```diff
@@ -0,0 +1,159 @@
...
+        boolean terminator = false;
+    }
+
+    public RegexTrie() {
+    }
+
+    public RegexTrie(List<String> strings) {
+        addAll(strings);
+    }
+
```
brianloss: Not sure where else this could be used. Is it worth adding a `RegexTrie(String... strings)` constructor too?

warehouse/query-core/src/test/java/datawave/query/util/regex/RegexTrieTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.util.Arrays;
+import java.util.regex.Pattern;
+
+public class RegexTrieTest {
+    private static final Logger log = Logger.getLogger(RegexTrieTest.class);
+
+    @BeforeClass
+    public static void setUpClass() throws Exception {
+        Logger.getRootLogger().setLevel(Level.OFF);
```
brianloss: There is a lot of direct logger level manipulation here. Is there a reason this can't go in the log4j.properties in the test/resources folder for this module?

warehouse/query-core/src/main/java/datawave/query/util/regex/RegexTrie.java
ivakegg: that would be reasonable.  will do

warehouse/query-core/src/main/java/datawave/query/util/regex/RegexTrie.java
ivakegg: ok

PR #435, opened 2019-05-08
fixes #418: QueryExecutorBean.getPlan

### LINKED ISSUE #418: Need to ability to retrieve a query plan prior to index expansion
It has been requested to be able to retrieve a query plan prior to creating a query, and without expanding the query against the index.  Hence things like term normalizations and function expansions would be shown.

ivakegg:
* Created a mechanism by which one can get the plan for a query without creating it for execution
* By default fields are expanded but values are not
* Updated the ShardQueryLogic and DefaultQueryPlanner to appropriately avoid expanding fields and/or values


bmwmaestoso:
My concern is that by duplicating some of the existing query tests, there is additional technical dept that is being created. There are several existing query tests that were copied for planner tests. Would it be better to incorporate tests for the plan within the existing tests rather than duplicating the test? Appropriate tests could then test the plan as well as the query. Is the goal to have all or most query tests run the plan and the query? Since I am working on some adding unit tests, I am finding that the plan must be examined to ensure that tests are performing the correct query. If a new test is added, will it need to be added to both the plan and query unit tests.

warehouse/query-core/src/test/java/datawave/query/planner/AnyFieldQueryPlannerNoFieldExpansionTest.java
```diff
@@ -0,0 +1,337 @@
...
+    
+    @Test
+    public void testAnd() throws Exception {
+        log.info("------  testAnd  ------");
+        String cont = "europe";
+        for (final TestCities city : TestCities.values()) {
+            String cityPhrase = EQ_OP + "'" + city.name() + "'";
+            String contPhrase = EQ_OP + "'" + cont + "'";
+            String query = Constants.ANY_FIELD + cityPhrase + AND_OP + Constants.ANY_FIELD + contPhrase;
+            String anyQuery = Constants.ANY_FIELD + cityPhrase + " && " + Constants.ANY_FIELD + contPhrase;
```
bmwmaestoso: The AND_OP constant can be changed from " AND " to " && ". In the same context OR could also be changed to " || ".

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -393,7 +393,7 @@ private DateIndexHelper getDateIndexHelper(final ShardQueryConfiguration config)
         }
         
         final QueryStopwatch timers = config.getTimers();
-        
+        System.out.println(PrintingVisitor.formattedQueryString(queryTree));
```
bmwmaestoso: Should this be removed or written to the log?

warehouse/query-core/src/test/java/datawave/query/testframework/AbstractFunctionalQuery.java
```diff
@@ -414,6 +420,35 @@ protected GenericQueryConfiguration setupConfig(final String queryStr) throws Ex
...
+     *            whether to use the index for anyfield expansion
+     * @param expandValues
+     *            whether to use the index for regex/range expansion
+     * @return query configuration
+     * @throws Exception
+     *             error condition from query initialization
+     */
+    protected String getPlan(final String queryStr, boolean expandFields, boolean expandValues) throws Exception {
+        Date[] startEndDate = this.dataManager.getShardStartEndDate();
+        log.debug("  query[" + queryStr + "]  start(" + YMD_DateFormat.format(startEndDate[0]) + ")  end(" + YMD_DateFormat.format(startEndDate[1]) + ")");
```
FineAndDandy: if is debug enabled?

warehouse/query-core/src/test/java/datawave/query/testframework/AbstractFunctionalQuery.java
```diff
@@ -465,4 +500,30 @@ protected void ivaratorConfig(final int hdfsLocations, final boolean fst) throws
...
+    
+    /**
+     * assertQuery is almost the same as Assert.assertEquals except that it will allow for different orderings of the terms within an AND or and OR.
+     * 
+     * @param expected
+     *            The expected query
+     * @param query
+     *            The query being tested
+     */
+    protected void assertPlanEquals(String expected, String query) throws ParseException {
```
FineAndDandy: doesn't this belong in a test class?

web-services/query/src/main/java/datawave/webservice/query/logic/QueryLogic.java
```diff
@@ -23,6 +23,32 @@
...
+     * @param runtimeQueryAuthorizations
+     *            - authorizations that have been calculated for this query based on the caller and server.
+     * @param expandFields
+     *            - should unfielded terms be expanded
+     * @param expandValues
+     *            - should regex/ranges be expanded into discrete values
+     */
+    default String getPlan(Connector connection, Query settings, Set<Authorizations> runtimeQueryAuthorizations, boolean expandFields, boolean expandValues)
+                    throws Exception {
+        if (expandFields && expandValues) {
```
FineAndDandy: ||

warehouse/query-core/src/test/java/datawave/query/planner/AnyFieldQueryPlannerNoFieldExpansionTest.java
```diff
@@ -0,0 +1,337 @@
...
+import datawave.query.exceptions.FullTableScansDisallowedException;
+import datawave.query.testframework.AbstractFunctionalQuery;
+import datawave.query.testframework.AccumuloSetupHelper;
+import datawave.query.testframework.CitiesDataType;
+import datawave.query.testframework.CitiesDataType.CityEntry;
+import datawave.query.testframework.CitiesDataType.CityField;
+import datawave.query.testframework.DataTypeHadoopConfig;
+import datawave.query.testframework.FieldConfig;
+import datawave.query.testframework.GenericCityFields;
+import org.apache.log4j.Logger;
```
apmoriarty: SLF4J?

apmoriarty: SLF4j?

web-services/query/src/main/java/datawave/webservice/query/logic/QueryLogic.java
ivakegg: Actually that is what I intended, but in hindsight this should probably throw an unsupported operation exception instead as there really is no good default implementation of this.

warehouse/query-core/src/test/java/datawave/query/testframework/AbstractFunctionalQuery.java
bmwmaestoso: This is the base class that all test classes that use the test framework extend. It needs to be available to any test class.

warehouse/query-core/src/test/java/datawave/query/planner/AnyFieldQueryPlannerNoFieldExpansionTest.java
ivakegg: That would be cleaner....I will work that

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
ivakegg: will remove

warehouse/query-core/src/test/java/datawave/query/testframework/AbstractFunctionalQuery.java
ivakegg: ok

warehouse/query-core/src/test/java/datawave/query/planner/AnyFieldQueryPlannerNoFieldExpansionTest.java
ivakegg: I thought we were using log4j everywhere.... did I miss something?

warehouse/query-core/src/test/java/datawave/query/planner/AnyFieldQueryPlannerNoFieldExpansionTest.java
ivakegg: Actually that would break the Lucene query test.  I will make a separate constant.

ivakegg: making JEXL_AND_OP and JEXL_OR_OP

warehouse/query-core/src/test/java/datawave/query/planner/AnyFieldQueryPlannerNoFieldExpansionTest.java
ivakegg: In terms of our tests, everything appears to be using log4j instead of slf4j.  The only exception is QueryExceptionTest in the base-responses service.  leaving log4j for now

ivakegg:
I will merge this Thurs if nobody has any more comments

ivakegg:
@bmwmaestoso OK, I merged the planner tests into the query tests.  What do you think?

Issue #419, opened 2019-04-25
add additional query tests


bmwmaestoso:
There are a few additional query tests that were added to the original code base.

MaxExpansionQueryTest
- Duplicate testMaxValueOrFst, adding the following to the beginning of the query:
     (city == 'paris' || city == 'venice' || city == 'turin') &&
- add some negation test conditions
    - anyfield == x && ! (anyfield =~ regex && field == val)
    - anyfield == x && ! (anyfield !~ regex)
    - x == y && (a == b || c == d) && (e !~ regex)

PushdownQueryTest
- Duplicate testErrorFilterIncludeRegex but do not disable expansion.

PR #437, opened 2019-05-15
#419: max expansion query tests update

bmwmaestoso:
Updated max expansion query tests:
- Added new query tests for max expansion threshold.
- Discovered several existing max expansion query tests were not corrected.
- Moved several unit tests into separate classes to use different indexes.
- Added method to tests generated query plan.


ivakegg:
The parsePlan mechanism is a fairly weak mechanism for verifying that a plan is doing what you want it to do, but probably ok in these cases.

PR #448, opened 2019-05-31
Test TupleToRange

apmoriarty:
Renamed RangeFactory to RangeFactoryForTests, 
added RangeFactory for use by TupleToRange, 
added test for TupleToRange and RangeFactory.

### DIFF
--- a/common-test/src/main/java/datawave/common/test/utils/query/RangeFactoryForTests.java
+++ b/common-test/src/main/java/datawave/common/test/utils/query/RangeFactoryForTests.java
@@ -4,7 +4,11 @@ import org.apache.accumulo.core.data.Key;
 import org.apache.accumulo.core.data.PartialKey;
 import org.apache.accumulo.core.data.Range;
 
-public class RangeFactory {
+/**
+ * Utility class that generates ranges for unit tests.
+ */
+public class RangeFactoryForTests {
+    
     public static final String NULL_BYTE_STRING = "\u0000";
     public static final String MAX_UNICODE_STRING = new String(Character.toChars(Character.MAX_CODE_POINT));
     
--- a/warehouse/query-core/src/main/java/datawave/query/index/lookup/TupleToRange.java
+++ b/warehouse/query-core/src/main/java/datawave/query/index/lookup/TupleToRange.java
@@ -54,7 +54,7 @@ public class TupleToRange implements Function<Tuple2<String,IndexInfo>,Iterator<
         
         if (isDocumentRange(indexInfo)) {
             
-            return createDocumentRanges(queryNode, shard, indexInfo);
+            return createDocumentRanges(queryNode, shard, indexInfo, config.isTldQuery());
             
         } else if (isShardRange(shard)) {
             
@@ -73,7 +73,7 @@ public class TupleToRange implements Function<Tuple2<String,IndexInfo>,Iterator<
      *            - object built from matches in the index.
      * @return - true if we can build document range(s).
      */
-    public boolean isDocumentRange(IndexInfo indexInfo) {
+    public static boolean isDocumentRange(IndexInfo indexInfo) {
         return !indexInfo.uids().isEmpty();
     }
     
@@ -82,7 +82,7 @@ public class TupleToRange implements Function<Tuple2<String,IndexInfo>,Iterator<
      * @param shard
      * @return - true if the shard string is a shard range
      */
-    public boolean isShardRange(String shard) {
+    public static boolean isShardRange(String shard) {
         return shard.indexOf('_') >= 0;
     }
     
@@ -92,16 +92,17 @@ public class TupleToRange implements Function<Tuple2<String,IndexInfo>,Iterator<
      * @param queryNode
      * @param shard
      * @param indexMatches
+     * @param isTldQuery
      * @return
      */
-    public Iterator<QueryPlan> createDocumentRanges(JexlNode queryNode, String shard, IndexInfo indexMatches) {
+    public static Iterator<QueryPlan> createDocumentRanges(JexlNode queryNode, String shard, IndexInfo indexMatches, boolean isTldQuery) {
         List<QueryPlan> ranges = Lists.newArrayListWithCapacity(indexMatches.uids().size());
         
         for (IndexMatch indexMatch : indexMatches.uids()) {
             
             String docId = indexMatch.getUid();
             Range range;
-            if (config.isTldQuery()) {
+            if (isTldQuery) {
                 range = RangeFactory.createTldDocumentSpecificRange(shard, docId);
             } else {
                 range = RangeFactory.createDocumentSpecificRange(shard, docId);
@@ -125,7 +126,7 @@ public class TupleToRange implements Function<Tuple2<String,IndexInfo>,Iterator<
         return ranges.iterator();
     }
     
-    public Iterator<QueryPlan> createShardRange(JexlNode queryNode, String shard, IndexInfo indexInfo) {
+    public static Iterator<QueryPlan> createShardRange(JexlNode queryNode, String shard, IndexInfo indexInfo) {
         JexlNode myNode = queryNode;
         if (indexInfo.getNode() != null) {
             myNode = indexInfo.getNode();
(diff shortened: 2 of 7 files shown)


warehouse/query-core/src/main/java/datawave/query/index/lookup/TupleToRange.java
```diff
@@ -35,79 +35,120 @@
...
+    public boolean isDocumentRange(IndexInfo indexInfo) {
+        return !indexInfo.uids().isEmpty();
+    }
+    
+    /**
+     *
+     * @param shard
+     * @return - true if the shard string is a shard range
+     */
+    public boolean isShardRange(String shard) {
```
alerman: Arguably, we know exactly where the "_" should be, no? If we do, should we be checking that the character at that index, instead of running an indexOf?

warehouse/query-core/src/main/java/datawave/query/index/lookup/TupleToRange.java
```diff
@@ -35,79 +35,120 @@
...
+        String shard = tuple.first();
+        IndexInfo indexInfo = tuple.second();
         
         JexlNode queryNode = currentScript;
-        if (ii.getNode() != null) {
-            if (log.isTraceEnabled()) {
-                log.trace("Got it from tuple " + JexlStringBuildingVisitor.buildQuery(ii.getNode()));
-            }
-            
+        if (indexInfo.getNode() != null && log.isTraceEnabled()) {
```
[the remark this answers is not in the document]
apmoriarty: Flipped.

warehouse/query-core/src/main/java/datawave/query/index/lookup/TupleToRange.java
apmoriarty: In theory we do know, however the standard seems to be an indexOf() call.

warehouse/query-core/src/main/java/datawave/query/ranges/RangeFactory.java
```diff
@@ -0,0 +1,87 @@
...
+ * This class provides a standard way of building ranges with clear intent.
+ */
+public class RangeFactory {
+    
+    public static final String NULL_BYTE_STRING = "\u0000";
+    public static final String MAX_UNICODE_STRING = new String(Character.toChars(Character.MAX_CODE_POINT));
+    private static final String FIRST_SHARD = "_0";
+    
+    /**
+     * Builds a tld document range that can be passed to the {@link datawave.query.iterator.QueryIterator}
```
jschmidt10: Copy/paste error I think. This one does not build a TLD document range.

warehouse/query-core/src/test/java/datawave/query/index/lookup/TupleToRangeTest.java
```diff
@@ -1,87 +1,235 @@
...
-        IndexInfo info = new IndexInfo(Lists.newArrayList(new IndexMatch("a"), new IndexMatch("b"), new IndexMatch("c"), new IndexMatch("d")));
-        info.applyNode(queryNode);
-        Tuple2<String,IndexInfo> tuple = Tuples.tuple("s", info);
-        ShardQueryConfiguration config = new ShardQueryConfiguration();
+    public void testIsShardRange() {
+        String shardRange = "20190314_0";
+        String dayRange = "20190314";
+        
+        TupleToRange tupleToRange = new TupleToRange(null, null);
+        assertTrue(tupleToRange.isShardRange(shardRange));
```
jschmidt10: It reads a little strange that the isShardRange, isDocumentRange, etc. are TupleToRange methods. The implementations of those methods don't use any state from the TupleToRange instance. What if we made the test methods static methods on your RangeFactory? Like `RangeFactory.isShardRange(shardRange)`.

warehouse/query-core/src/test/java/datawave/query/index/lookup/TupleToRangeTest.java
```diff
@@ -1,87 +1,235 @@
...
+        // Assert ranges against expected ranges
+        eval(expectedRanges, ranges);
+    }
+    
+    private void eval(List<Range> expectedRanges, Iterator<QueryPlan> ranges) {
+        Iterator<Range> expectedIter = expectedRanges.iterator();
+        while (expectedIter.hasNext()) {
+            
+            Range expectedRange = expectedIter.next();
+            Range generatedRange = ranges.next().getRanges().iterator().next();
```
jschmidt10: Awesome tests! This is small so feel free to disregard. We are missing a scenario where the iterator returned from `QueryPlan.getRanges()` returns more than one result. We could consider ensuring that the iterator `hasNext()` returns false after we pull out the one range.

warehouse/query-core/src/test/java/datawave/query/index/lookup/TupleToRangeTest.java
apmoriarty: The tricky part here is these methods are specific implementations of the TupleToRange's function of transforming index information to ranges for query. The RangeFactory's sole responsibility is to build ranges, I think adding identity method like isShardRange could clutter the class.

Put another way, the TupleToRange is responsible for identifying and extracting information to build the ranges, the RangeFactory ensures those ranges are built in a standard way.

warehouse/query-core/src/main/java/datawave/query/ranges/RangeFactory.java
apmoriarty: Fixed in next commit

warehouse/query-core/src/test/java/datawave/query/index/lookup/TupleToRangeTest.java
apmoriarty: I'll add both of those cases

warehouse/query-core/src/test/java/datawave/query/index/lookup/TupleToRangeTest.java
jschmidt10: The building of a range and the testing of a range are logically tied together concepts. Consider if we changed what a shard range looks like. You'd have to modify the creation of it in RangeFactory and the testing of it in TupleToRange. I think the fact that you need to instantiate a `new TupleToRange(null, null)` to call this method is another hint that it may be in the wrong place. You could have called `new TupleToRange(someNode, someConfig)` and it'd have no impact on the result of the `isShardRange` and `isDocumentRange` function call. My first preference would be to move them to RangeFactory. If you believe they should be in TupleToRange, I would ask we at least make them static.

warehouse/query-core/src/test/java/datawave/query/index/lookup/TupleToRangeTest.java
jschmidt10: I get that. I'm saying the Iterator returned by `ranges.next().getRanges().iterator()` is never checked to ensure it has no more elements. If you captured it in a variable `Iterator<Range> generatedIter = ranges.next().getRanges().iterator()`, then you could do a final `assertFalse(generatedIter.hasNext())` if you wanted.

apmoriarty: Ah, I see what you're saying -- I'll make that assertion.

warehouse/query-core/src/test/java/datawave/query/index/lookup/TupleToRangeTest.java
apmoriarty: Let's go with the static option

common-test/src/main/java/datawave/common/test/utils/query/RangeFactoryForTests.java
```diff
@@ -4,7 +4,11 @@
 import org.apache.accumulo.core.data.PartialKey;
 import org.apache.accumulo.core.data.Range;
 
-public class RangeFactory {
+/**
+ * Utility class that generates ranges for unit tests.
+ */
+public class RangeFactoryForTests {
```
alerman: Nitpick: I kind of hate this name. Not sure I have a better one. Maybe just TestRangeFactory or somthing. Not sure why but the ForTests is really bothering me

common-test/src/main/java/datawave/common/test/utils/query/RangeFactoryForTests.java
```diff
@@ -4,7 +4,11 @@
...
+    
     public static final String NULL_BYTE_STRING = "\u0000";
```
alerman: Why does this have this still if you have added NULL_BYTE_STRING to Constants?

warehouse/query-core/src/main/java/datawave/query/ranges/RangeFactory.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.accumulo.core.data.Key;
+import org.apache.accumulo.core.data.PartialKey;
+import org.apache.accumulo.core.data.Range;
+
+/**
+ * This class provides a standard way of building ranges with clear intent.
+ */
+public class RangeFactory {
+    
+    public static final String NULL_BYTE_STRING = "\u0000";
```
alerman: Should this use Constants.NULL_BYTE_STRING?

PR #458, opened 2019-06-10
Test and small refactor for AncestorEquality and TLDEquality.

apmoriarty:


warehouse/query-core/src/main/java/datawave/query/function/AncestorEquality.java
```diff
@@ -1,21 +1,29 @@
...
- * For example, two IDs `h1.h2.h3.a.b.c.d` and `h1.h2.h3.e.f` would be considered equal by this check.
- * 
+ * A key equality implementation that compares key prefixes to determine ancestor equality.
  *
+ * For example, given a parent key "h1.h2.h3.a.b.c" a valid ancestor would be key "h1.h2.h3.a.b.c.d.e"
  */
 public class AncestorEquality implements Equality {
     
+    /**
+     * Determines if the first key an ancestor of the other key by checking for a common prefix.
```
jwomeara: Determines if the first key **is** an ancestor of the other key by checking for a common prefix.

apmoriarty: Fixed

PR #460, opened 2019-06-11
Re #376 - Initial commit for spotbugs plugin...

keith-ratcliffe:
...and fixes for "High" threshold findings. Lower priority findings to be addressed later


warehouse/common/src/main/java/datawave/common/cl/OptionBuilder.java
```diff
@@ -19,8 +19,8 @@
     /** Whether or not the next Option's argument is optional; defaults to false. */
     public boolean optionalArg;
     
-    /** The next Option's type; defaults to null. */
-    public Object type;
+    /** The next Option's type; defaults to {@link String#getClass()}. */
+    public Class<?> type = String.class;
```
ivakegg: why are we changing the default here?

warehouse/query-core/src/main/java/datawave/query/iterator/QueryInformationIterator.java
```diff
@@ -31,7 +31,7 @@
 public class QueryInformationIterator extends WrappingIterator {
     private static final Logger log = Logger.getLogger(QueryInformationIterator.class);
     
-    private static final String REPORT_ERRORS_OPT = "REPORT_ERRORS";
+    protected static final String REPORT_ERRORS_OPT = "REPORT_ERRORS";
```
ivakegg: maybe public?

web-services/common-util/src/test/java/datawave/security/util/AuthorizationsUtilTest.java
```diff
@@ -49,6 +49,15 @@ public void initialize() {
         DatawaveUser p2 = new DatawaveUser(p2dn, UserType.SERVER, Sets.newHashSet("A", "F", "G"), null, null, System.currentTimeMillis());
         
         principal = new DatawavePrincipal(Lists.newArrayList(user, p1, p2));
+        
+        Collection<Collection<String>> cbAuths = new HashSet<>();
```
ivakegg: what are we trying to do here?  these cbAuths are not used anywhere else.

keith-ratcliffe: Doh! Thanks. Leftover cruft from my testing. Will remove

warehouse/common/src/main/java/datawave/common/cl/OptionBuilder.java
keith-ratcliffe: Sorry, I couldn't recall the flagged issue....tried to revert back and reproduce the finding but couldn't reproduce it...which is when I realized this one was actually a "Medium" threshold, not "High" as advertised by this PR. So, this one slipped thru, as it was one of the first I encountered b4 raising to High. My bad

At any rate, I now see that the finding was....

```
Field only ever set to null: datawave.common.cl.OptionBuilder.type [datawave.common.cl.OptionBuilder] At OptionBuilder.java:[line 55] UWF_NULL_FIELD
```
So,  I decided the easiest and safest remedy was simply to mirror the corresponding default setup for org.apache.commons.cli.Option.type. I.e., ...
```
public class Option implements Cloneable, Serializable {
   ....
   private Class<?> type = String.class;
   ....
}
```

keith-ratcliffe:
I'll try to add the original spotbugs finding in comments here where needed. I realize it's difficult to review this without that context

warehouse/query-core/src/main/java/datawave/query/iterator/QueryInformationIterator.java
keith-ratcliffe: This was just part of the larger refactoring of the ErrorReportingIterator (subclass), which was masking protected fields here and also duplicating a lot of code needlessly.

The spotbugs findings were....
```
[ERROR] Field ErrorReportingIterator.errorKey masks field in superclass datawave.query.iterator.QueryInformationIterator [datawave.query.iterator.errors.ErrorReportingIterator] In ErrorReportingIterator.java MF_CLASS_MASKS_FIELD
[ERROR] Field ErrorReportingIterator.returned masks field in superclass datawave.query.iterator.QueryInformationIterator [datawave.query.iterator.errors.ErrorReportingIterator] In ErrorReportingIterator.java MF_CLASS_MASKS_FIELD
```
...so I also used the opportunity to get rid of ErrorReportingIterator's duplicated methods, and also to delete it's duplicate private `REPORT_ERRORS_OPT`, which is why I made this one protected here in the superclass. Figured it was better to keep the visibility limited

warehouse/query-core/src/main/java/datawave/query/cardinality/CardinalityRecord.java
```diff
@@ -41,7 +41,7 @@ public CardinalityRecord(Set<String> recordedFields, DateType dateType) {
     }
     
     public CardinalityRecord(CardinalityRecord other) {
-        synchronized (cardinalityMap) {
```
keith-ratcliffe: [ERROR] Synchronization on CardinalityRecord.cardinalityMap in futile attempt to guard it [datawave.query.cardinality.CardinalityRecord] At CardinalityRecord.java:[line 49] ML_SYNC_ON_FIELD_TO_GUARD_CHANGING_THAT_FIELD

services/pom.xml
```diff
@@ -107,6 +107,42 @@
...
+                            <goals>
+                                <goal>check</goal>
+                            </goals>
+                        </execution>
+                    </executions>
+                    <configuration>
+                        <effort>Max</effort>
+                        <threshold>High</threshold>
+                        <xmlOutput>true</xmlOutput>
+                        <excludeFilterFile>${project.basedir}/../spotbugs-exclude.xml</excludeFilterFile>
```
brianloss: Doesn't this leak parent files into the services folder? If we move the services dir to a new repo, this would stop working, for example.

warehouse/query-core/src/main/java/datawave/query/attributes/Geometry.java
```diff
@@ -102,10 +102,12 @@ public int compareTo(Geometry other) {
         if (geometry == null) {
```
keith-ratcliffe: [ERROR] Possible null pointer dereference of Geometry.geometry in datawave.query.attributes.Geometry.compareTo(Geometry) [datawave.query.attributes.Geometry, datawave.query.attributes.Geometry, datawave.query.attributes.Geometry] Dereferenced at Geometry.java:[line 108]Null value at Geometry.java:[line 102]Known null at Geometry.java:[line 103] NP_NULL_ON_SOME_PATH

services/pom.xml
brianloss: Oh, nm, I see. This is relative to the sub-dir of services. That's been flaky for us in the past. Might be nicer if the file could be pulled in from a lib or something. Or there's a plugin for determining the dir of a parent pom.

warehouse/query-core/src/main/java/datawave/query/function/FacetedGrouping.java
```diff
@@ -177,7 +177,7 @@ public FacetedGrouping(FacetedConfiguration config) {
         
         if (fixNewList.size() > config.getMaximumFacetGroupCount()) {
             // we've exceeded, so let's get a minimum adjustment factor
-            final int groupAdjustmentFactor = (int) Math.ceil(fixNewList.size() / config.getMaximumFacetGroupCount());
+            final int groupAdjustmentFactor = (int) Math.ceil((double) fixNewList.size() / config.getMaximumFacetGroupCount());
```
keith-ratcliffe: [ERROR] Integral value cast to double and then passed to Math.ceil in datawave.query.function.FacetedGrouping.adjustAttributeGrouping(Collection, Set) [datawave.query.function.FacetedGrouping] At FacetedGrouping.java:[line 180] ICAST_INT_CAST_TO_DOUBLE_PASSED_TO_CEIL

Spotbugs caught several examples of this.

warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
```diff
@@ -238,20 +238,18 @@ protected void shutdownThreads() {
...
                 if (queryStream.context() == StreamContext.VARIABLE) {
                     context = StreamContext.PRESENT;
-                } else
+                } else {
                     context = queryStream.context();
+                }
                 
                 if (log.isDebugEnabled()) {
                     log.debug("Query returned a stream with a context of " + this.context);
-                    if (queryStream != null) {
```
keith-ratcliffe: [ERROR] Nullcheck of RangeStream.queryStream at line 253 of value previously dereferenced in datawave.query.index.lookup.RangeStream.iterator() [datawave.query.index.lookup.RangeStream, datawave.query.index.lookup.RangeStream, datawave.query.index.lookup.RangeStream] At RangeStream.java:[line 239]Redundant null check at RangeStream.java:[line 253]Another occurrence at RangeStream.java:[line 246] RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE

warehouse/query-core/src/main/java/datawave/query/iterator/ResultCountingIterator.java
```diff
@@ -43,9 +44,9 @@ public boolean hasNext() {
     }
     
     private Key addKeyCount(Key key) {
-        resultCount++;
```
keith-ratcliffe: [ERROR] Increment of volatile field datawave.query.iterator.ResultCountingIterator.resultCount in datawave.query.iterator.ResultCountingIterator.addKeyCount(Key) [datawave.query.iterator.ResultCountingIterator] At ResultCountingIterator.java:[line 46] VO_VOLATILE_INCREMENT

warehouse/query-core/src/main/java/datawave/query/iterator/errors/ErrorReportingIterator.java
```diff
@@ -22,20 +22,10 @@
  */
 public class ErrorReportingIterator extends QueryInformationIterator {
     
-    private static final String REPORT_ERRORS_OPT = "REPORT_ERRORS";
```
keith-ratcliffe: See previous comments

services/type-utils/src/main/java/datawave/data/type/util/PhoneNumber.java
```diff
@@ -382,6 +382,13 @@ public boolean equals(Object o) {
         }
     }
     
+    @Override
+    public int hashCode() {
+        int result = originalPhoneNumber != null ? originalPhoneNumber.hashCode() : 0;
```
brianloss: The equals method compares the result of toNormalizedString, however we are generating a hash code based on the unnormalized value.  If two different unnormalized values normalize to the same thing, then equals would return true but we'd compute different hash codes.

brianloss: Ok, not good at reading today. I see the normalized number is added to the hash below. Still, it seems off to be that we're using different fields (both unnormalized and normalized) for the hashCode vs equals (only the normalized number is used).

warehouse/query-core/src/main/java/datawave/query/language/functions/jexl/Chain.java
```diff
@@ -137,8 +138,13 @@ public String getParams() {
             if (split2.length == 1) {
                 variableValue = variableName;
             } else {
-                String variableValueString = (String) split2[1];
+                String variableValueString = split2[1];
                 if (variableValueString.startsWith("[") && variableValueString.endsWith("]")) {
+                    // TODO: This code needs further review. Spotbugs flagged the split("|") below as
+                    // RE_POSSIBLE_UNINTENDED_PATTERN. There are no tests that exercise this path, so the expected
```
keith-ratcliffe: Not sure how to resolve this one. Anyone familiar with the Chain function usage in question? I'm just not familiar with it at all.

Or should this class be deprecated/removed(?)

warehouse/query-core/src/main/java/datawave/query/tables/edge/EdgeQueryLogic.java
```diff
@@ -257,7 +257,7 @@ protected VisitationContext normalizeJexlQuery(String query, boolean getFullNorm
         
         // These strings are going to be parsed again in the iterator but if there is a problem with
         // normalizing the query we want to fail here instead of over on the server side
-        if (!visitationContext.getNormalizedQuery().equals("")) {
```
keith-ratcliffe: [ERROR] Call to StringBuilder.equals(String) in datawave.query.tables.edge.EdgeQueryLogic.normalizeJexlQuery(String, boolean) [datawave.query.tables.edge.EdgeQueryLogic] At EdgeQueryLogic.java:[line 260] EC_UNRELATED_TYPES
[ERROR] Call to StringBuilder.equals(String) in datawave.query.tables.edge.EdgeQueryLogic.normalizeJexlQuery(String, boolean) [datawave.query.tables.edge.EdgeQueryLogic] At EdgeQueryLogic.java:[line 273] EC_UNRELATED_TYPES

warehouse/query-core/src/main/java/datawave/query/tld/TLDIndexBuildingVisitor.java
```diff
@@ -233,7 +233,7 @@ public Range getKeyRange(Entry<Key,Document> from) {
             
             @Override
             public EventDataQueryFilter clone() {
-                return this.clone();
```
keith-ratcliffe: IL_INFINITE_RECURSIVE_LOOP

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/error/ErrorShardedDataTypeHandler.java
```diff
@@ -125,7 +121,6 @@
     
     @Override
     public void setup(TaskAttemptContext context) {
-        markingFunctions = MarkingFunctionsFactory.createMarkingFunctions();
```
brianloss: Might want to investigate where this was originally added. I could be wrong, but I feel like it might have been put there for the side effects.

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -2815,8 +2814,6 @@ private void updateQuery(GenericResponse<String> response, RunningQuery runningQ
                             maxResultsOverride, parameters);
             
             // Fire off an audit prior to updating
-            Set<String> methodAuths = new HashSet<>(Arrays.asList(q.getQueryAuthorizations().split("\\s*,\\s*")));
-            cbAuths.retainAll(methodAuths);
```
keith-ratcliffe: Bug was...
```
[ERROR] java.util.HashSet<java.lang.String> is incompatible with expected argument type java.util.HashSet<java.util.Collection<java.lang.String>> .... at QueryExecutorBean.java:[line 2819] GC_UNRELATED_TYPES
```
...but I then discovered the real bug was that `cbAuths` is never used again in this method. So, this update necessitated the change to ExtendedQueryExecutorBeanTest.java

services/pom.xml
keith-ratcliffe: No this allows the child modules under `services/` dir to read `services/spotbugs-exclude.xml`

I had considered having a single shared excludes file, but opted not to for exactly the reason you mentioned. Figured it was better to keep root modules and service modules completely independent for this

keith-ratcliffe: Sorry didn't see your follow-up for some reason. Yeah, this won't work for any future modules that get nested as grandchildren, etc, etc of services. Mulling over options

services/type-utils/src/main/java/datawave/data/type/util/PhoneNumber.java
keith-ratcliffe: No, you're right, should've just been based on the normalized value.  I think I was starting to suffer from repetitive-stress injury at this point, from clicking `Code --> Generate... --> .... ` in Intellij ;)

keith-ratcliffe: The equals impl is somewhat suspect as well. It doesn't guard against NPE, which could easily occur if null was passed into constructor.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/error/ErrorShardedDataTypeHandler.java
keith-ratcliffe: Yep, `ShardedDataTypeHandler.setup` instantiates markingFunctions, but slightly differently. It uses `MarkingFunctions.Factory` rather than `MarkingFunctionsFactory`, which is compatible for webserver usage. I can revert this line. Spotbugs should still be happy b/c field masking was the root issue, not this particular assignment.

But since `ErrorShardedDataTypeHandler` is ostensibly compatible with both webserver use and with normal ingest deployment, that should mean that `MarkingFunctionsFactory` should be as well, right? In other words, what if we just changed `ShardedDataTypeHandler` to use `MarkingFunctionsFactory` instead? And leave this code as-is?

The main diff between the two `createMarkingFunctions` methods is that they load slightly different Spring config files. Seems like that could be refactored somehow, so that all the work is delegated to the `MarkingFunctions.Factory` version, as long as we make that version smart enough to overridde Spring config locations

Myeah, probably at task for a different ticket. lol

keith-ratcliffe: I'll test the both ingest and web deployment for this a little more thoroughly before I update this PR. Need to make sure the field masking isn't required to keep everything working. Kinda worried that it might be

services/type-utils/src/main/java/datawave/data/type/util/PhoneNumber.java
keith-ratcliffe: Updated both equals and hashCode in latest commit

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/error/ErrorShardedDataTypeHandler.java
keith-ratcliffe: Ingest deploy/test works fine. Was originally thinking ErrorShardedDataTypeHandler might get used or subclassed somewhere in a web service context, e.g., for QueryMetrics ingest, etc, but that does not appear to be the case. So, I think this is g2g, as-is

warehouse/ingest-core/src/main/java/datawave/ingest/config/RawRecordContainerImpl.java
```diff
@@ -437,6 +433,23 @@ public boolean equals(Object other) {
...
+        result = 31 * result + (errors != null ? errors.hashCode() : 0);
+        result = 31 * result + (visibility != null ? visibility.hashCode() : 0);
+        result = 31 * result + (rawFileName != null ? rawFileName.hashCode() : 0);
+        result = 31 * result + (int) (rawFileTimeStamp ^ (rawFileTimeStamp >>> 32));
+        result = 31 * result + (int) (rawRecordNumber ^ (rawRecordNumber >>> 32));
+        result = 31 * result + (ids != null ? ids.hashCode() : 0);
+        result = 31 * result + Arrays.hashCode(rawData);
+        result = 31 * result + (auxData != null ? auxData.hashCode() : 0);
+        result = 31 * result + (securityMarkings != null ? securityMarkings.hashCode() : 0);
+        return result;
```
brianloss: Looks like this hashCode method doesn't include auxMap whereas equals compares it.

services/pom.xml
keith-ratcliffe: According to their [javadoc](https://github.com/spotbugs/spotbugs-maven-plugin/blob/spotbugs/src/main/groovy/org/codehaus/mojo/spotbugs/SpotBugsMojo.groovy#L278) , the filter files can be resolved from classpath, URI, and file path.

So, I guess we could give each jar it's own distinct _excludes_ file, or have one jar/one file that we add to each module's classpath when the spotbugs profile is activated. Not sure I love either of those options.

Wouldn't it be less intrusive just to use the same trick that we use for _${datawave.root}_, e.g., _${services.root}_ ?

warehouse/ingest-core/src/main/java/datawave/ingest/config/RawRecordContainerImpl.java
keith-ratcliffe: Thanks. Fixed in next commit

warehouse/query-core/src/main/java/datawave/query/tld/TLDIndexBuildingVisitor.java
brianloss: Hmm, not familiar enough with this part of the code, but having clone not actually clone doesn't seem great to me.  Maybe @FineAndDandy knows?

warehouse/query-core/src/main/java/datawave/query/tld/TLDQueryIterator.java
```diff
@@ -144,7 +144,7 @@ public Range getKeyRange(Map.Entry<Key,Document> from) {
             
             @Override
             public EventDataQueryFilter clone() {
-                return this.clone();
+                return this;
```
brianloss: Same comment as above... @FineAndDandy, any input?

warehouse/query-core/src/main/java/datawave/query/tld/TLDIndexBuildingVisitor.java
keith-ratcliffe: Yeah, I assumed that cloning wasn't really happening for these objects, otherwise we'd have run into OOM at some point.

keith-ratcliffe: To your point though, perhaps throwing **UnsupportedOperationException** is the better option in these cases

ivakegg:
Lets also change this to go against master

warehouse/query-core/src/main/java/datawave/query/language/functions/jexl/Chain.java
ivakegg: Ask Bill.  I think we can remove it.

PR #466, opened 2019-06-13
Fixes #391 - ShardedTableMapFile NPE for every table

### LINKED ISSUE #391: ShardedTableMapFile NPE for every table
ShardedTableMapFile.getLocations() is repeatedly throwing NPEs until it exhausts its retries and eventually returns an incomplete map.

The line that originates the NPE is:
`tabletIds.forEach(tId -> splitToLocation.put(tId.getEndRow(), locations.getTabletLocation(tId)))`

For the final tablet in the table, there will be no endRow. This results in us putting a null key into a TreeMap. The default Comparator used by a TreeMap cannot accommodate a null key. 

One solution could be to use a comparator that orders nulls at the end:
`new TreeMap<>(Comparator.nullsLast(Comparator.naturalOrder()))`

jzgithub1:
This fix prevents TabletId's with null endrows from populating the splitlocation map.  This is effectively what the old code did but it it would would retry until the number of retries ran out in the failure of the writeSplitsFile function.   I try to filter out the tabletId with the null end row with a Java 8 lambda but I am getting an UnsupportedOperationError which leads me to think that the Java version in my IDE is not the same as what DataWave is using.  On line 360 I had to do an ugly old school iteration and testing of the tabletId for the null end row because the removif lambda doensn't work and putting the test in the tabletIds.forEach lambda makes the code unreadable.

### DIFF
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
@@ -349,7 +349,9 @@ public class ShardedTableMapFile {
                     Range range = new Range();
                     Locations locations = tableOps.locate(shardedTableName, Collections.singletonList(range));
                     List<TabletId> tabletIds = locations.groupByRange().get(range);
-                    tabletIds.forEach(tId -> splitToLocation.put(tId.getEndRow(), locations.getTabletLocation(tId)));
+                    
+                    tabletIds.stream().filter(tId -> tId.getEndRow() != null)
+                                    .forEach(tId -> splitToLocation.put(tId.getEndRow(), locations.getTabletLocation(tId)));
                 }
                 // made it here, no errors so break out
                 keepRetrying = false;


warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
```diff
@@ -349,7 +350,19 @@ public static Path createShardedMapFile(Logger log, Configuration conf, Path wor
...
+                    } catch (UnsupportedOperationException use) {
+                        log.warn(use.getMessage() + " could not filter null end rows with the removeIf lambda function.");
+                    }
+                    
+                    for (TabletId tabletId : tabletIds)
+                        if (tabletId.getEndRow() != null)
+                            splitToLocation.put(tabletId.getEndRow(), locations.getTabletLocation(tabletId));
+                    // Would like to use line 365 but it does not filter out the null endRow tabletId
+                    // and line 355 with tabletIds.removeIf(tabletId -> tabletId.getEndRow() == null) fails to do it.
+                    // tabletIds.forEach(tId -> splitToLocation.put(tId.getEndRow(), locations.getTabletLocation(tId)));
```
brianloss: Why not something like:
```
// @formatter:off
tabletIds.stream()
    .filter(tId -> tId.getEndRow() != null)
    .forEach(tId -> splitToLocation.put(tId.getEndRow(), locations.getTabletLocation(tId)));
// @formatter:on
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
```diff
@@ -349,7 +350,19 @@ public static Path createShardedMapFile(Logger log, Configuration conf, Path wor
                     Range range = new Range();
                     Locations locations = tableOps.locate(shardedTableName, Collections.singletonList(range));
                     List<TabletId> tabletIds = locations.groupByRange().get(range);
-                    tabletIds.forEach(tId -> splitToLocation.put(tId.getEndRow(), locations.getTabletLocation(tId)));
+                    try {
```
brianloss: Please remove this try block since it will just print superfluous warnings. The underlying class returns an unmodifiable map of Range to unmodifiable list of TableId. removeIf mutates the list, and since you are trying to modify an unmodifiable list, you get an UnsupportedOperationException.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
```diff
@@ -334,7 +335,7 @@ public static Path createShardedMapFile(Logger log, Configuration conf, Path wor
      */
     public static Map<Text,String> getLocations(Logger log, AccumuloHelper accumuloHelper, String shardedTableName) {
         // split (endRow) -> String location mapping
-        Map<Text,String> splitToLocation = new TreeMap<>();
+        TreeMap<Text,String> splitToLocation = new TreeMap<>(Comparator.nullsLast(Comparator.naturalOrder()));
```
brianloss: Does the order of nulls really matter if we're preventing nulls from going in the map?

jzgithub1:
TreeMap will just not allow a null key value.  Here is the de-compiled code.  The else statement beginning on line 348 of ShardedTableMapFile.java will never not throw an NPE as far as I can tell and the while loop will only quit because MAX_RETRY_ATTEMPTS get hit.

public V put(K var1, V var2) {
        TreeMap.Entry var3 = this.root;
        if (var3 == null) {
            this.compare(var1, var1);
            this.root = new TreeMap.Entry(var1, var2, (TreeMap.Entry)null);
            this.size = 1;
            ++this.modCount;
            return null;
        } else {
            Comparator var6 = this.comparator;
            int var4;
            TreeMap.Entry var5;
            if (var6 != null) {
                do {
                    var5 = var3;
                    var4 = var6.compare(var1, var3.key);
                    if (var4 < 0) {
                        var3 = var3.left;
                    } else {
                        if (var4 <= 0) {
                            return var3.setValue(var2);
                        }

                        var3 = var3.right;
                    }
                } while(var3 != null);
            } else {
                if (var1 == null) {
                    throw new NullPointerException();
                }


This is the stack trace in the master branch code .  I put in an e.printStackTrace before the line number 357:   log.warn(e.getMessage() + " ... retrying ...");

WARN  [ClientConfiguration:loadFromSearchPath] Found no client.conf in default paths. Using default client configuration values.
java.lang.NullPointerException
	at java.util.TreeMap.put(TreeMap.java:563)
	at datawave.ingest.mapreduce.job.ShardedTableMapFile.lambda$getLocations$0(ShardedTableMapFile.java:352)
	at java.util.ArrayList.forEach(ArrayList.java:1257)
	at java.util.Collections$UnmodifiableCollection.forEach(Collections.java:1080)
	at datawave.ingest.mapreduce.job.ShardedTableMapFile.getLocations(ShardedTableMapFile.java:352)
	at datawave.ingest.mapreduce.job.ShardedTableMapFile.createShardedMapFile(ShardedTableMapFile.java:308)
	at datawave.ingest.mapreduce.job.ShardedTableMapFile.loadMap(ShardedTableMapFile.java:239)
	at datawave.ingest.mapreduce.job.ShardedTableMapFile.setupFile(ShardedTableMapFile.java:60)
	at datawave.ingest.mapreduce.job.ShardedTableMapFileTest.testWriteSplitsToAccumuloAndReadThem(ShardedTableMapFileTest.java:109)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:298)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:292)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.lang.Thread.run(Thread.java:748)
WARN  [ShardedTableMapFile:getLocations] null ... retrying ...

brianloss:
> TreeMap will just not allow a null key value. Here is the de-compiled code. The else statement beginning on line 348 of ShardedTableMapFile.java will never not throw an NPE as far as I can tell and the while loop will only quit because MAX_RETRY_ATTEMPTS get hit.

I can't tell--was this in reply to some particular comment? Yes, TreeMap doesn't support null keys.  It appears this code was refactored recently and this error was introduced.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
jzgithub1: That is the way to go.  I am still getting comfortable with the stream API.  I will make the change tomorrow and make a new pull request.

jschmidt10: I won't be around an IDE for a few days so I'm not exactly sure how this function is being used. If we drop the tablet with a null end key, then we'll never return the location of the final tablet for each table. If TreeMap won't work with the overridden comparator then maybe we need a new data structure.

brianloss: @jschmidt10, have a look at the [original PR](https://github.com/NationalSecurityAgency/datawave/pull/321) that caused this bug. Previously we were using some Accumulo internal APIs to get the locations.  If you look at the [convertToTextStringMap](https://github.com/NationalSecurityAgency/datawave/pull/321/files#diff-955e637d394bf9ae192422a56c43fd37L430) function that we passed the results through, it already filtered out locations with a null end row. One could argue that was always a bug although for our sharded tables where we always expect things to be pre-split, we never use the last tablet with the null end row.

jschmidt10: Ah, fair enough. Thanks!

jzgithub1:
I pushed the change that you requested Brian.  Thanks again for your recommendation.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
brianloss: Please remove this line, as it is pointless (we don't need a comparator that orders nulls last if the comparator will never compare any nulls since none can go in the treemap).

jzgithub1:
I got the null comparator out of the TreeMap constructor @brianloss.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
```diff
@@ -33,6 +33,7 @@
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.Comparator;
```
brianloss: Please remove the unused import. Then I think we'll be all set. Thanks!

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/ShardedTableMapFile.java
```diff
@@ -349,7 +350,9 @@ public static Path createShardedMapFile(Logger log, Configuration conf, Path wor
                     Range range = new Range();
                     Locations locations = tableOps.locate(shardedTableName, Collections.singletonList(range));
                     List<TabletId> tabletIds = locations.groupByRange().get(range);
-                    tabletIds.forEach(tId -> splitToLocation.put(tId.getEndRow(), locations.getTabletLocation(tId)));
+                    
+                    tabletIds.stream().filter(tId -> tId.getEndRow() != null)
+                                    .forEach(tId -> splitToLocation.put(tId.getEndRow(), locations.getTabletLocation(tId)));
```
brianloss: Have you run a build with this file? The formatter that runs as a part of the build will likely re-format this line.

jzgithub1:
I ran this command: mvn -Pdev -Ddeploy -Dtar -DskipTests clean install  

All tests in ShardedTableMapFileTest pass.  

I think you are correct in that I must have missed the "style" goal.  I will do that.

jzgithub1:
Import statement removed.  Thanks for catching that.

brianloss:
> I ran this command: mvn -Pdev -Ddeploy -Dtar -DskipTests clean install
> 
> All tests in ShardedTableMapFileTest pass.
> 
> I think you are correct in that I must have missed the "style" goal. I will do that.

As long as you ran the build, the formatting should have happened, so this is fine then.

PR #468, opened 2019-06-17
fixes #467

### LINKED ISSUE #467: RebuildingScannerTestHelper doesn't handle some rebuilt ivarator cases
When an ivarated document is rebuilt and it's not using sorteduids its possible for multiple keys within the document to hit within the range of the ivarator, which on a rebuild will cause an assertion error since the same document will be returned again. 

Detect this case by comparing the rebuilt key/document to the last key/document. If the documents match, but the keys are different do not throw the assertion error.

FineAndDandy:


warehouse/query-core/src/test/java/datawave/query/testframework/AccumuloSetupHelper.java
```diff
@@ -90,7 +90,7 @@ public AccumuloSetupHelper(final Collection<DataTypeHadoopConfig> types, FileTyp
      */
     public Connector loadTables(final Logger parentLog) throws AccumuloException, AccumuloSecurityException, IOException, InterruptedException,
                     TableExistsException, TableNotFoundException, URISyntaxException {
-        return loadTables(parentLog, RebuildingScannerTestHelper.TEARDOWN.EVERY_OTHER);
+        return loadTables(parentLog, RebuildingScannerTestHelper.TEARDOWN.ALWAYS);
```
ivakegg: I fear this may not catch errors that occur when the iterator is not torn down.  Basically in this mode we never call next.

FineAndDandy: reverted

PR #470, opened 2019-06-18
fixes #469: Ivarator IterationInterruptedException causes query failure

### LINKED ISSUE #469: Interrupted Ivarators cause queries to fail
In QueryIterator.handleException() continue through the exception cause loop looking for an IterationInterruptedException even if a IOException or TabletClosedException is found. This exception needs to be surfaced to the Tablet to prevent query failure due to tablet migration when using an Ivarator. 

Add interrupt framework RebuildingScannerTestHelper that will surface IterationIterruptedException from the iterator stack to test this condition alongside rebuilds.

FineAndDandy:
depends on #467


warehouse/query-core/src/test/java/datawave/query/RebuildingScannerTestHelper.java
```diff
@@ -5,6 +5,9 @@
 import datawave.accumulo.inmemory.InMemoryInstance;
 import datawave.accumulo.inmemory.InMemoryScanner;
 import datawave.accumulo.inmemory.ScannerRebuilder;
+import datawave.query.attributes.Document;
+import datawave.query.function.deserializer.KryoDocumentDeserializer;
+import datawave.accumulo.inmemory.*;
```
apmoriarty: No star imports, same comment for several other files.

contrib/in-memory-accumulo/src/main/java/datawave/accumulo/inmemory/InMemoryBatchScanner.java
```diff
@@ -97,21 +97,27 @@ public void setRanges(Collection<Range> ranges) {
...
+            for (Range range : ranges) {
+                if (found) {
+                    newRanges.add(range);
+                } else if (range.contains(lastKey)) {
+                    found = true;
+                    Range newRange = new Range(lastKey, false, range.getEndKey(), range.isEndKeyInclusive());
+                    newRanges.add(newRange);
+                }
             }
+            if (!ranges.isEmpty() && newRanges.isEmpty()) {
```
ivakegg: Can ranges every actually be empty initially?

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -491,7 +491,7 @@ private void handleException(Exception e) throws IOException {
             tce = (TabletClosedException) reason;
         
         int depth = 1;
-        while (tce == null && iie == null && ioe == null && reason.getCause() != null && reason.getCause() != reason && depth < 100) {
+        while (iie == null && reason.getCause() != null && reason.getCause() != reason && depth < 100) {
```
ivakegg: I am wondering if we should even remove the iie==null check?

warehouse/query-core/src/test/java/datawave/query/RebuildingScannerTestHelper.java
alerman: Do we not have a rule set up to enforce this in the pom's? If not, we should.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
FineAndDandy: if we reach an iie we always will be re-running so there is no reason to keep looking

contrib/in-memory-accumulo/src/main/java/datawave/accumulo/inmemory/InMemoryBatchScanner.java
FineAndDandy: this is more a sanity check than anything else. In early testing due to code errors it was possible to get an empty range here

PR #472, opened 2019-06-20
Avoid read ahead for non-aggregating column families (#446)

matthpeterson:
* Avoid read ahead for non-aggregating column families

* Remove unused code

* minor test cleanup

* Adding test

* formatter changes

PR #473, opened 2019-06-20
Avoid read ahead for non-aggregating column families (#446)

matthpeterson:
* Avoid read ahead for non-aggregating column families

* Remove unused code

* minor test cleanup

* Adding test

* formatter changes


matthpeterson:
This is a cherry-pick of a commit on master.

brianloss:
Is this the same as PR #472? If so, you should withdraw this one. If we merge 472 in, then when we merge up from release/version2.5 into release/version2.6, your changes will be there.

PR #474, opened 2019-06-25
Replaced use of Junit TemporaryFolder with Files.createTempDirectory

jzgithub1:
in a few tests to prevent random test failures.  This was recommended on the Junit4 thread to fix this temporary issue.  This code change has been tested on a single machine and the code builds with:
'mvn -Pdev -Ddeploy -Dtar  clean install'

### DIFF
--- a/warehouse/query-core/src/test/java/datawave/query/tables/IndexQueryLogicTest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/tables/IndexQueryLogicTest.java
@@ -25,6 +25,7 @@ import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -60,7 +61,7 @@ public class IndexQueryLogicTest extends AbstractFunctionalQuery {
     }
     
     @Before
-    public void querySetUp() {
+    public void querySetUp() throws IOException {
         log.debug("---------  querySetUp  ---------");
         
         // Super call to pick up authSet initialization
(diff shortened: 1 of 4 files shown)


warehouse/query-core/src/test/java/datawave/query/tables/IndexQueryLogicTest.java
```diff
@@ -60,7 +61,7 @@ public static void setupClass() throws Exception {
     }
     
     @Before
-    public void querySetUp() {
+    public void querySetUp() throws IOException {
```
ivakegg: Why adding this throws?

ivakegg: Oh, nevermind.  I see.

jzgithub1:
Because Files.createTempDirectory throws and IOException

warehouse/query-core/src/test/java/datawave/query/testframework/AbstractFunctionalQuery.java
```diff
@@ -139,7 +138,9 @@ protected ShardQueryLogic createQueryLogic() {
     }
     
     @Before
-    public void querySetUp() {
+    public void querySetUp() throws IOException {
+        temporaryFolder = Files.createTempDirectory("tmp");
```
brianloss: Does this folder ever get cleaned up?

jzgithub1: Thanks for spotting that.  I need an @After function where I do that.  I did this in two other classes so it got a little confusing.

jzgithub1: It seemed get wiped out anyways.

jzgithub1: Pushed the code fix. I hit the ' temporaryFolder.toFile().deleteOnExit();' line when I put a breakpoint on it in the Intellij debugger and run IndexQueryLogicTest which uses AbstractFunctionalQuery.

PR #486, opened 2019-06-28
Fixes #477 - Index only fields in data dictionary vs data model

### LINKED ISSUE #477: Index only fields in data dictionary vs data model
The "index only" column in the datawave model is not used.  That field should be dropped altogether as well as the index only column in the model webpage endpoint.

jzgithub1:
Removed the "index only" column in the datawave model which is not used. That field was dropped as well as the index only column in the model webpage endpoint.


ivakegg:
I think you missed what the issues was requesting.  The index-only field should be dropped from the model, not the data dictionary.  It is correct in the data dictionary, but not in the model.  Also this pull request appears to include all of the stuff from another pull request dealing with the temporary folders in test cases.

jzgithub1:
I'll just close the pull request.  I did a string search for 'index only' in the code and thought I found the right thing to edit.

PR #487, opened 2019-06-28
Update the typeFieldMap to avoid repeated calls to check for values

alerman:


warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
```diff
@@ -603,6 +603,10 @@ public static Matcher compileFieldNamePattern(String fieldNamePattern) {
         
         if (types.isEmpty()) {
             types.addAll(typeFieldMap.get(null));
+        } else {
+            synchronized (typeFieldMap) {
```
keith-ratcliffe: Not sure why locking on typeFieldMap is needed here. If it is needed, then read/writes to typeFieldMap elsewhere in this class would also need to be guarded, right?

keith-ratcliffe: Also, locking on a field to guard read/write on the same field is generally not a good idea, since the lock itself corresponds to the referenced object, not to the field itself. E.g., if the field ever gets reassigned to a new/different instance, then the existing lock becomes worthless. Wouldn't be an issue here, since there's only ever one assignment and the field is private, but still probably better to lock on `this` or something that's final wrt to the current instance

brianloss: I'm not sure Keith's question re why is the locking needed in the first place has been addressed. If there is a need to lock typeFieldMap, then there are many race conditions in this class, including in the getDataTypes method that has been modified. The map is queried earlier in the method, but it is possible that between that check and where you are updating, the map could have been updated by a different thread. The local types array is then added to the map, but could it be different than one another thread put there?

Bill345: It appears it's now synchronized(this) instead of synchronized(typeFieldMap) .

brianloss: Yes, but the question of why the synchronization is needed in the first place is still outstanding. If that synchronization is truly necessary, then there are probably more serious problems in this class that need to be fixed.

keith-ratcliffe: I'm not aware of any multi-threaded contexts for IngestHelpers, but I can't rule out that possibility. These are used primarily by ColumnBaseHandlers, which are leveraged during the map task, which *should* be single-threaded, unless we're doing something weird

alerman: I think we can probably do without it. @ivakegg and I were discussing solutions and he threw this in quick without looking much at what was going on, but the more I look at it the less I think we need to synchronize here. @ivakegg concur?

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
```diff
@@ -603,6 +603,10 @@ public static Matcher compileFieldNamePattern(String fieldNamePattern) {
         
         if (types.isEmpty()) {
             types.addAll(typeFieldMap.get(null));
+        } else {
+            synchronized (this) {
+                typeFieldMap.putAll(typeFieldName, types);
```
brianloss: It's probably not a huge deal, but whenever the typeFieldMap already contains an entry for typeFieldName, we'll re-add the types list. `putAll` on a multimap iterates over the entire collection you are adding. I would probably be better to only insert into the map if we never found it there in the first place.

ivakegg: I think I would have to say that 
1) we do not need to synchronize.  If we do then we need to add synchronization in several other places
2) instead of an else clause here, simply change line 599 as follows:
     Collection<datawave.data.type.Type<?>> patternTypes = typeCompiledPatternMap.get(patternMatcher);
     types.addAll(patternTypes);
     typeFieldMap.putAll(typeFieldName, patternTypes);

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
```diff
@@ -596,7 +596,9 @@ public static Matcher compileFieldNamePattern(String fieldNamePattern) {
             for (Matcher patternMatcher : typeCompiledPatternMap.keySet()) {
                 
                 if (patternMatcher.reset(fieldName).matches()) {
-                    types.addAll(typeCompiledPatternMap.get(patternMatcher));
+                    Collection<datawave.data.type.Type<?>> patternTypes = typeCompiledPatternMap.get(patternMatcher);
+                    types.addAll(patternTypes);
+                    typeFieldMap.putAll(fieldName, patternTypes);
```
Bill345: Looks good!

PR #488, opened 2019-07-01
Fixes #477 - Dropped index only column in the model.

### LINKED ISSUE #477: Index only fields in data dictionary vs data model

jzgithub1:
Removed the requested change to the ModelBean and made all needed changes to webservice usages and ModelKeyParserTest.java

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/model/util/LoadModelFromXml.java
+++ b/warehouse/query-core/src/main/java/datawave/query/model/util/LoadModelFromXml.java
@@ -46,9 +46,6 @@ public class LoadModelFromXml {
             switch (mapping.getDirection()) {
                 case FORWARD:
                     model.addTermToModel(mapping.getModelFieldName(), mapping.getFieldName());
-                    if (mapping.getIndexOnly()) {
-                        model.addUnevaluatedField(mapping.getFieldName());
-                    }
                     break;
                 case REVERSE:
                     model.addTermToReverseModel(mapping.getFieldName(), mapping.getModelFieldName());
--- a/web-services/model/src/main/java/datawave/webservice/query/model/ModelKeyParser.java
+++ b/web-services/model/src/main/java/datawave/webservice/query/model/ModelKeyParser.java
@@ -107,6 +107,8 @@ public class ModelKeyParser {
         } else {
             m = new Mutation(mapping.getModelFieldName());
             m.putDelete(modelName + dataType, mapping.getFieldName() + NULL_BYTE + mapping.getDirection().getValue(), cv, System.currentTimeMillis());
+            m.putDelete(modelName + dataType, mapping.getFieldName() + NULL_BYTE + "index_only" + NULL_BYTE + mapping.getDirection().getValue(), cv,
+                            System.currentTimeMillis());
             return m;
         }
     }
--- a/web-services/model/src/main/java/datawave/webservice/query/model/ModelKeyParser.java
+++ b/web-services/model/src/main/java/datawave/webservice/query/model/ModelKeyParser.java
@@ -16,7 +16,6 @@ public class ModelKeyParser {
     
     public static final String NULL_BYTE = "\0";
     public static final Value NULL_VALUE = new Value(new byte[0]);
-    public static final String INDEX_ONLY = "index_only";
     private static Logger log = Logger.getLogger(ModelKeyParser.class);
     
     public static FieldMapping parseKey(Key key, Set<Authorizations> auths) {
@@ -49,8 +48,6 @@ public class ModelKeyParser {
                 modelField = row;
             }
         } else if (3 == colq.length && Direction.FORWARD == Direction.getDirection(colq[2])) {
-            // For now, index_only is in the middle
-            indexOnlyField = colq[1].trim().equalsIgnoreCase(INDEX_ONLY);
             dataField = colq[0];
             modelField = row;
             direction = Direction.FORWARD; // we already checked it in the if condition
@@ -64,7 +61,6 @@ public class ModelKeyParser {
         mapping.setDirection(direction);
         mapping.setFieldName(dataField);
         mapping.setModelFieldName(modelField);
-        mapping.setIndexOnly(indexOnlyField);
         return mapping;
     }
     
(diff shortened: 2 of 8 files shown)


jzgithub1:
@ivakegg please check this in the GUI because I don't have all the components of the GUI running right now.  Thanks.

jzgithub1:
I think I have to get indexOnly out of json_Model.html too.

warehouse/query-core/src/main/java/datawave/query/model/util/LoadModelFromXml.java
```diff
@@ -46,9 +46,7 @@ public static QueryModel loadModelFromXml(InputStream stream) throws Exception {
             switch (mapping.getDirection()) {
                 case FORWARD:
                     model.addTermToModel(mapping.getModelFieldName(), mapping.getFieldName());
-                    if (mapping.getIndexOnly()) {
-                        model.addUnevaluatedField(mapping.getFieldName());
-                    }
+                    model.addUnevaluatedField(mapping.getFieldName());
```
brianloss: This line should go away, and in addition, the concept of unevaluated (aka index-only) fields should be removed from the QueryModel class entirely. In addition (question for @ivakegg), I'm wondering if we need to write a utility to convert an old format model into a new one (e.g., to re-write the model in the metadata table)?

jzgithub1: My QueryModel.java file is read-only for some reason.  Am I barred from editing it for some reason?  I could change the perms and edit it but I just want to see that I am not barred for some reason.

web-services/model/src/main/java/datawave/webservice/query/model/ModelKeyParser.java
```diff
@@ -105,16 +102,14 @@ public static Mutation createDeleteMutation(FieldMapping mapping, String modelNa
         ColumnVisibility cv = new ColumnVisibility(mapping.getColumnVisibility());
         Mutation m;
         String dataType = StringUtils.isEmpty(mapping.getDatatype()) ? "" : NULL_BYTE + mapping.getDatatype().trim();
-        String indexOnly = mapping.getIndexOnly() ? NULL_BYTE + INDEX_ONLY : "";
+        
         if (Direction.REVERSE.equals(mapping.getDirection())) {
             m = new Mutation(mapping.getFieldName());
-            m.putDelete(modelName + dataType, mapping.getModelFieldName() + indexOnly + NULL_BYTE + mapping.getDirection().getValue(), cv,
-                            System.currentTimeMillis());
+            m.putDelete(modelName + dataType, mapping.getModelFieldName() + NULL_BYTE + mapping.getDirection().getValue(), cv, System.currentTimeMillis());
```
ivakegg: We may need to add two delete mutations here: 1 for the original model entry and 1 for the new model entry.

jzgithub1:
Bottom line up front: 
I am pretty sure we are fine with this one delete mutation after tracing the master branch code and new code while running the ModelBeanTest and ModelKeyParserTest.  The reason is because the same Mutation is created since indexOnly is false all the time which causes only empty string to be added to the column qualifier on line 111 of ModelKeyParser (in the call to mutation.putDelete) .  I am still evaluating this and should have a final comment by the end of day.  

I am tracing the function createDeleteMutation in the master branch code while running the ModelBeanTest.  If mapping.getIndexOnly() on line 108 in ModelKeyParser always evaluates to false (since it is never used as is written in the issue ticket)  then the 'indexOnly' string always evaluates to empty string.  So without indexOnly inside the field mapping the Mutation created is really the same as it always was before for legacy Models and will be in the future.  I did try to add a function called CreateDeleteMutationLegacyIndexOnly and got ModelKeyParser test to succeed.  I would call that function on line 452 in the ModelBean.java after the ModelKeyParser.createDeleteMutation call.  That way I could add both the Mutations to the batch writer inside of that loop surrounding line 452 that loops through the fields in the FieldMapping.  The ModelBeanTest.testModelDelete test would fail for no other reason than an improper number of calls to System.currentTimeMillis() because of easy mock. 

After doing the above I realized I should step back to what I have pushed already because indexOnly is always and always has been added to the delete Mutation as empty string.

jzgithub1:
I have verified that insert and delete operations on the Model bean still work via the Poster REST client.  I have added the columnVisibility="VIS1" attribute (and removed the indexOnly atttibutes) to the test data in the '<datawave_proj>/warehouse/query-core/src/test/resources/models' directory. This attribute is required for inserts and deletes to succeed.   If the indexOnly attribute is still in the XML post, then the insert or delete will still work. The indexOnly attribute is vestigial now.  Verified the additions and deletions in the accumulo shell of course.  Need to give 'root' user the VIS1 authorization. I need to put this in the documentation.

jzgithub1:
I am going to change the columnVisibility="VIS1" to columnVisibility="PUBLIC" in the test data.  Wait for the next commit please before merging.

web-services/model/src/main/java/datawave/webservice/query/model/ModelKeyParser.java
brianloss: We only need the two delete mutations (with and without index_only) in the forward section (the else block). The change in this section is fine.

warehouse/query-core/src/main/java/datawave/query/model/util/LoadModelFromXml.java
ivakegg: @brianloss A conversion utility is not a bad thought.  We are trying to be backward compatible but it would be good to save the extra bits in the metadata table.  @jzgithub1 How about create a separate ticket for creating this utility please.

jzgithub1: Will create a ticket to write a utility to convert an old format model into a new one (e.g., to re-write the model in the metadata table).  

I checked for backward compatibility in the pull request by doing an insert model with the indexOnly attribute in the xml that I posted.  The insert still worked.

web-services/model/src/main/java/datawave/webservice/query/model/ModelKeyParser.java
brianloss: > 
> 
> We only need the two delete mutations (with and without index_only) in the forward section (the else block). The change in this section is fine.

It doesn't appear this has been addressed yet.

jzgithub1:
I am on vacation now Brian.   I tested this exhaustively with Poster application doing REST call for insert and delete of models and checking out the datawave.metadata table in the Accumulo shell.  It all works for me.  I will look at it again if you want next week when I come back.

brianloss:
> 
> 
> I am on vacation now Brian. I tested this exhaustively with Poster application doing REST call for insert and delete of models and checking out the datawave.metadata table in the Accumulo shell. It all works for me. I will look at it again if you want next week when I come back.

Yes, when you get back, the specific comment about 2 delete mutations (the comment I quoted) needs to be addressed. Everything will test correctly without the change, but it won't be backwards compatible with old data that had the index only marker.

jzgithub1:
Thanks for the clarification.

jzgithub1:
Made the requested change @brianloss  and tested it.  Sorry for the delay on this.

PR #491, opened 2019-07-09
Fixes an issue where backslashes were being removed from the original query string by the Jexl parser.

jwomeara:


warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveJexlEngine.java
```diff
@@ -37,4 +38,10 @@ protected Interpreter createInterpreter(JexlContext context, boolean strictFlag,
     public ASTJexlScript parse(CharSequence expression) {
         return super.parse(expression, null, null);
     }
+    
+    @Override
+    protected ASTJexlScript parse(CharSequence expression, JexlInfo info, Scope frame) {
+        expression = expression.toString().replace("\\\\", "\\\\\\\\");
```
ivakegg: replacing 4 slashes with 8 seems weird.  I think instead we should be replacing every 2 slashes with 4.  You have been focused on the double backslash thing but don't we have the same issues with the single backslash?

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -107,20 +107,26 @@
...
+    public static ASTJexlScript parseJexlQuery(String query, boolean preserveEscapes) throws ParseException {
         // Instantiate a parser and visitor
         Parser parser = new Parser(new StringReader(";"));
         
         String caseFixQuery = query.replaceAll("\\s+[Aa][Nn][Dd]\\s+", " and ");
         caseFixQuery = caseFixQuery.replaceAll("\\s+[Oo][Rr]\\s+", " or ");
         caseFixQuery = caseFixQuery.replaceAll("\\s+[Nn][Oo][Tt]\\s+", " not ");
         
+        if (preserveEscapes) {
+            caseFixQuery = caseFixQuery.replace("\\\\", "\\\\\\\\");
```
ivakegg: replace every 2 with 4 instead?

warehouse/query-core/src/test/resources/input/generic-cities.csv
```diff
@@ -13,4 +13,5 @@
 20151010,ldn-usa-mi-10,London,Michigan,United States,North America,USA,NA,140
 20151111,par-ita-11,Paris,Lazio,Italy,Europe,ITA,Euro,30
 20151111,rom-bel-11,Rome,Hainaut,Belgium,Europe,BEL,Euro,50
-20151111,ldn-fra-lle-11,London,lle-de-France,France,Europe,FRA,Euro,20
\ No newline at end of file
+20151111,ldn-fra-lle-11,London,lle-de-France,France,Europe,FRA,Euro,20
+20151111,edge-case-id-1,\\Edge-City,Edgeville,Edgeland,Edgeope,EDG,NA,220
```
ivakegg: lets add a single backslash and perhaps a triple backslash case and verify those

jwomeara: I added the two cases you requested.  They work out fine, however I think this needs to bake a little more for other reasons...

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveJexlEngine.java
jwomeara: Check the latest commit.  I decided to move away from fixing the slashes here.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
jwomeara: Check the latest commit.  I decided to move away from fixing the slashes here.

warehouse/query-core/src/test/resources/input/generic-cities.csv
jwomeara: Added even more edge cases.  Check the latest commit.

jwomeara:
This should be good to go now.  The current implementation is definitely cleaner.

PR #500, opened 2019-07-18
eliminate visitor or recursion from filling stack

4a4f484e:


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/TreeFlatteningRebuildingVisitor.java
```diff
@@ -228,34 +235,39 @@ protected ASTReferenceExpression advanceReferenceExpression(JexlNode jexlNode) {
...
         if (!currentNode.getClass().equals(newNode.getClass())) {
             log.warn("newNode is not the same type as currentNode ... something has probably gone horribly wrong");
         }
         
-        for (int i = 0; i < currentNode.jjtGetNumChildren(); i++) {
-            JexlNode node = (JexlNode) currentNode.jjtGetChild(i).jjtAccept(this, null);
-            JexlNode dereferenced = JexlASTHelper.dereference(node);
-            
-            if (acceptableNodesToCombine(currentNode, dereferenced, !node.equals(dereferenced))) {
-                flattenTree(dereferenced, newNode, data);
```
ivakegg: By eliminating the recursion here, are we now  failing to flatten the subtrees?

4a4f484e: Yes, I added an additional test with additional nesting.

ivakegg:
> Can we avoid reordering all of the terms in the query, please?

I will have to strongly agree on this one.  reordering is going to cause much grief with many test cases as well.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/TreeFlatteningRebuildingVisitor.java
ivakegg: I bet it will fail to flatten the AND tree within the OR tree however.  This test works with the current code:

    @Test
    public void multipleNestingTest() throws Exception {
        String query = "((a && (b && (c && d))) || b || (c || d || e || (f || g || (h || i || (((j || k)))))))";
        String expected = "((a && b && c && d) || b || c || d || e || f || g || h || i || j || k)";
        JexlNode node = TreeFlatteningRebuildingVisitor.flatten(JexlASTHelper.parseJexlQuery(query));
        Assert.assertEquals(expected, JexlStringBuildingVisitor.buildQuery(node));
    }

4a4f484e: It flattens, it is only the OR order that is different. I will address the ordering.

jwomeara:
> > Can we avoid reordering all of the terms in the query, please?
> 
> I will have to strongly agree on this one. reordering is going to cause much grief with many test cases as well.

I submitted a similar PR which preserves the order if you want to check it out.  #527

brianloss:
Seems like maybe this should be targeted for release/version2.6 instead of master?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/TreeFlatteningRebuildingVisitor.java
brianloss: Seems like maybe Ivan's example above would be a good test to add.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/TreeFlatteningRebuildingVisitor.java
```diff
@@ -70,44 +72,78 @@ public Object visit(ASTOrNode node, Object data) {
         ASTOrNode orNode = JexlNodes.newInstanceOfType(node);
         orNode.image = node.image;
         
-        return flattenTree(node, orNode, data);
+        LinkedList<JexlNode> stackA = new LinkedList<>();
```
ivakegg: Just a thought, but should we do the same for AND nodes?  I realize that very large OR nodes may be more common so I could go without that.

jwomeara:
Per Brian's request, this really should be targeted for release/version2.6

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/TreeFlatteningRebuildingVisitor.java
```diff
@@ -70,44 +72,78 @@ public Object visit(ASTOrNode node, Object data) {
...
+                            stackB.push(dereferenced);
+                        } else {
+                            stackB.push(child);
+                        }
+                    }
+                } else {
+                    stackB.push(currentNode);
+                }
+            } while (!stackA.isEmpty());
+            stackA = new LinkedList<>(stackB);
```
jwomeara: I'm worried about the efficiency of this implementation.  The way this is written, you are iterating over all of your nodes (even the ones you know aren't OR nodes) repeatedly until you complete an iteration without encountering an OR node.  Would it not be better to separate out the known non-OR nodes into a separate stack as you work?

The implementation I shared with you would avoid the repeated traversals over nodes which you already know are not OR nodes, and would save us time during the query planning phase (especially since flatten tends to get called multiple times during planning).  It might be good to add a test which iterates over an OR node of 100000+ nodes to get a feel for the performance.

PR #505, opened 2019-07-19
Added an #EVALUATION_ONLY Lucene function which can be used to defer …

jwomeara:
…part of a query until the Jexl evaluation phase.


warehouse/query-core/src/test/java/datawave/query/language/parser/jexl/TestLuceneToJexlParser.java
```diff
@@ -9,71 +8,62 @@
 public class TestLuceneToJexlParser {
     
     @Test
-    public void test1() {
```
apmoriarty: This is beyond the scope of your changes and won't prevent an approval, but it would be nice to see well named tests (perhaps with comments).

Issue #502, opened 2019-07-19
BufferedFileBackedSortedSet compaction could be spread out


ivakegg:
Currently the BufferedFileBackedSortedSet will do all of the compaction when the iterator() method is called.  However we should be compacting along the way to avoid an issue with the number of files in the first place.  Also need to try an reduce the number of temporary objects during this process as this was causing garbage collection issues during the compaction phase.

PR #506, opened 2019-07-19
re #502: Updated to do ivarator compactions as the set is being built instead of all at the end.

ivakegg:


warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
```diff
@@ -156,56 +158,91 @@ public boolean containsAll(Collection<?> c) {
...
+                }
+            });
+            
+            // newSet will be the final multiset
+            MultiSetBackedSortedSet<E> newSet = new MultiSetBackedSortedSet<>();
+            
+            for (int r = 0; r < rounds; r++) {
                 
-                MultiSetBackedSortedSet<E> newSet = new MultiSetBackedSortedSet<>();
+                // create the
```
jschmidt10: Incomplete comment.

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
```diff
@@ -156,56 +158,91 @@ public boolean containsAll(Collection<?> c) {
...
+            List<SortedSet<E>> sets = new ArrayList<>(set.getSets());
+            
+            // calculate the number of sets to compact per round
+            int numSets = sets.size();
+            int excessSets = numSets - (maxFiles / 2); // those over 50% of maxFiles
+            int filesToCompact = excessSets + rounds; // Add in a few extra to account for the compacted sets being added back in
+            int setsPerCompaction = (filesToCompact + rounds - 1) / rounds; // divide into sets to compact, rounding up
+            
+            // sort the sets by size (compact up smaller sets first)
+            sets.sort(new Comparator<SortedSet<E>>() {
```
jschmidt10: Java 8 opportunity: `sets.sort(Comparator.comparing((Set<Integer> s) -> s.size()).reversed());`

jschmidt10: The comparator looks like it sorts larger sets first but then when we compact, we pull from `sets` starting at index 0. If we want the smallest set compacted first, I think you want to just sort on size without reversing it.

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
ivakegg: got it

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
ivakegg: My comparator was sorting smaller sets first.  I will use the new syntax, but without the reversed() to get the same affect.

warehouse/query-core/src/test/java/datawave/query/util/sortedset/BufferedFileBackedSortedSetTest.java
```diff
@@ -0,0 +1,233 @@
...
+            sortedOrder[i * 2 + 1] = sortedTemplate[i];
+        }
+        set = new BufferedFileBackedSortedSet<>(new ByteArrayComparator(), 5, 7, new BufferedFileBackedSortedSet.SortedSetFileHandlerFactory() {
+            @Override
+            public FileSortedSet.SortedSetFileHandler createHandler() throws IOException {
+                return new SortedSetTempFileHandler();
+            }
+        });
+        
+        // adding in the data set multiple times to create underlying files with dyuplicate values making the
```
jschmidt10: Typo `duplicate`

warehouse/query-core/src/test/java/datawave/query/util/sortedset/BufferedFileBackedSortedSetTest.java
```diff
@@ -0,0 +1,233 @@
...
+            set.add(data[i]);
+            assertFalse(set.isEmpty());
+        }
+    }
+    
+    @Test
+    public void testClear() {
+        set.clear();
+        assertTrue(set.isEmpty());
+        for (int i = 0; i < data.length; i++) {
```
jschmidt10: This part of the test duplicates the test above. I think we can drop this for loop portion.

warehouse/query-core/src/test/java/datawave/query/util/sortedset/BufferedFileBackedSortedSetTest.java
```diff
@@ -0,0 +1,233 @@
...
+            assertTrue(Arrays.equals(expected, value));
+        }
+        set.clear();
+        for (byte[] value : set) {
+            fail();
+        }
+    }
+    
+    @Test
+    public void testIteratortRemove() {
```
jschmidt10: Typo `IteratorRemove`

warehouse/query-core/src/test/java/datawave/query/util/sortedset/BufferedFileBackedSortedSetTest.java
```diff
@@ -0,0 +1,233 @@
...
+                size--;
+                assertEquals(size, set.size());
+            } catch (Exception e) {
+                // expected that some of the underlying FileSortedSets are persisted and hence the remove will fail
+                failCount++;
+                assertTrue(set.contains(value));
+                assertEquals(size, set.size());
+            }
+        }
+        assertTrue(failCount > 0);
```
jschmidt10: There is nothing that verifies we successfully removed any. The failCount could equal the initial `set.size()`. Do we want to ensure at least one remove succeeded?

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
```diff
@@ -156,56 +158,78 @@ public boolean containsAll(Collection<?> c) {
...
+            List<SortedSet<E>> sets = new ArrayList<>(set.getSets());
+            
+            // calculate the number of sets to compact per round
+            int numSets = sets.size();
+            int excessSets = numSets - (maxFiles / 2); // those over 50% of maxFiles
+            int filesToCompact = excessSets + rounds; // Add in a few extra to account for the compacted sets being added back in
+            int setsPerCompaction = (filesToCompact + rounds - 1) / rounds; // divide into sets to compact, rounding up
+            
+            // sort the sets by size (compact up smaller sets first)
+            sets.sort(Comparator.comparing((SortedSet<E> s) -> s.size()));
```
brianloss: Doesn't this work?
```java
sets.sort(Comparator.comparing(SortedSet::size));
```

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
```diff
@@ -156,56 +158,78 @@ public boolean containsAll(Collection<?> c) {
...
-                    if (log.isDebugEnabled()) {
-                        long delta = System.currentTimeMillis() - start;
-                        log.debug("Compacted " + setToCompact + " -> " + compaction + " in " + delta + "ms");
-                    }
-                    newSet.addSet(compaction);
-                    setToCompact.clear();
+                
+                // remove the sets being compacted from our list
+                // creating new array list to allow previous list to be GC'ed
+                sets = new ArrayList<>(sets.subList(setsPerCompaction, sets.size()));
```
brianloss: You're already working with a copy (made on line 182). Shouldn't you just be able to do
```java
sets = sets.subList(setsPerCompaction, sets.size());
```
and avoid the extra object creation?

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
```diff
@@ -156,56 +158,78 @@ public boolean containsAll(Collection<?> c) {
...
+                
+                // add the compacted set to our final multiset
+                newSet.addSet(compaction);
+                
+                // clear the compactions set to remove the files that were compacted
+                setToCompact.clear();
+            }
+            
+            // now add in the sets we did not compact
+            for (int i = 0; i < sets.size(); i++) {
```
brianloss: Why not?
```java
newSet.addAll(sets);
```

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
jwomeara: Or on line 201 why not just call remove until you have the number you want:
```
setToCompact.addSet(sets.remove(0));
```

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
jschmidt10: +1

This change was my suggestion and I was having trouble getting the compiler happy but I may have been using `Collection::size` instead of `SortedSet::size`. Nice!

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
ivakegg: Since part of the goal was to avoid some memory issues we were having, I was attempting to ensure that the old sets could be garbage collected as we compacted things away.  However now that I think about it, the global set is still holding on to these references so it does not matter.  I actually was using subList originally.  I will revert back to that.

ivakegg: I kinda like the remove idea.  Perhaps I will reverse the order of the list to allow a more efficient remove or changed to a LinkedList.....

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
ivakegg: because addSet() is different than add()....

warehouse/query-core/src/test/java/datawave/query/util/sortedset/BufferedFileBackedSortedSetTest.java
ivakegg: that is reasonable.....

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
```diff
@@ -156,56 +159,74 @@ public boolean containsAll(Collection<?> c) {
...
-                // number of files is under maxFiles
-                int setsPerCompaction = Math.max(2, Math.min(maxOpenFiles, Math.round((float) numSets / maxFiles)));
-                int iterationsWithExtra = numSets / setsPerCompaction;
+            // create a copy of the set list (sorting below)
+            List<SortedSet<E>> sets = new ArrayList<>(set.getSets());
+            
+            // calculate the number of sets to compact per round
+            int numSets = sets.size();
+            int excessSets = numSets - (maxFiles / 2); // those over 50% of maxFiles
+            int filesToCompact = excessSets + rounds; // Add in a few extra to account for the compacted sets being added back in
```
brianloss: What is rounds is very large? Maybe we need some safety checks here.

ivakegg: agreed... I will work that

ivakegg: Also going to remove the concept of "rounds" since it is hard coded to 1.

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
```diff
@@ -156,56 +158,68 @@ public boolean containsAll(Collection<?> c) {
...
+            if (log.isDebugEnabled()) {
+                long delta = System.currentTimeMillis() - start;
+                log.debug("Compacted " + setToCompact + " -> " + compaction + " in " + delta + "ms");
+            }
+            
+            // add the compacted set to our final multiset
+            newSet.addSet(compaction);
+            
+            // clear the compactions set to remove the files that were compacted
+            setToCompact.clear();
```
brianloss: Is there some side effect to this operation? On a regular collection, I don't think this would be necessary since it's a local stack variable and all references only through it become available for GC once this method call is complete.

ivakegg: The side effect is in the comment preceding: deleting the physical files for the old sets.

PR #509, opened 2019-07-23
Fixes #475 - Enable text wrap on the Data Dictionary page for long

### LINKED ISSUE #475: Enable text wrap on the Data Dictionary page for long field names
The field names could potentially get very long.  It is requested that the data dictionary page allows the field names to wrap.

jzgithub1:
field names.

Added attribute:   word-break: break-all  to the ```<TD>``` element style in the screen.css.    Tested in Firefox.  Works as requested in #475

### DIFF
--- a/services/spring-boot-starter-datawave/src/main/resources/static/css/screen.css
+++ b/services/spring-boot-starter-datawave/src/main/resources/static/css/screen.css
@@ -63,6 +63,7 @@ td {
 	padding-bottom: 0.15em;
 	color: #333333;
     font-size: 10pt;
+    word-wrap:break-all
 }
 
 th {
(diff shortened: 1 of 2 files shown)


services/spring-boot-starter-datawave/src/main/resources/static/css/screen.css
```diff
@@ -63,6 +63,7 @@ td {
 	padding-bottom: 0.15em;
 	color: #333333;
     font-size: 10pt;
+    word-wrap:break-all
```
brianloss: I wonder if we want to have character-level wrapping for every cell in every table. It might be better to limit this to a td with a class assigned, and then where the behavior is desired, we add the class on to the column.

ivakegg: BTW: Jeff is currently this as requested Brian.....

PR #511, opened 2019-07-24
Replace code in PushdownFunction with Accumulo API

milleruntime:
* The binRanges method was doing the work of tableOps.locate()
* Fix mixing of Accumulo table ID and table name

This is one option to migrate DW to build with Accumulo 2.0.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/scheduler/PushdownFunction.java
+++ b/warehouse/query-core/src/main/java/datawave/query/scheduler/PushdownFunction.java
@@ -12,21 +12,11 @@ import datawave.query.config.ShardQueryConfiguration;
 import datawave.query.planner.QueryPlan;
 import org.apache.accumulo.core.client.AccumuloException;
 import org.apache.accumulo.core.client.AccumuloSecurityException;
-import org.apache.accumulo.core.client.Instance;
 import org.apache.accumulo.core.client.IteratorSetting;
-import org.apache.accumulo.core.client.TableDeletedException;
 import org.apache.accumulo.core.client.TableNotFoundException;
-import org.apache.accumulo.core.client.TableOfflineException;
-import org.apache.accumulo.core.client.impl.ClientContext;
-import org.apache.accumulo.core.client.impl.Tables;
-import org.apache.accumulo.core.client.impl.TabletLocator;
-import org.apache.accumulo.core.client.security.tokens.AuthenticationToken;
-import org.apache.accumulo.core.client.security.tokens.PasswordToken;
-import org.apache.accumulo.core.conf.AccumuloConfiguration;
-import org.apache.accumulo.core.data.impl.KeyExtent;
+import org.apache.accumulo.core.client.admin.Locations;
+import org.apache.accumulo.core.data.TabletId;
 import org.apache.accumulo.core.data.Range;
-import org.apache.accumulo.core.master.state.tables.TableState;
-import org.apache.accumulo.core.client.impl.Credentials;
 import org.apache.commons.jexl2.parser.ParseException;
 import org.apache.hadoop.io.Text;
 import org.apache.log4j.Logger;
@@ -56,25 +46,18 @@ public class PushdownFunction implements Function<QueryData,List<ScannerChunk>> 
     private ShardQueryConfiguration config;
     
     /**
-     * Tablet locator
-     */
-    private TabletLocator tl;
-    
-    /**
      * Set of query plans
      */
     protected Set<Integer> queryPlanSet;
     protected Collection<IteratorSetting> customSettings;
     
-    protected String tableId = "0";
+    protected String tableName;
     
-    public PushdownFunction(TabletLocator tl, ShardQueryConfiguration config, Collection<IteratorSetting> settings, String tableId) {
-        this.tl = tl;
+    public PushdownFunction(ShardQueryConfiguration config, Collection<IteratorSetting> settings, String tableName) {
         this.config = config;
         queryPlanSet = Sets.newHashSet();
         this.customSettings = settings;
-        this.tableId = tableId;
-        
+        this.tableName = tableName;
     }
     
     public List<ScannerChunk> apply(QueryData qd) {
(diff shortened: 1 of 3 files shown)


milleruntime:
I am seeing lots of tests failures so hold off on reviewing until further notice...

milleruntime:
> I am seeing lots of tests failures so hold off on reviewing until further notice...

So the failures were due to the UnsupportedOperationException in InMemoryAccumulo.  I ported the code over in b3ea5ebdbfb2a45dc8c4cfa75d60e0d34e63be96 but I don't think supporting InMemoryAccumulo is sustainable.  I am thinking we need to replace it with [MiniAccumuloCluster](https://static.javadoc.io/org.apache.accumulo/accumulo-minicluster/1.9.3/org/apache/accumulo/minicluster/package-summary.html).

brianloss:
> ... I don't think supporting InMemoryAccumulo is sustainable. I am thinking we need to replace it with [MiniAccumuloCluster](https://static.javadoc.io/org.apache.accumulo/accumulo-minicluster/1.9.3/org/apache/accumulo/minicluster/package-summary.html).

We use InMemoryAccumulo... in memory. It is used for table caches in the web service. Dropping support for it isn't really an option for us at this time.

warehouse/query-core/src/main/java/datawave/query/scheduler/PushdownScheduler.java
```diff
@@ -75,9 +75,9 @@
     protected List<Function<IteratorSetting,IteratorSetting>> customizedFunctionList;
     
     /**
-     * Local instance of the table ID
+     * Local instance of the table name
      */
-    protected String tableId;
+    protected String tableName;
```
brianloss: It looks like this member variable isn't being used--could be replaced with a local variable.

warehouse/query-core/src/main/java/datawave/query/scheduler/PushdownScheduler.java
```diff
@@ -108,40 +108,23 @@ public void addSetting(IteratorSetting customSetting) {
...
-        
         Instance instance = config.getConnector().getInstance();
         if (instance instanceof InMemoryInstance) {
-            tl = new InMemoryTabletLocator();
-            tableId = config.getTableName();
-        } else {
-            tableId = Tables.getTableId(instance, tableName);
-            Credentials credentials = new Credentials(config.getConnector().whoami(), new PasswordToken(config.getAccumuloPassword()));
-            tl = TabletLocator.getLocator(new ClientContext(instance, credentials, AccumuloConfiguration.getDefaultConfiguration()), tableId);
+            tableName = config.getTableName();
```
brianloss: I think this block can go away. As far as I can tell, `config.getTableName()` and `config.getShardTableName()` will return the same thing, so the set on line 129/119 should take care of it.

milleruntime:
> > ... I don't think supporting InMemoryAccumulo is sustainable. I am thinking we need to replace it with [MiniAccumuloCluster](https://static.javadoc.io/org.apache.accumulo/accumulo-minicluster/1.9.3/org/apache/accumulo/minicluster/package-summary.html).
> 
> We use InMemoryAccumulo... in memory. It is used for table caches in the web service. Dropping support for it isn't really an option for us at this time.

I was thinking pull out the parts of InMemoryAccumulo that are used for cacheing and then use MiniAccumulo for everything else (like Tests, where you just need to simulate an Accumulo instance).  We should just cache the table information, I am not sure how this could be done but something more sustainable than trying to replicate every Accumulo Operation.

milleruntime:
Pushed some updates from feedback but I'm seeing failure in QueryExecutorBeanTest.  Not sure if its related to the InMemoryAccumulo changes or the the query changes.

milleruntime:
> Pushed some updates from feedback but I'm seeing failure in QueryExecutorBeanTest. Not sure if its related to the InMemoryAccumulo changes or the the query changes.

Nevermind, all tests are passing now.  I just needed to mvn install contrib/in-memory-accumulo/

milleruntime:
I am not sure of the performance implications of this change but I put it up as an option for upgrading to Accumulo 2.0.  Another option would be to continue using the internal types but this would require some refactoring as well, since a lot of the types used in this class have changed or have been dropped.

milleruntime:
I squashed these commits to make it easier to cherry pick these changes to my 2.0 branch.  I may just close this PR in favor of pushing them to a 2.0 branch.

brianloss:
You'll need to update the version in the contrib/in-memory-accumulo pom, and then change the top-level pom to start using that new version.

warehouse/query-core/src/main/java/datawave/query/scheduler/PushdownFunction.java
```diff
@@ -167,58 +144,24 @@ protected void redistributeQueries(Multimap<String,QueryPlan> serverPlan, Tablet
         
     }
     
-    protected Map<String,Map<KeyExtent,List<Range>>> binRanges(TabletLocator tl, Instance instance, List<Range> ranges) throws AccumuloException,
-                    AccumuloSecurityException, TableNotFoundException {
-        Map<String,Map<KeyExtent,List<Range>>> binnedRanges = new HashMap<>();
-        
-        int lastFailureSize = Integer.MAX_VALUE;
-        
-        while (true) {
```
ivakegg: I think we potentially lost some functionality here.  The retry until we have a stable state of tablet locations was critical because this is working outside of a tserver, and hence rfiles are continually moving.  We wanted some sort of certainty that we have a stable state before launching the scans.  If the tableOperations().locate(...) does the same thing underneath then I am ok with this.  Otherwise I would prefer seeing if we could do this in a similar manner with the 2.0 API (e.g. persistent client connection).

milleruntime: I agree.  I will port over the retry functionality and look further into the behavior of the tableOps.locate() method. 

I still think this class is a good opportunity to migrate critical code to Accumulo API, ahead of a full upgrade to 2.0.

milleruntime: Sorry I forgot I had already looked into this...

Just from looking at the implementation of ```locate()``` I believe it behaves the same.  Meaning, it will keep trying to find locations until there are no more failures.  Take a look at [this loop](https://github.com/apache/accumulo/blob/2ca12d9b2bcfe3c199ce355648bf7cd2a6ccf55d/core/src/main/java/org/apache/accumulo/core/clientImpl/TableOperationsImpl.java#L1753).

milleruntime:
@ivakegg @brianloss @hlgp The functionality of the code I replaced appears to behave the same.  I think this would be a good change to merge, deploy and probably break some things, prior to a full 2.0 upgrade.

milleruntime:
I rebased onto the latest master and re-ran tests.  I saw an intermittent failure with MaxExpansionRegexQueryTest but ran it again and it passed.

PR #515, opened 2019-07-29
Added test for the CaseSensitivityVisitor.

apmoriarty:


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/CaseSensitivityVisitor.java
```diff
@@ -15,9 +15,8 @@
 
 /**
  * Upper case all identifier nodes as our field names are always upper case
- * 
- * 
- * 
+ * <p>
+ * Example: (foo == 'bar') -> (FOO == 'bar')
```
brianloss: Have you done a build with `-Ddist` enabled? My guess is this won't pass. You probably need:
```
Example: {@code (foo == 'bar') -> (FOO == 'bar')}
```

apmoriarty: It definitely fails the -Ddist build. Fixed in latest commit.

PR #517, opened 2019-07-29
Move config out of execution to allow config to apply when calling so…

alerman:
…rtpom:sort


brianloss:
Can you re-target to the release/version2.6 branch? The pom version has been released from there more recently (this is an example of why we probably need to move services to its own repo).

alerman:
Closing this in favor of using the release/version2.6 branch

PR #518, opened 2019-07-29
Updated the TypedValue sizeInBytes calculation to more accurately com…

jwomeara:
…pute the value size


jwomeara:
what do you think @FineAndDandy ?

PR #522, opened 2019-08-01
Added test for FieldMissingFromSchemaVisitorTest.

apmoriarty:


warehouse/query-core/src/test/java/datawave/query/jexl/visitors/FieldMissingFromSchemaVisitorTest.java
```diff
@@ -0,0 +1,327 @@
...
+        // Setup MockMetadataHelper
+        MockMetadataHelper helper = new MockMetadataHelper();
+        helper.addFieldsToDatatypes(fieldsToDatatypes);
+        
+        // Setup datatype filter
+        Set<String> dataTypeFilter = Sets.newHashSet("datatype1", "datatype2");
+        
+        // Case 1 - check with a datatype filter
+        Set<String> actual = FieldMissingFromSchemaVisitor.getNonExistentFields(helper, script, dataTypeFilter, specialFields);
+        Set<String> expected = Sets.newHashSet("FOO3", "FOO4");
```
jschmidt10: Since this `expected` is the same for all 3 of your cases, I would bump it up above the Case 1 comment to highlight that it's not tied to just Case 1.

apmoriarty: Fixed in latest commit.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/FieldMissingFromSchemaVisitorTest.java
```diff
@@ -0,0 +1,328 @@
...
+        actual = FieldMissingFromSchemaVisitor.getNonExistentFields(helper, script, null, specialFields);
+        assertEquals(Collections.emptySet(), actual);
+        
+        // Case 3 - check with an empty datatype filter, implies all fields.
+        actual = FieldMissingFromSchemaVisitor.getNonExistentFields(helper, script, Collections.emptySet(), specialFields);
+        assertEquals(Collections.emptySet(), actual);
+    }
+    
+    /**
+     * Test query with two field:datatype pairs, none of which does not exist in the metadata helper.
```
jwomeara: None of which does _not_ exist?  So none of them do not exist in the metadata helper?  Or, both of which do not exist in the metadata helper?

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/FieldMissingFromSchemaVisitorTest.java
```diff
@@ -0,0 +1,328 @@
...
+        MockMetadataHelper helper = new MockMetadataHelper();
+        helper.addFieldsToDatatypes(fieldsToDatatypes);
+        
+        // Setup datatype filter
+        Set<String> dataTypeFilter = Sets.newHashSet("datatypeX");
+        
+        // Case 1 - check with a datatype filter
+        Set<String> actual = FieldMissingFromSchemaVisitor.getNonExistentFields(helper, script, dataTypeFilter, specialFields);
+        Set<String> expected = Sets.newHashSet("FOO");
+        assertEquals(expected, actual);
```
jwomeara: Shouldn't the dataTypeFilter being set to 'datatypeX' have prevented FOO of type 'datatype1' from being returned?  It looks like the EvaluationPhaseFilterFunctionsDescriptor ignored the dataTypeFilter that's passed in, so fields of all types can be returned.  Is this desired, or did you uncover a bug?

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/FieldMissingFromSchemaVisitorTest.java
apmoriarty: Should be "none of which exist". Fixed in next commit.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/FieldMissingFromSchemaVisitorTest.java
apmoriarty: This looks like a bug.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/FieldMissingFromSchemaVisitorTest.java
```diff
@@ -0,0 +1,328 @@
...
+        String query = "FOO3 == 'bar' && FOO4 == 'bar'";
+        ASTJexlScript script = JexlASTHelper.parseJexlQuery(query);
+        
+        // Fields to datatypes for our MetadataHelper
+        Multimap<String,String> fieldsToDatatypes = HashMultimap.create();
+        fieldsToDatatypes.put("FOO", "datatype1");
+        fieldsToDatatypes.put("FOO2", "datatype2");
+        
+        // Setup MockMetadataHelper
+        MockMetadataHelper helper = new MockMetadataHelper();
```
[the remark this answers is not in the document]
apmoriarty: Added in next commit.

PR #524, opened 2019-08-01
fixes #480 - Show whether field is tokenized on DataDictionary page.

### LINKED ISSUE #480: Data Dictionary doc page should show whether a field is tokenized
The data dictionary web page should show whether a field is tokenized ('tf' cf in the DatawaveMetadata table).

friedlou:


services/dictionary-api/src/main/java/datawave/webservice/query/result/metadata/DefaultMetadataField.java
```diff
@@ -51,9 +52,13 @@
     
     @XmlAttribute
     private Boolean normalized = false;
+    
     @XmlAttribute
     private Boolean reverseIndexed = false;
     
+    @XmlAttribute
+    private Boolean tokenized = false;
```
brianloss: You need to add support for this field in the protostuff serialization code below (see code starting at line 171/184).

friedlou: Added in the latest commit.

PR #529, opened 2019-08-08
Include ModelBean in API documentation

### LINKED ISSUE #498: ModelBean REST API does not appear in the DataWave/doc/apidocs/index.html
The usage for insert, list, get, delete for the ModelBean is not on the https://localhost:8443/DataWave/doc/apidocs/index.html#resources  web page.

The rest API need to be added like it is for all the other Java Beans.

lbschanno:
Add datawave-ws-model as a dependency for datawave-docs so that the
source files (including ModelBean) will be read and included by the
enunciate plugin.

Fixes #498

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -132,6 +132,12 @@
         </dependency>
         <dependency>
             <groupId>gov.nsa.datawave.webservices</groupId>
+            <artifactId>datawave-ws-model</artifactId>
+            <version>${project.version}</version>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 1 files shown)


brianloss:
I know this is approved, but maybe it would be better to rebase this on release/version2.6 rather than master.

lbschanno:
Is there any reason it shouldn't be in both?

brianloss:
Looks like Ivan already merged so it’s moot now but the benefit of going to 2.6 is we’d then just merge the change up to master. To go the other way we have to cherry pick the commit back and then there will be a duplicate commit (commit, not code) on master. So the first way keeps the history cleaner.

PR #530, opened 2019-08-09
Fixes #481 - Add Search to Model Names page

### LINKED ISSUE #481: Add search to the model pages
The query model page needs a search box just like we have in the data dictionary page.

jzgithub1:
Added JQuery support for searching models by title in the Model List page.

### DIFF
--- a/web-services/client/src/main/java/datawave/webservice/model/Model.java
+++ b/web-services/client/src/main/java/datawave/webservice/model/Model.java
@@ -1,8 +1,10 @@
 package datawave.webservice.model;
 
 import java.io.Serializable;
+import java.text.MessageFormat;
 import java.util.TreeSet;
 
+import javax.inject.Inject;
 import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
 import javax.xml.bind.annotation.XmlAttribute;
@@ -22,7 +24,23 @@ import org.apache.commons.lang.builder.ToStringBuilder;
 public class Model extends BaseResponse implements Serializable, HtmlProvider {
     
     private static final long serialVersionUID = 1L;
+    private String jqueryUri;
+    private String dataTablesUri;
     private static final String TITLE = "Model Description", EMPTY = "";
+    private static final String DATA_TABLES_TEMPLATE = "<script type=''text/javascript'' src=''{0}''></script>\n"
+                    + "<script type=''text/javascript'' src=''{1}''></script>\n"
+                    + "<script type=''text/javascript''>\n"
+                    + "$(document).ready(function() '{' $(''#myTable'').dataTable('{'\"bPaginate\": false, \"aaSorting\": [[3, \"asc\"]], \"bStateSave\": true'}') '}')\n"
+                    + "</script>\n";
+    
+    public Model(String jqueryUri, String datatablesUri) {
+        this.jqueryUri = jqueryUri;
+        this.dataTablesUri = datatablesUri;
+        
+    }
+    
+    // Only used in ModelBeanTest now
+    public Model() {};
     
     @XmlAttribute(name = "name", required = true)
     private String name = null;
@@ -96,7 +114,7 @@ public class Model extends BaseResponse implements Serializable, HtmlProvider {
      */
     @Override
     public String getHeadContent() {
-        return EMPTY;
+        return MessageFormat.format(DATA_TABLES_TEMPLATE, jqueryUri, dataTablesUri);
     }
     
     /*
@@ -108,30 +126,27 @@ public class Model extends BaseResponse implements Serializable, HtmlProvider {
     public String getMainContent() {
         StringBuilder builder = new StringBuilder();
         
-        builder.append("<table>\n");
+        builder.append("<div>\n");
+        builder.append("<div id=\"myTable_wrapper\" class=\"dataTables_wrapper no-footer\">\n");
+        builder.append("<table id=\"myTable\" class=\"dataTable no-footer\" role=\"grid\" aria-describedby=\"myTable_info\">\n");
         
-        builder.append("<tr><th>Visibility</th><th>FieldName</th><th>DataType</th><th>ModelFieldName</th><th>Direction</th><th>IndexOnly</th></tr>");
+        builder.append("<thead><tr><th>Visibility</th><th>FieldName</th><th>DataType</th><th>ModelFieldName</th><th>Direction</th></tr></thead>");
+        builder.append("<tbody>");
         
-        int x = 0;
         for (FieldMapping f : this.getFields()) {
-            // highlight alternating rows
-            if (x % 2 == 0) {
-                builder.append("<tr class=\"highlight\">");
-            } else {
-                builder.append("<tr>");
-            }
-            x++;
-            
             builder.append("<td>").append(f.getColumnVisibility()).append("</td>");
             builder.append("<td>").append(f.getFieldName()).append("</td>");
             builder.append("<td>").append(f.getDatatype()).append("</td>");
             builder.append("<td>").append(f.getModelFieldName()).append("</td>");
             builder.append("<td>").append(f.getDirection()).append("</td>");
-            builder.append("<td>").append(f.getIndexOnly()).append("</td>");
             builder.append("</tr>");
         }
         
-        builder.append("</table>\n");
+        builder.append("</tbody>");
+        builder.append("  </table>\n");
+        builder.append("  <div class=\"dataTables_info\" id=\"myTable_info\" role=\"status\" aria-live=\"polite\"></div>\n");
+        builder.append("</div>\n");
+        builder.append("</div>");
         
         return builder.toString();
     }
--- a/web-services/client/src/main/java/datawave/webservice/model/ModelList.java
+++ b/web-services/client/src/main/java/datawave/webservice/model/ModelList.java
@@ -1,8 +1,10 @@
 package datawave.webservice.model;
 
 import java.io.Serializable;
+import java.text.MessageFormat;
 import java.util.HashSet;
 
+import javax.inject.Inject;
 import javax.xml.bind.annotation.XmlElement;
 import javax.xml.bind.annotation.XmlElementWrapper;
 import javax.xml.bind.annotation.XmlRootElement;
@@ -13,8 +15,23 @@ import datawave.webservice.result.BaseResponse;
 @XmlRootElement(name = "ModelList")
 public class ModelList extends BaseResponse implements Serializable, HtmlProvider {
     
+    private String jqueryUri;
+    private String dataTablesUri;
+    private String modelTableName;
+    
     private static final long serialVersionUID = 1L;
-    private static final String TITLE = "Model Names", EMPTY = "";
+    private static final String TITLE = "Model Names";
+    private static final String DATA_TABLES_TEMPLATE = "<script type=''text/javascript'' src=''{0}''></script>\n"
+                    + "<script type=''text/javascript'' src=''{1}''></script>\n"
+                    + "<script type=''text/javascript''>\n"
+                    + "$(document).ready(function() '{' $(''#myTable'').dataTable('{'\"bPaginate\": false, \"aaSorting\": [[0, \"asc\"]], \"bStateSave\": true'}') '}')\n"
+                    + "</script>\n";
+    
+    public ModelList(String jqueryUri, String datatablesUri, String dataWaveMetadateTableName) {
+        this.jqueryUri = jqueryUri;
+        this.dataTablesUri = datatablesUri;
+        this.modelTableName = dataWaveMetadateTableName;
+    }
     
     @XmlElementWrapper(name = "ModelNames")
     @XmlElement(name = "ModelName")
@@ -55,7 +72,7 @@ public class ModelList extends BaseResponse implements Serializable, HtmlProvide
      */
     @Override
     public String getHeadContent() {
-        return EMPTY;
+        return MessageFormat.format(DATA_TABLES_TEMPLATE, jqueryUri, dataTablesUri);
     }
     
     /*
@@ -70,22 +87,24 @@ public class ModelList extends BaseResponse implements Serializable, HtmlProvide
         if (this.getNames() == null || this.getNames().isEmpty()) {
             builder.append("No models available.");
         } else {
-            builder.append("<table>\n");
+            builder.append("<div>\n");
+            builder.append("<div id=\"myTable_wrapper\" class=\"dataTables_wrapper no-footer\">\n");
+            builder.append("<table id=\"myTable\" class=\"dataTable no-footer\" role=\"grid\" aria-describedby=\"myTable_info\">\n");
+            builder.append("<thead><tr><th>Model Name</th></tr></thead>");
+            builder.append("<tbody>");
             
-            int x = 0;
             for (String name : this.getNames()) {
                 // highlight alternating rows
-                if (x % 2 == 0) {
-                    builder.append("<tr class=\"highlight\">");
-                } else {
-                    builder.append("<tr>");
-                }
-                x++;
-                builder.append("<td><a href=\"/DataWave/Model/").append(name).append("\">").append(name).append("</a></td>");
-                builder.append("</tr>");
+                builder.append("<tr>");
+                builder.append("<td><a href=\"/DataWave/Model/").append(name).append("?modelTableName=").append(modelTableName);
+                builder.append("\">").append(name).append("</a></td>");
+                builder.append("</tr>\n");
             }
-            
-            builder.append("</table>\n");
+            builder.append("</tbody>");
+            builder.append("  </table>\n");
+            builder.append("  <div class=\"dataTables_info\" id=\"myTable_info\" role=\"status\" aria-live=\"polite\"></div>\n");
+            builder.append("</div>\n");
+            builder.append("</div>");
         }
         return builder.toString();
     }
--- a/web-services/client/src/main/java/datawave/webservice/model/ModelList.java
+++ b/web-services/client/src/main/java/datawave/webservice/model/ModelList.java
@@ -27,16 +27,16 @@ public class ModelList extends BaseResponse implements Serializable, HtmlProvide
                     + "$(document).ready(function() '{' $(''#myTable'').dataTable('{'\"bPaginate\": false, \"aaSorting\": [[0, \"asc\"]], \"bStateSave\": true'}') '}')\n"
                     + "</script>\n";
     
-    public ModelList(String jqueryUri, String datatablesUri, String dataWaveMetadateTableName) {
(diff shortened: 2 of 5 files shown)


jzgithub1:
To the tester.  If you are testing this in the Quickstart you must add the metadata table name that the Quickstart uses as a query string parameter to the Model list URL:

https://localhost:8443/DataWave/Model/list?modelTableName=datawave.metadata

ivakegg:
Search on the model names is great (might as well keep that), but what we really need is a search capability on the model contents page.

brianloss:
I think this should go into release/version2.6 rather than master--you'll need to rebase your branch onto release/version2.6 and then update the PR target...

web-services/client/src/main/java/datawave/webservice/model/Model.java
```diff
@@ -96,7 +102,8 @@ public String getPageHeader() {
      */
     @Override
     public String getHeadContent() {
-        return EMPTY;
+        String dataTablesHeader = MessageFormat.format(DATA_TABLES_TEMPLATE, "/", "/");
```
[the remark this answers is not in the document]
brianloss: Yeah, maybe an `@ConfigProperty` variable passed to the constructor of this class--something like that.

web-services/client/src/main/java/datawave/webservice/model/ModelList.java
```diff
@@ -55,7 +61,8 @@ public String getPageHeader() {
      */
     @Override
     public String getHeadContent() {
-        return EMPTY;
+        String dataTablesHeader = MessageFormat.format(DATA_TABLES_TEMPLATE, "/", "/");
```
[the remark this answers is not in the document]
jzgithub1: That is a good idea.  I will give it a try.   I need to modify this pull request because I implemented search on the ModelList page and not the Model Page that you get to by clicking a link on the ModelList page.  That is what @ivakegg really wanted.  
 
By the way, the Quickstart uses an accumulo metadata table name called datawave.metadata which is not the default.  I need to pass the table name in on the query string to get the Model List page to render correctly I have to add the non-default tablename to the URLs for the Models that are generated in the Model List page too.

This PR needs is almost done but I am trying to resolve the merging conflict with Model.java.

jzgithub1:
Just rebased these changes onto version 2.6.  Simplified the initialization of the DATA_TABLES_TEMPLATE variable in the ModelList and Model classes to just be a hard coded string which is all it really is with one or two slight changes from page to page.  The MessageFormat class was just requiring the DATA_TABLES_TEMPLATE variable to way more complication than needed with all the escapes and single ticks and was not really getting us anything but complexity.  The JQuery and Datatable files are where the Spring framework puts them so why confuse things with more Java code that makes things harder to maintain.   If folks really want me to spend more time figuring out a way to make path insertions more dynamic I won't resist.  What I really need is a way to know what the metadata table name is inside the ModelList class so I can build the URL to the Model pages properly so they will have the '?modelTableName=datawave.metadata' in them but I am looking all over the call stack in the debugger to see how to do this and can't find a way.    A Sesson object in Spring or Rest Easy that you could access like  'Session['datawaveMetaTableName']  in Active Server Pages would be nice.

brianloss:
> 
> 
> @brianloss , I just rebased these changes onto version 2.6. Simplified the initialization of the DATA_TABLES_TEMPLATE variable in the ModelList and Model classes to just be a hard coded string which is all it really is with one or two slight changes from page to page. The MessageFormat class was just requiring the DATA_TABLES_TEMPLATE variable to way more complication than needed with all the escapes and single ticks and was not really getting us anything but complexity. The JQuery and Datatable files are where the Spring framework puts them so why confuse things with more Java code that makes things harder to maintain. If folks really want me to spend more time figuring out a way to make path insertions more dynamic I won't resist. What I really need is a way to know what the metadata table name is inside the ModelList class so I can build the URL to the Model pages properly so they will have the '?modelTableName=datawave.metadata' in them but I am looking all over the call stack in the debugger to see how to do this and can't find a way. A Sesson object in Spring or Rest Easy that you could access like 'Session['datawaveMetaTableName'] in Active Server Pages would be nice.

You are mixing things up. The JQuery and Datatable files are where we put them. The Model pages are served by the Wildfly webservice, not by Spring microservices.

As I mentioned in my previous comment, use `@ConfigProperty`. It should be pretty easy. Something like making a member variable:

```java
  @Inject @ConfigProperty(name = "dw.cdn.jquery.uri", defaultValue = "/jquery.min.js")
  private String jqueryUri;

  @Inject @ConfigProperty(name = "dw.cdn.dataTables.uri", defaultValue = "/jquery.min.js")
  private String dataTablesUri;
```

The same trick would probably work to handle the model table. You'd remove the `@DefaultValue` annotation, and instead in the method check if the value is null and if so, substitute in the default, like this:

```java
  @Inject @ConfigProperty(name = "dw.model.defaultTableName", defaultValue = "DatawaveMetadata")
  private String defaultModelTableName;

  ...

  public ModelList listModelNames(@QueryParam("modelTableName") String modelTableName) {
    if (modelTableName == null) {
        modelTableName = defaultModelTableName;
    }
  ...
```
You'd have to do this for every rest method using the DEFAULT_MODEL_TABLE_NAME constant.

Once this is all done, these properties are controllable by setting them in the wildfly configuration (in standalone/configuration/standalone-full.xml). See [here](https://github.com/NationalSecurityAgency/datawave/blob/release/version2.6/web-services/deploy/application/src/main/wildfly/add-datawave-configuration.cli#L46) for examples.
```

**Comment by jzgithub1** [comment]
Thanks.  Did I do the re-basing correctly?

**Comment by jwomeara** [comment]
> Thanks. Did I do the re-basing correctly?

Rebase looks ok to me.  Just need to add in the configurable base URIs.

**Comment by jzgithub1** [comment]
@brianloss ,  that   @ConfigProperty annotation did not work for me but I will keep working on this.  Thanks again for all you help here.

**Comment by brianloss** [comment]
> 
> 
> @brianloss , that @ConfigProperty annotation did not work for me but I will keep working on this. Thanks again for all you help here.

Oops, yeah I forgot that those classes aren't instantiated by CDI so therefore injection won't work. Looks like maybe everything is created via the ModelBean which is an EJB and therefore CDI injection will work. Therefore, you could turn those variables into constructor parameters for Model and ModelList, inject them in ModelBean, and pass them through when the object is constructed there.

**Comment by jzgithub1** [comment]
Thanks @brianloss.  I will do it that way.

File: web-services/client/src/main/java/datawave/webservice/model/ModelList.java [hunk 3]
```diff
@@ -55,7 +72,8 @@ public String getPageHeader() {
      */
     @Override
     public String getHeadContent() {
-        return EMPTY;
+        String dataTablesHeader = MessageFormat.format(DATA_TABLES_TEMPLATE, jqueryUri, dataTablesUri);
```
brianloss: Get rid of the unused local variable and change this to `return MessageFormat.format(DATA_TABLES_TEMPLATE, jqueryUri, dataTablesUri);` please.

web-services/model/src/main/java/datawave/webservice/query/model/ModelBean.java
```diff
@@ -110,7 +124,12 @@
     @GZIP
     @Interceptors(ResponseInterceptor.class)
     public ModelList listModelNames(@QueryParam("modelTableName") @DefaultValue(DEFAULT_MODEL_TABLE_NAME) String modelTableName) {
-        ModelList response = new ModelList();
+        
+        if (modelTableName == null) {
```
brianloss: This will never be true because of the `@DefaultValue(DEFAULT_MODEL_TABLE_NAME)` annotation on line 112/126. You have to remove that annotation for this approach to work, and do the same thing in every other method using the approach.

web-services/model/src/main/java/datawave/webservice/query/model/ModelBean.java
```diff
@@ -110,7 +124,12 @@
...
+            modelTableName = defaultModelTableName;
+        }
+        
+        ModelList response = new ModelList(modelTableName, dataTablesUri, defaultModelTableName);
```
brianloss: You shouldn't be passing in `defaultModelTableName` here. It should be `modelTableName` so that if the user calls this method with something other than the default model table name, it will pass through and be used in the generated links.

web-services/client/src/main/java/datawave/webservice/model/ModelList.java
```diff
@@ -13,8 +15,23 @@
 @XmlRootElement(name = "ModelList")
 public class ModelList extends BaseResponse implements Serializable, HtmlProvider {
     
+    private String jqueryUri;
+    private String dataTablesUri;
+    private String defaultModelTableName;
```
brianloss: This should be called `modelTableName` not `defaultModelTableName` since it is the selected model table name requested by the user (which could be the default, but isn't guaranteed to be).

jzgithub1:
@brianloss , this last push should work but I have not been able to test it because of issue #539  
I am going out of town next week and maybe somebody can test this code if they know of an easier way without the QuickStart setup which is not working for me now on the 2.6 branch.   I will keep on trying to test it on my system in any case.  I tried rolling back to a previous local branch where the quickstart was working and checking out the files I want to test but was not successful because the hadoop version and URLs seemed to changed on me. The the old 3.1.1 hadoop link is not good anymore.

web-services/client/src/main/java/datawave/webservice/model/ModelList.java
jzgithub1: Yes.  Agreed.

jzgithub1:
@brianloss,  requested change made.

brianloss:
> 
> 
> @brianloss, requested change made.

You didn't make all requested changes. Please read all the comments before pushing changes. Thanks.

jzgithub1:
Thanks for the fast review.

web-services/deploy/application/src/main/wildfly/add-datawave-configuration.cli
```diff
@@ -88,6 +88,7 @@ module add --name=com.mysql.driver --dependencies=javax.api,javax.transaction.ap
 /system-property=dw.metrics.reporter.port:add(value=${metrics.reporter.port})
 /system-property=dw.metrics.reporter.class:add(value=${metrics.reporter.class})
 /system-property=dw.cache.tableNames:add(value="${tables.to.cache}")
+/system-property=dw.model.defaultTableName:add(value=${table.name.metadata})
```
brianloss: I don't see a change to remove the property in `remove-datawave-configuration.cli`. Is it here and I missed it?

jzgithub1:
I made all of the requested changes.  Added a default no-argument constructor in the Model class so the mock object could get built in the ModelBeanTest class.

web-services/model/src/main/java/datawave/webservice/query/model/ModelBean.java
```diff
@@ -109,8 +123,13 @@
...
     @Interceptors(ResponseInterceptor.class)
-    public ModelList listModelNames(@QueryParam("modelTableName") @DefaultValue(DEFAULT_MODEL_TABLE_NAME) String modelTableName) {
-        ModelList response = new ModelList();
+    public ModelList listModelNames(@QueryParam("modelTableName") String modelTableName) {
+        
+        if (modelTableName == null) {
+            modelTableName = defaultModelTableName;
+        }
+        
+        ModelList response = new ModelList(modelTableName, dataTablesUri, modelTableName);
```
brianloss: Looks like you're passing `modelTableName` twice here. That doesn't match up with the `ModelList` parameters. The first param should be `jqueryUri`.

web-services/client/src/main/java/datawave/webservice/model/ModelList.java
```diff
@@ -13,8 +15,23 @@
...
     private static final long serialVersionUID = 1L;
-    private static final String TITLE = "Model Names", EMPTY = "";
+    private static final String TITLE = "Model Names";
+    private static final String DATA_TABLES_TEMPLATE = "<script type=''text/javascript'' src=''{0}''></script>\n"
+                    + "<script type=''text/javascript'' src=''{1}''></script>\n"
+                    + "<script type=''text/javascript''>\n"
+                    + "$(document).ready(function() '{' $(''#myTable'').dataTable('{'\"bPaginate\": false, \"aaSorting\": [[0, \"asc\"]], \"bStateSave\": true'}') '}')\n"
+                    + "</script>\n";
+    
+    public ModelList(String jqueryUri, String datatablesUri, String dataWaveMetadateTableName) {
```
brianloss: Shouldn't the last parameter be called `modelTableName`?

web-services/deploy/application/src/main/wildfly/add-datawave-configuration.cli
jzgithub1: Let me look.

jzgithub1: I tried to keep the alphabetizing scheme in tact: 

/system-property=dw.metrics.pool.admin.size:remove
/system-property=dw.model.defaultTableName:remove

web-services/client/src/main/java/datawave/webservice/model/ModelList.java
jzgithub1: The models are being stored in the metadata data table is what I am seeing.

web-services/deploy/application/src/main/wildfly/add-datawave-configuration.cli
jzgithub1: OK.  I see what happened.  The modification to the remove-datawave-configuration.cli was erased when I built the project again.
I will get that back in.  I did not know that could happen.

web-services/model/src/main/java/datawave/webservice/query/model/ModelBean.java
jzgithub1: The modelTable name has multiple points of entry.  On the query string or or dependency injection.  Or if it can't check the modelTable name (the metadata table name) it uses the DEFAULT_MODEL_TABLE_NAME in the ModelBean.java.

web-services/client/src/main/java/datawave/webservice/model/ModelList.java
brianloss: > 
> 
> The models are being stored in the metadata data table is what I am seeing.

I don't understand this comment. I was pointing out that your parameter `dataWaveMetadateTableName` is misnamed. First, you probably meant it to be `dataWaveMetadataTableName` but second, the parameter should really be called modelTableName because that's what it is (and that is also the member variable name into which you are storing the parameter). The fact that currently in the system we use the datawave metadata table to store model info doesn't mean that will always be the case.

web-services/model/src/main/java/datawave/webservice/query/model/ModelBean.java
brianloss: > 
> 
> The modelTable name has multiple points of entry. On the query string or or dependency injection. Or if it can't check the modelTable name (the metadata table name) it uses the DEFAULT_MODEL_TABLE_NAME in the ModelBean.java.

What does this have to do with my comment? You are still passing the wrong values to the ModelList constructor here.

jzgithub1:
It does not seem that the remove-datawave-configuration.cli is under source control because when I modify it, save it and do a git status, I see a 'working tree clean' message.

brianloss:
> 
> 
> It does not seem that the remove-datawave-configuration.cli is under source control because when I modify it, save it and do a git status, I see a 'working tree clean' message.

I promise you it's in source control: https://github.com/NationalSecurityAgency/datawave/blob/release/version2.6/web-services/deploy/application/src/main/wildfly/remove-datawave-configuration.cli

You must have some pattern in your global git ignore file. I can change the file and have it show up as modified just fine.

jzgithub1:
Let me check.

jzgithub1:
Found the one that in under source control here:
web-services/deploy/application/src/main/wildfly/remove-datawave-configuration.cli

jzgithub1:
@brianloss .  I just pushed that last mod.   Hope we are good now.

brianloss:
> 
> 
> @brianloss . I just pushed that last mod. Hope we are good now.

Nope, sorry. You still haven't addressed my comments in ModelBean.java, regarding this line:
```java
ModelList response = new ModelList(modelTableName, dataTablesUri, modelTableName);
```

web-services/client/src/main/java/datawave/webservice/model/ModelList.java
```diff
@@ -13,13 +15,28 @@
...
+    private static final String DATA_TABLES_TEMPLATE = "<script type=''text/javascript'' src=''{0}''></script>\n"
+                    + "<script type=''text/javascript'' src=''{1}''></script>\n"
+                    + "<script type=''text/javascript''>\n"
+                    + "$(document).ready(function() '{' $(''#myTable'').dataTable('{'\"bPaginate\": false, \"aaSorting\": [[0, \"asc\"]], \"bStateSave\": true'}') '}')\n"
+                    + "</script>\n";
+    
+    public ModelList(String jqueryUri, String datatablesUri, String dataWaveMetadataTableName) {
+        this.jqueryUri = jqueryUri;
+        this.dataTablesUri = datatablesUri;
+        this.modelTableName = dataWaveMetadataTableName;
```
brianloss: It seems my last comment on this was deleted. The parameter to ModelList should be named `modelTableName` not `dataWaveMetadataTableName`. It is the name of the model table. By default that happens to be the metadata table, but that might not always be the case. In other words, the function of that parameter is to supply the name of the table used for storing the model, not to supply the name of the datawave metadata table.

jzgithub1:
@brianloss .  The second parameter in the ModelList constructor is the path the to the DataTables javascript library and has nothing to do with the model name at all. That is handled by the 'names' hashset which gets used on line 100 to build the table.  

I fixed the misspelling of the third parameter.

web-services/model/src/main/java/datawave/webservice/query/model/ModelBean.java
brianloss: > 
> 
> @brianloss . The second parameter in the ModelList constructor is the path the to the DataTables javascript library and has nothing to do with the model name at all. That is handled by the 'names' hashset which gets used on line 100 to build the table.
> 
> I fixed the misspelling of the third parameter.

You're still missing it. First off, see my other comment about the name of the third parameter. The name does not match its purpose and I would like it changed. Regarding the error, I don't know how to express it more clearly.
Look at the 3 parameters you are passing:

1. modelTableName
2. dataTablesUri
3. modelTableName

Now look at what the constructor takes:

1.  jqueryUri
2. dataTablesUri
3. (misnamed) dataWaveMetadataTableName

Don't you see how passing modelTableName to a parameter that expects to receive the jqueryUri is wrong?

jzgithub1:
@brianloss.  Thanks Brian.  Sorry to irritate you.  Thanks for all of your help here.  Just wanted to get this ticket done before I left for a week.

PR #531, opened 2019-08-09
Clean up boolean logic in AuditController

### LINKED ISSUE #218: AuditController ignores false response from messageChannel.send()
The AuditController should react to messageChannel.send() being false with a RuntimeException to prevent a 2xx response code to the caller

lbschanno:
Throw a RuntimeException in AuditController.sendMessage() when
messageChannel.send returns false in order to prevent 2xx response
codes.

Fixes #218

### DIFF
--- a/services/audit-service/src/main/java/datawave/microservice/audit/AuditController.java
+++ b/services/audit-service/src/main/java/datawave/microservice/audit/AuditController.java
@@ -122,7 +122,7 @@ public class AuditController {
             }
             
             boolean success = messageChannel.send(MessageBuilder.withPayload(AuditMessage.fromParams(parameters)).setCorrelationId(auditId).build());
-
+            
             if (auditProperties.isConfirmAckEnabled()) {
                 try {
                     success = success && latch.await(auditProperties.getConfirmAckTimeoutMillis(), TimeUnit.MILLISECONDS);
@@ -132,7 +132,7 @@ public class AuditController {
                     correlationLatchMap.remove(auditId);
                 }
             }
-
+            
             return success;
         }
         
--- a/services/audit-service/src/main/java/datawave/microservice/audit/AuditController.java
+++ b/services/audit-service/src/main/java/datawave/microservice/audit/AuditController.java
@@ -112,7 +112,7 @@ public class AuditController {
      *            The audit parameters to be sent
      */
     private boolean sendMessage(AuditParameters parameters) {
-        if ((healthChecker != null && healthChecker.isHealthy()) || healthChecker == null) {
+        if (healthChecker == null || healthChecker.isHealthy()) {
             String auditId = parameters.getAuditId();
             
             CountDownLatch latch = null;
@@ -122,7 +122,7 @@ public class AuditController {
             }
             
             boolean success = messageChannel.send(MessageBuilder.withPayload(AuditMessage.fromParams(parameters)).setCorrelationId(auditId).build());
-            
+
             if (auditProperties.isConfirmAckEnabled()) {
                 try {
                     success = success && latch.await(auditProperties.getConfirmAckTimeoutMillis(), TimeUnit.MILLISECONDS);
@@ -132,7 +132,7 @@ public class AuditController {
                     correlationLatchMap.remove(auditId);
                 }
             }
-            
+
             return success;
         }
         


services/audit-service/src/main/java/datawave/microservice/audit/AuditController.java
```diff
@@ -122,17 +122,22 @@ private boolean sendMessage(AuditParameters parameters) {
             }
             
             boolean success = messageChannel.send(MessageBuilder.withPayload(AuditMessage.fromParams(parameters)).setCorrelationId(auditId).build());
-            
+
+            // Prevent a 2xx response code to the caller.
+            if (!success) {
```
jwomeara: When this ticket was created the audit service had no concept of writing audit messages that failed to publish to the message channel to disk.  We have come a long way since then, so now when an audit message fails to send, we attempt to write it to disk before returning a failed status.  If we start throwing an exception here, we completely skip the logic that attempts to write the audit message to disk.  

The reality is that this was a stale ticket that we resolved a while ago which should have been closed.

services/audit-service/src/main/java/datawave/microservice/audit/AuditController.java
```diff
@@ -112,7 +112,7 @@ public void processConfirmAck(Message<?> message) {
      *            The audit parameters to be sent
      */
     private boolean sendMessage(AuditParameters parameters) {
-        if ((healthChecker != null && healthChecker.isHealthy()) || healthChecker == null) {
+        if (healthChecker == null || healthChecker.isHealthy()) {
```
jwomeara: Yup.  This is cleaner than the previous statement.  Good change.

services/audit-service/src/main/java/datawave/microservice/audit/AuditController.java
```diff
@@ -122,17 +122,22 @@ private boolean sendMessage(AuditParameters parameters) {
...
+
+            // Prevent a 2xx response code to the caller.
+            if (!success) {
+                throw new RuntimeException("Failed to send audit message " + auditId);
+            }
+
             if (auditProperties.isConfirmAckEnabled()) {
                 try {
-                    success = success && latch.await(auditProperties.getConfirmAckTimeoutMillis(), TimeUnit.MILLISECONDS);
+                    success = latch.await(auditProperties.getConfirmAckTimeoutMillis(), TimeUnit.MILLISECONDS);
```
jwomeara: I agree that this change makes sense if we add the RuntimeException above, but per my previous comment, we should not start doing that.  So, I think this should remain as-is.  Otherwise, the message send failure will not carry forward.

services/audit-service/src/main/java/datawave/microservice/audit/AuditController.java
lbschanno: Understood, I'll modify the commit to only address cleaning up the boolean logic that you noted below, and revert my other changes.

lbschanno:
Forced push an edit that reverts the unnecessary changes and updated the commit message to reflect the new change details.

services/audit-service/src/main/java/datawave/microservice/audit/AuditController.java
jwomeara: Awesome!  Thanks!

services/audit-service/src/main/java/datawave/microservice/audit/AuditController.java
```diff
@@ -122,7 +122,7 @@ private boolean sendMessage(AuditParameters parameters) {
             }
             
             boolean success = messageChannel.send(MessageBuilder.withPayload(AuditMessage.fromParams(parameters)).setCorrelationId(auditId).build());
-            
+
```
brianloss: These whitespace changes seem to indicate that the formatter hasn't been run on this file. Can you ensure that you've run the maven build and then include any files that got reformated as a result?

lbschanno: I ran the build and no files were reformatted. My IDE is set to remove indents with empty lines (this is after importing the datawave code-style settings). Should empty-line indents be preserved?

brianloss: I think the formatter puts the indents on empty lines. You've saved the file in the IDE after building, then you would get the problem. If you have your IDE closed, build, and see no modified files, then everything's ok.

brianloss: Yeah, I built the target of this PR, and I do end up with a modified AuditController.java file.

lbschanno: Confirmed, the issue was an IDE setting. I restored the whitespace and updated my IDE to preserve indents.

PR #538, opened 2019-08-15
Added logging to flagmaker load function

jkrdev:
Some simple logging for information on the file to be loaded

### DIFF
--- a/warehouse/ingest-core/src/main/java/datawave/util/flag/FlagMaker.java
+++ b/warehouse/ingest-core/src/main/java/datawave/util/flag/FlagMaker.java
@@ -276,7 +276,7 @@ public class FlagMaker implements Runnable, Observer, SizeValidator {
     }
 
     protected void logFileInfo(FlagDataTypeConfig fc, FileStatus status) {
-        log.info("File " + fc.getDataName() + " : " + status);
+        log.info("File {} : {}", fc.getDataName(), status);
     }
     
     private boolean shouldOnlyCreateFullFlags(FlagDataTypeConfig fc) {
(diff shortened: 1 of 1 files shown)


warehouse/ingest-core/src/main/java/datawave/util/flag/FlagMaker.java
```diff
@@ -266,13 +266,19 @@ void loadFilesForDistributor(FlagDataTypeConfig fc, FileSystem fs) throws IOExce
...
                     } catch (UnusableFileException e) {
                         log.warn("Skipping unusable file " + status.getPath(), e);
                     }
                 }
             }
         }
     }
+
+    protected void logFileInfo(FlagDataTypeConfig fc, FileStatus status) {
+        log.info("File type is " + fc.getDataName());
```
ivakegg: Lets combine these lines into one to avoid too much log bloat:
log.info("File " + fc.getDataName() + " : " + status);

jkrdev: sounds good changes made in update

warehouse/ingest-core/src/main/java/datawave/util/flag/FlagMaker.java
```diff
@@ -266,13 +266,18 @@ void loadFilesForDistributor(FlagDataTypeConfig fc, FileSystem fs) throws IOExce
...
                     } catch (UnusableFileException e) {
                         log.warn("Skipping unusable file " + status.getPath(), e);
                     }
                 }
             }
         }
     }
+
+    protected void logFileInfo(FlagDataTypeConfig fc, FileStatus status) {
+        log.info("File " + fc.getDataName() + " : " + status);
```
brianloss: This should probably be:
```
log.info("File {} : {}", fc.getDataName(), status);
```
So that we don't have to compute the combined string if we're not logging to the info level.

jkrdev: done.

Issue #533, opened 2019-08-12
Accumulo 2.0 Migration


milleruntime:
Accumulo 2.0 is now released: https://accumulo.apache.org/release/accumulo-2.0.0/
Opening this issue to put notes and as a reference for work to be done to get Datawave working with 2.0. 

Currently the code doesn't compile for 2.0 so that is probably a good place to start.

milleruntime:
I have got Datawave master branch to compile with Accumulo 2.0 on a branch [here](https://github.com/milleruntime/datawave/tree/accumulo-2.0).  Changes I found that were not specific to 2.0 I have made into pull requests.  I will try to keep this branch up to date with latest master until it can be brought into the Datawave repo.   The next biggest requirement will be upgrading Guava to something that works, see #514

milleruntime:
I saw this Exception in some reflection code when installing ingest:
```
Exception in thread "main" java.lang.RuntimeException: Unable to call generateInitialTableProperties
	at datawave.ingest.table.config.AbstractTableConfigHelper.generateInitialTableProperties(AbstractTableConfigHelper.java:122)
	at datawave.ingest.table.config.AbstractTableConfigHelper.areAggregatorsConfigured(AbstractTableConfigHelper.java:146)
	at datawave.ingest.table.config.AbstractTableConfigHelper.setAggregatorConfigurationIfNecessary(AbstractTableConfigHelper.java:86)
	at datawave.ingest.table.config.ShardTableConfigHelper.configureShardTable(ShardTableConfigHelper.java:167)
	at datawave.ingest.table.config.ShardTableConfigHelper.configure(ShardTableConfigHelper.java:140)
	at datawave.ingest.mapreduce.job.IngestJob.configureTablesIfNecessary(IngestJob.java:1225)
	at datawave.ingest.mapreduce.job.IngestJob.createAndConfigureTablesIfNecessary(IngestJob.java:1162)
	at datawave.ingest.mapreduce.job.IngestJob.configureTables(IngestJob.java:743)
	at datawave.ingest.mapreduce.job.IngestJob.run(IngestJob.java:298)
	at org.apache.hadoop.util.ToolRunner.run(ToolRunner.java:76)
	at datawave.ingest.mapreduce.job.IngestJob.main(IngestJob.java:210)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #540, opened 2019-08-19
Replace accumulo MockInstance with MiniAccumuloCluster

milleruntime:
This will aid in upgrading to 2.0 as Mock was removed.  Related to #533

### DIFF
--- a/warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/TableConfigHelperFactoryTest.java
+++ b/warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/TableConfigHelperFactoryTest.java
@@ -8,13 +8,12 @@ import org.apache.accumulo.core.client.Connector;
 import org.apache.accumulo.core.client.admin.TableOperations;
 import org.apache.accumulo.minicluster.MiniAccumuloCluster;
 import org.apache.accumulo.minicluster.MiniAccumuloConfig;
+import org.apache.commons.io.FileUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.log4j.Logger;
-import org.junit.After;
-import org.junit.Before;
-import org.junit.Rule;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
 import org.junit.Test;
-import org.junit.rules.TestName;
 
 import static org.hamcrest.core.Is.is;
 import static org.hamcrest.core.IsNull.nullValue;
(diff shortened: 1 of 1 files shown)


warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/TableConfigHelperFactoryTest.java
```diff
@@ -6,29 +6,48 @@
...
+    public void startCluster() throws Exception {
+        File macDir = new File(System.getProperty("user.dir") + "/target/mac/" + TableConfigHelperFactoryTest.class.getName() + "-" + name.getMethodName());
+        macDir.mkdirs();
+        mac = new MiniAccumuloCluster(new MiniAccumuloConfig(macDir, "pass"));
+        mac.start();
+    }
+    
     @Before
     public void setup() throws Exception {
+        startCluster();
```
brianloss: Is it possible to put this in a `@BeforeClass` and the shutdown in an `@AfterClass` method so we don't have to pay the startup/shutdown penalty on every test method? Looks like it's only 2 in this class, but the change would then let it run twice as fast.

milleruntime: Yeah I just had to make each test delete and create the tables and make everything static.

milleruntime: Tests takes about 5s now on my dev box.

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/TableConfigHelperFactoryTest.java
```diff
@@ -55,6 +73,9 @@ private void recreateTable(TableOperations tops, String table) throws Exception
     
     @Test
     public void shouldSetupTableWithOverrides() throws Exception {
+        recreateTable(tops, TableName.SHARD);
```
brianloss: The `@Before` method would still get run, before every test, so I think it would be better to have these two lines in a method annotated with that, rather than in each test.

milleruntime: Fixed in 3f4c0fe

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/TableConfigHelperFactoryTest.java
```diff
@@ -6,27 +6,44 @@
...
     private TableOperations tops;
     
     private static final String TEST_SHARD_TABLE_NAME = "testShard";
     
+    @BeforeClass
+    public static void startCluster() throws Exception {
+        File macDir = new File(System.getProperty("user.dir") + "/target/mac/" + TableConfigHelperFactoryTest.class.getName());
+        if (macDir.exists())
+            FileUtils.deleteDirectory(macDir);
+        macDir.mkdirs();
```
brianloss: Should we have a deleteOnExit call for this, or delete it in the shutdown method?

milleruntime: The mini accumulo files are useful for debugging failures in the Test.  The directory is in target so can easily be removed on a "mvn clean".

Issue #543, opened 2019-08-22
Datawave quickstart docker build is failing


zachmullen:
Platform: ubuntu 16.04
Docker version: 19.03.1

Relevant log output from running `docker-build.sh latest`:

```
Step 14/25 : RUN /bin/bash -c "/usr/bin/nohup /usr/sbin/sshd -D &" &&     /bin/bash -c "source ~/.bashrc && allInstall && datawaveStart && datawaveWebTest && allStop" &&     echo "0.0.0.0" > contrib/datawave-quickstart/accumulo/conf/monitor &&     rm -rf contrib/datawave-quickstart/datawave-ingest/logs/* &&     rm -rf contrib/datawave-quickstart/hadoop/logs/* &&     rm -rf contrib/datawave-quickstart/accumulo/logs/* &&     rm -rf contrib/datawave-quickstart/wildfly/standalone/log/*
 ---> Running in caeafaf48a90
[DW-INFO] - Maven installed
cp: cannot stat '': No such file or directory
[DW-INFO] - Maven is already installed
cp: cannot stat '': No such file or directory
cp: cannot stat '': No such file or directory
[DW-FATAL] - Hadoop tarball not found
Aborting install.sh
cp: cannot stat '': No such file or directory
cp: cannot stat '': No such file or directory
[DW-FATAL] - Accumulo requires that Hadoop be installed
Aborting install.sh
cp: cannot stat '': No such file or directory
cp: cannot stat '': No such file or directory
[DW-FATAL] - DataWave Ingest requires that Hadoop be installed
Aborting install-ingest.sh
cp: cannot stat '': No such file or directory
cp: cannot stat '': No such file or directory
[DW-FATAL] - DataWave Web requires that Accumulo be installed
Aborting install-web.sh
The command '/bin/sh -c /bin/bash -c "/usr/bin/nohup /usr/sbin/sshd -D &" &&     /bin/bash -c "source ~/.bashrc && allInstall && datawaveStart && datawaveWebTest && allStop" &&     echo "0.0.0.0" > contrib/datawave-quickstart/accumulo/conf/monitor &&     rm -rf contrib/datawave-quickstart/datawave-ingest/logs/* &&     rm -rf contrib/datawave-quickstart/hadoop/logs/* &&     rm -rf contrib/datawave-quickstart/accumulo/logs/* &&     rm -rf contrib/datawave-quickstart/wildfly/standalone/log/*' returned a non-zero code: 1
[...]

PR #544, opened 2019-08-22
Address HTTP 404 from hadoop tarball download

zachmullen:
Version 3.1.1 is not hosted on that server, but 3.1.2 is.

Refs #543


zachmullen:
Fixes #543

PR #558, opened 2019-09-03
Deprecate eventPerDayThreshold and shardsPerDayThreshold properties.

apmoriarty:
After taking a look at some of the index lookup code and running some experiments I found that rolling index ranges from event or shard ranges to day ranges had no benefit. In fact, rolling to day ranges incurs a small penalty.

This PR deprecates the mechanisms to roll event and shard ranges to a day ranges during the index lookup portion of query execution. Removal will occur in a separate PR.


apmoriarty:
Rather than removing the underlying events/shards per day threshold properties, I'm opting to deprecate them.

The shardsPerDayThreshold is removed from the RangeStreamScanner. This ensures we do not roll to a day range based on this property.

warehouse/query-core/src/test/java/datawave/query/DataTypeQueryTest.java
```diff
@@ -126,6 +126,7 @@ public void testEventPerDay() throws Exception {
         String dtFilter = CityEntry.generic.getDataType();
         qOptions.put(QueryParameters.DATATYPE_FILTER_SET, dtFilter);
         
+        // Setting this property has no effect on test results.
```
ivakegg: Then don't set it?

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
```diff
@@ -960,18 +962,22 @@ public void setUnevaluatedFields(String unevaluatedFieldList) {
...
         return eventPerDayThreshold;
     }
     
+    @Deprecated
     public void setEventPerDayThreshold(int eventPerDayThreshold) {
         this.eventPerDayThreshold = eventPerDayThreshold;
     }
     
+    @Deprecated
     public int getShardsPerDayThreshold() {
```
ivakegg: Are we going to remove its use in the RangeStream?

warehouse/query-core/src/main/java/datawave/query/tables/RangeStreamScanner.java
```diff
@@ -313,8 +312,7 @@ protected int scannerInvariant(final Iterator<Entry<Key,Value>> iter) {
                         
                         stats.addValue(info.count());
                         
-                        if (currentQueue.size() <= shardsPerDayThreshold
```
ivakegg: Since we are not actually removing the functionality yet, move the changes in this class to you change which will rework the index lookup altogether.

apmoriarty: Can do

PR #562, opened 2019-09-09
Replaced TreeHashVisitor with an enhanced JexlStringBuildingVisitor t…

jwomeara:
…o fix a NPE and updated usages throughout.


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlStringBuildingVisitor.java
```diff
@@ -154,14 +196,14 @@ public Object visit(ASTAndNode node, Object data) {
             sb.append("(");
         }
         
-        int lastsize = sb.length();
+        Collection<String> childStrings = (sortDedupeChildren) ? new TreeSet<>() : new ArrayList<>();
```
apmoriarty: The ArrayList could take the number of child nodes as an argument.

jwomeara: Sure.  See latest commit.

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
```diff
@@ -692,4 +691,8 @@ private boolean isInfinite() {
     public void setNode(JexlNode currNode) {
         myNode = currNode;
     }
+    
+    private String nodeToKey(JexlNode node) {
+        return JexlStringBuildingVisitor.buildQueryWithoutParse(TreeFlatteningRebuildingVisitor.flatten(node), true);
```
ivakegg: we call this a lot.  Is it possible to flatten the tree once?

jwomeara: Per our conversation, I verified that in all of our test cases the node passed in to nodeToKey is already flattened.  We should be able to safely remove the flatten call from this method.

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
```diff
@@ -83,9 +82,9 @@ public static boolean isMatched(Object scriptExecuteResult) {
     
     @Override
     public Object visit(ASTFunctionNode node, Object data) {
-        TreeHashNode hash = TreeHashVisitor.getNodeHash(node);
+        String nodeString = JexlStringBuildingVisitor.buildQuery(node);
```
ivakegg: buildQueryWithoutParse?

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
```diff
@@ -692,4 +691,9 @@ private boolean isInfinite() {
     public void setNode(JexlNode currNode) {
         myNode = currNode;
     }
+    
+    private String nodeToKey(JexlNode node) {
+        // Note: This method assumes that the node passed in is already flattened.
```
ivakegg: The cost of not being flattened is....

PR #566, opened 2019-09-10
fixes #565 Add PeekingPrediciate

### LINKED ISSUE #565: TLDEventDataFilter.keep() should not impact the parent filter
When TLDEventDataFilter.keep() is called it should not trigger a change in state in the filter. It is using the super.apply() to asses inclusion of child data when non-event fields are in use, but this will cause the underlying Expression filter to potentially change state.

Add a new interface PeekingPredicate which can be used to peek at the result of a Predicate.apply() method without actually applying.

FineAndDandy:


warehouse/query-core/src/main/java/datawave/query/predicate/PeekingPredicate.java
```diff
@@ -0,0 +1,15 @@
+package datawave.query.predicate;
+
+import com.google.common.base.Predicate;
+
+import javax.annotation.Nullable;
+
+public interface PeekingPredicate<T> extends Predicate<T> {
+    /**
+     * peek at the result of applying an input without actually applying it
```
ivakegg: Lets make this clear that this is equivalent to apply unless the underlying implementation saves state affecting subsequent apply calls.  Basically peek is intended to execute apply without incurring any side effects.

ivakegg: Perhaps even a pointer (using @see) denoting an instance where this actually makes a difference.

Issue #559, opened 2019-09-06
Track active queries in tserver


billoley:
To make it easier to find problematic queries, we should begin tracking which queries are active on a tserver including:

queryId
duration of query
bytes returned
seeks/next/sources

To limit logging, we should log the active queries periodically through a Timer.

PR #568, opened 2019-09-11
Re #559 Track active queries in the QueryIterator

billoley:
We want a way to see in the logs which queries are putting a load on a specific tserver.

Example logging output:

[GAPZW-47TWB-TUDP7-W7JHA] (tot/cur) 4123/3130 (call num/max/avg/min) seek=2/6/4/3 next=11/1/0/0 (lastDoc bytes/sources/seek/next) 352/1/4/22

The frame of reference is successive seek and next calls on a specific tserver for the same queryId
[query_id] 
total_call_time/current_call_time 
numSeeks/maxMsec/avgMsec/minMsec 
numNexts/maxMsec/avgMsec/minMsec lastDocReturnedSizeButes/numSources/numSourceSeeks/numSourceNexts

- Query records that have not been accessed in maxIdle time will be dropped (configurable)
- The timer in ActiveQueryLog will print this message at the DEBUG level for each query that is in a call (seek or next) at a configurable period.
- The max/avg/min calculations for seek and next are based on a sliding window of the last N calls.  N defaults to 10 and is configurable via AccumuloConfiguration.

datawave.query.active.maxIdle - default=900000 (15 minutes)
datawave.query.active.logPeriod - default=60000 (1 minute)
datawave.query.active.windowSize - default=10

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQuery.java
+++ b/warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQuery.java
@@ -5,6 +5,8 @@ import com.codahale.metrics.Timer;
 import datawave.query.attributes.Document;
 import datawave.query.iterator.profile.QuerySpan;
 import org.apache.accumulo.core.data.Range;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.HashMap;
 import java.util.HashSet;
@@ -13,6 +15,8 @@ import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
 public class ActiveQuery {
+    
+    private static Logger LOG = LoggerFactory.getLogger(ActiveQuery.class);
     
     private String queryId = null;
     private long initTs = 0;
@@ -70,10 +74,14 @@ public class ActiveQuery {
     }
     
     synchronized public void recordStats(Document document, QuerySpan querySpan) {
-        this.documentSizeBytes = document.sizeInBytes();
-        this.lastSourceCount = querySpan.getSourceCount();
-        this.lastSeekCount = querySpan.getSeekCount();
-        this.lastNextCount = querySpan.getNextCount();
+        if (document != null) {
+            this.documentSizeBytes = document.sizeInBytes();
+        }
+        if (querySpan != null) {
+            this.lastSourceCount = querySpan.getSourceCount();
+            this.lastSeekCount = querySpan.getSeekCount();
+            this.lastNextCount = querySpan.getNextCount();
+        }
     }
     
     synchronized public int removeRange(Range range) {
@@ -128,6 +136,7 @@ public class ActiveQuery {
             newNumCalls = numCalls.intValue() - 1;
         }
         if (newNumCalls < 0) {
+            LOG.info("inCall count for callType:" + type.toString() + "=" + newNumCalls + ", resetting to 0");
             newNumCalls = 0;
         }
         this.inCallMap.put(type, newNumCalls);
--- a/warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQuery.java
+++ b/warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQuery.java
@@ -1,25 +1,24 @@
 package datawave.query.tracking;
 
 import com.codahale.metrics.SlidingWindowReservoir;
-import com.codahale.metrics.Snapshot;
 import com.codahale.metrics.Timer;
 import datawave.query.attributes.Document;
 import datawave.query.iterator.profile.QuerySpan;
 import org.apache.accumulo.core.data.Range;
 
-import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
-import java.util.function.Predicate;
 
 public class ActiveQuery {
     
     private String queryId = null;
     private long initTs = 0;
-    private QuerySpan lastQuerySpan = null;
+    private long lastSourceCount = 0;
+    private long lastNextCount = 0;
+    private long lastSeekCount = 0;
     private long documentSizeBytes = 0;
     private int windowSize = 0;
     
@@ -39,13 +38,9 @@ public class ActiveQuery {
         this.initTs = System.currentTimeMillis();
     }
     
-    synchronized public long getNumCalls(CallType type) {
-        Long val = this.numCallsMap.get(type);
-        if (val == null) {
-            val = 0l;
-            this.numCallsMap.put(type, val);
-        }
-        return val;
+    synchronized public ActiveQuerySnapshot snapshot() {
+        return new ActiveQuerySnapshot(this.queryId, this.lastSourceCount, this.lastNextCount, this.lastSeekCount, this.documentSizeBytes,
+                        this.activeRanges.size(), this.totalElapsedTime(), this.isInCall(), this.currentCallTime(), this.numCallsMap, this.timerMap);
     }
     
     synchronized public void beginCall(Range range, CallType type) {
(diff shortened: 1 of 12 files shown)


warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
```diff
@@ -0,0 +1,203 @@
...
+        if (conf != null) {
+            checkSettings(conf, true);
+        } else {
+            // use the default values
+            setLogPeriod(this.logPeriod);
+            setMaxIdle(this.maxIdle);
+        }
+    }
+    
+    public void setLogPeriod(long logPeriod) {
```
ivakegg: make private or synchronize

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
```diff
@@ -0,0 +1,203 @@
...
+        public void run() {
+            synchronized (this) {
+                List<ActiveQuery> activeQueryList = new ArrayList<>();
+                for (ActiveQuery q : CACHE.asMap().values()) {
+                    activeQueryList.add(q);
+                }
+                activeQueryList.sort(activeQueryList, new TotalTimeComparator());
+
+                for (ActiveQuery q : CACHE.asMap().values()) {
+                    if (q.isInCall()) {
```
ivakegg: Add an option to see all calls....

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQuery.java
```diff
@@ -0,0 +1,200 @@
...
+    private Map<CallType,Integer> inCallMap = new HashMap<>();
+    private Set<Range> activeRanges = new HashSet<>();
+    
+    public ActiveQuery(String queryId, int windowSize) {
+        this.queryId = queryId;
+        this.windowSize = windowSize;
+        this.initTs = System.currentTimeMillis();
+    }
+    
+    synchronized public long getNumCalls(CallType type) {
```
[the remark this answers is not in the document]
billoley: synchronized per ActiveQuery object for writes and snapshots.
Snapshot objects are used for logging to enable sorting, sublists while minimizing synchronization and allowing consistent numbers throughout the sort/sublist/etc.

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQuery.java
```diff
@@ -0,0 +1,200 @@
...
+            Long start = rangeMap.get(range);
+            if (start != null) {
+                getTimer(type).update(System.currentTimeMillis() - start, TimeUnit.MILLISECONDS);
+                rangeMap.remove(range);
+                setInCall(type, false);
+            }
+        }
+    }
+    
+    public void recordStats(Document document, QuerySpan querySpan) {
```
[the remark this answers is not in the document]
billoley: Made synchronized

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
billoley: synchronized

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
billoley: Took out the check is isInCall, so all active queries are eligible to be displayed.

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQuery.java
```diff
@@ -0,0 +1,151 @@
...
+            numCalls = new Integer(0);
+        }
+        int newNumCalls;
+        if (inCall) {
+            newNumCalls = numCalls.intValue() + 1;
+        } else {
+            newNumCalls = numCalls.intValue() - 1;
+        }
+        if (newNumCalls < 0) {
+            newNumCalls = 0;
```
ivakegg: Perhaps this requires a warning message as it implies we are ending more calls than we are starting.....

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
```diff
@@ -0,0 +1,198 @@
...
+    public static final String LOG_MAX_QUERIES = "datawave.query.active.logMaxQueries";
+    public static final String WINDOW_SIZE = "datawave.query.active.windowSize";
+    
+    // Changeable via Accumulo properties
+    private long maxIdle = 900000;
+    private long logPeriod = 60000;
+    private int logMaxQueries = 5;
+    private int windowSize = 10;
+    
+    private Cache<String,ActiveQuery> CACHE = null;
```
ivakegg: Given it appears that setMaxIdle could never be called, then the CACHE could remain null and then the get() methods would throw an NPE.  I think we should give it an initial value here to guarentee that cannot happen.

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
```diff
@@ -0,0 +1,198 @@
...
+    synchronized public void remove(String queryId, Range range) {
+        ActiveQuery activeQuery = get(queryId);
+        int numActiveRanges = activeQuery.removeRange(range);
+        if (numActiveRanges == 0) {
+            this.CACHE.invalidate(queryId);
+        }
+    }
+    
+    synchronized public ActiveQuery get(String queryId) {
+        return this.CACHE.get(queryId, s -> new ActiveQuery(queryId, this.windowSize));
```
ivakegg: Is there any way to avoid this method from being synchronized.  Optimally we only synchronize when we need to add a new ActiveQuery (i.e on write).

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+    
+    class ActiveQueryTimerTask extends TimerTask {
+        
+        public ActiveQueryTimerTask() {}
+        
+        @Override
+        public void run() {
+            List<ActiveQuerySnapshot> activeQueryList = new ArrayList<>();
+            for (ActiveQuery q : CACHE.asMap().values()) {
```
ivakegg: If we have to synchronize on read in the get() method, do we need to here as well?

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
billoley: The private constructor (called from getInstance()) of ActiveQueryLog calls either 
setMaxIdle directly
or
checkSettings [with useDefaults=true] (which calls setMaxIdle with either the default value of the value from conf)

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
billoley: CACHE operations are threadsafe per the API.  The syncronization was due to the possibility of replacing the CACHE object itself when setMaxIdle is changed.  I removed the syncronized on get(), remove(), and setMaxIdle() and used a ReentrandReadWriteLock when using the CACHE object (excpet for the null check in setMaxIdle() which is re-checked inside the locked section anyway.

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
billoley: Good point -- I will use the ReentrantReadWriteLock.readLock() here as well.

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
ivakegg: What if the conf does not have a value for maxIdle?

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQuery.java
billoley: changed from info to warn.

PR #573, opened 2019-09-17
Adds ability to plot geo query geometries on a map via the query metrics endpoint.

jwomeara:


ivakegg:
Wow, this may take a minute or two to review....

web-services/client/src/main/java/datawave/webservice/query/metric/BaseQueryMetricListResponse.java
```diff
@@ -59,6 +61,14 @@ public void setAdministratorMode(boolean administratorMode) {
         this.administratorMode = administratorMode;
     }
     
+    public boolean isGeoQuery() {
```
ivakegg: Since we are changing the response, do we need to start versioning such that we can support a v1 and v2 response?  @brianloss What do you think?

jwomeara: The XML and JSON responses are the same.  The HTML response is the only one that changed, FWIW.

jwomeara: And the HTML response is ONLY changed if the query contains a geo term...  Again, just providing more food for thought.

brianloss: Yeah, since this is not part of the serialized response, there's no need to version anything.

web-services/client/src/main/java/datawave/webservice/query/map/QueryGeometryResponse.java
```diff
@@ -0,0 +1,105 @@
...
+    public QueryGeometryResponse() {
+        this(null, null);
+    }
+    
+    public QueryGeometryResponse(String queryId, String basemaps) {
+        this.queryId = queryId;
+        this.basemaps = basemaps;
+    }
+    
+    @XmlElement(name = "queryId")
```
brianloss: Should this have `, nillable = true` added?

web-services/client/src/main/java/datawave/webservice/query/map/QueryGeometry.java
```diff
@@ -0,0 +1,68 @@
...
+    
+    public String getGeometry() {
+        return geometry;
+    }
+    
+    public void setGeometry(String geometry) {
+        this.geometry = geometry;
+    }
+    
+    String toGeoJsonFeature() {
```
brianloss: Should this be `protected`?

jwomeara: What is going on with gitlab?  I added a comment on here saying that it can be public, but it isn't showing up...

PR #588, opened 2019-10-01
fixes #587: Updated global index lookups to restrict results by fieldname.

### LINKED ISSUE #587: Ignore unindexed fields found in the index
If the planner thinks a field is not-indexed, then any index entries it finds should be ignored.  This is to handle fields that were indexed at one time and then we stop indexing them and remove the index marker in the Datawavemetadata.  The consequence of not doing this is that we may expand unfielded terms with a non-indexed field resulting in a non-executable query.

ivakegg:
Ensure configuration contains both indexed and reverse indexed fields.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
+++ b/warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
@@ -223,6 +223,7 @@ public class DefaultQueryPlanner extends QueryPlanner {
     private static Multimap<String,Type<?>> normalizedFieldAsDataTypeMap;
     
     private static Set<String> cachedIndexedFields = null;
+    private static Set<String> cachedReverseIndexedFields = null;
     private static Set<String> cachedNormalizedFields = null;
     
     protected List<PushDownRule> rules = Lists.newArrayList();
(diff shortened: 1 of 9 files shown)


warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -1218,18 +1222,21 @@ protected ASTJexlScript processTree(final ASTJexlScript originalQueryTree, Shard
...
+                            && (null != cachedReverseIndexedFields && null != reverseIndexedFields)) {
                 indexedFields.addAll(cachedIndexedFields);
+                reverseIndexedFields.addAll(cachedReverseIndexedFields);
                 normalizedFields.addAll(cachedNormalizedFields);
                 
                 return;
             }
             
             cachedIndexedFields = metadataHelper.getIndexedFields(null);
+            cachedReverseIndexedFields = metadataHelper.getReverseIndexedFields(null);
```
ivakegg: I am thinking this should be filtered by datatype here.....  Or we remove this whole caching mechanism altogether.

drewfarris:
I could be missing something, but it looks like the list of literals that is supported in the filter implementation isn't really ever used because in the end, the `RegexTrie` is always used from `ShardIndexQueryTableStaticFunctions`. 

It would be really nice if we could avoid using patterns for this *always*. Do we know what point does a `RegexTrie` become more efficient than a large list of strings? 

The bottom line is, if we've never going to use literals in the iterator/filter implementation, does it make sense to have this code in place?

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
drewfarris: Yeah, I can't really tell what's supposed to be happening in the `DefaultQueryPlanner` where this method is called.

ivakegg:
> It would be really nice if we could avoid using patterns for this _always_. Do we know what point does a `RegexTrie` become more efficient than a large list of strings?

I create a performance test that tests matching against a set of literals vs a regex.  The regex was always slower and in fact the gap increased as we added more fieldnames to the list.  I will remove the regex mechanism.  I need to ensure the field name list is compressed however as that list could get large.

ivakegg:
Closed until I can determine whether the datatype cache can be removed, and change the use of the RegexTrie

PR #592, opened 2019-10-07
Update Guava to 22

milleruntime:
* Replaced various dropped methods
* Replaced Stopwatch contstructor & elapsedMillis
* Replaced Iterators.emptyIterator with java Collections
* Catch new IllegalStateException in ScannerSession

This will get DW closer to Accumulo 2.0

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/tables/ScannerSession.java
+++ b/warehouse/query-core/src/main/java/datawave/query/tables/ScannerSession.java
@@ -279,8 +279,15 @@ public class ScannerSession extends AbstractExecutionThreadService implements It
             // until we've completed the start process
             if (null != stats)
                 initializeTimers();
-            startAndWait();
             
+            // these two guava methods replaced behavior of startAndWait() from version 15 but
+            // will now throw an exception if another thread closes the session so catch and ignore
+            startAsync();
+            try {
+                awaitRunning();
+            } catch (IllegalStateException e) {
+                log.debug("Session was closed while waiting to start up.");
+            }
         }
         
         // isFlushNeeded is only in the case of when we are finished
@@ -403,7 +410,7 @@ public class ScannerSession extends AbstractExecutionThreadService implements It
                 flush();
                 return;
             }
-            stop();
+            stopAsync();
             
             return;
         }
@@ -665,7 +672,7 @@ public class ScannerSession extends AbstractExecutionThreadService implements It
     
     public void close() {
         forceClose = true;
-        stop();
+        stopAsync();
         synchronized (sessionDelegator) {
             if (null != delegatedResource) {
                 try {
--- a/warehouse/query-core/src/main/java/datawave/query/tables/facets/MergedReadAhead.java
+++ b/warehouse/query-core/src/main/java/datawave/query/tables/facets/MergedReadAhead.java
@@ -43,8 +43,14 @@ public class MergedReadAhead<T> extends AbstractExecutionThreadService implement
         
         buf = QueueUtils.synchronizedQueue(new CircularFifoQueue(1));
         
+        // these two guava methods replaced behavior of startAndWait() from version 15 but
+        // will now throw an exception if another thread closes the session so catch and ignore
         startAsync();
-        awaitRunning();
+        try {
+            awaitRunning();
+        } catch (IllegalStateException e) {
+            log.debug("Thread was closed while waiting to start up.");
+        }
     }
     
     /*
--- a/warehouse/query-core/src/main/java/datawave/query/tables/facets/MergedReadAhead.java
+++ b/warehouse/query-core/src/main/java/datawave/query/tables/facets/MergedReadAhead.java
@@ -43,8 +43,8 @@ public class MergedReadAhead<T> extends AbstractExecutionThreadService implement
         
         buf = QueueUtils.synchronizedQueue(new CircularFifoQueue(1));
         
-        startAndWait();
-        
+        startAsync();
+        awaitRunning();
     }
     
     /*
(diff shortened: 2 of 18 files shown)


warehouse/query-core/src/main/java/datawave/query/tables/facets/MergedReadAhead.java
```diff
@@ -43,8 +43,8 @@ public MergedReadAhead(FacetedConfiguration facetedConfig, final Iterator<T> ite
         
         buf = QueueUtils.synchronizedQueue(new CircularFifoQueue(1));
         
-        startAndWait();
-        
+        startAsync();
+        awaitRunning();
```
apmoriarty: Does this need a try-catch, per your comment in the ScannerSession?

milleruntime: I didn't see it thrown in any of the tests but I guess it would be better to have the try-catch to try and preserve the behavior.

milleruntime: Thanks for the review but my branch was old and I forgot to rebase with latest before opening the PR.

warehouse/query-core/src/main/java/datawave/query/tables/ScannerSession.java
```diff
@@ -403,7 +410,7 @@ protected void findTop() throws Exception {
                 flush();
                 return;
             }
-            stop();
+            stopAsync();
```
keith-ratcliffe: Not sure exactly what is causing this, but looking over the recent pipeline runs for this PR in Gitlab-CI, I noticed there's a sporadic  failure occurring with the test below, likely due to race condition introduced by the new `stopAsync()` behavior.
```
156768 [ERROR] Tests run: 10, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 2.666 s <<< FAILURE! - in datawave.webservice.query.runner.QueryExecutorBeanTest
156769 [ERROR] testCloseActuallyCloses(datawave.webservice.query.runner.QueryExecutorBeanTest)  Time elapsed: 0.173 s  <<< FAILURE!
java.lang.AssertionError: 
createQuery thread died before reaching initialize: java.lang.AssertionError: 
  Unexpected method call Persister.remove(datawave.webservice.query.QueryImpl@5d9ef7be[queryLogicName=EventQueryLogic,queryName=<null>,expirationDate=Tue Oct 08 18:23:15 GMT 2019,uuid=962018b3-5610-4add-ab31-4fb007a55da3,pagesize=10,pageTimeout=0,maxResultsOverride=NA,query=FOO == BAR,queryAuthorizations=PRIVATE,PUBLIC,userDN=CN=Guy Some Other soguy, OU=MY_SUBDIVISION, OU=MY_DIVISION, O=ORG, C=US,owner=<null>,parameters=[],dnList=<null>,columnVisibility=<null>,beginDate=Mon Oct 07 18:23:15 GMT 2019,endDate=Mon Oct 07 18:23:15 GMT 2019]):
    Persister.create("cn=guy some other soguy, ou=my_subdivision, ou=my_division, o=org, c=us", [cn=guy some other soguy, ou=my_subdivision, ou=my_division, o=org, c=us], PRIVATE|PUBLIC, "EventQueryLogic", datawave.webservice.query.QueryParametersImpl@d9a26113, {audit.userDn=[cn=guy some other soguy, ou=my_subdivision, ou=my_division, o=org, c=us], audit.columnVisibility=[PRIVATE|PUBLIC], audit.logicClass=[EventQueryLogic]}): expected: 1, actual: 1
    Persister.findById(<any>): expected: at least 0, actual: 0
	at datawave.webservice.query.runner.QueryExecutorBeanTest.testCloseActuallyCloses(QueryExecutorBeanTest.java:733)
```
Even though the build passes after one or more retries, it probably needs some investigation to figure out whether there's an actual issue with the code here, or if test above just needs to accommodate the new behavior. FWIW, now that you've rebased against the latest master, the deploy phase and it's query tests seem to pass consistently on the first try. Although, I'd be curious to see if Wildfly logs are seeing any new stacktraces.

milleruntime: I have yet to see that test Fail locally for me but I do see ```MaxExpansionRegexQueryTest``` fail on occasion.

keith-ratcliffe: > I have yet to see that test Fail locally for me but I do see `MaxExpansionRegexQueryTest` fail on occasion.

To follow up on this, I've recently observed the exact same failure with `testCloseActuallyCloses` on existing branches, so that issue is obviously unrelated to changes here. In light of that, can't see any reason not to approve this

milleruntime: Sounds good.  I rebased on current master and merged the conflict

brianloss:
I thought the reason we were locked to an old version of Guava is that Accumulo 1.9.3 uses Guava 14, so 15 was the highest we could use and still have code in iterators not barf due to missing methods (since iterators are stuck using the guava version provided by Accumulo).

Did something change? If not, then this change really needs to go in the Accumulo 2 branch since it will cause problems with iterators that use guava functions.

milleruntime:
> I thought the reason we were locked to an old version of Guava is that Accumulo 1.9.3 uses Guava 14, so 15 was the highest we could use and still have code in iterators not barf due to missing methods (since iterators are stuck using the guava version provided by Accumulo).
> 
> Did something change? If not, then this change really needs to go in the Accumulo 2 branch since it will cause problems with iterators that use guava functions.

You are correct, Accumulo 1.9.3 uses Guava 14 so this should go into the 2 branch.

PR #597, opened 2019-10-10
fixes #596 SourceManager properly manages more sources than initialSize

### LINKED ISSUE #596: SourceManager incorrectly manages sources
When a SourceManager re-uses a source (createdSize >= initialSize) the SourceManager incorrectly manages the underlying sources.

FineAndDandy:


warehouse/query-core/src/test/java/datawave/query/iterator/SortedListKeyValueIterator.java
```diff
@@ -38,6 +39,14 @@ public SortedListKeyValueIterator(List<Map.Entry<Key,Value>> sourceList) {
         currentIndex = 0;
     }
     
+    public SortedListKeyValueIterator(SortedMap<Key,Value> map) {
+        this.sourceList = new ArrayList<>(map.entrySet().size());
+        currentIndex = 0;
+        for (Map.Entry<Key,Value> entry : map.entrySet()) {
+            sourceList.add(entry);
+        }
```
apmoriarty: Could this be a sourceList.addAll(map.entrySet())?
Does the currentIndex indicate another possible constructor where we preserve the current index?

warehouse/query-core/src/test/java/datawave/query/iterator/SourceManagerTest.java
```diff
@@ -51,6 +55,112 @@ public void setup() throws ParseException, IOException {
...
+            copy1.next();
+            copy2.next();
+        }
+        
+        assertFalse(copy1.hasTop());
+        assertFalse(copy2.hasTop());
+        assertEquals(copy1Count, alternatingCount);
+        
+        // different ranges
+        copy1.seek(new Range(new Key("20121126_2"), true, new Key("20121126_3"), false), Collections.emptyList(), false);
```
apmoriarty: I would prefer to see this test broken out into the cases of "same range" - "different range". If a future change would only break one behavior it will be obvious up front which behavior is broken.

warehouse/query-core/src/main/java/datawave/query/iterator/SourceManager.java
```diff
@@ -178,8 +210,16 @@ public void seek(Range range, Collection<ByteSequence> columnFamilies, boolean i
             
             try {
                 lastKey = child.getTopKey();
+                lastValue = child.getTopValue();
+                rangeDone = false;
```
ivakegg: Lets move the setting of rangeDone to false to the top of the method.  This will allow us to remove a mess of sets below and guarentee we did not forget to set it on some path therein.

ivakegg: Perhaps setting the lastKey and lastValue to null at the top would also be good

warehouse/query-core/src/main/java/datawave/query/iterator/SourceManager.java
```diff
@@ -105,40 +132,48 @@ public void init(SortedKeyValueIterator<Key,Value> source, Map<String,String> op
...
             
             child.next();
             
             try {
                 lastKey = child.getTopKey();
+                lastValue = child.getTopValue();
+                
+                // need to understand when to expect no more keys for this range to differentiate from the start case
+                if (lastKey == null) {
+                    rangeDone = true;
```
ivakegg: set lastValue to null here or move the child.getTopValue() into and else clause

warehouse/query-core/src/main/java/datawave/query/iterator/SourceManager.java
```diff
@@ -178,8 +210,16 @@ public void seek(Range range, Collection<ByteSequence> columnFamilies, boolean i
             
             try {
                 lastKey = child.getTopKey();
+                lastValue = child.getTopValue();
+                rangeDone = false;
+                
+                if (lastKey == null) {
+                    rangeDone = true;
+                }
             } catch (Exception e) {
```
ivakegg: no need to catch exception....in fact this is wrong (good catch @FineAndDandy )

warehouse/query-core/src/main/java/datawave/query/iterator/SourceManager.java
FineAndDandy: lastValue should be set to null already if lastKey is set to null

FineAndDandy:
there is also no reason hasTop() shouldn't just check the lastKey != null. This may save on some reseeking

warehouse/query-core/src/test/java/datawave/query/iterator/SourceManagerTest.java
FineAndDandy: done

PR #598, opened 2019-10-10
Updates for Accumulo 2.0

milleruntime:
* Replace impl packages with new location
* Made BulkIngestMapFileLoader use AccumuloClient
* Replace DefaultConfiguration & CachedConfiguration
* Use TableId instead of string & make QueryIter implement SKVI
* Dropped DistributedTrace & ZooUtil references, dropped ExtentParser
* Dropped RFileScanner and RfileResource
* Updated in-memory-accumulo for 2.0
* Copied over dropped Tracer classes into webservice.query.runner
* Update services code and dependencies
* Many other various fixes

This should be a good start to #533.  Compiles with ```mvn -Pdev -Ddeploy -Dtar -DskipTests clean install```.  The microservices also compile under datawave/services/build-parent.


brianloss:
FYI, I am in the process of moving everything under the services folder out to separate repositories. That may cause some conflicts with this PR when I complete the move. Sorry for the inconvenience this will likely cause.

keith-ratcliffe:
Figured it was worth making a dedicated branch here for this particular WIP, so I'm going to go ahead and merge into the new accumulo-2.0 branch that I just pushed...

brianloss:
> Figured it was worth making a dedicated branch here for this particular WIP, so I'm going to go ahead and merge into the new accumulo-2.0 branch that I just pushed...

A bunch of this will likely get lost when I delete the services folder.

milleruntime:
> > Figured it was worth making a dedicated branch here for this particular WIP, so I'm going to go ahead and merge into the new accumulo-2.0 branch that I just pushed...
> 
> A bunch of this will likely get lost when I delete the services folder.

No problem.  Thanks for the heads up.

brianloss:
> 
> 
> > > Figured it was worth making a dedicated branch here for this particular WIP, so I'm going to go ahead and merge into the new accumulo-2.0 branch that I just pushed...
> > 
> > 
> > A bunch of this will likely get lost when I delete the services folder.
> 
> No problem. Thanks for the heads up.

It's a little tricky, but I'm trying to pull this change into accumulo-2.0 branches on the affected service repos by adding the main repo as a remote and cherry-picking your commit. So far that seems to be working.

services/accumulo-utils/pom.xml
```diff
@@ -30,6 +30,11 @@
                 <artifactId>authorization-api</artifactId>
                 <version>${version.authorization-api}</version>
             </dependency>
+            <dependency>
```
brianloss: This is a huge no-no. We don't want any dependencies in the microservices on the datawave query code as-is. If there is a common dependency, then it needs to be moved to one of the appropriate utils packages instead.

milleruntime: This was only done because of Span and Trace. They were dropped so I just stuck them in datawave.webservice.query.runner since that package used them the most... We shouldn't be using them at all but I don't know where they should go if we have to copy them over.

PR #600, opened 2019-10-11
Cleanup of ResourceQueue code.

drewfarris:
- final because this class shouldn't be used as an extension point
- cleaned up constructors to enable final fields
- Generics for `GenericObjectPool`
- Eliminated some redundant fields
- `ScannerQueueFactory` now a static nested class (also private final, not to be resued.)
- Removed unused Queue from ScannerQueueFactory.
- Renamed ScannerQueueFactory to AccumuloResourceFactory
- Cleaned up some unnecessary auto-generated method comments.


drewfarris:
We'll need a fix to this as it broke a downstream test.

PR #609, opened 2019-10-18
fixed #571: forced evaluation only of bad regexes

### LINKED ISSUE #571: force extremely bad regex expressions to be pushed down
There are circumstances where a regex may produce extremely large sets which may result in an ivarator (DatawaveFieldIndexCachingIterator) to take an extremely long time to run and eventually fail due to time or data size constraints.  We need to start developing mechanisms to determine this in the planning phase.  A separate ticket was written to detect this during execution, however if a query is going to fail because of this we would like to know up front.  Possible approaches may include sampling techniques coupled with a configuration of known bad regex forms.

ivakegg:
Added the ability to configure the query planner with a set of node transform rules

* Creates a rule that allows us to avoid any index lookups for regex expressions based on the form of the regex.
* Created test cases


warehouse/query-core/src/main/java/datawave/query/planner/rules/NodeTransformRule.java
```diff
@@ -0,0 +1,18 @@
+package datawave.query.planner.rules;
+
+import datawave.query.config.ShardQueryConfiguration;
+import org.apache.commons.jexl2.parser.JexlNode;
+
+public interface NodeTransformRule {
+    /**
+     * Basically every planning rule will get applied to every node. If nothing is to be modified, then simply return the passed in node. I something is to be
```
apmoriarty: If something is to be done.

warehouse/query-core/src/main/java/datawave/query/planner/rules/NodeTransformVisitor.java
```diff
@@ -0,0 +1,126 @@
+package datawave.query.planner.rules;
+
+import com.sun.tools.javac.util.Assert;
```
apmoriarty: Is this the right assert?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ParallelIndexExpansion.java
```diff
@@ -330,10 +330,9 @@ public Object visit(ASTERNode node, Object data) {
                 }
                 if (markedParents != null) {
                     for (JexlNode markedParent : markedParents) {
-                        if (ExceededValueThresholdMarkerJexlNode.instanceOf(markedParent) || ExceededOrThresholdMarkerJexlNode.instanceOf(markedParent)
-                                        || ASTDelayedPredicate.instanceOf(markedParent) || IndexHoleMarkerJexlNode.instanceOf(markedParent)
-                                        || ASTEvaluationOnly.instanceOf(node))
-                            return node;
+                        if (QueryPropertyMarker.instanceOf(markedParent, null))
+                            ;
```
hlgp: stray ; ?

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -1061,6 +1093,18 @@ protected ASTJexlScript processTree(final ASTJexlScript originalQueryTree, Shard
...
+        if (!getTransformRules().isEmpty()) {
+            stopwatch = timers.newStartedStopwatch("DefaultQueryPlanner - Apply Node Transform Rules");
+            
+            queryTree = applyNodeTransformRules(queryTree, getTransformRules(), config);
+            
+            if (log.isDebugEnabled()) {
+                logQuery(queryTree, "Query after function index queries were expanded:");
+            }
+            stopwatch.stop();
+        }
```
hlgp: same exact code block 3 times.  same timer message, same log message.  Consider making its own method?  Or at least give some indication of which iteration we're on for logging.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ParallelIndexExpansion.java
ivakegg: woops....will remove

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
ivakegg: that is fair....will do

warehouse/query-core/src/main/java/datawave/query/planner/rules/NodeTransformRule.java
ivakegg: 👍

warehouse/query-core/src/main/java/datawave/query/planner/rules/RegexPushdownTransformRule.java
```diff
@@ -0,0 +1,45 @@
...
+                    log.error("RegexPushdownTransformRule.apply: Forcing evaluation only for " + regex);
+                    return ASTEvaluationOnly.create(node);
+                }
+            }
+        }
+        return node;
+    }
+    
+    public void setRegexPatterns(List<String> patterns) {
+        this.patterns = patterns.stream().map(p -> Pattern.compile(p)).collect(Collectors.toList());
```
brianloss: No need to change, but as an FYI for using streams, this is  a case where you could use a method reference:
```java
this.patterns = patterns.stream().map(Pattern::compile).collect(Collectors.toList());
```

warehouse/query-core/src/test/java/datawave/query/planner/rules/NodeTransformVisitorTest.java
```diff
@@ -0,0 +1,185 @@
...
+        @Override
+        public JexlNode apply(JexlNode node, ShardQueryConfiguration config) {
+            if (node instanceof ASTAndNode) {
+                // reverse the children
+                ArrayList<JexlNode> children = newArrayList();
+                for (JexlNode child : children(node)) {
+                    children.add(child);
+                }
+                Collections.reverse(children);
+                return children(node, children.toArray(new JexlNode[children.size()]));
```
brianloss: Doesn't matter since this is a test, but the current guidance I've seen is that creating an array of size 0 is actually faster (and safer--no race condition) that making an array of the list size. That is, this is preferred:
```java
return children(node, children.toArray(new JexlNode[0]));
```
[Here's](https://shipilev.net/blog/2016/arrays-wisdom-ancients/) a good article on the subject.

warehouse/query-core/src/main/java/datawave/query/planner/rules/RegexPushdownTransformRule.java
ivakegg: thanks

warehouse/query-core/src/main/java/datawave/query/planner/rules/NodeTransformVisitor.java
```diff
@@ -0,0 +1,125 @@
...
+    @Override
+    public Object visit(ASTFunctionNode node, Object data) {
+        return applyTransforms(super.visit(node, data));
+    }
+    
+    @Override
+    public Object visit(ASTReference node, Object data) {
+        // do not recurse on a marker node
+        if (QueryPropertyMarker.instanceOf(node, null)) {
+            return applyTransforms(node);
```
jwomeara: Instead of applying the transform on the node that was passed in, we should instead apply the transforms to a copy of that node (using the base RebuildingVisitor), right?  That way we don't affect the original tree that was passed in.

jwomeara: Also, should recursing on a marked node be configurable in case we someday have a transform which we expect to run against a marked node?

warehouse/query-core/src/main/java/datawave/query/planner/rules/RegexPushdownTransformRule.java
```diff
@@ -0,0 +1,45 @@
...
+    public JexlNode apply(JexlNode node, ShardQueryConfiguration config) {
+        if (node instanceof ASTERNode) {
+            final String regex = String.valueOf(JexlASTHelper.getLiteralValue(node));
+            if (patterns.stream().anyMatch(p -> p.matcher(regex).matches())) {
+                if (JexlASTHelper.getIdentifierNames(node).contains(Constants.ANY_FIELD)) {
+                    log.error("RegexPushdownTransformRule.apply: Not allowing _ANYFIELD =~ " + regex);
+                    throw new DatawaveFatalQueryException("Not allowing _ANYFIELD =~ \" + regex");
+                } else {
+                    log.error("RegexPushdownTransformRule.apply: Forcing evaluation only for " + regex);
+                    return ASTEvaluationOnly.create(node);
```
FineAndDandy: does this need to check for index only?

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -1570,6 +1587,25 @@ public ASTJexlScript applyRules(ASTJexlScript queryTree, ScannerFactory scannerF
         return pushDownPlanner.applyRules(queryTree);
     }
     
+    /*
+     * Apply the configured node transforms
+     */
+    public ASTJexlScript applyNodeTransformRules(ASTJexlScript queryTree, List<NodeTransformRule> rules, ShardQueryConfiguration config, String instance) {
+        
+        if (!getTransformRules().isEmpty()) {
```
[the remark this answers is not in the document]
ivakegg: doh, ok

warehouse/query-core/src/main/java/datawave/query/planner/rules/NodeTransformVisitor.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.List;
+
+public class NodeTransformVisitor extends RebuildingVisitor {
+    
+    private final ShardQueryConfiguration config;
+    private final List<NodeTransformRule> rules;
+    
+    public NodeTransformVisitor(ShardQueryConfiguration config, List<NodeTransformRule> rules) {
+        assert (rules != null);
+        assert (config != null);
```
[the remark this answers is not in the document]
ivakegg: true.  converting to runtime checks

warehouse/query-core/src/main/java/datawave/query/planner/rules/NodeTransformVisitor.java
ivakegg: Will do re: copy of node.  Interesting idea on the marked node.  I do not believe that is what is wanted in this case.  I think a marked node needs to always be handled as a single structure.  Applying a node transform outside the context of the marking is asking for trouble.

warehouse/query-core/src/main/java/datawave/query/planner/rules/RegexPushdownTransformRule.java
ivakegg: The pull up code will do that for us, however it is reasonable to do that check here.  I will work that.  Also I think its a "non-event" check instead of an index only check.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RebuildingVisitor.java
```diff
@@ -9,6 +9,7 @@
 
 import datawave.query.exceptions.DatawaveFatalQueryException;
 import datawave.query.jexl.JexlNodeFactory;
+import datawave.query.jexl.nodes.QueryPropertyMarker;
```
FineAndDandy: imported but never used

PR #611, opened 2019-10-21
Add GitHub action workflow to run tests.

brianloss:
We have access to GitHub actions which might be a viable CI replacement for the private GitLab repo that we are currently using.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,68 @@
+name: Tests
+
+on:
+  push:
+    paths-ignore: ['*.md', 'CODEOWNERS', 'LICENSE']
+    branches:
+    - 'master'
+    - 'release/version*'
+  pull_request:
+    paths-ignore: ['*.md', 'CODEOWNERS', 'LICENSE']
+
+jobs:
+  build-and-test:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -0,0 +1,85 @@
...
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    # This is a hack to pre-populate the maven repo and reduce downloads. This can
+    # go away whenever GitHub actions support caching (supposedly in November 2019).
+    - name: Set up maven repo
+      run: |
+        echo ${{ secrets.GITHUB_TOKEN }} | docker login docker.pkg.github.com --username ${GITHUB_ACTOR} --password-stdin
+        mkdir ~/.m2 || true
+        docker run --rm -u $(id -u):$(id -g) -v ~/.m2:/m2 docker.pkg.github.com/brianloss/maven_repo/repo:v1
```
brianloss: This is a work-around until Github Actions supports caching (which they claim will happen in November). This is a docker image that has a pre-populated maven repo and we just copy it to the worker users `~/.m2` dir. This takes about 1'15" or so, but the total build times in the few tests I've done tend to be about 7 minutes faster.

PR #620, opened 2019-10-24
Added a generic constructor for flagMaker to allow for extension

jkrdev:


warehouse/ingest-core/src/main/java/datawave/util/flag/FlagMaker.java
```diff
@@ -130,12 +131,23 @@ public static void main(String... args) throws Exception {
...
         
     }
-    
+
+    private static FlagMaker createFlagMaker(FlagMakerConfig fc) {
+        try {
+            Class<? extends FlagMaker> c = (Class<? extends FlagMaker>) Class.forName(fc.getFlagMakerClass());
+            Constructor<? extends FlagMaker> constructor = c.getConstructor(FlagMakerConfig.class);
+            return constructor.newInstance(fc);
+        } catch (Exception e) {
```
jschmidt10: We might add an explicit catch for `NoSuchMethodException` so we can give a friendlier error message. We have basically created a mandatory, undocumented required constructor. Maybe add something like: `catch (NoSuchMethodException e) { throw new RuntimeException("Subclasses of FlagMaker must implement a constructor that takes a FlagMakerConfig", e); }` just for clarity?

jkrdev: Added in latest commit.

PR #621, opened 2019-10-24
Added JexlNodeset, a utility class that mimics Set operations for a c…

apmoriarty:
…ollection of Jexl nodes.

We duplicate a lot of effort to ensure uniqueness among collections of Jexl nodes. This class is an attempt to cut down on that duplicated effort.

Any and all suggestions welcome.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
+++ b/warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
@@ -1,11 +1,11 @@
 package datawave.query.language.parser.jexl;
 
-import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import datawave.query.jexl.JexlASTHelper;
 import org.apache.commons.jexl2.parser.JexlNode;
 
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -30,7 +30,7 @@ public class JexlNodeSet implements Set<JexlNode> {
     
     /**
      * Method to access all the Jexl nodes in the set.
-     * 
+     *
      * @return - the underlying node map values.
      */
     public Collection<JexlNode> getNodes() {
@@ -67,7 +67,7 @@ public class JexlNodeSet implements Set<JexlNode> {
     
     @Override
     public Iterator<JexlNode> iterator() {
-        return Lists.newArrayList(nodeMap.values()).listIterator();
+        return Collections.unmodifiableCollection(nodeMap.values()).iterator();
     }
     
     @Override
@@ -77,7 +77,7 @@ public class JexlNodeSet implements Set<JexlNode> {
     
     @Override
     public <T> T[] toArray(T[] ts) {
-        throw new UnsupportedOperationException("JexlNodeSet does not support toArray() operations.");
+        throw new UnsupportedOperationException("JexlNodeSet does not support toArray() calls to pre-allocated arrays.");
     }
     
     @Override
@@ -91,6 +91,9 @@ public class JexlNodeSet implements Set<JexlNode> {
     }
     
     /**
+     * Remove the specified object from the set if it is a Jexl node.
+     *
+     * Builds a node key for the Jexl node and delegates to {@link #remove(String, Object)}
      *
      * @param o
      *            - object to be removed from the set of Jexl nodes.
--- a/warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
+++ b/warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
@@ -1,0 +1,185 @@
+package datawave.query.language.parser.jexl;
+
+import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
+import datawave.query.jexl.JexlASTHelper;
+import org.apache.commons.jexl2.parser.JexlNode;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Utility class that implements the {@link Set} interface for use with a collection of Jexl nodes.
+ *
+ * Element uniqueness is computed using {@link JexlASTHelper#nodeToKey(JexlNode)} for keys.
+ *
+ * Important Note: The set is considered invalid if any of the underlying Jexl nodes change.
+ */
+public class JexlNodeSet implements Set<JexlNode> {
+    
+    // Internal map of node keys to nodes;
+    private final Map<String,JexlNode> nodeMap;
+    
+    public JexlNodeSet() {
+        this.nodeMap = new HashMap<>();
+    }
+    
+    /**
+     * Method to access all the Jexl nodes in the set.
+     * 
+     * @return - the underlying node map values.
+     */
+    public Collection<JexlNode> getNodes() {
+        return nodeMap.values();
+    }
+    
+    /**
+     * Get the set of generated node keys.
+     * 
+     * @return - the underlying node map keySet.
+     */
+    public Set<String> getNodeKeys() {
+        return nodeMap.keySet();
+    }
+    
+    @Override
+    public int size() {
+        return nodeMap.size();
+    }
+    
+    @Override
+    public boolean isEmpty() {
+        return nodeMap.isEmpty();
+    }
+    
+    @Override
+    public boolean contains(Object o) {
+        if (o instanceof JexlNode) {
+            String nodeKey = JexlASTHelper.nodeToKey((JexlNode) o);
+            return nodeMap.containsKey(nodeKey);
+        }
+        return false;
+    }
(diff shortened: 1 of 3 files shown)


warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
```diff
@@ -0,0 +1,185 @@
...
+    
+    @Override
+    public boolean isEmpty() {
+        return nodeMap.isEmpty();
+    }
+    
+    @Override
+    public boolean contains(Object o) {
+        if (o instanceof JexlNode) {
+            String nodeKey = JexlASTHelper.nodeToKey((JexlNode) o);
```
brianloss: This doesn't build--looks like this method was removed [here](https://github.com/NationalSecurityAgency/datawave/commit/d65a509315f42fda7cd9b6f989f8fca4bc348351#diff-dd6995b5d33081727484b35ba74f8b0bL1253).

apmoriarty: Updated to add that method back in.

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
```diff
@@ -0,0 +1,185 @@
...
+            return nodeMap.containsKey(nodeKey);
+        }
+        return false;
+    }
+    
+    @Override
+    public Iterator<JexlNode> iterator() {
+        return Lists.newArrayList(nodeMap.values()).listIterator();
```
jschmidt10: We're creating a new list each time this is invoked. If this is for defense against modification, we could do `Collections.unmodiableCollection(nodeMap.values()).iterator()`

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+    
+    @Override
+    public Object[] toArray() {
+        return nodeMap.entrySet().toArray();
+    }
+    
+    @Override
+    public <T> T[] toArray(T[] ts) {
+        throw new UnsupportedOperationException("JexlNodeSet does not support toArray() operations.");
```
jschmidt10: This comment is inconsistent. It'd make sense if we threw UnsupportedOperationException in the above toArray() as well.

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
```diff
@@ -0,0 +1,185 @@
...
+        String nodeKey = JexlASTHelper.nodeToKey(node);
+        if (!nodeMap.containsKey(nodeKey)) {
+            nodeMap.put(nodeKey, node);
+            return true;
+        }
+        return false;
+    }
+    
+    /**
+     *
```
jschmidt10: Missing a description.

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
apmoriarty: That's the intent, I'll use your suggestion

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
apmoriarty: I'll update this to be more descriptive

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
```diff
@@ -0,0 +1,185 @@
...
+import java.util.Set;
+
+/**
+ * Utility class that implements the {@link Set} interface for use with a collection of Jexl nodes.
+ *
+ * Element uniqueness is computed using {@link JexlASTHelper#nodeToKey(JexlNode)} for keys.
+ *
+ * Important Note: The set is considered invalid if any of the underlying Jexl nodes change.
+ */
+public class JexlNodeSet implements Set<JexlNode> {
```
jschmidt10: We could achieve the same functionality here if we just overrode `equals()` and `hashCode()` in `JexlNode`. Then we could use any of the Java collections we wanted (like HashSet for this case). It'd be less code and also enable us to switch collection implementations easier if we ever needed to.

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
apmoriarty: Will update

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
```diff
@@ -0,0 +1,185 @@
...
+    public boolean retainAll(Collection<?> collection) {
+        Set<String> retainKeys = new HashSet<>();
+        if (collection != null) {
+            for (Object o : collection) {
+                retainKeys.add(JexlASTHelper.nodeToKey((JexlNode) o));
+            }
+        }
+        
+        boolean modified = false;
+        for (String key : Sets.newHashSet(nodeMap.keySet())) {
```
jschmidt10: Seems unnecessary to create an extra HashSet here if we are only iterating over it.

warehouse/query-core/src/test/java/datawave/query/language/parser/jexl/JexlNodeSetTest.java
```diff
@@ -0,0 +1,266 @@
...
+public class JexlNodeSetTest {
+    
+    @Test
+    public void testTheReasonForThisClass() {
+        JexlNode node1 = JexlNodeFactory.buildEQNode("FOO", "bar");
+        JexlNode node2 = JexlNodeFactory.buildEQNode("FOO", "bar");
+        
+        Set<JexlNode> nodeSet = Sets.newHashSet(node1, node2);
+        assertEquals(2, nodeSet.size());
+        assertTrue(nodeSet.contains(node1));
```
jschmidt10: Again, not to harp on it, but we are getting this behavior b/c of the default implementation of `equals()` and `hashCode()`. This test would fail if we overrode those to use your `nodeToKey` function.

warehouse/query-core/src/test/java/datawave/query/language/parser/jexl/JexlNodeSetTest.java
```diff
@@ -0,0 +1,266 @@
...
+        assertTrue(nodeSet.contains(node2));
+        
+        assertTrue(nodeSet.add(node3));
+        assertFalse(nodeSet.add(node3));
+        assertFalse(nodeSet.add(node3));
+        
+        assertEquals(2, nodeSet.size());
+        assertTrue(nodeSet.contains(node1));
+        assertTrue(nodeSet.contains(node2));
```
jschmidt10: We never assert that `nodeSet.contains(node3)`.

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
apmoriarty: We could, and that's probably the best long term solution. However, we're working with apache jexl nodes, so we would have to either submit a PR against that project or override the class.

The danger we run into is when the JexlNode represents a large tree and it becomes expensive to compute the node key. With this implementation of JexlNodeSet, we can limit where we perform this potentially expensive operation (i.e., only pay the cost when we need it vs. everywhere all the time).

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
apmoriarty: Nope, I'll remove that.

warehouse/query-core/src/test/java/datawave/query/language/parser/jexl/JexlNodeSetTest.java
```diff
@@ -0,0 +1,266 @@
...
+        
+        JexlNodeSet nodeSet = new JexlNodeSet();
+        nodeSet.addAll(nodes);
+        
+        assertTrue(nodeSet.contains(node1));
+        assertTrue(nodeSet.contains(node2));
+        assertTrue(nodeSet.contains(node3));
+        
+        nodeSet.clear();
+        assertFalse(nodeSet.contains(node1));
```
jschmidt10: assert isEmpty()

apmoriarty: Got it

warehouse/query-core/src/test/java/datawave/query/language/parser/jexl/JexlNodeSetTest.java
apmoriarty: Will add

PR #623, opened 2019-10-24
fixes #622 Account for cache layers when moving NestedIterator

### LINKED ISSUE #622: NestedIterator doesn't handle move() properly

FineAndDandy:


warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -463,6 +465,15 @@ public void seek(Range r, Collection<ByteSequence> columnFamilies, boolean inclu
             // the start of this range is beyond the end of the last range seeked
             // we must reset keyValues to null and empty the underlying collection
             clearRowBasedHdfsBackedSet();
+        } else {
+            // inside the original range, so potentially need to reposition keyValues
+            if (keyValues != null) {
+                Key startKey = r.getStartKey();
+                // decide if keyValues needs to be rebuilt or can be reused
+                if (!keyValues.hasNext() || (keyValues.peek().getKey().compareTo(startKey) > 0)) {
```
ivakegg: Do we need to peek here or can we just use topKey?

ivakegg: Should we consider whether startInclusive is true or not?

warehouse/query-core/src/main/java/datawave/query/iterator/logic/AndIterator.java
```diff
@@ -126,29 +131,43 @@ public boolean hasNext() {
...
+     * @throws IllegalStateException
+     *             if prev >= minimum
+     */
     public T move(T minimum) {
         if (null == includeHeads) {
             throw new IllegalStateException("initialize() was never called");
         }
         
-        Set<T> headSet = includeHeads.keySet().headSet(minimum);
+        if (prev != null && prev.compareTo(minimum) >= 0) {
```
ivakegg: If we are not allowed to move backwards, then why all of that work in the ivarator iterator?

warehouse/query-core/src/main/java/datawave/query/iterator/DocumentIterator.java
```diff
@@ -15,6 +15,13 @@
...
     Document document();
     
-    // move forward
+    /**
+     * Move the iterator forward to the first Key >= pointer
+     *
+     * @param pointer
+     * @throws IOException
+     * @throws IllegalStateException
+     *             if getTopKey() >= pointer
```
brianloss: The build is failing due to this invalid javadoc.

warehouse/query-core/src/main/java/datawave/query/iterator/NestedIterator.java
```diff
@@ -19,7 +19,9 @@
      * Tells the underlying iterator to return the first element that is greater than or equal to <code>minimum</code>.
      * 
      * @param minimum
-     * @return
+     * @return the first Key in the iterator >= minimum or null if no Key exists
```
brianloss: The build is failing due to this invalid javadoc.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/AndIterator.java
```diff
@@ -126,29 +131,43 @@ public boolean hasNext() {
...
     }
     
+    /**
+     * Test all layers of cache for the minimum, then if necessary advance heads
+     * 
+     * @param minimum
+     *            the minimum to return
+     * @return the first >= minimum or null if none exists
+     * @throws IllegalStateException
+     *             if prev >= minimum
```
brianloss: The build is failing due to this invalid javadoc.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/IndexIterator.java
```diff
@@ -369,6 +369,15 @@ public void seek(Range range, Collection<ByteSequence> columnFamilies, boolean i
...
     private final Text newColumnQualifier = new Text(new byte[128]);
     
+    /**
+     * Advance the source to the Key specified by pointer then fetch the next tk/tv from that point
+     * 
+     * @param pointer
+     *            the minimum point to advance source to
+     * @throws IOException
+     * @throws IllegalStateException
+     *             if getTopKey() >= pointer
```
brianloss: The build is failing due to this invalid javadoc.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/IndexIteratorBridge.java
```diff
@@ -55,55 +60,58 @@ public Key next() {
...
     
+    /**
+     * Advance to the next Key in the iterator that is >= minimum. First check the cached value in next, then check the delegate cached value in getTopValue(),
+     * finally advance the delegate
+     * 
+     * @param minimum
+     *            the minimum key to advance to
+     * @return the first Key >= minimum found
+     * @throws IllegalStateException
+     *             if prevKey >= minimum
```
brianloss: The build is failing due to this invalid javadoc.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/OrIterator.java
```diff
@@ -115,36 +122,51 @@ public T next() {
...
     }
     
+    /**
+     * Test all layers of cache for the minimum, then if necessary advance heads
+     * 
+     * @param minimum
+     *            the minimum to return
+     * @return the first >= minimum or null if none exists
+     * @throws IllegalStateException
+     *             if prev >= minimum
```
brianloss: The build is failing due to this invalid javadoc.

warehouse/query-core/src/main/java/datawave/query/iterator/CachingIterator.java
```diff
@@ -0,0 +1,46 @@
...
+        } else {
+            next = null;
+        }
+        
+        return toReturn;
+    }
+    
+    @Override
+    public void remove() {
+        // no-op
```
brianloss: Is there a reason you wouldn't delegate this method to the delegate iterator? Looks like PeekingIterator will throw an UnsupportedOperationException here, so we've turned that into a no-op. Is that desired?

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
FineAndDandy: down below in findTop() next() is called at least once so we have to check the peek value instead of the curren tvalue

FineAndDandy: since we are checking for having to rebuild the iterator we only have to be concerned if the peek Key is > the start key, since in findTop() if we are under by one Key it will be automatically advanced until it fits the criteria

warehouse/query-core/src/main/java/datawave/query/iterator/logic/AndIterator.java
FineAndDandy: The work in the ivarator is to ensure it properly adheres to the seek() contract. It should never seek backwards from this point since move is forward only.

Its possible that means it never matters with the fix below

warehouse/query-core/src/main/java/datawave/query/iterator/CachingIterator.java
ivakegg: good point

PR #628, opened 2019-10-29
Fixes #627: Double escape before a Lucene wildcard can cause incorrec…

### LINKED ISSUE #627: Double escape before a Lucene wildcard can cause incorrect translation to JEXL
WildcardFieldedTerm.getFirstWildcardIndex was using a single character lookback when encountering a wildcard character to determine if it was escaped.  This is incorrect when the escape itself is escaped.

When the only (or all) wildcards in a term were masked in this way, the translation determined that there was no unescaped wildcard in the term and created an EXACT SelectorNode instead of a WILDCARDS SelectorNode.

billoley:
WildcardFieldedTerm.getFirstWildcardIndex was using a single character lookback when encountering a wildcard character to determine if it was escaped. This is incorrect when the escape itself is escaped.

When the only (or all) wildcards in a term were masked in this way, the translation determined that there was no unescaped wildcard in the term and created an EXACT SelectorNode instead of a WILDCARDS SelectorNode.


brianloss:
LGTM, but it looks like there are conflicts. Can you re-base off release/version2.6 and fix the conflicts?

Issue #635, opened 2019-10-31
IndexInfo should use JexlNodeSet


apmoriarty:
The IndexInfo object should delegate responsibility of Jexl node uniqueness to the JexlNodeSet.

This will cover modifying the JexlNodeSet to prefer the ASTDelayedPredicate version of source nodes.

apmoriarty:
Merged in https://github.com/NationalSecurityAgency/datawave/commit/65300a82a09012d878f3589fa1d0af01d12b530c

PR #636, opened 2019-10-31
Feature/index info uses jexl node set

apmoriarty:
Addresses Issue #635


warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
```diff
@@ -143,75 +137,44 @@ public IndexInfo union(IndexInfo o) {
...
         merged.count = -1;
         merged.uids = ImmutableSortedSet.of();
-        if (internalNodeList.isEmpty()) {
-            
+        if (nodeSet.isEmpty()) {
             merged.myNode = null;
         } else {
-            
-            merged.myNode = JexlNodeFactory.createUnwrappedOrNode(internalNodeList);
+            merged.myNode = JexlNodeFactory.createUnwrappedOrNode(nodeSet.getNodes());
```
jwomeara: This call will reparent the nodes returned by nodeSet.getNodes().  Are we sure we want to do that, or should we create copies of these nodes, so that we don't change the lineage of the original tree?

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
```diff
@@ -382,15 +297,14 @@ public IndexInfo union(IndexInfo o, List<JexlNode> delayedNodes) {
...
                 log.trace("internalNodeList node  is " + node);
             }
         }
-        if (internalNodeList.isEmpty()) {
+        if (nodeSet.isEmpty()) {
             merged.myNode = null;
         } else {
-            merged.myNode = TreeFlatteningRebuildingVisitor.flatten(JexlNodeFactory.createUnwrappedOrNode(internalNodeList));
-            
+            merged.myNode = TreeFlatteningRebuildingVisitor.flatten(JexlNodeFactory.createUnwrappedOrNode(nodeSet.getNodes()));
```
jwomeara: Same comment here as above w.r.t. reparenting nodes.

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
```diff
@@ -499,17 +416,19 @@ protected IndexInfo intersect(long maxPossibilities, Iterable<IndexMatch> matchI
             // NoData - B == 'c'
             // IndexMatch - C == infinite
             // the merge nodes matches now just represent (A && C)
-            if (matchNodes.values().size() > 1) {
-                newNode = TreeFlatteningRebuildingVisitor.flatten(JexlNodeFactory.createOrNode(matchNodes.values()));
+            if (nodeSet.size() > 1) {
+                newNode = TreeFlatteningRebuildingVisitor.flatten(JexlNodeFactory.createOrNode(nodeSet.getNodes()));
```
jwomeara: Same comment here w.r.t. reparenting nodes.

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
```diff
@@ -499,17 +416,19 @@ protected IndexInfo intersect(long maxPossibilities, Iterable<IndexMatch> matchI
...
             newNode = TreeFlatteningRebuildingVisitor.flatten(matches.iterator().next().getNode());
         } else {
-            List<JexlNode> allNodes = Lists.newArrayList(infiniteNodes);
-            allNodes.add(matchNode);
-            newNode = TreeFlatteningRebuildingVisitor.flatten(JexlNodeFactory.createAndNode(allNodes));
+            JexlNodeSet nodeSet = new JexlNodeSet();
+            nodeSet.addAll(infiniteNodeSet.getNodes());
+            nodeSet.add(matchNode);
+            
+            newNode = TreeFlatteningRebuildingVisitor.flatten(JexlNodeFactory.createAndNode(nodeSet.getNodes()));
```
jwomeara: Same comment here w.r.t. reparenting nodes.

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
```diff
@@ -622,12 +538,13 @@ public IndexInfo intersect(IndexInfo o, List<JexlNode> delayedNodes, UidIntersec
...
         if (null != myNode && myNode != o.myNode)
-            internalNodeList.add(myNode);
+            nodeSet.add(myNode);
         if (null != o.myNode)
-            internalNodeList.add(o.myNode);
-        internalNodeList.addAll(delayedNodes);
-        merged.myNode = TreeFlatteningRebuildingVisitor.flatten(JexlNodeFactory.createAndNode(internalNodeList));
+            nodeSet.add(o.myNode);
+        nodeSet.addAll(delayedNodes);
+        merged.myNode = TreeFlatteningRebuildingVisitor.flatten(JexlNodeFactory.createAndNode(nodeSet.getNodes()));
```
jwomeara: Same comment here w.r.t. reparenting nodes.

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
```diff
@@ -12,26 +17,32 @@
...
 
+import static datawave.query.jexl.JexlASTHelper.nodeToKey;
+
 /**
  * Utility class that implements the {@link Set} interface for use with a collection of Jexl nodes.
  *
  * Element uniqueness is computed using {@link JexlASTHelper#nodeToKey(JexlNode)} for keys.
  *
- * Important Note: The set is considered invalid if any of the underlying Jexl nodes change.
+ * If a Jexl node and it's {@link ASTDelayedPredicate} equivalent are both added to the set, the set will accept the delayed predicate.
```
jwomeara: In what cases will the set unwrap the marked node, and instead add the source node?  Are ASTDelayed nodes the only type which are handled this way?  Some more info here would be useful.

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
```diff
@@ -82,7 +93,42 @@ public boolean contains(Object o) {
...
+            }
+            // Add the delayed node with a node key of just the source node.
+            return add(nodeToKey(sourceNode), node);
+        } else {
+            return add(nodeToKey(node), node);
+        }
+    }
+    
+    /**
+     * You had better know what you're doing if you use this method.
```
FineAndDandy: outline why instead of just a warning

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
```diff
@@ -82,7 +93,42 @@ public boolean contains(Object o) {
...
+                    log.trace("Source of delayed node already exists in node set.");
+                }
+                if (remove(sourceNode)) {
+                    if (log.isTraceEnabled()) {
+                        log.trace("Removing source node " + sourceNode + " from node set prior to adding delayed predicate: " + node);
+                    }
+                }
+            }
+            // Add the delayed node with a node key of just the source node.
+            return add(nodeToKey(sourceNode), node);
```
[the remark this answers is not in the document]
jwomeara: Per our conversation, I understand a little better now.  I think it's safe to remove the contains & remove logic though.

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
apmoriarty: This preserves the original implementation, but I like the idea of copying the nodes. I can experiment with that.

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
jwomeara: +1

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
```diff
@@ -96,28 +142,33 @@ public boolean add(JexlNode node) {
...
      * @param nodeKey
-     *            - key for the object to be removed.
+     *            key for the object to be removed.
      * @param o
-     *            - the object to be removed.
-     * @return - True, if the set contained the specified element.
+     *            the object to be removed.
+     * @return True, if the set contained the specified element.
      */
     private boolean remove(String nodeKey, Object o) {
```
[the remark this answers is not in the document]
apmoriarty: Good catch, this was leftover from an earlier iteration

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
```diff
@@ -82,7 +93,42 @@ public boolean contains(Object o) {
     
     @Override
     public boolean add(JexlNode node) {
-        String nodeKey = JexlASTHelper.nodeToKey(node);
+        if (isDelayed(node)) {
+            if (log.isTraceEnabled()) {
+                log.trace("Trying to add a delayed node: " + nodeToKey(node));
+            }
+            JexlNode sourceNode = getSourceNode(node);
+            if (contains(sourceNode)) {
```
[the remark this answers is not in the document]
apmoriarty: Makes sense

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
```diff
@@ -686,11 +606,4 @@ private boolean isInfinite() {
     public void setNode(JexlNode currNode) {
         myNode = currNode;
     }
-    
-    private String nodeToKey(JexlNode node) {
-        // Note: This method assumes that the node passed in is already flattened.
-        // If not, and our tree contains functionally equivalent subtrees, we would
-        // be duplicating some of our efforts which is bad, m'kay?
```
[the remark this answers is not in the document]
apmoriarty: Can do

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
```diff
@@ -185,4 +236,46 @@ public boolean removeAll(Collection<?> collection) {
...
+    /**
+     * Copied from IndexInfo so that JexlNodeSet can handle adding delayed predicates.
+     *
+     * (FOO == 'bar') and (DELAYED &amp;&amp; FOO == 'bar') are logically equivalent.
+     *
+     * @param node
+     *            a Jexl node.
+     * @return the source Jexl node if this node is delayed.
+     */
+    private JexlNode getSourceNode(JexlNode node) {
```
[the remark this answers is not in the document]
apmoriarty: Looks like this works just great.

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
```diff
@@ -185,4 +236,46 @@ public boolean removeAll(Collection<?> collection) {
...
+        } else if (IndexHoleMarkerJexlNode.instanceOf(node)) {
+            
+            return IndexHoleMarkerJexlNode.getIndexHoleSource(node);
+        } else {
+            return node;
+        }
+    }
+    
+    // Is a node delayed?
+    protected boolean isDelayed(JexlNode node) {
```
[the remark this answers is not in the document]
apmoriarty: This works out just fine too.

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
```diff
@@ -22,25 +18,39 @@
...
+    // Determines if we unwrap QueryPropertyMarker nodes and use their source node to determine node uniqueness.
+    private boolean useSourceNodeForKeys = false;
+    
     // Internal map of node keys to nodes;
     private final Map<String,JexlNode> nodeMap;
     
-    private static final Logger log = Logger.getLogger(JexlNodeSet.class);
-    
     public JexlNodeSet() {
+        this(false);
```
jwomeara: I hate to ask this, but why not make the default 'true' since that seems to be how this is used everywhere?

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
```diff
@@ -93,84 +103,45 @@ public boolean contains(Object o) {
...
+            // If the node mapped to our node key is delayed, do not overwrite.
+            if (isDelayed(nodeMap.get(nodeKey))) {
+                return false;
+            } else {
+                
+                // If we are not delayed but a delayed version of our node already exists, do not add.
+                if (!isDelayed(node)) {
+                    return false;
                 }
             }
```
jwomeara: This could be simplified to a single if statement.

jwomeara: Not sure why this is considered 'outdated'...

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
apmoriarty: In the index lookup code node equality is synonymous with unwrapping QueryPropertyMarker nodes. If you feel this should be the default behavior everywhere I'm comfortable making that change.

Someone using this class for the first time might reasonably assume that an EQNode and the wrapped version of that EQNode are separate and distinct nodes.

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlNodeSet.java
apmoriarty: Can do. Weird that it shows as outdated...it's definitely not..

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
ivakegg: The original logic was only sticking only putting a delayed node or one of the other nodes into merge.myNode.  What is the effect of including all of the nodes here?

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
ivakegg: Same comment as above w.r.t. including the entire set of nodes here.

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
ivakegg: Same comment here w.r.t. include all nodes.

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
apmoriarty: The original logic is still preserved, although it has been pushed down into the JexlNodeSet. Node keys for delayed nodes were (and still are) built from source nodes. If the JexlNodeSet encounters a delayed version of a source node, it will retain the delayed node over the source node.

The effect of calling nodeSet.getNodes() is the same as creating the unwrapped ORNode from the internal node list.

jwomeara: To me, it looks like the original logic checked the list of nodes stored for a given key to see if any of them were delayed.  If a delayed version was found, that version would be added to the internalNodeList.  If none of them were delayed, it would just add the first node for that key.  With the new logic, we are no longer dealing with a single value map, not a multimap.  Because of that you added logic which will ensure that if a marked node is found for a given key, that node will be the one saved in the map.  So, unless I'm missing something, this new implementation looks like it matches the previous implementation.

Issue #637, opened 2019-10-31
periodic test failures


ivakegg:
This is to track test failures that we see periodically so that we collaborate on fixing them:

ivakegg:
```
[ERROR] Tests run: 20, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 2.426 s <<< FAILURE! - in datawave.query.FilterFieldsQueryTest
[ERROR] testEqCityAndEqContinentBlackListWithHitList(datawave.query.FilterFieldsQueryTest)  Time elapsed: 0.093 s  <<< FAILURE!
java.lang.AssertionError: missing return field(CITY)
        at datawave.query.FilterFieldsQueryTest.runTest(FilterFieldsQueryTest.java:380)
        at datawave.query.FilterFieldsQueryTest.runTest(FilterFieldsQueryTest.java:325)
        at datawave.query.FilterFieldsQueryTest.testEqCityAndEqContinentBlackListWithHitList(FilterFieldsQueryTest.java:271)

[ERROR] testEqCityAndEqContinentBlackList(datawave.query.FilterFieldsQueryTest)  Time elapsed: 0.074 s  <<< FAILURE!
java.lang.AssertionError: missing return field(CODE)
        at datawave.query.FilterFieldsQueryTest.runTest(FilterFieldsQueryTest.java:380)
        at datawave.query.FilterFieldsQueryTest.runTest(FilterFieldsQueryTest.java:325)
        at datawave.query.FilterFieldsQueryTest.testEqCityAndEqContinentBlackList(FilterFieldsQueryTest.java:236)

[ERROR] testBlackListMultiValueIncluded(datawave.query.FilterFieldsQueryTest)  Time elapsed: 0.066 s  <<< FAILURE!
java.lang.AssertionError: missing return field(COUNTRY)
        at datawave.query.FilterFieldsQueryTest.runTest(FilterFieldsQueryTest.java:380)
        at datawave.query.FilterFieldsQueryTest.runTest(FilterFieldsQueryTest.java:330)
        at datawave.query.FilterFieldsQueryTest.testBlackListMultiValueIncluded(FilterFieldsQueryTest.java:207)
```

ivakegg:
```
[...]

Issue #156, opened 2018-08-02
(ExceededValueThresholdMarkerJexlNode = true needs to check value for maxValueExpansionThreshold


bmwmaestoso:
The following code extracted from a test case while correct causes an exception to occur if the maxValueExpansionThreshold is set to 1, There should be a check in the code base to ensure the value of maxValueExpansionThreshold to ensure that it is > 1. In this case, _ANYFIELD_ =~ 'paris' resolves the ERNODE into a single EQNODE by the DefaultQueryPlanner, which causes the query to fail (DatawaveFatalQueryException throw from IteratorBuildVisitor). If the value is set > 1, the query works correctly.

String query = "((ExceededValueThresholdMarkerJexlNode = true) and _ANYFIELD_ =~ 'paris') && COUNTRY =~ 'un.*";
this.logic.setMaxValueExpansionThreshold(1);
ivaratorConfig();  // configures the ivarator cache
runTest(query, expect);

jwomeara:
Setting maxValueExpansionThreshold to 1 is used from time to time in our tests to force the ivarator to run; see [CompositeIndexTest](https://github.com/NationalSecurityAgency/datawave/blob/master/warehouse/query-core/src/test/java/datawave/query/planner/CompositeIndexTest.java#L523).

I don't have all of the context for this test that is failing, but it seems to me that we would want to make it work with any positive value for maxValueExpansionThreshold.

bmwmaestoso:
This may also be of interest. Setting the maxValueExpansionThreshold to 1, without configuring the ivarator causes an exception occurs but this exception is swallowed and the query does not return any results. Seems like this is a misleading response.

SethSmucker:
Fixed by #638. ivarators can run against `ANYFIELD` expressions now (IvaratorRequiredVisitor handles the check). So the maxValueExpansionThreshold case no longer throws, closing.

PR #638, opened 2019-11-01
re #637: periodic test failures

ivakegg:
This one fixes the MaxExpansionRegexQueryTest and resulted in fixing #156 as well.

* Updated to use google Files to create temporary directories (avoid known bug in nio Files temporary dir creation)
* Updating planner to correctly fail when attempting to ivarator when not configured.
* Fixed to allow ivarating against _ANYFIELD_ expressions (#156)

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexRegexIteratorJexl.java
+++ b/warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexRegexIteratorJexl.java
@@ -1,8 +1,10 @@
 package datawave.core.iterators;
 
+import datawave.data.ColumnFamilyConstants;
 import datawave.query.Constants;
 import datawave.query.parser.JavaRegexAnalyzer;
 import datawave.query.parser.JavaRegexAnalyzer.JavaRegexParseException;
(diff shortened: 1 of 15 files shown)


ivakegg:
This may have also fixed the FilterFieldsQueryTest as that uses the AbstractFunctionQuery for which I replaced the temporary directory creation mechanism for the more reliable google implementation

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexRegexIteratorJexl.java
```diff
@@ -93,6 +95,10 @@ public String toString() {
             startKey = new Key(rowId, fiName);
             endKey = new Key(rowId, new Text(fiName.toString() + '\0'));
             return new RangeSplitter(new Range(startKey, true, endKey, true), getMaxRangeSplit());
+        } else if (ANY_FINAME.equals(fiName)) {
```
FineAndDandy: what happens if we have a negated ANY_FINAME

ivakegg: You are correct.  Moving anyfield logic up

PR #639, opened 2019-11-01
Updated the ivarators to be able to use alternate filesystems and pat…

jwomeara:
…hs when persisting cached entries to disk.


warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -326,68 +330,67 @@ public DatawaveFieldIndexCachingIteratorJexl() {
...
         this.scanThreshold = 10000;
         this.hdfsBackedSetBufferSize = 10000;
         this.maxOpenFiles = 100;
+        this.numRetries = 2;
         this.maxRangeSplit = 11;
         
         this.sortedUIDs = true;
     }
     
     protected DatawaveFieldIndexCachingIteratorJexl(Builder builder) {
```
FineAndDandy: seems like builder can have a variety of accepted forms, can you add javadoc to formalize these?

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -326,68 +330,67 @@ public DatawaveFieldIndexCachingIteratorJexl() {
...
+        this.negated = builder.negated;
+        this.returnKeyType = builder.returnKeyType;
+        this.timeFilter = builder.timeFilter;
+        this.datatypeFilter = builder.datatypeFilter;
+        
+        this.ivaratorCacheDirs = builder.ivaratorCacheDirs;
+        
+        if (ivaratorCacheDirs.size() > 0) {
+            this.controlFs = ivaratorCacheDirs.get(0).getFs();
+            this.controlDir = new Path(ivaratorCacheDirs.get(0).getPathURI());
```
FineAndDandy: is there any reason not to make these a separate config instead of by convention using the first?

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
```diff
@@ -58,25 +62,96 @@ public BufferedFileBackedSortedSet(BufferedFileBackedSortedSet<E> other) {
...
+            if (fsStatus != null) {
+                long availableStorageMB = fsStatus.getRemaining() / 0x100000L;
+                double availableStoragePercent = (double) fsStatus.getRemaining() / fsStatus.getCapacity();
+                
+                // if we are using less than our storage limit, the cache dir is valid
+                isCacheDirValid = availableStorageMB >= ivaratorCacheDir.getConfig().getMinAvailableStorageMB()
+                                && availableStoragePercent >= ivaratorCacheDir.getConfig().getMinAvailableStoragePercent();
+            }
+        } else {
+            isCacheDirValid = true;
```
FineAndDandy: need to check available space for non-hdfs directories?

jwomeara: We only ever create SortedSetHdfsFileHandlerFactory objects, so this case will never be reached in a deployment of DataWave.  However, for testing, there are tests which do NOT create SortedSetHdfsFileHandlerFactory objects, so we need to be able to handle that when we reach this logic.

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
jwomeara: It's a little bit trickier than that actually.  Prior to the ivarator cache directories being passed to the query iterator, their order is randomized (respecting priority of course).  So, if you have 5 ivarator cache dirs with priority 0, they will be shuffled, and the first one will be used as the 'controlDir'.  This is intended to spread out the load among the various ivarator cache directories.  Hope that clears things up.

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
jwomeara: Sure.  Are you saying that you want me to just list the various extensions of the builder here?

warehouse/query-core/src/main/java/datawave/query/iterator/ivarator/IvaratorCacheDirConfig.java
```diff
@@ -0,0 +1,138 @@
...
+        this.priority = priority;
+        this.minAvailableStorageMB = minAvailableStorageMB;
+        this.minAvailableStoragePercent = minAvailableStoragePercent;
+    }
+    
+    @JsonIgnore
+    public boolean isValid() {
+        boolean result = true;
+        
+        if (!basePathURI.startsWith("file:/") && !basePathURI.startsWith("hdfs:/")) {
```
keith-ratcliffe: Null check on basePathURI first, since it's null by default?

warehouse/query-core/src/main/java/datawave/query/iterator/ivarator/IvaratorCacheDirConfig.java
```diff
@@ -0,0 +1,138 @@
...
+        ObjectMapper objectMapper = new ObjectMapper();
+        objectMapper.configure(SerializationFeature.WRITE_SINGLE_ELEM_ARRAYS_UNWRAPPED, true);
+        return objectMapper.writeValueAsString(ivaratorCacheDirConfigs);
+    }
+    
+    public static List<IvaratorCacheDirConfig> fromJson(String jsonArray) throws JsonProcessingException {
+        ObjectMapper objectMapper = new ObjectMapper();
+        objectMapper.configure(DeserializationFeature.ACCEPT_SINGLE_VALUE_AS_ARRAY, true);
+        return new ArrayList<>(Arrays.asList(objectMapper.readValue(jsonArray, IvaratorCacheDirConfig[].class)));
+    }
```
keith-ratcliffe: Could leverage a single static final ObjectMapper instead, and just configure once via static block

jwomeara: I thought about that.  Is ObjectMapper thread safe though?  Does it even need to be the way I'm using it...  Let me think on this.

warehouse/query-core/src/main/java/datawave/query/iterator/ivarator/IvaratorCacheDirConfig.java
jwomeara: Agreed.

warehouse/query-core/src/main/java/datawave/query/iterator/ivarator/IvaratorCacheDirConfig.java
keith-ratcliffe: Yep it's threadsafe, as long as you don't modify it's config after the fact

keith-ratcliffe: > Yep it's threadsafe, as long as you don't modify it's config after the fact

i.e., as long as different threads don't need to alter the config for whatever reason

jwomeara: Got it!  pr updated.

warehouse/query-core/src/main/java/datawave/query/iterator/ivarator/IvaratorCacheDirConfig.java
jwomeara: fixed in latest commit.

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
jwomeara: Added some comments.  Good to go?

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
```diff
@@ -58,25 +62,96 @@ public BufferedFileBackedSortedSet(BufferedFileBackedSortedSet<E> other) {
...
         this.comparator = comparator;
-        this.handlerFactory = handlerFactory;
+        this.handlerFactories = handlerFactories;
         this.bufferPersistThreshold = bufferPersistThreshold;
+        this.numRetries = numRetries;
         this.maxOpenFiles = maxOpenFiles;
     }
     
+    private SortedSetFileHandler createFileHandler(SortedSetFileHandlerFactory handlerFactory) throws IOException {
+        boolean isCacheDirValid = false;
```
ivakegg: Curious as to your design decision here.  Why did you not encapsulate the logic related to choosing file systems within the SortedSetFileHandler implementation?

warehouse/query-core/src/main/java/datawave/query/iterator/builder/IvaratorBuilder.java
```diff
@@ -1,35 +1,65 @@
...
     protected CompositeMetadata compositeMetadata;
     protected int compositeSeekThreshold;
     
-    public FileSystem getHdfsFileSystem() {
-        return hdfsFileSystem;
+    protected void validateIvaratorCacheDirs(List<IvaratorCacheDir> ivaratorCacheDirs) {
+        if (ivaratorCacheDirs.isEmpty())
+            throw new IllegalStateException("No ivarator cache dirs defined");
+        
+        IvaratorCacheDir ivaratorCacheDir = ivaratorCacheDirs.get(0);
```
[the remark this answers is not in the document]
jwomeara: I think originally I had the IvaratorCacheDirConfig objects being validated here using their isValid() method, but I ended up moving that into the IteratorBuildingVisitor instead.  So, by this point, the ivarator cache dir configs have already been validated.  This function is just making sure that the control directory exists, since we know that path/fs will be needed by the ivarator.

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
jwomeara: I liked the idea of a file handler factory per filesystem, but I could be convinced to switch it up if you feel strongly about it.

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
jwomeara: done

warehouse/query-core/src/main/java/datawave/query/iterator/builder/IvaratorBuilder.java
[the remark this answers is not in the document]
jwomeara: updated.

properties/default.properties
```diff
@@ -401,7 +401,26 @@ atom.connection.pool.name=WAREHOUSE
...
+##        paths preference over others.  For paths which share the same priority, the order will be determined at
+##        random.  Otherwise, paths are sorted in ascending order by priority.
+## 3) The third argument can be specified either as an exact number of MB (long), or as a percent (double).
+##   - minAvailableStorageMB - (Optional, Default: 0) A long, greater than or equal to 0, which specifies the minimum
+##        amount of available storage space required to persist to this ivarator path.  If less than this amount is
+##        available, we will not use this ivarator path.
+##   - minAvailableStoragePercent - (Optional, Default: 0.0) A double, between 0.0 and 1.0, which specifies the
+##        minimum percent of available storage space required to persist to this ivarator path.  If less than this
+##        percent is available, we will not use this ivarator path.
+ivarator.cache.dir.config= \
```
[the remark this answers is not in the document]
jwomeara: added another example with multiple ivarator cache dirs.

properties/default.properties
```diff
@@ -401,7 +401,41 @@ atom.connection.pool.name=WAREHOUSE
...
+          <bean class="datawave.query.iterator.ivarator.IvaratorCacheDirConfig"> \
+\n            <constructor-arg value="hdfs:///IvaratorCache" /> \
+\n        </bean>
+
+# By comparison, here is an example configuration which specifies two ivarator cache dirs.
+# - The first cache directory is on the local disk, has a priority of 0, and has minAvailableStorageMB set to 4096.
+# - The second cache directory is in hdfs, has a priority of 1, and has minAvailableStoragePercent set to 0.33 (i.e. 33%)
+#ivarator.cache.dir.config= \
+#          <bean class="datawave.query.iterator.ivarator.IvaratorCacheDirConfig"> \
```
ivakegg: +1

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
```diff
@@ -58,25 +62,96 @@ public BufferedFileBackedSortedSet(BufferedFileBackedSortedSet<E> other) {
...
+            FsStatus fsStatus = null;
+            try {
+                fsStatus = ivaratorCacheDir.getFs().getStatus();
+            } catch (IOException e) {
+                log.warn("Unable to determine status of the filesystem: " + ivaratorCacheDir.getFs());
+            }
+            
+            // determine whether this fs is a good candidate
+            if (fsStatus != null) {
+                long availableStorageMB = fsStatus.getRemaining() / 0x100000L;
```
jwomeara: @ivakegg 
told you

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
FineAndDandy: maybe this should be delegated then instead of implemented here?

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
ivakegg: You got me....awesome

ivakegg: Since this is MB and not MiB, I think you should be dividing by 100000

jwomeara: Oh, who are you, NIST???  I'll fix it, haha.

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
jwomeara: I'll create a validate method.

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
jwomeara: fixed.

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
jwomeara: done.

Issue #641, opened 2019-11-04
Test IndexMatch read/write methods


apmoriarty:

apmoriarty:
Merged in https://github.com/NationalSecurityAgency/datawave/pull/642

PR #642, opened 2019-11-04
Test read/write methods for the IndexMatch object.

apmoriarty:
Addresses https://github.com/NationalSecurityAgency/datawave/issues/641

### DIFF
--- a/warehouse/query-core/src/test/java/datawave/query/index/lookup/IndexMatchTest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/index/lookup/IndexMatchTest.java
@@ -19,6 +19,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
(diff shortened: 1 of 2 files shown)


warehouse/query-core/src/test/java/datawave/query/index/lookup/IndexMatchTest.java
```diff
@@ -216,4 +223,46 @@ public void testHashCode() {
...
+        assertEquals("uid", other.uid);
+    }
+    
+    @Test(expected = NullPointerException.class)
+    public void testWriteReadIndexMatchWithNullUid() throws IOException {
+        IndexMatch match = new IndexMatch(null);
+        
+        // Cannot write a null uid.
+        IndexMatch other = writeRead(match);
+        assertNull(other.uid);
```
jschmidt10: This `assertNull` is probably never reached if the `match.write()` NPE's. I'd just drop it.

apmoriarty: Makes sense. I'll replace it with a fail("Should not have made it this far") kind of assertion.

PR #652, opened 2019-11-14
re DATAWAVE-559: ActiveQueryLog timer task not being scheduled due to…

billoley:
ActiveQueryLog timer task not being scheduled due to logic issue

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
+++ b/warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
@@ -35,7 +35,7 @@ public class ActiveQueryLog {
     
     private Cache<String,ActiveQuery> CACHE = null;
     private ReentrantReadWriteLock cacheLock = new ReentrantReadWriteLock();
-    private Timer timer = new Timer("ActiveQueryLog");
+    private Timer timer = null;
     
     synchronized public static void setConfig(AccumuloConfiguration conf) {
         if (conf != null) {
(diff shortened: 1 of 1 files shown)


warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
```diff
@@ -35,7 +35,7 @@
     
     private Cache<String,ActiveQuery> CACHE = null;
     private ReentrantReadWriteLock cacheLock = new ReentrantReadWriteLock();
-    private Timer timer = new Timer("ActiveQueryLog");
+    private Timer timer = null;
```
billoley: timer needs to start as null so that we can differentiate between when it is first being initialized and when it is being reinitialized

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
```diff
@@ -73,9 +73,9 @@ synchronized public void setLogPeriod(long logPeriod) {
             if (logPeriod != this.logPeriod || this.timer == null) {
                 if (this.timer != null) {
                     this.timer.cancel();
-                    this.timer = new Timer("ActiveQueryLog");
                 }
                 this.logPeriod = logPeriod;
+                this.timer = new Timer("ActiveQueryLog");
```
billoley: this.timer should be set with a new Timer for either condition - logPeriod != this.logPeriod || this.timer == null
It's the cancel that shouldn't be executed when this.timer is null to avoid an NPE

PR #657, opened 2019-11-20
fixes #656: Updated to avoid nested JexlASTScript nodes in tree

### LINKED ISSUE #656: GeoFunctions expansion creates JexlScript Nodes nested in tree
The GeoFunctions expansion now creates JexlScript nodes nested within the tree which will mess up subsequent tree visitors during planning.

ivakegg:


jwomeara:
> the negatives in this comment confuse me. "since we do not want the script not the reference"
I think he meant 'nor'

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GeoFunctionsDescriptor.java
```diff
@@ -300,8 +300,10 @@ public JexlNode rebuildNode(ShardQueryConfiguration settings, MetadataHelper met
                             
                             // if there are other fields, recreate the geo function node
                             if (!otherFields.isEmpty()) {
-                                JexlNode geoNode = JexlASTHelper.parseJexlQuery(namespace + ":" + name + "(" + getFieldParam(otherFields) + ", '"
-                                                + args.get(1).image + "', '" + args.get(2).image + "')");
+                                // using dereference of the child node since we do not want the JexlASTScript not the JexlASTReference parent nodes
```
alerman:
```suggestion
                                // using dereference of the child node since we do not want the JexlASTScript nor the JexlASTReference parent nodes
```

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GeoFunctionsDescriptor.java
```diff
@@ -353,8 +355,11 @@ public JexlNode toGeoWaveFunction(Set<String> fields) throws Exception {
                 }
             }
             
-            if (wkt != null)
-                return JexlASTHelper.parseJexlQuery("geowave:intersects(" + getFieldParam(fields) + ", '" + wkt + "')");
+            if (wkt != null) {
+                // using dereference of the child node since we do not want the JexlASTScript not the JexlASTReference parent nodes
```
alerman:
```suggestion
                // using dereference of the child node since we do not want the JexlASTScript nor the JexlASTReference parent nodes
```

jwomeara:
> Is it possible that this would return without children? Do we need a check to confirm that a child is present before we call getChild

I don't think that is possible...  no.

alerman:
added suggestions to nor since @jwomeara thought that was what was meant

ivakegg:
> Is it possible that this would return without children? Do we need a check to confirm that a child is present before we call getChild

That is not possible

PR #660, opened 2019-11-26
fixes #659: ContentOrderedEvaluator fixed and new JUnits created

### LINKED ISSUE #659: Phrase and other content-related queries may result in missed hits
The bad behavior is related to ContentOrderedEvaluator, and deals with cases where a document contains both partial and full matches for the terms being queried

After encountering a partial-match dead end, the [ContentOrderedEvaluator.traverseFailure](https://github.com/NationalSecurityAgency/datawave/blob/master/warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentOrderedEvaluator.java#L339) method is used to continue traversal at the point of failure

However, since *traverseFailure* doesn't communicate subsequent full matches by way of the the caller's "found" list, the caller is unaware of that success and only reports the partial match failure back up the call chain

There are two new junits in my PR that will reproduce the bug...
```java
   @Test
   public void evaluate_traverseFailureFORWARDTest() { ... }
   @Test
   public void evaluate_traverseFailureREVERSETest() { ... }
```
...when run using the unpatched  ContentOrderedEvaluator

keith-ratcliffe:


warehouse/query-core/src/test/java/datawave/query/jexl/functions/ContentOrderedEvaluatorTest.java
```diff
@@ -0,0 +1,262 @@
...
+        offsets.add(offsetsB);
+        offsets.add(offsetsC);
+        
+        evaluator = new WrappedContentOrderedEvaluator(null, 1, new HashMap<>(), "a", "b", "c");
+        
+        Assert.assertTrue(evaluator.evaluate(offsets));
+    }
+    
+    @Test
+    public void evaluate_livePruneTest() {
```
ivakegg: Can we add a test where we need to alternate the match please
A = 1, 10, 20
B = 21, 24, 30
C = 3, 8, 22

keith-ratcliffe: Added 2 new tests for this, for FORWARD and REVERSE variants

PR #661, opened 2019-11-26
re #559 - Cancel and create TimerTasks instead of Timers, cancel Time…

billoley:
When datawave is upgraded and a new classloader is created, a new ActiveQueryLog instance gets created with a new Timer and Timer thread and the old Timer thread is still running.  

Cancel and create TimerTasks instead of Timers, cancel Timer if class not used in 24 hours (classloader issue)

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
+++ b/warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
@@ -12,6 +12,7 @@ import java.util.Timer;
 import java.util.TimerTask;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 public class ActiveQueryLog {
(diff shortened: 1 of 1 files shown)


warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
```diff
@@ -26,16 +26,21 @@
...
+    private long maxIdle = MINUTES_15_MS;
+    private long logPeriod = MINUTES_1_MS;
     private int logMaxQueries = 5;
     private int windowSize = 10;
     
     private Cache<String,ActiveQuery> CACHE = null;
     private ReentrantReadWriteLock cacheLock = new ReentrantReadWriteLock();
     private Timer timer = null;
+    private TimerTask timerTask = null;
+    private long lastAccess = System.currentTimeMillis();
```
ivakegg: This needs to be marked volatile

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
```diff
@@ -68,20 +78,49 @@ private ActiveQueryLog(AccumuloConfiguration conf) {
         }
     }
     
+    synchronized private void touch() {
```
ivakegg: since we are simply changing a long here, I am not sure we need to synchronize.  If two threads are setting this at the same time, then it really does not matter that they execute simultaneously.  The setting of the long value is atomic.

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
```diff
@@ -68,20 +78,49 @@ private ActiveQueryLog(AccumuloConfiguration conf) {
...
+        this.lastAccess = System.currentTimeMillis();
+    }
+    
+    synchronized long getLastAccess() {
```
ivakegg: no need to synchronize here

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
```diff
@@ -26,16 +26,21 @@
...
     public static final String LOG_MAX_QUERIES = "datawave.query.active.logMaxQueries";
     public static final String WINDOW_SIZE = "datawave.query.active.windowSize";
+    private static long HOURS_24_MS = TimeUnit.MILLISECONDS.convert(24, TimeUnit.HOURS);
+    private static long MINUTES_15_MS = TimeUnit.MILLISECONDS.convert(15, TimeUnit.MINUTES);
+    private static long MINUTES_1_MS = TimeUnit.MILLISECONDS.convert(1, TimeUnit.MINUTES);
     
     // Changeable via Accumulo properties
-    private long maxIdle = 900000;
-    private long logPeriod = 60000;
+    private long maxIdle = MINUTES_15_MS;
```
ivakegg: set the longs and int values to be volatile if not final

PR #662, opened 2019-11-27
Fixed geo evaluation to properly handle bounds which cross the antime…

jwomeara:
…ridian.


warehouse/query-core/src/main/java/datawave/query/jexl/functions/GeoFunctions.java
```diff
@@ -98,7 +110,14 @@ public static boolean within_bounding_box(Object lonField, Object latField, doub
             return false;
         }
         
-        boolean lonMatch = lonValues.stream().anyMatch(lon -> lon >= minLon && lon <= maxLon);
+        // @formatter:off
+        boolean lonMatch = lonValues.stream().anyMatch(lon ->
```
ivakegg: We should probably add a test for this case

jwomeara: done

Issue #670, opened 2019-12-04
Don't add e columns for shard exclusions


hlgp:
We're adding e columns in the metadata table for shard exclusion fields, causing them not to be treated as index-only fields.

PR #671, opened 2019-12-05
Addresses #670  don't add e columns for excluded fields

hlgp:


warehouse/ingest-core/src/main/java/datawave/ingest/metadata/EventMetadata.java
```diff
@@ -216,6 +216,11 @@ protected void addEventField(IngestHelperInterface helper, RawRecordContainer ev
             return;
         }
         
+        if (helper.isShardExcluded(fieldName)) {
+            log.debug("{} is an excluded field, not adding to event", fieldName);
```
alerman: Should this message be "not adding to event **fields**"?

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
```diff
@@ -551,6 +551,11 @@ public void addIndexOnlyField(String fieldName) {
         indexOnlyFields.add(fieldName);
     }
     
+    @Override
+    public void addShardExclusionField(String fieldName) {
```
ivakegg: Lets use this method around line 393

warehouse/ingest-core/src/main/java/datawave/ingest/metadata/EventMetadata.java
hlgp: no, it should say "not adding event or frequency metadata columns," but I was just following precedent on lines 215 and 210.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
hlgp: Considering that we have an addIndexOnlyField method and we don't call it in the equivalent spot at 411, I didn't think it mattered.

PR #674, opened 2019-12-06
fixes #673 - Updated the regex function visitor to avoid creating or …

### LINKED ISSUE #673: Regex Function Visitor is creating nodes with null children.
In the RegexFunctionVisitor, in the ASTFunctionNode visitor, when we create regex nodes from an include/exclude regex, we are using the index of the original node, not the recreated node when adding children to the or node.  We should update the 'adopt' method to insert the child using the current number of children as the index.

jwomeara:
…nodes with null children.


jwomeara:
@ivakegg @ejrgilbert

jwomeara:
> Can we add a unit test for this?

Your wish is my command.

PR #677, opened 2019-12-10
fixes #676 Allow index only fields to be delayed

### LINKED ISSUE #676: Allow delayed index only lookups
Addresses #658 by allowing these negations to be delayed. Also allows for the query planner to make decisions about pushing down (delaying) terms without concern for causing the query to become non-executable.

FineAndDandy:


warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -937,18 +938,27 @@ public T next() {
...
-            if (exceededOrEvaluationCache != null)
-                contextCreator.addAdditionalEntries(exceededOrEvaluationCache);
-            
-            final Iterator<Tuple3<Key,Document,DatawaveJexlContext>> itrWithDatawaveJexlContext = Iterators.transform(itrWithContext, contextCreator);
-            Iterator<Tuple3<Key,Document,DatawaveJexlContext>> matchedDocuments = statelessFilter(itrWithDatawaveJexlContext, jexlEvaluationFunction);
-            if (log.isTraceEnabled()) {
-                log.trace("arithmetic:" + arithmetic + " range:" + getDocumentRange(documentSource) + ", thread:" + Thread.currentThread());
+            try {
+                IteratorBuildingVisitor iteratorBuildingVisitor = createIteratorBuildingVisitor(getDocumentRange(documentSource), false, this.sortedUIDs);
+                DelayedIndexOnlySubTreeVisitor delayedIndexOnlySubTreeVisitor = DelayedIndexOnlySubTreeVisitor.processDelayedSubTrees(script,
```
ivakegg: index only or non-event...

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -937,18 +938,27 @@ public T next() {
...
-            
-            if (exceededOrEvaluationCache != null)
-                contextCreator.addAdditionalEntries(exceededOrEvaluationCache);
-            
-            final Iterator<Tuple3<Key,Document,DatawaveJexlContext>> itrWithDatawaveJexlContext = Iterators.transform(itrWithContext, contextCreator);
-            Iterator<Tuple3<Key,Document,DatawaveJexlContext>> matchedDocuments = statelessFilter(itrWithDatawaveJexlContext, jexlEvaluationFunction);
-            if (log.isTraceEnabled()) {
-                log.trace("arithmetic:" + arithmetic + " range:" + getDocumentRange(documentSource) + ", thread:" + Thread.currentThread());
+            try {
+                IteratorBuildingVisitor iteratorBuildingVisitor = createIteratorBuildingVisitor(getDocumentRange(documentSource), false, this.sortedUIDs);
```
ivakegg: We need to double check, but any additional iterators you add in here must adhere to the contract stated at the top of the createDocumentPipeline method: "It is very important that this pipeline can handle resetting the bottom iterator with a new value. This means that hasNext() needs to call the next iterator. The only state that can be maintained is the next value ready after hasNext() has been called. Once next returns the value, the next hasNext() call must call the next iterator again. So for example Iterators.filter() cannot be used as it uses a google commons AbstractIterator that maintains an iterator state (failed, ready, done); use statelessFilter above instead."

warehouse/query-core/src/main/java/datawave/query/jexl/DelayedIndexOnlyContext.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Responsible for retrieving delayed fields for the specified docRange on demand and merging it with any values already in the delegate. Use the
+ * IteratorBuildingVisitor on each delayed sub tree to generate iterators over the docRange.
+ */
+public class DelayedIndexOnlyContext extends DatawaveJexlContext {
```
ivakegg: DelayedNonEventContext ?

warehouse/query-core/src/main/java/datawave/query/jexl/DelayedIndexOnlyContext.java
```diff
@@ -0,0 +1,131 @@
...
+    private List<Document> fetchOnDemand(String name) throws IOException {
+        List<Document> documentList = new ArrayList<>();
+        
+        // limit the ranges to use to the current document
+        iteratorBuildingVisitor.limit(docRange);
+        
+        // for each sub tree build the nested iterator
+        for (JexlNode subTree : delayedSubTrees) {
+            // short circuit trees that have no identifiers
+            if (JexlASTHelper.getIdentifiers(subTree).size() == 0) {
```
ivakegg: This does a full scan of the underlying tree.  Is that worth it?

warehouse/query-core/src/main/java/datawave/query/function/IndexOnlyContextCreator.java
```diff
@@ -168,6 +182,18 @@ protected DatawaveJexlContext newDatawaveJexlContext(final Tuple3<Key,Document,M
             newContext = parentContext;
         }
         
+        // see if there are any delayed subtrees that are index only
+        if (delayedIndexOnlySubTreeVisitor != null && delayedIndexOnlySubTreeVisitor.getFoundIndexOnlyFields().size() > 0) {
+            // build the current document range from the document Key to end of the document, even though for some query logics this may be too large a range,
+            // it will be narrowed with equality later
+            Key startKey = new Key(from.first().getRow(), from.first().getColumnFamily());
```
ivakegg: Are we creating a document range or a field index range here?

warehouse/query-core/src/main/java/datawave/query/jexl/DelayedIndexOnlyContext.java
```diff
@@ -0,0 +1,131 @@
...
+                // get all the leaf nodes, discarding the rest
+                Collection<NestedIterator<Key>> subTreeLeaves = subTreeNestedIterator.leaves();
+                // for each leaf, see if its a match for the target field
+                for (NestedIterator<Key> leaf : subTreeLeaves) {
+                    // grab the field off the leaf and make sure it matches
+                    if (leaf instanceof FieldedIterator && name.equals(((FieldedIterator) leaf).getField())) {
+                        // init/seek the leaf
+                        leaf.initialize();
+                        if (leaf instanceof SeekableIterator) {
+                            ((SeekableIterator) leaf).seek(docRange, null, false);
```
ivakegg: As I noted above where we create this range, are we scanning the field index or the events.

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/JexlNodes.java
```diff
@@ -199,4 +199,37 @@ public static ASTNumberLiteral literal(Number n) {
...
+    public static boolean removeFromParent(JexlNode parent, JexlNode childToRemove) {
+        boolean found = false;
+        JexlNode[] children = new JexlNode[parent.jjtGetNumChildren() - 1];
+        int childIndex = 0;
+        for (int i = 0; i < parent.jjtGetNumChildren(); i++) {
+            JexlNode child = parent.jjtGetChild(i);
+            if (!child.equals(childToRemove)) {
+                children[childIndex++] = child;
+            } else {
+                // clear the praent of the node that is removed
```
ivakegg: praent -> parent

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/JexlNodes.java
```diff
@@ -199,4 +199,37 @@ public static ASTNumberLiteral literal(Number n) {
...
+     * @return true if childToRemove was successfully found and removed from parent, false otherwise
+     */
+    public static boolean removeFromParent(JexlNode parent, JexlNode childToRemove) {
+        boolean found = false;
+        JexlNode[] children = new JexlNode[parent.jjtGetNumChildren() - 1];
+        int childIndex = 0;
+        for (int i = 0; i < parent.jjtGetNumChildren(); i++) {
+            JexlNode child = parent.jjtGetChild(i);
+            if (!child.equals(childToRemove)) {
+                children[childIndex++] = child;
```
ivakegg: If we do not find the child, then I think we will hit an array out of bounds exception here

FineAndDandy: good catch

warehouse/query-core/src/main/java/datawave/query/function/IndexOnlyContextCreator.java
FineAndDandy: IteratorBuildingVisitor expects document ranges

warehouse/query-core/src/main/java/datawave/query/jexl/DelayedIndexOnlyContext.java
FineAndDandy: IndexIterator calls buildIndexRange(), which converts a document key to an fi range

FineAndDandy: This needs to be updated to pass the columnFamilies/include flag from the QueryIterator

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/DelayedIndexOnlySubTreeVisitor.java
```diff
@@ -0,0 +1,165 @@
...
+ * modify/destroy the tree so a copy should always be used
+ */
+public class DelayedIndexOnlySubTreeVisitor extends BaseVisitor {
+    private static final Logger log = Logger.getLogger(DelayedIndexOnlySubTreeVisitor.class);
+    
+    private Set<String> nonEventFields;
+    private Set<JexlNode> delayedSubTrees;
+    private Set<String> foundIndexOnlyFields;
+    
+    public static DelayedIndexOnlySubTreeVisitor processDelayedSubTrees(ASTJexlScript script, Set<String> nonEventFields) {
```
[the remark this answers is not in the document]
FineAndDandy: agree

warehouse/query-core/src/main/java/datawave/query/jexl/DelayedIndexOnlyContext.java
FineAndDandy: agree

warehouse/query-core/src/main/java/datawave/query/jexl/DelayedIndexOnlyContext.java
FineAndDandy: This case is actually covered in the DelayedNonEventSubTreeVisitor now, so will remove

warehouse/query-core/src/main/java/datawave/query/jexl/DelayedIndexOnlyContext.java
FineAndDandy: This looks like its okay for ivarators as well

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/JexlNodes.java
FineAndDandy: yes

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/JexlNodes.java
```diff
@@ -199,4 +202,39 @@ public static ASTNumberLiteral literal(Number n) {
...
+     */
+    public static boolean removeFromParent(JexlNode parent, JexlNode childToRemove) {
+        boolean found = false;
+        // at most as many children as currently exist
+        List<JexlNode> children = new ArrayList<>(parent.jjtGetNumChildren());
+        JexlNode[] nodeArray = new JexlNode[0];
+        
+        for (int i = 0; i < parent.jjtGetNumChildren(); i++) {
+            JexlNode child = parent.jjtGetChild(i);
+            if (!child.equals(childToRemove)) {
```
jwomeara: .equals isn't implemented for JexlNode or it's descendants.  This could probably just be a simple '=='.  Also, if child is null, this will throw an exception.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/DelayedNonEventSubTreeVisitor.java
```diff
@@ -0,0 +1,165 @@
...
+public class DelayedNonEventSubTreeVisitor extends BaseVisitor {
+    private static final Logger log = Logger.getLogger(DelayedNonEventSubTreeVisitor.class);
+    
+    private Set<String> nonEventFields;
+    private Set<JexlNode> delayedSubTrees;
+    private Set<String> foundNonEventFields;
+    
+    public static DelayedNonEventSubTreeVisitor processDelayedSubTrees(ASTJexlScript script, Set<String> nonEventFields) {
+        // create a safe copy to rip apart
+        ASTJexlScript copy = (ASTJexlScript) RebuildingVisitor.copy(script);
```
jwomeara: Better yet, if you have no objections could you flatten the tree to avoid potential stack overflows?  In the past we saw the occasional stack overflow in the DataWaveInterpreter.  I don't think the tree has been flattened prior to this call either, but I may be wrong.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/DelayedNonEventSubTreeVisitor.java
```diff
@@ -0,0 +1,165 @@
...
+     * @param node
+     * @param data
+     * @return
+     */
+    @Override
+    public Object visit(ASTAndNode node, Object data) {
+        if (JexlASTHelper.isDelayedPredicate(node)) {
+            // check for expected form
+            if (node.jjtGetNumChildren() == 2) {
+                // grab the sub tree off the delayed node for processing, the second node is always the one that
```
jwomeara: always the one that...  (the suspense is killing me!)

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/DelayedNonEventSubTreeVisitor.java
```diff
@@ -0,0 +1,165 @@
...
+
+import java.util.HashSet;
+import java.util.NoSuchElementException;
+import java.util.Set;
+
+/**
+ * Visitor builds a list of all delayed sub trees within a script, and removes any nodes which are not based on index only lookups. This visitor will
+ * modify/destroy the tree so a copy should always be used
+ */
+public class DelayedNonEventSubTreeVisitor extends BaseVisitor {
```
jwomeara: So, you get all the delayed subtrees in this visitor, but then you are only using the leaf nodes (which can contain includes and excludes) when this makes it's way into DelayedNonEventIndexContext.  Why not just get the leaf nodes at this point.  Then you don't need to carry subtrees all the way through.  Also, you could create a multimap of field name to leaf nodes which would be useful later on when you need to do a lookup (i.e. no need to check all delayed subtrees to find what you're looking for).  You also wouldn't have to worry about pruning the tree if you took this approach.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/DelayedNonEventSubTreeVisitor.java
```diff
@@ -0,0 +1,165 @@
...
+     * 
+     * @param node
+     * @param data
+     * @return
+     */
+    @Override
+    public Object visit(ASTAndNode node, Object data) {
+        if (JexlASTHelper.isDelayedPredicate(node)) {
+            // check for expected form
+            if (node.jjtGetNumChildren() == 2) {
```
jwomeara: This assumption will break if you start flattening the tree up front.

FineAndDandy:
I'm expecting to add at least one more commit on this

FineAndDandy:
This is now ready for review. There are a few notable changes

- nonEvent fields can now be delayed
- composite fields can now be delayed, but not be made up of non-event fields
- all ivarators can now be delayed
- ExecutionDeterminationVisitor will no longer return a state of error if a nonEvent field is present in a delayed subtree

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExecutableDeterminationVisitor.java
```diff
@@ -394,7 +394,7 @@ else if (ExceededValueThresholdMarkerJexlNode.instanceOf(node) || ExceededOrThre
         else if (ASTDelayedPredicate.instanceOf(node) || ASTEvaluationOnly.instanceOf(node)) {
             if (isNoFieldOnly(node)) {
                 state = STATE.IGNORABLE;
-            } else if (isNonEvent(node)) {
+            } else if (ASTEvaluationOnly.instanceOf(node) && isNonEvent(node)) {
```
jwomeara: Should we add a check for delayed here as well?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandCompositeTerms.java
```diff
@@ -173,8 +174,8 @@ public Object visit(ASTOrNode node, Object data) {
     public Object visit(ASTAndNode node, Object data) {
         ExpandData parentData = (ExpandData) data;
         
-        // ignore marked nodes
-        if (QueryPropertyMarkerVisitor.instanceOfAny(node))
+        // ignore marked nodes except delays
+        if (QueryPropertyMarkerVisitor.instanceOfAny(node) && !QueryPropertyMarkerVisitor.instanceOf(node, ASTDelayedPredicate.class, null))
```
jwomeara: I think you could improve performance here by passing a list of acceptable markers instead of doing one pass checking for any, and another checking for delayed predicate.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandCompositeTerms.java
```diff
@@ -291,8 +292,8 @@ public Object visit(ASTFunctionNode node, Object data) {
     // don't descend into delayed predicates
     @Override
     public Object visit(ASTReference node, Object data) {
-        // ignore marked nodes
-        if (!QueryPropertyMarkerVisitor.instanceOfAny(node)) {
+        // ignore marked nodes except delays
+        if (!QueryPropertyMarkerVisitor.instanceOfAny(node) || QueryPropertyMarkerVisitor.instanceOf(node, ASTDelayedPredicate.class, null)) {
```
jwomeara: same comment here as above.

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
```diff
@@ -206,29 +206,17 @@ public ScannerChunk apply(@Nullable ScannerChunk input) {
...
-                                    }
-                                }
-                            }
-                            
-                        } else {
-                            if (input.getRanges().size() == 1) {
-                                if (log.isTraceEnabled()) {
-                                    log.trace("Ensuring max pipelines is set to 1");
-                                    
+                        if (!evaluatedPreviously) {
```
jwomeara: Just trying to understand this...  So in the past, if our ranges were not all doc specific (i.e. we have some shard ranges), and we haven't previously encountered this exact query plan, we will do our large fielded list work (and then later cache this remade query).  OK, cool.  I don't know why we only did this in the past when we had some shard ranges in our query.  This seems like it would be good to do all the time (per your change).  Am I following that correctly?

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/JexlNodes.java
```diff
@@ -199,4 +202,44 @@ public static ASTNumberLiteral literal(Number n) {
...
+    /**
+     * Remove childToRemove from parent, updating all references on both the parent and childToRemove to be consistent
+     * 
+     * @param parent
+     *            the parent to remove from
+     * @param childToRemove
+     *            the child to remove from parent
+     * @return true if childToRemove was successfully found and removed from parent, false otherwise
+     */
+    public static boolean removeFromParent(JexlNode parent, JexlNode childToRemove) {
```
jwomeara: I would love to see this renamed to 'orphanChild'.   I'm 50/50 on whether that's a joke or not.

warehouse/query-core/src/test/resources/input/generic-cities.csv
```diff
@@ -1,16 +1,16 @@
...
+20150808,par-usa-mo-8,Paris,Missouri,United States,North America,USA,NA,110,39.48_-92.02
+20150808,rom-usa-mo-8,Rome,Missouri,United States,North America,USA,NA,110,36.84_-92.79
+20150808,ldn-usa-mo-8,London,Missouri,United States,North America,USA,NA,110,39.58_-91.42
+20150909,par-usa-oh-9,Paris,Ohio,United States,North America,USA,NA,100,40.80_-81.18
+20151010,par-usa-ma-10,Paris,Maine,United States,North America,USA,NA,120,44.24_-70.56
+20151010,rom-usa-ms-10,Rome,Mississippi,United States,North America,USA,NA,130,33.96_-90.50
+20151010,ldn-usa-mi-10,London,Michigan,United States,North America,USA,NA,140,42.04_-83.67
+20151111,par-ita-11,Paris,Lazio,Italy,Europe,ITA,Euro,30,41.91_12.26
+20151111,rom-bel-11,Rome,Hainaut,Belgium,Europe,BEL,Euro,50,50.54_3.25
+20151111,ldn-fra-lle-11,London,lle-de-France,France,Europe,FRA,Euro,20,46.63_5.08
```
jwomeara: I respect the fact that you added legitimate lat/lon values for each of these cities.  I also totally understand why 0/0 was used everywhere else, haha.

warehouse/query-core/src/test/java/datawave/query/testframework/AbstractDataManager.java
```diff
@@ -127,30 +128,31 @@ public String convertAnyField(final String phrase, final String op) {
...
-                if (getNormalizer(field) instanceof NumberNormalizer) {
-                    // remove quotes from phrase
-                    String num = "";
-                    String numPhrase = phrase;
-                    int start = phrase.indexOf('\'');
-                    if (0 <= start) {
-                        num = phrase.substring(start + 1);
-                        if (num.endsWith("'")) {
-                            num = num.substring(0, num.length() - 1);
+                if (!(getNormalizer(field) instanceof GeoNormalizer)) {
```
jwomeara: Ugh, sorry you had to update any of this...

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RangeConjunctionRebuildingVisitor.java
```diff
@@ -112,7 +112,7 @@ public RangeConjunctionRebuildingVisitor(ShardQueryConfiguration config, Scanner
     
     @Override
     public Object visit(ASTReference node, Object data) {
-        if (ASTDelayedPredicate.instanceOf(node) || IndexHoleMarkerJexlNode.instanceOf(node) || ASTEvaluationOnly.instanceOf(node)) {
+        if (IndexHoleMarkerJexlNode.instanceOf(node) || ASTEvaluationOnly.instanceOf(node)) {
```
jwomeara: Similar to comment above.  We could probably make this run faster by doing one run of the visitor and setting a list with IndexHoleMarker and EvaluationOnly as the two types we check for.  Not a big deal though.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandCompositeTerms.java
```diff
@@ -301,8 +302,8 @@ public Object visit(ASTReference node, Object data) {
     // don't descend into delayed predicates
     @Override
     public Object visit(ASTReferenceExpression node, Object data) {
-        // ignore marked nodes
-        if (!QueryPropertyMarkerVisitor.instanceOfAny(node))
+        // ignore marked nodes except delays
+        if (!QueryPropertyMarkerVisitor.instanceOfAny(node) || QueryPropertyMarkerVisitor.instanceOf(node, ASTDelayedPredicate.class, null))
```
jwomeara: same comment here as above.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExecutableDeterminationVisitor.java
FineAndDandy: a nonEvent delayed is exactly what we can now process

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandCompositeTerms.java
FineAndDandy: good idea

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
FineAndDandy: That is how I read it also, but I was thinking you or @ivakegg would catch me if that was wrong

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RangeConjunctionRebuildingVisitor.java
FineAndDandy: agree

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
```diff
@@ -1585,4 +1605,8 @@ public IteratorBuildingVisitor setExceededOrEvaluationCache(Map<String,Object> e
         this.exceededOrEvaluationCache = exceededOrEvaluationCache;
         return this;
     }
+
```
apmoriarty:
```suggestion
// Null out the root JexlNode.
```

jwomeara:
Let me know when this is ready for re-review!

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
jwomeara: Man, I love how this was new to me when I originally saw your PR, but now I've seen it and experienced it, and still don't know why we shouldn't ever do this for doc specific ranges.  Fun stuff.

FineAndDandy:
consolidated in https://github.com/NationalSecurityAgency/datawave/pull/755

PR #680, opened 2019-12-11
Fixes #678 - Failed Query Plans need to be sent to metrics

### LINKED ISSUE #678: Send Failed Query Plan to Metrics
Non executable queries need to have plans sent to metrics.

JustDoWhitt:


warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -418,16 +417,6 @@ private DateIndexHelper getDateIndexHelper(final ShardQueryConfiguration config)
...
-        // a full table scan is required if
-        final boolean isFullTable = queryRanges.second();
-        
-        // abort if we cannot handle full table scans
-        if (isFullTable && !config.getFullTableScanEnabled()) {
-            PreConditionFailedQueryException qe = new PreConditionFailedQueryException(DatawaveErrorCode.FULL_TABLE_SCAN_REQUIRED_BUT_DISABLED);
-            throw new FullTableScansDisallowedException(qe);
-        }
         
         TraceStopwatch stopwatch = timers.newStartedStopwatch("DefaultQueryPlanner - Rebuild JEXL String from AST");
```
apmoriarty: Let's update this comment to say something like "Saving state of query string prior to global index lookup."

JustDoWhitt: Updated.

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -445,6 +459,8 @@ public void initialize(ShardQueryConfiguration config, Connector connection, Que
         
         config.setQueryString(getQueryPlanner().getPlannedScript());
         
+        this.setQueryPlan(getQueryPlanner().getPlannedScript());
```
ivakegg: does this not simply set the query plan on the configuration?

JustDoWhitt: No, this sets the query plan onto a field i added into the logic. If the configuration fails, the plan will be set into this field. If the configuration initialization is successful, I just assigned it to both fields which isn't completely necessary, but at least it keeps them the same. It could be removed at this point since its already set in the configuration.

web-services/query/src/main/java/datawave/webservice/query/logic/BaseQueryLogic.java
```diff
@@ -26,6 +26,7 @@
     private GenericQueryConfiguration baseConfig;
     private String logicName = "No logicName was set";
     private String logicDescription = "Not configured";
+    protected String queryPlan = "No Query Plan was set.";
```
ivakegg: I guess the point I was trying to make is that if we can set this value in the configuration, then we do not want it also set here.  We used to have almost all of the configuration both in the logic and in the configuration which was causing a nightmare in terms of ensure values got set in both places.  Lets have this set only in the configuration if at all possible or give me a good reason why.

JustDoWhitt: Ah okay, I see what you were saying now. The reason I set this in the query logic is because when a query plan fails it returns back the configuration as null, which is why it can't be sent to metrics. To only set it in the configuration would require to handle the exceptions within the configuration and still return the configuration when the query plan fails. Is that preferable?

ivakegg: In the RunningQuery where you handle the exception case, you use logic.getQueryPlan().  There is no reason why logic.getQueryPlan() couldn't get the plan from the config object instead of from a local class member.  If you look at almost all of the other setters and getters in the logic, they simply defer to the configuration object.

JustDoWhitt: I believe I just made this change.

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
```diff
@@ -171,6 +169,9 @@ public void setConnection(Connector connection) throws Exception {
             this.getMetric().setError(e);
             throw e;
         } finally {
+            // get the query plan from logic config object
```
apmoriarty:
```suggestion
            // get the query plan from logic config object
            // set metric query plan from the query logic
```

PR #681, opened 2019-12-11
Merge CreateTLDUidsIterator into CreateUidsIterator

apmoriarty:
Moved functionality from CreateTLDUidsIterator into CreateUidsIterator. Any query can now parse root TLD uids by enabling the parseRootUids option.


warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
```diff
@@ -457,12 +456,7 @@ public ScannerStream visit(ASTEQNode node, Object data) {
                 
                 if (setCondenseUids) {
```
jwomeara: Is setCondenseUids synonymous with parseRootUids?

warehouse/query-core/src/main/java/datawave/query/index/lookup/CreateUidsIterator.java
```diff
@@ -62,8 +70,10 @@
     private static final Logger log = Logger.getLogger(CreateUidsIterator.class);
     
     public static final String COLLAPSE_UIDS = "index.lookup.collapse";
+    public static final String PARSE_ROOT_UIDS = "index.lookup.parse.root.uids";
```
jwomeara: This turns TLD parsing on or off, right?  So why not name it TLD... something?

apmoriarty: I can change that to PARSE_TLD_UIDS

warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
apmoriarty: No, not exactly...

apmoriarty:
> 
> 
> I would be in favor of having 'TLD' somewhere in the name the setting that turns this on/off. Thoughts?
> 
> Also, do we have any tests which verify that everything still works in the non-TLD case?

I can rename variables/methods from "parse root uids" to "parse tld uids". Queries run as non-tld by default, so yes we're covered in the non-tld case.

PR #683, opened 2019-12-13
Enhance JexlNodeSet bulk operations

apmoriarty:
JexlNodeSet now has optional directAdd methods to bypass generating a new node key each time a node needs to be added. This enables node key re-use in certain cases.


warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
```diff
@@ -389,7 +389,7 @@ protected IndexInfo intersect(long maxPossibilities, Iterable<IndexMatch> matchI
             for (String uid : ids.keySet()) {
                 JexlNodeSet nodeSet = new JexlNodeSet();
                 nodeSet.addAll(ids.get(uid));
-                nodeSet.addAll(infiniteNodeSet.getNodes());
```
jwomeara: imho this would look cleaner as:

```
nodeSet.addAll(infiniteNodeSet);
```

apmoriarty:
> 
> 
> Do we really want these directAdd methods to be public? I would think that we would want to prevent people from doing that as it could be abused, and it seems to me that you really just need the ability to combine sorted sets which is possible via addAll.

That's a good point, I'll make the direct add methods private and change usages of  directAddAll() to addAll().

PR #686, opened 2019-12-17
Fixes #541 - NullPointerException in FlagMaker when collectMetrics is…

### LINKED ISSUE #541: NullPointerException in FlagMaker when collectMetrics is true
In the FlagMaker, the StandaloneTaskAttemptContext is never initialized, so when we reach the place where metrics are recorded, a NullPointerException is thrown.

See: https://github.com/NationalSecurityAgency/datawave/blob/0cbc0441d3b7a9114386c0495be3380fbfc194d8/warehouse/ingest-core/src/main/java/datawave/util/flag/FlagMaker.java#L486

This may be as a result of #451 being incomplete at this point in time.

jzgithub1:
… true.   Added a null check for the ctx variable.


warehouse/ingest-core/src/main/java/datawave/util/flag/FlagMaker.java
```diff
@@ -507,7 +507,7 @@ protected File write(Collection<InputFile> flagging, FlagDataTypeConfig fc, Stri
             if (fc.getFileListMarker() == null) {
                 String sep = " ";
                 for (InputFile inFile : flagging) {
-                    if (fc.isCollectMetrics())
+                    if (fc.isCollectMetrics() && ctx != null)
```
ivakegg: Looking at the history of the FlagMaker.java, I see that the whole concept of metrics got encapsolated into a metrics object of type FlagMetrics.  In there you will see a context and the updating of metrics.  The ctx object in this class should be completely removed, and the calls at like 511 and 531 should instead be calls on the metrics object.

warehouse/ingest-core/src/main/java/datawave/util/flag/FlagMaker.java
```diff
@@ -502,13 +499,17 @@ protected File write(Collection<InputFile> flagging, FlagDataTypeConfig fc, Stri
             throw new IOException("Unable to create flag file " + f);
         }
         
+        final FileSystem fs = getHadoopFS();
+        final FlagMetrics metrics = new FlagMetrics(fs, fc.isCollectMetrics());
```
ivakegg: So now we have created a new metrics object, and updated counters therein, but never write the metrics out.  So 1) are these metrics even needed and 2) if they are then we should be writing the metrics out.

jzgithub1: The metrics get written out on line 475 in FlagMaker.java .  I added the FlagMetrics object as a parameter in the write and writeFlagFile functions so it only has to get initialized once in the processFlags function.  I checked the call stacks for those functions and it looks like we are all good now.

warehouse/ingest-core/src/main/java/datawave/util/flag/FlagMaker.java
```diff
@@ -238,9 +237,10 @@ public void run() {
     protected void processFlags() throws IOException {
         FileSystem fs = getHadoopFS();
         log.trace("Querying for files on {}", fs.getUri().toString());
+        
         for (FlagDataTypeConfig fc : fmc.getFlagConfigs()) {
-            long startTime = System.currentTimeMillis();
             String dataName = fc.getDataName();
+            final FlagMetrics metrics = new FlagMetrics(fs, fc.isCollectMetrics());
```
ivakegg: We do not want to create one metrics file across all flag files we write.  We need a new one per flag file.  You were correct passing the instance into the write() method, but not correct passing in an instance into the writeFlagFile() method.

warehouse/ingest-core/src/main/java/datawave/util/flag/FlagMaker.java
```diff
@@ -397,14 +398,15 @@ public FileVisitResult visitFile(java.nio.file.Path path, BasicFileAttributes at
...
      * @throws IOException
      */
     //@formatter:on
-    void writeFlagFile(final FlagDataTypeConfig fc, Collection<InputFile> inFiles) throws IOException {
+    void writeFlagFile(final FlagDataTypeConfig fc, Collection<InputFile> inFiles, FlagMetrics metrics) throws IOException {
         File flagFile = null;
         final FileSystem fs = getHadoopFS();
         long now = System.currentTimeMillis();
-        final FlagMetrics metrics = new FlagMetrics(fs, fc.isCollectMetrics());
+
```
ivakegg: we should still be creating a new metrics instance here.

warehouse/ingest-core/src/main/java/datawave/util/flag/FlagMaker.java
```diff
@@ -428,10 +430,10 @@ void writeFlagFile(final FlagDataTypeConfig fc, Collection<InputFile> inFiles) t
             Path first = flagging.iterator().next().getCurrentDir();
             String baseName = fmc.getFlagFileDirectory() + File.separator + df.format(now / 1000) + "_" + fc.getIngestPool() + "_" + fc.getDataName() + "_"
                             + first.getName() + "+" + flagging.size();
-            flagFile = write(flagging, fc, baseName);
+            flagFile = write(flagging, fc, baseName, metrics);
```
ivakegg: this part is good

jzgithub1:
@ivakegg Requested changes made.

PR #688, opened 2019-12-18
Flatten the query tree after parsing from string.

apmoriarty:
Flatten the query tree after parsing from string. This reduces the number of nodes that need to be visited in subsequent steps.


warehouse/query-core/src/main/java/datawave/audit/DatawaveSelectorExtractor.java
```diff
@@ -22,13 +22,14 @@
         ASTJexlScript jexlScript = null;
         
         try {
-            jexlScript = JexlASTHelper.parseJexlQuery(query.getQuery());
+            // Parse & Flatten to reduce the number of node traversals in later method calls
+            jexlScript = JexlASTHelper.parseAndFlattenJexlQuery(query.getQuery());
```
ivakegg: You need to remove the call to the TreeFlatteningRebuildingVisitor.flatten below on line 39

apmoriarty: Done.

PR #689, opened 2019-12-18
fixes #687 correctly build the upper range when not inclusive

### LINKED ISSUE #687: DatawaveFieldIndexFilterIteratorJexl.buildBoundingFiRanges() is wrong
When a range is not upper inclusive, the resulting endKey is not cast to a char so the endKey range will be a number rather than the char decremented by one + \uffff

FineAndDandy:


warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexFilterIteratorJexl.java
```diff
@@ -115,7 +115,7 @@ public String toString() {
         } else {
             String upperString = upperBound.toString();
             this.boundingFiRangeStringBuilder.append(upperString.substring(0, upperString.length() - 1));
-            this.boundingFiRangeStringBuilder.append(upperString.charAt(upperString.length() - 1) - 1);
+            this.boundingFiRangeStringBuilder.append((char) (upperString.charAt(upperString.length() - 1) - 1));
```
apmoriarty: Should line 117 also have a cast to a char?

ivakegg: line 117 is appending a string, not a char

PR #691, opened 2019-12-19
Fixes #690 - 'yarn' keys now filtered in TypeRegistry setup

### LINKED ISSUE #690: On Hadoop 3, TypeRegistry can pick up 'yarn' cruft inadvertently
M/R jobs without `ingest.data.types` defined in the conf will cause TypeRegistry to try and auto-discover data types. In that case, it may inadvertently try to load 'yarn' classes, based on keys such as  `yarn.timeline-service.reader.class`, which it tries to interpret as a DW Ingest-specific RecordReader

keith-ratcliffe:


warehouse/ingest-core/src/main/java/datawave/ingest/data/TypeRegistry.java
```diff
@@ -291,6 +291,7 @@ private TypeRegistry(Configuration config) {
         names.remove("file");
         names.remove("accumulo");
         names.remove("num");
+        names.remove("yarn");
```
javease: I know we discussed but I'm still unclear. It looks like the loop above is going to add "yarn" and yarn.timeline-service" to the set and this looks like it will only remove the "yarn" entry and not the "yarn.timeline-service" entry. Won't we still encounter the problem?

javease: I attempted a quick test locally and I still saw the error when i just did a remove of "yarn" but the error went away when i also added the remove for "yarn.timeline-service".

keith-ratcliffe: Ok. Sorry about that. The best way forward I think is for us to ditch the hard coded filter there altogether

In the meantime, as a workaround, you could easily preconfigure `ingest.data.types` in your job conf XML with the comma-delimited list of type names that you need in the TypeRegistry. Or any arbitrary names, for that matter, if you're only counting on the side effect of TypeRegistry initialization

keith-ratcliffe:
Closing this PR, but leaving issue open. Need to reconsider the approach here, e.g., implementing configurable blacklist instead...

PR #694, opened 2019-12-30
JexlNodeSet now supports equality

apmoriarty:
JexlNodeSet now supports equality, used by IndexMatch to determine equality of the nodes backing a hit in the global index.


warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexMatch.java
```diff
@@ -89,7 +89,12 @@ public boolean equals(Object obj) {
         if (null == obj)
             return false;
         if (obj instanceof IndexMatch) {
-            return uid.equals(((IndexMatch) obj).uid);
+            IndexMatch other = (IndexMatch) obj;
```
ivakegg: Can we have a case where the uid is the same and yet the node set is different?

apmoriarty: Yes. If two distinct terms (or multiple terms in different sub-trees) hit on the same document then we have a case of two identical uids and two different JexlNodeSets.

apmoriarty:
@ivakegg @jwomeara If you could review again that would be great

PR #695, opened 2020-01-02
Accumulo 2 API Updates

brianloss:
This depends on SNAPSHOT versions of the microservice API/utils and in-memory Accumulo. You will need to build the release/version2.0 branch of each of the dependent microservice projects.

Changes include:
* Replacing Connector with AccumuloClient (and closing it when done, where feasible)
* Using new, non-deprecated AccumuloInputFormat/AccumuloOutputFormat.
* Using htrace instead of accumulo trace, which is now gone.

Note that the htrace changes haven't really been tested at all.


warehouse/ingest-core/src/test/resources/datawave/ingest/mapreduce/job/IngestJob-test-type.xml
```diff
@@ -19,7 +19,7 @@
   <value>ingest-test</value>
   <description>This is the type of data being ingested.</description>
 </property>
-
+<!--
```
ivakegg: Lets get @keith-ratcliffe to look at this one

warehouse/ops-tools/config-compare/pom.xml
```diff
@@ -12,6 +12,12 @@
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
         </dependency>
+        <!-- This is normally included via hadoop-common, but we have everything excluded in the warehouse parent. -->
+        <dependency>
+            <groupId>commons-collections</groupId>
+            <artifactId>commons-collections</artifactId>
+            <version>3.2.2</version>
```
ivakegg: lets move this version at least into the dependency configuration at the top level

warehouse/query-core/src/main/java/datawave/query/util/DateIndexHelper.java
```diff
@@ -131,44 +129,31 @@ public static DateIndexHelper getInstance() {
...
-        return null;
-    }
-    
     /**
      * Initializes the instance with a provided update interval.
      * 
-     * @param connector
-     *            A Connector to Accumulo
+     * @param client
+     *            A client connectoin to Accumulo
```
ivakegg: connection

warehouse/ingest-core/src/test/resources/datawave/ingest/mapreduce/job/IngestJob-test-type.xml
brianloss: Turns out this was un-necessary. Must have been something else broken while I was in the middle of making changes. I've removed this change.

warehouse/ops-tools/config-compare/pom.xml
brianloss: Fixed in latest commit.

warehouse/query-core/src/main/java/datawave/query/util/DateIndexHelper.java
brianloss: Fixed in latest commit.

PR #727, opened 2020-01-24
Updated the ivarators to wait to add runnables to the ivarator execut…

jwomeara:
…or until those runnables have a valid source.


warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -888,20 +902,38 @@ protected boolean isTimedOut() {
     /**
      * Get a source copy. If the source is setup as a ThreadLocalPooledSource, then no copy is needed.
```
ivakegg: fix comment

ivakegg: perhaps add comment that this should only be used for the unsorted case

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1328,7 +1329,8 @@ protected IteratorBuildingVisitor createIteratorBuildingVisitor(Class<? extends
                         .setIvaratorCacheBufferSize(this.getIvaratorCacheBufferSize())
                         .setIvaratorCacheScanPersistThreshold(this.getIvaratorCacheScanPersistThreshold())
                         .setIvaratorCacheScanTimeout(this.getIvaratorCacheScanTimeout()).setMaxRangeSplit(this.getMaxIndexRangeSplit())
-                        .setIvaratorMaxOpenFiles(this.getIvaratorMaxOpenFiles()).setIvaratorSources(this, this.getMaxIvaratorSources())
```
ivakegg: Does the QueryIterator still need to implement SourceFactory after these changes?

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1328,7 +1329,8 @@ protected IteratorBuildingVisitor createIteratorBuildingVisitor(Class<? extends
...
+                        .setIvaratorMaxOpenFiles(this.getIvaratorMaxOpenFiles()).setIvaratorSource(this.sourceForDeepCopies)
```
ivakegg: maybe setIvaratorSource should be setUnsortedIvaratorSource

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1337,6 +1339,14 @@ protected IteratorBuildingVisitor createIteratorBuildingVisitor(Class<? extends
         // TODO: .setStatsPort(this.statsdHostAndPort);
     }
     
+    protected ArrayBlockingQueue<SortedKeyValueIterator<Key,Value>> createIvaratorSourcePool(SortedKeyValueIterator<Key,Value> sourceForDeepCopies,
+                    int maxIvaratorSources, IteratorEnvironment env) {
+        ArrayBlockingQueue<SortedKeyValueIterator<Key,Value>> ivaratorSourcePool = new ArrayBlockingQueue<>(maxIvaratorSources, true);
+        for (int i = 0; i < maxIvaratorSources; i++)
+            ivaratorSourcePool.add(sourceForDeepCopies.deepCopy(env));
```
ivakegg: does this deepCopy do the opening of the RFiles, or is that lazy and not done until the first seek?  Just curious if we need to create these lazily.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownLargeFieldedListsVisitor.java
```diff
@@ -127,14 +127,14 @@ public Object visit(ASTOrNode node, Object data) {
                 List<JexlNode> markers = new ArrayList<>();
                 
                 try {
-                    // if we have an hdfs cache directory and if past the fst threshold, then create the fst and replace the list with an assignment
-                    if (fstHdfsUri != null && (eqNodes.size() >= config.getMaxOrExpansionFstThreshold())) {
+                    // if we have an hdfs cache directory and if past the fst/list threshold, then create the fst/list and replace the list with an assignment
+                    if (eqNodes.size() >= config.getMaxOrExpansionThreshold()) {
+                        markers.add(ExceededOrThresholdMarkerJexlNode.createFromValues(field, values));
+                        eqNodes = null;
+                    } else if (fstHdfsUri != null && (eqNodes.size() >= config.getMaxOrExpansionFstThreshold())) {
```
ivakegg: I think the fst check needs to be first the way we were doing before.  If we are favoring the fst too much then the fst threshold needs to be set larger.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
jwomeara: Yes, because it is used by the SourceManager which determines how many non-ivarator iterators can be made.

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
jwomeara: done.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
jwomeara: sure.  done.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
jwomeara: Now you've done it...  Fix coming soon...

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownLargeFieldedListsVisitor.java
jwomeara: i suppose you're right.  fixed.

PR #735, opened 2020-01-31
fixes #734: Ensure we move to the correct next field when using the reverse index

### LINKED ISSUE #734: DiscoveryLogic times out when searching reverse index
The DiscoveryLogic seems to take forever to scan the shardReverseIndex when given a regex with a leading wildcard.

ivakegg:


hlgp:
Very nice.  How about a unit test?

ivakegg:
In creating the unit test cases, I was unable to reproduce the error until I figured out that there was a second error in the DiscoveryIterator.  I think I now have this one complete.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryLogic.java
```diff
@@ -369,6 +372,9 @@ public ShardIndexQueryTable clone() {
             String literal = literalAndField.getKey(), field = literalAndField.getValue();
             // if we're _ANYFIELD_, then use null when making the literal range
             field = Constants.ANY_FIELD.equals(field) ? null : field;
+            if (field != null) {
+                familiesToSeek.add(new Text(field));
+            }
```
jwomeara: Yikes!  Good catch.  Does this mean we were potentially returning things that we shouldn't have before this fix?

PR #737, opened 2020-02-04
Fixes #703 - Created query wizard in Quickstart WebUI

### LINKED ISSUE #703: We need a simple query UI
The DATAWAVE quickstart should also stand up a simple UI through which a user can send a query and get the results.  I expect the use case would be something like this:

1) enter query and parameters and hit submit button -> returns a page containing the query plan and a next button
2) hit next button -> returns a page containing a page of results and a next button
3) repeat 2 until there are no more results in which case you get a close button

jzgithub1:
Created a three step query wizard in the Quickstart WebUI (https://localhost:8443/DataWave/doc).
It creates the specific form required for each query type.  Although the "begin" and "end" options in the form may appear in the optional parameter section they should usually be filled in ( use values: 19700101, 20990101).   Also make sure that the proper query syntax is selected in the optional parameter section.  This feature is accessed in by selecting the "Datawave" main menu item and then selecting the "Basic Query" submenu item.  The  "auths" and "visibilities" are pre-populated for ease of use and can be edited.


keith-ratcliffe:
Cool. Thanks for working on this!

web-services/query/src/main/java/datawave/webservice/query/runner/BasicQueryBean.java
```diff
@@ -0,0 +1,384 @@
...
+        String queryId = createResponse.getResult();
+        CreateQuerySessionIDFilter.QUERY_ID.set(queryId);
+        queryWizardStep3Response.setQueryId(queryId);
+        GenericResponse<String> planResponse;
+        try {
+            planResponse = queryExecutor.plan(queryId);
+        } catch (Exception e) {
+            queryWizardStep3Response.setErrorMessage(e.getMessage());
+            return queryWizardStep3Response;
+        }
```
keith-ratcliffe: When I ran this, I got a "Datawave could not generate a plan for the query" message when I tried ContentQuery logic type, which makes sense, because non-ShardQueryLogic types like ContentQuery typically won't have a planner involved. But in this case, you get stuck on the /showQueryWizardStep3 page with no way to retrieve the content result.

To replicate, do an EventQuery to pull back wikipedia records (e.g., "anarchy" unfielded should work, if you have the default data loaded), copy the RECORD_ID field's value (e.g., 20130305_0/enwiki/ibrtlu.qead3h.uz468c), then just plug that value into the query field for a ContentQuery

keith-ratcliffe: I think in all cases you should have a Query ID though, so it'd be good to display that on the step3 page. E.g., its extremely handy to plug that value into a QueryMetricsQuery to troubleshoot queries as needed

jzgithub1: @keith-ratcliffe ,  initially this is working well and was tested mainly with EdgeQuery and EventQuery with the examples from the QuickStart guide since I have limited experience.

The queries I used were:
EventQuery    GENRES:action OR GENRES:adv*   (LUCENE)
EdgeQuery    SOURCE == 'kevin bacon' && TYPE == 'TV_COSTARS'   (JEXL) 

Those queries will get you some output for sure if you haven't seen any yet.  I will work on the suggestions you made but I have to leave this issue for a few days to get back on Accumulo .

web-services/deploy/docs/docs/index.html
```diff
@@ -32,6 +32,7 @@
                         <li><a href="/DataWave/DataDictionary">Data Dictionary</a></li>
                         <li><a href="/DataWave/EdgeDictionary">Edge Dictionary</a></li>
                         <li><a href="/DataWave/Query/listQueryLogic">Deployed Query Logics</a></li>
+                        <li><a href="/DataWave/BasicQuery/showQueryWizard">Basic Query</a></li>
```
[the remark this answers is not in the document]
jzgithub1: You are right.  I should just be Query.

web-services/query/src/main/java/datawave/webservice/query/runner/BasicQueryBean.java
```diff
@@ -0,0 +1,388 @@
...
+            createResponse = queryExecutor.createQuery(logicName, queryParameters, httpHeaders);
+        } catch (Exception e) {
+            queryWizardStep3Response.setErrorMessage(e.getMessage());
+            return queryWizardStep3Response;
+        }
+        String queryId = createResponse.getResult();
+        CreateQuerySessionIDFilter.QUERY_ID.set(queryId);
+        queryWizardStep3Response.setQueryId(queryId);
+        
+        if (!logicName.equals("ContentQuery")) {
```
keith-ratcliffe: Sorry, finally getting back to reviewing this. For the comparison here, it would be better to use your queryLogicFactory instance to get the logicClass, rather than use the logicName. The name is subject to change based on user preference. For example, the user might decide they like the name "ContentLookup" better and reconfigure QLF.xml. But if we rename **datawave.query.tables.content.ContentQueryTable** that'd at least get caught by the compiler here

Second issue: I did try out the update, and I did get past the original issue that I reported, but I then noticed the ContentQuery response fields get truncated to just a few visible chars in the results view. However, the d-column entries in this case are not well-suited for row/col table viewing anyway....because they can be huge, they're base64 encoded, etc. So, I think I'm fine with leaving that problem for another release down the road

jzgithub1: Thank you @keith-ratcliffe.  I will get these changes made soon.

jzgithub1: @keith-ratcliffe , I am testing a solution to the logic name issue you mention above.   Basically I get the query logic class from the queryLogicFactory bean.  The problem is that I have to string match the object's class name against "datawave.query.tables.content.ContentQueryTable" which will be detected at runtime, not compile time.  I could do an "instanceof" comparison with ContentQueryTable but that will cause a circular dependency issue with datawave-query-core and datawave-ws-query (by having to import ContentQueryTable in my bean) .  I will push the code for you to review.

web-services/query/src/main/java/datawave/webservice/query/runner/BasicQueryBean.java
```diff
@@ -0,0 +1,408 @@
...
+            try {
+                planResponse = queryExecutor.plan(queryId);
+            } catch (Exception e) {
+                queryWizardStep3Response.setErrorMessage(e.getMessage());
+                return queryWizardStep3Response;
+            }
+            
+            queryWizardStep3Response.setQueryPlan(planResponse.getResult());
+        } else
+            queryWizardStep3Response.setQueryPlan(" No query plan for is required for this query");
```
keith-ratcliffe: Change to:
```
queryWizardStep3Response.setQueryPlan("No plan required for this query");
```

web-services/query/src/main/java/datawave/webservice/query/runner/BasicQueryBean.java
```diff
@@ -0,0 +1,408 @@
...
+        
+        return queryWizardStep3Response;
+    }
+    
+    /**
+     * Gets the next page of results from the query object. If the object is no longer alive, meaning that the current session has expired, then this fail. The
+     * response object type is dynamic, see the listQueryLogic operation to determi"ne what the response type object will be.
```
keith-ratcliffe: Typo: `determi"ne`

Issue #740, opened 2020-02-06
Handle orphaned files from killed MapFileLoader


JustDoWhitt:
When MapFileLoader is killed, orphaned files can be left in flagged directory.

SethSmucker:
Fixed in #980, more cleanup in #1153 and #1191. Closing.

PR #741, opened 2020-02-06
#740: Add Cleanup_File_Marker when files are ready for cleanup. Clean…

JustDoWhitt:
… directories upon startup.


brianloss:
Looks like there are format problems in BulkIngestMapFileLoader and BulkIngestMapFileLoaderTest. Please build locally and update your PR.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/BulkIngestMapFileLoader.java
```diff
@@ -489,6 +496,26 @@ public void run() {
         log.info("Bulk map file loader shutting down.");
     }
     
+    private void cleanJobDirectoriesOnStartup() throws IOException {
+        Path[] cleanupDirectories = getJobDirectories(new Path(workDir, jobDirPattern + '/' + CLEANUP_FILE_MARKER));
+        for (int i = 0; i < cleanupDirectories.length; i++) {
+            getFileSystem(srcHdfs).delete(cleanupDirectories[i], true);
```
hlgp: Again, you cannot simply assume that we are ready to delete the directories at this point.  We don't know if we successfully moved all the source files yet.

JustDoWhitt: I thought that as the point of marking them as job.cleanup. Meaning we know they are successfully moved, therefore they can be deleted?

So it's possible they haven't moved so we would like to run MarkSourceFilesLoaded() again?

hlgp: No, reread your ticket.  The point is that we've seen jobs only partially complete, which leaves source files hanging around.  Notice that you are marking it as job.cleanup before we move the files, not after.  In the past, if the process was interrupted there, the state would still say job.loading and we had no idea at what point the process terminated.  Did it actually finish loading?  Did it finish moving its files and just not get to the final cleanup step?  We had to do manual verification each time and often the job was successful, but didn't finish moving all its files.  Job.cleanup allows us to safely move the files and cleanup automatically.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/BulkIngestMapFileLoader.java
```diff
@@ -1127,6 +1162,19 @@ public void markSourceFilesLoaded(Path jobDirectory) throws IOException {
         }
     }
     
+    public boolean markDirectoryForCleanup(Path jobDirectory, URI destFs) {
+        boolean success = false;
+        try {
+            success = getFileSystem(destFs).rename(new Path(jobDirectory, LOADING_FILE_MARKER), new Path(jobDirectory, CLEANUP_FILE_MARKER));
+            System.out.println("renamed successful");
```
hlgp: use the logger, not system.out

PR #743, opened 2020-02-11
Improvements to various quickstart scripts for execution on Mac OS X

drewfarris:
-  For Yarn, `MR_INTER_DIR` and `MR_DONE_DIR` are on HDFS, so no need to use the `DW_CLOUD_DATA` directory as a prefix for these.
-  Eliminated whitespace in `-d` arg for various process checking functions that would cause malfunction on OSX.
- Added explicit `-i'bak'` for `sed` command in `accumulo/install.sh` to address malfunction on OSX.
-  Changed process names in `pgrep` targets for Hadoop to work around process name length limitation on OSX.


drewfarris:
@keith-ratcliffe - you're probably the best person to review these.

drewfarris:
Probably needs some more tweaking:

```
[DW-INFO] - Hadoop capacity-scheduler.xml written
sed: 1: "/Users/andrewfarris/pro ...": command a expects \ followed by text
mkdir: /jobhist/inter: Permission denied
[DW-FATAL] - [assertCreateDir] configured base directory /jobhist/inter does not exist
Aborting install.sh
```

contrib/datawave-quickstart/bin/services/accumulo/install.sh
```diff
@@ -59,9 +59,9 @@ fi
 assertCreateDir "${DW_ACCUMULO_JVM_HEAPDUMP_DIR}"
 
 # Update tserver and other options in accumulo-env.sh
-sed -i "s~\(ACCUMULO_TSERVER_OPTS=\).*$~\1\"${DW_ACCUMULO_TSERVER_OPTS}\"~g" ${DW_ACCUMULO_CONF_DIR}/accumulo-env.sh
-sed -i "s~\(export JAVA_HOME=\).*$~\1\"${JAVA_HOME}\"~g" ${DW_ACCUMULO_CONF_DIR}/accumulo-env.sh
-sed -i "s~\(export ACCUMULO_MONITOR_OPTS=\).*$~\1\"\${POLICY} -Xmx2g -Xms512m\"~g" ${DW_ACCUMULO_CONF_DIR}/accumulo-env.sh
+sed -i'bak' "s~\(ACCUMULO_TSERVER_OPTS=\).*$~\1\"${DW_ACCUMULO_TSERVER_OPTS}\"~g" ${DW_ACCUMULO_CONF_DIR}/accumulo-env.sh
+sed -i'bak' "s~\(export JAVA_HOME=\).*$~\1\"${JAVA_HOME}\"~g" ${DW_ACCUMULO_CONF_DIR}/accumulo-env.sh
+sed -i'bak' "s~\(export ACCUMULO_MONITOR_OPTS=\).*$~\1\"\${POLICY} -Xmx2g -Xms512m\"~g" ${DW_ACCUMULO_CONF_DIR}/accumulo-env.sh
```
ivakegg: Won't this leave backup files around that we need to clean up?

alerman: I think you can do `sed -i''`

brianloss: Yup, that seems to work in my local testing. (Tested in a CentOS docker container too in order to make sure that doesn't break there.)

keith-ratcliffe: Not an issue, really. Any cruft in DW_ACCUMULO_CONF_DIR will get obliterated automatically on an uninstall/reinstall, along with any other paths created by the quickstart.

keith-ratcliffe:
Weird. Was expecting the failed gitlab pipeline to be due to the '/jobhist/inter: Permission denied' issue failing the deploy stage, but it was actually test failures in DatawaveFieldIndexIteratorJexlTest. Can't be due to these changes, so retrying it now to see if it was a fluke

keith-ratcliffe:
> Probably needs some more tweaking:
> 
> ```
> [DW-INFO] - Hadoop capacity-scheduler.xml written
> sed: 1: "/Users/andrewfarris/pro ...": command a expects \ followed by text
> mkdir: /jobhist/inter: Permission denied
> [DW-FATAL] - [assertCreateDir] configured base directory /jobhist/inter does not exist
> Aborting install.sh
> ```
These are from **bin/services/hadoop/install.sh** 

I didn't observe the **sed: 1** issue when I ran these changes locally on CentOS, so seems like OSX must not like something about
```
sed -i "s|.*\(export JAVA_HOME=\).*|\1${JAVA_HOME}|g" ${HADOOP_CONF_DIR}/hadoop-env.sh
```
The **mkdir** problem can be fixed by just deleting these two lines, since the local dirs aren't needed. They get created automatically in HDFS by hadoop/yarn...
```
assertCreateDir "${DW_HADOOP_MR_INTER_DIR}"
assertCreateDir "${DW_HADOOP_MR_DONE_DIR}"
```

contrib/datawave-quickstart/bin/services/hadoop/install.sh
```diff
@@ -89,13 +89,10 @@ else
 fi
 
 # Ensure that $JAVA_HOME is observed by all hadoop scripts
-sed -i "s|.*\(export JAVA_HOME=\).*|\1${JAVA_HOME}|g" ${HADOOP_CONF_DIR}/hadoop-env.sh
+sed -i'.bak' "s|.*\(export JAVA_HOME=\).*|\1${JAVA_HOME}|g" ${HADOOP_CONF_DIR}/hadoop-env.sh && rm ${HADOOP_CONF_DIR}/hadoop-env.sh.bak
```
brianloss: Why not just `sed -i ''` instead of `sed -i '.bak'` and then you don't have to remove the `.bak` file separately?

contrib/datawave-quickstart/bin/services/accumulo/install.sh
```diff
@@ -59,9 +59,9 @@ fi
 assertCreateDir "${DW_ACCUMULO_JVM_HEAPDUMP_DIR}"
 
 # Update tserver and other options in accumulo-env.sh
-sed -i "s~\(ACCUMULO_TSERVER_OPTS=\).*$~\1\"${DW_ACCUMULO_TSERVER_OPTS}\"~g" ${DW_ACCUMULO_CONF_DIR}/accumulo-env.sh
-sed -i "s~\(export JAVA_HOME=\).*$~\1\"${JAVA_HOME}\"~g" ${DW_ACCUMULO_CONF_DIR}/accumulo-env.sh
-sed -i "s~\(export ACCUMULO_MONITOR_OPTS=\).*$~\1\"\${POLICY} -Xmx2g -Xms512m\"~g" ${DW_ACCUMULO_CONF_DIR}/accumulo-env.sh
+sed -i'.bak' "s~\(ACCUMULO_TSERVER_OPTS=\).*$~\1\"${DW_ACCUMULO_TSERVER_OPTS}\"~g" ${DW_ACCUMULO_CONF_DIR}/accumulo-env.sh && rm ${DW_ACCUMULO_CONF_DIR}/accumulo-env.sh.bak
+sed -i'.bak' "s~\(export JAVA_HOME=\).*$~\1\"${JAVA_HOME}\"~g" ${DW_ACCUMULO_CONF_DIR}/accumulo-env.sh && rm ${DW_ACCUMULO_CONF_DIR}/accumulo-env.sh.bak
+sed -i'.bak' "s~\(export ACCUMULO_MONITOR_OPTS=\).*$~\1\"\${POLICY} -Xmx2g -Xms512m\"~g" ${DW_ACCUMULO_CONF_DIR}/accumulo-env.sh && rm ${DW_ACCUMULO_CONF_DIR}/accumulo-env.sh.bak
```
brianloss: Why not just `sed -i ''` instead of `sed -i '.bak'` and then you don't have to remove the `.bak` file separately?

contrib/datawave-quickstart/bin/services/hadoop/install.sh
```diff
@@ -89,13 +89,10 @@ else
...
 
 verifySSHConfig
 
-assertCreateDir "${DW_HADOOP_MR_INTER_DIR}"
-assertCreateDir "${DW_HADOOP_MR_DONE_DIR}"
-
```
brianloss: What's the reason for removing these lines?

contrib/datawave-quickstart/bin/services/accumulo/install.sh
drewfarris: It would have been nice if `sed -i''` would have worked, but it didn't. I've added a `sed ... && rm ....bak` to the end of these to clean up the backup file if the sed completes successfully.

drewfarris: I'll re-run and capture the specific error..

contrib/datawave-quickstart/bin/services/hadoop/install.sh
drewfarris: They're not needed because the directories are on HDFS - `assertCreateDir` checks local filesystem directories.

brianloss: Ah, sounds good. Thanks.

drewfarris:
I have come to hate the subtle differences between common linux and original osx executables. The bottom line here is that I'm relying on `pgrep` and `readlink' from the homebrew packages `proctools` and `coreutils`. I suspect this is fine because homebrew is sufficiently ubiquitous across OS X users.

contrib/datawave-quickstart/bin/services/accumulo/install.sh
drewfarris: Here's the bizarre error that I get regardless of whether I'm using a bare `-i` or `-i''`
```
sed: 1: "/Users/andrewfarris/pro ...": command a expects \ followed by text
```

drewfarris: Ultimately, this works:
```
sed -i'' -e 's|.*\(export JAVA_HOME=\).*|\1'${JAVA_HOME}'|g' ${HADOOP_CONF_DIR}/hadoop-env.sh
```
Whereas this fails:
```
sed -i'' "s|.*\(export JAVA_HOME=\).*|\1${JAVA_HOME}|g" ${HADOOP_CONF_DIR}/hadoop-env.sh`
```

PR #749, opened 2020-02-13
Fixed an issue were full field index scan ranges might exclude the fi…

jwomeara:
…nal day in the range.


jwomeara:
@ivakegg @FineAndDandy Would like to get your take on this.

PR #759, opened 2020-02-25
Bugfix/issue 740 Make it so orphan files are not left during bulk ingest

JustDoWhitt:


warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/BulkIngestMapFileLoader.java
```diff
@@ -489,6 +496,27 @@ public void run() {
         log.info("Bulk map file loader shutting down.");
     }
     
+    private void cleanJobDirectoriesOnStartup() throws IOException {
+        Path[] cleanupDirectories = getJobDirectories(new Path(workDir, jobDirPattern + '/' + CLEANUP_FILE_MARKER));
+        for (int i = 0; i < cleanupDirectories.length; i++) {
+            markSourceFilesLoaded(cleanupDirectories[i]);
```
ivakegg, line 502: I think we may have issues here if we have multiple map loaders working against the same warehouse.  Perhaps   this should attempt the "grab control" of the directory by renaming the cleanup file and only when successfully doing that will actually implement the cleanup.

JustDoWhitt: I believe I made changes that address this. Please let me know what you think.

FineAndDandy: think this is covered with the REMOVE_FILE_MARKER transition

FineAndDandy (approved):

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/BulkIngestMapFileLoader.java
```diff
@@ -489,6 +498,40 @@ public void run() {
...
+            }
+            
+            if (success) {
+                markSourceFilesLoaded(cleanupDirectories[i]);
+                getFileSystem(srcHdfs).delete(cleanupDirectories[i], true);
+                runCleanUpScript(cleanupDirectories[i], true);
+            }
+        }
+        
+        Path[] failedDirectories = getJobDirectories(new Path(workDir, jobDirPattern + '/' + LOADING_FILE_MARKER));
```
hlgp, line 522: If we start an additional loader later on, there will be jobs already in loading.  Best to handle this case from the start script after we verify there are no loaders currently running.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/BulkIngestMapFileLoader.java
```diff
@@ -489,6 +498,40 @@ public void run() {
...
+    // take ownership by renaming files and then running cleanup script
+    private void cleanJobDirectoriesOnStartup() throws IOException {
+        Path[] cleanupDirectories = getJobDirectories(new Path(workDir, jobDirPattern + '/' + CLEANUP_FILE_MARKER));
+        for (int i = 0; i < cleanupDirectories.length; i++) {
+            boolean success = false;
+            
+            try {
+                success = getFileSystem(srcHdfs).rename(new Path(cleanupDirectories[i], CLEANUP_FILE_MARKER),
+                                new Path(cleanupDirectories[i], REMOVE_FILE_MARKER));
+                log.info("Renamed " + cleanupDirectories[i] + '/' + CLEANUP_FILE_MARKER + " to " + REMOVE_FILE_MARKER);
```
hlgp, line 510: There's no need to "take ownership" here.  If multiple loaders try to move the files, hdfs provides the atomicity.  Adding another "state" for ownership is yet another state we need to check for at startup if we do it this way.

PR #760, opened 2020-02-25
Fixes #705 - Debug queries during exection of the VisitorFunction.apply method

### LINKED ISSUE #703: We need a simple query UI

### LINKED ISSUE #705: We need the ability to debug queries out of the VisitorFunction
The VisitorFunction is the final place where a specific query plan may be updated before sending on its way to the tservers against the shard table (QueryIterator).  We need the ability to debug what those queries are and which ranges they are against.  As this could be quite numerous, this debugging should be off by default.

jzgithub1:
Added debugging logic to the apply function in the VisitorFunction class.  I am not sure that anything else needs to be done.


warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
```diff
@@ -246,6 +241,14 @@ public ScannerChunk apply(@Nullable ScannerChunk input) {
                     if (log.isDebugEnabled()) {
```
ivakegg: Given the debugging below, I think we can remove the -> newQuery  part of this debug statement.

ivakegg:
@brianloss @jwomeara @FineAndDandy Can somebody else please take a quick look at this one.  Thanks.

PR #761, opened 2020-02-26
Return query plans in random order.

apmoriarty:
Add option to return query plans in random order for better scan distribution.


warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -481,6 +481,22 @@ private DateIndexHelper getDateIndexHelper(final ShardQueryConfiguration config)
...
+                queryPlanComparators = new ArrayList<>();
+            
+            queryPlanComparators.add(new DefaultQueryPlanComparator() {
+                @Override
+                public int compare(QueryPlan o1, QueryPlan o2) {
+                    int left = o1.getQueryString().hashCode() + o1.getRanges().iterator().next().getStartKey().hashCode();
+                    int right = o2.getQueryString().hashCode() + o2.getRanges().iterator().next().getStartKey().hashCode();
+                    return left - right;
+                }
+            });
```
jwomeara: I love the idea of a random ordering of the ranges, but I do not like the idea that two identical queries would run in exactly the same order.  Check out this stack overflow post which gives an example of a way to create a uniquely random ordering using a comparator even if against the same input: https://stackoverflow.com/questions/29241746/shuffle-random-comparator

warehouse/query-core/src/main/java/datawave/query/planner/ThreadedRangeBundlerIterator.java
```diff
@@ -91,8 +91,12 @@ private ThreadedRangeBundlerIterator(Builder builder) {
...
         int maxCapacity = (int) maxRanges > 0 ? (int) maxRanges : 1000;
         if (builder.getQueryPlanComparators() != null && !builder.getQueryPlanComparators().isEmpty()) {
-            Comparator<QueryPlan> comparator = (builder.getQueryPlanComparators().size() > 1) ? new MultiComparator<>(builder.getQueryPlanComparators())
-                            : builder.getQueryPlanComparators().iterator().next();
+            Comparator<QueryPlan> comparator;
+            if (builder.getQueryPlanComparators().size() > 1) {
+                comparator = new MultiComparator<>(builder.getQueryPlanComparators());
+            } else {
+                comparator = builder.getQueryPlanComparators().iterator().next();
+            }
```
jwomeara: Why exactly are we doing this?

jwomeara:
If we enable this by default, should we not also set a better default value for the amount of time to wait to buffer ranges?

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
apmoriarty: Sounds reasonable

warehouse/query-core/src/main/java/datawave/query/planner/ThreadedRangeBundlerIterator.java
apmoriarty: Readability

apmoriarty:
Closing PR.

PR #764, opened 2020-02-26
fixes #763 Add UUIDTransform and parser/LookupUUIDTune processing

### LINKED ISSUE #763: Allow UUIDTransforms as part of UUIDType
Allow a configurable number of transforms to be applied to UUIDTypes. Also allow parsers to be injected into the LookupUUIDTune profiles.

FineAndDandy:


warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/LuceneToJexlUUIDQueryParser.java
```diff
@@ -2,7 +2,11 @@
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
+import com.sun.org.apache.bcel.internal.generic.Select;
```
ivakegg: This import looks suspicious.  Is this really what you wanted?

warehouse/core/src/main/java/datawave/query/data/UUIDTransform.java
```diff
@@ -0,0 +1,43 @@
+package datawave.query.data;
+
+import java.util.regex.Pattern;
+
+/**
+ * Transform UUID rule, defines a pattern and replacement to be applied to a UUID value
+ */
+public class UUIDTransform {
```
ivakegg: It seems to me that we could create a dedicated transform for what you want.  Using a regex and replacement mechanism seems expensive.

PR #771, opened 2020-03-02
Failed Query Plans need to be sent to metrics

JustDoWhitt:


warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -442,6 +431,19 @@ private DateIndexHelper getDateIndexHelper(final ShardQueryConfiguration config)
...
+        
+        Tuple2<CloseableIterable<QueryPlan>,Boolean> queryRanges = getQueryRanges(scannerFactory, metadataHelper, config, queryTree);
+        
+        // a full table scan is required if
+        final boolean isFullTable = queryRanges.second();
+        
+        // abort if we cannot handle full table scans
+        if (isFullTable && !config.getFullTableScanEnabled()) {
+            PreConditionFailedQueryException qe = new PreConditionFailedQueryException(DatawaveErrorCode.FULL_TABLE_SCAN_REQUIRED_BUT_DISABLED);
+            throw new FullTableScansDisallowedException(qe);
```
ivakegg: Won't the query plan fail to be set if this exeception is thrown?

JustDoWhitt: Oh, you're right, it would just be whatever "config.setQueryString(jexlQueryString);" is from the ShardQueryLogic.

I can update it to set the plan there before that exception is thrown.

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -914,14 +914,6 @@ private QueryData validateQuery(String queryLogicName, MultivaluedMap<String,Str
...
                 log.error("Exception occured while closing query logic; may be innocuous if scanners were running.", e);
             }
-            
-            if (null != connection) {
-                try {
-                    connectionFactory.returnConnection(connection);
-                } catch (Exception e) {
-                    log.error("Error returning connection on failed create", e);
-                }
-            }
```
FineAndDandy, line 924: why remove this?

FineAndDandy (changes requested):

matthpeterson:
See https://github.com/NationalSecurityAgency/datawave/pull/1290

PR #774, opened 2020-03-03
Updated the tree flattener to better handle queries with large or'd l…

jwomeara:
…ists, and updated the selector extractor to flatten before extracting selectors.


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/TreeFlatteningRebuilder.java
```diff
@@ -182,6 +190,32 @@ private JexlNode copyTree(JexlNode node, Deque<JexlNode> postOrderDeque) {
...
+        
+        while (!stack.isEmpty()) {
+            JexlNode currNode = stack.pop();
+            
+            // only add children if
+            // 1) this is the original node, or
+            // 2) this node is the same type as the root node and
+            // -- a) this is an OR node, or
+            // -- b) this is an AND node (but not a bounded range)
+            if (currNode == node || node.getClass().isInstance(currNode)
```
ivakegg: Make the operation precedence obvious between the initial || and the subsequent &&

jwomeara: So... an extra set of parens???  I thought the comment made it obvious.

jwomeara: Fixed.  What more do you want from me!!!???  I seek your approval!

PR #775, opened 2020-03-04
Better way to configure unique terms within an expression

apmoriarty:
Push the configuration for the UniqueExpressionTermsVisitor down to the logic level vs. the DefaultQueryPlanner level.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/config/LookupUUIDTune.java
+++ b/warehouse/query-core/src/main/java/datawave/query/config/LookupUUIDTune.java
@@ -125,7 +125,6 @@ public class LookupUUIDTune implements Profile {
             if (maxShardsPerDayThreshold != -1) {
                 rsqc.setShardsPerDayThreshold(maxShardsPerDayThreshold);
             }
-            rsqc.setEnforceUniqueTermsWithinExpressions(enforceUniqueTermsWithinExpressions);
             
             // we need this since we've finished the deep copy already
             rsqc.setSpeculativeScanning(speculativeScanning);
--- a/warehouse/query-core/src/main/java/datawave/query/config/LookupUUIDTune.java
+++ b/warehouse/query-core/src/main/java/datawave/query/config/LookupUUIDTune.java
@@ -54,8 +54,7 @@ public class LookupUUIDTune implements Profile {
             rsq.setSpeculativeScanning(speculativeScanning);
             rsq.setCacheModel(enableCaching);
             rsq.setPrimaryToSecondaryFieldMap(primaryToSecondaryFieldMap);
-            if (enforceUniqueTermsWithinExpressions)
-                rsq.setEnforceUniqueTermsWithinExpressions(true);
+            rsq.setEnforceUniqueTermsWithinExpressions(enforceUniqueTermsWithinExpressions);
             
             if (querySyntaxParsers != null) {
                 rsq.setQuerySyntaxParsers(querySyntaxParsers);
@@ -68,8 +67,6 @@ public class LookupUUIDTune implements Profile {
                 SeekingQueryPlanner planner = new SeekingQueryPlanner();
                 planner.setMaxFieldHitsBeforeSeek(maxFieldHitsBeforeSeek);
                 planner.setMaxKeysBeforeSeek(maxKeysBeforeSeek);
-                if (enforceUniqueTermsWithinExpressions)
-                    planner.setEnforceUniqueTermsWithinExpressions(true);
                 
                 rsq.setQueryPlanner(planner);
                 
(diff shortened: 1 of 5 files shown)


warehouse/query-core/src/main/java/datawave/query/config/LookupUUIDTune.java
```diff
@@ -54,6 +54,8 @@ public void configure(BaseQueryLogic<Entry<Key,Value>> logic) {
             rsq.setSpeculativeScanning(speculativeScanning);
             rsq.setCacheModel(enableCaching);
             rsq.setPrimaryToSecondaryFieldMap(primaryToSecondaryFieldMap);
+            if (enforceUniqueTermsWithinExpressions)
+                rsq.setEnforceUniqueTermsWithinExpressions(true);
```
jwomeara: Is the conditional here really necessary?

warehouse/query-core/src/main/java/datawave/query/config/LookupUUIDTune.java
```diff
@@ -66,6 +68,9 @@ public void configure(BaseQueryLogic<Entry<Key,Value>> logic) {
                 SeekingQueryPlanner planner = new SeekingQueryPlanner();
                 planner.setMaxFieldHitsBeforeSeek(maxFieldHitsBeforeSeek);
                 planner.setMaxKeysBeforeSeek(maxKeysBeforeSeek);
+                if (enforceUniqueTermsWithinExpressions)
+                    planner.setEnforceUniqueTermsWithinExpressions(true);
```
jwomeara: Same comment as above.

warehouse/query-core/src/main/java/datawave/query/config/LookupUUIDTune.java
```diff
@@ -125,6 +130,8 @@ public void configure(GenericQueryConfiguration configuration) {
             if (maxShardsPerDayThreshold != -1) {
                 rsqc.setShardsPerDayThreshold(maxShardsPerDayThreshold);
             }
+            if (enforceUniqueTermsWithinExpressions)
+                rsqc.setEnforceUniqueTermsWithinExpressions(true);
```
jwomeara: Same comment as above.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -746,7 +746,8 @@ protected ASTJexlScript updateQueryTree(ScannerFactory scannerFactory, MetadataH
         
         stopwatch.stop();
         
-        if (enforceUniqueTermsWithinExpressions) {
+        // Enforce if either DefaultQueryPlanner or ShardQueryConfig is true
+        if (enforceUniqueTermsWithinExpressions || config.getEnforceUniqueTermsWithinExpressions()) {
```
jwomeara: Why does this term exist in two places?  Why not only have it in the shard query configuration?

apmoriarty: My understanding is that the LookupUUIDTune is used to override a select number of query configs, most of the options here are duplicates.

apmoriarty:
> 
> 
> Why have this property in both places? I can guarantee you that's going to bite someone in the future when they think they've enabled/disabled it, but they missed the other config location. Also, why does it need to be moved to the shard config at all? Both the shard config and planner config are accessible in QueryLogicFactory.xml.

My understanding is that the LookupUUIDTune is used to override a select number of query configs, most of the options here are duplicates.

warehouse/query-core/src/main/java/datawave/query/config/LookupUUIDTune.java
apmoriarty: Nope, resolving in next commit.

warehouse/query-core/src/main/java/datawave/query/config/LookupUUIDTune.java
apmoriarty: Resolved

warehouse/query-core/src/main/java/datawave/query/config/LookupUUIDTune.java
```diff
@@ -125,6 +125,8 @@ public void configure(GenericQueryConfiguration configuration) {
             if (maxShardsPerDayThreshold != -1) {
                 rsqc.setShardsPerDayThreshold(maxShardsPerDayThreshold);
             }
+            rsqc.setEnforceUniqueTermsWithinExpressions(enforceUniqueTermsWithinExpressions);
+
```
FineAndDandy: we only need to set this once. If the value is stored in the BaseQueryLogic, no need to set it again in the ShardQueryConfiguration

apmoriarty: I'll make that change, thanks

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -746,7 +740,8 @@ protected ASTJexlScript updateQueryTree(ScannerFactory scannerFactory, MetadataH
         
         stopwatch.stop();
         
-        if (enforceUniqueTermsWithinExpressions) {
+        // Enforce unique terms within an AND or OR expression.
+        if (config.getEnforceUniqueTermsWithinExpressions()) {
```
jwomeara: YES!

PR #776, opened 2020-03-04
Collapse uids by threshold, option added to ShardQueryConfig

apmoriarty:


warehouse/query-core/src/main/java/datawave/query/index/lookup/CreateUidsIterator.java
```diff
@@ -104,27 +143,40 @@ public boolean hasTop() {
...
-                        if (log.isTraceEnabled())
-                            log.trace("Adding uid " + StringUtils.split(uid, '\u0000')[1]);
-                        uids.add(uid);
-                    }
+                
+                // Three things must be true in order to return a document specific range
+                // 1. The uids stored in the index must have the ignore flag set to FALSE
+                // 2. The option to collapse all document ranges must be false.
+                // 3. The document count must be under the document threshold.
+                if (ignore || !collapseUids || (count <= collapseUidsThreshold)) {
```
ivakegg: this used to be if (!ignore).  Are we sure we got this right?

apmoriarty: Good catch, that should be !ignore.

apmoriarty:
Resolved merge conflict in the range stream that stemmed from removing the condense code path.

warehouse/query-core/src/main/java/datawave/query/index/lookup/CreateUidsIterator.java
```diff
@@ -80,17 +82,54 @@
     public void init(SortedKeyValueIterator<Key,Value> source, Map<String,String> options, IteratorEnvironment env) throws IOException {
         src = source;
         if (null != options) {
+            
             final String collapseOpt = options.get(COLLAPSE_UIDS);
-            if (null != collapseOpt) {
+            if (StringUtils.isNotBlank(collapseOpt)) {
```
ivakegg, line 87: seems like we are repeating this parsing of the option code for a boolean and then an integer.  Perhaps refactor into separate methods?

warehouse/query-core/src/main/java/datawave/query/index/lookup/CreateUidsIterator.java
```diff
@@ -104,27 +143,40 @@ public boolean hasTop() {
...
-                        if (log.isTraceEnabled())
-                            log.trace("Adding uid " + StringUtils.split(uid, '\u0000')[1]);
-                        uids.add(uid);
-                    }
+                
+                // Three things must be true in order to return a document specific range
+                // 1. The uids stored in the index must have the ignore flag set to FALSE
+                // 2. The option to collapse all document ranges must be false.
+                // 3. The document count must be under the document threshold.
+                if (!ignore || !collapseUids || (count <= collapseUidsThreshold)) {
```
ivakegg, line 167: The comment says that 3 thrings must be true.  However the boolean expression says that only one of those 3 things must be true.  Which is it?

warehouse/query-core/src/main/java/datawave/query/index/lookup/CreateUidsIterator.java
apmoriarty: This way of parsing the argument will tell you if a non-boolean value was passed in. I think that's sufficient reason to keep it as is.

warehouse/query-core/src/main/java/datawave/query/index/lookup/CreateUidsIterator.java
apmoriarty: That comment is wrong, I'll fix it.

warehouse/query-core/src/main/java/datawave/query/index/lookup/CreateUidsIterator.java
```diff
@@ -161,9 +161,10 @@ public void next() throws IOException {
                 ignore |= uidInfo.second();
                 
                 // Three things must be true in order to return a document specific range
-                // 1. The uids stored in the index must have the ignore flag set to FALSE
-                // 2. The option to collapse all document ranges must be false.
+                // 1. The uids stored in the index must NOT have the ignore flag set to FALSE
+                // 2. The option to collapse all document ranges cannot be set.
                 // 3. The document count must be under the document threshold.
+                // If any of those is false, do not add the uids.
```
ivakegg, line 167: The wealth of double and even triple negatives is awesome.  Now the comment on line 163 and line 167 are in direct conflict with each other.  It seems to me that neither the comment nor the code was correct.  My reasoning is as follows:
1) If the uids stored in the index have the ignore flag set to FALSE, then there are no uids to add, hence ignore MUST be false if we are to add the UIDs, regardless of collapseUids state.
2) I agree that either the collapseUids should be false OR the count is less than the threshold in order to add the uids.
Hence I believe the correct comment and comparison should be:
```
// Two things must be true in order to return a document specific range:
// 1) The ignore flag must be false
// 2) Either the collapse flag is false, or the document count is under the document threshold
if (!ignore && (!collapseUids || (count <= collapseUidsThreshold))) {
```

apmoriarty: Good call, I'll make that update.

ivakegg:
@FineAndDandy If you could give this one a look over I would appreciate it.

warehouse/query-core/src/main/java/datawave/query/index/lookup/CreateUidsIterator.java
```diff
@@ -80,17 +82,54 @@
...
+            
             final String collapseOpt = options.get(COLLAPSE_UIDS);
-            if (null != collapseOpt) {
+            if (StringUtils.isNotBlank(collapseOpt)) {
                 try {
-                    collapseUids = Boolean.valueOf(collapseOpt);
+                    this.collapseUids = Boolean.parseBoolean(collapseOpt);
                 } catch (Exception e) {
-                    collapseUids = false;
+                    String errMsg = "Option [" + COLLAPSE_UIDS_THRESHOLD + "] expected a boolean, received [" + collapseOpt + "].";
```
d-hwang, line 91: COLLAPSE_UIDS expected a boolean?

warehouse/query-core/src/main/java/datawave/query/index/lookup/CreateUidsIterator.java
```diff
@@ -80,17 +82,54 @@
...
+                    log.error(errMsg);
+                    throw new IllegalArgumentException(errMsg);
                 }
             }
+            
+            final String collapseThresholdOpt = options.get(COLLAPSE_UIDS_THRESHOLD);
+            if (StringUtils.isNotBlank(collapseThresholdOpt)) {
+                try {
+                    int threshold = Integer.parseInt(collapseThresholdOpt);
+                    if (threshold == -1 || threshold == Integer.MAX_VALUE) {
```
d-hwang, line 101: this.collapseUidsThreshold already defaults to MAX_VALUE, so I believe you just need to handle the else-case?

d-hwang (approved):
Just minor comments.

apmoriarty:
I don't think this is worthwhile at this point

PR #781, opened 2020-03-13
Added diagnostic information to the query iterator and ivarators.

jwomeara:


warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexListIteratorJexl.java
```diff
@@ -101,7 +100,11 @@ public DatawaveFieldIndexListIteratorJexl(DatawaveFieldIndexListIteratorJexl oth
     @Override
     public String toString() {
         StringBuilder builder = new StringBuilder();
-        builder.append("DatawaveFieldIndexFSTIteratorJexl{fName=").append(getFieldName()).append(", negated=").append(isNegated()).append("}");
+        if (fst != null)
+            builder.append("DatawaveFieldIndexFSTIteratorJexl");
```
drewfarris: Is there a reason you don't want to append the rest of the diagnostic output (e.g: queryId, field name) if `fst != null`?

jwomeara: The rest of the diagnostic output is always appended.  That part is outside of the if/else logic.

PR #782, opened 2020-03-16
FieldConfigurationHelper updates: interface and index-only fields

drewfarris:
. Extracted `FieldConfigHelper` interface
 . Implementation of `FieldConfigurationHelper` is `XMLFieldConfigHelper`
 . `FieldConfigHelper` now used to configure index-only fields.
 . index-only fields are `store=false` and `indexed=true`


drewfarris:
This could pave the way for reading this configuration from the `DatawaveMetadata` table, e.g.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/XMLFieldConfigHelper.java
```diff
@@ -0,0 +1,567 @@
...
+                this.fieldHelper.addTokenizedField(name);
+            }
+            
+            if (reverseTokenized) {
+                this.fieldHelper.addReverseTokenizedField(name);
+            }
+            
+            if (fieldType != null) {
+                if (this.ingestHelper != null) {
+                    this.ingestHelper.updateDatawaveTypes(name, fieldType);
```
keith-ratcliffe: This refactor looks good to me. WRT our recent conversation, I'm guessing that the given XML schema _could_ support multiple normalizers for a single field by just making the **indexType** attribute a comma-delimited list(?) The updateDatawaveTypes method will parse your fieldType var that way at least. I guess a new junit in XMLFieldConfigHelperTest could test that hypothesis, if multiple normalizers per-field is something we need to continue to support

drewfarris: Agreed, I think I'll tackle this in a separate ticket.

Issue #786, opened 2020-03-20
ValidPatternVisitor does not check ASTFunction nodes


apmoriarty:
The ValidPatternVisitor does not check ASTFunction nodes. This is an issue since the #INCLUDE and #EXCLUDE functions are expected to contain regexes.

PR #787, opened 2020-03-20
ValidPatternVisitor now checks #INCLUDE and #EXCLUDE functions.

apmoriarty:
Addresses https://github.com/NationalSecurityAgency/datawave/issues/786


billoley:
LGTM

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidPatternVisitor.java
```diff
@@ -69,22 +56,73 @@ public Object visit(ASTERNode node, Object data) {
...
     @Override
     public Object visit(ASTNRNode node, Object data) {
+        parseAndPutLiteral(node);
+        return data;
+    }
+    
+    @Override
+    public Object visit(ASTFunctionNode node, Object data) {
+        
+        if (node.jjtGetNumChildren() >= 4) {
```
ivakegg, line 66: I will have to say that a more robust solution here is to push this functionality into the function descriptor mechanism.  That way we can catch any function that might have a regex......

ivakegg: The JexlArgumentDescriptor already has a regexArguments() method to determine if the literal arguments are to be treated as regexes.  We should use that along with fieldsForNormalization to determine which arguments to test.

ivakegg: And then we can apply this to all functions that return true for regexArguments().....

apmoriarty: I took a look at some of the existing descriptors and see where you're going with this. Should have an update for this pr tomorow sometime

ivakegg: An example of how to do this can be found in the datawave.query.jexl.visitors.FunctionNormalizationRebuildingVisitor.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/QueryFunctions.java
```diff
@@ -22,6 +22,8 @@
     public static final String OPTIONS_FUNCTION = "options";
     public static final String UNIQUE_FUNCTION = "unique";
     public static final String GROUPBY_FUNCTION = "groupby";
+    public static final String INCLUDE_FUNCTION = "include";
+    public static final String EXCLUDE_FUNCTION = "exclude";
```
apmoriarty, line 27: Pretty sure this is the right place to put these constants, let me know if that needs to change.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidPatternVisitor.java
```diff
@@ -69,22 +59,65 @@ public Object visit(ASTERNode node, Object data) {
...
+     */
+    public void parseAndPutPattern(JexlNode node) {
+        // Catch the situation where a user might enter FIELD1 !~ VALUE1
+        Object literalValue = JexlASTHelper.getLiteralValue(node);
         if (literalValue != null && String.class.equals(literalValue.getClass())) {
             String literalString = (String) literalValue;
-            if (patternCache.containsKey(literalString)) {
-                return data;
+            // Only parse literals that contain a regex char
+            if (StringUtils.containsAny(literalString, "?.*^+-_[](){}")) {
```
apmoriarty, line 109: Tried to scope down the number of patterns that get parsed. Might not be necessary.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/QueryFunctions.java
ivakegg: These are EvaluationPhaseFilterFunctions, not QueryFunctions

ivakegg: and the functions are called includeRegex and excludeRegex

warehouse/query-core/src/main/java/datawave/query/jexl/functions/QueryFunctionsDescriptor.java
```diff
@@ -138,6 +138,10 @@ private static void verify(String name, int numArgs) {
             if (numArgs == 0) {
                 throw new IllegalArgumentException("Expected at least one argument to the " + name + " function");
             }
+        } else if (name.equals(QueryFunctions.INCLUDE_FUNCTION) || name.equals(QueryFunctions.EXCLUDE_FUNCTION)) {
```
ivakegg, line 141: Wrong functions descriptor.

warehouse/query-core/src/main/java/datawave/query/language/functions/jexl/Exclude.java
```diff
@@ -1,11 +1,12 @@
 package datawave.query.language.functions.jexl;
 
+import datawave.query.jexl.functions.QueryFunctions;
 import datawave.query.language.functions.QueryFunction;
 import datawave.query.search.WildcardFieldedFilter;
 
 public class Exclude extends AbstractEvaluationPhaseFunction {
     public Exclude() {
-        super("exclude");
+        super(QueryFunctions.EXCLUDE_FUNCTION);
```
ivakegg, line 9: I would leave this the way it was.

ivakegg: and ditto for the other language function classes

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidPatternVisitor.java
ivakegg: I don't think this is necessary, and in fact could be problematic.  A Regex is not guaranteed to contain one of those characters.  Perhaps a better way to go is if the literal contains only alphanumeric characters.

warehouse/query-core/src/main/java/datawave/query/language/functions/jexl/Exclude.java
apmoriarty: I'll revert that to the way it was

warehouse/query-core/src/main/java/datawave/query/jexl/functions/QueryFunctionsDescriptor.java
apmoriarty: Got it.

PR #789, opened 2020-03-23
fixes #750 Re-enable AuditBean in Embedded, Filter out shaded guava c…

### LINKED ISSUE #750: BulkResults (via MapReduce API) broken after Datawave refactoring
Additional jars need to be added to the Map/Reduce job context, some classes need to be excluded, some classes need to be created to support running in a map job.

billoley:
fixes #750 Re-enable AuditBean in Embedded, Filter out shaded guava classes from jboss-client.jar while writing to job distributed cache

### DIFF
--- a/web-services/map-reduce/src/main/java/datawave/webservice/mr/configuration/MapReduceJobConfiguration.java
+++ b/web-services/map-reduce/src/main/java/datawave/webservice/mr/configuration/MapReduceJobConfiguration.java
@@ -25,6 +25,8 @@ import java.net.URL;
 import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -35,6 +37,8 @@ import java.util.jar.JarInputStream;
 import java.util.jar.JarOutputStream;
 import java.util.regex.Pattern;
 import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+import java.util.zip.ZipOutputStream;
 
 public class MapReduceJobConfiguration {
     
@@ -249,7 +253,7 @@ public class MapReduceJobConfiguration {
             } else {
                 // Jars within the deployed EAR in Wildfly use the VFS protocol, and need to be handled specially.
                 if (jarFile.startsWith("vfs:")) {
-                    Set<String> files = new HashSet<>();
+                    Set<String> files = new LinkedHashSet<>();
                     try (InputStream urlInputStream = new URL(jarFile).openStream()) {
                         JarInputStream jarInputStream = (JarInputStream) urlInputStream;
                         for (JarEntry jarEntry = jarInputStream.getNextJarEntry(); jarEntry != null; jarEntry = jarInputStream.getNextJarEntry()) {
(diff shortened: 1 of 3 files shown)


web-services/map-reduce/src/main/java/datawave/webservice/mr/configuration/MapReduceJobConfiguration.java
```diff
@@ -302,6 +316,40 @@ protected void prepareClasspath(String jobId, Job job, Path jobDir) throws Excep
...
+                    for (ZipEntry zipEntry = zipInputStream.getNextEntry(); zipEntry != null; zipEntry = zipInputStream.getNextEntry()) {
+                        boolean include = true;
+                        Iterator<Pattern> patternItr = patterns.iterator();
+                        while (patternItr.hasNext() && include == true) {
+                            Pattern p = patternItr.next();
+                            if (p.matcher(zipEntry.getName()).matches()) {
+                                include = false;
+                            }
+                        }
+                        if (include) {
```
brianloss: Not required, but you can simplify this quite a bit with streams...
```suggestion
                        final String entryName = zipEntry.getName();
                        if (patterns.stream().noneMatch(p -> p.matcher(entryName).matches())) {
```

web-services/map-reduce/src/main/java/datawave/webservice/mr/configuration/MapReduceJobConfiguration.java
```diff
@@ -302,6 +316,40 @@ protected void prepareClasspath(String jobId, Job job, Path jobDir) throws Excep
...
+                            zipOutputStream.putNextEntry(zipEntry);
+                            byte[] buffer = new byte[2048];
+                            int len;
+                            while ((len = zipInputStream.read(buffer)) > 0) {
+                                zipOutputStream.write(buffer, 0, len);
+                            }
```
brianloss: Another way...
```suggestion
                            ByteStreams.copy(zipInputStream, zipOutputStream);
```

billoley: Thanks for the suggestions.  This way is much cleaner.  It took me a while to figure out the Zip(Input/Output)Stream semantics and I didn't circle back to clean it up.

Oh the irony that we are using a guava class in the wildfly jvm to remove the wildfly classes from a shaded jar.  I combined both suggestions and pushed a commit.

brianloss: Figured you'd appreciate that. ;)

web-services/map-reduce/src/main/java/datawave/webservice/mr/configuration/MapReduceJobConfiguration.java
```diff
@@ -284,7 +288,17 @@ protected void prepareClasspath(String jobId, Job job, Path jobDir) throws Excep
         File[] jarFiles = libDir.listFiles(jarFilter);
         if (jarFiles != null) {
             for (File jar : jarFiles) {
-                addSingleFile(jar, new Path(classpath, jar.getName()), jobId, job, fs);
+                // remove guava classes from jboss-client.jar
+                if (jar.getName().equals("jboss-client.jar")) {
+                    List<Pattern> patterns = new ArrayList<>();
```
ivakegg: ugh, but ok for now I guess....

billoley: I know, but the alternative is storing a modified version of that file in our version control.  That's not great either.  The dependency is needed for our MapReduce jobs to run, but because of the inability to order the user classpath to the job we have to eliminate the newer guava classes that are shaded into that jar.  

I considered adding configuration to take care of this (which jar to transfer, which regexes to filter out).  I can do that later.  The java would be cleaner at least.   I ran out of time at the end of the week and this will work for now.

PR #791, opened 2020-03-24
WIP Completes #790: Fixes for query tracing

### LINKED ISSUE #790: Query tracing seems to be broken
Trace info is not getting written to Accumulo as expected when using `trace=true` as a query param. Appears that our trace client config in `AccumuloConnectionFactoryBean` is no longer initializing things properly.

keith-ratcliffe:
Query traces are now persisted to Accumulo `trace`  table, as intended.

However, traces still don't appear in Monitor's "Recent traces" view, so perhaps still more to do here. FWIW, Accumulo's own traces don't seem to be getting displayed as intended, so maybe an issue with the monitor itself


web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -2051,11 +2045,10 @@ private BaseQueryResponse next(final String id, boolean checkForContentLookup) {
...
                 log.error("Error committing transaction: Unknown error", e);
-            } finally {
-                // Stop timing on this trace, if any
-                if (span != null) {
-                    span.stop();
-                }
+            }
+            // Stop timing on this trace, if any
+            if (span != null) {
+                span.stop();
```
brianloss, line 2051: I'm curious as to why this was moved outside of the finally block? Don't we want to stop the "next" call span even if the call itself fails with an exception?

keith-ratcliffe: In this case the enclosing scope is already a _finally_ block, starting around L2029, so the extra _finally_ didn't seem to buy us much here, since any exceptions will still be suppressed above. I suppose the span still might not be stopped here, if some java.lang.Error-derived Throwable is raised in the preceding _try_. Moving this block above the _try_ would mitigate that though. I'll do that

ivakegg:
Let's rebase against 2.9 please. Thanks.

ivakegg:
Please resolve conflicts.  Also there seems to be more commits than I expected for this merge request.

keith-ratcliffe:
> Let's rebase against 2.9 please. Thanks.



> Please resolve conflicts. Also there seems to be more commits than I expected for this merge request.

We discovered that there were additional aspects of tracing that were broken in DW. E.g., thread contexts external to QueryExecutorBean were not delivering spans to Accumulo (i.e., QueryIterator) as expected, so this is still a WIP.

The tracing breakage there is apparently due to our nested ExecutorServices. In short, our scan threads need the same "DistributedTrace.enable(...)" logic that we previously only had inside AccumuloConnectionFactoryBean

I have that fixed, and will update this PR once I've done a little more testing at scale

keith-ratcliffe:
OBE. Tracing in Accumulo via htrace is no longer a thing

PR #792, opened 2020-03-27
Whitespace-only formatting cleanup for QueryLogicFactory.xml

drewfarris:


drewfarris:
Cleaned up some tabs and trailing whitespace.

PR #793, opened 2020-03-27
FacetHandler ingest code + FacetedQueryLogic Configuration

drewfarris:


drewfarris:
Ok, this is ready for review - if folks are interested in doing so.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetValue.java
```diff
@@ -0,0 +1,86 @@
+package datawave.ingest.mapreduce.handler.facet;
+
+import com.clearspring.analytics.stream.cardinality.HyperLogLogPlus;
+import com.clearspring.analytics.stream.frequency.CountMinSketch;
+import jj2000.j2k.util.FacilityManager;
```
drewfarris, line 5: Bad import? Probably.

billoley:
If I'm reading correctly.... In FacetedQueryPlanner, FacetCheck, and FacetQueryPlanVisitor, it seems that the facet tables are hardcoded to FacetsNating and FacetsNatingMetadata (needs to be changed), whereas your PR is naming those datawave.facets and datawave.facetMetadata.

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/handler/facet/FacetHandlerTest.java
```diff
@@ -0,0 +1,595 @@
...
+     */
+    private void processEvent(RawRecordContainer event, Multimap<String,NormalizedContentInterface> eventFields,
+                    ExtendedDataTypeHandler<Text,BulkIngestKey,Value> handler) {
+        
+        assertNotNull("Event was null", event);
+        
+        try {
+            handler.process(null, event, eventFields, sctx, contextWriter);
+        } catch (IOException | InterruptedException e) {
+            throw new RuntimeException("Error setting up context writer", e);
```
billoley, line 206: This error message seems like it was copied from elsewhere in the test & should probably say that there was an error in processing

drewfarris:
> If I'm reading correctly.... In FacetedQueryPlanner, FacetCheck, and FacetQueryPlanVisitor, it seems that the facet tables are hardcoded to FacetsNating and FacetsNatingMetadata (needs to be changed), whereas your PR is naming those datawave.facets and datawave.facetMetadata.

Yup, this needs to be fixed.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetHandler.java
```diff
@@ -0,0 +1,366 @@
...
+    public static final String FACET_CATEGORY_DELIMITER = ".facet.category.delimiter";
+    public static final String FACET_FIELD_PREDICATE_CLASS = ".facet.field.predicate.class";
+    
+    public static final String FACET_CATEGORY_PREFIX_REGEX = "\\.facet\\.category\\.name\\..*";
+    
+    private static final Text PV = new Text("pv");
+    private static final String NULL = "\0";
+    private static final Value EMPTY_VALUE = new Value(new byte[] {});
+    
+    // protected Multimap<String,NormalizedContentInterface> fields = HashMultimap.create();
```
apmoriarty, line 75: Remove commented code

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetHandler.java
```diff
@@ -0,0 +1,366 @@
...
+     * 
+     * @param pivotFieldName
+     * @param facetFieldName
+     * @return
+     */
+    protected Text createColumnFamily(String pivotFieldName, String facetFieldName) {
+        return new Text(pivotFieldName + NULL + facetFieldName);
+    }
+    
+    /** TODO */
```
apmoriarty, line 324: Extra TODO comment?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetHandler.java
```diff
@@ -0,0 +1,366 @@
...
+    @Override
+    public FacetValue estimate(RawRecordContainer input) {
+        // precision value: 10, sparse set disabled.
```
cawaring, line 327: Are you avoiding using the sparse setting for HLL++ because of some of the issues with size in certain cases when serializing it to disk?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetHandler.java
```diff
@@ -0,0 +1,366 @@
...
+        HyperLogLogPlus card = new HyperLogLogPlus(10);
+        card.offer(input.getId().toString());
+        
+        return new FacetValue(card, new CountMinSketch(10, 1, 1));
```
cawaring, line 331: I might not be reading this correctly, but using this constructor we're creating a CMS with depth of 10 and width of 1.  We should probable make these parameters configurable, likewise for hLL++.  That way one could set the parameters based on the error and probability they want to support.  Could add some bounds checking to make it reasonable.

warehouse/ingest-core/src/test/resources/config/facet-config.xml
```diff
@@ -0,0 +1,28 @@
...
+        <value>datawave.facetMetadata</value>
+    </property>
+    <property>
+        <name>facet.metadata.table.loader.priority</name>
+        <value>40</value>
+    </property>
+    <property>
+        <name>facet.hash.table.name</name>
+        <value>datawave.facetHashes</value>
+    </property>
```
cawaring, line 23: We may want to include some configurations for hll++ and cms here.

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/handler/facet/FacetHandlerTest.java
drewfarris: Resolved

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetHandler.java
drewfarris: I need to take a close look at this - the sparse setting seems a little wasteful, I'm not certain when the sparse setting becomes more efficient than the alternative (?dense?) setting.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetHandler.java
drewfarris: Good point. I'll work on the plumbing for this. I'm wondering if we might want to do this on a per-field basis or along some other dimension than "It works this way for everything". I'm also wondering if these parameters get encoded in the serialized object so that we can track this for individual estimates.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetHandler.java
```diff
@@ -0,0 +1,363 @@
...
+    /* Global configuration fields */
+    
+    protected Text facetTableName;
+    protected Text facetMetadataTableName;
+    protected Text facetHashTableName;
+    
+    /* Per-datatype configuration fields */
+    
+    protected int facetHashThreshold;
+    protected String categoryDelimiter = "/";
```
ivakegg, line 84: Just a nit, but '/' suggests a path of some sort which is not the case here.  This is more like a definition so perhaps colon, equals or some sort of bracket may be better.  Again, just a nit.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetHandler.java
```diff
@@ -0,0 +1,363 @@
...
+        return markingFunctions == null ? vis.flatten() : markingFunctions.flatten(vis);
+    }
+    
+    @Override
+    public long process(KEYIN key, RawRecordContainer event, Multimap<String,NormalizedContentInterface> fields,
+                    TaskInputOutputContext<KEYIN,? extends RawRecordContainer,KEYOUT,VALUEOUT> context, ContextWriter<KEYOUT,VALUEOUT> contextWriter)
+                    throws IOException, InterruptedException {
+        
+        final String shardId = shardIdFactory.getShardId(event).substring(0, 8);
+        Text dateColumnQualifier = new Text(shardId); // TODO: Makes an assumption about the structure of the shardId.
```
ivakegg, line 219: use ShardIdFactory.getDateString(shardId)

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetHandler.java
```diff
@@ -0,0 +1,363 @@
...
+            Text reflexiveCf = createColumnFamily(pivotFieldName, pivotFieldName);
+            for (NormalizedContentInterface pivotTypes : eventFields.get(pivotFieldName)) {
+                for (String facetFieldName : pivotMap.get(pivotFieldName)) {
+                    if (pivotFieldName.equals(facetFieldName))
+                        continue;
+                    
+                    Text generatedCf = createColumnFamily(pivotFieldName, facetFieldName);
+                    Text myCf = generatedCf;
+                    
+                    if (HashTableFunction.isReduced(pivotTypes))
```
ivakegg, line 254: Shouldn't this check be moved outside of this for loop but inside the containing loop (i.e move to line 247)?  Alternatively the eventFields could be prefiltered to not include any "reduced" members.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetValue.java
```diff
@@ -0,0 +1,85 @@
...
+        return super.get();
+    }
+    
+    @Override
+    public void readFields(DataInput in) throws IOException {
+        int cardSize = in.readInt();
+        byte[] cardinality = new byte[cardSize];
+        in.readFully(cardinality);
+        
+        // TODO: frequencyEstimate
```
ivakegg, line 53: yup, missing that

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetHandler.java
drewfarris: @ivakegg Thanks, I understand where you're coming from. It's a nit, but it's worth doing the work up-front to get it right. 

In use, we currently have configuration property values like: 
`NETWORK_NAME/GENRES,EMBEDDED_CAST_PERSON_GENDER,RATING_AVERAGE` .

Thinking about other delimiters that make sense in this context, `;` may be the best bet from a semantic perspective:
`NETWORK_NAME;GENRES,EMBEDDED_CAST_PERSON_GENDER,RATING_AVERAGE`

(I considered '|', '#', and ';', although '|' might be better visually it is also semantically overloaded)

What do you think? The changes I need to make to address this are distributed across code, tests and configuration, so I want to make sure I get something we're satisfied with the first time.

_I won't get started about embedding structured data in a property value in that it forces you do to special parsing on the value itself - considering alternatives I think it's the lesser of evils in this particular case._

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetHandler.java
drewfarris: Yes `HashTableFunction.isReduced(pivotTypes)` can be moved out. 

For what it's worth, we do use the reduced members in `eventFields` for the inner loop that iterates over `facetFieldName` members when generating facets, so I can't remove all reduced members from `eventFields` (see snippet below). I suppose I could create separate collections of pivot and facet fields, but is that any better?

```
for (NormalizedContentInterface facetTypes : eventFields.get(facetFieldName)) {
                        if (HashTableFunction.isReduced(facetTypes)) {
                            myCf.append(HashTableFunction.FIELD_APPEND_BYTES, 0, HashTableFunction.FIELD_APPEND_BYTES.length);
                        }
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetValue.java
drewfarris: Strangely, `FacetValue` is never used in this code.  [`FacetHandler.estimate(..)`](https://github.com/drewfarris/datawave/blob/facethandler/warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetHandler.java#L323-L329)  is the only place it's created. Instead of the `FacetValue` a [Simpler Value](https://github.com/drewfarris/datawave/blob/facethandler/warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetHandler.java#L241-L241) is used for the facets. 

Bottom line: I need to do more digging to uncover the original intent here, but I'm hoping that might not prevent merging the current implementation

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetHandler.java
ivakegg: Your point about embedding structured data in a value is a very good one.  I have been a huge contributor to this evil practice unfortunately but we all have to learn somehow.  If we are not going to use a separate property, then I vote for a semicolon I guess.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetHandler.java
ivakegg: your choice.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetValue.java
ivakegg: ok

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetHandler.java
ivakegg: final String shardId = ShardIdFactory.getDateString(shardId);

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetHandler.java
drewfarris: Done. This is now a semicolon.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetHandler.java
drewfarris: Done. Thanks for the pointer to how to do this correctly.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/facet/FacetHandler.java
drewfarris: Done. I just moved `HashTableFunction.isReduced(pivotTypes)` outside of the loop it was in now.

PR #795, opened 2020-04-01
Fix potential NPE in TermFrequencyIndexIterator.

apmoriarty:


ivakegg:
I certainly see the potential for an NPE, however I am fearing that we are using this for fields that are index only but are not tokens.  What kind of field got us into this situation?

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyIndexIterator.java
```diff
@@ -187,12 +187,20 @@ public void next() throws IOException {
             
             for (int i = 0; i < 256 && source.hasTop() && key.getFieldName().compareTo(field) < 0; ++i) {
                 source.next();
-                key = new DatawaveKey(source.getTopKey());
+                
+                tk = source.getTopKey();
```
ivakegg, line 191: Reusing tk here is not good here.  Say we get through this loop without tk being null, and then the check at line 204 is true.  This would result in a seek and we would immediately pop out of the outer while loop because tk is not null and would not be holding the correct field name!  Use a separate Key variable instead.

apmoriarty: Will do

ivakegg:
If this tests ok on your system, then lets pull her in.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyIndexIterator.java
```diff
@@ -184,15 +184,24 @@ public void next() throws IOException {
...
             DatawaveKey key = new DatawaveKey(top);
+            Key nextTop = null;
             
             for (int i = 0; i < 256 && source.hasTop() && key.getFieldName().compareTo(field) < 0; ++i) {
                 source.next();
-                key = new DatawaveKey(source.getTopKey());
+                
+                nextTop = source.getTopKey();
+                if (nextTop == null)
+                    break;
```
brianloss, line 194: I wouldn't hold it up over this, and some could argue it's too esoteric, but it seems like this could actually be a good use of a [labelled continue](https://www.developer.com/java/data/understanding-the-java-labeled-statement.html) statement. Label the outer while loop, change this to `continue <label>` and remove the addition at line 202.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyIndexIterator.java
```diff
@@ -184,15 +184,24 @@ public void next() throws IOException {
...
+                
+                key = new DatawaveKey(nextTop);
                 if (log.isTraceEnabled()) {
                     log.trace("Have key " + key + " < " + field);
                 }
             }
             
+            if (nextTop == null)
+                continue;
```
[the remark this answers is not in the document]
apmoriarty: I'll have that fix up in a moment

PR #802, opened 2020-04-08
fixes #801: Added a fix for reusing completed ivarator directory sort…

### LINKED ISSUE #801: ivarators failing to reload completed directory
ivarators were built to reload a completed directory when being rebuilt from an iterator tear-down/rebuild.  However that mechanism was broken back when the unsortedUIDs mechanism was added.  This needs to be fixed.

ivakegg:
…ed sets along with a test.


jwomeara:
Wow!  What a test!  Great work!

PR #804, opened 2020-04-10
Fixes #785: fixes for DedupingIterator and UniqueTransform,

### LINKED ISSUE #785: Unreachable code in both UniqueTransform and DedupingIterator caused by "DEBUG" field, and other issues
Both UniqueTransform and DedupingIterator classes have unreachable code caused by their respective  "DEBUG" var usage...

**UniqueTransform**: 
```
private final boolean DEBUG = false;
```
**DedupingIterator**:
```
private final boolean DEBUG = true;
```
**Both** have the following identical block of code:
```
    if (DEBUG && !seen.contains(byteSeq)) {
          throw new IllegalStateException("This event is 1 in 1Q!");
    } else {
          return true;
    }
```
Neither class provides a way for its DEBUG value to be changed, so UniqueTransform can never throw the exception, and DedupingIterator can never return true.

In testing, we've seen DedupingIterator throw the exception and cause the given query to fail, which   could be the result of having exceeded 500K results, based on DedupingIterator's BloomFilter definition (expectedInsertions: 500000).

In both classes, it seems like an appropriate fix here would be to provide some mechanism to override the DEBUG value, and possibly also to override the BloomFilter config

Anyone have any thoughts on this one?

keith-ratcliffe:
new config options, and junits added

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/tables/DedupingIterator.java
+++ b/warehouse/query-core/src/main/java/datawave/query/tables/DedupingIterator.java
@@ -3,32 +3,30 @@ package datawave.query.tables;
 import com.google.common.hash.BloomFilter;
 import com.google.common.hash.Funnel;
 import com.google.common.hash.PrimitiveSink;
-import java.io.Serializable;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.Map.Entry;
-import org.apache.accumulo.core.data.ArrayByteSequence;
 import org.apache.accumulo.core.data.ByteSequence;
 import org.apache.accumulo.core.data.Key;
 import org.apache.accumulo.core.data.Value;
 
-/**
- * Created on 9/13/16.
- */
+import java.io.Serializable;
+import java.util.Iterator;
+import java.util.Map.Entry;
+
 class DedupingIterator implements Iterator<Entry<Key,Value>> {
+    static final int BLOOM_EXPECTED_DEFAULT = 500000;
+    static final double BLOOM_FPP_DEFAULT = 1e-15;
+    
     private Iterator<Entry<Key,Value>> delegate;
     private Entry<Key,Value> next;
     private BloomFilter<byte[]> bloom = null;
-    private HashSet<ByteSequence> seen;
-    private final boolean DEBUG = true;
+    
+    public DedupingIterator(Iterator<Entry<Key,Value>> iterator, int bloomFilterExpected, double bloomFilterFpp) {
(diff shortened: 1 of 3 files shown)


web-services/query/src/main/java/datawave/webservice/query/configuration/GenericQueryConfiguration.java
```diff
@@ -47,6 +47,8 @@
     
     protected boolean bypassAccumulo;
     
+    protected boolean iteratorDebug = true;
```
apmoriarty, line 50: I think this should default to false

web-services/query/src/test/java/datawave/webservice/query/configuration/TestBaseQueryLogic.java
```diff
@@ -54,6 +54,7 @@ public void testConstructor_Copy() throws Exception {
         expect(this.copy.getRoleManager()).andReturn(roleManager);
         expect(this.copy.getSelectorExtractor()).andReturn(null);
         expect(this.copy.getBypassAccumulo()).andReturn(false);
+        expect(this.copy.getIteratorDebug()).andReturn(true);
```
apmoriarty, line 57: I think this should also expect a false-by-default value.

web-services/query/src/main/java/datawave/webservice/query/configuration/GenericQueryConfiguration.java
keith-ratcliffe: I'm cool with defaulting to false, if there are no objections. But I'm curious to get your thoughts on DedupingIterator's `DEBUG` behavior in general.

The reason for defaulting to `true` was so that the existing behavior of DedupingIterator would be retained (assuming that's what we wanted), and to avoid the need for QLF.xml changes for defaulting to `true` there

And this new config setting was just to have a way of disabling it...mainly to prevent the `This event is 1 in 1Q!` error, which is something you're likely to run into on DW clusters that allow more than 500K results to be returned for a given query

So, if we default to false, I guess the question is, do we want to enable by default in QLF.xml or leave it disabled? If we leave it disabled, then perhaps the better question is, should we consider just getting rid of the `DEBUG` path inside DedupingIterator altogether and also this new config setting? That's the strategy I took with UniqueTransform in this PR, for the simple fact that the `true` path was never used and there was no way to enable it.

If this is something that will rarely be enabled in the future, then I think I'd much prefer to remove the debug behavior, and avoid adding yet another config setting.

Thoughts?

Issue #805, opened 2020-04-17
quickstart build fails


szemicael:
In release/version 2.10, when running env.sh:
[ERROR] Failed to execute goal on project datawave-ws-security: Could not resolve dependencies for project gov.nsa.datawave.webservices:datawave-ws-security:ejb:2.10.0-SNAPSHOT: Could not transfer artifact javax.transaction:jta:jar:1.0.1B from/to boundless (https://repo.boundlessgeo.com/main/): repo.boundlessgeo.com: Unknown host repo.boundlessgeo.com -> [Help 1]
Appears repo at https://repo.boundlessgeo.com/main/ no longer exists.
Replacing repo with https://repo.osgeo.org/repository/release/ in main pom.xml fixed the issue.

keith-ratcliffe:
Reopening temporarily until dd04d37 is backported to other branches. Noticed PR builds for 2.7 and 2.8 were failing due to the same problem

PR #809, opened 2020-04-20
fixes #797: unsorted and yielding do not play well together

### LINKED ISSUE #797: The QueryIterator unsorted mechanism and yielding do not play well together
We have witnessed that if the yielding mechanism kicks in whilst using the unsorted mechanism in the QueryIterator (QueryOptions.SORTED_UIDS = false) on a shard range can result in yielding to a position outside of the initial range.  This appears to only happen after being torn down once.  In addition the seek range as well as the yield key appear to be lacking the "count" which is added to the beginning of the column family (see the datawave.query.iterator.ResultCountIterator).  So somehow the logic is bypassing this iterator.  The only way I can see that happening is that the seek range is being detected as being a document range which should be impossible in this scenario because the end key contains only a row.

ivakegg:
* Created a test case that recreates the issue.
 * Fixed the edge case.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/iterator/profile/FinalDocumentTrackingIterator.java
+++ b/warehouse/query-core/src/main/java/datawave/query/iterator/profile/FinalDocumentTrackingIterator.java
@@ -111,11 +111,7 @@ public class FinalDocumentTrackingIterator implements Iterator<Map.Entry<Key,Val
     
     @Override
     public boolean hasNext() {
-        // if we yielded, then leave gracefully
-        if (yield != null && yield.hasYielded()) {
-            return false;
-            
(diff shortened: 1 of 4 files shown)


drewfarris:
Looks like this is having problems due to #805 Does this still need to be targeted at 2.7 or can we rebase to master?

ivakegg:
@drewfarris This is a critical bug fix so I was pulling it in as early as possible.  Why do you say this is having problems due to #805?

drewfarris:
@ivakegg when I looked at this last week, the checks weren't passing because of the repository settings that were in place at the time. @keith-ratcliffe has since back-ported #805 to the 2.7 branch so it should be g2g. I was looking at this last week locally and had inadvertently rebased against master. Bottom line - this LGTM.

PR #811, opened 2020-04-28
Non-functional tweaks to generics usage

drewfarris:
...to avoid build failures in VSCode.

Some commentary: These are all changes to the way generics were managed and thus have implications to compile time as opposed to runtime. IntelliJ and even `javac` as we execute from maven don't care about these, but the syntax/language engine in VSCode does.

In fact, upon analysis, this seems that each of these implementations differ in how they handle a non-parameterized type, e.g: `datawave.data.type.Type` vs. a wildcard parameterized type `datawave.data.type.Type<?>`. 

The case in the mock for `ExpandCompositeTermsTest` is especially interesting because of the use of `Object` in the `compareTo` method. This doesn't adhere to either of the parameter constraints established in `BaseType`, that the templated object be `Comparable` or `Serializable`. So that seems more clearly incorrect (at compile time)...

### DIFF
--- a/warehouse/ingest-core/src/main/java/datawave/IdentityDataType.java
+++ b/warehouse/ingest-core/src/main/java/datawave/IdentityDataType.java
@@ -1,15 +1,16 @@
 package datawave;
 
 import java.util.Collection;
+import datawave.data.type.Type;
 
-public class IdentityDataType implements datawave.data.type.Type {
+public class IdentityDataType implements Type<String> {
     @Override
     public String normalize() {
         throw new UnsupportedOperationException();
     }
     
     @Override
-    public Comparable denormalize() {
+    public String denormalize() {
         throw new UnsupportedOperationException();
     }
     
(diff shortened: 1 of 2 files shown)


warehouse/query-core/src/test/java/datawave/query/jexl/visitors/ExpandCompositeTermsTest.java
```diff
@@ -1289,7 +1287,7 @@ public boolean producesFixedLengthRanges() {
         }
         
         @Override
-        public int compareTo(Object o) {
```
drewfarris, line 1292: Original Error was:
```
The method compareTo(Object) of type ExpandCompositeTermsTest.MockDiscreteIndexType must override or implement a supertype method.  Java(67109498 [1292, 20]
```

warehouse/ingest-core/src/main/java/datawave/IdentityDataType.java
```diff
@@ -1,15 +1,16 @@
 package datawave;
 
 import java.util.Collection;
+import datawave.data.type.Type;
 
-public class IdentityDataType implements datawave.data.type.Type {
```
drewfarris, line 5: Original error was in `EventMetadataTest` 
```
Type mismatch: cannot convert from List<IdentityDataType> to List<Type<?>>`
```

Previouly, `IdentityDataType` wasn't parameterized and resulted in a warning.

PR #832, opened 2020-06-17
fixes #824 SessionScanner should handle BatchResource correctly

### LINKED ISSUE #824: Expansion inconsistency
There seems to be a case where an unfielded regex will sometimes incorrectly expand out to a subset of the terms instead of hitting the exceeded value threshold.  It seems to occur spuriously.  One interesting note is that in the log messages from the DefaultQueryPlanner, the Expand Regex timing is significantly shorter when the query hits this condition as it only expands one of the fields to a portion of the total values.  When the query correctly marks the fields as an ExceededValueThreshold node the Expand Regex portion takes significantly longer.

FineAndDandy:
The SessionScanner was making assumptions about lastKey and tearing down the scanner which cannot be done with a BatchResource set.


warehouse/query-core/src/test/java/datawave/query/tables/ScannerSessionTest.java
```diff
@@ -0,0 +1,231 @@
...
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.SortedSet;
+import java.util.TreeSet;
+
+/**
+ * This test spins up a mini accumulo to accurately test the effect of underlying Scanner/Batch scanners against the ScannerSession. InMemoryAccumulo makes some
+ * simplications that in the past have masked bugs
```
apmoriarty, line 40: simplifications?

FineAndDandy: fixed

ivakegg:
You had mentioned something about a BatchResource being used incorrectly with a regular scanner session.  Is that going to be another pull request or did I misunderstand?

PR #833, opened 2020-06-18
fixes #826 Stop looping over terms once threshold is reached

### LINKED ISSUE #826: LookupTermsFromRegex incorrectly checks exceeded value threshold
The iterator will continue to run until it times out or exhausts all keys

FineAndDandy:


warehouse/query-core/src/main/java/datawave/query/jexl/lookups/LookupTermsFromRegex.java
```diff
@@ -256,15 +258,25 @@ public IndexLookupMap lookup(ShardQueryConfiguration config, ScannerFactory scan
...
+                            // if the limits haven't already been surpassed
+                            if (!fieldsToValues.isKeyThresholdExceeded()
+                                            && (fieldsToValues.get(field) == null || !fieldsToValues.get(field).isThresholdExceeded())) {
+                                // We are only returning a mapping of field value to field name, no need to
+                                // determine cardinality and such at this point.
+                                boolean added = fieldsToValues.put(field, term);
+                                
+                                // if there was a failure to add, and the key threshold isn't exceeded, increment the value exceeded count
+                                if (!added && !fieldsToValues.isKeyThresholdExceeded()) {
+                                    valueExceededCount++;
```
ivakegg, line 270: I don't think this is correct.  Using the return value from fieldsToValues.put being false does not mean that fieldsToValues.get(field).isThresholdExceeded().  The return value could be false because that value was already in the set.

FineAndDandy: you are right, if the key is already present in the underlying value hashset it will return false.

PR #837, opened 2020-06-19
fixes #836: Simplify regex expressions

### LINKED ISSUE #836: Simplify regexes
It has come to my attention that executing a regex of the following form: ".*.*.*.*.*.*.*bla" or ".*?.*?.*?.*?.*?.*?.*?bla" can have serious performance issues.  I found that there is an exponential growth relative to the number of "/*" or "/*?" in the regex and to the size of the value in terms of performance.  We have discovered this regex  running against 1000+ character values to actually take days!  This has caused significant issue in terms of hanging up tservers.  This is obviously a user error, however the system should be able to account for this.  We need a mechanism to simplify this regex to avoid the performance implications.

ivakegg:
* Created a NodeTransform to simplify regex expressions for performance reasons

### DIFF
--- a/warehouse/query-core/src/test/java/datawave/query/planner/rules/NodeTransformVisitorTest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/planner/rules/NodeTransformVisitorTest.java
@@ -26,6 +26,7 @@ public class NodeTransformVisitorTest {
     
     private static final List<String> PATTERNS = Arrays.asList(new String[] {".\\.\\*", "\\.\\*.", "\\.\\*<[^<>]+>"});
     private static final RegexPushdownTransformRule regexPushdownRule = new RegexPushdownTransformRule();
+    private static final RegexSimplifierTransformRule regexSimplifier = new RegexSimplifierTransformRule();
     private static final NodeTransformRule reverseAndRule = new NodeTransformRule() {
         @Override
         public JexlNode apply(JexlNode node, ShardQueryConfiguration config, MetadataHelper helper) {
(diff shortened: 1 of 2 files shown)


warehouse/query-core/src/test/java/datawave/query/planner/rules/NodeTransformVisitorTest.java
```diff
@@ -112,7 +117,24 @@ public void regexPushdownAnyfieldTransformRuleTest() throws Exception {
...
+
+    @Test
+    public void regexSimplifierTransformRuleTest() throws Exception {
+        // @formatter:off
+        String query = "BLA == '.*?.*?x' && " +
+                "BLA =~ 'ab.*.*' && " +
+                "BLA =~ 'a.*.*.*.*?.*?' && " +
+                "BLA =~ '.*?.*?.*bla.*?.*?blabla' && " +
+                "_ANYFIELD_ =~ '.*.*?.*?<bla>'";
+        String expected = "BLA == '.*?.*?x' && " +
```
FineAndDandy, line 129: shouldn't this be simplified to `BLA == '.*?x'`?

ivakegg: That is an equals node, not a regex node.

ivakegg: That is an equals node, not a regex node.

FineAndDandy: ha, right. You got me

PR #838, opened 2020-06-23
Fixes #713 auto close failed queries

### LINKED ISSUE #713: auto close failed queries
We now automatically close a query when a 204 is returned (no more results).  We should also automatically close a query when a failure condition is returned (400+).

rcarterjr:
Closes query in an else condition that left the query open after an exception is thrown.


ivakegg:
It appears we have some failing test cases.... hmm...

rcarterjr:
> It appears we have some failing test cases.... hmm...

https://github.com/NationalSecurityAgency/datawave/blob/ff75a77b703096889477a74081b2ad8733ce53b4/web-services/query/src/test/java/datawave/webservice/query/runner/ExtendedQueryExecutorBeanTest.java#L2664

This is the line of the assetEquals that failed

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -2026,6 +2026,7 @@ private BaseQueryResponse next(final String id, boolean checkForContentLookup) {
                 close(id);
                 closedQueryCache.add(id); // remember that we auto-closed this query
             } else {
+                close(id);
```
ivakegg, line 2029: It appears that this close may throw an exception in certain cases.  So this close should be wrapped in a try catch and log the error:
```
try {
  close(id);
} catch (Exception ce) {
  log.error(qe, ce);
}
```

PR #842, opened 2020-06-27
Add support for multiple ActiveQueryLog instances

### LINKED ISSUE #817: Active query log should handle multiple tables separately
The Active query log should track queries separately against the shard, errorShard, and queryMetricsE tables.

lbschanno:
Currently an ActiveQueryLog singleton is used for logging events from
all active queries. It can be difficult to easily search the logs for
queries that are running on specific tables.

Add support to be able to retrieve a new ActiveQueryLog instance
associated with a configurable name, and include this name in log
messages generated by the ActiveQueryLog. Keep support for a default
ActiveQueryLog instance that can be shared by queries when additional
search details is not needed.

In addition, add the ability to specifiy a source for the active query
log name for ShardQueryLogic.

Fixes #817


lbschanno:
In `QueryOptions.describeOptions`, the value for METADATA_TABLE_NAME is set twice. Which should be kept? 
line 1032: `options.put(METADATA_TABLE_NAME, "The name of the metadata table");`
line 1103: `options.put(METADATA_TABLE_NAME, this.metadataTableName);`

lbschanno:
In `ActiveQueryLog.setMaxIdle(long maxIdle)`, this.maxIdle is never set to the new maxIdle. Is this desired?

warehouse/query-core/src/test/java/datawave/query/config/ShardQueryConfigurationTest.java
```diff
@@ -434,17 +436,46 @@ public void testIsTldQuery() {
...
+            
             objectCount++;
         }
         
-        Assert.assertEquals("New variable was added to or removed from the ShardQueryConfiguration", expectedObjectCount, objectCount);
+        assertEquals("New variable was added to or removed from the ShardQueryConfiguration", expectedObjectCount, objectCount);
+    }
+    
+    @Test
+    public void whenRetrievingActiveQueryLogName_givenTableNameSource_thenReturnsTableName() {
```
ivakegg, line 455: ouch!  How about a shorter name and some comments above it.

warehouse/query-core/src/test/java/datawave/query/tracking/ActiveQueryLogTest.java
```diff
@@ -90,6 +95,33 @@ public void run() {
         }
     }
     
+    @Test
+    public void whenRetrievingInstance_givenNullName_shouldReturnsDefaultInstance() {
```
ivakegg, line 99: ouch! How about a shorter name and some comments above it.

ivakegg:
@billoley Please take a look at this one.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -465,21 +477,6 @@ public void deepCopy(QueryOptions other) {
         
         this.sortedUIDs = other.sortedUIDs;
         
-        this.compressedMappings = other.compressedMappings;
```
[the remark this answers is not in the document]
lbschanno: Heh, you got me there.

warehouse/query-core/src/test/java/datawave/query/config/ShardQueryConfigurationTest.java
lbschanno: I agree that one shouldn't use overly long method names when writing product code, but I don't think the same should apply to unit tests. By using a when<behavior_under_test>_given<factor>_then<expected_behavior>, I'm able to state a single technical requirement being tested (which works well for TDD) and make the method self documenting. If a test fails during a build, I'll be able to know at a glance in the logs what was being tested and which requirement failed. I think this naming style also makes it easier for other developers to clearly understand each test case.

I have found that when testing multiple methods with multiple technical requirements for each method, writing short but still clear method names can get pretty tricky. Granted, that's not necessarily the case here and I can shorten the names a bit. I hope I've been able to explain my reasoning clearly; this is a habit formed from many years of TDD.

ivakegg:
> In `ActiveQueryLog.setMaxIdle(long maxIdle)`, this.maxIdle is never set to the new maxIdle. Is this desired?

That appears to be an oversight as far as I can tell.  @billoley do you agree?

ivakegg:
> In `QueryOptions.describeOptions`, the value for METADATA_TABLE_NAME is set twice. Which should be kept?
> line 1032: `options.put(METADATA_TABLE_NAME, "The name of the metadata table");`
> line 1103: `options.put(METADATA_TABLE_NAME, this.metadataTableName);`

I would have to say line 1032 is incorrect and can be removed.  Good find.

lbschanno:
> I would have to say line 1032 is incorrect and can be removed. Good find.

@ivakegg no problem, I've just removed line 1032.

billoley:
> > In `ActiveQueryLog.setMaxIdle(long maxIdle)`, this.maxIdle is never set to the new maxIdle. Is this desired?
> 
> That appears to be an oversight as far as I can tell. @billoley do you agree?

I agree -- looks like an oversight that should be fixed

billoley:
LGTM

Just fix the bug in setMaxIdle noted above

One area that I was looking into is that when we change classloaders on Datawave upgrades, static objects and the things they reference hang around.  Before this PR, that would be the AQL instance and the timer threads.  Now it's the cache of AQLs and their timer threads.  The periodic check and shutdown of those timer threads should work they way they did before.... just that there might be more of them after this PR.

warehouse/query-core/src/test/java/datawave/query/tracking/ActiveQueryLogTest.java
lbschanno: I shortened the test case names a bit.

lbschanno:
Fixed the bug in setMaxIdle and shortened the test case names a bit.

FineAndDandy (approved):

ivakegg (approved):

ivakegg:
@lbschanno if you can resolve those conflicts then I will merge this one in.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -1013,7 +1030,6 @@ public IteratorOptions describeOptions() {
         options.put(QUERY, "The JEXL query to evaluate documents against");
         options.put(QUERY_ID, "The UUID of the query");
         options.put(TYPE_METADATA, "A mapping of field name to a set of DataType class names");
-        options.put(METADATA_TABLE_NAME, "The name of the metadata table");
```
billoley, line 1016: In describeOptions, this:
options.put(METADATA_TABLE_NAME, "The name of the metadata table");
is the correct line to leave in and should remain
options.put(METADATA_TABLE_NAME, this.metadataTableName);
below is not descriptive and should be deleted

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
```diff
@@ -2061,6 +2073,35 @@ public void setEvaluationOnlyFields(Set<String> evaluationOnlyFields) {
...
+     * Returns the name to use for the active query log for query iterators, derived from the value {@link #activeQueryLogNameSource}. Cases:
+     * <ul>
+     * <li>{@value TABLE_NAME_SOURCE}: returns the shard table name</li>
+     * <li>{@value QUERY_LOGIC_NAME_SOURCE}: returns the name of the shard query logic class</li>
+     * <li>otherwise returns a blank value</li>
+     * </ul>
+     *
+     * @return the custom active query name to use, or a blank value if the default active query log should be used
+     */
+    @JsonIgnore
```
billoley, line 2094: can be converted into a switch/case structure for readability

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -314,7 +314,7 @@ public void enableYielding(YieldCallback<Key> yieldCallback) {
     
     @Override
     public void next() throws IOException {
-        ActiveQueryLog.getInstance().get(getQueryId()).beginCall(this.originalRange, ActiveQuery.CallType.NEXT);
+        ActiveQueryLog.getInstance(getActiveQueryLogName()).get(getQueryId()).beginCall(this.originalRange, ActiveQuery.CallType.NEXT);
```
billoley, line 317: We should reduce unnecessary calls and cache accesses by declaring 
final ActiveQueryLog activeQueryLog = ActiveQueryLog.getInstance(getActiveQueryLogName());
and using  activeQueryLog in place of ActiveQueryLog.getInstance(getActiveQueryLogName()) in this method

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -345,10 +345,10 @@ public void seek(Range range, Collection<ByteSequence> columnFamilies, boolean i
         // preserve the original range for use with the Final Document tracking iterator because it is placed after the ResultCountingIterator
         // so the FinalDocumentTracking iterator needs the start key with the count already appended
         originalRange = range;
-        ActiveQueryLog.getInstance().get(getQueryId()).beginCall(this.originalRange, ActiveQuery.CallType.SEEK);
+        ActiveQueryLog.getInstance(getActiveQueryLogName()).get(getQueryId()).beginCall(this.originalRange, ActiveQuery.CallType.SEEK);
```
billoley, line 348: We should reduce unnecessary calls and cache accesses by declaring 
final ActiveQueryLog activeQueryLog = ActiveQueryLog.getInstance(getActiveQueryLogName());
and using  activeQueryLog in place of ActiveQueryLog.getInstance(getActiveQueryLogName()) in this method

billoley (changes requested):

billoley:
> > In `QueryOptions.describeOptions`, the value for METADATA_TABLE_NAME is set twice. Which should be kept?
> > line 1032: `options.put(METADATA_TABLE_NAME, "The name of the metadata table");`
> > line 1103: `options.put(METADATA_TABLE_NAME, this.metadataTableName);`
> 
> I would have to say line 1032 is incorrect and can be removed. Good find.

Looks like I was wrong here.  The description in 1032 should stay and 1103 get deleted.   Sorry.

lbschanno:
@billoley I've implemented your suggestions, let me know if there's anything else that should be done.

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
```diff
@@ -100,10 +100,9 @@ public static ActiveQueryLog getInstance(String name) {
         }
         
         // If no log currently exists for the name, create one.
```
billoley, line 102: This should go back to the way it was before the last commit so that we check for the named AQL in the logCache and add it if necessary all inside the synchronized block

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
```diff
@@ -4,83 +4,139 @@
...
-    synchronized public static void setConfig(AccumuloConfiguration conf) {
+    private final String name;
+    
+    synchronized public static void setConfig(final AccumuloConfiguration conf) {
         if (conf != null) {
             if (ActiveQueryLog.conf == null || conf.getUpdateCount() > ActiveQueryLog.conf.getUpdateCount()) {
                 ActiveQueryLog.conf = conf;
-                ActiveQueryLog.getInstance().checkSettings(conf, false);
+            }
+            // Do not allow access to the cache while updating each log's settings.
```
billoley, line 57: This is where I thought the  if (logCache != null) check could be outside of the synchronized

billoley (changes requested):

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
billoley: The other changes look fine.  I just have these two issues with the synchronization changes in AQL.

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
lbschanno: Fixed.

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
lbschanno: Fixed.

keith-ratcliffe:
Updating PR branch which is out of date with current release/version3.2. That should fix the quickstart deploy test

billoley (approved):

ivakegg (approved):

PR #855, opened 2020-07-17
fixes #854: QueryPruningVisitor reduces AndNode and OrNode

### LINKED ISSUE #854: QueryPruningVisitor should reduce And/Or nodes
QueryPruningVisitor will attempt to prune whole AndNode and OrNode structures but will not reduce children. 

When a child is false within an ORNode it can be pruned

When a child is true within an AndNode it can be pruned

FineAndDandy:


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPruningVisitor.java
```diff
@@ -516,10 +540,28 @@ private void replaceAndAssign(JexlNode toReplace, String queryString, JexlNode b
...
                 
-                JexlNodes.swap(parent, toReplace, baseReplacement);
+                if (baseReplacement != null) {
+                    JexlNodes.swap(parent, toReplace, baseReplacement);
+                } else {
+                    // remove the node entirely
+                    List<JexlNode> children = new ArrayList<>(parent.jjtGetNumChildren() - 1);
+                    for (int i = 0; i < parent.jjtGetNumChildren(); i++) {
+                        JexlNode child = parent.jjtGetChild(i);
+                        if (!child.equals(toReplace)) {
```
ivakegg, line 555: should/can this be (child != toReplace) instead of (!child.equals(toReplace))

FineAndDandy: yep it should be the exact object so we can use !=

Issue #886, opened 2020-07-31
Include query ids in failed query creates


ivakegg:
When queries fail in the create phase, the query id is not returned in the response object making it very difficult to get back to the query metrics.  We should return the queryid in the failure response somehow.

ivakegg:
Sorry @jzgithub1, but I think I need to take a different approach on this one.

PR #889, opened 2020-08-06
Include query ids in failed query creates #886  - Depends on base-rest-responses change

jzgithub1:
Added query id to the BaseResponse for queries.  This PR depends on changes in the base-rest-reponses project.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,7 +80,7 @@
         <version.microservice.accumulo-utils>1.3</version.microservice.accumulo-utils>
         <version.microservice.audit-api>1.4</version.microservice.audit-api>
         <version.microservice.authorization-api>1.4</version.microservice.authorization-api>
-        <version.microservice.base-rest-responses>1.2</version.microservice.base-rest-responses>
+        <version.microservice.base-rest-responses>1.2-SNAPSHOT</version.microservice.base-rest-responses>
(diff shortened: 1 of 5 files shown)


jzgithub1:
Solves #886

pom.xml
```diff
@@ -80,7 +80,7 @@
         <version.microservice.accumulo-utils>1.3</version.microservice.accumulo-utils>
         <version.microservice.audit-api>1.4</version.microservice.audit-api>
         <version.microservice.authorization-api>1.4</version.microservice.authorization-api>
-        <version.microservice.base-rest-responses>1.2</version.microservice.base-rest-responses>
```
keith-ratcliffe, line 83: I might be confused. There's already a 1.2 base-rest-responses jar deployed to github, so I'm not understanding how this PR can be dependent on a 1.2-SNAPSHOT. This PR should depend on a new 1.3 release. Likewise, your current base-rest-responses PR should be targeting 1.3-SNAPSHOT at the moment

Right?

jzgithub1: @keith-ratcliffe ,  @ivakegg  wanted me to do it this way I am pretty sure.  I had to change some code in that project  (base-rest-responses) for this pull request.  That is to say that there is an associated pull request in that project which I mention above.

ivakegg: I will fix the version when I merge this in.

PR #890, opened 2020-08-10
Update protostuff, type-utils, and metadata-utils to ensure they are …

jwomeara:
…in alignment.

@ivakegg


ivakegg:
Manually updated the type utils in release/version2.9 to 1.7 as additional change was required.

PR #897, opened 2020-08-17
Address spotbugs issues with TableConfigCache code

drewfarris:


warehouse/ingest-core/src/main/java/datawave/ingest/config/TableConfigCache.java
```diff
@@ -22,12 +23,13 @@
...
     protected static Map<String,String> configMap;
     
-    protected static Logger log = Logger.getLogger("datawave.ingest");
+    protected static final Logger log = Logger.getLogger("datawave.ingest");
     
     public TableConfigCache(Configuration conf) {
         super(conf);
     }
     
+    @SuppressFBWarnings("ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD")
```
drewfarris, line 32: It seems like this was the best action to take here - it was complaining that we were writing to a static field from a non-static method. Perhaps there's another way to do this that would not result in a warning?

hlgp: I'll investigate this when I get back. Approving for now. Long term we shouldn't  need this suppression.

jschmidt10: Looks like `configMap` is static but we're setting it in a non-static method. The method actually has no dependency on any instance state. Seems like the fix might be to make this method static.

drewfarris: @jschmidt10 - I suspected the same thing but was going to leave the design up to @hlgp.

jschmidt10: Sounds good!

PR #899, opened 2020-08-18
Add QueryPlanMetadata, QueryPlanMetadataVisitor. 

apmoriarty:
Add QueryPlanMetadata, QueryPlanMetadataVisitor. Integreate into DefaultQueryPlanner.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
+++ b/warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
@@ -56,6 +56,7 @@ import datawave.query.jexl.visitors.GeoWavePruningVisitor;
 import datawave.query.jexl.visitors.IsNotNullIntentVisitor;
 import datawave.query.jexl.visitors.IvaratorRequiredVisitor;
 import datawave.query.jexl.visitors.JexlStringBuildingVisitor;
+import datawave.query.jexl.visitors.QueryPlanMetadataVisitor;
 import datawave.query.jexl.visitors.QueryPruningVisitor;
 import datawave.query.jexl.visitors.RewriteNegationsVisitor;
 import datawave.query.jexl.visitors.ParallelIndexExpansion;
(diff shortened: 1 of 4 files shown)


warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -889,7 +890,12 @@ protected ASTJexlScript processTree(final ASTJexlScript originalQueryTree, Shard
         
         TraceStopwatch stopwatch = null;
         
-        if (!disableExpandIndexFunction) {
+        // Parse out some metadata about the query plan to make informed decisions as the query tree is processed.
+        stopwatch = timers.newStartedStopwatch("DefaultQueryPlanner - Build QueryPlanMetadata");
+        QueryPlanMetadata planMetadata = QueryPlanMetadataVisitor.getQueryPlanMetadata(queryTree);
+        stopwatch.stop();
+        
+        if (planMetadata.hasFunctionNodes() && !disableExpandIndexFunction) {
```
ivakegg, line 898: Does this actually save us any time if we simply need to re-gather the plan metadata once anything has changed?

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -1166,108 +1173,141 @@ protected ASTJexlScript processTree(final ASTJexlScript originalQueryTree, Shard
             stopwatch.stop();
         }
         
-        if (!disableBoundedLookup) {
-            stopwatch = timers.newStartedStopwatch("DefaultQueryPlanner - Expand bounded query ranges");
-            
-            // Expand any bounded ranges into a conjunction of discrete terms
+        ParallelIndexExpansion regexExpansion = null;
+        if (!disableBoundedLookup && (planMetadata.hasRegexNodes() || planMetadata.hasAnyFieldNodes())) {
```
ivakegg, line 1177: Wouldn't we need to recalculate the plan metadata just before this since many steps have occurred since its calculation?

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -1166,108 +1173,141 @@ protected ASTJexlScript processTree(final ASTJexlScript originalQueryTree, Shard
...
+            }
+        }
+        
+        if (!disableBoundedLookup && planMetadata.boundedRangesPossible()) {
+            try {
+                stopwatch = timers.newStartedStopwatch("DefaultQueryPlanner - Expand bounded query ranges");
                 queryTree = RangeConjunctionRebuildingVisitor.expandRanges(config, scannerFactory, metadataHelper, queryTree, config.isExpandFields(),
                                 config.isExpandValues());
-                if (log.isDebugEnabled()) {
+            } catch (TableNotFoundException | ExecutionException e) {
```
ivakegg, line 1204: The range conjunction rebuilding visitor may end up removing bounded ranges if they expanded to the empty set making the planmetadata invalid yet again.

ivakegg (changes requested):

ivakegg:
I am not convinced this is more efficient nor more readable.

apmoriarty:
I agree -- I don't think this actually provides any value

PR #900, opened 2020-08-18
Increase DefaultQueryPlanner stopwatch granularity

### LINKED ISSUE #865: More granular logging for ParallelIndexExpansion
There is a section of the DefaultQueryPlanner that does not log timing details with enough granularity.  Currently the times for each stage of the planning are currently logged and the method for doing that is well used.  We simply need to separate each step within the "DefaultQueryPlanner - Expand bounded query ranges" stage with a separate stopwatch.

lbschanno:
Increase the granularity of logging the timing details for expanding
bounded query ranges in the DefaultQueryPlanner with a separate
stopwatch.

Fixes #865

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
+++ b/warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
@@ -115,6 +115,7 @@ import org.apache.commons.jexl2.parser.ParseException;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.time.DateUtils;
 import org.apache.log4j.Logger;
+import org.graalvm.compiler.core.common.alloc.Trace;
 
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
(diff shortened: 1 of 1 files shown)


apmoriarty:
This is pretty close to what we want but there is one problem. You should be using the timers provided by the shard query config so that the timers get logged at the end ([see this section of code in the ShardQueryLogic](https://github.com/NationalSecurityAgency/datawave/blob/master/warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java#L556-L562)).

I suppose there is a question if we even need the surrounding timer tracking the total time, but I'll leave that to @ivakegg

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -115,6 +115,7 @@
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.time.DateUtils;
 import org.apache.log4j.Logger;
+import org.graalvm.compiler.core.common.alloc.Trace;
```
alerman, line 118: Am I missing where this is getting used?

lbschanno: You're not, it was accidentally imported. I'll remove it.

lbschanno:
@apmoriarty the timers used to create the new stopwatch seems to originate from a call from updateQueryTrees(), which in turn gets the timers from a supplied ShardQueryConfiguration argument.

I wasn't sure if the surrounding timer should be preserved, but the ticket itself specified a separate stopwatch, so I decided to play it safe.

apmoriarty:
My mistake, you are indeed using the stopwatch provided from the query config. 

LGTM.

Issue #788, opened 2020-03-20
Reduce size of the ModelBean.getModel()


ivakegg:
The (Query) Models can be fairly large causing issues in browsers. One idea is to allow adding a filter parameter to the REST endpoints to reduce the data being returned. Another idea is to provide a configurable paging mechanism within the returned pages.

PR #901, opened 2020-08-21
Add pagination support for ModelBean API

lbschanno:
This PR provides a partial fix for #788; namely, it provides a back-end implementation that supports pagination, sorting, and filtering of the field mappings returned when performing a GET against /DataWave/Model/{name} endpoint. This effectively fixes the issue when retrieving JSON, XML, etc., but not when retrieving the default HTML page with the embedded jquery.dataTables table. I tried to make a fix for that, but I found myself struggling and too unfamiliar with jquery.datatables to get it working. Hopefully someone else will build on the work here to update the HTML response.

The following parameters are now available:
* limit - total number of field mappings to return. -1 for no limit (default).
* offset - total number of initial field mappings to skip. 0 for no offset (default).
* sort - the sort order to apply to the field mappings in the model. This can be specified in several ways:
  * sort="asc" - Sort fields in natural order (default). 
  * sort="desc" - Sort fields in reverse natural order.
  * sort="asc:{field}" - Sort in ascending order by the specified property.
  * sort="desc:{field}" - Sort in descending order by the specified property.
  * Supported properties: fieldName, modelFieldName, dataType, direction, visibility
* search - case-insensitive, partial matching against the fieldName, modelFieldName, dataType, direction, visibility.


web-services/model/src/main/java/datawave/webservice/query/model/ModelKeyParser.java
```diff
@@ -5,39 +5,37 @@
...
     
     public static final String NULL_BYTE = "\0";
     public static final Value NULL_VALUE = new Value(new byte[0]);
-    private static Logger log = Logger.getLogger(ModelKeyParser.class);
     
-    public static FieldMapping parseKey(Key key, Set<Authorizations> auths) {
-        FieldMapping mapping = new FieldMapping();
+    private static final Logger log = Logger.getLogger(ModelKeyParser.class);
+    
+    public static FieldMapping parseKey(Key key) {
```
drewfarris, line 21: I haven't read the entire PR yet, so the answer to this may be elsewhere, but I'm curious why we have dropped the columnVisibility from the `FieldMapping` produced by this method.

From what I can tell `auths` was unused so it might sense to get rid of them here, but I wonder if there are subclasses elsewhere that may depend on this method signature.

web-services/model/src/main/java/datawave/webservice/query/model/ModelKeyParser.java
```diff
@@ -5,39 +5,37 @@
...
         String row = key.getRow().toString();
-        String[] colf = key.getColumnFamily().toString().split(NULL_BYTE);
         String[] colq = key.getColumnQualifier().toString().split(NULL_BYTE);
-        String cv = key.getColumnVisibility().toString();
+        String visbility = key.getColumnVisibility().toString();
```
lbschanno, line 24: I need to correct the spelling of this variable.

web-services/model/src/main/java/datawave/webservice/query/model/ModelKeyParser.java
lbschanno: @drewfarris I did not have any issues with subclasses needing the original `parseKey(Key key, Set<Authorizations> auths)` method signature, but I can add it back as an overloaded method and call `parseKey(Key key)` within it.

ivakegg: It appears that the visibility was added back in.  I am checking to see if there are any other dependencies on parseKey(key, auths)

ivakegg: confirmed that the only place parseKey was called is in the ModelBean.  So dropping the arg is ok

web-services/model/src/main/java/datawave/webservice/query/model/ModelBean.java
```diff
@@ -326,70 +305,160 @@ public VoidResponse cloneModel(@Required("name") @FormParam("name") String name,
...
-        Set<Authorizations> cbAuths = new HashSet<>();
-        if (p instanceof DatawavePrincipal) {
-            DatawavePrincipal cp = (DatawavePrincipal) p;
-            user = cp.getShortName();
-            for (Collection<String> auths : cp.getAuthorizations()) {
-                cbAuths.add(new Authorizations(auths.toArray(new String[auths.size()])));
-            }
-        }
-        log.trace(user + " has authorizations " + cbAuths);
+    public Model getModel(@Required("name") @PathParam("name") String name, @QueryParam("modelTableName") String tableName,
```
ivakegg, line 308: It is imperative that we still support the previous interface.  So all additional parameters need to have defaults or we need to create a completely separate endpoint.

ivakegg (changes requested):
Please ensure backward compatibility.  If needed lets create a "v2" interface.

PR #903, opened 2020-08-24
Do not swallow DatawaveInterpreter.invocationFailed exceptions

### LINKED ISSUE #887: Exceptions thrown when evaluating functions are silently swallowed
DatawaveInterpreter uses strict=false during evaluation which results in no error being thrown if an exception is encountered during a method invocation (any function).  There are multiple places in the interpreter code which test strict before throwing exceptions.  Specifically the invocationFailed() is responsible for deciding to throw exceptions from inside called functions.  This may result in missed hits.

If a function in the query throws an exception, it should error the query.  Note that setting strict=false may be required for other reasons (to be verified).

lbschanno:
Currently during a evaluation when strict==false, DatawaveInterpreter
swallows errors when an exception is encountered during a method
invocation (any function).

Override invocationsFails in DatawaveInterpreter and make it always
throw the given exception.

Fixes #887.


lbschanno:
@ivakegg all test failures have been fixed.

ivakegg (approved):

ivakegg (dismissed):

ivakegg (approved):

PR #904, opened 2020-08-24
fixes #858 Add EventFieldIterator to merge values with tf iterator for document ranges

### LINKED ISSUE #858: Document Specific ranges against regex'd TF fields should match against full value as well as tokens
Shard ranges will match against the full value because the FI holds the full value and the tokens, but the TF only holds the tokens. The behavior should be consistent regardless of the range.

FineAndDandy:


FineAndDandy:
realized I didn't deal with applying normalizers to the field... will be another commit to fix, hold reviews until then please

FineAndDandy:
Adding normalization support ended up taking a bit longer than expected. PR should be ready for review now.

FineAndDandy:
not sure what is going on with the tests here. I have no test failures locally. I will checkout the PR and see if I get errors building it.

FineAndDandy:
a checkout of the PR builds clean for me. Can anyone else verify a problem?

FineAndDandy:
I'm going to reopen a new PR and see if that doesn't fix the weird build issue here.

PR #908, opened 2020-08-27
fixes #907 Implement removeIf on HdfsBackedSortedSet

### LINKED ISSUE #907: HdfsBackedSortedSet should override java.util.Collection.removeIf(Predicate)
Calling the removeIf method on HdfsBackedSortedSet results in an exception if the set if persisted.  Just as we do in the remove() method, we should load the underlying FileSortedSet if persisted to enable executing the remove.

ivakegg:
* Implemented removeIf on the supper class BufferedFileBackedSortedSet

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
+++ b/warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
@@ -394,7 +394,7 @@ public class BufferedFileBackedSortedSet<E> implements SortedSet<E> {
         }
         return false;
     }
-
+    
     @Override
     public boolean remove(Object o) {
         boolean removed = false;
@@ -476,7 +476,7 @@ public class BufferedFileBackedSortedSet<E> implements SortedSet<E> {
         }
         return modified;
     }
-
+    
     @Override
     public boolean removeIf(Predicate<? super E> filter) {
         boolean removed = false;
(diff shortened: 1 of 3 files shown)


warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
```diff
@@ -476,6 +477,33 @@ public boolean removeAll(Collection<?> c) {
...
+        boolean removed = false;
+        for (SortedSet<E> subSet : set.getSets()) {
+            FileSortedSet<E> fileSet = (FileSortedSet<E>) subSet;
+            if (fileSet.isPersisted()) {
+                try {
+                    fileSet.load();
+                    if (fileSet.removeIf(filter)) {
+                        removed = true;
+                    }
+                    fileSet.persist();
```
FineAndDandy, line 491: only persist if removed?

FineAndDandy (changes requested):

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
ivakegg: nope, I wanted to put it back into a persisted state if it was previously persisted.

ivakegg: But I get your point, if nothing changed then why rewrite the files.  I will create a separate ticket to add a mechanism to flip from loaded to persisted using the existing underlying files.

ivakegg: see #949

PR #909, opened 2020-08-31
fixes #858 repost

### LINKED ISSUE #858: Document Specific ranges against regex'd TF fields should match against full value as well as tokens

FineAndDandy:


warehouse/query-core/src/main/java/datawave/query/iterator/logic/EventFieldNormalizingIterator.java
```diff
@@ -0,0 +1,248 @@
...
+import java.util.SortedSet;
+import java.util.TreeSet;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Iterator expects Event Key/Value pairs and will filter to only matching field and apply normalizations. This will result in one Key generated per normalizer
+ * applied to a field
+ */
+public class EventFieldNormalizingIterator implements SortedKeyValueIterator<Key,Value> {
```
ivakegg, line 35: It seems to me that we are doing a lot of work to try and keep things sorted when all we are doing is aggregating these into a Document.  Would a better approach be to modify the aggregator to simply do the normalization?

warehouse/query-core/src/main/java/datawave/query/iterator/EventFieldIterator.java
```diff
@@ -0,0 +1,163 @@
...
+        
+        // test current source key to determine state
+        if (!source.hasTop()) {
+            // no key can match the underlying source is empty
+            return null;
+        }
+        
+        Key top = source.getTopKey();
+        
+        if (minimum.compareTo(top) < -1) {
```
apmoriarty, line 63: Should be `compareTo() < 0`

warehouse/query-core/src/main/java/datawave/query/iterator/logic/EventFieldNormalizingIterator.java
```diff
@@ -0,0 +1,248 @@
...
+    
+    private boolean initialized = false;
+    
+    public EventFieldNormalizingIterator(String field, SortedKeyValueIterator<Key,Value> delegate, TypeMetadata typeMetadata, String defaultTypeClass) {
+        this.field = field;
+        this.delegate = delegate;
+        this.typeMetadata = typeMetadata;
+        this.defaultTypeClass = defaultTypeClass;
+    }
+
```
apmoriarty, line 64: the consistency in these two constructors is killing me.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EventFieldAggregator.java
```diff
@@ -0,0 +1,67 @@
...
+        return columnFamily;
+    }
+    
+    @Override
+    public Key apply(SortedKeyValueIterator<Key,Value> itr, Document doc, AttributeFactory attrs) throws IOException {
+        Key result = super.apply(itr, doc, attrs);
+        
+        // for each thing in the doc, mark it as to-keep false because it will ultimately come from the document aggregation, otherwise there will be duplicates
+        for (Attribute<?> attr : doc.getDictionary().values()) {
+            attr.setToKeep(false);
```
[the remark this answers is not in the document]
FineAndDandy: that's fair, I will create a new document that I pass to apply() then merge it into doc.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/EventFieldNormalizingIterator.java
```diff
@@ -0,0 +1,248 @@
...
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Iterator expects Event Key/Value pairs and will filter to only matching field and apply normalizations. This will result in one Key generated per normalizer
+ * applied to a field
+ */
+public class EventFieldNormalizingIterator implements SortedKeyValueIterator<Key,Value> {
+    // speedy cache loading for types, duplicated from AttributeFactory with caching of types rather than classes
+    protected static LoadingCache<String,Type<?>> typeCache = CacheBuilder.newBuilder().maximumSize(128).expireAfterAccess(1, TimeUnit.HOURS)
+                    .build(new CacheLoader<String,Type<?>>() {
```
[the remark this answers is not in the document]
FineAndDandy: This is the same hard coded value as the AttributeFactory cache.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/EventFieldNormalizingIterator.java
FineAndDandy: I originally started there, but because it has a 1:many relationship due to normalizers it was going to require interface changes to `IdentityAggregator`. We can still go with that if you prefer.

warehouse/query-core/src/main/java/datawave/query/iterator/EventFieldIterator.java
FineAndDandy: yep

warehouse/query-core/src/main/java/datawave/query/iterator/logic/EventFieldNormalizingIterator.java
ivakegg: If it makes the code simpler, then I am all for it.  We have added several new classes here whereas an api change and one simpler extension would be easier to understand and maintain.  I guess the question is how many other places will this affect my changing the interface to return 1:many relationships.

FineAndDandy: It requires changing the parserFieldValues() to return a list rather than a single value, then updating all current methods to return singletons.

warehouse/query-core/src/main/java/datawave/query/iterator/EventFieldIterator.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+    
+    @Override
+    public void setContext(Key context) {
+        // no-op
+    }
+    
+    @Override
+    public boolean hasNext() {
+        // do the actual seeking now if it hasn't be done yet
```
ivakegg, line 107: be -> been

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EventFieldAggregator.java
```diff
@@ -0,0 +1,144 @@
...
+    @Override
+    protected List<Tuple2<String,String>> parserFieldNameValue(Key topKey) {
+        String cq = topKey.getColumnQualifier().toString();
+        int nullIndex1 = cq.indexOf('\u0000');
+        String field = cq.substring(0, nullIndex1);
+        String value = cq.substring(nullIndex1 + 1);
+        
+        int dataTypeEnd = -1;
+        ByteSequence cf = topKey.getColumnFamilyData();
+        for (int i = 0; i < cf.getBackingArray().length; i++) {
```
ivakegg, line 60: The valid bytes in cf.getBackingArray() only go through cf.length().  The backing array could be longer.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EventFieldAggregator.java
```diff
@@ -0,0 +1,144 @@
...
+        for (String normalizedValue : normalizedValues) {
+            fieldValuePairs.add(new Tuple2<>(field, normalizedValue));
+        }
+        
+        return fieldValuePairs;
+    }
+    
+    @Override
+    protected ByteSequence parseFieldNameValue(ByteSequence cf, ByteSequence cq) {
+        ArrayList<Integer> nulls = TLD.instancesOf(0, cq, 1);
```
ivakegg, line 83: Seems to me this whole method could be replaced with simply returning cq.  If you need a copy of the bytes then return new ArrayByteSequence(Arrays.copyOf(cq.getBackingArray(), cq.length()), 0, cq.length()).

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EventFieldAggregator.java
```diff
@@ -0,0 +1,144 @@
...
+        
+        // fetch all Types for the field/dataType combination
+        Collection<String> typeClasses = typeMetadata.getTypeMetadata(fieldName, dataType);
+        
+        // if its not found add the default
+        if (typeClasses.size() == 0) {
+            typeClasses.add(defaultTypeClass);
+        }
+        
+        // transform the key for each type and add it to sorted
```
ivakegg, line 130: This is not actually doing any sorting.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/iterator/EventFieldIterator.java
FineAndDandy: got it

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EventFieldAggregator.java
FineAndDandy: good catch

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EventFieldAggregator.java
FineAndDandy: you are right, there is no reason to do this. I don't think there is any reason not to just return the cq.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EventFieldAggregator.java
FineAndDandy: fixed docs

apmoriarty (dismissed):
I think all comments have been addressed

ivakegg (dismissed):

FineAndDandy:
what is this waiting on for merge?

drewfarris:
Let's retarget this to a newer branch, @ivakegg will be creating a release/version3.2 branch

FineAndDandy:
I will retarget this against release/version3.2

FineAndDandy:
I'm working on resolving conflicts. Having some issues with unit tests after the merge

FineAndDandy:
we will need to wait for https://github.com/NationalSecurityAgency/datawave-metadata-utils/pull/7 to be merged and available in release/version3.2 for this ticket to be merged.

FineAndDandy:
this is dependent on #1104

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/TLDEventFieldAggregator.java
```diff
@@ -0,0 +1,34 @@
...
+        final int start = nulls.get(0) + 1;
+        
+        // uid is from the null byte to the end of the cf
+        ByteSequence uid = columnFamily.subSequence(start, columnFamily.length());
+        
+        // find the end of the tld if it exists
+        ArrayList<Integer> dots = TLD.instancesOf('.', uid);
+        if (dots.size() > 2) {
+            // reduce to the TLD
+            return columnFamily.subSequence(0, start + dots.get(2));
```
ivakegg, line 28: I think this is making an assumption about how we create UIDs.  If we turned on the timestamp portion of the UID then this would break or if we decide to substitute an alternate UID mechanism as it is currently pluggable.  Can we simply look for the last dot and strip that off?

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/TLDEventFieldAggregator.java
apmoriarty: What if there's more than one dot.

FineAndDandy: I updated the code to use UID to get the base UID

FineAndDandy: Generally this query logic is probably not configured right if the UID scheme changes

ivakegg (approved):

apmoriarty (approved):

PR #917, opened 2020-09-09
Fixes #839 Cap query end date - WIP

### LINKED ISSUE #839: Cap query end date in certain cases, responsibly
It's possible to specify a date range that exceeds the limit for data stored. A specific example may be seen in RangeStream.createFullFieldIndexScanList(). 

It should be possible to scan the metadata table to determine the most recent shards and then limit the query end date based on that.

rcarterjr:
Once I have a test with the necessary metadata dates to compare I should be able to fix the MetadataHelper function.

Thanks @ivakegg


warehouse/query-core/src/test/java/datawave/query/index/lookup/RangeStreamTest.java
```diff
@@ -60,20 +63,49 @@
     @BeforeClass
     public static void setupAccumulo() throws Exception {
         
+        // just like this shard index i am going to need to
+        // create a metadata table
+        
+        final String METADATA_TABLE = "metadataTable";
```
apmoriarty, line 69: I realize this is a WIP, you can use the static names defined in TableName instead of defining these locally.

Issue #880, opened 2020-07-31
Multiple  JexlNode rebuilding visitors are not preserving proper parentage


ivakegg:
There is a new validateLineage method being added to JexlASTHelper which attempts to validate that for a given Jexl tree:
1) every child node has a non-null parent equal to the parent who listed that node as a child
2) all children are non-null
Using this method we can determine where in the planning process (DefaultQueryPlanner) that we are creating invalid trees.  Those steps/rebuilding visitors need to be fixed.

One known area is when when getQueryPropertySource is called on the QueryPropertyMarker.

Once everything is fixed, we should be able to call validateLineage before and after every step in the DefaultQueryPlanner with no errors.

lbschanno:
Closing this ticket since the bulk of visitors now produce trees with correct lineages. If new issues are discovered with visitors, new issues can be created as needed.

Issue #825, opened 2020-05-28
Handle the cases were a field is both indexed and not indexed within a time range 


cawaring:
Fields can be added or removed from the index over time.  We do not always handle cases where a field spans a time range where it's both indexed and not index.  We need to handle this case appropriately.  This would include using the field as an indexed field for the portion of time it is indexed and avoid using it to prune queries and ranges when it's not indexed.

ivakegg:
Suggest marking the time ranges in the datawave metadata table, and then using that info in the DefaultQueryPlanner appropriately.  The list of indexed fields sent to the QueryIterator will need to be adjusted for each range as appropriate.

ivakegg:
The Data Dictionary should also show this information.

ivakegg:
FYI, the PushdownMissingIndexRangeNodesVisitor is sort of an initial implementation of this capability.  However instead of pushing those nodes down, perhaps range stream index lookup can return a stream of shards for the holes, and also modify the query for those ranges to push the terms down to the evaluation phase.  The comment I put into another pull request may help one understand this: https://github.com/NationalSecurityAgency/datawave/pull/925#issuecomment-706319147.

ivakegg:
BTW having the ingest automatically fill in the metadata for when a field is indexed would be great as well.  This should probably be encoded in the same way we are condensing the F column here: https://github.com/NationalSecurityAgency/datawave/pull/828.  For existing systems we would need to manually update (a tool would be nice) the metadata for data already ingested.

jzgithub1:
> BTW having the ingest automatically fill in the metadata for when a field is indexed would be great as well. This should probably be encoded in the same way we are condensing the F column here: #828. For existing systems we would need to manually update (a tool would be nice) the metadata for data already ingested.

This is done with the new IndexColumnIterator class.

jzgithub1:
> BTW having the ingest automatically fill in the metadata for when a field is indexed would be great as well. This should probably be encoded in the same way we are condensing the F column here: #828. For existing systems we would need to manually update (a tool would be nice) the metadata for data already ingested.

I put up another ticket this.  The tool will be another Accumulo Iterator.  Won't take long to implement.

jzgithub1:
[...]

PR #925, opened 2020-09-26
WIP Validate lineage for all JexlNode rebuilding visitors

lbschanno:
WIP for #880 for verifying/fixing JexlNode rebuilding visitors to ensure that they preserve proper parentage. 

I have fixed/verified parentage for the following visitors:

* `AllTermsIndexedVisitor`
* `BooleanOptimizationVisitor`
* `FacetCheck`
* `DateIndexCleanupVisitor`
* `ExpandMultinormalizedTerms`
* `TreeFlatteningRebuildingVisitor`
* `FixNegativeNumbersVisitor`
* `GeowavePruningVisitor`
* `QueryModelVisitor`
* `RangeCoalescingVisitor`
* `RegexFunctionVisitor`

I am currently working on fixing parentage for the following visitors:

* `UniqueExpressionTermsVisitor`
* `ExpandCompositeTerms`

There are a number of rebuilding visitors that do not have any unit tests associated with them that I could find, nor was there enough documentation for me to be able to extrapolate test cases for them. I would appreciate any help with examples or desired test cases so that I can write tests for each of the following and verify parentage.

* `BoundedRangeDetectionVisitor`
* `FunctionIndexQueryExpansionVisitor`
* `FunctionNormalizationRebuildingVisitor`
* `IsNotNullIntentVisitor`
* `ParallelIndexExpansion`
* `PruneLessSelectiveFieldsVisitor`
* `PushdownLargeFieldedListsVisitor`
* `PushdownMissingIndexRangeNodesVisitor`
* `PushFunctionsIntoExceededValueRanges`
* `RangeConjunctionRebuildingVisitor`
* `RangeExpansionThresholdRebuildingVisitor`

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/AllTermsIndexedVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/AllTermsIndexedVisitor.java
@@ -28,25 +28,29 @@ import org.apache.commons.jexl2.parser.ASTNRNode;
 import org.apache.commons.jexl2.parser.ASTOrNode;
 import org.apache.commons.jexl2.parser.JexlNode;
 import org.apache.commons.jexl2.parser.JexlNodes;
-import org.apache.commons.jexl2.parser.Node;
-import org.apache.commons.jexl2.parser.ParserTreeConstants;
 import org.apache.log4j.Logger;
 
 import java.text.MessageFormat;
+import java.util.Arrays;
 import java.util.NoSuchElementException;
+import java.util.Objects;
+import java.util.function.BiFunction;
+
+import static org.apache.commons.jexl2.parser.JexlNodes.promote;
 
 /**
- * 
+ *
  */
 public class AllTermsIndexedVisitor extends RebuildingVisitor {
+    
     private static final Logger log = Logger.getLogger(AllTermsIndexedVisitor.class);
     
     protected final ShardQueryConfiguration config;
     protected final MetadataHelper helper;
     
     public AllTermsIndexedVisitor(ShardQueryConfiguration config, MetadataHelper helper) {
-        Preconditions.checkNotNull(config);
-        Preconditions.checkNotNull(helper);
+        Preconditions.checkNotNull(config, "ShardQueryConfiguration must not be null");
+        Preconditions.checkNotNull(helper, "MetadataHelper must not be null");
         
         this.config = config;
         this.helper = helper;
@@ -54,7 +58,7 @@ public class AllTermsIndexedVisitor extends RebuildingVisitor {
     
     @SuppressWarnings("unchecked")
     public static <T extends JexlNode> T isIndexed(T script, ShardQueryConfiguration config, MetadataHelper helper) {
-        Preconditions.checkNotNull(script);
+        Preconditions.checkNotNull(script, "JEXL script must not be null");
         
         AllTermsIndexedVisitor visitor = new AllTermsIndexedVisitor(config, helper);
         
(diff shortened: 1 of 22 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/AllTermsIndexedVisitor.java
```diff
@@ -63,143 +67,91 @@ public AllTermsIndexedVisitor(ShardQueryConfiguration config, MetadataHelper hel
...
+        // @formatter:off
+        JexlNode[] children = Arrays.stream(JexlNodes.children(node))
+                        .map(n -> (JexlNode) n.jjtAccept(this, data)) // Visit the node
+                        .filter(Objects::nonNull) // Filter out null nodes
+                        .filter(n -> (!JexlNodes.isAnd(n) && !JexlNodes.isOr(n)) || JexlNodes.isNotChildless(n)) // Filter out empty junction nodes
+                        .toArray(JexlNode[]::new);
+        // @formatter:on
+        JexlNode copy = JexlNodes.newInstanceOfType(node);
+        copy.image = node.image;
+        System.out.println("Copy: " + copy);
```
ivakegg, line 116: Please remove System.out.println.  Use the logger if anything.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/AllTermsIndexedVisitor.java
```diff
@@ -63,143 +67,91 @@ public AllTermsIndexedVisitor(ShardQueryConfiguration config, MetadataHelper hel
...
                 return child;
             default:
-                return newNode;
+                return copy;
         }
     }
     
+    private JexlNode getCopyWithVisitedChildren(JexlNode node, Object data) {
+        // @formatter:off
+        JexlNode[] children = Arrays.stream(JexlNodes.children(node))
```
ivakegg, line 108: nice

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/BooleanOptimizationRebuildingVisitor.java
```diff
@@ -54,27 +50,22 @@ protected JexlNode optimizeTree(JexlNode currentNode, JexlNode newNode, Object d
...
-            
             JexlNode toAttach = nodes.second();
             toAttach = TreeFlatteningRebuildingVisitor.flatten(toAttach);
+            ASTOrNode newNode = new ASTOrNode(ParserTreeConstants.JJTORNODE);
+            newNode.image = currentNode.image;
             
             for (int i = 0; i < toAttach.jjtGetNumChildren(); i++) {
                 JexlNode node = copy(prunedNode);
                 JexlNode attach = (JexlNode) toAttach.jjtGetChild(i).jjtAccept(this, data);
+                attach.jjtSetParent(node);
```
ivakegg, line 61: I see lots of refactoring here but this is the only line that actually fixes a parenting problem for this class.  Refactoring for the sake of refactoring may do more harm than good if anybody else is working on these visitors for any other reason.  It also makes it very difficult for anybody to actually accept the entire package and merge.  I think you may need to pull back a little from the refactoring efforts.  Better yet a separate pull request per visitor that is only refactoring would be considerably more palatable.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryModelVisitor.java
```diff
@@ -1,16 +1,14 @@
...
-import com.google.common.collect.FluentIterable;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Sets;
-import datawave.query.model.QueryModel;
 import datawave.query.exceptions.DatawaveFatalQueryException;
 import datawave.query.jexl.JexlASTHelper;
 import datawave.query.jexl.JexlNodeFactory;
 import datawave.query.jexl.JexlNodeFactory.ContainerType;
+import datawave.query.model.QueryModel;
```
ivakegg, line 11: OK, I am stumped in this visitor as to where any lineage problems were fixed.  Is this one purely refactoring?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RangeCoalescingVisitor.java
```diff
@@ -90,12 +50,12 @@ public Object visit(ASTAndNode node, Object data) {
             JexlNodes.ensureCapacity(andNode, node.jjtGetNumChildren());
             for (int i = 0; i < node.jjtGetNumChildren(); i++) {
                 JexlNode newChild = (JexlNode) node.jjtGetChild(i).jjtAccept(this, data);
-                
                 andNode.jjtAddChild(newChild, i);
                 newChild.jjtSetParent(andNode);
             }
         }
-        
+
```
ivakegg, line 57: This visitor looks good.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RegexFunctionVisitor.java
```diff
@@ -48,55 +50,59 @@ public Object visit(ASTFunctionNode node, Object data) {
...
-                JexlNode regexNode = buildRegexNode((ASTIdentifier) node0, functionMetadata.name(), arguments.get(1).image);
-                if (regexNode != null) {
-                    returnNode = regexNode;
-                }
-            } else {
-                JexlNode newParent = null;
-                if (functionMetadata.name().equals("excludeRegex")) {
-                    newParent = new ASTAndNode(ParserTreeConstants.JJTANDNODE);
+    
+        switch (functionMetadata.name()) {
```
ivakegg, line 54: much cleaner, thanks

ivakegg:
First I must explain what a (query property) "marker" node is.  A portion of a query (jexl subtree) can be marked as have some attribute by attaching an assignment node to that subtree.  So if I want to mark some node A with a marker N, then I replace A with ((N = true) && (A)).  These markers can be thought of as query execution hints that the planner puts in place to allow the execution to do certain things.

Second I must explain what "push down" means.  To "push down" some portion of a query (jexl subtree) is to use the field index instead of the global index, or pushing down farther to the query evaluation phase.  Query execution basically started with global index lookup, then field index lookup (if required), then query evaluation meaning bouncing the boolean logic against all of the values for a document.

Here is some description of these visitors:

- BoundedRangeDetectionVisitor
This is supposed to detect whether there are bounded ranges, however it appears to only be looking for bounds.  I am actually changing this visitor here: https://github.com/NationalSecurityAgency/datawave/pull/926
- FunctionIndexQueryExpansionVisitor
This will expand all of the functions within the query to include something that we can run against the index.  Basically every function node in the tree will be replaced with (function && (<some query)).  The index query created will be dependent on the query descriptor for the query in question.
- FunctionNormalizationRebuildingVisitor
This visitor will normalize the values within the arguments of a function node.  The reason this is separated out is that the function descriptor needs to be used to determine which arguments are actually values and for which field do they apply.  The field name is needed to lookup the data type and subsequently the normalization required.  Normalization of values is required for index lookup.
- IsNotNullIntentVisitor
if an ER (regular expression) node is of the form FIELD =~ '*', then replace with: FIELD != null
- ParallelIndexExpansion
This is used for expanding regex nodes into the set of discrete values.  It is also used for unfielded term expansion (FixUnfieldedTermsVisitor) by extension.  If a regex (or range) cannot be expanded into descrete values, then it is wrapped with an ExceededValueThresholdMarkerJexlNode marker.
- PruneLessSelectiveFieldsVisitor
Not used for now.  You can ignore this one for now until we start populating the metadata with more extensive metrics at which time this visitor may be changed extensively.
- PushdownLargeFieldedListsVisitor
This will take a very large list of values OR'ed together for one field into an ExceededOrThresholdMarkerJexlNode construct.  This essentially will allow an "Ivarator" (DatawaveFieldIndedListIteratorJexl) to be used instead of creating a separate source for each term (see QueryIterator).
- PushdownMissingIndexRangeNodesVisitor
This will push down terms as "evaluation only" (meaning the index is not used) when there is an entry in the metadata table that denotes that the field is not indexed for some time range.  This is an initial form of a solution for #825.
- PushFunctionsIntoExceededValueRanges
If we have an ExceededValueThresholdMarkerJexlNode on a range which had been produced as a result of the FunctionIndexQueryExpansionVisitor above, then the function is pushed into the marker node construct to allow the "Ivarator" to evaluation the function when scanning the field index.
- RangeConjunctionRebuildingVisitor
This will expand ranges into discrete values as the ParallelIndexExpansion visitor does for regex nodes.
- RangeExpansionThresholdRebuildingVisitor
This one is not used and can be ignored/dropped.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/AllTermsIndexedVisitor.java
lbschanno: Fixed.

lbschanno:
@ivakegg I have removed any refactoring not related to fixes. I will create separate PRs for that work as we discussed so that this PR is focused on fixes only.

lbschanno:
Closing this PR to break it up into individual PRs for each visitor.

PR #926, opened 2020-09-28
Change to mark bounded ranges in JEXL

ivakegg:
The initial symptom was that we simply drop bounded ranges for the same field when they are and'ed together.  To alleviate this, we determined that bounded ranges need to be explicitly marked in JEXL to parallel the explicit bounded range construct in LUCENE.  So the primary changes are:

1) Bounded ranges in JEXL now look like this :  ((BoundedRange = true) && (x < a && x > b))
2) We no longer coalesce bounds for the same field into bounded range constructs

Other changes that came with this are:

1) Reworked the getBoundedRanges in the JexlASTHelper to use a builder pattern to avoid the excessive number of variants.
2) JexlASTHelper getIdentifiers and getLiterals will avoid getting the identifier and literal from assignments
3) Updated the JexlNodeFactory.createExpression to only create one if needed.  This avoid many of the extra parens showing up.
4) A bounded range that expands to nothing is now replaced with a FALSE literal which is now handled appropriately by the QueryPruningVisitor which was just enabled by default in 2.9.10.
5) An UnmarkedBoundedRangeDetectionVisitor was added to fail if marked bounded ranges are found that are not valid ranges, and to find bounded ranges that are unmarked.  The later was primarily to find any test cases that needed to mark ranges.  It is debatable whether that part should be left.
6) A flag was added to the configuration called "generatePlanOnly" which is used when requesting a plan.  Previously the range stream was being created even when values or fields were not being expanded which now that bounded ranges are being processed correctly would result in plans requiring full table scans.  This flag allows the getPlan endpoint to avoid that.


jwomeara:
If you can get this updated, I'll review it.

dmgall:
Your filling my inbox ☺

Miss seeing everyone!
Deeanna

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/BoundedRangeDetectionVisitor.java
```diff
@@ -28,29 +38,45 @@ public static boolean mustExpandBoundedRange(ShardQueryConfiguration config, Met
...
-        return false;
-        
     }
     
     @Override
-    public Object visit(ASTNRNode node, Object data) {
-        if (null != data) {
-            AtomicBoolean hasBounded = (AtomicBoolean) data;
-            hasBounded.set(true);
+    public Object visit(ASTERNode node, Object data) {
```
FineAndDandy, line 62: shouldn't these be LT/LTE and GT/GTE in order to detect a bounded range?

ivakegg: All I did here was replace the visit(ASTLTNode) and visit(ASTGTNode) with the detection of the BoundedRange marker in visit(ASTReference).  If you look where this thing is used, it is also looking for ER and NR nodes.  I do concur that the initial implementation was missing LE and GE node checks.  By changing this to look for BoundedRange markers, I essentially have made that change.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
```diff
@@ -356,4 +290,20 @@ protected JexlNode expandNodeForNormalizers(JexlNode node, Object data) {
...
         return nodeToReturn;
     }
+    
+    /**
+     * method to return if the current node is an instance of a delayed predicate
+     *
+     * @param currNode
+     * @return
+     */
+    protected boolean isDelayedPredicate(JexlNode currNode) {
```
apmoriarty, line 300: Looks like there are four different instances of a method with this name. Is there a need to maintain different copies? If so, what is unique about this instance?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPruningVisitor.java
```diff
@@ -253,6 +253,11 @@ public Object visit(ASTOrNode node, Object data) {
     
     @Override
     public Object visit(ASTAndNode node, Object data) {
```
apmoriarty, line 255: Looks like this import and commented code can get removed.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RangeConjunctionRebuildingVisitor.java
```diff
@@ -117,335 +111,43 @@ public Object visit(ASTReference node, Object data) {
...
-        for (JexlNode child : children(node)) {
-            if (ParserTreeConstants.JJTANDNODE == id(child)) {
-                collapseAndSubtrees((ASTAndNode) child, subTrees);
-            } else {
-                subTrees.add(child);
-            }
+        try {
+            fieldsToTerms = lookup.lookup(config, scannerFactory, config.getMaxIndexScanTimeMillis());
+        } catch (IllegalRangeArgumentException e) {
+            log.info("Cannot expand "
```
apmoriarty, line 141: Should this be log.error?

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -670,6 +676,11 @@ protected ASTJexlScript updateQueryTree(ScannerFactory scannerFactory, MetadataH
         
         stopwatch.stop();
         
+        // TEMPORARY?: CHECK FOR UNMARKED BOUNDED RANGES
```
apmoriarty, line 679: Is it temporary or not?

warehouse/query-core/src/test/java/datawave/query/DelayedIndexOnlyQueryTest.java
```diff
@@ -7,6 +7,7 @@
 import datawave.query.testframework.DataTypeHadoopConfig;
 import datawave.query.testframework.FieldConfig;
 import datawave.query.testframework.GenericCityFields;
+import org.apache.log4j.Level;
```
apmoriarty, line 10: stray import

warehouse/query-core/src/test/java/datawave/query/GroupedFlattenQueryTest.java
```diff
@@ -114,13 +114,26 @@ public void testCounty() throws Exception {
...
+    public void testFoundedRangeUnbounded() throws Exception {
         log.info("------  testFoundedRange  ------");
         String start = "1850";
         String end = "1860";
         String city = "'AuStiN'";
-        String query = GroupedField.CITY.name() + EQ_OP + city + AND_OP + "(" + GroupedField.FOUNDED.name() + GT_OP + start + AND_OP
-                        + GroupedField.FOUNDED.name() + LT_OP + end + ")";
+        String query = GroupedField.CITY.name() + EQ_OP + city + AND_OP + GroupedField.FOUNDED.name() + GT_OP + start + AND_OP + GroupedField.FOUNDED.name()
+                        + LT_OP + end;
+        // all entries have at least one founded less than end and one founded greater than start, just not the same value.
```
apmoriarty, line 124: Founded is such a fun word.

apmoriarty (changes requested):
Just a few questions

FineAndDandy (changes requested):

drewfarris:
@ivakegg when you have a chance, plz retarget to 3.1.x

apmoriarty (approved):
LGTM

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
ivakegg: removed

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RangeConjunctionRebuildingVisitor.java
ivakegg: sure, why not

warehouse/query-core/src/test/java/datawave/query/DelayedIndexOnlyQueryTest.java
ivakegg: making not temporary

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownLargeFieldedListsVisitor.java
```diff
@@ -216,15 +218,10 @@ private void copyChildren(Collection<JexlNode> children, Collection<JexlNode> co
...
-        } else if ((node.jjtGetNumChildren() == 1) && (node instanceof ASTReferenceExpression || node instanceof ASTReference || node instanceof ASTAndNode)) {
-            return rangeNodeToRange(node.jjtGetChild(0));
-        } else if ((node.jjtGetNumChildren() == 2) && node instanceof ASTAndNode) {
-            JexlNode leftChild = node.jjtGetChild(0);
-            JexlNode rightChild = node.jjtGetChild(1);
-            return new Range(new Key(String.valueOf(JexlASTHelper.getLiteralValue(leftChild))), leftChild instanceof ASTGENode, new Key(
-                            String.valueOf(JexlASTHelper.getLiteralValue(rightChild))), rightChild instanceof ASTLENode);
+        LiteralRange range = JexlASTHelper.findRange().getRange(node);
+        if (range != null) {
+            return new Range(new Key(String.valueOf(range.getLower())), range.isLowerInclusive(), new Key(String.valueOf(range.getUpper())),
```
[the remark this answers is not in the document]
ivakegg: I don't get this comment.  LiteralRange does not include Key/

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPruningVisitor.java
```diff
@@ -253,6 +253,11 @@ public Object visit(ASTOrNode node, Object data) {
...
+        // do not process query property markers
+        // if (QueryPropertyMarker.instanceOf(node, null)) {
+        // return TruthState.UNKNOWN;
+        // }
+
```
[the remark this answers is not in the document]
ivakegg: good catch.  I think I was debugging something.  I will remove or reenable as appropriate.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RangeConjunctionRebuildingVisitor.java
```diff
@@ -117,335 +111,43 @@ public Object visit(ASTReference node, Object data) {
...
-            
-            JexlNode orNode = JexlNodeFactory.createNodeTreeFromFieldsToValues(JexlNodeFactory.ContainerType.OR_NODE, new ASTEQNode(
-                            ParserTreeConstants.JJTEQNODE), onlyRangeNodes, fieldsToTerms, expandFields, expandValues, false);
-            
-            // Set the parent and child pointers accordingly
-            orNode.jjtSetParent(newNode);
-            newNode.jjtAddChild(orNode, index++);
-            
-        }
+        IndexLookup lookup = ShardIndexQueryTableStaticMethods.expandRange(range);
```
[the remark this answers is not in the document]
ivakegg: nope

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/UnmarkedBoundedRangeDetectionVisitor.java
```diff
@@ -0,0 +1,45 @@
...
+        if (BoundedRange.instanceOf(node)) {
+            if (!JexlASTHelper.findRange().isRange(node)) {
+                if (null != data) {
+                    AtomicBoolean hasBounded = (AtomicBoolean) data;
+                    hasBounded.set(true);
+                }
+            }
+            
+            return false;
+        } else if (JexlASTHelper.findRange().notDelayed().notMarked().isRange(node)) {
```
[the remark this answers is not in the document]
ivakegg: adding comments

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -793,492 +798,364 @@ private static void getERNodes(JexlNode node, List<ASTERNode> erNodes) {
...
-            } else if (literal instanceof BigInteger) {
-                range = getBigIntegerBoundedRange(rangeNodes, thisRangesNodes, new LiteralRange<>(fieldName, LiteralRange.NodeOperand.AND));
-            } else if (literal instanceof Float) {
-                range = getFloatBoundedRange(rangeNodes, thisRangesNodes, new LiteralRange<>(fieldName, LiteralRange.NodeOperand.AND));
-            } else if (literal instanceof Double) {
-                range = getDoubleBoundedRange(rangeNodes, thisRangesNodes, new LiteralRange<>(fieldName, LiteralRange.NodeOperand.AND));
-            } else if (literal instanceof BigDecimal) {
-                range = getBigDecimalBoundedRange(rangeNodes, thisRangesNodes, new LiteralRange<>(fieldName, LiteralRange.NodeOperand.AND));
+            JexlNode[] children = new JexlNode[] {child1, child2};
+            if (literal1 instanceof String || literal2 instanceof String) {
```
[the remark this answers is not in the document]
ivakegg: nah, I like literal1 and literal2

warehouse/query-core/src/test/java/datawave/query/RangeQueryTest.java
```diff
@@ -131,26 +139,28 @@ public void testSingleValueAndMultiFieldWithParens() throws Exception {
         }
     }
     
-    @Test
+    @Test(expected = FullTableScansDisallowedException.class)
```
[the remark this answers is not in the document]
ivakegg: Previously this would pull together the bounds as a bounded range.  Now we required ranges to be bounded.  Hence this results only non-executable separate bounds and hence the query requires a full table scan required.

warehouse/query-core/src/test/java/datawave/query/jexl/JexlASTHelperTest.java
```diff
@@ -247,38 +249,81 @@ public void testApplyNormalization() throws Throwable {
...
-        ranges = JexlASTHelper.getBoundedRangesIndexAgnostic((ASTAndNode) (script.jjtGetChild(0)), nonRangeChildNodes, true);
+        range = JexlASTHelper.findRange().getRange(script.jjtGetChild(0));
         
-        Assert.assertEquals(1, ranges.size());
-        Assert.assertEquals(2, ranges.values().iterator().next().size());
-        Assert.assertEquals(1, nonRangeChildNodes.size());
-        Assert.assertEquals(ASTNotNode.class, nonRangeChildNodes.get(0).getClass());
+        Assert.assertNotNull(range);
+        Assert.assertNotNull(range.getLowerNode());
+        Assert.assertNotNull(range.getUpperNode());
```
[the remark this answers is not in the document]
ivakegg: good idea

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownMissingIndexRangeNodesVisitor.java
```diff
@@ -72,78 +74,43 @@ public PushdownMissingIndexRangeNodesVisitor(ShardQueryConfiguration config, Met
...
         }
         
-        return newNode;
     }
     
     @Override
     public Object visit(ASTReferenceExpression node, Object data) {
         // if not already delayed somehow
-        if (!QueryPropertyMarker.instanceOf(node, null)) {
+        if (!QueryPropertyMarkerVisitor.instanceOfAnyExcept(node, Collections.singletonList(BoundedRange.class))) {
```
jwomeara, line 113: Previously, QueryPropertyMarkerVisitor.instanceOf(node, null) implied that we were searching for any delayed nodes.  Now that we have BoundedRange as an additional marker, it implies something different as that isn't delayed.  Are we sure that there aren't any other places in the code where we don't need to do a similar update to what you've done here to ensure that BoundedRange isn't treated as a delayed node?  Should we create some kind of isInstanceOfDelayed method to make this clearer?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/UnmarkedBoundedRangeDetectionVisitor.java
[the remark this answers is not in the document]
jwomeara: If I understand correctly, this case indicates that we have a bounded range, but it's not marked, and not delayed, whereas the previous case will find a bounded range that's already been marked.

warehouse/query-core/src/test/java/datawave/query/jexl/functions/GeoFunctionsDescriptorTest.java
```diff
@@ -28,7 +29,7 @@ public void antiMeridianTest2() throws Exception {
         JexlArgumentDescriptor argDesc = new GeoFunctionsDescriptor().getArgumentDescriptor((ASTFunctionNode) node.jjtGetChild(0).jjtGetChild(0));
         JexlNode queryNode = argDesc.getIndexQuery(null, null, null, null);
         Assert.assertEquals(
-                        "(((LON_FIELD >= '170.0' && LON_FIELD <= '180') && (LAT_FIELD >= '40.0' && LAT_FIELD <= '50.0')) && ((LON_FIELD >= '-180' && LON_FIELD <= '-170.0') && (LAT_FIELD >= '40.0' && LAT_FIELD <= '50.0')))",
+                        "((((BoundedRange = true) && (LON_FIELD >= '170.0' && LON_FIELD <= '180')) && ((BoundedRange = true) && (LAT_FIELD >= '40.0' && LAT_FIELD <= '50.0'))) && (((BoundedRange = true) && (LON_FIELD >= '-180' && LON_FIELD <= '-170.0')) && ((BoundedRange = true) && (LAT_FIELD >= '40.0' && LAT_FIELD <= '50.0'))))",
```
jwomeara, line 32: This is going to blow up our query plans for geo queries.  Gulp...

jwomeara (changes requested):

warehouse/query-core/src/main/java/datawave/query/composite/CompositeRange.java
```diff
@@ -305,9 +308,9 @@ public boolean isValid() {
     
     public boolean contains(JexlNode node) {
         boolean success = true;
-        if (node instanceof ASTAndNode)
-            for (int i = 0; i < node.jjtGetNumChildren(); i++)
-                success &= this.jexlNodeListLowerBound.contains(node.jjtGetChild(i)) || this.jexlNodeListUpperBound.contains(node.jjtGetChild(i));
+        LiteralRange range = JexlASTHelper.findRange().getRange(node);
+        if (range != null)
+            success &= this.jexlNodeListLowerBound.contains(range.getLowerNode()) && this.jexlNodeListUpperBound.contains(range.getUpperNode());
```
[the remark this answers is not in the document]
ivakegg: ok

warehouse/query-core/src/test/java/datawave/query/jexl/functions/GeoFunctionsDescriptorTest.java
ivakegg: I think you create a ticket for that.....

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownMissingIndexRangeNodesVisitor.java
ivakegg: Agreed.  I did validate all calls as such.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownLargeFieldedListsVisitor.java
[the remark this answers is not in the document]
ivakegg: String.valueOf will not create a new string if the arg is already a string.

PR #932, opened 2020-10-07
fixes #927 Always move the next highest key in AndIterator

### LINKED ISSUE #927: Improve AndIterator intersection algorithm
Intersection currently will always seek the lowest key to the highest, instead move the next highest to the highest. The next highest is more likely to be a lower cardinality term which will skip over more terms when the lowest is finally moved. Also when moving an AndIterator adapt the move to key based on the highest seek of all iterators being moved, bypassing extra seeks that are known to not be matches.

FineAndDandy:


apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/iterator/logic/AndIterator.java
```diff
@@ -350,11 +354,46 @@ public T move(T minimum) {
...
+        while (highest != null && !includeHeads.isEmpty()) {
+            // if include heads is higher than the target key, adjust the target key to the highest since an intersection is not possible at anything but the
+            // highest
+            T highestInternal = includeHeads.keySet().last();
+            if (highestInternal.compareTo(to) > 0) {
+                to = highestInternal;
+            }
+            
+            // advance the iterators
+            includeHeads = moveIterators(highest, to);
```
ivakegg, line 383: Are we risking a stack overflow here?

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/iterator/logic/AndIterator.java
FineAndDandy: this call is not on the set, but on a single value. This is not recursive.

ivakegg (approved):

FineAndDandy:
closing this in favor of a later more stable release

PR #940, opened 2020-10-13
Fixed Decor.

ghost:


keith-ratcliffe:
Thanks for the PR. Unfortunately, I'm going to have to close it, because it appears the only purpose is to add whitespace, which is not something we'd typically merge, unless it was to fix a failing "check-code-formatting" test. These changes introduce new breakage for that particular test

PR #942, opened 2020-10-15
Fixed!  Added a flatten step to EdgeQueryLogic.configureRanges to showcase te…

jwomeara:
…st errors in ExtendedEdgeQueryLogicTest and EdgeQueryFunctionalTest.  The test failures occur because EdgeTableRangeBuildingVisitor cannot handle a flattened tree.

This is broke... as a joke!

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EdgeTableRangeBuildingVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EdgeTableRangeBuildingVisitor.java
@@ -232,6 +232,7 @@ public class EdgeTableRangeBuildingVisitor extends BaseVisitor implements EdgeMo
                         }
                     }
                     mergedContext = childContext;
+                    continue;
                 } else if (((List<QueryContext>) mergedContext).get(0).getRowContext() != null) {
                     for (QueryContext qContext : ((List<QueryContext>) mergedContext)) {
                         if (!(qContext.combineQueryContexts(((List<QueryContext>) childContext), false))) {
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EdgeTableRangeBuildingVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EdgeTableRangeBuildingVisitor.java
@@ -20,6 +20,7 @@ import org.apache.commons.jexl2.parser.ASTNRNode;
 import org.apache.commons.jexl2.parser.ASTOrNode;
 import org.apache.commons.jexl2.parser.ASTReference;
 import org.apache.commons.jexl2.parser.ASTReferenceExpression;
+import org.apache.commons.jexl2.parser.JexlNode;
 import org.apache.commons.jexl2.parser.SimpleNode;
 import org.apache.log4j.Logger;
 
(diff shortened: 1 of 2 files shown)


jwomeara:
Spent a little time looking at this tonight.  I think I have it fixed now.  Let me know what you think.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EdgeTableRangeBuildingVisitor.java
```diff
@@ -168,75 +171,86 @@ public Object visit(ASTAndNode node, Object data) {
...
+                    mergedContext = childContext;
+                    continue;
+                } else if (((List<QueryContext>) mergedContext).get(0).getRowContext() != null) {
+                    for (QueryContext qContext : ((List<QueryContext>) mergedContext)) {
+                        if (!(qContext.combineQueryContexts(((List<QueryContext>) childContext), false))) {
+                            log.error("And node had unexpected return type");
+                            throw new IllegalArgumentException("Error: problem with query syntax");
+                        }
                     }
                 }
```
ivakegg, line 243: need a continue here

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EdgeTableRangeBuildingVisitor.java
```diff
@@ -168,75 +171,86 @@ public Object visit(ASTAndNode node, Object data) {
...
-                    }
+            if ((childContext.get(0) instanceof IdentityContext) && (mergedContext.get(0) instanceof IdentityContext)) {
+                QueryContext qContext = new QueryContext();
+                
+                qContext.packageIdentities((List<IdentityContext>) childContext);
+                qContext.packageIdentities((List<IdentityContext>) mergedContext);
+                
+                ArrayList<QueryContext> aList = new ArrayList<>();
+                aList.add(qContext);
+                mergedContext = aList;
```
ivakegg, line 200: how about set "processed = true" as well and then test it at the error message below.

jwomeara: Check out my latest commit.  I think that by putting the thrown exception in an else statement we fix the issue and preserve the original functionality.

ivakegg (approved):

drewfarris (commented):
The error message are the only potential issue that stuck out to me, but I'm not as familiar with this code as I could be.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EdgeTableRangeBuildingVisitor.java
```diff
@@ -168,75 +171,86 @@ public Object visit(ASTAndNode node, Object data) {
...
                     }
+                } else {
+                    log.error("Problem parsing query");
+                    throw new IllegalArgumentException("Error: problem with query syntax");
                 }
-                return contexts2;
+            } else {
+                
+                log.error("And node had unexpected return type");
+                throw new IllegalArgumentException("Error: problem with query syntax");
```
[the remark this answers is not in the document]
ivakegg: This is what was there before.....

PR #945, opened 2020-10-16
fixes #943: Updated to keep the original _ANYFIELD_ node within negations

### LINKED ISSUE #943: Unfielded negations should remain unfielded along with any expansions
#572 was implemented to allow us to disable expanding unfielded terms within a negation.  Enabling this feature can still cause issues in terms of matches against index-only terms or in the case of a TLD query where the terms only exist in child documents.  Regardless whether this feature is enabled, we need to leave the unfielded term within the negation along with any expansions.

ivakegg:


jwomeara:
lgtm

jwomeara (approved):

hlgp (approved):

PR #948, opened 2020-10-19
fixes #946 Check nullHeadMap for entries that match the current context

### LINKED ISSUE #946: NestedIteratorContextUtil.applyContext() should check nullHeadMap
Not checking for null sources out of the `processmoves()` doesn't catch sources that were already previously evaluated null at the current context.

FineAndDandy:


warehouse/query-core/src/main/java/datawave/query/iterator/logic/NestedIteratorContextUtil.java
```diff
@@ -174,6 +174,9 @@
             headMapKey = headMap.isEmpty() ? null : headMap.keySet().last();
         }
         
+        // grab any exist null heads that match the context (not requiring a move due to already being there)
```
apmoriarty, line 177: existing

apmoriarty (approved):

ivakegg (approved):

PR #951, opened 2020-10-19
Addresses #950 ingest job percent error threshold

hlgp:


warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
```diff
@@ -423,20 +424,25 @@ public int run(String[] args) throws Exception {
         // output the counters to the log
         Counters counters = job.getCounters();
         log.info(counters);
+        JobClient jobClient = new JobClient((org.apache.hadoop.mapred.JobConf) job.getConfiguration());
```
ivakegg, line 427: We still need to close this jobClient which is what the try block was doing for us previously.

hlgp: true!

ivakegg (approved):

drewfarris (approved):

PR #954, opened 2020-10-20
Fix lineage for AllTermsIndexedVisitor

lbschanno:
AllTermsIndexedVisitor did not return a query tree with a valid lineage.
Ensure that lineage is properly established and add an assert for this
in related tests.

Related to #880


ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/AllTermsIndexedVisitor.java
```diff
@@ -63,111 +73,56 @@ public AllTermsIndexedVisitor(ShardQueryConfiguration config, MetadataHelper hel
...
+                        .map(n -> (JexlNode) n.jjtAccept(this, data)) // Visit the node.
+                        .filter(Objects::nonNull) // Remove null nodes.
+                        .filter(n -> (!(n instanceof ASTAndNode) && !(n instanceof ASTOrNode)) || JexlNodes.isNotChildless(n)) // Remove empty junction nodes.
+                        .toArray(JexlNode[]::new);
+        // @formatter:on
+        JexlNode copy = JexlNodes.newInstanceOfType(node);
+        copy.image = node.image;
+        JexlNodes.children(copy, children);
+        return copy;
+    }
```
jwomeara, line 124: This method is essentially the same as the copy method in the base RebuildingVisitor.  The exception to this is the filtering step where it drops AND or OR node without any children.  That *shouldn't* happen, and if it is happening that points to problems elsewhere in the code base.  So, if we assume that we won't encounter an AND or OR node without any children (which I think is a fairly safe bet) then we could just use the RebuildingVisitor's copy method.  Even if we *DO* encounter an AND or OR node with 0 children, the switch statement in the and/or node visitors will handle that by returning null, which gets dropped by the rebuilding visitor's copy method.  

Sorry for the wall of text... I think that using the visit methods of the parent RebuildingVisitor is a little bit cleaner and simplifies the code a bit.  So, in each place where you call 'getCopyWithVisitedChildren' you can instead call super.visit(node, data) which calls RebuildingVisitor.copy, which visits the entire subtree and creates copies.

jwomeara (changes requested):
1 comment, otherwise LGTM!

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/AllTermsIndexedVisitor.java
lbschanno: Fixed.

ivakegg (dismissed):

PR #955, opened 2020-10-20
Fix lineage for BooleanOptimizationRebuildingVisitor

lbschanno:
BooleanOptimizationRebuildingVisitor does not return a query tree with a
valid lineage. Ensure that lineage is properly established and an assert
for this in related tests.

Related to #880.


ivakegg (approved):

jwomeara (approved):
LGTM!

PR #958, opened 2020-10-20
Verify lineage for ExpandMultinormalizedTerms

lbschanno:
Add an assert to tests for ExpandMultinormalizedTerms to verify that it
returns a query tree with valid lineage.

Part of #880.


ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
```diff
@@ -155,12 +155,9 @@ public Object visit(ASTReference node, Object data) {
...
      */
     protected boolean isDelayedPredicate(JexlNode currNode) {
-        if (ASTDelayedPredicate.instanceOf(currNode) || ExceededOrThresholdMarkerJexlNode.instanceOf(currNode)
+        return ASTDelayedPredicate.instanceOf(currNode) || ExceededOrThresholdMarkerJexlNode.instanceOf(currNode)
                         || ExceededValueThresholdMarkerJexlNode.instanceOf(currNode) || ExceededTermThresholdMarkerJexlNode.instanceOf(currNode)
-                        || IndexHoleMarkerJexlNode.instanceOf(currNode) || ASTEvaluationOnly.instanceOf(currNode))
-            return true;
-        else
-            return false;
+                        || IndexHoleMarkerJexlNode.instanceOf(currNode) || ASTEvaluationOnly.instanceOf(currNode);
```
jwomeara, line 160:
```
        return QueryPropertyMarkerVisitor.instanceOf(currNode, Arrays.asList(ASTDelayedPredicate.class, ExceededOrThresholdMarkerJexlNode.class,
                        ExceededValueThresholdMarkerJexlNode.class, ExceededTermThresholdMarkerJexlNode.class, IndexHoleMarkerJexlNode.class,
                        ASTEvaluationOnly.class), null);
```

This method is a dupe of (erroneously protected) JexlASTHelper.isDelayedPredicate.  We should rip this out, update the JexlASTHelper method to be public, and use that instead. 

Also, instead of visiting the tree six times for each instanceof call, we should pass the list of types to the QueryPropertyMarkerVisitor, and run instanceof once (as shown above);

jwomeara (changes requested):
1 comment, but otherwise LGTM!

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
lbschanno: I've deduped the method and used QueryPropertyMarkerVisitor. One change I made was to use QueryPropertyMarkerVisitor.instanceOfAny(node) instead, since that will default to the TYPE_IDENTIFIERS set that already contains the delayed predicate instance types.

jwomeara: Good call.  I didn't realize that there were only 6 marker nodes.  D'oh!  Nice work!

jwomeara (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -1204,13 +1201,15 @@ private static void getERNodes(JexlNode node, List<ASTERNode> erNodes) {
...
-            return false;
+    /**
+     * Return whether to not the provided node is an instance of a delayed predicate.
+     * 
+     * @param node
+     *            the node
+     * @return true if the node is an instance of a delayed predicate, or false otherwise
+     */
+    public static boolean isDelayedPredicate(JexlNode node) {
+        return QueryPropertyMarkerVisitor.instanceOfAny(node);
```
ivakegg, line 1212: I don't believe this should be changed.  isDelayedPredictate is specific to a certain set of query property markers.  BoundedRange is coming along in #926 which is specifically not a delayed predicate.  Please revert this to it original state.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
lbschanno: Fixed.

jwomeara: Wait wait wait!  I still think we should avoid multiple recursions of the entire jexl tree.  Can we revert the revert, but update the isDelayedPredicate call to be...

```
return QueryPropertyMarkerVisitor.instanceOf(currNode, Arrays.asList(ASTDelayedPredicate.class, ExceededOrThresholdMarkerJexlNode.class,
                        ExceededValueThresholdMarkerJexlNode.class, ExceededTermThresholdMarkerJexlNode.class, IndexHoleMarkerJexlNode.class,
                        ASTEvaluationOnly.class), null);
```

I think this satisfies both my and Ivan's issue, right?

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -1204,13 +1204,10 @@ private static void getERNodes(JexlNode node, List<ASTERNode> erNodes) {
...
-    protected static boolean isDelayedPredicate(JexlNode currNode) {
-        if (ASTDelayedPredicate.instanceOf(currNode) || ExceededOrThresholdMarkerJexlNode.instanceOf(currNode)
+    public static boolean isDelayedPredicate(JexlNode currNode) {
+        return ASTDelayedPredicate.instanceOf(currNode) || ExceededOrThresholdMarkerJexlNode.instanceOf(currNode)
                         || ExceededValueThresholdMarkerJexlNode.instanceOf(currNode) || ExceededTermThresholdMarkerJexlNode.instanceOf(currNode)
-                        || IndexHoleMarkerJexlNode.instanceOf(currNode) || ASTEvaluationOnly.instanceOf(currNode))
-            return true;
-        else
-            return false;
+                        || IndexHoleMarkerJexlNode.instanceOf(currNode) || ASTEvaluationOnly.instanceOf(currNode);
```
jwomeara, line 1210: If we do it this way, we only traverse the tree once:
```
return QueryPropertyMarkerVisitor.instanceOf(currNode, Arrays.asList(ASTDelayedPredicate.class, ExceededOrThresholdMarkerJexlNode.class,
                        ExceededValueThresholdMarkerJexlNode.class, ExceededTermThresholdMarkerJexlNode.class, IndexHoleMarkerJexlNode.class,
                        ASTEvaluationOnly.class), null);
```

jwomeara (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
lbschanno: Third time's the charm? Fixed.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
lbschanno: Fixed in latest commit.

lbschanno:
Per a request from @ivakegg earlier today, I've moved the isDelayedPredicate() method to QueryPropertyMarkerVisitor.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -1204,14 +1206,9 @@ private static void getERNodes(JexlNode node, List<ASTERNode> erNodes) {
...
     
-    protected static boolean isDelayedPredicate(JexlNode currNode) {
-        if (ASTDelayedPredicate.instanceOf(currNode) || ExceededOrThresholdMarkerJexlNode.instanceOf(currNode)
-                        || ExceededValueThresholdMarkerJexlNode.instanceOf(currNode) || ExceededTermThresholdMarkerJexlNode.instanceOf(currNode)
-                        || IndexHoleMarkerJexlNode.instanceOf(currNode) || ASTEvaluationOnly.instanceOf(currNode))
-            return true;
-        else
-            return false;
-    }
+    private static final List<Class<? extends QueryPropertyMarker>> DELAYED_PREDICATE_TYPES = Collections.unmodifiableList(Arrays.asList(
```
ivakegg, line 1209: This constant is not needed?

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
lbschanno: Good catch, I forgot to remove this. Fixed.

PR #959, opened 2020-10-20
Verify lineage for TreeFlatteningRebuildingVisitor

lbschanno:
Add an assert to tests for TreeFlatteningRebuildingVisitor to verify
that it returns a query with valid lineage.

Part of #880.


ivakegg (approved):

jwomeara (changes requested):
Aside from my minor nit this LGTM!

lbschanno:
@jwomeara could you describe what you would like changed for this PR?

jwomeara:
Sorry, it looks like my comment didn't take :/  
I'll add the comment to the line in question.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/TreeFlatteningRebuilder.java
```diff
@@ -150,9 +148,8 @@ else if (hasChildren && node == parentStack.peek()) {
      *            the node to be copied
      * @param postOrderDeque
      *            the post order traversal of the copied tree
-     * @return the copied tree
      */
-    private JexlNode copyTree(JexlNode node, Deque<JexlNode> postOrderDeque) {
+    private void copyTree(JexlNode node, Deque<JexlNode> postOrderDeque) {
```
jwomeara, line 152: If we're going to remove the return for the method, then we should probably rename it as it no longer returns a copied tree.  When I originally wrote this I thought it would be useful to have an iterative method for copying a Jexl tree.  So, that's why it returns the copied tree even though we don't actually use it anywhere - yet!  I could see an iterative copy method being useful in the future though, so my thought would be to leave this method as-is and return the copied tree.  Just my 2 cents.

lbschanno: @jwomeara thanks for clarifying. I've restored the return value for this method.

ivakegg (approved):

jwomeara (approved):

PR #962, opened 2020-10-20
Verify lineage for QueryModelVisitor

lbschanno:
Add asserts to tests for QueryModelVisitor to verify that it returns a
query tree for valid lineage.

Part of #880.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryModelVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryModelVisitor.java
@@ -2,15 +2,14 @@ package datawave.query.jexl.visitors;
 
 import com.google.common.base.Function;
 import com.google.common.collect.ArrayListMultimap;
-import com.google.common.collect.FluentIterable;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Sets;
-import datawave.query.model.QueryModel;
 import datawave.query.exceptions.DatawaveFatalQueryException;
 import datawave.query.jexl.JexlASTHelper;
 import datawave.query.jexl.JexlNodeFactory;
 import datawave.query.jexl.JexlNodeFactory.ContainerType;
+import datawave.query.model.QueryModel;
 import datawave.webservice.common.logging.ThreadConfigurableLogger;
 import datawave.webservice.query.exception.DatawaveErrorCode;
 import datawave.webservice.query.exception.QueryException;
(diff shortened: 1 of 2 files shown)


ivakegg (approved):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryModelVisitor.java
```diff
@@ -395,20 +388,18 @@ protected JexlNode expandBinaryNodeFromModel(JexlNode node, Object data) {
          * need to ensure that if we create a logical structure ( such as an or ) -- each literal references a unique identifier from the right. Otherwise,
          * subsequent visitors will reference incorrection sub trees, and potentially negate the activity of the query model visitor
          */
-        Set<List<Object>> newSet = Sets.newHashSet(FluentIterable.from(product).transform(new ProductTransformer()));
+        Set<List<Object>> newSet = product.stream().map(new ProductTransformer()::apply).collect(Collectors.toSet());
```
jwomeara, line 391:
```suggestion
        Set<List<JexlNode>> newSet = product.stream().map(list -> list.stream().map(RebuildingVisitor::copy).collect(Collectors.toList()))
                        .collect(Collectors.toSet());
```

I like this change, but it had me wondering if we actually still needed the ProductTransformer.  After playing around with this a little, I settled on this implementation (which causes a ripple-effect of changes in a few places).

The transformer handled the case where one of the elements in the list is NOT a JexlNode.  However, after looking at all the ways we populate the 'left' and 'right' sets, we are only ever populating them with JexlNodes, so instead of typing everything as 'object', I typed everything as JexlNode, which simplified the ProductTransformer logic down to straight 'copy' on each sublist.  That's what is reflected here.  

This change led to changes method signature changes in JexlNodeFactory, namely:
1) createNodeTreeFromPairs - the pairs argument changed to Set<List<JexlNode>> pairs
2) buildUntypedBinaryNode - the left and right arguments changed to JexlNode

Both of those methods accept simple objects as inputs, but they don't actually work for anything that doesn't extend JexlNode, so it makes sense to me to be more specific with our typing.  I know that this isn't your code, and it may be a little outside the scope of this PR, but I think this would be a good change if you're up for it.

jwomeara (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryModelVisitor.java
lbschanno: @jwomeara I made the changes necessary to remove ProductTransformer.

jwomeara (approved):
Nice work!

PR #963, opened 2020-10-20
Fix lineage for RangeCoalescingVisitor

lbschanno:
RangeCoalescingVisitor does not return a query tree with a valid
lineage. Ensure that lineage is properly established, and modify unit
tests to assert this.

Part of #880.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RangeCoalescingVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RangeCoalescingVisitor.java
@@ -1,31 +1,21 @@
 package datawave.query.jexl.visitors;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-
 import datawave.query.jexl.JexlASTHelper;
 import datawave.query.jexl.JexlNodeFactory;
 import datawave.query.jexl.LiteralRange;
 import datawave.webservice.common.logging.ThreadConfigurableLogger;
-
 import org.apache.commons.jexl2.parser.ASTAndNode;
-import org.apache.commons.jexl2.parser.ASTERNode;
-import org.apache.commons.jexl2.parser.ASTGENode;
-import org.apache.commons.jexl2.parser.ASTGTNode;
-import org.apache.commons.jexl2.parser.ASTLENode;
-import org.apache.commons.jexl2.parser.ASTLTNode;
-import org.apache.commons.jexl2.parser.ASTNRNode;
 import org.apache.commons.jexl2.parser.JexlNode;
 import org.apache.commons.jexl2.parser.JexlNodes;
 import org.apache.commons.jexl2.parser.ParserTreeConstants;
 import org.apache.log4j.Logger;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
 /**
  * Visits an JexlNode tree, and coalesces bounded ranges into separate AND expressions.
- *
- * 
- *
  */
 public class RangeCoalescingVisitor extends RebuildingVisitor {
     private static final Logger log = ThreadConfigurableLogger.getLogger(RangeCoalescingVisitor.class);
@@ -41,36 +31,6 @@ public class RangeCoalescingVisitor extends RebuildingVisitor {
         RangeCoalescingVisitor visitor = new RangeCoalescingVisitor();
         
         return (T) script.jjtAccept(visitor, null);
-    }
-    
-    @Override
-    public Object visit(ASTERNode node, Object data) {
-        return node;
-    }
-    
-    @Override
-    public Object visit(ASTNRNode node, Object data) {
-        return node;
-    }
-    
-    @Override
-    public Object visit(ASTLTNode node, Object data) {
-        return node;
-    }
-    
-    @Override
-    public Object visit(ASTGTNode node, Object data) {
-        return node;
-    }
-    
-    @Override
-    public Object visit(ASTLENode node, Object data) {
-        return node;
-    }
-    
-    @Override
-    public Object visit(ASTGENode node, Object data) {
-        return node;
     }
     
     @Override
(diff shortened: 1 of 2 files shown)


ivakegg (approved):

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/RangeCoalescingVisitorTest.java
```diff
@@ -113,11 +85,48 @@ public void testCoalesceManyNestedAndUnorderedTermsWithAnchorTerm() throws Parse
...
+    private void assertVisitorResult(String original, String expected) throws ParseException {
+        ASTJexlScript originalScript = JexlASTHelper.parseJexlQuery(original);
+        ASTJexlScript expectedScript = JexlASTHelper.parseJexlQuery(expected);
+        ASTJexlScript actualScript = RangeCoalescingVisitor.coalesceRanges(originalScript);
         
-        String newQuery = JexlStringBuildingVisitor.buildQuery(newScript);
-        assertTrue("Expected [" + expectedQuery1 + "] or [" + expectedQuery2 + "] but was " + newQuery,
-                        (expectedQuery1.equals(newQuery) || expectedQuery2.equalsIgnoreCase(newQuery)));
+        assertScriptEquality(actualScript, expectedScript);
+        assertTrue(JexlASTHelper.validateLineage(actualScript, true));
```
jwomeara, line 97: This is 100% not required for this PR, but I would LOVE to see us validate the lineage of the originalScript after running the visitor as well.  In my opinion, a rebuilding visitor should _not_ be destructive.  That is to say that the originalScript that's passed in should not be modified in any way by any of our rebuilding visitors.  The changes should be evident in the rebuilt actualScript only.  

I will also add that this is definitely NOT the case with many of our rebuilding visitors.  They are certainly destructive (I know, because I've written a few that way, d'oh!).  I'm hoping that this is something that we can move away from.  If you're not feeling up to that for this PR, please do me a favor and make a separate issue so that we can track that as well.  

BTW, I loaded up your PR and added the validateLineage step against the originalScript...  This is definitely a destructive rebuilding visitor.  FYI.

jwomeara (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RangeCoalescingVisitor.java
```diff
@@ -90,12 +50,12 @@ public Object visit(ASTAndNode node, Object data) {
             JexlNodes.ensureCapacity(andNode, node.jjtGetNumChildren());
             for (int i = 0; i < node.jjtGetNumChildren(); i++) {
                 JexlNode newChild = (JexlNode) node.jjtGetChild(i).jjtAccept(this, data);
-                
                 andNode.jjtAddChild(newChild, i);
                 newChild.jjtSetParent(andNode);
             }
         }
         
+        JexlNodes.replaceChild(node.jjtGetParent(), node, andNode);
```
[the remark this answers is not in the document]
jwomeara: I tested this change locally, and tests still pass perfectly without this line.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RangeCoalescingVisitor.java
```diff
@@ -137,8 +99,9 @@ protected JexlNode coalesceBoundedRanges(Map<LiteralRange<?>,List<JexlNode>> ran
         
         // If we had no other nodes than this bounded range, we can strip out the original parent
         if (newNode.jjtGetNumChildren() == 1) {
-            newNode.jjtGetChild(0).jjtSetParent(newNode.jjtGetParent());
-            return newNode.jjtGetChild(0);
+            JexlNode child = newNode.jjtGetChild(0);
+            JexlNodes.replaceChild(newNode.jjtGetParent(), newNode, child);
```
jwomeara, line 103: This reparenting is also not needed as reparenting is handled by the base RebuildingVisitor.  

I tested this change locally as well, and tests still pass perfectly if you simply return newNode.jjtGetChild(0).

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RangeCoalescingVisitor.java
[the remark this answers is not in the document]
lbschanno: Fixed.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RangeCoalescingVisitor.java
lbschanno: Fixed.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/RangeCoalescingVisitorTest.java
lbschanno: I will create a separate ticket to encapsulate this request to make it easier to track the effort for it across the relevant visitors.

jwomeara: Thanks!

jwomeara (approved):

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/RangeCoalescingVisitorTest.java
jwomeara: I didn't see an issue yet, so I created one.

ivakegg (approved):

jwomeara (changes requested):
On line 43 of the RangeCoalescingVisitor, there is a statement which sets the parent of the rebuilt node to the original node's parent.  However, given that this is a rebuilding visitor, the original node's parent will end up being rebuilt, and parentage for this node will be reassigned when that happens, so line 43 just isn't necessary.  I tested this change with your unit test, and everything ran successfully.  I know that this wasn't part of your change, but it seemed related to me, so I figured while we're here we should change it.  Other than that, this looks great!

lbschanno:
@jwomeara I've removed the unnecessary line.

jwomeara (approved):

PR #969, opened 2020-10-22
ValidPatternVisitor now checks #INCLUDE and #EXCLUDE functions. (#787)

apmoriarty:
Backport fix for valid pattern visitor to 2.9, original feature was merged into master.

* ValidPatternVisitor now checks #INCLUDE and #EXCLUDE functions.

* Address PR comments.

* Address PR comments.

Co-authored-by: Brian Loss <brianloss@gmail.com>
Co-authored-by: Ivan Bella <ivan@bella.name>

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidPatternVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidPatternVisitor.java
@@ -1,16 +1,21 @@
 package datawave.query.jexl.visitors;
 
-import java.util.Map;
-import java.util.NoSuchElementException;
-import java.util.regex.Pattern;
-
-import datawave.query.jexl.JexlASTHelper;
-
-import org.apache.commons.jexl2.parser.ASTERNode;
-import org.apache.commons.jexl2.parser.ASTNRNode;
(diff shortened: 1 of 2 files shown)


ivakegg (approved):

FineAndDandy (approved):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidPatternVisitor.java
```diff
@@ -69,22 +59,64 @@ public Object visit(ASTERNode node, Object data) {
...
+    }
+    
+    /**
+     * Parse a literal value and put into the pattern cache if it does not exist.
+     *
+     * @param node
+     */
+    public void parseAndPutPattern(JexlNode node) {
+        // Catch the situation where a user might enter FIELD1 !~ VALUE1
+        Object literalValue = JexlASTHelper.getLiteralValue(node);
```
ivakegg, line 107: As it turns out, the initial implementation wraps the call to getLiteralValue with a try-catch block.  This was allowing us to handle FIELD1 =~ FIELD2 constructs correctly.  Now this visitor will simply throw an exception in that case which is not what we want.  I am fixing this in the master branch.

PR #978, opened 2020-10-29
fixes #976: Added a visitor to validate comparisons are between a field…

### LINKED ISSUE #976: Queries containing <literal> <op> <literal> should fail fast
If a query contains an expression of the form \<literal\> \<op\> \<literal\>, then they most certainly were intending something else (i.e. the field name got interpreted as a literal because of a typo).  This query will actually get through the planner currently but will finally fail on the tserver side.  We should fail fast with an appropriate error message.

ivakegg:
Added a visitor to ensure comparisons are between a field and a literal or between two fields.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidComparisonVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidComparisonVisitor.java
@@ -2,80 +2,209 @@ package datawave.query.jexl.visitors;
 
 import datawave.query.exceptions.DatawaveFatalQueryException;
 import datawave.query.jexl.JexlASTHelper;
+import org.apache.commons.jexl2.parser.ASTAdditiveNode;
+import org.apache.commons.jexl2.parser.ASTAssignment;
+import org.apache.commons.jexl2.parser.ASTDivNode;
 import org.apache.commons.jexl2.parser.ASTEQNode;
 import org.apache.commons.jexl2.parser.ASTERNode;
+import org.apache.commons.jexl2.parser.ASTFalseNode;
+import org.apache.commons.jexl2.parser.ASTFunctionNode;
 import org.apache.commons.jexl2.parser.ASTGENode;
 import org.apache.commons.jexl2.parser.ASTGTNode;
 import org.apache.commons.jexl2.parser.ASTLENode;
 import org.apache.commons.jexl2.parser.ASTLTNode;
+import org.apache.commons.jexl2.parser.ASTMethodNode;
+import org.apache.commons.jexl2.parser.ASTModNode;
+import org.apache.commons.jexl2.parser.ASTMulNode;
 import org.apache.commons.jexl2.parser.ASTNENode;
 import org.apache.commons.jexl2.parser.ASTNRNode;
+import org.apache.commons.jexl2.parser.ASTNullLiteral;
+import org.apache.commons.jexl2.parser.ASTNumberLiteral;
+import org.apache.commons.jexl2.parser.ASTStringLiteral;
+import org.apache.commons.jexl2.parser.ASTTrueNode;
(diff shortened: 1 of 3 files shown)


jwomeara (approved):
Noticed that CI was failing, so I poked around a bit and think I have a fix.  All tests are passing now, and a few other cases were added to the new test.  LGTM, but we should probably get another set of eyes on this.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/ValidComparisonVisitorTest.java
```diff
@@ -0,0 +1,136 @@
...
+        ValidComparisonVisitor.check(script);
+    }
+    
+    @Test(expected = DatawaveFatalQueryException.class)
+    public void testInvalidAssignment() throws ParseException {
+        String queryString = "'ASTEvaluationOnly' = true";
+        ASTJexlScript script = JexlASTHelper.parseJexlQuery(queryString);
+        ValidComparisonVisitor.check(script);
+    }
+
```
jwomeara, line 87: You removed the concept of passing found identifiers back up to parent nodes as I was doing with the list implementation.  Add this test and you'll see how it fails.

```suggestion
    
    @Test(expected = DatawaveFatalQueryException.class)
    public void testMethodNodeNested() throws ParseException {
        String queryString = "(AG.greaterThan(39).size() >= 1) == 'false'";
        ASTJexlScript script = JexlASTHelper.parseJexlQuery(queryString);
        ValidComparisonVisitor.check(script);
    }
    
```

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidComparisonVisitor.java
```diff
@@ -0,0 +1,166 @@
...
+            ((MutableInt) data).increment();
+        }
+        return data;
+    }
+    
+    private void validateExpression(JexlNode node, Object data) {
+        if ((data instanceof MutableInt) && (((MutableInt) data).intValue() > 1)) {
+            throw new DatawaveFatalQueryException("Cannot compare two literals.  Invalid expression: " + JexlStringBuildingVisitor.buildQuery(node));
+        }
+    }
```
jwomeara, line 164: Doing something like this to pass found identifiers back up to parent nodes fixes the test case I presented...

```suggestion
    private void validateExpression(JexlNode node, Object data, Object parentData) {
        if ((data instanceof MutableInt)) {
            if ((((MutableInt) data).intValue() > 1)) {
                throw new DatawaveFatalQueryException("Cannot compare two literals.  Invalid expression: " + JexlStringBuildingVisitor.buildQuery(node));
            } else if (parentData instanceof MutableInt) {
                ((MutableInt)parentData).add((((MutableInt) data).intValue()));
            }
        }
    }
```

You will obviously have to pass the parent data as the third argument in the various places where this is called.

jwomeara (changes requested):

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/ValidComparisonVisitorTest.java
ivakegg: I contemplated that, and I don't believe that your added test query should actually fail here.  I am only trying to catch the case where it is a simple literal op literal expression where it is highly likely that the user made an error specifying the query.  If it is something more complicated like this one, then I am more inclined to believe the user knows what they are doing.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidComparisonVisitor.java
ivakegg: See previous comment.

jwomeara (approved):

PR #980, opened 2020-10-30
Handle orphaned files from killed MapFileLoader (#740)

hlgp:


warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/BulkIngestMapFileLoader.java
```diff
@@ -433,9 +434,16 @@ public void run() {
...
         Path[] jobDirectories = new Path[0];
         int nextJobIndex = 0;
+        boolean firstRun = true;
+        
         try {
             while (true) {
                 try {
+                    if (firstRun) {
+                        
+                        cleanJobDirectoriesOnStartup();
```
ivakegg, line 444: Why not put this before the while loop and avoid the firstRun flag altogether?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/BulkIngestMapFileLoader.java
```diff
@@ -534,6 +542,18 @@ public void run() {
         log.info("Bulk map file loader shutting down.");
     }
     
+    protected void cleanJobDirectoriesOnStartup() throws IOException {
+        Path[] cleanupDirectories = getJobDirectories(new Path(workDir, jobDirPattern + '/' + CLEANUP_FILE_MARKER));
+        for (int i = 0; i < cleanupDirectories.length; i++) {
+            
+            markSourceFilesLoaded(cleanupDirectories[i]);
```
ivakegg, line 549: I realize that the markSourceFilesLoaded can supposedly be run by multiple instances concurrently, but how about the runCleanUpScript?  Since that is being passed in, I am thinking this could be a problem.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/BulkIngestMapFileLoader.java
```diff
@@ -534,6 +542,18 @@ public void run() {
         log.info("Bulk map file loader shutting down.");
     }
     
+    protected void cleanJobDirectoriesOnStartup() throws IOException {
+        Path[] cleanupDirectories = getJobDirectories(new Path(workDir, jobDirPattern + '/' + CLEANUP_FILE_MARKER));
```
ivakegg, line 546: I don't think this will work if the srcHdfs is different than the destHdfs, in which case the directory will have been discp'ed to the destHdfs prior to having been processed.  See the routine below that marks the directory for cleanup which works with the destHdfs.

ivakegg (changes requested):

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/BulkIngestMapFileLoader.java
hlgp: Why not, indeed.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/BulkIngestMapFileLoader.java
hlgp: true!

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/BulkIngestMapFileLoader.java
hlgp: We don't actually use this script option anywhere.  Might be an artifact from a bygone era.  Thinking to remove the logic and option for it since it seems hokey to call a bash script like this anyway. Thoughts?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/BulkIngestMapFileLoader.java
```diff
@@ -527,13 +524,26 @@ public void run() {
...
         }
         log.info("Bulk map file loader shutting down.");
     }
     
+    protected void cleanJobDirectoriesOnStartup() throws IOException {
+        Path[] cleanupDirectories = getJobDirectories(destHdfs, new Path(workDir, jobDirPattern + '/' + CLEANUP_FILE_MARKER));
+        for (int i = 0; i < cleanupDirectories.length; i++) {
+            
+            markSourceFilesLoaded(cleanupDirectories[i]);
+            getFileSystem(destHdfs).delete(cleanupDirectories[i], true);
```
ivakegg, line 541: Will this throw an exception is some other process already deleted the directory?

hlgp: Ah, you're right.  I'll adjust that try/catch

ivakegg:
Also please rebase.  Somehow this is showing way more commits than it should....

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/BulkIngestMapFileLoader.java
```diff
@@ -527,13 +528,29 @@ public void run() {
...
     
+    protected void cleanJobDirectoriesOnStartup() throws IOException {
+        Path[] cleanupDirectories = getJobDirectories(destHdfs, new Path(workDir, jobDirPattern + '/' + CLEANUP_FILE_MARKER));
+        for (int i = 0; i < cleanupDirectories.length; i++) {
+            
+            markSourceFilesLoaded(cleanupDirectories[i]);
+            try {
+                getFileSystem(destHdfs).delete(cleanupDirectories[i], true);
+            } catch (IOException e) {
+                log.warn("Unable to delete directory " + cleanupDirectories[i] + ".  Another process already cleaned it up.");
```
ivakegg, line 547: How do you know that is the reason for the IOException?  Perhaps the client could not reach the NN at the time or something else.  Include the message from the IOException as the reason.

ivakegg (changes requested):

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/BulkIngestMapFileLoader.java
hlgp: updated.

ivakegg (approved):

PR #982, opened 2020-11-02
fixes #981 by adding the third param

### LINKED ISSUE #981: Graceful shutdown code of Wildfly is broken
In wildfly 16+, a third parameter was added to the shutdown call for the JMX bean in wildfly. Since we call this after we have waited for our own graceful shutdown, we should just add a 0 to shutdown immediately.

alerman:


alerman:
After further review, that third param should be optional. Digging.

alerman:
> After further review, that third param should be optional. Digging.

While the param is optional, adding it in causes an immediate shutdown once queries are complete. I believe that since we are sending the signature we have to include it. We are setting the default values so we could also just send null,null.

apmoriarty (approved):

alerman:
Rebased on 2.9

apmoriarty (approved):

ivakegg (approved):

Issue #968, opened 2020-10-21
Run validateLineage on original jexl tree AFTER visitor executes


jwomeara:
As a general rule, we should try to code our rebuilding visitors such that they don't modify the original jexl tree that's passed in.  We have many visitors that do not follow this pattern at the moment.  This can cause issues if any of our logic assumes that the original query tree would be unchanged after being processed by a rebuilding visitor.  

We should add a test for each rebuilding visitor which tests whether or not the original query tree is modified after executing that visitor.  Any rebuilding visitors that modify the original tree should be updated to only reflect changes in the resulting output tree.

lbschanno:
Thus far I've verified that the following visitors modify the original tree:

- BooleanOptimizationRebuildingVisitor
- RangeCoalescingVisitor
- RegexFunctionVisitor

lbschanno:
Closing this ticket since we now enable AST validation in our test QLFs, and queries are validated after each visitor via TimedVisitorManager.

PR #993, opened 2020-11-05
Do not modify input tree in RegexFunctionVisitor

lbschanno:
Fix RegexFunctionVisitor so that it does not modify the original input
JEXL tree, and verify that it maintains a valid lineage.

Part of #968


apmoriarty (approved):
LGTM

PR #994, opened 2020-11-05
Fix lineage for UniqueExpressionTermsVisitor

lbschanno:
Fix UniqueExpressionTermsVisitor to ensure that it returns a JEXL tree
with a valid lineage, and does not modify the original input tree.

Part of #880 and #968


warehouse/query-core/src/test/java/datawave/query/jexl/visitors/UniqueExpressionTermsVisitorTest.java
```diff
@@ -236,21 +239,31 @@ public void testDuplicateSubTreeDisjunction() throws ParseException {
...
-        
-        // Assert that both original and expected query strings parse as expected
-        assertEquals("Original query did not parse cleanly" + original, original, JexlStringBuildingVisitor.buildQuery(originalScript));
-        assertEquals("Expected query did not parse cleanly" + expected, expected, JexlStringBuildingVisitor.buildQuery(expectedScript));
         
         // Remove duplicate terms from within expressions.
         ASTJexlScript visitedScript = UniqueExpressionTermsVisitor.enforce(originalScript);
         
-        // De-duped query should match expected
-        assertEquals(expected, JexlStringBuildingVisitor.buildQuery(visitedScript));
```
apmoriarty, line 253: This check should be left in. There have been cases were the tree equality visitor said trees were equal but the queries parsed into different strings.

apmoriarty (changes requested):
Good refactor, I would like to see an assertion added back to the test.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/UniqueExpressionTermsVisitorTest.java
lbschanno: No problem, it's been added back in.

ivakegg (approved):

apmoriarty (approved):
LGTM

PR #1002, opened 2020-11-25
Fixes an issue where ExceededOrThreshold nodes couldn't be evaluated …

jwomeara:
…for doc-specific ranges.

### DIFF
--- a/warehouse/query-core/src/test/java/datawave/query/jexl/nodes/ExceededOrThresholdMarkerJexlNodeTest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/jexl/nodes/ExceededOrThresholdMarkerJexlNodeTest.java
@@ -121,6 +121,7 @@ public class ExceededOrThresholdMarkerJexlNodeTest {
     private static final String POINT_10 = "POINT (1 3)";
     private static final String POINT_11 = "POINT (2 0)";
     private static final String POINT_12 = "POINT (1 0)";
+    private static final String POINT_13 = "POINT (20 20);POINT (20 30)";
     
     private static final String INDEX_1 = "1f0aaaaaaaaaaaaaaa";
     private static final String INDEX_2 = "1f1ffc54fefc54fefc";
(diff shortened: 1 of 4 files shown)


jwomeara:
In the VisitorFunction we used to prohibit the use of FSTs and list ivarators if all of the ranges in a query were doc-specific.  With the move from 2.8 to 2,9, we removed that safeguard thinking it was unnecessary...  As it turns out, it was!

If you want to prove this to yourself, go check out the seek call in the QueryIterator.  In the event that we come into that with a document range, we will not execute the 'buildDocumentIterator' method, which in turn sets up the IteratorBuildingVisitor, which handles population of the exceededOrEvaluationCache, which (until now) was required in order to evaluate ExceededOrNodes during Jexl evaluation.  

The fix contained herein will parse the ExceededOrNode params at evaluation time if they can't be found in the evaluation context.

warehouse/query-core/src/test/java/datawave/query/jexl/nodes/ExceededOrThresholdMarkerJexlNodeTest.java
```diff
@@ -424,15 +431,49 @@ public void valueListTest() throws Exception {
...
             // ensure that this is one of the ingested events
-            Assert.assertTrue(pointList.remove(wkt));
+            Assert.assertTrue(pointList.removeAll(wkt));
+        }
+        
+        Assert.assertEquals(0, pointList.size());
+    }
+    
+    @Test
+    public void docSpecificValueListTest() throws Exception {
```
apmoriarty, line 449: Won't hold up an approval, but it might be nice to see a non-geo case that exercises this code path.

apmoriarty (approved):

alerman (approved):

ivakegg (approved):

warehouse/query-core/src/test/java/datawave/query/jexl/nodes/ExceededOrThresholdMarkerJexlNodeTest.java
jwomeara: Aside from the fact that these values happen to be geohashes, this isn't really a geo-specific test.  We aren't using any geo functions, so it's just a simple term equality evaluation being performed here.

PR #1005, opened 2020-12-02
Fixes an issue where large field lists cannot be executed against …

ivakegg:
Fixes an issue where large field lists cannot be executed against fields with leading numerics


FineAndDandy (approved):

hlgp (approved):

keith-ratcliffe (approved):
Aside from the one typo, looks good to me, assuming that there are no others places, aside from PLFLVisitor, where we need to avoid the identifier deconstruct. Based on our discussion, sounds like we're good there

PR #1006, opened 2020-12-10
Guard against regex and range expansion by default

### LINKED ISSUE #888: Add QueryPlanMetadata to guide DefaultQueryPlanner
The DefaultQueryPlanner will attempt to expand bounded ranges by default, even if no regex nodes exist in the query. 
Add a simple class QueryPlanMetadata that tracks the counts of supported jexl nodes. 
A simple QueryPlanMetadataVisitor should be relatively straightforward to implement.

Building the plan metadata
```
QueryPlanMetadata planMetadata = QueryPlanMetadata.buildPlanMetadata(script)
```

and guarding against regex expansion by default
```
if (queryPlanMetadata.getRegexNodeCount() > 0)
    // then expand bounded ranges
```
or
```
if (queryPlanMetadata.hasRegexNodes())
    // then expand bounded ranges
```

Additional thoughts welcome.

lbschanno:
The DefaultQueryPlanner will attempt to expand regex and bounded ranges
by default, even when no regex nodes exist in the query.

Add a visitor to count each type of node in a query, and use this to
only expand the query when the relevant nodes are present.

Fixes #888


warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -1177,22 +1178,33 @@ protected ASTJexlScript processTree(final ASTJexlScript originalQueryTree, Shard
...
+                    log.debug("No regex to expand");
                 }
-                innerStopwatch.stop();
                 
-                innerStopwatch = timers.newStartedStopwatch("DefaultQueryPlanner - Expand ranges");
-                queryTree = RangeConjunctionRebuildingVisitor.expandRanges(config, scannerFactory, metadataHelper, queryTree, config.isExpandFields(),
-                                config.isExpandValues());
-                if (log.isDebugEnabled()) {
-                    logQuery(queryTree, "Query after expanding ranges:");
+                if (nodeCount.hasPossibleBoundedRange()) {
```
ivakegg, line 1197: Given #926 has been merged in, we can now detect bounded ranges precisely.  Update the node counter to also count the different types of QueryPropertyMarkers.  This can then be if (nodeCount.asBoundedRange()).

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -1235,19 +1251,30 @@ protected ASTJexlScript processTree(final ASTJexlScript originalQueryTree, Shard
...
+                        if (log.isDebugEnabled()) {
+                            logQuery(queryTree, "Query after expanding regex again:");
+                        }
                     }
-                    queryTree = RangeConjunctionRebuildingVisitor.expandRanges(config, scannerFactory, metadataHelper, queryTree, config.isExpandFields(),
-                                    config.isExpandValues());
-                    if (log.isDebugEnabled()) {
-                        logQuery(queryTree, "Query after expanding ranges again:");
+                    
+                    if (nodeCount.hasPossibleBoundedRange()) {
```
ivakegg, line 1264: see comment above

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
lbschanno: Done.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
lbschanno: Done.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -1188,22 +1194,29 @@ protected ASTJexlScript processTree(final ASTJexlScript originalQueryTree, Shard
...
             try {
-                innerStopwatch = timers.newStartedStopwatch("DefaultQueryPlanner - Expand regex");
                 ParallelIndexExpansion regexExpansion = new ParallelIndexExpansion(config, scannerFactory, metadataHelper, expansionFields,
                                 config.isExpandFields(), config.isExpandValues(), config.isExpandUnfieldedNegations());
-                queryTree = (ASTJexlScript) regexExpansion.visit(queryTree, null);
-                if (log.isDebugEnabled()) {
-                    logQuery(queryTree, "Query after expanding regex:");
+                
+                // Check if there is any regex to expand.
+                if (NodeTypeCountVisitor.countNodes(queryTree).hasAny(ASTNRNode.class, ASTERNode.class)) {
```
ivakegg, line 1201: Do we have to call the visitor every time or can we count the nodes once?  There may be places where we need to recount but certainly not at line 1211

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
lbschanno: @ivakegg I made a modification so that the nodes are recounted only when we know the query tree has been modified. Let me know if this works for you.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -1236,22 +1243,34 @@ protected ASTJexlScript processTree(final ASTJexlScript originalQueryTree, Shard
...
+                if (nodeCount.hasAny(ASTNRNode.class, ASTERNode.class)) {
+                    innerStopwatch = timers.newStartedStopwatch("DefaultQueryPlanner - Expand regex");
+                    queryTree = (ASTJexlScript) regexExpansion.visit(queryTree, null);
+                    if (log.isDebugEnabled()) {
+                        logQuery(queryTree, "Query after expanding regex:");
+                    }
+                    innerStopwatch.stop();
+                    
+                    // The query tree was modified. Recount nodes.
+                    nodeCount = NodeTypeCountVisitor.countNodes(queryTree);
```
ivakegg, line 1260: I don't think we need to do this here.  Just do it once after expanding regexes and ranges.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -1275,12 +1294,18 @@ protected ASTJexlScript processTree(final ASTJexlScript originalQueryTree, Shard
                 }
                 innerStopwatch.stop();
                 
-                innerStopwatch = timers.newStartedStopwatch("DefaultQueryPlanner - Expand pushing functions into exceeded value ranges");
-                queryTree = PushFunctionsIntoExceededValueRanges.pushFunctions(queryTree, metadataHelper, config.getDatatypeFilter());
-                if (log.isDebugEnabled()) {
-                    logQuery(queryTree, "Query after expanding pushing functions into exceeded value ranges:");
+                // The query tree was modified. Recount nodes.
+                nodeCount = NodeTypeCountVisitor.countNodes(queryTree);
```
ivakegg, line 1298: may not need this call as this would not affect the call at 1301.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -1307,30 +1332,49 @@ protected ASTJexlScript processTree(final ASTJexlScript originalQueryTree, Shard
...
-                        logQuery(queryTree, "Query after expanding regex again:");
+                    nodeCount = NodeTypeCountVisitor.countNodes(queryTree);
+                    // Check if there is any regex to expand.
+                    if (nodeCount.hasAny(ASTNRNode.class, ASTERNode.class)) {
+                        queryTree = (ASTJexlScript) regexExpansion.visit(queryTree, null);
+                        if (log.isDebugEnabled()) {
+                            logQuery(queryTree, "Query after expanding regex again:");
+                        }
+                        // The query tree was modified. Recount nodes.
+                        nodeCount = NodeTypeCountVisitor.countNodes(queryTree);
```
ivakegg, line 1343: Don't need this call.  Just call once after expanding regexes and ranges

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -1307,30 +1332,49 @@ protected ASTJexlScript processTree(final ASTJexlScript originalQueryTree, Shard
...
+                    
+                    // Check if there are any bounded ranges to expand.
+                    if (nodeCount.isPresent(BoundedRange.class)) {
+                        queryTree = RangeConjunctionRebuildingVisitor.expandRanges(config, scannerFactory, metadataHelper, queryTree, config.isExpandFields(),
+                                        config.isExpandValues());
+                        if (log.isDebugEnabled()) {
+                            logQuery(queryTree, "Query after expanding ranges again:");
+                        }
+                        // The query tree was modified. Recount nodes.
+                        nodeCount = NodeTypeCountVisitor.countNodes(queryTree);
```
ivakegg, line 1354: probably dont' need to do this here.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
ivakegg: I believe we are still calling it too often.  If we expand regexs and ranges, that will only produce more equality nodes and maybe ExceededValue markers.  I don't think any of the subsequent calls actually depend on those counts so why recount?

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
lbschanno: Removed.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
lbschanno: Removed.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
lbschanno: Removed.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
lbschanno: Removed.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
lbschanno: Edited the code to only count the nodes twice, once before the first round of expanding regex and ranges, and once before the second round of expanding regex and ranges after pulling up delayed predicates.

lbschanno:
Rebased to target branch release/version3.3.

ivakegg (approved):

jschmidt10 (approved):

PR #1011, opened 2020-12-13
Exclude dom4j 1.6.1 as a dependency

drewfarris:
* Not necessary for runtime operation
* Triggers a dependabot warning


drewfarris:
See: https://github.com/NationalSecurityAgency/datawave/network/alert/warehouse/pom.xml/dom4j:dom4j/open

ivakegg (approved):

PR #1012, opened 2020-12-14
Cache expressions instead of reparsing for every key.

apmoriarty:


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EventDataQueryExpressionVisitor.java
```diff
@@ -167,78 +182,43 @@ private boolean apply(Key key, boolean update) {
...
-                            } else {
-                                // can't normalize so add the original matcher
-                                normalizedPatternMatchers.add(fieldPatterns.get(fieldPattern));
-                            }
-                        } catch (Exception e) {
-                            // can't normalize this pattern, add the original matcher
-                            normalizedPatternMatchers.add(fieldPatterns.get(fieldPattern));
-                        }
+                    
+                    boolean seenType = seenTypes.contains(type);
```
lbschanno, line 188: The following operations only execute if `seenType` is false. Additionally, `fieldValues`, `fieldPatterns`, and `fieldRanges` are initialized in the constructor, so the check for null is unecessary. Thus, this section can be simplified to:

```java

if(!seenTypes.contains(type)) {
	
    // Check for value normalization
    if(!fieldValues.isEmpty())
        allNormalizedValues.addAll(normalizeValuesForType(type));

    // Check for pattern normalization
    if (!fieldPatterns.isEmpty())
        allNormalizedPatterns.addAll(normalizePatternsForType(type));

    // Check for range normalization
    if(!fieldRanges.isEmpty())
        allNormalizedRanges.addAll(normalizeRangesForType(type));

    // Add this type to seen types.
    seenTypes.add(type);
}
```

apmoriarty: Good suggestion, implemented while incorporated while resolving merge conflicts from bounded range changes.

FineAndDandy (approved):

d-hwang (approved):

alerman (commented):
LGTM

alerman (approved):

PR #1013, opened 2020-12-14
IndexStreams support seeking

apmoriarty:
IndexStreams now support seek-like operation on underlying iterators, improves performance of selective query terms on the global index. Seeking to max is adaptive.

Conflicts:
	warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java


apmoriarty:
Closing in favor of https://github.com/NationalSecurityAgency/datawave/pull/1175

PR #1014, opened 2020-12-14
fixes #1010 fix SpringCDIExtension applicationContext and serialization/deserialization of DatawavePrincipals

### LINKED ISSUE #1010: BulkResults not properly initializing CDI injected objects
SpringCDIExtension not properly initializing the ClassPathXmlApplicationContext -- must split the resource string on commas

This was noticed when a running QueryLogic reached a point in the DocumentTransformer where it was adding cardinality information and called Document.getTimestamp() whiched caused a metadata update and a reference to MarkingFunctions which had to be resolved through Weld/CDI/Spring and failed.

CallerPrincipal and ServerPrincipal encode/decode isn't working

Also -- use an Apache Base64 class for encoding/decoding instead of infinispan

billoley:
SpringCDIExtension not properly initializing the ClassPathXmlApplicationContext -- must split the resource string on commas

This was noticed when a running QueryLogic reached a point in the DocumentTransformer where it was adding cardinality information and called Document.getTimestamp() whiched caused a metadata update and a reference to MarkingFunctions which had to be resolved through Weld/CDI/Spring and failed.

CallerPrincipal and ServerPrincipal encode/decode isn't working

Also -- use an Apache Base64 class for encoding/decoding instead of infinispan

### DIFF
--- a/web-services/map-reduce-embedded/src/main/java/datawave/security/system/EmbeddedCallerPrincipalProducer.java
+++ b/web-services/map-reduce-embedded/src/main/java/datawave/security/system/EmbeddedCallerPrincipalProducer.java
@@ -7,7 +7,11 @@ import javax.enterprise.inject.Produces;
 import javax.interceptor.Interceptor;
 
 import datawave.security.authorization.DatawavePrincipal;
-import org.infinispan.commons.util.Base64;
+import org.apache.commons.codec.binary.Base64;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.ObjectInputStream;
 
 /**
  * Caller principal producer supplied just for Embedded mode (e.g., inside of MapReduce jars). This archive should not be included for normal web applications.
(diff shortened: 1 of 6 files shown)


web-services/map-reduce-embedded/src/main/java/datawave/security/system/EmbeddedCallerPrincipalProducer.java
```diff
@@ -7,7 +7,11 @@
 import javax.interceptor.Interceptor;
 
 import datawave.security.authorization.DatawavePrincipal;
-import org.infinispan.commons.util.Base64;
+import org.apache.commons.codec.binary.Base64;
```
keith-ratcliffe, line 10: Does this allow us to remove the infinispan dependency from the POM?

web-services/map-reduce-embedded/src/main/java/datawave/security/system/EmbeddedCallerPrincipalProducer.java
```diff
@@ -31,8 +35,21 @@ public DatawavePrincipal produceCallerPrincipal() {
...
     private void initializeCallerPrincipal() {
         String encodedCallerPrincipal = System.getProperty("caller.principal");
-        if (encodedCallerPrincipal == null)
-            throw new IllegalStateException("System property caller.principal must be set to a serialized, gzip'd, base64 encoded principal.");
-        callerPrincipal = (DatawavePrincipal) Base64.decodeToObject(encodedCallerPrincipal);
+        if (encodedCallerPrincipal == null) {
+            throw new IllegalStateException("System property caller.principal must be set to a serialized, base64 encoded principal.");
+        }
+        ByteArrayInputStream bais = new ByteArrayInputStream(Base64.decodeBase64(encodedCallerPrincipal));
+        try {
```
keith-ratcliffe, line 42: Try-with-resources instead?

billoley: Updated to use try with resources

web-services/map-reduce-embedded/src/main/java/datawave/security/system/EmbeddedCallerPrincipalProducer.java
billoley: No - it looks like infinispan is used elsewhere.

billoley: Looks like we can probably delete it from the map-reduce-embedded pom

keith-ratcliffe (dismissed):

keith-ratcliffe (approved):

PR #1019, opened 2020-12-18
Groupby function should group fields based on value/name

d-hwang:
Currently, groupby code only compares the value which is breaking when groupby fields have the same value.

### DIFF
--- a/warehouse/query-core/src/test/java/datawave/query/transformer/GroupingTest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/transformer/GroupingTest.java
@@ -195,8 +195,8 @@ public abstract class GroupingTest {
         
         for (EventBase event : eventQueryResponse.getEvents()) {
             
-            String genderKey = "";
-            String ageKey = "";
+            String firstKey = "";
+            String secondKey = "";
             Integer value = null;
             for (Object field : event.getFields()) {
                 FieldBase fieldBase = (FieldBase) field;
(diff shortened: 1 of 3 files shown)


apmoriarty (dismissed):

warehouse/query-core/src/test/java/datawave/query/transformer/GroupingTest.java
```diff
@@ -205,28 +205,31 @@ protected BaseQueryResponse runTestQueryWithGrouping(Map<String,Integer> expecte
...
+                        firstKey = fieldBase.getValueString();
                         break;
                     case "AGE":
-                        ageKey = fieldBase.getValueString();
+                        secondKey = fieldBase.getValueString();
                         break;
                     case "AG":
-                        ageKey = fieldBase.getValueString();
+                        secondKey = fieldBase.getValueString();
                         break;
```
drewfarris, line 221: In theory you could collapse these, but not absolutely required:

```
case "GENDER":
case "GEN":
case "BIRTHDAY":
      firstKey = fieldBase.getValueString()
      break;
case "AGE":
case "AG":
       secondKey = fieldBase.getValueString()
      break;
```


```

drewfarris (dismissed):

warehouse/query-core/src/test/java/datawave/query/transformer/GroupingTest.java
d-hwang: Got it. I rebased and incorporated your suggestion. Thanks!!

PR #1033, opened 2021-01-04
Bugfix for TermFrequencyIterator

apmoriarty:
Bugfix for TermFrequencyIterator: do not prune index only fields, properly calculate when to seek, do not parse keys with DatawaveKey.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyIterator.java
+++ b/warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyIterator.java
@@ -8,8 +8,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
 
+import datawave.core.iterators.key.TFKey;
 import datawave.query.Constants;
-import datawave.query.data.parsers.DatawaveKey;
 
 import org.apache.accumulo.core.data.ArrayByteSequence;
 import org.apache.accumulo.core.data.ByteSequence;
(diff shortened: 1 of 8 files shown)


warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyIterator.java
```diff
@@ -191,36 +190,28 @@ protected void findTop() throws IOException {
                 break;
             }
             
-            if (count > MAX_SCAN_BEFORE_SEEK) {
+            tfKey.parse(k);
```
ivakegg, line 193: Much cleaner, however I wish we could avoid a new class member and instead pass this down into the functions.  It just makes it harder to verify that the members are filled/set wherever they are used.

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyIterator.java
```diff
@@ -324,8 +332,17 @@ public void setKeys(Set<Key> keys) {
     public TreeSet<String> buildUidsFromKeys(Collection<Key> keys) {
         TreeSet<String> uids = new TreeSet<>();
         for (Key key : keys) {
-            DatawaveKey datawaveKey = new DatawaveKey(key);
-            uids.add(datawaveKey.getUid());
+            // keys are CF=datatype\x00uid
+            byte[] backing = key.getColumnFamilyData().getBackingArray();
+            int index = -1;
+            for (int i = 0; i < backing.length; i++) {
```
ivakegg, line 338: Although backing.length will work, it would be safer to use key.getColumnFamilyData().getLength().

warehouse/query-core/src/main/java/datawave/core/iterators/key/TFKey.java
```diff
@@ -0,0 +1,90 @@
...
+                int stop = nulls.get(nulls.size() - 1);
+                value = new String(backing, start, (stop - start));
+            }
+        }
+        return value;
+    }
+    
+    public String getField() {
+        if (field == null) {
+            if (nulls.size() >= 3) {
```
FineAndDandy, line 76: 4

FineAndDandy (changes requested):

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyIterator.java
apmoriarty: No such method?

warehouse/query-core/src/main/java/datawave/core/iterators/key/TFKey.java
apmoriarty: Unless I'm missing something the TF CQ has four pieces and three split points. This is checking the number of split points found, which should be three.

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyIterator.java
apmoriarty: I would tend to agree with you, however in this case everything passes through the hasTop() method so there's a single entry point for new keys.

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyIterator.java
apmoriarty: Got it.

FineAndDandy (approved):

ivakegg (approved):

PR #1032, opened 2021-01-04
Added a pruning step to the QueryIterator init method which removes i…

jwomeara:
…varator cache directories which are not present on the node.

PR #1035, opened 2021-01-06
Prune non-existant Ivarator Cache Base URIs

drewfarris:
. Modifies the QueryIterator to remove configured Ivarator Cache Base URIS that don't exist.
  Similar to #1032 but for the 2.8 branch, which predates IvaratorCacheDirs

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
+++ b/warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
@@ -91,7 +91,8 @@ import org.apache.commons.jexl2.parser.ASTJexlScript;
 import org.apache.commons.lang.builder.CompareToBuilder;
 import org.apache.commons.pool.BasePoolableObjectFactory;
 import org.apache.commons.pool.impl.GenericObjectPool;
-import org.apache.commons.pool2.PooledObject;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.io.Text;
 import org.apache.log4j.Logger;
 import org.apache.zookeeper.server.quorum.QuorumPeerConfig.ConfigException;
@@ -99,6 +100,7 @@ import org.apache.zookeeper.server.quorum.QuorumPeerConfig.ConfigException;
 import javax.annotation.Nullable;
 import java.io.IOException;
 import java.net.MalformedURLException;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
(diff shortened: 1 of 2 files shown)


warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -263,6 +258,32 @@ public void init(SortedKeyValueIterator<Key,Value> source, Map<String,String> op
         if (env != null) {
             ActiveQueryLog.setConfig(env.getConfig());
         }
+        
+        pruneIvaratorCacheBaseURIs();
+    }
+    
+    public void pruneIvaratorCacheBaseURIs() {
+        if (ivaratorCacheBaseURIAlternatives != null) {
+            List<String> newAlternatives = new ArrayList<>(ivaratorCacheBaseURIAlternatives);
```
drewfarris, line 267: Seems like it should be needed, but the list implementation in `ivaratorCacheBaseURIAlternatives` doesn't support `remove`.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -92,22 +92,17 @@
...
 import java.io.IOException;
 import java.net.MalformedURLException;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
drewfarris, line 104: Need to fix wildcard import :-p

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -263,6 +265,32 @@ public void init(SortedKeyValueIterator<Key,Value> source, Map<String,String> op
         if (env != null) {
             ActiveQueryLog.setConfig(env.getConfig());
         }
+        
+        pruneIvaratorCacheBaseURIs();
+    }
+    
+    public void pruneIvaratorCacheBaseURIs() {
+        if (ivaratorCacheBaseURIAlternatives != null) {
```
ivakegg, line 273: to be consistent, use this.ivaratorCacheBaseURIAlternatives

ivakegg (approved):

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
jwomeara: That's whack yo!  But I get it!

jwomeara (changes requested):
Happy to approve once the wildcard import is taken out.

PR #1043, opened 2021-01-15
Make ParallelIndexExpansion max threads configurable

### LINKED ISSUE #1031: ParallelIndexExpansion capping threads at 10
The ParallelIndexExpansion is capping the number of threads at 10.  This need to be configurable during system deployment via  the QueryLogicFactory.xml.

lbschanno:
The ParallelIndexExpansion caps the number of index lookup threads at
10. Make this configurable via the QueryLogicFactory.xml.

Fixes #1031

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
+++ b/warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
@@ -333,9 +333,9 @@ public class ShardQueryConfiguration extends GenericQueryConfiguration implement
     private List<String> contentFieldNames = Collections.emptyList();
     
     /**
-     * The default maximum number of threads to use when performing an index lookup. Default value: 10.
+     * The default minimum number of threads to use when performing an index lookup. Default value: 10.
      */
-    private int defaultMaxNumIndexLookupThreads = 10;
+    private int minIndexLookupThreads = 10;
     
     /**
      * Default constructor
@@ -511,7 +511,7 @@ public class ShardQueryConfiguration extends GenericQueryConfiguration implement
         this.setTrackSizes(other.isTrackSizes());
         this.setContentFieldNames(null == other.getContentFieldNames() ? null : Lists.newArrayList(other.getContentFieldNames()));
         this.setEvaluationOnlyFields(other.getEvaluationOnlyFields());
-        this.setDefaultMaxNumIndexLookupThreads(other.getDefaultMaxNumIndexLookupThreads());
+        this.setMinIndexLookupThreads(other.getMinIndexLookupThreads());
     }
     
     /**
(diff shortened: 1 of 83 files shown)


warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
```diff
@@ -332,6 +332,11 @@
     
     private List<String> contentFieldNames = Collections.emptyList();
     
+    /**
+     * The default maximum number of threads to use when performing an index lookup. Default value: 10.
+     */
+    private int defaultMaxNumIndexLookupThreads = 10;
```
keith-ratcliffe, line 338: Overall the PR looks good, except I think the choice of field name here may be kind of confusing for users, given its relationship to `numIndexLookupThreads` in QLF.xml.

The way the code currently works, this value serves as a default minimum number of threads rather than max, because you can't go lower than this value....e.g., if the user tries to set `numIndexLookupThreads = 1`, they're going to get 10 instead.

Maybe `defaultIndexLookupThreads` instead? Or `minIndexLookupThreads`?

keith-ratcliffe (changes requested):

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
keith-ratcliffe: Or alternatively, rather than have two distinct config parameters for this, maybe we could just stick with `numIndexLookupThreads` as the only configurable property, and then change the default minimum be a local private constant here, setting its value to 1 instead of 10.  That would resolve the original limitation with the code here, right?

lbschanno: @keith-ratcliffe good catch there about the actual behavior. I've renamed the property to `minIndexLookupThreads` instead to make it clearer.

I'm hesitant to change the default minimum to a constant of 1, because it seems that the intent is to enforce a particular minimum number of threads, rather than simply relying on `numIndexLookupThreads`.

keith-ratcliffe: I think `numIndexLookupThreads` is something that typically will get overridden already in QLF.xml to be 10 or greater. Setting `numIndexLookupThreads` to a value of 1 does have particular value when debugging, and that's something the code has never allowed us to do, based on the current hardcoded default min of 10. That said, I _think_ it's safe to have a constant value of 1 here instead, but I guess I'd defer to @ivakegg on this one. Not sure what he had in mind exactly

ivakegg: OK, being of sound mind and thought, I am of the opinion that I must have been hallucinating when I wrote that ticket.

That being said lets see if there is anything to do here.  If we are to configure a minimum number of threads, then allocate a ThreadPoolExecutor directly and supply a separate "corethreads" from the "maxthreads" parameter.  Otherwise lets change the 10 to a 1 and be done with it.  What do you guys think (@keith-ratcliffe , @lbschanno)?

keith-ratcliffe: @ivakegg I guess my vote would be for just having a private/const 1 as the minimum and stick with the existing config param as-is

lbschanno: @ivakegg I'm all for keeping it simple, so my vote is for just changing the 10 to a 1.

keith-ratcliffe (approved):
Looks good

ivakegg (approved):

PR #1044, opened 2021-01-19
Shorten labels for QueryPropertyMarker classes

lbschanno:
Currently it is possible for a query plan to get too large due to the
large class names of some QueryPropertyMarker subclasses.

Implement an abbreviated labeling system for QueryPropertyMarker classes
that will take the place of the class names in queries.


ivakegg:
I like shorter labels, but I feel those are too short.  How about:
List (for exceededOrThreshold), Bounded, Delay, Hole, Eval, Term, and Value.
Please let @jwomeara weigh in before you go changing them again.

lbschanno:
@ivakegg sure thing, I'll wait.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPropertyMarkerVisitor.java
```diff
@@ -57,42 +64,47 @@
...
+    private List<JexlNode> sourceNodes;
     
-    private boolean identifierFound = false;
-    
-    protected Set<String> rejectedIdentifiers = new HashSet<>();
-    private boolean rejectedIdentifiersFound = false;
-    
-    private QueryPropertyMarkerVisitor() {}
+    private boolean whitelistedTypeFound = false;
+    private boolean blacklistedTypeFound = false;
```
jwomeara, line 72: Consider using alternative terminology like allow and deny vs. what you have here.

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/ExceededOrThresholdMarkerJexlNode.java
```diff
@@ -34,6 +34,12 @@
     public static final String EXCEEDED_OR_FIELD = "field";
     public static final String EXCEEDED_OR_PARAMS = "params";
     
+    private static final String LABEL = "EOTM";
```
jwomeara, line 37: Consider shortening this to EOT, or even EO since the 'threshold' and 'marker' parts are sort of implied.

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/ExceededTermThresholdMarkerJexlNode.java
```diff
@@ -8,6 +8,12 @@
  */
 public class ExceededTermThresholdMarkerJexlNode extends QueryPropertyMarker {
     
+    private static final String LABEL = "ETTM";
```
jwomeara, line 11: Same as above - consider shortening to ETT or ET.

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/ExceededValueThresholdMarkerJexlNode.java
```diff
@@ -8,6 +8,12 @@
  */
 public class ExceededValueThresholdMarkerJexlNode extends QueryPropertyMarker {
     
+    private static final String LABEL = "EVTM";
```
jwomeara, line 11: Same as above - consider shortening to EVT or EV.

jwomeara (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/IndexHoleMarkerJexlNode.java
```diff
@@ -8,6 +8,12 @@
  */
 public class IndexHoleMarkerJexlNode extends QueryPropertyMarker {
     
+    private static final String LABEL = "IOM";
```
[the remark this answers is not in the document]
lbschanno: Good catch, this is a typo and it was meant to be "IHM".

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPropertyMarkerVisitor.java
lbschanno: Sure, no problem.

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/ExceededOrThresholdMarkerJexlNode.java
lbschanno: Making this change wouldn't be a problem, but first, did you see @ivakegg's comment above about using shorter labels that aren't as short as acronyms? Any thoughts on that?

ivakegg: @jwomeara We need your feedback relative to coming up with a compromise between acronyms and something a little more understandable by the non-query savvy bystander .

jwomeara: I'm all about them acronyms.  *shrug*

What would the middle-ground look like?  Can you give me an example @ivakegg ?

jwomeara (approved):
I'm happy with the changes functionally.  We need to decide whether we want to shorten the markers further (which I am in favor of) or come up with some kind of middle-ground between that and what they used to look like.

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/ExceededOrThresholdMarkerJexlNode.java
lbschanno: @jwomeara, @ivakegg provided the follow examples above:

> I like shorter labels, but I feel those are too short. How about:
List (for exceededOrThreshold), Bounded, Delay, Hole, Eval, Term, and Value.

ivakegg: @jwomeara pinged me on the side and is OK with whatever labels we agree to.  So lets use the short word labels instead.  Start with the ones I suggested.  Thanks.

ivakegg: @lbschanno Important safety tip!  I just remembered that these labels actually translate to fields that are added into the JexlContext during the evaluation of the query.  So please add an underscore at the beginning and end of each of these labels.  Thanks.

lbschanno: @ivakegg good advice, will do.

lbschanno: @ivakegg I changed the labels.

PR #1048, opened 2021-01-23
Return singular Instance from QueryPropertyMarkerVisitor

### LINKED ISSUE #1022: Update QueryPropertyMarker/QueryPropertyMarkerVisitor to return the classes it finds
Instead of passing in a class type to the QueryPropertyMarkerVisitor, we should update it to just return all of the QueryPropertyMarker instances that it finds for a given node.  Then we can make a determination based on what it finds.  This will also discourage people from making multiple instanceOf calls back to back against a single node.  

We should also consider creating a specific method which determines whether the node is a delayed node as we will soon have more than just delayed property markers with the new BoundedRange marker.

lbschanno:
The current API for QueryPropertyMarkerVisitor does not discourage
people from making multiple instanceOf calls back to back against a
single node.

Update the API to instead return a QueryPropertyMarker.Instance that can
be used to identify the marker's type (if present) and the marker's
source.

Fixes #1022


warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
```diff
@@ -192,19 +192,10 @@ private JexlNode getOrNode(JexlNode node) {
...
-            
-            return ExceededValueThresholdMarkerJexlNode.getExceededValueThresholdSource(delayedNode);
-        } else if (ExceededTermThresholdMarkerJexlNode.instanceOf(delayedNode)) {
-            
-            return ExceededTermThresholdMarkerJexlNode.getExceededTermThresholdSource(delayedNode);
-        } else if (IndexHoleMarkerJexlNode.instanceOf(delayedNode)) {
-            
-            return IndexHoleMarkerJexlNode.getIndexHoleSource(delayedNode);
+        QueryPropertyMarker.Instance instance = QueryPropertyMarker.findInstance(delayedNode);
+        if (instance.isAnyTypeOf(ASTDelayedPredicate.class, ExceededOrThresholdMarkerJexlNode.class, ExceededValueThresholdMarkerJexlNode.class,
```
ivakegg, line 196: nice

ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
lbschanno: Thanks!

jwomeara (commented):
Still looking through this, and I like what I see so far.  You definitely captured what I was thinking of when I created the ticket.  

I did have a question though - does this handle the case where a node might have multiple markers applied to it?  I don't know how valid a case like that would be, but I figured I would ask.

lbschanno:
@jwomeara I'm not sure, I didn't test for that case since I thought it was only possible for there to be a single marker type within a marker node. Would that look 
`(ASTDelayedPredicate = true && IndexHoleMarker=true) && (FOO == 'a')` 
or 
`((ASTDelayedPredicate=true) && (IndexHoleMarker=true)) && (FOO == 'a')`
or something else?

lbschanno:
@jwomeara I confirmed with @ivakegg that a node cannot have multiple markers applied to it, so this ticket does not need to handle that case.

lbschanno:
Rebase branch to target release/version3.3.

jwomeara (approved):

ivakegg (approved):

PR #1053, opened 2021-01-28
Restrict access by DN in query execution service

### LINKED ISSUE #1027: Update the query execution service to filter by DN
The Query object contains the userDN and a DNlist.  The query executor bean need to be able to filter by a set of allowed DNs per query logic.  Hence the BaseQueryLogic needs an additional member called allowedDNs.  Then the QueryExecutorService needs to be able to deny access to a query logic if that list is non-empty and does not match and of the DNs in the DNlist.

lbschanno:
Support restricting access to query logics that have defined sets of
authorized DNs to only calling principals with an authorized DN.

Fixes #1027

### DIFF
--- a/web-services/query/src/main/java/datawave/webservice/query/logic/QueryLogic.java
+++ b/web-services/query/src/main/java/datawave/webservice/query/logic/QueryLogic.java
@@ -1,9 +1,5 @@
 package datawave.webservice.query.logic;
 
-import java.security.Principal;
-import java.util.List;
-import java.util.Set;
-
 import datawave.audit.SelectorExtractor;
 import datawave.marking.MarkingFunctions;
 import datawave.validation.ParameterValidator;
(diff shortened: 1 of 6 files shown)


web-services/query/src/main/java/datawave/webservice/query/logic/QueryLogic.java
```diff
@@ -297,4 +297,35 @@ default String getResponseClass(Query query) throws QueryException {
...
+     * <li>The set of authorized DNs is not empty, and the provided collection contains a DN that is also found within the set of authorized DNs.</li>
+     * </ul>
+     * 
+     * @param dns
+     *            the DNs to determine access rights for
+     * @return true if the collection contains at least one DN that has access to this query logic, or false otherwise
+     */
+    default boolean containsDNWithAccess(Collection<String> dns) {
+        Set<String> authorizedDNs = getAuthorizedDNs();
+        return authorizedDNs == null || authorizedDNs.isEmpty() || (dns != null && dns.stream().anyMatch(authorizedDNs::contains));
```
ivakegg, line 329: nice

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -488,7 +489,17 @@ private QueryData validateQuery(String queryLogicName, MultivaluedMap<String,Str
             Arrays.sort(dns);
             qd.dnList = Arrays.asList(dns);
             qd.proxyServers = dp.getProxyServers();
+            
+            // Verify that the calling principal has access to the query logic.
+            if (!qd.logic.containsDNWithAccess(qd.dnList)) {
+                UnauthorizedQueryException qe = new UnauthorizedQueryException("User " + qd.userDn + " is not part of the authorized DNs for the query logic "
```
ivakegg, line 495: This message is a little misleading as we are looking for an intersection with qd.dnList, not just the qd.userDn.  Perhaps something like "None of the DNs used have access to this query logic: " + qd.dnList.

lbschanno: Good point. I've modified the exception message as requested.

ivakegg (approved):

PR #1058, opened 2021-02-03
WIP Fetch TermFrequency Offsets via precomputed column qualifiers

apmoriarty:
Compute term frequency search space per document, fetch offsets via precomputed TF column qualifiers.

The cost to compute the search space per document is extremely low compared to the cost to search all *possible* term frequency hits for a document. 

The TermFrequencyIterator is replaced by the TermFrequencyOffsetIterator.


warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/FieldValue.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+    
+    public static FieldValue getFieldValue(Text cqText) {
+        if (cqText == null) {
+            return null;
+        }
+        return getFieldValue(cqText.toString());
+    }
+    
+    public static FieldValue getFieldValue(String cq) {
```
lbschanno, line 101: Consider returning `Optional<FieldValue>` for these static methods instead. This will imply that it's possible for a null value to be returned, and reduce the chance of NPEs. It will also help in any functional programming usages.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/FieldValue.java
```diff
@@ -0,0 +1,131 @@
...
+import datawave.util.StringUtils;
+import org.apache.accumulo.core.data.Key;
+import org.apache.hadoop.io.Text;
+
+/**
+ * A field name and value which is sorted on {@code <value>\0<name>}
+ */
+@Deprecated
+public class FieldValue implements Comparable<FieldValue> {
+    private int nullOffset;
```
lbschanno, line 12: These can be final.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/FieldValue.java
```diff
@@ -0,0 +1,131 @@
...
+        } else {
+            return bytes[index];
+        }
+    }
+    
+    public String getValueField() {
+        return valueField;
+    }
+    
+    public String getField() {
```
lbschanno, line 59: I'm curious, why not also store the value and field as separate attributes for this class? Then you won't need to perform the `substring()` call every time, and it will negate the need for `nullOffset`.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/FieldValue.java
```diff
@@ -0,0 +1,131 @@
...
+        String value = "";
+        String field = "";
+        if (cqParts.length >= 4) {
+            field = cqParts[cqParts.length - 1];
+            value = cqParts[2];
+            // in case the value had null characters therein
+            for (int i = 3; i < (cqParts.length - 1); i++) {
+                value = value + '\0' + cqParts[i];
+            }
+        } else if (cqParts.length == 3) {
```
lbschanno, line 124: This can be a simple else because `cqParts` is guaranteed to be 3; checks have already been performed to see if `cqParts` is <= 2 or >= 4.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermFrequencyHitFunction.java
```diff
@@ -0,0 +1,331 @@
...
+        return fields;
+    }
+    
+    // A node is special if it is not the fields being searched. That is, the first node is a number or a variable 'termOffsetMap'
+    private boolean isFirstNodeSpecial(JexlNode node) {
+        if (node instanceof ASTNumberLiteral) {
+            return true;
+        } else if (node instanceof ASTReference) {
+            List<ASTIdentifier> ids = JexlASTHelper.getIdentifiers(node);
+            if (ids.size() == 1 && ids.get(0).image.equals("termOffsetMap")) {
```
lbschanno, line 274: This can be simplified to `return ids.size() == 1 && ids.get(0).image.equals("termOffsetMap");`

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/FieldValue.java
```diff
@@ -0,0 +1,131 @@
+package datawave.query.postprocessing.tf;
+
+import datawave.util.StringUtils;
+import org.apache.accumulo.core.data.Key;
+import org.apache.hadoop.io.Text;
+
+/**
+ * A field name and value which is sorted on {@code <value>\0<name>}
+ */
+@Deprecated
```
lbschanno, line 10: Do you mean for this to be deprecated?

apmoriarty: Short answer yes. Long answer, the FieldValue class was a private static class within the TermFrequencyIterator. This merge request moves it out of the way for a lighter weight parsing class, TFKey

lbschanno: Ah I see, thanks for explaining. Feel free to disregard any of my comments about this class if it's being phased out.

apmoriarty: Actually, if you want to take a look at TFKey and drop any comments on there that would be great. Always happy to have a second set of eyes on new code

lbschanno: The main thing that jumps out at me is `TFKey.parse()`. I feel it would be more intuitive for this to be a static method that returns a new TFKey, rather than modify an existing TFKey. But that's mainly just a design preference.

apmoriarty:
quickstart action blocked by https://github.com/NationalSecurityAgency/datawave/issues/1111

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyIterator.java
```diff
@@ -191,6 +191,10 @@ protected void findTop() throws IOException {
             }
             
             tfKey.parse(k);
+            if (!tfKey.isValid()) {
+                log.error("Malformed term frequency key: " + k.toStringNoTime());
+                source.next();
```
ivakegg, line 196: do we really want to skip it?

ivakegg (commented):
If I am not mistaken, this one actually degrades the performance of the EventQuery, but avoids some very bad performance on TLD queries.  If that is the case, then can we avoid doing the prefetch when doing a simply EventQuery?

apmoriarty:
> 
> 
> If I am not mistaken, this one actually degrades the performance of the EventQuery, but avoids some very bad performance on TLD queries. If that is the case, then can we avoid doing the prefetch when doing a simply EventQuery?

[The code block here](https://github.com/NationalSecurityAgency/datawave/pull/1058/files#diff-f5879684435d05cf161eb782b1cf68495741cf947aa07847e2271f0f98e375ccR126-R139) does precisely what you're asking about. Only negated terms are fetched, and only in the case of a TLD query.

drewfarris:
This may be out of the scope of the review, but `TFKey's` usage of an `ArrayList` to hold the nulls makes me feel a little weird. Could we possibly build the list in an ArrayList, but convert it to an `int[]` when we're done? I have no empirical evidence this will improve performance, but elimination the virtual function lookup and stack traversal seems like it would be reasonable to do. Depending on the design of TFKey, we could always make this int[3] and ignore 'extra' nulls because they would appear in the value.

drewfarris:
> This may be out of the scope of the review, but `TFKey's`...

One other thought on the design of `TFKey` - from the design, it looks like we're focused on 'lazy' parsing, do you have a sense as to whether this is giving us a significant performance advantage (e.g: it is common the create a TFKey and then only read the Uid but not the value?)

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyIterator.java
drewfarris: If we want to skip an invalid key, don't we want an explicit `continue` after `source.next()` to restart the while loop at 180? It is not clear that to me whether nothing will hapen in lines 199-209 below, with the worst case possibly calling source.next() again in the else clause.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/TermFrequencyList.java
```diff
@@ -81,17 +82,31 @@ public String toString() {
...
+                if (index == 0) {
+                    index = i; // first null byte
+                } else {
+                    index = i; // second null byte
+                    break;
+                }
+            }
+        }
+        
+        eventId.append(new String(backing.subSequence(0, index).toArray()));
```
drewfarris, line 109: I think you can avoid the array copy implicit in the `ArrayByteSequence.subSequence()/toArray()` call if you just use the `String(byte[], int, int)` constructor on the original ByteSequence's backing here instead.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermOffsetFunction.java
```diff
@@ -1,35 +1,42 @@
...
-    private Set<String> tfIndexOnlyFields;
     
-    public TermOffsetFunction(TermOffsetPopulator tfPopulator, Set<String> tfIndexOnlyFields) {
+    // Responsible for reducing the term frequency search space
+    private TermFrequencyHitFunction hitFunction;
+    
+    public TermOffsetFunction(TermOffsetPopulator tfPopulator, ASTJexlScript script, SortedKeyValueIterator<Key,Value> source, boolean isTld) {
         this.tfPopulator = tfPopulator;
-        this.tfIndexOnlyFields = tfIndexOnlyFields;
+        this.hitFunction = new TermFrequencyHitFunction(script, tfPopulator.getTermFrequencyFieldValues());
```
drewfarris, line 32: I'm still getting my head around this, but when I see the `TermFrequencyHitFunction` embedded within `TermOffsetFunction` - I wonder if we could make them 'siblings' instead of 'parent/child' functions; Meaning: is there an advantage if the caller of `TermOffsetFunction` also called `TermFrequencyHitFunction` and passes the data from one to another in the appropriate sequence. Does this make either easier to unit test?

drewfarris: It looks like there's a layer of abstraction here with `TFFactory` and we effectively have the consumer calling a specifically functional interface, so disregard my original comment. However, there's something about the relationship between `TermOffsetFunction`, `TermOffsetPopulator` and `TermFrequencyHitFunction` that seems a little asymmetrical. Perhaps it is as simple as constructing the `TermFrequencyHitFunction` in the `TFFactory` implementation the same way we do the `TermOffsetPopulator`. The other thought would be to restructure these three classes slightly so we essentially have a `TermOffsetFunction` implementation that does it all.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TFFactory.java
```diff
@@ -40,29 +43,34 @@
...
+     *            a source copy used to fetch term frequencies
+     * @param secondSource
+     *            a source copy used to fetch delayed values from the field index (negated content function terms)
      * @return
      */
     public static com.google.common.base.Function<Tuple2<Key,Document>,Tuple3<Key,Document,Map<String,Object>>> getFunction(ASTJexlScript query,
                     Set<String> contentExpansionFields, Set<String> termFrequencyFields, Multimap<String,Class<? extends Type<?>>> dataTypes,
-                    Equality equality, EventDataQueryFilter evaluationFilter, SortedKeyValueIterator<Key,Value> sourceDeepCopy, Set<String> tfIndexOnlyFields) {
+                    Equality equality, EventDataQueryFilter evaluationFilter, SortedKeyValueIterator<Key,Value> sourceDeepCopy,
+                    SortedKeyValueIterator<Key,Value> secondSource, boolean isTld) {
```
drewfarris, line 64: Consider the api ergonomics here. Is it reasonable to expect the caller to know that it should `deepCopy` the source twice? Or should the caller just send `TFFactory` a source, and `TFFactory` is responsible for deepCopying it as many times as the implementation requires?  It also makes me wonder whether we should bundle all of these 'configuration' objects like `contentExpansionFields`, `termFrequencyFields`, `tfIndexOnlyFields` together into a single parent configuration object.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/PushdownNegationVisitorTest.java
```diff
@@ -282,4 +283,15 @@ public void testApplyDeMorgansNegateOrRoot() {
...
         Assert.assertEquals("!((!(f1 == 'v1') && !(f2 == 'v2')))", JexlStringBuildingVisitor.buildQuery(result));
     }
+    
+    @Test
+    public void testPushdownNegationIntoContentFunction() throws ParseException {
+        String query = "FOO == 'few' && !(content:phrase('TEXT', termOffsetMap, 'bar', 'baz') && (TEXT == 'bar' && TEXT == 'baz'))";
+        ASTJexlScript script = JexlASTHelper.parseJexlQuery(query);
+        
+        JexlNode pushed = PushdownNegationVisitor.pushdownNegations(script);
+        String expected = "FOO == 'few' && ((!(content:phrase('TEXT', termOffsetMap, 'bar', 'baz')) || !(TEXT == 'bar') || !(TEXT == 'baz')))";
```
drewfarris, line 293: I feel like I'm missing something here, but I'll ask the question anyway: The first form enforces that the query does not match "bar baz", "bar" and "baz". The second form seems to convert this to not match "bar baz" or not match "bar" or not match "baz". 

The user's intent in the first form is to exclude "bar baz" from the results, but it seems like the second form would exclude "bar" and/or "baz" if they appear by themselves outside the context of a phrase.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -982,12 +982,13 @@ public T next() {
...
+                // First source fetches term offsets
+                SortedKeyValueIterator<Key,Value> firstSource = sourceDeepCopy.deepCopy(myEnvironment);
+                // Second source fetches tf field value pairs from the field index for delayed (negated) phrases
+                SortedKeyValueIterator<Key,Value> secondSource = sourceDeepCopy.deepCopy(myEnvironment);
                 
-                Function<Tuple2<Key,Document>,Tuple3<Key,Document,Map<String,Object>>> tfFunction;
-                tfFunction = TFFactory.getFunction(getScript(documentSource), getContentExpansionFields(), getTermFrequencyFields(), this.getTypeMetadata(),
-                                super.equality, getEvaluationFilter(), sourceDeepCopy.deepCopy(myEnvironment), tfIndexOnlyFields);
+                Function<Tuple2<Key,Document>,Tuple3<Key,Document,Map<String,Object>>> tfFunction = buildTfFunction(getScript(documentSource), firstSource,
+                                secondSource);
```
drewfarris, line 991: Perhaps we allow the function to `deepCopy` the iterators if it needs to, not all function implementations would need two sources?

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyOffsetIterator.java
```diff
@@ -0,0 +1,163 @@
...
+        if (nextCQ == null)
+            return null;
+        
+        // Update the start key's column qualifier
+        Key start = initialRange.getStartKey();
+        start = new Key(start.getRow(), start.getColumnFamily(), nextCQ);
+        
+        // Update the initial range with the new start key
+        return new Range(start, true, initialRange.getEndKey(), initialRange.isEndKeyInclusive());
+    }
```
drewfarris, line 130: Consider minimizing the diffs between this class and `TermFrequencyIterator` so that it is easy to compare implementations. For example consider moving this entire section to the end of the class.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/DelayedFieldIndexIterator.java
```diff
@@ -0,0 +1,112 @@
...
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.Map;
+
+/**
+ * Iterator that fetches all keys defined by the seek range. Used by the {@link TermFrequencyHitFunction} to determine if a field value pair is present in a
+ * document. In the TLD case this iterator will also fetch all child uids associatd with the TLD.
+ */
+public class DelayedFieldIndexIterator extends WrappingIterator {
```
drewfarris, line 20: As discussed, consider dropping this entirely? It's just a wrapper around a source and doesn't do anything special. If we do indeed just need a wrapper, perhaps there's another existing implementation of `WrappingIterator` we can use so we don't need to implement and maintain this one?

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/FieldValue.java
drewfarris, line 10: Consider whether we should just delete this outright.

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyOffsetIterator.java
```diff
@@ -0,0 +1,163 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
+
+import static datawave.query.Constants.TERM_FREQUENCY_COLUMN_FAMILY;
+
+/**
+ * Alternate implementation of the {@link TermFrequencyIterator} that operates on precomputed fully-qualified column qualifiers
```
drewfarris, line 24: Clarify comment? I'm not sure what a 'precomputed fully-qualified column qualifier is' - or what would make a column qualified `partially-qualified`.

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyOffsetIterator.java
```diff
@@ -0,0 +1,163 @@
...
+    
+    private Key tk = null;
+    private Value tv = null;
+    
+    // A set of full column qualifiers that describe the search space
+    private final TreeSet<Text> searchSpace;
+    
+    private Range initialRange;
+    
+    private final ByteSequence tfBytes = new ArrayByteSequence(TERM_FREQUENCY_COLUMN_FAMILY.getBytes(), 0, TERM_FREQUENCY_COLUMN_FAMILY.getLength());
```
drewfarris, line 38: Could this be static as well?

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyOffsetIterator.java
```diff
@@ -0,0 +1,163 @@
...
+    protected Collection<ByteSequence> seekCFs = Collections.singleton(tfBytes);
```
drewfarris, line 39: Should this be `final`?

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyOffsetIterator.java
```diff
@@ -0,0 +1,163 @@
...
+    
+    private SortedKeyValueIterator<Key,Value> source;
+    
+    public TermFrequencyOffsetIterator(Set<Text> searchSpace) {
```
drewfarris, line 43: Consider dropping unused constructor?

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermFrequencyHitFunction.java
```diff
@@ -0,0 +1,689 @@
...
+import java.util.TreeSet;
+
+/**
+ * Function that determines the search space for a query's content functions given a document.
+ *
+ * The search space is a set of fully qualified TermFrequency column qualifiers
+ */
+public class TermFrequencyHitFunction {
+    
+    private final Logger log = Logger.getLogger(TermFrequencyHitFunction.class);
```
drewfarris, line 49: `static`?

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermFrequencyHitFunction.java
```diff
@@ -0,0 +1,689 @@
...
+    private final Multimap<String,String> tfFVs;
+    
+    private boolean hasNegatedFunctions;
+    
+    // Document built from any negated content function terms that hit
+    private Document negatedDocument;
+    
+    private SortedKeyValueIterator<Key,Value> source;
+    
+    // use this one weird trick to avoid an obnoxious number of source deep copies!
```
drewfarris, line 72: 🙄 Hah.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermFrequencyHitFunction.java
```diff
@@ -0,0 +1,689 @@
...
+    private List<SubQuery> parseFunction(Function function) {
+        // functions may take different forms..
+        // within = {field, number, termOffsetMap, terms...}
+        // within = {number, termOffsetMap, terms...}
+        // adjacent = {field, termOffsetMap, terms...}
+        // adjacent = {termOffsetMap, terms...}
+        // phrase = {field, termOffsetMap, terms...}
+        // phrase = {termOffsetMap, terms...}
+        List<JexlNode> args = function.args();
+        int index = function.name().equals("within") ? 3 : 2;
```
drewfarris, line 561: It makes me feel a little strange that we go to the trouble to enumerate our expectations here in a comment, but we don't entirely enforce them in code. I would certainly feel more comfortable if we enforced that the function was one of 'within', 'adjacent' or 'phrase' if those are truly the only functions we plan to handle, and then somehow gracefully degrade if we don't understand the function we've encountered.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermFrequencyHitFunction.java
```diff
@@ -0,0 +1,689 @@
...
+    }
+    
+    private List<String> parseArg(JexlNode node) {
+        List<String> parsed = new LinkedList<>();
+        List<Object> values = JexlASTHelper.getLiteralValues(node);
+        for (Object value : values) {
+            if (value instanceof String) {
+                parsed.add((String) value);
+            } else {
+                throw new IllegalStateException("Expected literal value to be String cast-able");
```
drewfarris, line 664: How will this be handled elsewhere?

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermFrequencyHitFunction.java
```diff
@@ -0,0 +1,689 @@
...
+    }
+    
+    /**
+     * A multi-fielded function can be thought of as the conjunction of multiple sub queries.
+     */
+    private class SubQuery {
+        int id;
+        boolean isNegated;
+        String field;
+        TreeSet<String> fieldValues;
```
drewfarris, line 677: Consider making these all `final` - immutability ftw.

warehouse/query-core/src/test/java/datawave/core/iterators/key/TFKeyTest.java
```diff
@@ -4,6 +4,7 @@
 import org.junit.Test;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
 
 public class TFKeyTest {
```
FineAndDandy, line 9: add isValid test?

FineAndDandy (changes requested):

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermOffsetPopulator.java
```diff
@@ -175,57 +148,43 @@ protected Range getRange(Set<Key> keys) {
...
             
-            // no need to apply the evaluation filter here as the TermFrequencyIterator above is already doing more filtering than we can do here.
-            // So this filter is simply extraneous. However if the an EventDataQueryFilter implementation gets smarter somehow, then it can be added back in
-            // here.
-            // For example the AncestorQueryLogic may require this....
-            // if (evaluationFilter == null || evaluationFilter.apply(Maps.immutableEntry(key, StringUtils.EMPTY_STRING))) {
+            // Malformed keys will fail to parse, skip them.
+            if (field == null || value == null) {
+                log.error("Malformed term frequency key: " + key.toStringNoTime());
+                continue;
```
drewfarris, line 166: I believe we need to call `offsetIter.next()` prior to `continue` in this case or we'll just get stuck evaluating the same key when an error is encountered.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermOffsetPopulator.java
```diff
@@ -234,11 +193,48 @@ protected Range getRange(Set<Key> keys) {
...
+    /**
+     * Build a range from the search space.
+     *
+     * @param row
+     *            the shard
+     * @param searchSpace
+     *            a sorted set of TF column qualifiers
+     * @return a range encompassing the full search space
+     */
+    public static Range getRangeFromSearchSpace(Text row, TreeSet<Text> searchSpace) {
```
drewfarris, line 209: Perhaps we can indeed use `SortedSet` instead of `TreeSet`?

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyIterator.java
apmoriarty: An explicit continue is the right thing to do here, I'll add that in.

apmoriarty:
> 
> 
> This may be out of the scope of the review, but `TFKey's` usage of an `ArrayList` to hold the nulls makes me feel a little weird. Could we possibly build the list in an ArrayList, but convert it to an `int[]` when we're done? I have no empirical evidence this will improve performance, but elimination the virtual function lookup and stack traversal seems like it would be reasonable to do. Depending on the design of TFKey, we could always make this int[3] and ignore 'extra' nulls because they would appear in the value.

I used [the TLD class](https://github.com/NationalSecurityAgency/datawave/blob/master/warehouse/query-core/src/main/java/datawave/query/tld/TLD.java) as a blue print for this. It's probably worth tossing that into a micro benchmark and trying this implementation against the one you're suggesting.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/TermFrequencyList.java
apmoriarty: Good catch, making this change

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/PushdownNegationVisitorTest.java
apmoriarty: This is a fun one. The first form says that both 'bar' and 'baz' must be present to evaluate the function. If either term is missing then the entire content function cannot be evaluated, which in turn evaluates that branch of the query to true (candidate document was *not* excluded).

In the second form we need the second part to evaluate to true (!content:function || !termA || !termB). The absence of term B will force the entire subtree to a True. The presence of term B will force (NOT TRUE) into false, and we continue evaluating the union. So it should all work out in the end.

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyOffsetIterator.java
apmoriarty: It irks me that the `TermFrequencyIterator` has getters and setters before the main functions like `next`, `seek`, and `getNextSeekRange`. Going to ask forgiveness in advance for the nasty diff between the two classes and leave it as-is.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/DelayedFieldIndexIterator.java
apmoriarty: Dropping like a hot potato! Shame that the original idea didn't pan out, but on the bright side less code to be maintained.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/FieldValue.java
apmoriarty: If for some reason we need to flip back to this as opposed to the `TFKey` I would like to leave this in. We can always remove this later as part of a sweep for deprecated classes.

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyOffsetIterator.java
apmoriarty: Sure can.

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyOffsetIterator.java
apmoriarty: I'll clarify.

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyOffsetIterator.java
apmoriarty: Mhmm

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyOffsetIterator.java
apmoriarty: Done.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermFrequencyHitFunction.java
apmoriarty: Yes

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermFrequencyHitFunction.java
apmoriarty: 😄

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermFrequencyHitFunction.java
apmoriarty: This is a reasonable change, I'll make it so.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermFrequencyHitFunction.java
apmoriarty: Done.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermFrequencyHitFunction.java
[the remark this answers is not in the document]
apmoriarty: removed this iterator and comment.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/FieldValue.java
```diff
@@ -0,0 +1,131 @@
...
+        }
+        return false;
+    }
+    
+    @Override
+    public String toString() {
+        return getField() + " -> " + getValue();
+    }
+    
+    public static FieldValue getFieldValue(Key key) {
```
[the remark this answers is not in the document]
apmoriarty: Will do

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/FieldValue.java
```diff
@@ -0,0 +1,131 @@
...
+    public String getValueField() {
+        return valueField;
+    }
+    
+    public String getField() {
+        return valueField.substring(nullOffset + 1);
+    }
+    
+    public String getValue() {
+        return valueField.substring(0, nullOffset);
```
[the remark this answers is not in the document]
apmoriarty: This is one of the motivators for the `TFKey` class, modifying the `FieldValue` class is beyond the scope of this merge request

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/FieldValue.java
```diff
@@ -0,0 +1,131 @@
...
+            return bytes[index];
+        }
+    }
+    
+    public String getValueField() {
+        return valueField;
+    }
+    
+    public String getField() {
+        return valueField.substring(nullOffset + 1);
```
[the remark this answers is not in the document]
apmoriarty: Beyond scope

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/FieldValue.java
```diff
@@ -0,0 +1,131 @@
+package datawave.query.postprocessing.tf;
+
+import datawave.util.StringUtils;
+import org.apache.accumulo.core.data.Key;
+import org.apache.hadoop.io.Text;
+
+/**
+ * A field name and value which is sorted on {@code <value>\0<name>}
```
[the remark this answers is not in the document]
apmoriarty: Updated.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermFrequencyHitFunction.java
```diff
@@ -0,0 +1,689 @@
...
+    }
+    
+    public void setIsTld(boolean isTld) {
+        this.isTld = isTld;
+    }
+    
+    // used to perform a lazy init
+    private void initializeSource() throws IOException {
+        iter = new DelayedFieldIndexIterator();
+        iter.init(source, null, null);
```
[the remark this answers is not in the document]
apmoriarty: This is getting reworked, comment no longer applicable.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermFrequencyHitFunction.java
```diff
@@ -0,0 +1,689 @@
...
+    // Document built from any negated content function terms that hit
+    private Document negatedDocument;
+    
+    private SortedKeyValueIterator<Key,Value> source;
+    
+    // use this one weird trick to avoid an obnoxious number of source deep copies!
+    private DelayedFieldIndexIterator iter;
+    
+    // Determines how the seek range for the field index is built, i.e., should it consider child documents
+    private boolean isTld = false;
```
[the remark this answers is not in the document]
apmoriarty: The `isTld` flag determines how a range's end key is built and how negated functions are handled. I'm leaning towards leaving this as-is compared to creating a TldTermFrequencyHitFunction.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermFrequencyHitFunction.java
```diff
@@ -0,0 +1,689 @@
...
+        if (specialCase) {
+            // field(s) were not present in the function, lookup fields by value.
+            fields = lookupFieldsByValues(values);
+        } else {
+            // function's first arg is the field, or fields in the form (FIELD_A || FIELD_B)
+            fields = parseField(args.get(0));
+        }
+        
+        // delayed = negated
+        boolean isNegated = isDelayed(function);
```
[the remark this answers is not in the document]
apmoriarty: It may be more precise to say that delayed == not part of the nested iterator bridge construct. The IteratorBuildingVisitor stops at delayed markers, so if this function is under a delayed marker it's terms have not been fetched as part of the field index lookup -- thus requiring a fetch here.

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyOffsetIterator.java
```diff
@@ -0,0 +1,160 @@
...
+                break;
+            }
+            
+            if (searchSpace.contains(key.getColumnQualifier())) {
+                this.tk = key;
+                this.tv = source.getTopValue();
+                break;
+            } else {
+                // Always seek, for now.
+                seekToNextHit(key);
```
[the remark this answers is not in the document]
apmoriarty: Opened issue https://github.com/NationalSecurityAgency/datawave/issues/1131

warehouse/query-core/src/test/java/datawave/core/iterators/key/TFKeyTest.java
apmoriarty: Added.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermOffsetPopulator.java
apmoriarty: The TermFrequencyOffsetIterator only returns keys that match it's search space, so in theory it will never return a malformed key. A continue here *could* risk skipping a key, and thus a hit. Either way, an error will be logged for investigation. I think skipping a key is preferable to getting stuck in an infinite loop. Adding in a next call here.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermOffsetPopulator.java
apmoriarty: The searchSpace object is passed to the TermFrequencyOffsetIterator which makes use of the TreeSet's `higher(E)` method. It's convent to leave this as a TreeSet.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
apmoriarty: Delaying the second deep copy makes sense, especially given that it should only happen if a query contains negations.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TFFactory.java
apmoriarty: Makes sense to bundle all of these into a single object that gets passed around.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TFFactory.java
```diff
@@ -40,29 +43,34 @@
...
     /**
      * Factory method for creating the TF function used for generating the map context.
      * 
      * @param query
+     *            the query tree
      * @param dataTypes
      * @param sourceDeepCopy
+     *            a source copy used to fetch term frequencies
+     * @param secondSource
+     *            a source copy used to fetch delayed values from the field index (negated content function terms)
```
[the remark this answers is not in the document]
FineAndDandy: fix javadoc

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermFrequencyHitFunction.java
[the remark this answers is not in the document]
FineAndDandy: please update the comments variables to reflect this then

FineAndDandy (changes requested):
would just like to see the last few comments addressed

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermFrequencyHitFunction.java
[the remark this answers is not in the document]
apmoriarty: Updated in next commit.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermFrequencyHitFunction.java
apmoriarty: As far as I can tell, the args in the jexl node are always strings. I have yet to find a numeric that isn't represented by a string.

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyOffsetIterator.java
```diff
@@ -0,0 +1,163 @@
...
+    
+    private boolean seekToNextHit(Key key) throws IOException {
+        Range seekRange = getNextSeekRange(key);
+        if (seekRange == null)
+            return false;
+        source.seek(seekRange, seekCFs, true);
+        return true;
+    }
+    
+    private Range getNextSeekRange(Key key) {
```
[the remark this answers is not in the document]
apmoriarty: Addressed in next commit

thebikerdude (dismissed):

FineAndDandy (approved):

apmoriarty:
Rebased from `feature/accumulo2.0` to `integration`, squashed existing commits, and handled accumulo api changes

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyOffsetIterator.java
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+    
+    @Override
+    public void seek(Range range, Collection<ByteSequence> cfs, boolean inclusive) throws IOException {
+        this.initialRange = new Range(range);
+        this.tk = null;
+        this.tv = null;
+        
+        source.seek(range, seekCFs, true);
```
d-hwang, line 71: Should this pass in inclusive flag?

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermFrequencyHitFunction.java
```diff
@@ -0,0 +1,689 @@
...
+        } else {
+            // function's first arg is the field, or fields in the form (FIELD_A || FIELD_B)
+            fields = parseField(args.get(0));
+        }
+        
+        // nodes under a delayed marker are not part of the field index lookup and do not have values in the document. Values from a content function delayed
+        // for any reason must be fetched by this function.
+        boolean isNegated = isDelayed(function);
+        if (isNegated) {
+            this.hasNegatedFunctions = true;
```
d-hwang, line 590: Should this be named as "hasDelayedFunctions"? this.hasDelayedFunctions = isDelayed(function) ??

d-hwang (commented):
Looks good to me, but I had two minor comments.

d-hwang:
Looks like one of the unit tests is failing.

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyOffsetIterator.java
apmoriarty: This is safe to hardcode, we always want to restrict the scan to the term frequency column family ("tf")

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermFrequencyHitFunction.java
apmoriarty: Seems reasonable, updated variables and comment for clarity

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IteratorBuildingVisitorTest.java
```diff
@@ -37,6 +39,8 @@
 import java.util.Map;
 import java.util.Set;
 
+import static datawave.query.Constants.NULL;
```
FineAndDandy, line 42: you imported the static then typed the value everywhere

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IteratorBuildingVisitorTest.java
```diff
@@ -832,23 +732,48 @@ public void visitAnd_ExceededValueThresholdMarker_RegexLeadingWildcardNegationAl
...
+        ASTJexlScript script = JexlASTHelper.parseJexlQuery(query);
+        
+        Key hit = new Key("row", "dataType\u0000123.345.456");
+        
+        List<Map.Entry<Key,Value>> source = new ArrayList<>();
+        source.add(new SimpleEntry(new Key("row", "fi\u0000FOO", "few\u0000dataType\u0000123.345.456"), new Value()));
+        source.add(new SimpleEntry(new Key("row", "fi\u0000TEXT", "bar\u0000dataType\u0000123.345.456"), new Value()));
+        source.add(new SimpleEntry(new Key("row", "fi\u0000TEXT", "baz\u0000dataType\u0000123.345.456"), new Value()));
+        // source.add(new SimpleEntry(new Key("row", "tf", "dataType\u0000123.345.456\u0000bar\u0000TEXT"), new Value()));
+        // source.add(new SimpleEntry(new Key("row", "tf", "dataType\u0000123.345.456\u0000baz\u0000TEXT"), new Value()));
```
FineAndDandy: delete

FineAndDandy (approved):
I would like to see the constants fixed in the tests, but won't hold up a merge

d-hwang (approved):

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IteratorBuildingVisitorTest.java
apmoriarty: ha, never completed that thought. Fixed.

d-hwang (approved):

FineAndDandy (dismissed):

ivakegg (dismissed):

hlgp:
approved back in November.  please address conflicts so we can get this in

apmoriarty:
Closing in favor of https://github.com/NationalSecurityAgency/datawave/pull/1632. Same branch, different repo

PR #1060, opened 2021-02-05
re #855: Ensure we reduce the query after expanding ranges

ivakegg:


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPruningVisitor.java
```diff
@@ -253,11 +253,6 @@ public Object visit(ASTOrNode node, Object data) {
     
     @Override
     public Object visit(ASTAndNode node, Object data) {
-        // do not process query property markers
```
ivakegg, line 256: I removed this because I see no reason why we can't reduce under a delayed expression.  The ASTAssignment returns UNKNOWN below, so if we have a query property marker and the source is FALSE, then we should replace the whole thing with FALSE.

PR #1064, opened 2021-02-10
Support custom result limits by DN in Execution service

### LINKED ISSUE #1028: Update Execution service to modify result limit by DN
Similar to #1027, we should allow configuring query logics to allow different result limits based on DN.  Hence the BaseQueryLogic need a new member which maps DNs to result limits.  Then the query execution should take this into account (in RunningQuery I think) when applying the maxResults limit.

lbschanno:
Allow query logics to be configured with custom limits for specific DNs.
Update the execution service to then use these custom limits instead of
the default maxResults when applicable.

Fixes #1028

### DIFF
--- a/web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
+++ b/web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
@@ -1,17 +1,5 @@
 package datawave.webservice.query.runner;
 
-import java.security.Principal;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Set;
-import java.util.UUID;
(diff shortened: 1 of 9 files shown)


ivakegg (dismissed):

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
```diff
@@ -39,6 +26,18 @@
 import org.apache.log4j.Logger;
 import org.jboss.logging.NDC;
 
+import java.security.Principal;
```
ivakegg, line 29: Just so you know, every time you reorganize imports as such it inevitably causes me to have to resolve conflicts when merging up.

ivakegg (approved):

PR #1040, opened 2021-01-11
Fixes #1036

keith-ratcliffe:
- Updated Accumulo from 2.0 to 2.0.1
 - Fixed ClassCastExceptions thrown in Pushdown[Function|Scheduler]
 - Fixed ClassNotFoundException during webservice EAR deployment
 - Fixed Tserver FileNotFoundException on TypeMetadataBridgeContext.xml
 - Fixed CLASSPATH setup errors in some ingest scripts

PR #1065, opened 2021-02-11
Fix compile error -- BaseIteratorEnvironment class is gone in 2.0.

brianloss:
BaseIteratorEnvironment is gone in Accumulo 2.0, replaced instead with
default implementations on the IteratorEnvironment interface. Provide
an empty implementation of that interface here, to provide the same
functionality as BaseIteratorEnvironment used to.

Also, update to Accumulo 2.0.1 since it includes a CVE fix.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -30,7 +30,7 @@
         <maven.compiler.target>1.8</maven.compiler.target>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <surefire.forkCount>1C</surefire.forkCount>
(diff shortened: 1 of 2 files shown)


keith-ratcliffe (approved):
Thanks @brianloss 
[PR 1040](https://github.com/NationalSecurityAgency/datawave/pull/1040) addresses the remaining deployment related bugs on the master branch. Once both of these are merged, clean builds on master should resume

ivakegg (approved):

Issue #1073, opened 2021-02-18
Order of Multi-Fielded Filter Functions Should Not Matter


apmoriarty:
The order of fields within a multi-fielded filter function such as filter:includeRegex or filter:excludeRegex should not influence the number of results for a query. 

This bug is the result of how the DatawaveInterpreter attempts to short circuit an OR node during evaluation. The following test run from the FilterFunctionQueryTest will exercise this bug.

```    // Order of fields should not affect the number of results
    @Test
    public void testExerciseBugWithHowOrNodesAreHandled() throws Exception {
        String orig = "CITY == 'london' and filter:includeRegex(STATE||NUM,'110')";
        String next = "CITY == 'london' and (STATE =~ '110' or NUM =~ '110')";
        runTest(orig, next);

        orig = "CITY == 'london' and filter:includeRegex(NUM||STATE,'110')";
        next = "CITY == 'london' and (NUM =~ '110' or STATE =~ '110')";
        runTest(orig, next);
    }

apmoriarty:
Fixed in #1074

PR #1074, opened 2021-02-18
Fix bug in DatawaveInterpreter that caused incorrect resolution of filter functions

apmoriarty:
Addresses https://github.com/NationalSecurityAgency/datawave/issues/1073


apmoriarty:
Rebasing onto master. Issue was discovered in 2.9 but there's an issue with the quick start action.

PR #1075, opened 2021-02-18
Let's see about fixing the 2.9 quickstart...

apmoriarty:


keith-ratcliffe (approved):
Thanks @apmoriarty

tmp.txt
```diff
@@ -0,0 +1 @@
+this is a test.
```
keith-ratcliffe, line 1: assuming you'll remove this file before merge

keith-ratcliffe:
Hmm, tarballs are good now, but [namenode format](https://github.com/NationalSecurityAgency/datawave/blob/release/version2.9/contrib/datawave-quickstart/bin/services/hadoop/install.sh#L97) seems to be failing consistently, causing jvm core dump. Not sure why

keith-ratcliffe:
@apmoriarty, the quickstart for this PR deploys and tests cleanly on a local, non-docker install. However, I can replicate the same issue locally with the docker build, using our docker-build.sh script. There's a container-specific issue when running the namenode and datanode JVMs. Doesn't appear to affect any other JVMs, strangely

PR #1084, opened 2021-02-24
fixes #1070: Apply depth and termcount limits at the end of the VisitorFunction

### LINKED ISSUE #1070: Query Term limit should be checked after query expansion
The DefualtQueryPlanner only checks maxTermThreshold after the initial query parse. It should be rechecked after query expansion to ensure the expanded query does not violate limits.

FineAndDandy:


keith-ratcliffe:
Merging latest changes from target, as that should resolve the failed quickstart test

apmoriarty (approved):

jwomeara:
@ivakegg @FineAndDandy So... we did decide to use the same term threshold before and after expansion???

PR #1102, opened 2021-03-05
fixes #1101: Apply end key bounded by document range

### LINKED ISSUE #1101: IndexOnlyKeyToDocumentData should apply document range to endKey
The current range is created against only the row cf+'\uffff' which will result in scanning all values to the end of the cf starting with the document range, instead of ending on the row cf:cq\u0000\uffff

FineAndDandy:


apmoriarty (approved):
Fix looks good.

keith-ratcliffe (approved):
LGTM

PR #1103, opened 2021-03-09
fixes #1079: Attempt query term reduction in visitor function by combining values and ranges

### LINKED ISSUE #1079: Attempt to create Ivarators to collapse query expansion when beyond term threshold
If a query is beyond the term limit after query planning attempt to collapse values into ivarators to reduce the query term count below the threshold.

FineAndDandy:


jwomeara:
Please add some tests.  We have some coverage for PushdownLargeFieldedListsVisitor in ExceededOrThresholdMarkerJexlNodeTest.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownLargeFieldedListsVisitor.java
```diff
@@ -108,13 +149,18 @@ public Object visit(ASTOrNode node, Object data) {
...
+                skip = true;
+            }
             
             Collection<JexlNode> eqNodes = eqNodesByField.get(field);
             Collection<JexlNode> rangeNodes = rangeNodesByField.get(field);
             
             // if "_ANYFIELD_" or "_NOFIELD_", then simply add the subset back into the children list
             // if past our threshold, then add a ExceededValueThresholdMarker with an OR of this subset to the children list
-            if (!Constants.ANY_FIELD.equals(field)
+            if (!skip
```
jwomeara, line 162: Could you update the comment to describe the significance of the additional `skip` boolean?

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
```diff
@@ -215,6 +217,65 @@ public ScannerChunk apply(@Nullable ScannerChunk input) {
                         }
                     }
                     
+                    int termCount = TermCountingVisitor.countTerms(script);
```
jwomeara, line 220: If we're going to run TermCountingVisitor after we've run through pushdown, then we should update it to count each pushed down "_List_" as a single term, should we not?

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
```diff
@@ -215,6 +217,65 @@ public ScannerChunk apply(@Nullable ScannerChunk input) {
...
+                                        int toReduce = termCount - config.getMaxTermThreshold();
+                                        while (toReduce > 0) {
+                                            // get the highest value field out of the map
+                                            Integer reduction = pushdownCapacity.lastKey();
+                                            List<String> fields = pushdownCapacity.get(reduction);
+                                            
+                                            // take the first field
+                                            String field = fields.remove(0);
+                                            fieldsToReduce.add(field);
+                                            toReduce -= reduction;
```
jwomeara, line 261: If we're counting the list ivarator nodes as a single term, shouldn't this be something like:
```
toReduce -= (reduction-1);
```

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
```diff
@@ -215,6 +217,65 @@ public ScannerChunk apply(@Nullable ScannerChunk input) {
...
+                                    URI hdfsQueryCacheUri = getFstHdfsQueryCacheUri(config, settings);
+                                    SortedMap<Integer,List<String>> pushdownCapacity;
+                                    FileSystem fs = null;
+                                    String fstHdfsUri = null;
+                                    if (hdfsQueryCacheUri != null) {
+                                        fs = VisitorFunction.fileSystemCache.getFileSystem(hdfsQueryCacheUri);
+                                        fstHdfsUri = hdfsQueryCacheUri.toString();
+                                    }
+                                    
+                                    pushdownCapacity = PushdownLargeFieldedListsVisitor.getPushdownCapacity(reduceConfig, script, fs, fstHdfsUri);
```
jwomeara, line 241: Instead of calculating this on the second pass of the PushdownLargeFieldedListsVisitor, couldn't we just generate this on the first pass for those terms which didn't get combined into a list ivarator?  Might save some time.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownLargeFieldedListsVisitor.java
```diff
@@ -75,12 +80,48 @@ public PushdownLargeFieldedListsVisitor(ShardQueryConfiguration config, FileSyst
...
+        // invert and sort
+        SortedMap<Integer,List<String>> sortedMap = new TreeMap<>();
+        for (String fieldName : pushdownCapacity.keySet()) {
+            Integer reduction = pushdownCapacity.get(fieldName);
+            List<String> fields = sortedMap.get(reduction);
+            
+            if (fields == null) {
+                fields = new ArrayList<>();
+                sortedMap.put(reduction, fields);
+            }
```
jwomeara, line 119:
```suggest
            List<String> fields = sortedMap.computeIfAbsent(reduction, k -> new ArrayList<>());

            fields.add(fieldName);
```

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownLargeFieldedListsVisitor.java
```diff
@@ -187,6 +238,26 @@ public Object visit(ASTOrNode node, Object data) {
...
+     * @param field
+     * @param reduction
+     */
+    private void track(Object data, String field, int reduction) {
+        if (data instanceof Map) {
+            Map trackingMap = (Map) data;
+            Integer count = 0;
+            if (trackingMap.get(field) != null) {
+                count = (Integer) trackingMap.get(field);
+            }
```
jwomeara, line 254: IDE warnings are annoying...

```suggestion
            @SuppressWarnings("unchecked")
            Map<String,Integer> trackingMap = (Map<String,Integer>) data;
            Integer count = 0;
            if (trackingMap.get(field) != null) {
                count = trackingMap.get(field);
            }
```

jwomeara (changes requested):
I like this, but I think this could be simplified a bit, with less executions of the pushdown visitor.  Please add some tests as well.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownLargeFieldedListsVisitor.java
```diff
@@ -133,9 +179,11 @@ public Object visit(ASTOrNode node, Object data) {
                     if (fstHdfsUri != null && (eqNodes.size() >= config.getMaxOrExpansionFstThreshold())) {
                         URI fstPath = createFst(values);
                         markers.add(ExceededOrThresholdMarkerJexlNode.createFromFstURI(field, fstPath));
+                        track(data, field, eqNodes.size() - 1);
```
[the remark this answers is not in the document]
FineAndDandy: I like this idea. I think we also have to compute how much further a pushed down term could be reduced even if we push it down.

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
FineAndDandy: more generally TermCountingVisitor should count each ivarator as a single term?

FineAndDandy:
@jwomeara can you revisit this when you have a chance please?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownLargeFieldedListsVisitor.java
```diff
@@ -108,17 +116,20 @@ public Object visit(ASTOrNode node, Object data) {
...
+            boolean canReduce = (this.fields == null || this.fields.contains(field));
             
             Collection<JexlNode> eqNodes = eqNodesByField.get(field);
             Collection<JexlNode> rangeNodes = rangeNodesByField.get(field);
             
             // if "_ANYFIELD_" or "_NOFIELD_", then simply add the subset back into the children list
             // if past our threshold, then add a ExceededValueThresholdMarker with an OR of this subset to the children list
-            if (!Constants.ANY_FIELD.equals(field)
+            if (canReduce
+                            && !Constants.ANY_FIELD.equals(field)
```
apmoriarty, line 128: This is kind of ugly, can we put an expression per line and wrap with formatter:off/on?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownLargeFieldedListsVisitor.java
```diff
@@ -178,15 +189,40 @@ public Object visit(ASTOrNode node, Object data) {
...
         
         return children(newNode, children.toArray(new JexlNode[children.size()]));
     }
     
+    /**
+     * If data is set update it for the given field to include the new possible reduction, otherwise do nothing
+     * 
+     * @param data
+     * @param field
+     * @param reduction
```
apmoriarty, line 211: Define reduction -- I assume it's the number of nodes eliminated from the query via pushdown?

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
```diff
@@ -279,14 +284,71 @@ protected ASTJexlScript pushdownLargeFieldedLists(ShardQueryConfiguration config
...
+                // check if the capacity is available to get under the term limit
+                // determine if its possible to reduce enough to meet the threshold
+                int capacitySum = 0;
+                for (Integer capacity : pushdownCapacity.values()) {
+                    capacitySum += capacity;
+                }
+                
+                if (termCount - capacitySum <= config.getMaxTermThreshold()) {
+                    // copy the config and set minimum thresholds for collapsing Ors
+                    ShardQueryConfiguration reduceConfig = new ShardQueryConfiguration(config);
```
apmoriarty, line 309: There's a lot to copy in the ShardQueryConfig, would you be opposed to storing off the current thresholds, setting a new threshold an resetting to default values at the end of this block?

warehouse/query-core/src/test/java/datawave/query/tables/async/event/VisitorFunctionTest.java
```diff
@@ -0,0 +1,292 @@
...
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.UUID;
+
+public class VisitorFunctionTest extends EasyMockSupport {
```
apmoriarty, line 34: Hey thanks for adding a test for the VisitorFunction

warehouse/query-core/src/test/java/datawave/query/tables/async/event/VisitorFunctionTest.java
```diff
@@ -0,0 +1,292 @@
...
+    private VisitorFunction function;
+    
+    private ShardQueryConfiguration config;
+    private MetadataHelper helper;
+    
+    @Before
+    public void setup() throws MalformedURLException {
+        // config = createMock(ShardQueryConfiguration.class);
```
apmoriarty, line 42: remove commented line

warehouse/query-core/src/test/java/datawave/query/tables/async/event/VisitorFunctionTest.java
```diff
@@ -0,0 +1,292 @@
...
+        verifyAll();
+        
+        Assert.assertNotEquals(chunk, updatedChunk);
+        String updatedQuery = updatedChunk.getOptions().getIterators().iterator().next().getOptions().get(QueryOptions.QUERY);
+        Assert.assertNotEquals(query, updatedQuery);
+        Assert.assertTrue(updatedQuery, updatedQuery.contains("_List_"));
+        Assert.assertTrue(updatedQuery, updatedQuery.contains("field = 'FIELD1'"));
+        Assert.assertTrue(updatedQuery, updatedQuery.contains("ranges\":[[\"(a\",\"z)\"]"));
+    }
+}
```
apmoriarty, line 292: Is the addition of the VisitorFunctionTest sufficient or should a test be added for the PushdownLargeFieldedListsVisitor?

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -582,7 +582,7 @@ private void markQueryStarted(ShardQueryConfiguration config, Query settings) th
     }
     
     /**
-     * NOT THREAD SAFE -- relies on QueryStopwatch is not thread safe
+     * NOT THREAD SAFE - relies on QueryStopwatch which is not thread safe
```
apmoriarty, line 585: Grammar fixes make me unreasonably happy

apmoriarty (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownLargeFieldedListsVisitor.java
```diff
@@ -178,15 +189,43 @@ public Object visit(ASTOrNode node, Object data) {
             else {
                 // recurse on the eq children in this subset
                 copyChildren(eqNodes, children, data);
+                track(data, field, eqNodes.size() - 1);
                 
                 // recurse on the range children in this subset
                 copyChildren(rangeNodes, children, data);
+                
+                int numBatches = (int) Math.ceil(rangeNodes.size() / (double) Math.max(1, config.getMaxRangesPerRangeIvarator()));
+                numBatches = Math.min(Math.max(1, config.getMaxOrRangeIvarators()), numBatches);
```
jwomeara, line 198: Instead of copying this code from above, make a method.  That way if we make changes in the future, we only need to change one part of the code.

Alternatively, you could remove these lines, set ShardQueryconfiguration.maxOrRangeIvarators to 1 during your second pushdown visitor pass in the visitor function and just subtract 1 from rangeNodes.size() in the following track call.

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
```diff
@@ -279,14 +284,79 @@ protected ASTJexlScript pushdownLargeFieldedLists(ShardQueryConfiguration config
...
+            if (termCount > config.getMaxTermThreshold()) {
+                // check if the capacity is available to get under the term limit
+                // determine if its possible to reduce enough to meet the threshold
+                int capacitySum = 0;
+                for (Integer capacity : pushdownCapacity.values()) {
+                    capacitySum += capacity;
+                }
+                
+                if (termCount - capacitySum <= config.getMaxTermThreshold()) {
+                    // preserve the original config and set minimum thresholds for collapsing Ors, do not use FST for performance reasons
```
jwomeara, line 308: This comment strikes me as odd.  Technically we aren't doing anything programmatically to prevent FSTs from running.  If the FST threshold is set to 2, you will absolutely create an FST instead of a list ivarator.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownLargeFieldedListsVisitor.java
jwomeara: Could you add this in please?

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
jwomeara: Did you determine whether or not TermCountingVisitor counts List or FST ivarators in it's count?  I suspect it doesn't, and if it doesn't, we should update it since we're relying on it to give an accurate term count.

jwomeara (changes requested):
Requested a few additional changes.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownLargeFieldedListsVisitor.java
FineAndDandy: got it

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
FineAndDandy: yep, thanks

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
FineAndDandy: I missed this comment on the last pass. It does not currently count these as terms. I will update the visitor to catch all ivarators and count them as a term each

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownLargeFieldedListsVisitor.java
FineAndDandy: ok

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownLargeFieldedListsVisitor.java
FineAndDandy: ok

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownLargeFieldedListsVisitor.java
```diff
@@ -108,17 +116,26 @@ public Object visit(ASTOrNode node, Object data) {
...
+            // @formatter:off
+            if (canReduce &&
+                    !Constants.ANY_FIELD.equals(field) &&
+                    !Constants.NO_FIELD.equals(field) &&
+                    (eqNodes.size() >= config.getMaxOrExpansionFstThreshold() ||
+                            eqNodes.size() >= config.getMaxOrExpansionThreshold() ||
+                            rangeNodes.size() >= config.getMaxOrRangeThreshold()
+                    ) &&
+                    isIndexed(field)) {
+                // @formatter:on
```
jwomeara, line 136: This is a personal preference thing, and I certainly won't reject the PR for this, but my preference would be to rewrite this chunk as follows:
```suggestion
            Collection<JexlNode> eqNodes = eqNodesByField.get(field);
            Collection<JexlNode> rangeNodes = rangeNodesByField.get(field);
            
            // if fields is set and the current field is not in fields, then don't add an ExceededValueThresholdMarker
            // if "_ANYFIELD_" or "_NOFIELD_", then simply add the subset back into the children list
            // if past our threshold, then add a ExceededValueThresholdMarker with an OR of this subset to the children list
            // @formatter:off
            if ((this.fields == null || this.fields.contains(field)) &&
                !Constants.ANY_FIELD.equals(field) &&
                !Constants.NO_FIELD.equals(field) &&
                (eqNodes.size() >= config.getMaxOrExpansionFstThreshold() ||
                 eqNodes.size() >= config.getMaxOrExpansionThreshold() ||
                 rangeNodes.size() >= config.getMaxOrRangeThreshold()) &&
                isIndexed(field)) {
                // @formatter:on
```

I think this makes it easier to see what all of the conditions are at a glance, instead of having to scroll up to see what `canReduce` is set to.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/TermCountingVisitor.java
```diff
@@ -39,6 +34,8 @@ public Object visit(ASTAndNode node, Object data) {
         if (JexlASTHelper.findRange().isRange(node)) {
             // count each bounded range as 1
             ((MutableInt) data).increment();
+        } else if (QueryPropertyMarkerVisitor.isIvarator(node)) {
+            ((MutableInt) data).increment();
```
jwomeara, line 38: With your `isIvarator` addition, ranges are already covered, so we don't need the first condition anymore.

```suggestion
        if (QueryPropertyMarkerVisitor.isIvarator(node)) {
            // count everything that ivarates as 1
            ((MutableInt) data).increment();
```

jwomeara (changes requested):
Term counting visitor can be simplified with your recent addition.  I also added a suggestion for the conditional in `PushdownLargeFieldedListsVisitor.java`

jwomeara (approved):
LGTM

apmoriarty (approved):
LGTM

PR #1116, opened 2021-03-16
WIP:  Datawave 825 IndexColumnIterator extends TypedValueCombiner

jzgithub1:
I have to call the fuction setColumns in the init() function of the updated iterator which is now a TypedValueCombiner now.  @ivakegg can you give me some pointers here.  I don't have an example to go from in the datawave code base.  Thanks.  Compile and run IndexColumnIteratorTest to see what happens.


warehouse/ingest-core/src/main/java/datawave/iterators/IndexColumnIterator.java
```diff
@@ -0,0 +1,108 @@
...
+    public IndexColumnIterator() {}
+    
+    public IndexColumnIterator(IndexColumnIterator aThis, IteratorEnvironment environment) {
+        super();
+        setSource(aThis.getSource().deepCopy(environment));
+    }
+    
+    @Override
+    public void init(SortedKeyValueIterator<Key,Value> source, Map<String,String> options, IteratorEnvironment env) throws IOException {
+        //TODO implement the function call below.
```
ivakegg, line 29: The columns are set by using the "columns" option.  See the Combiner class.

warehouse/ingest-core/src/main/java/datawave/iterators/IndexColumnIterator.java
```diff
@@ -0,0 +1,108 @@
...
+                // TODO Just deserialize with creating a Value object
+                return IndexedDatesValue.deserialize(new Value(bytes));
+        }
+    }
+    
+    @Override
+    public IteratorOptions describeOptions() {
+        IteratorOptions io = super.describeOptions();
+        io.addNamedOption("ageOffDate", "Indexed dates before this date are not aggregated.");
+        io.setName("ageOffDate");
```
ivakegg, line 90: Why call setName and setDescirption?  addNamedOption does everything that was required here.

warehouse/ingest-core/src/main/java/datawave/iterators/IndexColumnIterator.java
```diff
@@ -0,0 +1,108 @@
...
+    public IndexedDatesValue typedReduce(Key key, Iterator<IndexedDatesValue> iterator) {
+        
+        IndexedDatesValue indexedDatesValue = null;
+        
+        if (log.isTraceEnabled())
+            log.trace("IndexColumnIterator combining dates for range for key " + key.getRow().toString(), new Exception());
+        
+        while (iterator.hasNext()) {
+            indexedDatesValue = iterator.next();
+
```
ivakegg, line 58: missing the code in here that actually combines the values....

warehouse/ingest-core/src/main/java/datawave/iterators/IndexColumnIterator.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+        
+        @Override
+        public IndexedDatesValue decode(byte[] bytes) throws ValueFormatException {
+            
+            if (bytes == null || bytes.length == 0)
+                return new IndexedDatesValue();
+            else
+                // TODO Just deserialize with creating a Value object
+                return IndexedDatesValue.deserialize(new Value(bytes));
```
ivakegg, line 82: Do we really need to wrap the bytes in a Value object to deserialize?

hlgp:
OBE by #2094, etc

PR #1122, opened 2021-03-22
Fixes #1121 Use columnVisibility from query to populate query metric

### LINKED ISSUE #1121: Use columnVisibility from query to populate query metric
Query metrics should be stored using the same columnVisibility as the query that they are tracking

billoley:
Correctly implement markings and visibility

Fixes #1121 Use columnVisibility from query to populate query metric


billoley:
Many of the changes in the test class are because it originally had the expected & actual parameters to Assert backwards.

keith-ratcliffe (dismissed):

web-services/client/src/main/java/datawave/webservice/query/metric/BaseQueryMetric.java
```diff
@@ -905,6 +908,10 @@ public String getColumnVisibility() {
     
     public void setColumnVisibility(String columnVisibility) {
         this.columnVisibility = columnVisibility;
```
ivakegg, line 910: I we are putting this in the markings, do we still nee the columnVisibility member?

ivakegg (changes requested):

web-services/client/src/main/java/datawave/webservice/query/metric/BaseQueryMetric.java
billoley: Decided that the best solution was to leave the setter and getter for markings but not the settings member.  The columnVisibility will hold the value and will be used to polulate a new Map when getMarkings() is called.  Also added serialization tests for XML, JSON, protobuf.

keith-ratcliffe (approved):

PR #1124, opened 2021-03-24
fixes #1123: Fixed the parsing and generation of hit terms

### LINKED ISSUE #1123: Hit terms are sometimes missing and may be parsed incorrectly
The LimitFields mechanism when gathering the hit terms will tokenize on ':' and then only use the first two tokens.  If the value contains a ':' then only a partial value will result.  This is even worse if the result is to be parse (e.g. a date) and results in a value that cannot be parsed.
In addition, the hit terms are missing for bounded ranges, and large or lists.

ivakegg:
* Updated LimitFields to correctly parse hit term values
* Updated DatawaveInterpreter to correctly generate hit terms for bounded ranges and large or lists
* Added tests which demonstrated problems and now ensure fixes work


apmoriarty (approved):
LGTM

hlgp (approved):

PR #1125, opened 2021-03-26
Lazily create ivarator cache dirs

### LINKED ISSUE #1059: The ivarator cache dirs are precreating directories even if they are not used
The ivarator cache dirs configured on the QueryIterator will result in a directory being created in every one of them even if the directory is not actually used for HdfsBackedSortedSets.  We should avoid creating directories that are not needed and even avoid creating the FileSystem object if we never use that ivarator cache dir.  Hence the FileSystem creation and dir creation should be done as lazily as possible.  There are two known places where we pre-create the directories and create the FileSystem objects:

1) QueryIterator.pruneIvaratorCacheDirs()
2) IteratorBuildingVisitor.getIvaratorCacheDirs()

This should be delayed to where we actually need to create usable directories:
1) BufferedFileBackedSortedSet.persist()
2) BufferedFileBackedSortedSet.compact()

The FileSystem object should not be created until an HdfsBackedSortedSet actually needs them in its constructor.

lbschanno:
The ivarator cache directories as currently managed result in
directories being created for each QueryIterator even if they are not
actually used for HdfsBackedSortedSets.

Modify this so that the cache directories are lazily created only when
persisting data to the file system.

Note for reviewers, I modified `QueryIterator.pruneIvaratorCacheDirs()` to only remove configurations that are invalid. If there are any additional checks that should be added to this in light of the removal of actually creating the directories, please let me know.

Closes #1059

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
+++ b/warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
@@ -277,27 +277,9 @@ public class QueryIterator extends QueryOptions implements YieldingKeyValueItera
         pruneIvaratorCacheDirs();
     }
     
-    // this method will prune any ivarator cache directories that are not available on this node
-    private void pruneIvaratorCacheDirs() throws IOException {
-        ivaratorCacheDirConfigs.removeIf(this::pruneIvaratorCacheDir);
-    }
-    
-    private boolean pruneIvaratorCacheDir(IvaratorCacheDirConfig config) {
-        boolean fsExists = false;
-        
-        // first, make sure the cache configuration is valid
-        if (config.isValid()) {
-            Path basePath = new Path(config.getBasePathURI());
-            
-            try {
-                FileSystem fs = this.getFileSystemCache().getFileSystem(basePath.toUri());
-                fsExists = fs.mkdirs(basePath) || fs.exists(basePath);
-            } catch (Exception e) {
-                log.debug("Ivarator Cache Dir does not exist: " + basePath);
-            }
-        }
-        
-        return !fsExists;
+    // this method will prune any ivarator cache directories that do not have a valid configuration.
+    private void pruneIvaratorCacheDirs() {
+        ivaratorCacheDirConfigs.removeIf(config -> !config.isValid());
     }
     
     @Override
--- a/warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
+++ b/warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
@@ -96,13 +96,16 @@ import org.apache.commons.jexl2.parser.JexlNode;
 import org.apache.commons.lang.builder.CompareToBuilder;
 import org.apache.commons.pool.BasePoolableObjectFactory;
 import org.apache.commons.pool.impl.GenericObjectPool;
+import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.permission.FsAction;
 import org.apache.hadoop.io.Text;
 import org.apache.log4j.Logger;
 import org.apache.zookeeper.server.quorum.QuorumPeerConfig.ConfigException;
 
 import javax.annotation.Nullable;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.util.Collection;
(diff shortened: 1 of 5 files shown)


warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
```diff
@@ -112,13 +113,18 @@ public void persist() throws IOException {
             // go through the handler factories and try to persist the sorted set
             for (int i = 0; i < handlerFactories.size() && !buffer.isPersisted(); i++) {
                 SortedSetFileHandlerFactory handlerFactory = handlerFactories.get(i);
+                
+                // Lazily create the required directories that the files will be written to.
+                // handlerFactory.ensureDirsCreated();
```
apmoriarty, line 118: unused comment?

warehouse/query-core/src/test/java/datawave/query/iterator/QueryIteratorIT.java
```diff
@@ -557,17 +557,6 @@ public void event_leadingRegex_shardRange_secondEvent_test() throws IOException
         event_test(seekRange, query, false, null, addEvent(11, "123.345.457"), Arrays.asList(getBaseExpectedEvent("123.345.457")));
     }
     
-    @Test(expected = IOException.class)
```
apmoriarty, line 560: Why is this test being removed? Is there still a way to test an invalid config? For example, what if every configured ivarator directory is invalid? (instead of hdfs://blah it's hddfs://blah?). In that case everything should prune to zero, and we should get an appropriate failure message.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
```diff
@@ -1032,20 +1032,6 @@ public Object visit(ASTNumberLiteral node, Object o) {
...
-            throw new IOException("Unable to load hadoop configuration", e);
-        } catch (Exception e) {
-            log.warn("Unable to access " + path, e);
-            return false;
-        }
-        return true;
-    }
-    
     /**
      * Create a cache directory path for a specified regex node. If alternatives have been specified, then random alternatives will be attempted until one is
```
apmoriarty, line 1036: I think this comment needs to be updated. This method is no longer creating new directories if they do not exist, it is now building up a list of potential hdfs directories based on the config.

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
```diff
@@ -112,13 +113,18 @@ public void persist() throws IOException {
...
+                
                 SortedSetFileHandler handler = createFileHandler(handlerFactory);
                 
                 // if we have a valid handler, try to persist
                 if (handler != null) {
                     Exception cause = null;
                     for (int attempts = 0; attempts <= numRetries && !buffer.isPersisted(); attempts++) {
                         try {
+
```
apmoriarty, line 127: Remove additional line?

apmoriarty (changes requested):
Some general comments/question, overall very excited for this change

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
lbschanno: Fixed.

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
lbschanno: Fixed.

warehouse/query-core/src/test/java/datawave/query/iterator/QueryIteratorIT.java
lbschanno: Added back unit test and modified it to fail when the config is invalid.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
lbschanno: Comment has been updated.

apmoriarty (dismissed):
LGTM

warehouse/query-core/src/main/java/datawave/query/util/sortedset/HdfsBackedSortedSet.java
```diff
@@ -162,16 +165,42 @@ public boolean isValid() {
...
+            }
+        }
+        
+        private void ensureCreation(Path path) throws IOException {
+            try {
+                FileSystem fs = getFs();
+                if (!fs.exists(path)) {
+                    // Attempt to create the required directory if it does not exist.
+                    if (!fs.mkdirs(path)) {
+                        throw new IOException("Unable to mkdirs: fs.mkdir(" + path + ")->false");
```
ivakegg, line 193: If we cannot create the directory for whatever reason, the we need to be able to try the next directory in the configuration.

lbschanno: @ivakegg could you elaborate more on this? I'm not sure what you mean by the next directory. I see that `IvaratorCacheDirConfig` has a basePathURI. Are you referring to trying to create multiple unique paths?

ivakegg: Looking through the code it appears that the logic presumes that we can create the directory on the first valid ivarator cache dir config anyway (there can be multiple set on the QueryIterator).  I take back this comment.  I think we are good.

ivakegg (approved):

warehouse/query-core/src/main/java/datawave/query/util/sortedset/HdfsBackedSortedSet.java
lbschanno: Okay, great!

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -277,27 +277,9 @@ public void init(SortedKeyValueIterator<Key,Value> source, Map<String,String> op
...
-                fsExists = fs.mkdirs(basePath) || fs.exists(basePath);
-            } catch (Exception e) {
-                log.debug("Ivarator Cache Dir does not exist: " + basePath);
-            }
-        }
-        
-        return !fsExists;
+    // this method will prune any ivarator cache directories that do not have a valid configuration.
+    private void pruneIvaratorCacheDirs() {
+        ivaratorCacheDirConfigs.removeIf(config -> !config.isValid());
```
ivakegg, line 282: Is there anyway to verify a volume is writable without actually writing to it?

ivakegg: FileSystem.access or FileSystem.getAclStatus perhaps

ivakegg (changes requested):
need to at least check for writability

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
ivakegg: Also once we find the first cache directory that is valid and writable, there is no need to check the other cache directories.  This way we can avoid excessive HDFS namenode calls which was the original problem that produced this ticket.  I apologize that these extra requirements were not stated up front.  I had not thought about the problem enough at the time.

lbschanno: I will try figuring something out to test if a volume is writable.

With regards to your latest comment, just to make sure I understand, do you want the remaining entries in ivaratorCacheDirConfigs removed after we find the first valid and writable cache directory>

lbschanno: @ivakegg can it be assumed that if the base path of a config doesn't exist, its parent does, or its parent is root?

ivakegg: I think we can simply prune out the remaining ones once we find one that is valid and writable.
I am not sure we can assume the parent path exists.  The first ancestor that exists must be writable to allow this to work.

apmoriarty (approved):

ivakegg (approved):

PR #1132, opened 2021-03-31
issue 1062: Allow GROUP_BY to work on multi-value fields 

### LINKED ISSUE #1062: Allow GROUP_BY to work on multi-value fields
The function #GROUP_BY (f:groupby) only works on fields that contain a single value.  However it is desirable that this function work with multk-value fields as well.  Bunus points  if you give the user the option to either iterate over the values or string them together as a single value (the default should be to iterator).

friedlou:
fixes #1062 

Force include.grouping.context to true when using the groupby function.


d-hwang:
Hi, we are expected to handle multi-value fields even without grouping-context. Could you please test these different cases?
1) FOO=1[visibilityA],FOO=1[visbilityB]
2) FOO=1[visibilityA],FOO=2[visibilityA]

I believe you should be able to use GroupingTest class. Thanks!

friedlou:
@d-hwang: Please look at the latest commit. It should solve the issue you brought up in your comment.

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
```diff
@@ -569,4 +591,8 @@ public int hashCode() {
             return hcb.toHashCode();
         }
     }
+
```
apmoriarty, line 594: Remove this please

apmoriarty (changes requested):
LGTM once commented lines are removed

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
friedlou: Done.

apmoriarty: Please put the hash code function back and remove the commented out lines (now lines 588 to 590). Apologies for not being precise

friedlou: Done in last commit.

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
```diff
@@ -22,6 +22,7 @@
 import org.apache.commons.lang.builder.HashCodeBuilder;
 import org.slf4j.Logger;
 import org.springframework.util.Assert;
+import sun.reflect.generics.tree.Tree;
```
apmoriarty, line 25: Unused import?

friedlou: Removed

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
```diff
@@ -362,10 +362,31 @@ private void flatten(List<Document> documents) {
             if (this.groupFieldsSet.contains(shorterName)) {
                 expandedGroupFieldsList.add(shortName);
                 log.trace("{} contains {}", this.groupFieldsSet, shorterName);
-                GroupingTypeAttribute<?> created = makeGroupingTypeAttribute(shortName, field.getData());
-                created.setColumnVisibility(field.getColumnVisibility());
-                fieldMap.put(fieldName, created);
-                fieldToFieldWithContextMap.put(shortName, fieldName);
+                
+                if (field.getData() instanceof Collection<?>) {
+                    // This handles multi-valued entries that do not have gouping context
```
jschmidt10:
```suggestion
                    // This handles multi-valued entries that do not have grouping context
```

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
```diff
@@ -362,10 +362,31 @@ private void flatten(List<Document> documents) {
...
+                    // Create GroupingTypeAttribute and put in ordered map ordered on the attribute type
+                    Multimap<Type<?>,GroupingTypeAttribute<?>> attrSortedMap = TreeMultimap.create();
```
jschmidt10: Are we relying on the sorting here? Or does it matter? If we're relying on sorting for correctness, we should declare this as a `SortedMultimap<Type<?>,GroupingTypeAttribute<?>>` to prevent someone from changing this to a HashMultimap for instance.

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
```diff
@@ -362,10 +362,31 @@ private void flatten(List<Document> documents) {
...
+                    for (Object typeAttribute : ((Collection<?>) field.getData())) {
+                        Type<?> type = ((TypeAttribute<?>) typeAttribute).getType();
```
jschmidt10, line 373: If we use `getData()` here instead of `getType()`, then it'll directly parallel the case below when we only have a single value. It took me a bit of digging to realize those two methods are actually the same so I was wondering why it'd be different for a multi-valued value vs a single value.

warehouse/query-core/src/test/java/datawave/query/transformer/GroupingTest.java
```diff
@@ -385,10 +382,60 @@ public void testGrouping4() throws Exception {
...
+        
+        Date startDate = format.parse("20091231");
+        Date endDate = format.parse("20150101");
+        
+        String queryString = "UUID =~ '^[CS].*'";
+        
+        Map<String,Integer> expectedMap = ImmutableMap.of("1", 3, "2", 3, "3", 1);
+        
+        extraParameters.put("group.fields", "RECORD");
+        // extraParameters.put("group.fields.batch.size", "12");
```
jschmidt10: We can delete the commented out code.

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
friedlou: getData() gets me an Object. At this point in the code I need a Type. That is why I used getType(). In the single value, all I have is an Attribute, so all I have is getData(), I think it should stay as is.

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
friedlou: Will change it.

warehouse/query-core/src/test/java/datawave/query/transformer/GroupingTest.java
friedlou: Will remove it.

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
friedlou: Will change it

jschmidt10 (approved):

apmoriarty (approved):

PR #1143, opened 2021-04-09
Bugfix/max expansion regex query test

ivakegg:


plainolneesh (approved):

warehouse/query-core/src/main/java/datawave/query/tables/BatchScannerSession.java
```diff
@@ -541,10 +542,22 @@ public void onSuccess(Scan finishedScan) {
...
+        } else {
+            log.error("BatchScanerSession failed", t);
+        }
         uncaughtExceptionHandler.uncaughtException(Thread.currentThread().currentThread(), t);
+        stopAsync();
         Throwables.propagate(t);
-        
+    }
+    
+    private boolean isInterruptedException(Throwable t) {
```
keith-ratcliffe, line 555: Would it be worthwhile to introduce commons lang dependency on ExceptionUtils for this instead?

ivakegg: I will check it out

ivakegg: Actually the exception is not always the root cause, so I think I am going to stick with what I have.

PR #1144, opened 2021-04-12
Updated GeoWave range optimization to not rely on byteArrayRange.isSi…

jwomeara:
…ngleValue since it does not return the expected result.


keith-ratcliffe (dismissed):
LGTM

PR #1146, opened 2021-04-13
Misc fixes: support custom BASE_WORK_DIR, full table scan

brianloss:
* If BASE_WORK_DIR is set in ingest-env.sh, its value is not picked up
  for ingest jobs since the job cache environment is sourced before
  ingest-env and therefore BASE_WORK_DIR is its default value. Move the
  block that sources job-cache-env after sourcing ingest-libs (which
  pulls in ingest-env) to ensure that BASE_WORK_DIR has the right value
  if it has been customized.

* If full table scan is set to true in BaseEventQuery, then facet query
  reports an error when listing query logic (and probably when used)
  indicating it doesn't support full table scan. Explicitly disable
  full table scan for facet query so it doesn't break if full table scan
  is enabled in BaseEventQuery.

Signed-off-by: Brian Loss <brianloss@gmail.com>


ivakegg (approved):

jwomeara (approved):
LGTM

keith-ratcliffe (approved):

brianloss:
Thanks @keith-ratcliffe! I pulled in the main branch. Can you merge once tests pass, please? Will also need to do the same for  #1148 and #1149.

PR #1147, opened 2021-04-13
Improve load-job-cache performance

brianloss:
* Use the hadoop copyFromLocal command instead of xargs with a bunch
  of hdfs put calls. The copyFromLocal command accepts a number of
  threads to use and runs quickly. This is used to copy the entire job
  cache dir hierarchy at once. This works better with systems such as
  ADLS Gen 2 where a large number of simultaneous requests could lead
  to throttling.

* Only set the replication on the job cache files if the filesystem is
  hdfs. Others such as ABFS or local either don't support setting the
  replication at all, or it is otherwise unnecessary.

* Skip dotfiles in check-job-cache.sh check


keith-ratcliffe (approved):
LGTM

jwomeara (approved):
LGTM

brianloss:
Thanks @jwomeara! I updated the branch. Can I trouble you to merge this once the checks have passed? I don't have permission to do so.

PR #1148, opened 2021-04-13
Prevent int overflow when loading large JSON files

brianloss:
JsonRecordReader uses a CountingInputStream to track the position in the
file. That class provides a count that is an int, and a count that is a
long. Switch to the count using the long value. Without this, if one
attempted to load a large JSON file where the count reported by the
stream would exceed Integer.MAX_VALUE, ingest would fail since at that
point retrieving the count will throw an exception.


apmoriarty (approved):

jwomeara (approved):
LGTM

brianloss:
@jwomeara or @keith-ratcliffe, could you merge this one too, please? Thanks!

PR #1149, opened 2021-04-13
Send memory requests to Yarn

brianloss:
* Currently, the bulk-ingest script sets the VM heap size, but does not
  make any memory request to Yarn, which means the job will use whatever
  the default configuration is for map and reduce memory requests. This
  can lead to the scenario where the VM heap is larger than the Yarn
  memory request, and Yarn will then schedule more tasks than a VM can
  hold, leading to OOM errors.

* Update the quickstart yarn nodemanager memory. With the above change,
  CI jobs would fail because the scheduler wouldn't be able to find any
  space to run the ingest jobs given the default memory config. CI runs
  in Azure on a 7GiB VM, so indicating 6GiB available should be fine.

* While debugging the above, I decided to update the CI configuration to
  handle the new cache action, change run params to match Accumulo's CI
  (better way to not log file transfers, enable ANSI color for maven
  output even in CI) and to no install anything in the local repo (so
  it doesn't get cached). There is generally not a need to use the
  install maven goal except maybe when performing a release build.
  Otherwise, package or verify will suffice.

Signed-off-by: Brian Loss <brianloss@gmail.com>


jwomeara (approved):
LGTM

keith-ratcliffe (approved):

Issue #1150, opened 2021-04-13
Clean up orphaned partial bulk load directories


sbgift-dev:

sbgift-dev:
Needed to check writeDirectlyToDest to determine which fs to delete the workdir from

SethSmucker:
Done, microservices are on Spring Boot 2.7.1 now. Closing.

PR #1153, opened 2021-04-19
Cleanup orphaned partial bulk load directories (#1150)

sbgift-dev:


warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
```diff
@@ -338,6 +338,12 @@ public int run(String[] args) throws Exception {
                 configureBulkPartitionerAndOutputFormatter(job, cbHelper, conf, outputFs);
             } catch (Exception e) {
                 log.error(e);
+                log.error("Deleting orphaned directory: " + workDirPath);
```
jschmidt10, line 341: I'm wondering if this should be an INFO since this line itself is not an error. Usually when you see ERROR in the log, it's something you need to investigate (like the error logged above). This is more just telling us we're cleaning up due to a previous error.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
```diff
@@ -1300,18 +1306,18 @@ protected void distCpDirectory(Path workDir, FileSystem src, FileSystem dest, Co
...
-            .withCopyStrategy(distCpStrategy)
-            .withSyncFolder(true)
-            .preserve(DistCpOptions.FileAttribute.USER)
-            .preserve(DistCpOptions.FileAttribute.GROUP)
-            .preserve(DistCpOptions.FileAttribute.PERMISSION)
-            .preserve(DistCpOptions.FileAttribute.BLOCKSIZE)
-            .preserve(DistCpOptions.FileAttribute.CHECKSUMTYPE)
-            .withBlocking(true)
-            .build();
+                .withLogPath(logPath)
```
jschmidt10, line 1309: Did you reformat this intentionally? If so, that's cool. If this was due to an IDE auto-format - I would put it back, otherwise, these lines will keep getting  ping-ponged back and forth between the two styles.

sbgift-dev: Yeah must've been an auto-format, oops

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
```diff
@@ -338,6 +338,12 @@ public int run(String[] args) throws Exception {
                 configureBulkPartitionerAndOutputFormatter(job, cbHelper, conf, outputFs);
             } catch (Exception e) {
                 log.error(e);
+                log.info("Deleting orphaned directory: " + workDirPath);
+                try {
+                    inputFs.delete(workDirPath, true);
+                } catch (Exception er) {
+                    log.error("Unable to remove directory: " + workDirPath);
```
jschmidt10, line 345: Sorry, missed this on the first pass. Mind if we add the er to the log? That'll show us the "Caused By:" stacktrace.

```suggestion
                    log.error("Unable to remove directory: " + workDirPath, er);
```

jschmidt10 (approved):

apmoriarty (approved):

ivakegg (approved):

PR #1154, opened 2021-04-20
Make TreeEqualityVisitor accept JexlNodes for comparison

### LINKED ISSUE #914: TreeEquality-like visitor to support sub-tree evaluation.
While reviewing #910 to fix #883 it was noted that more work had to be done by the developer to utilize the TreeEqualityVisitor. Currently that visitor only accepts two ASTJexlScripts as an argument.

A quick hack of the TreeEqualityVisitor indicated that handling reference expressions might be a problem. 

Either add a new method to the TreeEqualityVisitor that accepts jexl nodes, or create a SubTreeEqualityVisitor that handles sub tree equality.

lbschanno:
The TreeEqualityVisitor currently only accepts two ASTJexlAcripts as
arguments for comparison. This forces developers to do more work to
utilize the TreeEqualityVisitor in tests.

Make the following changes:
* Modify the TreeEqualityVisitor to accept JexlNodes instead of
  ASTJexlScripts.
* Add a static TreeEqualityVisitor.isEqual method that does not require
  an instantiated Reason to be passed in.
* Add field equal to TreeEqualityVisitor.Reason so that the Reason class
  may be returned and used both to assert equality and to get the reason
  why two trees may not be equal.
* Add unit tests for the TreeEqualityVisitor.

Fixes #914

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ConjunctionEliminationVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ConjunctionEliminationVisitor.java
@@ -225,7 +225,7 @@ public class ConjunctionEliminationVisitor extends RebuildingVisitor {
     // Return whether or not the two JEXL queries are equivalent.
     private boolean isEquivalent(JexlNode node, ASTJexlScript script) throws ParseException {
         ASTJexlScript nodeScript = getScript(node);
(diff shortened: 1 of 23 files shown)


apmoriarty (approved):
LGTM, really like these changes.

FineAndDandy (approved):

hlgp (approved):

PR #1157, opened 2021-04-22
 Add timestamped job cache for deletion purposes

cj2543:
This pull request will add functionality for loading/deleting job caches with a timestamp. This will not replace the ingest job cache loading because of concerns about the frequency it hits zookeeper(locking mechanism). This will be used primarily for deletion jobs that run less frequently.

### DIFF
--- a/warehouse/common/src/test/java/datawave/common/io/FilesFinderTest.java
+++ b/warehouse/common/src/test/java/datawave/common/io/FilesFinderTest.java
@@ -1,0 +1,121 @@
+package datawave.common.io;
+
+import com.google.common.io.Files;
+
+import datawave.common.test.utils.FileUtils;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.StringJoiner;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+import java.util.stream.Stream;
+
+import static datawave.common.test.utils.FileUtils.createTemporaryFile;
+
+public class FilesFinderTest {
+    private static final String TEMP_DIR = Files.createTempDir().getAbsolutePath();
+    private static final String CLASSPATH_DELIMITER = ":";
+    private static final String FILE_PREFIX = "File";
+    
+    private static String TEST_LEVEL0_PROP_FILE;
+    private static String TEST_LEVEL1_JAR_FILE;
+    private static String TEST_LEVEL1_XML_FILE;
+    private static String TEST_LEVEL2_XML_FILE;
+    
+    @BeforeClass
+    public static void setup() throws IOException {
+        TEST_LEVEL0_PROP_FILE = createTemporaryFile(TEMP_DIR, 0, FILE_PREFIX, "properties");
+        TEST_LEVEL1_JAR_FILE = createTemporaryFile(TEMP_DIR, 1, FILE_PREFIX, "jar");
+        TEST_LEVEL1_XML_FILE = createTemporaryFile(TEMP_DIR, 1, FILE_PREFIX, "xml");
+        TEST_LEVEL2_XML_FILE = createTemporaryFile(TEMP_DIR, 2, FILE_PREFIX, "xml");
+    }
+    
+    @AfterClass
+    public static void tearDown() throws IOException {
+        FileUtils.deletePath(TEMP_DIR);
+    }
+    
+    @Test
+    public void testGetFilesFromClasspath() {
+        Collection<String> expectedFiles = sortedList(TEST_LEVEL1_XML_FILE, TEST_LEVEL2_XML_FILE);
+        StringJoiner joiner = new StringJoiner(CLASSPATH_DELIMITER);
+        expectedFiles.forEach(joiner::add);
+        
+        Collection<String> filesFound = FilesFinder.getFilesFromClasspath(joiner.toString(), CLASSPATH_DELIMITER);
+        Assert.assertEquals(expectedFiles, filesFound);
+    }
+    
+    @Test
+    public void testGetFilesFromClasspathWithRelativePaths() {
+        Collection<String> expectedFiles = sortedList(TEST_LEVEL0_PROP_FILE, TEST_LEVEL1_JAR_FILE);
+        String baseDir = TEMP_DIR + "/1/2/";
+        String testPaths = baseDir + "../../" + FILE_PREFIX + "0.properties:" + baseDir + "../../1/" + FILE_PREFIX + "1.jar";
+        
+        Collection<String> filesFound = FilesFinder.getFilesFromClasspath(testPaths, CLASSPATH_DELIMITER);
+        Assert.assertEquals(expectedFiles, filesFound);
+    }
+    
+    @Test(expected = IllegalArgumentException.class)
+    public void testGetFilesFromClasspathWithRelativePathFailures() {
+        String testPaths = "../../" + FILE_PREFIX + "0.properties:../../1/" + FILE_PREFIX + "1.jar";
+        FilesFinder.getFilesFromClasspath(testPaths, CLASSPATH_DELIMITER);
+    }
+    
+    @Test
+    public void testGetFilesFromPattern() throws IOException {
+        Collection<String> expectedFiles = sortedList(TEST_LEVEL1_XML_FILE, TEST_LEVEL2_XML_FILE);
+        Collection<String> filesFound = FilesFinder.getFilesFromPattern(TEMP_DIR, "**.xml", Integer.MAX_VALUE);
+        Assert.assertEquals(expectedFiles, filesFound);
+    }
+    
+    @Test
+    public void testGetFilesFromPatternByDepth() throws IOException {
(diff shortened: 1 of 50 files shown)


warehouse/common/src/test/java/datawave/common/io/FilesFinderTest.java
```diff
@@ -0,0 +1,115 @@
...
+    public static void setup() throws IOException {
+        TEST_LEVEL0_PROP_FILE = createTemporaryFile(TEMP_DIR, 0, FILE_PREFIX, "properties");
+        TEST_LEVEL1_JAR_FILE = createTemporaryFile(TEMP_DIR, 1, FILE_PREFIX, "jar");
+        TEST_LEVEL1_XML_FILE = createTemporaryFile(TEMP_DIR, 1, FILE_PREFIX, "xml");
+        TEST_LEVEL2_XML_FILE = createTemporaryFile(TEMP_DIR, 2, FILE_PREFIX, "xml");
+    }
+    
+    @AfterClass
+    public static void tearDown() throws IOException {
+        FileUtils.deletePath(TEMP_DIR);
```
jschmidt10, line 46: If you used the JUnit provided `TemporaryFolder` Rule, you wouldn't have to worry about the creation or deletion of your temporary directory.

warehouse/common/src/test/java/datawave/common/io/FilesFinderTest.java
```diff
@@ -0,0 +1,115 @@
...
+        Assert.assertEquals(expectedFiles, filesFound);
+    }
+    
+    @Test(expected = IllegalArgumentException.class)
+    public void testGetFilesFromClasspathWithRelativePathFailures() {
+        Collection<String> expectedFiles = Stream.of(TEST_LEVEL0_PROP_FILE, TEST_LEVEL1_JAR_FILE).sorted().collect(Collectors.toList());
+        String testPaths = "../../" + FILE_PREFIX + "0.properties:../../1/" + FILE_PREFIX + "1.jar";
+        
+        Collection<String> filesFound = FilesFinder.getFilesFromClasspath(testPaths, CLASSPATH_DELIMITER);
+        Assert.assertEquals(expectedFiles, filesFound);
```
jschmidt10, line 75: In this scenario, we're expecting `getFilesFromClasspath()` to throw an Exception. We'll never actually evaluate this assertion. Can we change the last two lines to just:

```
FilesFinder.getFilesFromClasspath(testPaths, CLASSPATH_DELIMITER);
```

warehouse/common/src/test/java/datawave/common/io/FilesFinderTest.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+    
+    @AfterClass
+    public static void tearDown() throws IOException {
+        FileUtils.deletePath(TEMP_DIR);
+    }
+    
+    @Test
+    public void testGetFilesFromClasspath() {
+        Collection<String> expectedFiles = Stream.of(TEST_LEVEL1_XML_FILE, TEST_LEVEL2_XML_FILE).sorted().collect(Collectors.toList());
```
jschmidt10, line 51: We could wrap this initializer into a method like:

`Collection<String> expectedFiles = sortedList(TEST_LEVEL1_XML_FILE, TEST_LEVEL2_XML_FILE);`

warehouse/ingest-core/src/main/java/datawave/ingest/util/cache/JobCacheFactory.java
```diff
@@ -0,0 +1,195 @@
...
+    
+    public JobCacheFactory(Configuration conf) throws IOException {
+        Path basePath = new Path(Args.notNull(conf.get(DW_JOB_CACHE_BASE_PATH), DW_JOB_CACHE_BASE_PATH));
+        jobCacheBasePath = new FileSystemPath(basePath.getFileSystem(conf), basePath);
+        
+        timestampPattern = Pattern.compile(Args.notNull(conf.get(DW_JOB_CACHE_TIMESTAMP_PATTERN), DW_JOB_CACHE_TIMESTAMP_PATTERN));
+        timestampDirPrefix = Args.notNull(conf.get(DW_JOB_CACHE_TIMESTAMP_DIR_PREFIX), DW_JOB_CACHE_TIMESTAMP_DIR_PREFIX);
+        
+        String lockFactoryMode = Args.notNull(conf.get(DW_JOB_CACHE_LOCK_FACTORY_MODE), DW_JOB_CACHE_LOCK_FACTORY_MODE);
+        lockFactory = new CacheLockConverter().convert(lockFactoryMode);
```
jschmidt10, line 52: I think this is a cool way to initialize this but the fact that we're delegating through a JCommander converter couples this utility to JCommander now. If we ever moved it around, for instance, we'll need to take JCommander with it.

Could you maybe bump that logic into the LockFactory.Mode instead? Like have a `LockFactory.Mode.fromString()` or something like that? Then have the converter and this both use that utility method.

warehouse/ingest-core/src/main/java/datawave/ingest/util/cache/lease/JobCacheZkLockFactory.java
```diff
@@ -0,0 +1,277 @@
...
+        }
+        
+        InterProcessSemaphoreMutex lock = new InterProcessSemaphoreMutex(client, getLockPath(id));
+        try {
+            if (lock.acquire(lockTimeoutInMs, TimeUnit.MILLISECONDS)) {
+                jobIdToLeases.put(id, lock);
+                return true;
+            }
+        } catch (Exception e) {
+            LOGGER.warn("Unable to acquire lock for {} ", id, e);
```
jschmidt10, line 76: This actually maps to `Logger.warn(String message, Object... arguments)`, so you won't get the nice stacktrace output. Instead, I think we want `LOGGER.warn("Unable to acquire lock for " + id, e)` to get the `Logger.warn(String message, Throwable cause)` functionality.

warehouse/common/src/test/java/datawave/common/io/FilesFinderTest.java
cj2543: I think TemporaryFolder had some concurrency issues with parallel junit test executions.  Last time I checked I think the low side had the junit version that contained the fix but high side did not.   I will have to check.

warehouse/common/src/test/java/datawave/common/io/FilesFinderTest.java
cj2543: Thanks!  Will update.

warehouse/common/src/test/java/datawave/common/io/FilesFinderTest.java
cj2543: Makes sense.  I will update.   I will see how often I use this in my test cases.

warehouse/ingest-core/src/main/java/datawave/ingest/util/cache/JobCacheFactory.java
cj2543: Thanks!  Will update.

warehouse/ingest-core/src/main/java/datawave/ingest/util/cache/lease/JobCacheZkLockFactory.java
cj2543: Thanks!  I may have made this mistake in multiple places.  I will take a look back as well.

warehouse/ingest-core/src/main/java/datawave/ingest/util/cache/JobCacheFactory.java
jschmidt10: Or maybe just make an alternative `JobCacheLockFactory.getLockFactory(String mode)` method that will throw an Exception if you give it an invalid String.

warehouse/ingest-core/src/main/java/datawave/ingest/util/cache/JobCacheFactory.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+    
+    /**
+     * Form the lock id by joining the passed in array with file.separator as the delimiter.
+     *
+     * @param lockIdElements
+     *            An array of lock id components that when joined form the lock id.
+     * @return A String representing the lock id.
+     */
+    String getLockIdPath(String... lockIdElements) {
```
jschmidt10, line 190: This method is really generic. It could even go in `FileUtils`. Maybe a `FileUtils.toPath(String... elements)` or something like that?

cj2543: I agree that it is generic but the FileUtils is under the commons-test.  I am not sure if I want to create a FileUtils under commons for one method.  I am not sure if it is a good fit under the Files class.  Ok to leave it here?

warehouse/ingest-core/src/main/java/datawave/ingest/util/cache/lease/JobCacheZkLockFactory.java
jschmidt10: Actually, I think I'm wrong here. It looks like Log4j might actually examine the last item to see if it extends Throwable. My bad!

warehouse/common/src/test/java/datawave/common/io/FilesFinderTest.java
jschmidt10: Ahh, okay. That's cool to leave it.

warehouse/ingest-core/src/main/java/datawave/ingest/util/cache/JobCacheFactory.java
jschmidt10: Ahh, sure - that's cool. We could rename the method since it's not necessarily specific to forming a LockId path but more just creating a file path from the pieces.

hlgp:
OBE

PR #1159, opened 2021-04-26
VisitorFunction reduces query terms below the source limit when configured.

apmoriarty:
- PushdownLargeFieldedListsVisitor now operates on max terms OR source limit thresholds
- VisitorFunction will pushdown based on source limit if configured in the query logic factory


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownLargeFieldedListsVisitor.java
```diff
@@ -58,12 +58,14 @@
...
     private Set<String> fields;
+    private long sourceLimit;
     
-    public PushdownLargeFieldedListsVisitor(ShardQueryConfiguration config, FileSystem fs, String fstHdfsUri, Set<String> fields) {
+    public PushdownLargeFieldedListsVisitor(ShardQueryConfiguration config, FileSystem fs, String fstHdfsUri, Set<String> fields, long sourceLimit) {
         this.config = config;
         this.fstHdfsUri = fstHdfsUri;
         this.fs = fs;
         this.fields = fields;
+        this.sourceLimit = sourceLimit;
```
FineAndDandy, line 68: sourceLimit is never used. Either this should be used down below to collapse nodes that are over a source limit (if specified/nor -1) or this should be removed

FineAndDandy (changes requested):

FineAndDandy (dismissed):

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
```diff
@@ -323,7 +332,14 @@ protected ASTJexlScript pushdownLargeFieldedLists(ShardQueryConfiguration config
                         
                         // sort from largest to smallest reductions and make reductions until under the threshold
                         Set<String> fieldsToReduce = new HashSet<>();
-                        int toReduce = termCount - config.getMaxTermThreshold();
+                        int toReduce;
+                        if (termCount > config.getMaxTermThreshold()) {
```
d-hwang, line 336: if it hits both MaxTermThreshold & sourceLimit, does it matter that it only reduces down to MaxTermThreshold?

ivakegg: indeed, the termCount count could  be over both of those values in which case we need to reduce it to the minimum of the threshold and limit.

apmoriarty:
may be worth revisiting in the future

PR #1162, opened 2021-04-29
Fix #1063: Skip TimingMetadata entries when applying Unique transform.

### LINKED ISSUE #1063: UNIQUE functions appears to be dropping results with missing fields
Initially the UNIQUE function was written to return unique instances of results given a set of fields.  It has been observed lately that results with missing fields are not being considered as unique.  Note that this may not be deterministic and hence we may have a threading/yielding issue that is resulting in this observation.  One example of a query with this problem was as follows:

(FIELD1:value1 AND (#ISNOTNULL(FIELD2) OR (#ISNOTNULL(FIELD3) AND (#ISNOTNULL(FIELD4) OR #ISNOTNULL(FIELD5)))) AND (#UNIQUE(FIELD4))

Without the #UNIQUE this returned 1 result where FIELD4 was empty/null.  With the #UNIQUE, this returned 0 results.

friedlou:
Fix #1063


apmoriarty (approved):

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -97,6 +99,10 @@ public UniqueTransform(BaseQueryLogic<Entry<Key,Value>> logic, Set<String> field
     @Override
     public Entry<Key,Document> apply(@Nullable Entry<Key,Document> keyDocumentEntry) {
         if (keyDocumentEntry != null) {
+            if (isTimingMetadata(keyDocumentEntry.getValue())) {
```
ivakegg, line 102: I don't think this is exactly what we need to do.  If we are collecting timing details then every document will have timing metadata (see PipelineQuerySpanCollectionIterator) in which case everything will be returned defeating the whole purpose of this transform.  What you actually need to do here is ignore the "Final Document" which is the entry that contains a summary after all documents are returned.  To test for that use FinalDocumentTrackingIterator.isFinalDocumentKey(Key k).

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
friedlou: Made changes per comment. Please see latest commit.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -108,6 +115,14 @@ public UniqueTransform(BaseQueryLogic<Entry<Key,Value>> logic, Set<String> field
         return keyDocumentEntry;
     }
     
+    /**
+     * Determine if the document is a TIMING_METADATA record. They should be skipped.
+     */
+    private boolean isTimingMetadata(Document document) {
```
ivakegg: Is this method needed anymore?

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
friedlou: Removed in latest commit.

PR #1165, opened 2021-05-06
Address extra parens introduced by several visitors. ASTDelayedPredicate detects if a node is already delayed.

apmoriarty:
- Added RemoveExtraParensVisitor
- PushdownNegationsVisitor now creates unwrapped And/Or nodes
- ASTDelayedPredicate will detect already-delayed nodes and subtrees (avoids double delays)


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownNegationVisitor.java
```diff
@@ -41,7 +39,10 @@ public static JexlNode pushdownNegations(JexlNode tree) {
         flattened.jjtAccept(new PushdownNegationVisitor(), new NegationState(false));
         
         // flatten the result
-        return TreeFlatteningRebuildingVisitor.flatten(flattened);
+        flattened = TreeFlatteningRebuildingVisitor.flatten(flattened);
+        
+        // remove extra parens and return
+        return RemoveExtraParensVisitor.remove(flattened);
```
ivakegg, line 45: Can't we fix the TreeFlatteningRebuildingVisitor to remove extra parens up front?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RemoveExtraParensVisitor.java
```diff
@@ -0,0 +1,372 @@
...
+     *
+     * @param node
+     *            an ASTReferenceExpression node
+     * @param data
+     *            don't need it
+     * @return
+     */
+    @Override
+    public Object visit(ASTReferenceExpression node, Object data) {
+        while (isDoubleWrapped(node)) {
```
ivakegg, line 81: What if it is triple wrapped?

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -1248,6 +1249,8 @@ protected ASTJexlScript processTree(final ASTJexlScript originalQueryTree, Shard
                 ParallelIndexExpansion regexExpansion = new ParallelIndexExpansion(config, scannerFactory, metadataHelper, expansionFields,
                                 config.isExpandFields(), config.isExpandValues(), config.isExpandUnfieldedNegations());
                 queryTree = (ASTJexlScript) regexExpansion.visit(queryTree, null);
+                // regex expansion picks up an extra set of parens, so quickly fix that here
```
ivakegg, line 1252: Can we fix the regex expansion to avoid the extra parens in the first place?

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/ASTDelayedPredicate.java
```diff
@@ -44,10 +44,18 @@ public String toString() {
...
      * @param node
-     * @return
+     *            the node to delay
+     * @return an ASTDelayedPredicate
      */
-    public static ASTDelayedPredicate create(JexlNode node) {
+    public static JexlNode create(JexlNode node) {
+        
+        // Do not delay and already-delayed sub tree
+        if (node instanceof ASTDelayedPredicate || ASTDelayedPredicate.isSubTreeAlreadyDelayed(node)) {
```
ivakegg, line 56: Can't we use instanceOf(node, ASTDelayedPredicate) here instead?

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/PushdownNegationVisitorTest.java
```diff
@@ -280,6 +287,30 @@ public void testApplyDeMorgansNegateOrRoot() {
...
+        Assert.assertEquals("!(!(f1 == 'v1') && !(f2 == 'v2'))", JexlStringBuildingVisitor.buildQuery(result));
+    }
+    
+    @Test
+    public void test_doubleWrappingUnionAfterPushdown() throws ParseException {
+        String query = "!(FOO == 'bar' || FOO == 'baz')";
+        ASTJexlScript script = JexlASTHelper.parseJexlQuery(query);
+        ASTJexlScript result = (ASTJexlScript) PushdownNegationVisitor.pushdownNegations(script);
+        
+        // Really? Double wrapping? That's very untidy and this is very passive aggressive.
```
ivakegg, line 299: I am not getting your passive aggressive comment here.  Please clean this up.

ivakegg (changes requested):

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/PushdownNegationVisitorTest.java
apmoriarty: I'll get rid of that

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownNegationVisitor.java
apmoriarty: I could try rolling this into the TreeFlatteningRebuildingVisitor

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RemoveExtraParensVisitor.java
apmoriarty: This loop effectively eats up extra parens, so triple wrapped should be taken care of. This should probably check for double refs and double refExpresssions though...

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
apmoriarty: An early iteration attempted to fix that but it's uhm, ugly.

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/ASTDelayedPredicate.java
apmoriarty: Can do!

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownNegationVisitor.java
apmoriarty: This approach is quickly proving to be too unwieldy.

FineAndDandy: Alternatively can we add this to the TreeFlatteningRebuildingVisitor.flatten() even as its own visitor so that the expected result is at least the same all the time.

FineAndDandy (changes requested):
call RemoveExtraParensVisitor from TreeFlatteningRebuildingVisitor if a full integration isn't reasonable.

FineAndDandy (approved):

FineAndDandy:
I like that we got a lot of 'free' unit tests by making this update

apmoriarty:
> I like that we got a lot of 'free' unit tests by making this update

Great suggestion

FineAndDandy (approved):

d-hwang (approved):

FineAndDandy (approved):

hlgp (approved):

PR #1170, opened 2021-05-11
Removed duplication of data in the logic members with those in the configuration.

tynyttie:
… from DiscoveryLogic.java. I replaced the variables with separateCountsByColVis = DiscoveryQueryConfiguration.getSeparateCountsByColVis() and DiscoveryQueryConfiguration.getShowReferenceCount(). This was to avoid duplication of data in the logic members with those in the conifgurtion.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryLogic.java
+++ b/warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryLogic.java
@@ -70,7 +70,7 @@ import com.google.common.collect.Sets;
 public class DiscoveryLogic extends ShardIndexQueryTable {
     
     private static final Logger log = Logger.getLogger(DiscoveryLogic.class);
-
+    
     public static final String SEPARATE_COUNTS_BY_COLVIS = "separate.counts.by.colvis";
     public static final String SHOW_REFERENCE_COUNT = "show.reference.count";
     public static final String REVERSE_INDEX = "reverse.index";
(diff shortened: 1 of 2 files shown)


warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryLogic.java
```diff
@@ -500,19 +498,38 @@ public ShardIndexQueryTable clone() {
     }
     
     public Boolean getSeparateCountsByColVis() {
-        return separateCountsByColVis;
+        
+        if (config != null) {
```
ivakegg, line 502: This is looking great.  The only thing I would change is to NOT check if config is null here or in the methods below.  I would rather get a NullPointerException if there is no config as that is a non-starter.  If this logic does not have a config then nothing is going to work well here.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryLogic.java
ivakegg: Alternatively check for null but throw an IllegalStateException or something like that.

ivakegg (approved):

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryLogic.java
```diff
@@ -71,12 +71,11 @@
     
     private static final Logger log = Logger.getLogger(DiscoveryLogic.class);
     
+    DiscoveryQueryConfiguration config;
```
ivakegg, line 74: The base class already has a config member.  What we should do is use setConfig() on the base class instead and then override the getConfig() method to return a DiscoveryQueryConfiguration instead of the GenericQueryConfiguration.

ivakegg (approved):

PR #1175, opened 2021-05-13
Task/seekable index streams

apmoriarty:


apmoriarty:
Rebased and force pushed

warehouse/query-core/src/main/java/datawave/query/index/lookup/BaseIndexStream.java
```diff
@@ -0,0 +1,122 @@
...
+            }
+            hasPeeked = true;
+        }
+        return peekedElement;
+    }
+    
+    @Override
+    public Tuple2<String,IndexInfo> next() {
+        if (!hasPeeked) {
+            return backingIter.next();
```
alerman, line 92: do we need a hasNext call here? Perhaps instead of this block just call peek() to avoid the duplication?

FineAndDandy: I think as this is it conforms to the contract. If you haven't called hasNext() before calling next() you may get a NoSuchMethodException

FineAndDandy (approved):

warehouse/query-core/src/main/java/datawave/query/index/lookup/BaseIndexStream.java
apmoriarty: Strictly speaking, no. This method still conforms to the contract of the PeekingIterator. There is a way you could force this to break, but to do so you would also be violating the method contract of an iterator (by calling next() before calling hasNext()).

PR #1183, opened 2021-05-20
Bugfix/issue 887 - Add more null checks in JEXL functions

billoley:
Add more null checks in JEXL functions to prevent empty fields (value=null) from causing InvocationExceptions from being propagated

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/functions/QueryFunctions.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/functions/QueryFunctions.java
@@ -68,11 +68,13 @@ public class QueryFunctions {
         if (field != null) {
             // we only want to test against the normalized variant to be consistent with the DatawaveArithmetic
             String fieldValue = ValueTuple.getNormalizedStringValue(field);
-            int leftComparison = fieldValue.compareTo(left);
-            if (leftComparison > 0 || (leftInclusive && leftComparison == 0)) {
-                int rightComparison = fieldValue.compareTo(right);
-                if (rightComparison < 0 || (rightInclusive && rightComparison == 0)) {
-                    return Collections.singleton(getHitTermString(field));
+            if (fieldValue != null) {
+                int leftComparison = fieldValue.compareTo(left);
+                if (leftComparison > 0 || (leftInclusive && leftComparison == 0)) {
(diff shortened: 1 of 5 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/functions/QueryFunctions.java
```diff
@@ -61,9 +65,9 @@ protected static String getHitTermString(Object valueTuple) {
     }
     
     public static Collection<?> between(Object field, String left, boolean leftInclusive, String right, boolean rightInclusive) {
-        // we only want to test against the normalized variant to be consistent with the DatawaveArithmetic
-        String fieldValue = ValueTuple.getNormalizedStringValue(field);
-        if (fieldValue != null) {
+        if (field != null) {
+            // we only want to test against the normalized variant to be consistent with the DatawaveArithmetic
+            String fieldValue = ValueTuple.getNormalizedStringValue(field);
```
keith-ratcliffe, line 70: Isn't the null check for `fieldValue` still needed, since ValueTuple.getNormalizedStringValue can return null?

ivakegg:
This should be against release/version3.2

warehouse/query-core/src/main/java/datawave/query/jexl/functions/QueryFunctions.java
ivakegg: can't hurt

keith-ratcliffe:
BTW, the failed quickstart actions, such as the one on this PR, can be ignored until I can push a fix for the quickstart's Dockerfile. Suddenly, at least one of the Dockerfile volume definitions (~/.m2/repository in this case) is no longer deemed valid by the docker runtime, likely due to a version change. Everything prior to the volume declaration succeeds, including the ingest and query integration tests. Ticket and fix for this coming shortly

warehouse/query-core/src/main/java/datawave/query/jexl/functions/QueryFunctions.java
billoley: I didn't look like ValueTuple.getNormalizedStringValue should return a null unless it was passed a null, but I added two additional null checks for safety and readability.

Also retargetted the PR to release/version3.2

ivakegg (approved):

PR #1191, opened 2021-05-26
Clean up orphaned partial bulk load directories #1150

sbgift-dev:
Since the last MR: added a check on writeDirectlyToDest to determine which FS to delete from.
Ran into a Wrong FS error previously


warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
```diff
@@ -340,7 +340,11 @@ public int run(String[] args) throws Exception {
                 log.error(e);
                 log.info("Deleting orphaned directory: " + workDirPath);
                 try {
-                    inputFs.delete(workDirPath, true);
+                    if (writeDirectlyToDest) {
```
javease, line 343: Not sure why you need this if condition. Believe this can just be outputFs.delete(workDirPath, true) because if !writeDirectlyToDest then outputFs=inputFs.

sbgift-dev: Yep that makes sense, thanks

hlgp (dismissed):

ivakegg (approved):

hlgp (approved):

ivakegg (dismissed):

jschmidt10:
@sbgift-dev Your commit history looks a little wigged out - it includes some commits that are not yours. We might need to cherry-pick or rebase this back against integration. Let me know if you need a hand with this.

jschmidt10:
> @sbgift-dev Your commit history looks a little wigged out - it includes some commits that are not yours. We might need to cherry-pick or rebase this back against integration. Let me know if you need a hand with this.

It does look like this commit tree has stuff that has already been committed. Not a huge deal since we can squash these all together.

jschmidt10 (approved):

matthpeterson (approved):

PR #1193, opened 2021-05-27
Sorted backport

hlgp:


warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/SplitsFileType.java
```diff
@@ -0,0 +1,47 @@
...
+     * tableA 2ef 
+     * tableA 3gh 
+     * tableA 4ij 
+     * tableA 5kl 
+     * tableA 6mn 
+     * tableA 7op 
+     * tableA 8qr
+     * tableA 9st
+     * 
+     * And the number wanted is 4 then the resulting splits would be
```
matthpeterson, line 32: It seems that in practice there would be 5 total groups covered by these 4 splits.  Data for tablet 0ab would be included in the group ending with 2ef.  Tablet 9st would need to be included in an unnamed "leftover" group.

If the goal is to have n groups, it would make sense to have n-1 splits.  Otherwise, to have the number of groups match the number of splits, it might be cleaner and simpler to always include the final entry:
```
     * tableA 2ef 
     * tableA 4ij 
     * tableA 6mn 
     * tableA 9st
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/SplitsFileType.java
```diff
@@ -0,0 +1,47 @@
...
+     * 
+     * And the number wanted is 4 then the resulting splits would be 
+     * tableA 2ef 
+     * tableA 4ij 
+     * tableA 6mn 
+     * tableA 8qr
+     * 
+     */
+    // @formatter:on
+    TRIMMEDBYNUMBER,
```
matthpeterson, line 40: Suggestion: to conform to Java's conventions for constants, rename to TRIMMED_BY_NUMBER.  Similarly, SPLITS_AND_LOCATIONS.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/partition/MultiTableRangePartitioner.java
```diff
@@ -33,21 +34,22 @@
     static TaskInputOutputContext<?,?,?,?> context = null;
     private static boolean collectStats = false;
     
-    private volatile boolean cacheFilesRead = false;
+    protected volatile boolean cacheFilesRead = false;
     private Text holder = new Text();
-    private ThreadLocal<Map<String,Text[]>> splitsByTable = new ThreadLocal<>();
+    protected ThreadLocal<Map<String,Text[]>> splitsByTable = new ThreadLocal<Map<String,Text[]>>();
```
matthpeterson, line 39: The right hand side can be just 
```
new ThreadLocal<>();
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/partition/MultiTableRangePartitioner.java
```diff
@@ -58,15 +60,21 @@ private void readCacheFilesIfNecessary() {
...
-            } catch (IOException e) {
+            } catch (Exception e) {
                 log.error("Failed to get localCacheFiles from context", e);
                 throw new RuntimeException("Failed to get localCacheFiles from context", e);
             }
             
             try {
-                NonShardedSplitsFile.Reader reader = new NonShardedSplitsFile.Reader(context.getConfiguration(), localCacheFiles, isTrimmed());
+                NonShardedSplitsFile.Reader reader = new NonShardedSplitsFile.Reader(context.getConfiguration(), localCacheFiles, getSplitsFileType());
+                if (getSplitsFileType().equals(SplitsFileType.SPLITSANDLOCATIONS)) {
```
matthpeterson, line 70: splitToLocationMap is only used in SortedTabletLocationPartitioner yet the field exists here and the code for initializing it is here.  Perhaps it would be worthwhile to move the field and initialization code to that partitioner by defining a new updateSplits method and calling it here, passing the reader to it.  It could then be overridden by subclasses.

It's just a bit confusing as this partitioner does not use it.  Maybe it's not worth the suggested refactor.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/SplitsFileType.java
hlgp: I interpreted TRIMMEDBYNUMBER to mean down-sampled to a specified number of splits with it already implied that the number of groups would be n+1

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/SplitsFileType.java
hlgp: sounds good

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/partition/MultiTableRangePartitioner.java
hlgp: thanks

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/partition/MultiTableRangePartitioner.java
hlgp: Yes, I found the inheritance structure of this puzzling.  I think the original idea was trying to limit duplicated code and also handle a theoretical case where we have several nonsharded tables configured to use different partitioners instead of all of them using the default, but resulting in the parent having to be aware of the type of file that its children need instead of the children simply overriding this method *shrug* I'm speculating.  I've noticed some odd logic in the way we separate the duties of the NonShardedSplitsFile and the ShardedTableMapFile, so I'm working a refactor where we have one SplitsFile to rule them all and each paritioner simply asks for what it needs from that cache and I plan to address this further weirdness there.

hlgp:
After integration testing, this partitioner makes things about 30% slower.  We can do better. So the partitioner's presence in this MR doesn't hold up the other changes (despite the partitioner being inactive) I will open a follow-on to capture the other changes separately.

PR #1194, opened 2021-06-01
Don't drop UIDs in GlobalIndexUidAggregator

### LINKED ISSUE #1173: GlobalIndexUidAggregator can drop UIDs in the remove list
GlobalIndexUidAggregator is typically driven by TotalAggregatingIterator, even though it extends Combiner and probably could be made to run independently as a Combiner. TotalAggregatingIterator appears to treat scan, minor, and non-full major compactions in the same way with special behavior for a full major compaction. If a non-full major compaction happens in the right order, then it is possible for UIDs in the removed list to be dropped (where the UID will remain in the UID list and just the removal will be dropped).

I simulated the right circumstances by starting with a table configured with TotalAggregatingIterator/GlobalIndexUidAggregator. Then, I inserted some keys and scanned to get the output that would be produced by a major compaction involving those keys. I then took that output and inserted in an empty table (to simulate the table now having the result of the non-full major compaction). I then added a new key and scanned.

```
> insert term "" "" '\x08\x00\x10\x01\x1A\x05uuid2' -ts 20
> insert term "" "" '\x08\x00\x10\x01\x1A\x05uuid3' -ts 30
> insert term "" "" '\x08\x00\x10\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\x01"\x05uuid1' -ts 40
> scan -st
term : [] 40    \x08\x00\x10\x01\x1A\x05uuid2\x1A\x05uuid3"\x05uuid1
```

Here we see the protocol buffer has uuid2, uuid3 in the UID list and uuid1 in the remove list. The total count is 1 since we had +1 each for uuid2 and uuid3 and -1 for uuid1. Now, I wiped the table and inserted two keys:

```
> insert term "" "" '\x08\x00\x10\x01\x1A\x05uuid2\x1A\x05uuid3"\x05uuid1' -ts 40
> insert term "" "" '\x08\x00\x10\x01\x1A\x05uuid1' -ts 10
> scan -st
term : [] 40    \x08\x00\x10\x02\x1A\x05uuid2\x1A\x05uuid1\x1A\x05uuid3
```

The first insert represents the result of the partial major compaction that would have happened had the files containing the first 3 inserts above been compacted together. Note that the timestamp of uuid1 (10) is older than the removal of uuid1 (40), so uuid1 should be gone. However, the final  protocol buffer has uuid1, uuid2, and uuid3 in the UID list (though the count of 2 is what it should be).

When I initially tried this test, I tried inserting just a single uuid, but found that it plus the delete yielded a count of 0 which is also not handled in the reduce method in GlobalIndexUidAggregator:

```
> insert term "" "" '\x08\x00\x10\x01\x1A\x05uuid2' -ts 20
> insert term "" "" '\x08\x00\x10\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\x01"\x05uuid1' -ts 40
> scan -st
term : [] 40    \x08\x00\x10\x00\x1A\x05uuid2"\x05uuid1
> compact -w
> scan -st
>
```

brianloss:
* Ensure that if there is a positive (including 0) count for an incoming
  protocol buffer that we handle UIDs in the REMOVEDUID list as well as
  those in the UID list. This scenario is possible if a partial major
  compaction combines files adding UID(s) to the UID list and adding
  other/different UID(s) to the REMOVEDUID list (where the adds of the
  UIDs being removed are in a different file that isn't involved in the
  major compaction).

* Switch logger to use slf4j and placeholders in the log messages.

* Add test cases for new functionality, and apply IDE clean up.

Fixes #1173

### DIFF
--- a/warehouse/core/src/main/java/datawave/ingest/protobuf/Uid.java
+++ b/warehouse/core/src/main/java/datawave/ingest/protobuf/Uid.java
@@ -73,27 +73,6 @@ public final class Uid {
          * <code>repeated string REMOVEDUID = 4;</code>
          */
         com.google.protobuf.ByteString getREMOVEDUIDBytes(int index);
-        
-        // repeated string QUARANTINEUID = 5;
-        /**
-         * <code>repeated string QUARANTINEUID = 5;</code>
-         */
-        java.util.List<java.lang.String> getQUARANTINEUIDList();
-        
-        /**
-         * <code>repeated string QUARANTINEUID = 5;</code>
-         */
-        int getQUARANTINEUIDCount();
-        
-        /**
-         * <code>repeated string QUARANTINEUID = 5;</code>
-         */
-        java.lang.String getQUARANTINEUID(int index);
-        
-        /**
-         * <code>repeated string QUARANTINEUID = 5;</code>
-         */
-        com.google.protobuf.ByteString getQUARANTINEUIDBytes(int index);
     }
     
     /**
@@ -172,14 +151,6 @@ public final class Uid {
                             rEMOVEDUID_.add(input.readBytes());
                             break;
                         }
-                        case 42: {
-                            if (!((mutable_bitField0_ & 0x00000010) == 0x00000010)) {
-                                qUARANTINEUID_ = new com.google.protobuf.LazyStringArrayList();
-                                mutable_bitField0_ |= 0x00000010;
-                            }
-                            qUARANTINEUID_.add(input.readBytes());
-                            break;
-                        }
                     }
                 }
             } catch (com.google.protobuf.InvalidProtocolBufferException e) {
@@ -192,9 +163,6 @@ public final class Uid {
                 }
                 if (((mutable_bitField0_ & 0x00000008) == 0x00000008)) {
                     rEMOVEDUID_ = new com.google.protobuf.UnmodifiableLazyStringList(rEMOVEDUID_);
-                }
-                if (((mutable_bitField0_ & 0x00000010) == 0x00000010)) {
-                    qUARANTINEUID_ = new com.google.protobuf.UnmodifiableLazyStringList(qUARANTINEUID_);
                 }
                 this.unknownFields = unknownFields.build();
                 makeExtensionsImmutable();
@@ -323,44 +291,11 @@ public final class Uid {
             return rEMOVEDUID_.getByteString(index);
         }
         
-        // repeated string QUARANTINEUID = 5;
-        public static final int QUARANTINEUID_FIELD_NUMBER = 5;
-        private com.google.protobuf.LazyStringList qUARANTINEUID_;
-        
-        /**
-         * <code>repeated string QUARANTINEUID = 5;</code>
-         */
-        public java.util.List<java.lang.String> getQUARANTINEUIDList() {
-            return qUARANTINEUID_;
-        }
-        
-        /**
-         * <code>repeated string QUARANTINEUID = 5;</code>
-         */
-        public int getQUARANTINEUIDCount() {
-            return qUARANTINEUID_.size();
-        }
-        
-        /**
-         * <code>repeated string QUARANTINEUID = 5;</code>
-         */
-        public java.lang.String getQUARANTINEUID(int index) {
-            return qUARANTINEUID_.get(index);
-        }
-        
-        /**
-         * <code>repeated string QUARANTINEUID = 5;</code>
-         */
-        public com.google.protobuf.ByteString getQUARANTINEUIDBytes(int index) {
-            return qUARANTINEUID_.getByteString(index);
-        }
-        
         private void initFields() {
             iGNORE_ = false;
             cOUNT_ = 0L;
             uID_ = com.google.protobuf.LazyStringArrayList.EMPTY;
             rEMOVEDUID_ = com.google.protobuf.LazyStringArrayList.EMPTY;
-            qUARANTINEUID_ = com.google.protobuf.LazyStringArrayList.EMPTY;
         }
         
         private byte memoizedIsInitialized = -1;
@@ -396,9 +331,6 @@ public final class Uid {
             for (int i = 0; i < rEMOVEDUID_.size(); i++) {
                 output.writeBytes(4, rEMOVEDUID_.getByteString(i));
             }
-            for (int i = 0; i < qUARANTINEUID_.size(); i++) {
-                output.writeBytes(5, qUARANTINEUID_.getByteString(i));
-            }
             getUnknownFields().writeTo(output);
         }
         
@@ -431,14 +363,6 @@ public final class Uid {
                 }
                 size += dataSize;
                 size += 1 * getREMOVEDUIDList().size();
-            }
-            {
-                int dataSize = 0;
-                for (int i = 0; i < qUARANTINEUID_.size(); i++) {
-                    dataSize += com.google.protobuf.CodedOutputStream.computeBytesSizeNoTag(qUARANTINEUID_.getByteString(i));
-                }
-                size += dataSize;
-                size += 1 * getQUARANTINEUIDList().size();
             }
             size += getUnknownFields().getSerializedSize();
             memoizedSerializedSize = size;
@@ -516,7 +440,8 @@ public final class Uid {
         
         @java.lang.Override
         protected Builder newBuilderForType(com.google.protobuf.GeneratedMessage.BuilderParent parent) {
-            return new Builder(parent);
+            Builder builder = new Builder(parent);
+            return builder;
         }
         
         /**
@@ -560,8 +485,6 @@ public final class Uid {
                 bitField0_ = (bitField0_ & ~0x00000004);
                 rEMOVEDUID_ = com.google.protobuf.LazyStringArrayList.EMPTY;
                 bitField0_ = (bitField0_ & ~0x00000008);
-                qUARANTINEUID_ = com.google.protobuf.LazyStringArrayList.EMPTY;
-                bitField0_ = (bitField0_ & ~0x00000010);
                 return this;
             }
             
@@ -607,11 +530,6 @@ public final class Uid {
                     bitField0_ = (bitField0_ & ~0x00000008);
                 }
                 result.rEMOVEDUID_ = rEMOVEDUID_;
-                if (((bitField0_ & 0x00000010) == 0x00000010)) {
-                    qUARANTINEUID_ = new com.google.protobuf.UnmodifiableLazyStringList(qUARANTINEUID_);
-                    bitField0_ = (bitField0_ & ~0x00000010);
-                }
-                result.qUARANTINEUID_ = qUARANTINEUID_;
                 result.bitField0_ = to_bitField0_;
                 onBuilt();
                 return result;
@@ -652,16 +570,6 @@ public final class Uid {
                     } else {
                         ensureREMOVEDUIDIsMutable();
                         rEMOVEDUID_.addAll(other.rEMOVEDUID_);
-                    }
-                    onChanged();
-                }
-                if (!other.qUARANTINEUID_.isEmpty()) {
-                    if (qUARANTINEUID_.isEmpty()) {
-                        qUARANTINEUID_ = other.qUARANTINEUID_;
-                        bitField0_ = (bitField0_ & ~0x00000010);
-                    } else {
-                        ensureQUARANTINEUIDIsMutable();
-                        qUARANTINEUID_.addAll(other.qUARANTINEUID_);
                     }
                     onChanged();
                 }
@@ -967,103 +875,6 @@ public final class Uid {
                 return this;
             }
             
-            // repeated string QUARANTINEUID = 5;
-            private com.google.protobuf.LazyStringList qUARANTINEUID_ = com.google.protobuf.LazyStringArrayList.EMPTY;
-            
-            private void ensureQUARANTINEUIDIsMutable() {
-                if (!((bitField0_ & 0x00000010) == 0x00000010)) {
-                    qUARANTINEUID_ = new com.google.protobuf.LazyStringArrayList(qUARANTINEUID_);
-                    bitField0_ |= 0x00000010;
-                }
-            }
-            
-            /**
-             * <code>repeated string QUARANTINEUID = 5;</code>
-             */
-            public java.util.List<java.lang.String> getQUARANTINEUIDList() {
-                return java.util.Collections.unmodifiableList(qUARANTINEUID_);
-            }
-            
-            /**
-             * <code>repeated string QUARANTINEUID = 5;</code>
-             */
-            public int getQUARANTINEUIDCount() {
-                return qUARANTINEUID_.size();
-            }
-            
-            /**
-             * <code>repeated string QUARANTINEUID = 5;</code>
-             */
-            public java.lang.String getQUARANTINEUID(int index) {
-                return qUARANTINEUID_.get(index);
-            }
-            
-            /**
-             * <code>repeated string QUARANTINEUID = 5;</code>
-             */
-            public com.google.protobuf.ByteString getQUARANTINEUIDBytes(int index) {
-                return qUARANTINEUID_.getByteString(index);
-            }
-            
-            /**
-             * <code>repeated string QUARANTINEUID = 5;</code>
-             */
-            public Builder setQUARANTINEUID(int index, java.lang.String value) {
-                if (value == null) {
-                    throw new NullPointerException();
-                }
-                ensureQUARANTINEUIDIsMutable();
-                qUARANTINEUID_.set(index, value);
-                onChanged();
-                return this;
-            }
-            
-            /**
-             * <code>repeated string QUARANTINEUID = 5;</code>
-             */
-            public Builder addQUARANTINEUID(java.lang.String value) {
-                if (value == null) {
-                    throw new NullPointerException();
-                }
-                ensureQUARANTINEUIDIsMutable();
-                qUARANTINEUID_.add(value);
-                onChanged();
-                return this;
-            }
-            
-            /**
-             * <code>repeated string QUARANTINEUID = 5;</code>
-             */
-            public Builder addAllQUARANTINEUID(java.lang.Iterable<java.lang.String> values) {
-                ensureQUARANTINEUIDIsMutable();
-                super.addAll(values, qUARANTINEUID_);
-                onChanged();
-                return this;
-            }
-            
-            /**
-             * <code>repeated string QUARANTINEUID = 5;</code>
-             */
-            public Builder clearQUARANTINEUID() {
-                qUARANTINEUID_ = com.google.protobuf.LazyStringArrayList.EMPTY;
-                bitField0_ = (bitField0_ & ~0x00000010);
-                onChanged();
-                return this;
-            }
-            
-            /**
-             * <code>repeated string QUARANTINEUID = 5;</code>
-             */
-            public Builder addQUARANTINEUIDBytes(com.google.protobuf.ByteString value) {
-                if (value == null) {
-                    throw new NullPointerException();
-                }
-                ensureQUARANTINEUIDIsMutable();
-                qUARANTINEUID_.add(value);
-                onChanged();
-                return this;
-            }
-            
             // @@protoc_insertion_point(builder_scope:datawave.ingest.protobuf.List)
         }
         
@@ -1084,17 +895,15 @@ public final class Uid {
     
     private static com.google.protobuf.Descriptors.FileDescriptor descriptor;
     static {
-        java.lang.String[] descriptorData = {"\n\tUid.proto\022\030datawave.ingest.protobuf\"]\n"
+        java.lang.String[] descriptorData = {"\n\tUid.proto\022\030datawave.ingest.protobuf\"F\n"
                         + "\004List\022\016\n\006IGNORE\030\001 \002(\010\022\r\n\005COUNT\030\002 \002(\004\022\013\n\003"
-                        + "UID\030\003 \003(\t\022\022\n\nREMOVEDUID\030\004 \003(\t\022\025\n\rQUARANT"
-                        + "INEUID\030\005 \003(\tB\034\n\030datawave.ingest.protobuf" + "H\001"};
+                        + "UID\030\003 \003(\t\022\022\n\nREMOVEDUID\030\004 \003(\tB\034\n\030datawav" + "e.ingest.protobufH\001"};
         com.google.protobuf.Descriptors.FileDescriptor.InternalDescriptorAssigner assigner = new com.google.protobuf.Descriptors.FileDescriptor.InternalDescriptorAssigner() {
             public com.google.protobuf.ExtensionRegistry assignDescriptors(com.google.protobuf.Descriptors.FileDescriptor root) {
                 descriptor = root;
                 internal_static_datawave_ingest_protobuf_List_descriptor = getDescriptor().getMessageTypes().get(0);
                 internal_static_datawave_ingest_protobuf_List_fieldAccessorTable = new com.google.protobuf.GeneratedMessage.FieldAccessorTable(
-                                internal_static_datawave_ingest_protobuf_List_descriptor, new java.lang.String[] {"IGNORE", "COUNT", "UID", "REMOVEDUID",
-                                        "QUARANTINEUID",});
+                                internal_static_datawave_ingest_protobuf_List_descriptor, new java.lang.String[] {"IGNORE", "COUNT", "UID", "REMOVEDUID",});
                 return null;
             }
         };
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
@@ -1,55 +1,41 @@
 package datawave.ingest.table.aggregator;
 
+import java.io.IOException;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.Map;
 
+import datawave.ingest.protobuf.Uid.List.Builder;
+import org.apache.accumulo.core.client.IteratorSetting;
 import org.apache.accumulo.core.data.Key;
 import org.apache.accumulo.core.data.Value;
-import org.apache.log4j.Logger;
 
 import com.google.protobuf.InvalidProtocolBufferException;
 
 import datawave.ingest.protobuf.Uid;
+import org.apache.accumulo.core.iterators.IteratorEnvironment;
+import org.apache.accumulo.core.iterators.SortedKeyValueIterator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Implementation of an Aggregator that aggregates objects of the type Uid.List. This is an optimization for the shardIndex and shardReverseIndex, where the
  * list of UIDs for events will be maintained in the global index for low cardinality terms.
- * 
- * 
- * 
  */
 public class GlobalIndexUidAggregator extends PropogatingCombiner {
-    private static final Logger log = Logger.getLogger(GlobalIndexUidAggregator.class);
-    private Uid.List.Builder builder = Uid.List.newBuilder();
+    private static final Logger log = LoggerFactory.getLogger(GlobalIndexUidAggregator.class);
+    private static final String TIMESTAMPS_IGNORED = "timestampsIgnored";
     
     /**
      * Using a set instead of a list so that duplicate UIDs are filtered out of the list. This might happen in the case of rows with masked fields that share a
      * UID.
      */
-    private HashSet<String> uids = new HashSet<>();
-    
-    public GlobalIndexUidAggregator(int max) {
-        this.maxUids = max;
-    }
-    
-    public GlobalIndexUidAggregator() {
-        this.maxUids = MAX;
-    }
+    private final HashSet<String> uids = new HashSet<>();
     
     /**
      * List of UIDs to remove.
      */
-    private HashSet<String> uidsToRemove = new HashSet<>();
-    
-    /**
-     * List of UIDs to remove.
-     */
-    private HashSet<String> quarantinedIds = new HashSet<>();
-    
-    /**
-     * List of UIDs to remove.
-     */
-    private HashSet<String> releasedUids = new HashSet<>();
+    private final HashSet<String> uidsToRemove = new HashSet<>();
     
     /**
      * flag for whether or not we have seen ignore
@@ -64,7 +50,7 @@ public class GlobalIndexUidAggregator extends PropogatingCombiner {
     /**
      * Maximum number of UIDs.
      */
-    public int maxUids = MAX;
+    public int maxUids;
     
     /**
      * representative count.
@@ -72,9 +58,17 @@ public class GlobalIndexUidAggregator extends PropogatingCombiner {
     private long count = 0;
     
     /**
-     * temporary set for removals.
+     * Indicates whether timestamps are "ignored" where it is assumed all keys aggregated by this class share the same timestamp value.
      */
-    protected HashSet<String> tempSet;
+    private boolean timestampsIgnored = true;
+    
+    public GlobalIndexUidAggregator(int max) {
+        this.maxUids = max;
+    }
+    
+    public GlobalIndexUidAggregator() {
+        this.maxUids = MAX;
+    }
     
     /**
      * @return True if we saw a "count only" protobuf during the last reduce operation.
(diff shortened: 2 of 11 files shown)


warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
```diff
@@ -150,18 +144,21 @@ public Value reduce(Key key, Iterator<Value> iter) {
...
-                // if delta > 0, we are collecting the uid list
-                // in the protobuf into our object's uid list.
-                if (delta > 0) {
+                // if delta >= 0, we are collecting the uid list in the protobuf into our object's uid list.
+                //
+                // Note that it is possible to have a delta of 0 if we were to have gone through
+                // a partial major compaction and combined a protocol buffer having a positive
+                // count of 1 and a negative count of 1 (leaving one UID in the UID list and one
+                // UID in the REMOVEDUID list.
+                if (delta >= 0) {
```
jschmidt10, line 161: Nice catch. I'm wondering if this could even happen with numbers less than 0 - like if one RFile contains an insert of uid1 and another RFile has a removeUidList with uid2 and uid3 in it.

brianloss: @jschmidt10 Good point. Maybe the same action needs to be taken regardless of the count in the incoming protocol buffer -- add the UIDs in the add list, and remove the ones in the remove list. I also have some stuff in here that changes behavior w.r.t timestamps (I'm assuming timestamp matters but I know by default datawave truncates timestamp to the day boundary so I'll need to guard that behavior change with an option). I'll take a look at all that at some point and push some updates.

warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
```diff
@@ -138,7 +132,7 @@ public Value aggregate() {
     @Override
     public Value reduce(Key key, Iterator<Value> iter) {
         if (log.isTraceEnabled())
```
javease, line 134: similar to elsewhere, could remove the if wrapping the log statement

warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
```diff
@@ -99,8 +95,7 @@ public Value aggregate() {
             uids.removeAll(quarantinedIds);
             
             if (!releasedUids.isEmpty()) {
-                if (log.isDebugEnabled())
-                    log.debug("Adding released UIDS");
+                log.debug("Adding released UIDS");
                 uids.addAll(releasedUids);
```
javease, line 99: I realize this isn't part of your change but couldn't this result in uids exceeding the max threshold?

warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
```diff
@@ -150,18 +144,21 @@ public Value reduce(Key key, Iterator<Value> iter) {
...
-                if (delta > 0) {
+                // if delta >= 0, we are collecting the uid list in the protobuf into our object's uid list.
+                //
+                // Note that it is possible to have a delta of 0 if we were to have gone through
+                // a partial major compaction and combined a protocol buffer having a positive
+                // count of 1 and a negative count of 1 (leaving one UID in the UID list and one
+                // UID in the REMOVEDUID list.
+                if (delta >= 0) {
                     
                     for (String uid : v.getQUARANTINEUIDList()) {
```
javease, line 163: I'm not fulling understanding the construction of this protobuf. For the case where delta is non-negative, we appear to be treating the QUARANTINEUIDList as "released UIDs" which we'll add back into the list of uids in the aggregator call. For negative delta, this is treated as a quarantine list not to be included in the uids. Is this a correct understanding?

warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
```diff
@@ -172,48 +169,49 @@ public Value reduce(Key key, Iterator<Value> iter) {
                     for (String uid : v.getUIDList()) {
                         
                         // check that a removal has not occurred
-                        // if it has, we decrement the count, from above.
                         if (!uidsToRemove.contains(uid) && !quarantinedIds.contains(uid)) {
                             
                             // add the UID iff we are under our MAX
                             if (uids.size() < maxUids)
                                 uids.add(uid);
```
javease, line 176: could add a break here if condition is false since we won't be adding any entries in the for loop once maxUids is reached.

warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
```diff
@@ -172,48 +169,49 @@ public Value reduce(Key key, Iterator<Value> iter) {
...
+                    }
+                    
+                    log.debug("Adding uids {} {}", delta, count);
                     
                     // if our delta is < 0, then we can remove, iff seenIgnore is false. If it is true, there is no need to proceed with removals
-                } else if (delta < 0 && !seenIgnore) {
+                } else if (!seenIgnore) {
                     
                     // so that we can perform the decrement
                     for (String uid : v.getREMOVEDUIDList()) {
```
javease, line 196: again, not fully understanding our use of this protobuf, but in the above block we don't appear to care about the REMOVEDUIDList when propogate is false but here we're updating uidsToRemove regardless of the propagate setting. Does that make sense?

warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
brianloss: Yeah, I'm not really sure about the quarantined/released lists. Maybe @phrocker can comment? Think he knows...

warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
brianloss: If the propogate [sic] var is false, then removals aren't passed along (basically propogate is false for a full major compaction only). So yeah, we should probably be checking the propogate var here too. The seenIgnore flag gets set when there's a protocol buffer whose count exceeds the max. In that case, there's nothing in either the UID or REMOVEDUID list, so we should probably be checking that in the add case as well. I'm not an expert on this class and was just trying to fix a specific error that I encountered. I'm hesitant to make changes that are too sweeping, but if there are other experts on this class who want to weigh in and comment to ensure I haven't broken anything, that would be good. Errors in this class could have some pretty serious negative consequences.

warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
brianloss: It does look like maybe the intent was for the QUARANTINEDUID list to be dual-purpose... You quarantine a UID by doing an insert with a count of -1 and the UID in the QUARANTINEDUID list. You release it by doing an insert with a count of 1 and the UID in the QUARANTINEDUID list. However, with partial major compactions, that approach won't work (or I'm just wrong about how the feature works) since the -1 count PB with the uid in the QUARANTINEDUID list could get combined with a PB having, say, a count of 2 and two other uids in the UID list. The result would be a PB with a count of 1, 2 uids in the UID list, and 1 in the QUARANTINEDUID list. The next time that PB is read in (for any of a scan, partial major compaction, or full major compaction) the delta will be positive and the quarantined UID will be released even though that probably wasn't the intent.

warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
jschmidt10: Sounds good. Thanks!

warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
phrocker: hmm, my recollection is that all keys would be propagated until a full major C, so that even if the partial major occurred all keys would exist until that FullMajorC was launched periodically.

brianloss: In the untouched version of this, the uids in the QUARANTINEDUID list are released when the delta is positive (see [here](https://github.com/NationalSecurityAgency/datawave/blob/master/warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java#L166)) and uids in the QUARANTINEDUID list are carried forward as quarantined when the delta is negative (see [here](https://github.com/NationalSecurityAgency/datawave/blob/master/warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java#L204)).

This seems like a problem for partial major compactions. In case 1, say two PBs are combined. One has a count of -1 and uid1 in the QUARANTINEDUID list. The other PB has a count of 2 and uid2 and uid3 in the UID list. The combined PB will have a count of 1, uid2 and uid3 in the UID list, and uid1 in the QUARANTINEDUID list. Then, next time there's a scan or another compaction, the incoming delta for this new PB will be positive and the quarantined uid1 will be released. For case 2, say a previously quarantined uid1 is released so there's a PB with a count of 1 and uid1 in the QUARANTINEDUID list. Say that PB gets combined with another during a partial major compaction where the other PB has a count of -2 with uid2 and uid3 in the REMOVEDUID list. The final PB will have a count of -1, uid1 in the UID list, and uid2 and uid3 in the REMOVEDUID list. Now when that PB eventually gets combined with the one that originally quarantined uid1, uid1 will be re-quarantined.

phrocker: Yep. I thought there being checks for testing if  partial, not doing any combining of keys at all, leaving scan and full major c to do the work as a safeguard....then periodic full majors doing the combining and reduction of keys...but again, my knowledge is vestigial.

brianloss: Perhaps the way to handle this is to modify the protobuf to add a RELEASEDUID list as well. When the intent is to release a previously quarantined UID, insert a PB with a count of 1 and the uid in the new RELEASEDUID list. Then the action to take in the code is clear and it will work with partial major compactions. We could add a special case to release a UID if the incoming PB contains only uids in the QUARANTINEDUID list and not any of the other lists. That would catch previously released quarantined uids that were never involved in a partial major compaction. For those that were involved in a partial majc, I don't think one can make any assumptions about uids in the QUARANTINEDUID list. @ivakegg, any thoughts? I have some local changes to attempt this and also fix the case @jschmidt10  identified. I'll go ahead and push those, and we can just drop the new commit if people don't like that approach.

ivakegg:
Would it be possible to redo this against the integration branch?  Sorry for the hassle but accumulo-2.0 is somewhat delayed and I would like to get this one out there....

brianloss:
> Would it be possible to redo this against the integration branch? Sorry for the hassle but accumulo-2.0 is somewhat delayed and I would like to get this one out there....

Sure, I'll take a look a rebasing it this week. I was in the process of reworking and need to write more tests, but then got busy on other tasks. Hopefully I can spend some time writing more tests this week.

brianloss:
> Would it be possible to redo this against the integration branch? Sorry for the hassle but accumulo-2.0 is somewhat delayed and I would like to get this one out there....

@ivakegg, I rebased my branch off integration, force pushed, and updated the PR target to integration. I also added another commit with new changes based on prior discussions in the PR. I think this needs more tests to confirm behavior, but wanted to get eyes on the code changes since it's a complicated class, and and error could have a big impact on a production system.

warehouse/ingest-core/src/main/java/datawave/iterators/PropogatingIterator.java
```diff
@@ -303,7 +305,10 @@ public boolean validateOptions(Map<String,String> options) {
         Preconditions.checkNotNull(env);
         Preconditions.checkNotNull(options);
         
-        shouldPropogate = !(env.getIteratorScope() == IteratorScope.majc && env.isFullMajorCompaction());
+        // Don't propagate for either scan or full major compaction. In either case, the aggregated result has combined
+        // all existing values for a key so we don't need to propagate temporary state that is only used to combine
+        // partial results with new info.
+        shouldPropogate = !(env.getIteratorScope() == IteratorScope.majc && env.isFullMajorCompaction()) && !(env.getIteratorScope() == IteratorScope.scan);
```
phrocker, line 311: I think this change makes it much more explicit which I think is important here.

warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
```diff
@@ -240,32 +288,47 @@ public void reset() {
...
+        copy.timestampsIgnored = timestampsIgnored;
+        copy.propogate = propogate;
+        // Not copying other fields that are all cleared in the reset() method.
+        return copy;
+    }
+    
+    @Override
+    public void init(SortedKeyValueIterator<Key,Value> source, Map<String,String> options, IteratorEnvironment env) throws IOException {
+        super.init(source, options, env);
+        if (options.containsKey(TIMESTAMPS_IGNORED)) {
```
jschmidt10, line 316: I hit an issue with trying to add options to GlobalIndexUidAggregator - since we're configuring it as an aggregator option on a TotalAggregatingIterator - the `GlobalIndexUidAggregator.init()` method is never actually invoked. Also, the TotalAggregatingIterator treats every option as a column so it'll treat `timestampsIgnored` as the column and try to instantiate a class of `true` for instance.

I also tried using GlobalIndexUidAggregator as a top-level iterator but there is an issue with that as the `propogateKey()` and `reset()` methods are invoked by TotalAggregatingIterator (via PropogatingIterator) - if you use GlobalIndexUidAggregator directly on the table, only `reduce()` gets invoked by Accumulo. 

I had considered maybe adding a hard-coded option to use as a flag to delegate options down into the aggregator like maybe `table.iterator.majc.UidAgg.opt.*.aggropt.timestampsIgnored`. I ended up just using a subclass of GlobalIndexUidAggregator instead though.

brianloss: @jschmidt10, PR #1091 would allow the options to be passed to aggregators via TotalAggregatingIterator. To use GlobalIndexUidAggregator as a Combiner, other work would probably be required.

jschmidt10: Great! Thanks.

jschmidt10 (dismissed):

warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
brianloss: That PR also allows the max UID count to be configured. I can roll just the changes that allow the options to be passed along to this PR if that will make it easier to test anything. (Or, I can make that a separate PR.)

jschmidt10: I think we're set. With ignoreTimestamp being the default, we don't need the option right away.

warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
brianloss: @javease, I believe my latest change should address your question and also resolve the discussion in this thread.

warehouse/ingest-core/src/test/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregatorTest.java
```diff
@@ -224,8 +225,353 @@ public void testCount() throws Exception {
         Uid.List resultList = Uid.List.parseFrom(result.get());
         
         assertEquals(5, resultList.getCOUNT());
-        assertEquals(false, resultList.getIGNORE());
+        assertFalse(resultList.getIGNORE());
+        assertEquals(1, resultList.getUIDCount());
+        
+    }
+
```
matthpeterson, line 235: Thanks for tackling this worthwhile effort to correct the inaccuracies with the aggregator!

The test below adds the same UID twice, highlighting the possible inconsistencies between getCount and the size of getUIDList.

Note that the test fails both before and after this PR's changes.  It's just highlighting another inaccuracy with the aggregator.

```
    @Test
    public void testAddUIDTwice() throws Exception {
        GlobalIndexUidAggregator localAgg = new GlobalIndexUidAggregator();
        IteratorSetting is = new IteratorSetting(19, "test", GlobalIndexUidAggregator.class);
        GlobalIndexUidAggregator.setCombineAllColumns(is, true);
        localAgg.validateOptions(is.getOptions());

        UUID uuid1 = UUID.randomUUID();
        ArrayList<Value> values = Lists.newArrayList();
        Builder b = createNewUidList();
        b.setIGNORE(false);
        b.setCOUNT(1);
        b.addUID(uuid1.toString());
        values.add(new Value(b.build().toByteArray()));

        b = createNewUidList();
        b.setIGNORE(false);
        b.setCOUNT(1);
        b.addUID(uuid1.toString());
        values.add(new Value(b.build().toByteArray()));

        Value result = localAgg.reduce(new Key("key"), values.iterator());
        Uid.List resultList = Uid.List.parseFrom(result.get());        
        assertEquals(1, resultList.getUIDCount());
        assertTrue(resultList.getUIDList().contains(uuid1.toString()));
        assertEquals(1, resultList.getUIDList().size());
        assertEquals(1, resultList.getCOUNT()); // Fails here getCount()==2
    }
```

There are use cases in which the same UID will be added multiple times so it's worth considering the behavior.  

The test above highlights how the aggregator will have an inconsistency between its count and list size.  The test below shows how the aggregator will assess that the max was reached without first checking if the UID is unique or already in the UIDList.  

It seems that the same UID added three times (while IGNORE == false) should result in a count of one and a UID list of size one.

```
    @Test
    public void testAddUIDThrice() throws Exception {
// lowered max to show problem more easily
        GlobalIndexUidAggregator localAgg = new GlobalIndexUidAggregator(2);
        IteratorSetting is = new IteratorSetting(19, "test", GlobalIndexUidAggregator.class);
        GlobalIndexUidAggregator.setCombineAllColumns(is, true);
        localAgg.validateOptions(is.getOptions());

        UUID uuid1 = UUID.randomUUID();
        ArrayList<Value> values = Lists.newArrayList();
        Builder b = createNewUidList();
        b.setIGNORE(false);
        b.setCOUNT(1);
        b.addUID(uuid1.toString());
        values.add(new Value(b.build().toByteArray()));

        b = createNewUidList();
        b.setIGNORE(false);
        b.setCOUNT(1);
        b.addUID(uuid1.toString());
        values.add(new Value(b.build().toByteArray()));

        b = createNewUidList();
        b.setIGNORE(false);
        b.setCOUNT(1);
        b.addUID(uuid1.toString());
        values.add(new Value(b.build().toByteArray()));

        Value result = localAgg.reduce(new Key("key"), values.iterator());
        Uid.List resultList = Uid.List.parseFrom(result.get());
        assertEquals(1, resultList.getUIDCount()); // Fails getUIDCount == 0
        assertTrue(resultList.getUIDList().contains(uuid1.toString()));
        assertEquals(1, resultList.getUIDList().size());
        assertEquals(1, resultList.getCOUNT());
    }
```

Perhaps the logic should not depend on count unless seenIgnore is true.  The size of the UID list would be a more accurate count to use, when available.

Note that before the UIDs are put into the protobuf List they are deduplicated in a HashSet within the aggregator.

Maybe this issue is outside the scope of this ticket.

brianloss: Sure, these are good points. I originally wanted only to correct the UIDs getting dropped, and was trying to change as little as possible to make that happen. Then more things cropped up and that lead to adding the released UID list and also re-working the inner logic. It's not unreasonable to try to solve issues with duplicate UIDs too. It gets tricky if you are trying to combine 2 protocol buffers that are near their limit (and have a large max UID count). Do you combine the two sets and have a possibly huge (albeit temporary list)? Do you compute an intersection look at the total count (which may save memory but you also have to do it every time and have the performance hit)?

warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
```diff
@@ -150,69 +155,113 @@ public Value reduce(Key key, Iterator<Value> iter) {
...
+                            releasedUids.add(uid);
+                        if (timestampsIgnored)
+                            quarantinedIds.remove(uid);
                     }
-                    
+                } else {
+                    // Add any quarantined UIDs to the internal list. If we're paying attention to timestamps, then we
+                    // don't want to quarantine any uid that we've already seen (i.e., with a newer timestamp) that has
+                    // been released.
+                    for (String uid : v.getQUARANTINEUIDList()) {
```
matthpeterson: I'm unaware of any uses of the quarantine and release capabilities.  Perhaps it would be best to deprecate and drop support, for the sake of simplifying this complicated and critical aggregator.  Thoughts?

brianloss: It would make the logic simpler, for sure.

brianloss: I will try removing that, and covering the duplicate UID cases.

warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
```diff
@@ -150,69 +155,113 @@ public Value reduce(Key key, Iterator<Value> iter) {
...
+                 * where compactions did not occur, and the indices are converted into the newer protobuf, we must use the UID list to maintain removals for
+                 * deltas less than 0
+                 */
+                if (delta < 0 && v.getREMOVEDUIDList().isEmpty()) {
+                    for (String uid : v.getUIDList()) {
+                        // add to uidsToRemove, and decrement count if the uid is in UIDS
+                        uidsToRemove.add(uid);
+                        uids.remove(uid);
+                    }
+                }
```
brianloss: This block was effectively moved from an earlier version. @matthpeterson, are you aware of any cases where this is actually needed? I'm wondering if this can go too?

warehouse/ingest-core/src/test/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregatorTest.java
brianloss: @matthpeterson it turns out there's actually a test that verifies that duplicate UIDs are counted! See [here](https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/ingest-core/src/test/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregatorTest.java#L226).

brianloss: @matthpeterson and @jschmidt10, I have pushed changes to attempt to address duplicate UIDs (and also get rid of the quarantine/release lists). It's still a tricky class to get the behavior right, but hopefully this is an improvement. Have a look (especially what's going on with counts in the code and tests) and let me know what you think.

matthpeterson: Interesting test.  I see now that a lot of the other tests also assert a discrepancy between getCount and getUIDCount().

If a system ever re-ingests the same data and then removes it, the existing logic will leave "dangling" index entries - even for a list that has never hit the max.  I'd rather the logic be changed to be as accurate as possible.  I don't know why we wanted to deliberately have inconsistencies between the count and UIDList sizes except to accommodate the quarantine / release logic.  Now that the logic is being updated to rely on the two separate lists for quarantine and release, I don't think we need to try to conflate the meaning of count.

> It gets tricky if you are trying to combine 2 protocol buffers that are near their limit (and have a large max UID count). Do you combine the two sets and have a possibly huge (albeit temporary list)? Do you compute an intersection look at the total count (which may save memory but you also have to do it every time and have the performance hit)?

The change below seems to correct the discrepancies between getCount() and getUIDCount() and make the two unit tests from my prior comment pass.
```
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
@@ -87,9 +87,9 @@ public class GlobalIndexUidAggregator extends PropogatingCombiner {

         // as a backup, we remove the intersection of the UID sets

-        builder.setCOUNT(count);
+        builder.setCOUNT(seenIgnore ? count : uids.size());

-        if (seenIgnore || count > maxUids) {
+        if (seenIgnore || uids.size() > maxUids) {
```

This change would require making the getCOUNT and getUIDCount assertions match for the test you mentioned and a few others.
```
@@ -224,13 +224,81 @@ public class GlobalIndexUidAggregatorTest {
         Value result = agg.reduce(new Key("key"), values.iterator());
         Uid.List resultList = Uid.List.parseFrom(result.get());

-        assertEquals(5, resultList.getCOUNT());
+        assertEquals(1, resultList.getCOUNT());
```
and
```
@@ -267,7 +335,7 @@ public class GlobalIndexUidAggregatorTest {
         Collections.reverse(values);
         Value result = localAgg.reduce(new Key("key"), values.iterator());
         Uid.List resultList = Uid.List.parseFrom(result.get());
-        assertEquals(1, resultList.getCOUNT());
+        assertEquals(2, resultList.getCOUNT());
         assertEquals(2, resultList.getUIDCount());
         assertEquals(2, resultList.getUIDList().size());
         assertEquals(0, resultList.getREMOVEDUIDList().size());
@@ -318,7 +386,7 @@ public class GlobalIndexUidAggregatorTest {
         Collections.reverse(values);
         Value result = localAgg.reduce(new Key("key"), values.iterator());
         Uid.List resultList = Uid.List.parseFrom(result.get());
-        assertEquals(-2, resultList.getCOUNT());
+        assertEquals(1, resultList.getCOUNT());
         assertEquals(1, resultList.getUIDCount());
         assertEquals(1, resultList.getUIDList().size());
         assertEquals(3, resultList.getREMOVEDUIDCount());
@@ -476,7 +544,7 @@ public class GlobalIndexUidAggregatorTest {
         Collections.reverse(values);
         Value result = localAgg.reduce(new Key("key"), values.iterator());
         Uid.List resultList = Uid.List.parseFrom(result.get());
-        assertEquals(0, resultList.getCOUNT());
+        assertEquals(1, resultList.getCOUNT());
         assertEquals(1, resultList.getUIDCount());
         assertEquals(1, resultList.getUIDList().size());
         assertEquals(1, resultList.getREMOVEDUIDList().size());
@@ -543,7 +611,7 @@ public class GlobalIndexUidAggregatorTest {
         Collections.reverse(values);
         Value result = localAgg.reduce(new Key("key"), values.iterator());
         Uid.List resultList = Uid.List.parseFrom(result.get());
-        assertEquals(1, resultList.getCOUNT());
+        assertEquals(2, resultList.getCOUNT());
```

warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
matthpeterson: I'm thinking that this backwards-compatibility code has been in place for so long now that any system having compacted its indexes since then would no longer need it.

brianloss: Ok, I'll push another commit that removes it.

warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
brianloss: The QUARANTINEDUID list has been removed, so this discussion is OBE. Resolving...

warehouse/ingest-core/src/test/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregatorTest.java
brianloss: > The change below seems to correct the discrepancies between getCount() and getUIDCount() and make the two unit tests from my prior comment pass.

I had already made a similar change in my latest commit. I think I must have pushed right as you were typing this up @matthpeterson. Have a look at the latest. It's a bit more of a refactoring since the quarantine list was also removed. It might be easier at this point to look at the raw file vs the diffs.

brianloss: I should add that I looked at enforcing the constraint that `getCOUNT() == (getUIDCount() - getREMOVEDUIDCount())` but found this to not be possible. When we go through the UID list to add them to the in-memory set, we first check if the UID is in the uidsToRemove set and take no action if that's the case. One might suggest removing the uid from uidsToRemove at that point since we've encountered the "add" that goes with the remove. However, it would not be safe to remove the uid from uidsToRemove since we don't yet know if there are any other adds for that uid (from other entries that we haven't seen yet).

warehouse/core/src/main/java/datawave/ingest/protobuf/Uid.java
```diff
@@ -1084,17 +895,15 @@ public Builder addQUARANTINEUIDBytes(com.google.protobuf.ByteString value) {
...
         com.google.protobuf.Descriptors.FileDescriptor.InternalDescriptorAssigner assigner = new com.google.protobuf.Descriptors.FileDescriptor.InternalDescriptorAssigner() {
             public com.google.protobuf.ExtensionRegistry assignDescriptors(com.google.protobuf.Descriptors.FileDescriptor root) {
                 descriptor = root;
                 internal_static_datawave_ingest_protobuf_List_descriptor = getDescriptor().getMessageTypes().get(0);
                 internal_static_datawave_ingest_protobuf_List_fieldAccessorTable = new com.google.protobuf.GeneratedMessage.FieldAccessorTable(
-                                internal_static_datawave_ingest_protobuf_List_descriptor, new java.lang.String[] {"IGNORE", "COUNT", "UID", "REMOVEDUID",
-                                        "QUARANTINEUID",});
+                                internal_static_datawave_ingest_protobuf_List_descriptor, new java.lang.String[] {"IGNORE", "COUNT", "UID", "REMOVEDUID",});
                 return null;
             }
```
matthpeterson, line 908: What happens if the new Uid protobuf (with QUARANTINEUID no longer present) is used to deserialized a legacy serialized Uid protobuf (that had the QUARANTINEUID in its schema, but unpopulated in practice)?

brianloss: I thought protocol buffers would just ignore unknown fields, but we should verify.

matthpeterson (dismissed):

jschmidt10 (dismissed):

matthpeterson:
https://github.com/brianloss/datawave/pull/6 was created with suggested changes to this PR.  If approved and merged, it will modify this PR's source branch and require another round of approvals here.

warehouse/core/src/main/java/datawave/ingest/protobuf/Uid.java
matthpeterson: @jschmidt10 verified that your understanding is correct.

matthpeterson (approved):

jschmidt10 (approved):

PR #1195, opened 2021-06-02
Fixes #879 - X-Partial Status 

### LINKED ISSUE #879: Correctly set X-Partial-Results
X-Partial-Results and PartialResults in the response should indicate whether we are returning less than the requested page size in a response.

plainolneesh:
Re: #879

### DIFF
--- a/web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
+++ b/web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
@@ -45,11 +45,11 @@ import java.util.concurrent.TimeoutException;
  *
  */
 public class RunningQuery extends AbstractRunningQuery implements Runnable {
-
+    
     private static final long serialVersionUID = 1L;
-
+    
     private static Logger log = Logger.getLogger(RunningQuery.class);
-
+    
     private transient AccumuloClient client = null;
     private AccumuloConnectionFactory.Priority connectionPriority = null;
     private transient QueryLogic<?> logic = null;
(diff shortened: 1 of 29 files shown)


web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
```diff
@@ -332,11 +331,8 @@ public ResultsPage next() throws Exception {
                 }
             }
         }
-        if (resultList.isEmpty()) {
-            return new ResultsPage();
-        } else {
-            return new ResultsPage(resultList, ((hitPageByteTrigger || hitPageTimeTrigger) ? ResultsPage.Status.PARTIAL : ResultsPage.Status.COMPLETE));
-        }
+        ResultsPage.Status status = (currentPageCount < maxPageSize) ? ResultsPage.Status.PARTIAL : ResultsPage.Status.COMPLETE;
```
ivakegg, line 334: We are so close here, however a subtlety was missed.  If the page being returned is the last page to be returned, then this is not a "partial page".  So if we have hit the max results, or it.hasNext() is false, then lets return COMPLETE in those cases.

ivakegg (changes requested):

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
```diff
@@ -332,13 +331,16 @@ public ResultsPage next() throws Exception {
                 }
             }
         }
-        if (resultList.isEmpty()) {
-            return new ResultsPage();
-        } else {
-            return new ResultsPage(resultList, ((hitPageByteTrigger || hitPageTimeTrigger) ? ResultsPage.Status.PARTIAL : ResultsPage.Status.COMPLETE));
+        ResultsPage.Status status;
+        if (this.maxResults >= numResults) {
```
ivakegg, line 335: My previous comment was "So if we have hit the max results, or it.hasNext() is false, then lets return COMPLETE in those cases.".  This if statement does not do that.  We "hit the max results" at lines 236 and 241.  In those cases we should return COMPLETE.  Also if !this.it.hasNext()  (i.e it.hasNext() is false) then we should also return COMPLETE.  Only after those checks should you return PARTIAL iff currentPageCount < maxPageSize.  So the code is something like this (provided hitMaxResults is set to true at lines 236 and 241):

if (hitMaxResults || !this.it.hasNext()) {
   status = ResultsPage.Status.COMPLETE;
} else {
   status = (currentPageCount < maxPageSize) ? ResultsPage.Status.PARTIAL : ResultsPage.Status.COMPLETE;
}
return new ResultsPage(resultsList, status);

ivakegg (changes requested):

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
ivakegg: Oh, and BTW this code does not even compile.  Please ensure that all code compiles, passes the formatting, and all tests pass before pushing it up.  Thanks.

plainolneesh: I honstly don't know how I made the mistake of pushing this one. The correct one that i wrote was

  if (resultList.isEmpty()) {
            return new ResultsPage();
        } else if (iter.hasNext() && numResults < this.maxResults && currentPageCount < maxPageSize)
            return new ResultsPage(resultList, ResultsPage.Status.PARTIAL);
        else {
            return new ResultsPage(resultList, ResultsPage.Status.COMPLETE);
        }
    }
I am updating this PR now.

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
```diff
@@ -26,6 +26,7 @@
 import org.apache.commons.lang.StringEscapeUtils;
 import org.apache.htrace.TraceInfo;
 import org.apache.log4j.Logger;
+import org.infinispan.configuration.parsing.ParseUtils;
```
lbschanno, line 29: This unused import can be removed.

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
```diff
@@ -332,10 +332,13 @@ public ResultsPage next() throws Exception {
                 }
             }
         }
+        
         if (resultList.isEmpty()) {
             return new ResultsPage();
-        } else {
-            return new ResultsPage(resultList, ((hitPageByteTrigger || hitPageTimeTrigger) ? ResultsPage.Status.PARTIAL : ResultsPage.Status.COMPLETE));
+        } else if (iter.hasNext() && numResults < this.maxResults && currentPageCount < maxPageSize)
```
lbschanno, line 338: The call to `iter.hasNext()` is causing a couple tests in `ExtendedRunningQueryTest` to fail. You'll need to update those tests to expect an additional call to this method. 

```
Error:  Failures: 
Error:    ExtendedRunningQueryTest.testNextMaxResults_HappyPathUsingDeprecatedConstructor:271 
  Unexpected method call TransformIterator.hasNext():
    TransformIterator.hasNext(): expected: 5, actual: 6
Error:    ExtendedRunningQueryTest.testNextWithDnResultLimit_HappyPathUsingDeprecatedConstructor:448 
  Unexpected method call TransformIterator.hasNext():
    TransformIterator.hasNext(): expected: 3, actual: 4
Error:    ExtendedRunningQueryTest.testNext_HappyPathUsingDeprecatedConstructor:184 
  Unexpected method call TransformIterator.hasNext():
    TransformIterator.hasNext(): expected: 3, actual: 4
Error:    ExtendedRunningQueryTest.testNext_NoResultsAfterCancellationUsingDeprecatedConstructor:322 
  Unexpected method call Query.getPagesize():
    Query.getUncaughtExceptionHandler(): expected: 3, actual: 1
    Query.getId(): expected: 2, actual: 1
    Query.getUserDN(): expected: 2, actual: 1
```

lbschanno (changes requested):

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
codebyneesh: Updated the tests please see new commit .

lbschanno (approved):

PR #1207, opened 2021-06-25
fixes #1206: Updated JexlStringBuildingVisitor to correctly render ne…

### LINKED ISSUE #1206: JexlStringBuildingVisitor doesn't render negative numbers correctly
JexlStringBuildingVisitor renders negative numbers without the negative sign.  So, if you convert a jexl node to string, and convert back to a jexl node, you get a different query.

jwomeara:
…gative numbers

Updated ContentFunctionArguments to correctly parse arguments which represent negative numbers.


warehouse/query-core/src/test/java/datawave/query/jexl/functions/ContentFunctionQueryTest.java
```diff
@@ -496,12 +496,16 @@ public boolean isTermFrequencyField(String field) {
         @Override
         public boolean isIndexListField(String field) {
             return false;
-        };
+        }
+
```
apmoriarty, line 500: Fix spacing before semi-colons?

FineAndDandy (dismissed):

apmoriarty (approved):

warehouse/query-core/src/test/java/datawave/query/jexl/functions/ContentFunctionQueryTest.java
jwomeara: That semicolon was already there - not sure why...  The autoformatter did that, but I have since taken the unnecessary semicolon out.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/ContentFunctionArguments.java
```diff
@@ -107,7 +108,7 @@ public ContentFunctionArguments(Function f) throws ParseException {
                 }
             }
             
-            String nextArg = JexlASTHelper.dereference(args.get(currentArg++)).image.trim();
+            String nextArg = getValue(args.get(currentArg++));
```
ivakegg, line 111: Fascinating.  I didn't realize negative numbers had this issue in jexl.  Do we have any other functions that process the arguments incorrectly as such?

ivakegg (approved):

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/ContentFunctionArguments.java
jwomeara: The geo functions which use northLat, southLat, eastLon, westLon have to have their arguments quoted in order to work.  That's the one that concerned me the most.  If you pass in the arguments unquoted it throws a NPE, which is good at least.  

I did not go through the effort of updating every function which accepts a number, but maybe that's something we should consider doing...

ivakegg (approved):

keith-ratcliffe (approved):
LGTM

PR #1209, opened 2021-06-28
Fixes #1204 - Ensure that relevant index-only fields are added to the Jexl context

### LINKED ISSUE #1204: Phrase query evaluation fails due to 'unknown, ambiguous or inaccessible method phrase'
Original query: `(RUNTIME:[34 TO 52] or TEXT_TOKEN:"gte corp") (SCENE:fire)`

In this case, it seems that `Interpreter.call` is reacting to an argument mismatch when it tries to resolve a `content:phrase` method on the ContentFunctions class.

Remote debugging with Intellij, it appears that the first `phrase` method argument (i.e., local var `argv[0]`), which should be 'TEXT_TOKEN', is actually null at this point, resulting in the stacktrace below

Still investigating
```
[tserver.TabletServer] WARN : Failed to get multiscan result
java.util.concurrent.ExecutionException: java.lang.RuntimeException: Failure for query 028ccc46-bed5-4568-a65a-2afab26906d2
 .....
Caused by: org.apache.commons.jexl2.JexlException$Method: datawave.query.function.JexlEvaluation.<init>@48![1,57]:
 '(content:phrase(TEXT_TOKEN, termOffsetMap, 'gte', 'corp') 
&& TEXT_TOKEN == 'gte' 
&& TEXT_TOKEN == 'corp' 
&& (SCENE == 'fire')) 
|| (((_Value_ = true) 
  && ((_Bounded_ = true) 
  && (RUNTIME >= '+bE3.4' && RUNTIME <= '+bE5.2'))) 
  && (SCENE == 'fire'));' unknown, ambiguous or inaccessible method phrase
        at org.apache.commons.jexl2.Interpreter.call(Interpreter.java:1062)
        at org.apache.commons.jexl2.Interpreter.visit(Interpreter.java:1110)
        at datawave.query.jexl.DatawaveInterpreter.visit(DatawaveInterpreter.java:97)
        at org.apache.commons.jexl2.parser.ASTFunctionNode.jjtAccept(ASTFunctionNode.java:18)
        at org.apache.commons.jexl2.Interpreter.visit(Interpreter.java:1317)
        at datawave.query.jexl.DatawaveInterpreter.visit(DatawaveInterpreter.java:476)
        at org.apache.commons.jexl2.parser.ASTReference.jjtAccept(ASTReference.java:18)
        at datawave.query.jexl.DatawaveInterpreter.visit(DatawaveInterpreter.java:370)
        at org.apache.commons.jexl2.parser.ASTAndNode.jjtAccept(ASTAndNode.java:18)
        at datawave.query.jexl.DatawaveInterpreter.visit(DatawaveInterpreter.java:370)
        at org.apache.commons.jexl2.parser.ASTAndNode.jjtAccept(ASTAndNode.java:18)
        at datawave.query.jexl.DatawaveInterpreter.visit(DatawaveInterpreter.java:370)
        at org.apache.commons.jexl2.parser.ASTAndNode.jjtAccept(ASTAndNode.java:18)
        at datawave.query.jexl.DatawaveInterpreter.visit(DatawaveInterpreter.java:228)
        at org.apache.commons.jexl2.parser.ASTOrNode.jjtAccept(ASTOrNode.java:18)
        at org.apache.commons.jexl2.Interpreter.visit(Interpreter.java:946)
        at org.apache.commons.jexl2.parser.ASTJexlScript.jjtAccept(ASTJexlScript.java:38)
        at org.apache.commons.jexl2.Interpreter.interpret(Interpreter.java:232)
        at org.apache.commons.jexl2.ExpressionImpl.execute(ExpressionImpl.java:107)
        at datawave.query.function.JexlEvaluation.apply(JexlEvaluation.java:70)
        at datawave.query.function.JexlEvaluation.apply(JexlEvaluation.java:22)
        at datawave.query.iterator.QueryIterator$1.computeNext(QueryIterator.java:790)
        at datawave.query.iterator.QueryIterator$1.hasNext(QueryIterator.java:800)
        at com.google.common.collect.TransformedIterator.hasNext(TransformedIterator.java:42)
        at com.google.common.collect.TransformedIterator.hasNext(TransformedIterator.java:42)
        at com.google.common.collect.TransformedIterator.hasNext(TransformedIterator.java:42)
        at com.google.common.collect.TransformedIterator.hasNext(TransformedIterator.java:42)
        at com.google.common.collect.TransformedIterator.hasNext(TransformedIterator.java:42)
        at datawave.query.iterator.QueryIterator$1.computeNext(QueryIterator.java:788)
        at datawave.query.iterator.QueryIterator$1.hasNext(QueryIterator.java:800)
        at com.google.common.collect.TransformedIterator.hasNext(TransformedIterator.java:42)
        at datawave.query.iterator.pipeline.Pipeline.run(Pipeline.java:67)
        at datawave.query.iterator.pipeline.SerialIterator.startPipeline(SerialIterator.java:85)
        at datawave.query.iterator.QueryIterator.seek(QueryIterator.java:467)
        ... 11 more
```

keith-ratcliffe:
Not 100% positive that this is the ideal solution, but it does fix several examples of `ASTFunctionNode` eval misfire that we've encountered in testing. In each case, debugging revealed that the culprit was the index-only field(s) being omitted from the jexl context

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/function/IndexOnlyContextCreator.java
+++ b/warehouse/query-core/src/main/java/datawave/query/function/IndexOnlyContextCreator.java
@@ -4,13 +4,13 @@ import com.google.common.base.Function;
 import com.google.common.collect.Multimap;
 import datawave.query.Constants;
 import datawave.query.attributes.Document;
+import datawave.query.collections.FunctionalSet;
 import datawave.query.composite.CompositeMetadata;
 import datawave.query.iterator.IndexOnlyFunctionIterator;
 import datawave.query.iterator.QueryOptions;
 import datawave.query.jexl.DatawaveJexlContext;
 import datawave.query.jexl.DelayedNonEventIndexContext;
 import datawave.query.jexl.IndexOnlyJexlContext;
-import datawave.query.jexl.visitors.DelayedNonEventSubTreeVisitor;
 import datawave.query.jexl.visitors.IteratorBuildingVisitor;
 import datawave.query.jexl.visitors.SetMembershipVisitor;
 import datawave.query.planner.DefaultQueryPlanner;
@@ -125,6 +125,7 @@ public class IndexOnlyContextCreator extends JexlContextCreator {
         final Tuple3<Key,Document,DatawaveJexlContext> tuple;
         if (null != from) {
             tuple = super.apply(from);
+            verifyContext(tuple.third());
         } else {
             tuple = null;
         }
@@ -197,6 +198,22 @@ public class IndexOnlyContextCreator extends JexlContextCreator {
         }
         
         return newContext;
+    }
+    
+    /**
+     * Ensures that downstream expectations are met with regard to the state of the given context
+     * 
+     * @param context
+     *            jexl context to verify, assumed to be non-null
+     */
+    private void verifyContext(DatawaveJexlContext context) {
+        // Make sure that index-only fields being queried are represented in the context, in
+        // order to avoid potential downstream eval misfire (see #1204)
+        for (String var : this.variables) {
+            if (this.indexOnlyFields.contains(var) && !context.has(var)) {
+                context.set(var, FunctionalSet.empty());
+            }
+        }
     }
     
     /*
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/DelayedNonEventIndexContext.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/DelayedNonEventIndexContext.java
@@ -82,7 +82,7 @@ public class DelayedNonEventIndexContext extends DatawaveJexlContext {
         
         return delegate.get(name);
     }
-
+    
     /**
      * Use the IteratorBuildingVisitor limit to the current docRange to parse all delayed sub trees of the query. From those delayed sub trees initialize all
      * iterators matching the target field and aggregate all partial Documents into a list
@@ -106,7 +106,6 @@ public class DelayedNonEventIndexContext extends DatawaveJexlContext {
             
             // reset the root
             iteratorBuildingVisitor.resetRoot();
-
             
             // construct the index iterator for this node
             delayedNonEventNode.jjtAccept(iteratorBuildingVisitor, null);
@@ -155,27 +154,30 @@ public class DelayedNonEventIndexContext extends DatawaveJexlContext {
             }
         }
     }
-
+    
     @Override
     public boolean has(String name) {
         return delegate.has(name);
     }
-
+    
     @Override
     public void clear() {
         this.delegate.clear();
     }
-
+    
     @Override
     public int size() {
         return this.delegate.size();
     }
-
+    
     @Override
     public boolean equals(Object o) {
-        if (this == o) return true;
-        if (o == null || getClass() != o.getClass()) return false;
-        if (!super.equals(o)) return false;
+        if (this == o)
+            return true;
+        if (o == null || getClass() != o.getClass())
+            return false;
+        if (!super.equals(o))
+            return false;
         DelayedNonEventIndexContext that = (DelayedNonEventIndexContext) o;
         // @formatter:off
         return inclusive == that.inclusive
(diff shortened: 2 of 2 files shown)


jwomeara (approved):

apmoriarty (dismissed):
LGMT -- should this target integration?

keith-ratcliffe:
> LGMT -- should this target integration?

Thanks. Good call, it should. However, there's another wrinkle with the queries that helped to expose this bug. The new issue may be completely unrelated, but I'd like to keep this open until I can confirm/deny

With this patch in place, the affected queries now run without error, but we've discovered that they consistently return fewer hits than they should 

For example, a query such as....
```
INDEXED_FIELD:blah and (EVENT_ONLY:blahblah or INDEXONLY_TOKEN:"foo bar")

Hit Count: N
```
...returns fewer results than this query...
```
INDEXED_FIELD:blah and (EVENT_ONLY:blahblah)

Hit Count: N + C
Where C is a constant that represents missed documents from the first query. The docs in subset C are always the same every time the query is run
```
Obviously, removing an OR'ed phrase shouldn't increase the hit count. Without the patch from this PR, the first query fails in the exact same manner as described in #1204, and the second query gets the same (N + C) hits as above

The issue was discovered on our test cluster, but yesterday I was able to replicate and debug on the quickstart. Still debugging, trying to reduce it to something unit testable

keith-ratcliffe:
> However, there's another wrinkle with the queries that helped to expose this bug. The new issue may be completely unrelated, but I'd like to keep this open until I can confirm/deny
> 
After a bit more debugging, I can at least say for sure that the secondary issue is unrelated to phrase- or content-related query processing...

The original issue surfaced because `DatawaveInterpreter` couldn't find a suitable `ContentFunctions.phrase` method for the arguments given by the JexlContext. For the second issue, I note that the problematic query term need not be content-specific. It only needs to be an index-only term.

That said, it's probably ok to call #1204 complete at this point

keith-ratcliffe:
>...
> 
> The original issue surfaced because `DatawaveInterpreter` couldn't find a suitable `ContentFunctions.phrase` method for the arguments given by the JexlContext. For the second issue, I note that the problematic query term need not be content-specific. It only needs to be an index-only term.
> 
> That said, it's probably ok to call #1204 complete at this point

Rebased onto integration branch...
@jwomeara , @apmoriarty , if you guys can re-approve, then we can get this one closed. Will open another ticket for the second issue

apmoriarty (approved):

warehouse/query-core/src/main/java/datawave/query/function/IndexOnlyContextCreator.java
```diff
@@ -199,6 +200,22 @@ protected DatawaveJexlContext newDatawaveJexlContext(final Tuple3<Key,Document,M
...
+     * Ensures that downstream expectations are met with regard to the state of the given context
+     * 
+     * @param context
+     *            jexl context to verify, assumed to be non-null
+     */
+    private void verifyContext(DatawaveJexlContext context) {
+        // Make sure that index-only fields being queried are represented in the context, in
+        // order to avoid potential downstream eval misfire (see #1204)
+        for (String var : this.variables) {
+            if (this.indexOnlyFields.contains(var) && !context.has(var)) {
```
keith-ratcliffe, line 213: @apmoriarty , @jwomeara it turns out that this method introduced the new issue (#1227) observed on our test system. That seems to be due to an existing shortcoming of **DelayedNonEventIndexContext** ...

That is, if the context passed in here is of type **DelayedNonEventIndexContext**, then `!context.has(var)` is _always_ true. That's because the **has** method isn't overridden to operate on the internal `delegate` context, unlike its **get** and **set** methods which do. It just interrogates the internal map owned by the instance itself, and that map is apparently always empty.

So, an index-only term that was previously fetched/added to the delegate context, and which would have constituted a match, is now being replaced in the delegate with an empty set...and the hit is lost

Is there a reason we shouldn't override the **has** method in **DelayedNonEventIndexContext** to operate on the delegate?

keith-ratcliffe: Latest commit adds overrides for several methods to **DelayedNonEventIndexContext**

apmoriarty (approved):

jwomeara (approved):

PR #1211, opened 2021-06-29
closes #1166: Adds ability to query Geo fields with GeoWave functions.

### LINKED ISSUE #1166: Add ability for geowave functions to be run against geo type fields.
We can run geo functions against geowave and point type fields.  We should also add the ability to run geowave functions against geo type fields.

The basic approach would be to just get the bounding envelope for any WKT that is passed in, and run against the geo type index with that.  Additionally, the more advanced approach would update the geowave evaluation functions to compare the geo type point data with the WKT passed in via the geowave function.

jwomeara:


warehouse/query-core/src/main/java/datawave/query/jexl/functions/GeoWaveFunctions.java
```diff
@@ -56,19 +66,52 @@ private static boolean intersectsGeometries(Iterable<?> values, Geometry[] geome
...
+        return geom;
+    }
+    
+    private static Geometry parseGeometryFromGeoType(String geoTypeString) {
+        Geometry geom = null;
+        try {
+            geom = geoPointToGeometry(GeoNormalizer.isNormalized(geoTypeString) ? GeoNormalizer.GeoPoint.decodeZRef(geoTypeString) : GeoNormalizer.GeoPoint
+                            .decodeZRef(geoNormalizer.normalize(geoTypeString)));
+        } catch (Exception e) {
+            // do nothing
```
ivakegg, line 106: really?

ivakegg (changes requested):

ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GeoWaveFunctions.java
jwomeara: we swallow the exception during the initial AbstractGeometryNormalizer.parseGeometry call on line 93 so that we can attempt to parse as GeoType via this function.  If the result of both of those is a null geometry, we throw an exception.  Do you want me to log something?  There currently isn't a logger in this class (I'm assuming for evaluation performance reasons?)

ivakegg (approved):

hlgp (approved):

ivakegg (approved):

Issue #1208, opened 2021-06-28
Change use of JBoss MultivaluedMap in Query, QueryImpl


billoley:
In creating a query metrics microservice, the populateMetric method gets called from the W/S and passed a Query object.
A subsequent call to Query.getOptionalQueryParameters returns javax.ws.rs.core.MultivaluedMap requiring JBoss dependencies (jboss-jaxrs-api, etc) in the microservice which is less than ideal.

Both Query.toMap and Query.getOptionalQueryParameters currently return javax.ws.rs.core.MultivaluedMap.

We could either use org.springframework.util.MultiValueMap or just Map<String, List<String>>
Both javax.ws.rs.core.MultivaluedMap and org.springframework.util.MultiValueMap extend Map<K, List<V>>

PR #1213, opened 2021-06-29
re #1208 Change use of JBoss MultivaluedMap in Query, QueryImpl

billoley:


keith-ratcliffe (approved):
LGTM

jwomeara:
This is very similar to what I did on our queryMicroservices branch - with some slight differences here and there.  But it looks like you need this now vice whenever we finish with the query microservices.  Fair enough!  LGTM

jwomeara (approved):

keith-ratcliffe (approved):

PR #1226, opened 2021-07-19
closes #1225: add uid aggregator that will keep countOnly keys when 0…

### LINKED ISSUE #1225: Allow GlobalUidIndexAggregator to retain count-only entries even when 0 or negative
We have a use case where we know that we can issue too many subtractions for count-only protobufs in the shardIndex table. We'd like to have a configuration that allows us to retain count-only entries, even once their count is 0 or below (which would usually result in the keys being dropped). 

Update GlobalUidIndexAggregator to allow keeping these keys via a configuration. The default should be to not retain these keys to preserve behavior for current users of the aggregator.

jschmidt10:
… or below

### DIFF
--- a/warehouse/ingest-configuration/src/main/resources/config/shard-ingest-config.xml
+++ b/warehouse/ingest-configuration/src/main/resources/config/shard-ingest-config.xml
@@ -89,4 +89,8 @@
     <name>partitioner.category.member.${table.name.shard}</name>
     <value>shardedTables</value>
 </property>
+<property>
+    <name>index.tables.keep.count.only.entries</name>
+    <value>${indexTables.keepCountOnlyEntries}</value>
+</property>
 </configuration>
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/KeepCountOnlyUidAggregator.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/KeepCountOnlyUidAggregator.java
@@ -1,8 +1,10 @@
 package datawave.ingest.table.aggregator;
 
 /**
- * An extension of {@link GlobalIndexUidAggregator} that will always preserve "count only" Uid.List protobufs, even when their count goes below 1. This iterator
- * will prevent the accidental deletion of count only protobufs in case extra deletions occur.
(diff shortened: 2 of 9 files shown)


warehouse/ingest-configuration/src/main/resources/config/shard-ingest-config.xml
```diff
@@ -89,4 +89,8 @@
     <name>partitioner.category.member.${table.name.shard}</name>
     <value>shardedTables</value>
 </property>
+<property>
+    <name>shard.keep.count.only.index.entries</name>
```
jschmidt10: Elizabeth requested a rename here to `index.tables.keep.count.only.entries`. I'll make that update.

apmoriarty (dismissed):

apmoriarty (dismissed):

FineAndDandy (dismissed):

warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/KeepCountOnlyUidAggregator.java
```diff
@@ -0,0 +1,12 @@
+package datawave.ingest.table.aggregator;
+
+/**
+ * An extension of {@link GlobalIndexUidAggregator} that will always preserve "count only" Uid.List protobufs, even when their count goes below 1. This iterator
+ * will prevent the accidental deletion of count only protobufs in case extra deletions occur.
```
ivakegg: Actually the names of things are probably fine here.  In this comment here however explain that seenIgnore means that the Uid.List count exceeded the max.

ivakegg (dismissed):

warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/KeepCountOnlyUidAggregator.java
jschmidt10: Sounds good. Will update.

matthpeterson (approved):

ElizzyD (approved):

ivakegg (approved):

PR #1228, opened 2021-07-21
Allow options to be passed to aggregators

brianloss:
* Allow options to be specified for the aggregators that are used with
  TotalAggregatingIterator/PropogatingIterator and AggregatingReducer.
* Clean up IDE warnings and switch logging to slf4j (to clean up log
  statements) in PropogatingIterator.

Note: this is a subset of the work in #1091 -- just the part that allows options to be passed, since #1194 also adds an option that some users might want to set (though the default value for the option in that case preserves existing behavior, so setting it is not necessary)

### DIFF
--- a/warehouse/ingest-core/src/main/java/datawave/iterators/PropogatingIterator.java
+++ b/warehouse/ingest-core/src/main/java/datawave/iterators/PropogatingIterator.java
@@ -1,6 +1,7 @@
 package datawave.iterators;
 
 import com.google.common.base.Preconditions;
+import com.google.common.base.Splitter;
 import com.google.common.collect.Maps;
 import datawave.ingest.table.aggregator.PropogatingCombiner;
 import org.apache.accumulo.core.data.ArrayByteSequence;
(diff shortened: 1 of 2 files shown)


matthpeterson (approved):

warehouse/ingest-core/src/main/java/datawave/iterators/PropogatingIterator.java
```diff
@@ -303,32 +277,34 @@ public boolean validateOptions(Map<String,String> options) {
...
+            int sepIdx = value.indexOf(' ');
+            String aggClass = (sepIdx < 0) ? value : value.substring(0, sepIdx);
+            Object agg = createAggregator(aggClass);
             if (agg instanceof PropogatingCombiner) {
-                propAgg = PropogatingCombiner.class.cast(agg);
+                PropogatingCombiner propAgg = (PropogatingCombiner) agg;
+                if (sepIdx > 0) {
+                    String encodedOpts = value.substring(sepIdx + 1);
+                    Map<String,String> aggOpts = Splitter.on(';').trimResults().withKeyValueSeparator('=').split(encodedOpts);
+                    propAgg.validateOptions(aggOpts);
```
matthpeterson, line 292: Should a unit test be added for the parsing of encoded options?

jschmidt10 (approved):

Issue #1231, opened 2021-07-27
Create MatchesInGroupLeftRange


phrocker:
Allow matching within a group utilizing a range Function. 

This was done in an effort to improve composite regex and range handling.

This piggy backs on  #188

phrocker:
I have a draft PR for this

PR #1232, opened 2021-07-27
Add matches in Group Left Range function

phrocker:
Draft PR to support #1231. 

Will clean up and add tests as needed. Any insight would be appreciated!

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/composite/CompositeTerm.java
+++ b/warehouse/query-core/src/main/java/datawave/query/composite/CompositeTerm.java
@@ -1,5 +1,6 @@
 package datawave.query.composite;
 
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Sets;
 import datawave.data.type.DiscreteIndexType;
 import datawave.query.jexl.JexlASTHelper;
@@ -17,6 +18,9 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 import java.util.Set;
 
 /**
@@ -85,7 +89,17 @@ public class CompositeTerm extends Composite {
      * @return
      */
     private String getAppendedExpressions() {
-        return String.join(separator, expressionList);
+        final List<String> newExpressions;
+        if (Iterables.getLast(jexlNodeList) instanceof ASTERNode && jexlNodeList.stream().filter(x -> x instanceof ASTERNode).count() == 1) {
+            // We have an ASTERNode in the last position, but the rest are not ERNodes ( and therefore must be EQNodes )
+            // escape all of the terms
+            newExpressions = expressionList.stream().map(Pattern::quote).collect(Collectors.toList());
+            // set the last ( the regex ) so that it is not escaped.
+            newExpressions.set(expressionList.size() - 1, expressionList.get(expressionList.size() - 1));
+        } else {
+            newExpressions = expressionList;
+        }
+        return String.join(separator, newExpressions);
     }
     
     // what this essentially boils down to is that the only valid nodes are equals or equals regex nodes
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
@@ -3,22 +3,19 @@ package datawave.query.jexl.functions;
 import com.google.common.base.CharMatcher;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
+import datawave.data.normalizer.DateNormalizer;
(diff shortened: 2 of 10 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
```diff
@@ -310,6 +307,72 @@ public static boolean excludeRegex(Iterable<?> values, String regex) {
         return FunctionalSet.unmodifiableSet(matches);
     }
     
+    private static class NaturalComparator<T extends Comparable<T>> implements Comparator<T> {
```
phrocker, line 310: I did this quickly back in April but I'll replace this with a lib function.  As memory serves there is a natural comparator in commons ComparatorUtils so naturally I should use that....get it?

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
```diff
@@ -310,6 +307,69 @@ public static boolean excludeRegex(Iterable<?> values, String regex) {
...
+     *            Generic type of the bounds arguments.
+     * @return
+     */
+    public static <T> Collection<ValueTuple> getWithinInclusiveRange(ValueTuple fieldValue, T lowerBound, T upperBound) {
+        try {
+            if (Range.between(lowerBound, upperBound, new NaturalComparator()).contains((T) ValueTuple.getNormalizedValue(fieldValue))) {
+                return FunctionalSet.singleton(getHitTerm(fieldValue));
+            }
+        } catch (ClassCastException e) {
+            if (fieldValue.getNormalizedValue() instanceof BigDecimal) {
```
phrocker, line 337: I would love input here. The expectation is that if I have a range that can be converted ( if a string) we allow that range execution to occur, otherwise the query will produce an exception by way of the class cast. 

Not looking at this after a few months it no longer passes the sniff test, so input is very welcome.

warehouse/query-core/src/main/java/datawave/query/composite/CompositeTerm.java
```diff
@@ -85,7 +89,17 @@ public void getNodesAndExpressions(List<Class<? extends JexlNode>> nodeClasses,
      * @return
      */
     private String getAppendedExpressions() {
-        return String.join(separator, expressionList);
+        final List<String> newExpressions;
+        if (Iterables.getLast(jexlNodeList) instanceof ASTERNode && jexlNodeList.stream().filter(x -> x instanceof ASTERNode).count() == 1) {
```
phrocker, line 93: If I were reviewing this I think I might ask this to be broken into a separate PR. This originates from my test branch so this and the matches range function were within the same branch, so think I'll break this change out into its own PR.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GroupingRequiredFilterFunctions.java
```diff
@@ -311,6 +313,89 @@ private static void manageMatchesInGroupRemainingArgs(Object fieldValue, String
...
+     */
+    public static Collection<?> matchesInGroupLeftRange(Object... args) {
+        if (log.isTraceEnabled()) {
+            log.trace("matchesInGroupLeftRange(" + Arrays.asList(args) + ")");
+        }
+        // positionFrom is either '0', or it is the integer value of the last argument
+        // when the argumentCount is odd
+        int positionFromLeft = 0;
+        if (args.length % 3 != 0) {
+            Object lastArgument = args[args.length - 1];
```
phrocker, line 333: much of this is boiler plate parsing for the purposes of extracting the components of the matches in group fields and bounds....

warehouse/query-core/src/main/java/datawave/query/composite/CompositeTerm.java
phrocker: Broken out into https://github.com/NationalSecurityAgency/datawave/pull/1234 , so will prune this change set from this PR today.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
hlgp: Only just started looking at this, but my initial thoughts are that it might be worth considering adding functionality to the FunctionNormalizationRebuildingVisitor or some other FunctionRangeHandlingVisitor, since there you have access to the normalization information and can determine rangeablitiy, as it were, maybe some validation that the lowerBound is indeed less than the upperBound, etc., and somehow pass down the type information you need here.  Also that way we could fail up in Planning if we are not dealing with a rangeable type, and other functions we might want to add range functionality to will inherit the goodness.  Maybe that's hokey.  Maybe I've misinterpreted this idea.  Idk, It's after 9 on a Friday.

hlgp:
This is still a desired function but has gotten stale.  Reopen when revived.

Issue #188, opened 2018-08-28
Need to revisit support for composites created from regex nodes


jwomeara:
At the moment, our support for composites containing regex is incomplete.  We are actively prohibiting anyfield expansion from including composite fields, and in most cases we are disallowing composites to be formed if they would contain a regex node.  

In the future, if composites prove to be performant, then we should revisit our support for composites containing regex.  Special focus needs to be given to the concept of an overloaded composite containing a regex, as this is a trickier case than most.  

For now, we will stick with our limited support, but we can probably do better.

phrocker:
FWIW @jwomeara composite queries are pretty amazing. kudos to @tomnelson

PR #1234, opened 2021-07-28
Support Composite regex expansion combination

phrocker:
Works toward Issue #188 . 

Since this is a combined field with ASTEQNodes w/ ERNodes, this will quote the EQNodes so the Regex Analyzer won't treat the EQNodes as regular expression characters ( as they will be escaped ).

Attempts composite term expansion, followed by regex expansion, again followed by composite term expansion. For cases where the regex expansion can help provide eq nodes, we can then attempt composite terms again to locate an optimal index lookup for the composite fields specified in query.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandCompositeTerms.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandCompositeTerms.java
@@ -12,8 +12,11 @@ import datawave.query.composite.Composite;
 import datawave.query.composite.CompositeRange;
 import datawave.query.composite.CompositeTerm;
 import datawave.query.composite.CompositeUtils;
+import datawave.query.tables.ScannerFactory;
 import datawave.query.config.ShardQueryConfiguration;
+import datawave.query.exceptions.CannotExpandUnfieldedTermFatalException;
 import datawave.query.exceptions.DatawaveFatalQueryException;
+import datawave.query.exceptions.DoNotPerformOptimizedQueryException;
 import datawave.query.jexl.JexlASTHelper;
 import datawave.query.jexl.JexlNodeFactory;
 import datawave.query.jexl.LiteralRange;
@@ -22,6 +25,7 @@ import datawave.query.util.MetadataHelper;
 import datawave.webservice.common.logging.ThreadConfigurableLogger;
 import datawave.webservice.query.exception.DatawaveErrorCode;
 import datawave.webservice.query.exception.QueryException;
+import org.apache.accumulo.core.client.TableNotFoundException;
 import org.apache.commons.jexl2.parser.ASTAndNode;
 import org.apache.commons.jexl2.parser.ASTDelayedPredicate;
 import org.apache.commons.jexl2.parser.ASTEQNode;
@@ -30,6 +34,7 @@ import org.apache.commons.jexl2.parser.ASTEvaluationOnly;
 import org.apache.commons.jexl2.parser.ASTFunctionNode;
 import org.apache.commons.jexl2.parser.ASTGENode;
 import org.apache.commons.jexl2.parser.ASTGTNode;
+import org.apache.commons.jexl2.parser.ASTJexlScript;
 import org.apache.commons.jexl2.parser.ASTLENode;
 import org.apache.commons.jexl2.parser.ASTLTNode;
 import org.apache.commons.jexl2.parser.ASTNENode;
@@ -39,7 +44,7 @@ import org.apache.commons.jexl2.parser.ASTOrNode;
 import org.apache.commons.jexl2.parser.ASTReference;
 import org.apache.commons.jexl2.parser.ASTReferenceExpression;
 import org.apache.commons.jexl2.parser.JexlNode;
-import org.apache.log4j.Logger;
+import org.slf4j.Logger;
 
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -55,6 +60,8 @@ import java.util.Map.Entry;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import static org.slf4j.LoggerFactory.getLogger;
+
 import static org.apache.commons.jexl2.parser.JexlNodes.children;
 
 /**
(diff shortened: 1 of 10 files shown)


warehouse/query-core/src/test/java/datawave/query/jexl/visitors/ExpandCompositeTermsTest.java
```diff
@@ -95,6 +95,18 @@ public void test2() throws Exception {
...
+         * would result in [term][regex][term]
+         */
+        query = "WINNER=='blue' && TEAM=='gold' && NAME=~'gold-?.*' && POINTS==11";
+        runTestQuery(query, "WINNER == 'blue' && NAME =~ 'gold-?.*' && TEAM_POINTS == 'gold,11'");
+        /**
+         * As part of boundary value testing we are ensuring that for TEAM_POINTS we aren't combining these terms during composite expansion since that would
+         * result in [regex][term]
+         */
+        query = "WINNER=='blue' && TEAM=~'gol.*' && NAME=='gold-8' && POINTS==11";
+        runTestQuery(query, "WINNER == 'blue' && TEAM =~ 'gol.*' && NAME == 'gold-8' && POINTS == 11");
```
apmoriarty, line 109: Can we break these into separate test cases like below? (test4a, test4b, etc). Same comment for tests below

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -1218,32 +1218,21 @@ protected ASTJexlScript processTree(final ASTJexlScript originalQueryTree, Shard
...
-        
         if (!disableBoundedLookup) {
+            /**
+             * This will allow composite terms to be checked and expanded; however, there are cases where the expansion cannot occur. For example if you have
+             * three fields and the second and third fields are regex. We may be able to expand the second field and composite a field of A*B*C, where C is a
+             * regex with the ExpandCompositeTerms visitor, at which point ParallelIndexExpansion can then attempt the composite field expansion of A*B*C, and
+             * (hopefully) result in discrete terms.
+             */
+            queryTree = checkAndExpandCompositeTerms(config, queryTree, timers, scannerFactory, metadataHelper, expansionFields);
+
```
apmoriarty, line 1229: Can we get some test cases where the only component being tested is the query plan? Something similar to the [TreeFlatteningRebuildingVisitorTest](https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/query-core/src/test/java/datawave/query/jexl/visitors/TreeFlatteningRebuildingVisitorTest.java) or the [QueryModelVisitorTest](https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/query-core/src/test/java/datawave/query/jexl/visitors/QueryModelVisitorTest.java)

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandCompositeTerms.java
```diff
@@ -106,6 +113,71 @@ private ExpandCompositeTerms(ShardQueryConfiguration config) {
...
+     *            The jexl node representing the query
+     * @return An expanded version of the passed-in script containing composite nodes
+     */
+    @SuppressWarnings("unchecked")
+    public static <T extends JexlNode> T expandTermsWithRegex(ShardQueryConfiguration config, T script, ScannerFactory scannerFactory,
+                    MetadataHelper metadataHelper, Set<String> expansionFields) throws InstantiationException, IllegalAccessException, TableNotFoundException {
+        
+        ExpandCompositeTerms visitor = new ExpandCompositeTerms(config);
+        
+        // need to flatten the tree so i get all and nodes at the same level
```
apmoriarty:
```suggestion
        // CompositeTerm expansion requires a flattened tree
```

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandCompositeTerms.java
```diff
@@ -106,6 +113,71 @@ private ExpandCompositeTerms(ShardQueryConfiguration config) {
...
+                /**
+                 * Don't promote a tree that isn't executable. If that is the case resort to returning the tree created by the composite expansion.
+                 */
+                if (ExecutableDeterminationVisitor.isExecutable(expandedTree, config, indexedFields, indexOnlyFields, nonEventFields, null, metadataHelper)) {
+                    return (T) expandedTree.jjtAccept(visitor, new ExpandData());
+                }
+            } catch (CannotExpandUnfieldedTermFatalException cee) {
+                /**
+                 * Exceptions in the executor of ParallelIndexExpansion wrap other exceptions. In the case where we have a do not perform query optimization
+                 * exception we want to pass that along. Therefore we will check the cause and return the original DoNotPerformOptimizedQueryException.
```
apmoriarty: The ParallelIndexExpansion wraps execution exceptions with CannotExpandUnfieldedTermFatalExceptions. In the case of an exception while expanding regexes the original cause will be DoNotPerformOptimizedQueryException. If we failed to expand a regex, persist the original DoNotPerformOptimizedQueryException.

Or put another way "Persist the original exception if a regex term failed to expand"

apmoriarty (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandCompositeTerms.java
phrocker: Thanks for the comments. For the life of me I cannot understand why I chose such a verbose way to explain a very simple piece of code. Thanks for pointing this out. At least I can laugh at myself.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/ExpandCompositeTermsTest.java
phrocker: Thanks for confirming what I was assuming (but wanted validation ) : that numerics followed by letters are related tests.

jwomeara: Ha, I'm to blame for that.  There were so many test cases to be written that I couldn't think of a unique name for each one.

jwomeara (commented):
I don't have time to really sink my teeth into this right now, but I would like to have that opportunity before anything is approved/merged.

phrocker:
> 
> 
> I don't have time to really sink my teeth into this right now, but I would like to have that opportunity before anything is approved/merged.

Hey! I made the changes with tests that @apmoriarty requested, but had to "run out the door" so I probably won't commit that until I get back home late Sunday or early Monday ( at the earliest ) -- so no rush in my opinion.

hlgp:
This MR has gotten stale and we have no current need for expansions against composite indices.
We can reopen it when the work is prioritized, conflicts are resolved, and it needs active attention.

PR #1235, opened 2021-07-28
additional work to stop double wrapping nodes

apmoriarty:
The original implementation for removing extra reference expressions broke a few assumptions made by various visitors. 

- Added a visitor to fix reference expressions that do not have a parent reference node
- Reworked implementation of RemoveExtraParensVisitor
- Simplified delayed predicate pushdown

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EnsureReferenceNodesVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EnsureReferenceNodesVisitor.java
@@ -1,0 +1,381 @@
+package datawave.query.jexl.visitors;
+
+import org.apache.commons.jexl2.parser.ASTAdditiveNode;
+import org.apache.commons.jexl2.parser.ASTAdditiveOperator;
+import org.apache.commons.jexl2.parser.ASTAmbiguous;
+import org.apache.commons.jexl2.parser.ASTAndNode;
+import org.apache.commons.jexl2.parser.ASTArrayAccess;
+import org.apache.commons.jexl2.parser.ASTArrayLiteral;
+import org.apache.commons.jexl2.parser.ASTAssignment;
+import org.apache.commons.jexl2.parser.ASTBitwiseAndNode;
+import org.apache.commons.jexl2.parser.ASTBitwiseComplNode;
+import org.apache.commons.jexl2.parser.ASTBitwiseOrNode;
+import org.apache.commons.jexl2.parser.ASTBitwiseXorNode;
+import org.apache.commons.jexl2.parser.ASTBlock;
+import org.apache.commons.jexl2.parser.ASTConstructorNode;
+import org.apache.commons.jexl2.parser.ASTDivNode;
+import org.apache.commons.jexl2.parser.ASTEQNode;
+import org.apache.commons.jexl2.parser.ASTERNode;
+import org.apache.commons.jexl2.parser.ASTEmptyFunction;
+import org.apache.commons.jexl2.parser.ASTFalseNode;
+import org.apache.commons.jexl2.parser.ASTFloatLiteral;
+import org.apache.commons.jexl2.parser.ASTForeachStatement;
+import org.apache.commons.jexl2.parser.ASTFunctionNode;
+import org.apache.commons.jexl2.parser.ASTGENode;
+import org.apache.commons.jexl2.parser.ASTGTNode;
+import org.apache.commons.jexl2.parser.ASTIdentifier;
+import org.apache.commons.jexl2.parser.ASTIfStatement;
+import org.apache.commons.jexl2.parser.ASTIntegerLiteral;
+import org.apache.commons.jexl2.parser.ASTJexlScript;
+import org.apache.commons.jexl2.parser.ASTLENode;
+import org.apache.commons.jexl2.parser.ASTLTNode;
+import org.apache.commons.jexl2.parser.ASTMapEntry;
+import org.apache.commons.jexl2.parser.ASTMapLiteral;
+import org.apache.commons.jexl2.parser.ASTMethodNode;
+import org.apache.commons.jexl2.parser.ASTModNode;
+import org.apache.commons.jexl2.parser.ASTMulNode;
+import org.apache.commons.jexl2.parser.ASTNENode;
+import org.apache.commons.jexl2.parser.ASTNRNode;
+import org.apache.commons.jexl2.parser.ASTNotNode;
+import org.apache.commons.jexl2.parser.ASTNullLiteral;
+import org.apache.commons.jexl2.parser.ASTNumberLiteral;
+import org.apache.commons.jexl2.parser.ASTOrNode;
+import org.apache.commons.jexl2.parser.ASTReference;
+import org.apache.commons.jexl2.parser.ASTReferenceExpression;
+import org.apache.commons.jexl2.parser.ASTReturnStatement;
+import org.apache.commons.jexl2.parser.ASTSizeFunction;
+import org.apache.commons.jexl2.parser.ASTSizeMethod;
+import org.apache.commons.jexl2.parser.ASTStringLiteral;
+import org.apache.commons.jexl2.parser.ASTTernaryNode;
+import org.apache.commons.jexl2.parser.ASTTrueNode;
+import org.apache.commons.jexl2.parser.ASTUnaryMinusNode;
+import org.apache.commons.jexl2.parser.ASTVar;
+import org.apache.commons.jexl2.parser.ASTWhileStatement;
+import org.apache.commons.jexl2.parser.JexlNode;
+import org.apache.commons.jexl2.parser.JexlNodes;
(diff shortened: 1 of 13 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EnsureReferenceNodesVisitor.java
```diff
@@ -0,0 +1,381 @@
...
+import org.apache.commons.jexl2.parser.JexlNodes;
+import org.apache.commons.jexl2.parser.ParserTreeConstants;
+import org.apache.commons.jexl2.parser.SimpleNode;
+
+/**
+ * Some visitors rely on the presence of an {@link org.apache.commons.jexl2.parser.ASTReference} node in the AST.
+ *
+ * Ensure that all {@link org.apache.commons.jexl2.parser.ASTReferenceExpression} nodes have a parent ASTReference.
+ */
+public class EnsureReferenceNodesVisitor extends BaseVisitor {
```
FineAndDandy, line 64: Should this be a RebuildingVisitor?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PullupUnexecutableNodesVisitor.java
```diff
@@ -242,9 +242,20 @@ public Object visit(ASTNotNode node, Object data) {
...
+                if (node.jjtGetNumChildren() > 1) {
+                    log.warn("Tried to pull up a delayed marker that had more than one child.");
+                }
+                
+                // This swap should be safe, the reference node should have a single child that is a reference expression
+                JexlNode source = ASTDelayedPredicate.getDelayedPredicateSource(marker);
+                JexlNode wrappedSource = JexlNodes.wrap(source);
+                JexlNodes.swap(node, marker.jjtGetChild(0), wrappedSource);
+            }
+            return marker;
```
FineAndDandy, line 258: is this out of scope?

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/PullupUnexecutableNodesVisitorTest.java
```diff
@@ -0,0 +1,64 @@
...
+        test(query, expected);
+    }
+    
+    private void test(String query, String expected) throws ParseException {
+        ASTJexlScript script = JexlASTHelper.parseAndFlattenJexlQuery(query);
+        
+        ASTJexlScript visitedScript = (ASTJexlScript) PullupUnexecutableNodesVisitor.pullupDelayedPredicates(script, true, config, indexedFields,
+                        indexOnlyFields, nonEventFields, metadataHelper);
+        ASTJexlScript expectedScript = JexlASTHelper.parseJexlQuery(expected);
+        // assertTrue(TreeEqualityVisitor.isEqual(expectedScript, visitedScript));
```
FineAndDandy, line 59: delete

FineAndDandy (approved):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
```diff
@@ -92,6 +92,9 @@ public ExpandMultiNormalizedTerms(ShardQueryConfiguration config, MetadataHelper
             throw new DatawaveFatalQueryException(qe);
         }
         
+        // inject reference nodes when needed
+        script = (T) EnsureReferenceNodesVisitor.ensureReferences(script);
```
ivakegg, line 96: Why do we need this here?  Is a previous visitor breaking this contract?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PullupUnexecutableNodesVisitor.java
```diff
@@ -242,9 +242,20 @@ public Object visit(ASTNotNode node, Object data) {
...
+            // continue to pull up delayed markers in the odd case that a double delay occurred
+            JexlNode marker = node;
+            while (ASTDelayedPredicate.instanceOf(marker)) {
+                if (node.jjtGetNumChildren() > 1) {
+                    log.warn("Tried to pull up a delayed marker that had more than one child.");
+                }
+                
+                // This swap should be safe, the reference node should have a single child that is a reference expression
+                JexlNode source = ASTDelayedPredicate.getDelayedPredicateSource(marker);
+                JexlNode wrappedSource = JexlNodes.wrap(source);
```
ivakegg, line 255: Won't this simply add extra parens?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EnsureReferenceNodesVisitor.java
jwomeara: I concur.  If the plan is for a visitor to modify the tree, it should probably be a rebuilding visitor.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RangeConjunctionRebuildingVisitor.java
```diff
@@ -92,6 +92,9 @@ public RangeConjunctionRebuildingVisitor(ShardQueryConfiguration config, Scanner
                 throw new DatawaveFatalQueryException(qe);
             }
             
+            // inject reference nodes when needed
+            script = (T) EnsureReferenceNodesVisitor.ensureReferences(script);
```
ivakegg, line 96: Is some visitor breaking this contract such that we need to do this?

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EnsureReferenceNodesVisitor.java
```diff
@@ -0,0 +1,381 @@
...
+        return data;
+    }
+    
+    /*
+     * Short circuits
+     */
+    
+    @Override
+    public Object visit(SimpleNode node, Object data) {
+        node.childrenAccept(this, data);
```
jwomeara, line 155: How is this a short-circuit?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PullupUnexecutableNodesVisitor.java
```diff
@@ -242,9 +242,20 @@ public Object visit(ASTNotNode node, Object data) {
...
-            return source;
+            
+            // continue to pull up delayed markers in the odd case that a double delay occurred
+            JexlNode marker = node;
+            while (ASTDelayedPredicate.instanceOf(marker)) {
+                if (node.jjtGetNumChildren() > 1) {
+                    log.warn("Tried to pull up a delayed marker that had more than one child.");
+                }
+                
+                // This swap should be safe, the reference node should have a single child that is a reference expression
```
jwomeara, line 253: You check, but you have not guaranteed that the reference node has a single child.  Consider throwing an exception if you encounter a reference node with multiple children, or just remove the check & logging since this should never happen.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
apmoriarty: ExpandMultiNormalizedTerms can produce a reference expression without a reference node

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RangeConjunctionRebuildingVisitor.java
apmoriarty: The RangeConjunctionRebuildingVisitor can produce a reference expression without a reference node

jwomeara (commented):
I still think that instead of adding in a new visitor to remove extra parens that we should use this as a tool to identify existing visitors which are double-wrapping nodes, and fix those visitors instead.  Adding an extra visitor means more work to do during query planning and ultimately doesn't fix the root issue.  That being said, we have already merged in broken code, which this apparently fixes...  So, I will approve, but I ask that you create a ticket to fix the root cause, please.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RangeConjunctionRebuildingVisitor.java
ivakegg: But you call it before the visitor is actually called?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
ivakegg: But you call it before the visitor is actually called?

ivakegg:
We determined there is a better way to go about fixing the root issues.  We will revert the commit that has broken the code base.

PR #1243, opened 2021-08-09
Fixes #1068 - Java 11 - Build/deploy DW 4x/Accumulo 2.0.1/Hadoop 3.3.1/Wildfly 17.0.1.Final on JDK 11

### LINKED ISSUE #1068: Make DW compatible with Java 11
Future Accumulo releases (e.g., 2.1) will require Java 11 compatibility for compile and runtime. Thus, DW will need to support Java 11 as well

keith-ratcliffe:
~~TODO:~~
~~1. Fix docs build~~ (fixed)
~~2. Fix ModelBeanTest~~ (fixed)

### DIFF
--- a/contrib/datawave-quickstart/bin/services/datawave/test-web/tests/QueryMetrics.test
+++ b/contrib/datawave-quickstart/bin/services/datawave/test-web/tests/QueryMetrics.test
@@ -13,7 +13,7 @@ setCurlData \
         begin=19701231 \
         end=20500101 \
         pagesize=1 \
-        auths=PUBLIC \
+        auths=PRIVATE \
         columnVisibility=PRIVATE \
         query.syntax=JEXL
 
@@ -65,7 +65,7 @@ setCurlData \
         begin=19701231 \
         end=20500101 \
         pagesize=1 \
-        auths=PUBLIC \
+        auths=PRIVATE \
         columnVisibility=PRIVATE \
         query.syntax=JEXL
 
--- a/contrib/datawave-quickstart/bin/services/datawave/test-web/tests/QueryMetrics.test
+++ b/contrib/datawave-quickstart/bin/services/datawave/test-web/tests/QueryMetrics.test
@@ -13,7 +13,7 @@ setCurlData \
         begin=19701231 \
         end=20500101 \
         pagesize=1 \
-        auths=PRIVATE \
+        auths=PUBLIC \
         columnVisibility=PRIVATE \
         query.syntax=JEXL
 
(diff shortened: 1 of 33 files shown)


contrib/datawave-quickstart/bin/services/datawave/test-web/tests/QueryMetrics.test
```diff
@@ -13,7 +13,7 @@ setCurlData \
         begin=19701231 \
         end=20500101 \
         pagesize=1 \
-        auths=PUBLIC \
+        auths=PRIVATE \
```
ivakegg: This is a curious change required for Java 11... why is this needed?

keith-ratcliffe: Found that the test currently fails without the PRIVATE auth, technically a separate issue. Will address in a different PR

web-services/model/src/test/java/datawave/webservice/query/model/ModelBeanTest.java
```diff
@@ -91,8 +91,24 @@ public void setup() throws Exception {
...
+         *
+         * "javax.xml.bind.UnmarshalException: unexpected element (uri:"http://webservice.datawave/v1", local:"Model").
+         *    Expected elements are <{}Model>"
+         *
+         * As a workaround, we can force it to load here in the same manner as in jdk 8's Package.getPackageInfo(),
+         * though there's likely a "better" (i.e., module-focused) way of handling this in jdk 9+
+         *
+         * E.g., see https://stackoverflow.com/questions/52157040/
+         */
+        Class.forName("datawave.webservice.model.package-info", false, this.getClass().getClassLoader());
```
keith-ratcliffe, line 107: I note that other classes have the same JAXB setup but don't have this problem, such as [here](https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/query-core/src/test/java/datawave/query/cardinality/TestCardinalityConfiguration.java#L31) and [here](https://github.com/NationalSecurityAgency/datawave/blob/integration/web-services/client/src/test/java/datawave/webservice/query/metric/QueryMetricTest.java#L149).

I also note that PowerMockRunner seems to be the main contributing factor here, given that when `@RunWith(PowerMockRunner.class)` is commented out, the JAXB deserialization works fine as-is, no workaround required

jwomeara (dismissed):
LGTM

ivakegg (dismissed):

keith-ratcliffe:
OBE

PR #1244, opened 2021-08-10
Support setting SHARDS_AND_DAYS hint from options function

### LINKED ISSUE #1030: SHARDS_AND_DAYS hint will fail to parse without parens
For a query like (FOO == 'bar' && SHARDS_AND_DAYS = '20200101_1'), the shards and days hit will be trimmed.  It appears that the shards and days hint is required to be inside it's own parens to work.  Unless this is a JEXL language issue, parens should not be required.

lbschanno:
Allow users to optionally choose to specify a SHARDS_AND_DAYS hint via
the options function.

Fixes #1030

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/AddShardsAndDaysVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/AddShardsAndDaysVisitor.java
@@ -22,7 +22,7 @@ import java.util.stream.Collectors;
  */
 public class AddShardsAndDaysVisitor extends RebuildingVisitor {
     
-    private static final Joiner JOINER = Joiner.on(',');
+    private static final Joiner JOINER = Joiner.on(',').skipNulls();
     
     @SuppressWarnings("unchecked")
     public static <T extends JexlNode> T update(T node, String shardsAndDays) {
@@ -110,6 +110,8 @@ public class AddShardsAndDaysVisitor extends RebuildingVisitor {
         // If they were all duplicates, return the original string. Otherwise append them.
         if (this.shardsAndDays.isEmpty()) {
             return shardsAndDays;
+        } else if (shardsAndDays.isEmpty()) {
+            return JOINER.join(this.shardsAndDays);
         } else {
             return shardsAndDays + "," + JOINER.join(this.shardsAndDays);
         }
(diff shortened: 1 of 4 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/AddShardsAndDaysVisitor.java
```diff
@@ -0,0 +1,120 @@
...
+    
+    private static final Joiner JOINER = Joiner.on(',');
+    
+    @SuppressWarnings("unchecked")
+    public static <T extends JexlNode> T update(JexlNode node, String shardsAndDays) {
+        if (node == null) {
+            return null;
+        }
+        
+        T copy = (T) copy(node);
```
ivakegg: Do we really need to make a copy?  We can return the original tree if no changes are made, and the node.jjtAccept(visitor, node) call below will make a copy.  Am I missing something here?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/AddShardsAndDaysVisitor.java
```diff
@@ -0,0 +1,120 @@
...
+        // Remove any duplicate shards and days hints from the initial list.
+        String[] values = StringUtils.split(shardsAndDays, ',');
+        for (String value : values) {
+            this.shardsAndDays.remove(value);
+        }
+        // If they were all duplicates, return the original string. Otherwise append them.
+        if (this.shardsAndDays.isEmpty()) {
+            return shardsAndDays;
+        } else {
+            return shardsAndDays + "," + JOINER.join(this.shardsAndDays);
```
ivakegg, line 116: I know it is an edge case and perhaps will never happen, but if the shardsAndDays is the empty string, then this will result is a leading comma.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/AddShardsAndDaysVisitor.java
lbschanno: Good point, I have removed the unnecessary copy.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/AddShardsAndDaysVisitor.java
lbschanno: Good catch, I have added a fix to handle this edge case.

ivakegg (approved):

apmoriarty (approved):

Issue #1171, opened 2021-05-11
ContentFunctionEvaluator can produce false-positive matches


keith-ratcliffe:
Given a content query such as this...
```
FIELD1:"some phrase" AND FIELD2:"something else"~5
```
...ContentFunctionEvaluator  may return a hit on "some phrase" when it doesn't actually exist in FIELD1, but does exist in FIELD2

keith-ratcliffe:
Fix merged into 3.2 via 3a34e53

PR #1246, opened 2021-08-10
Support for adding full phrases to the HIT_TERM

apmoriarty:


FineAndDandy (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionEvaluator.java
```diff
@@ -218,15 +218,15 @@ public boolean evaluate() {
                             log.trace(logPrefix + " satisfied the content function");
                         }
                         
-                        return true;
+                        return field;
```
d-hwang: Instead of returning the field, we can potentially collect set of fields that satisfy the query "content:phrase((foo||bar), ...)", but obviously that may make things slower. LGTM otherwise.

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
```diff
@@ -353,6 +369,44 @@ private void addHits(Object fieldValue) {
...
+                        cvs.add(attr.getColumnVisibility());
+                    }
+                }
+                
+                ColumnVisibility cv;
+                try {
+                    cv = MarkingFunctionsFactory.createMarkingFunctions().combine(cvs);
+                } catch (MarkingFunctions.Exception e) {
+                    log.error("Failed to combine column visibilities while generating HIT_TERM for phrase function");
+                    e.printStackTrace();
```
d-hwang: log.error("message", e)?

d-hwang (dismissed):
Just some minor comments, but it looks good to me.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionEvaluator.java
apmoriarty: Updated to address this. Found out I'll have to backport the work done in https://github.com/NationalSecurityAgency/datawave/issues/1171 as part of this work.

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
apmoriarty: Done

d-hwang (dismissed):
LGTM!

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
```diff
@@ -341,6 +357,71 @@ private void addHits(Object fieldValue) {
...
+     * @param hitListArithmetic
+     *            a JexlArithmetic that supports hit lists
+     */
+    private void addHitsForFunction(String field, ASTFunctionNode node, HitListArithmetic hitListArithmetic) {
+        ColumnVisibility cv;
+        // aggregate individual hits for the content function
+        Collection<ColumnVisibility> cvs = new HashSet<>();
+        Attributes source = new Attributes(true);
+        
+        try {
```
jschmidt10: It looks like this try is only protecting the MarkingsFunction.combine(). We should shrink this try to just cover that.

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
```diff
@@ -341,6 +357,71 @@ private void addHits(Object fieldValue) {
...
+                    }
+                }
+            }
+            
+            cv = MarkingFunctionsFactory.createMarkingFunctions().combine(cvs);
+        } catch (MarkingFunctions.Exception e) {
+            log.error("Failed to combine column visibilities while generating HIT_TERM for phrase function for field [" + field + "]");
+            log.error("msg: ", e);
+            // set to the first column visibility if the marking functions failed
+            cv = cvs.iterator().next();
```
jschmidt10: This feels unsafe. Should we just fail if the combining does not work?

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
apmoriarty: Can do

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
apmoriarty: I think if we fail to combine markings we can just abort from this method and continue with the evaluation. Testing now..

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
```diff
@@ -341,6 +357,71 @@ private void addHits(Object fieldValue) {
...
+     *            a JexlArithmetic that supports hit lists
+     */
+    private void addHitsForFunction(String field, ASTFunctionNode node, HitListArithmetic hitListArithmetic) {
+        ColumnVisibility cv;
+        // aggregate individual hits for the content function
+        Collection<ColumnVisibility> cvs = new HashSet<>();
+        Attributes source = new Attributes(true);
+        
+        try {
+            Set<String> values = new ContentFunctionsDescriptor().getArgumentDescriptor(node).getHitTermValues();
```
jschmidt10: We could save this in a variable and reuse: `new ContentFunctionsDescriptor().getArgumentDescriptor(node)`. We also use it down below.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionEvaluator.java
```diff
@@ -180,6 +183,9 @@ public boolean evaluate() {
                 
                 // Iterate over each collection of offsets (grouped by field) and try to find one that satisfies the phrase/adjacency
                 for (String field : offsetsByField.keySet()) {
+                    if (!fields.isEmpty() && !fields.contains(field)) {
```
jschmidt10, line 187: I believe this is equivalent to just `if (!fields.contains(field))`

warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionEvaluator.java
```diff
@@ -145,10 +145,13 @@ protected boolean initialize() {
     /**
      * Evaluate whether there is an unordered set of terms that are within the defined distance.
      * 
-     * @return true if found, false otherwise
+     * @return the field that hit if found, boolean false otherwise
      */
-    public boolean evaluate() {
+    public Object evaluate() {
```
jschmidt10: Requiring a type check on a return result feels a little tricky/hard-to-understand. What do you think of either:

1) Returning an empty collection for the negative result
2) Making a ContentFunResult class that an `isFound` flag along with the `foundFields`

apmoriarty: Leaving this as-is keeps the code somewhat simplified. I don't want to add a whole class to handle the truth state of a content function, and the handling of the result object in the interpreter is relatively straightforward.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionEvaluator.java
apmoriarty: This should be `if(fields != null && !fields.contains(field))`, fix forthcoming

apmoriarty: Interesting things happen when messing with this line, I'm going to leave this as-is and defer to @keith-ratcliffe on this one.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionEvaluator.java
jschmidt10: I feel kinda strongly about this one. I don't know many other methods where I have to check the type of the return value, and typically, if you do, it's to check for a subclass - not for two entirely different types. Mind if we get another dev or two to weigh in? 

Another option would be to make this return the boolean again and cache the result in the object so you could pull it out if evaluate() is true.

apmoriarty: I'll look into option 1

jschmidt10 (approved):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionEvaluator.java
jschmidt10: Thanks man. Looks good!

drewfarris (approved):

PR #1251, opened 2021-08-20
Add ability to override query node processor in LuceneToJexlQueryParser

drewfarris:
When implementing query handling strategies that differ from what's implemented in `LuceneToJexlQueryParser` it is necessary to override the `CustomQueryNodeProcessorPipeline` which is currently hard-coded. 

This change introduces a `QueryNodeProcessorPipelineFactory` that can be overridden and injected into `LuceneToJexlQueryParser` to change the way in which lucene `QueryNode`'s will be transformed into JEXL expressions.


warehouse/query-core/src/test/java/datawave/query/language/parser/jexl/TestLuceneToJexlQueryParser.java
```diff
@@ -567,5 +585,20 @@ public void testParseRegexsWithEscapes() throws ParseException {
...
     }
+
+    private static class TestQueryNodeProcessorFactory extends QueryNodeProcessorFactory {
+        @Override
+        public QueryNodeProcessor create(QueryConfigHandler configHandler) {
+            return new QueryNodeProcessorPipeline(configHandler);
+        }
+    }
+
+    @Test
```
drewfarris, line 596: This new test demonstrates how `LuceneToJexlQueryParser` behavior changes after injecting an alternate `QueryNodeProcessorFactory` implementation. In this case, it uses the default lucene `QueryNodeProcessorPipeline` that does not perform field tokenization in the same way as the datawave `CustomerQueryNodeProcessorPipeline`.

ivakegg (dismissed):

billoley (dismissed):

apmoriarty (approved):

ivakegg (approved):

PR #1256, opened 2021-08-23
[5/5] Clean up of provisional VM update, final commit

Montana:
It's worth noting that I also have a staging environment for the build activated which will not build, but the production environment builds, as pictured below. 

Screen Shot 2021-08-23 at 8 30 19 AM

Passing build: https://app.travis-ci.com/github/Montana/datawave/builds/236004234

Screen Shot 2021-08-23 at 8 55 35 AM

In the five PR's I've made - the current progress outline is as follows 

* Better syntax highlighting 
* Faster, smoother build times 
* More options given to the Datawave build 
* Working build, more verbose
* More caching options 
* Less memory usage 
* Gives option of having `install-jdk.sh` with it having permissions already granted via `chmod u+x`

(I apologize in advance this wasn't in one PR, that being said this is now an updated `.travis.yml` that works!)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,73 @@
-dist: precise
+---
+dist: trusty
+sudo: true
+
+# Setting `language: minimal`, in this case it can be both, (either/or).
+
 language: java
-jdk:
-  - oraclejdk8
+  
+# My tips for NSA's Datawave Travis optimization. 
+ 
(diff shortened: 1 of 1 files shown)


hlgp:
Thank you for your trouble, Montana, but the .travis.yml file has been removed as of  3.42.0

PR #1261, opened 2021-08-24
Re: Issue #454 - added FullTableScanDisallowedException

### LINKED ISSUE #454: MaxExpansionIndexOnlyQueryTest.testMaxValueNegAnyField random failure
This test will randomly fail when the MaxExpansionThreshold marker is set and an ivarator is not  configured. The execution should throw an exception but it appears that a race condition is occurring that causes the query to return without any results instead of throwing an exception. The exceptions from each of the threads are still logged.

This same condition is performed on numerous other tests without a problem. So fa this is the only test case where the random error condition occurs.

Currently, the unit test is marked as ignored.

codebyneesh:
Updated the randomly failing testMaxValueNegAnyField

### DIFF
--- a/warehouse/query-core/src/test/java/datawave/query/MaxExpansionIndexOnlyQueryTest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/MaxExpansionIndexOnlyQueryTest.java
@@ -1,6 +1,7 @@
 package datawave.query;
 
 import datawave.query.exceptions.DatawaveFatalQueryException;
+import datawave.query.exceptions.FullTableScansDisallowedException;
 import datawave.query.testframework.AbstractFunctionalQuery;
 import datawave.query.testframework.AccumuloSetup;
 import datawave.query.testframework.CitiesDataType;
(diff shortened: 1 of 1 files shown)


warehouse/query-core/src/test/java/datawave/query/MaxExpansionIndexOnlyQueryTest.java
```diff
@@ -126,7 +125,7 @@ public void testMaxValueNegAnyField() throws Exception {
         try {
             runTest(query, expect);
             Assert.fail("exception expected");
-        } catch (RuntimeException re) {
+        } catch (FullTableScansDisallowedException e) {
```
ivakegg, line 128: I am quite confused here.  If we were catching RuntimeException before and this was only failing periodically then how does replacing what we are catching going to make this any better?  Wouldn't we still need to catch a RuntimeException?  Also if we are only periodically throwing a FullTableScansDisallowedException, then why is that not the exception always being thrown?

ivakegg (changes requested):

warehouse/query-core/src/test/java/datawave/query/MaxExpansionIndexOnlyQueryTest.java
codebyneesh: From what I saw, this test has been ignored since ~2019.  When I ran it, it was only throwing the full table scan required error. I figured between then and now, something has been updated that could've possibly impacted the periodic failure.

ivakegg: Ah, I see.  Thanks.

ivakegg (approved):

jschmidt10 (approved):

PR #1267, opened 2021-08-26
Fix lineage for FieldToFieldComparisonVisitor

lbschanno:
FieldToFieldComparisonVisitor does not return a query with a valid
lineage. Ensure that lineage is properly established, and add unit tests
to verify this.

Part of #880.

### DIFF
--- a/warehouse/query-core/src/test/java/datawave/query/jexl/visitors/FieldToFieldComparisonVisitorTest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/jexl/visitors/FieldToFieldComparisonVisitorTest.java
@@ -3,39 +3,102 @@ package datawave.query.jexl.visitors;
 import datawave.query.jexl.JexlASTHelper;
 import org.apache.commons.jexl2.parser.ASTEvaluationOnly;
 import org.apache.commons.jexl2.parser.ASTJexlScript;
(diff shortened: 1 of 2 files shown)


apmoriarty (approved):

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/FieldToFieldComparisonVisitorTest.java
```diff
@@ -3,39 +3,102 @@
...
+        // Verify the script is as expected, and has a valid lineage.
+        assertScriptEquality(visitedScript, expected);
+        assertLineage(visitedScript);
+        
+        // Verify the original script was not modified, and still has a valid lineage.
+        assertScriptEquality(originalScript, original);
+        assertLineage(originalScript);
+    }
+    
+    private void assertScriptEquality(ASTJexlScript actualScript, String expected) throws ParseException {
```
jschmidt10, line 92: These two functions (this and the one below) seem like generalized utilities for asserting queries are equal with valid lineage. This would be a good candidate some kind of test utility that could be reused across all of these lineage verification/fix PRs.

jschmidt10 (approved):

PR #1268, opened 2021-08-27
Verify lineage for NodeTransformVisitor

lbschanno:
Modify tests for NodeTransformVisitor to viery that it returns a query
with a valid lineage.

Part of #880.

### DIFF
--- a/warehouse/query-core/src/test/java/datawave/query/planner/rules/NodeTransformVisitorTest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/planner/rules/NodeTransformVisitorTest.java
@@ -3,12 +3,15 @@ package datawave.query.planner.rules;
 import datawave.query.config.ShardQueryConfiguration;
 import datawave.query.jexl.JexlASTHelper;
 import datawave.query.jexl.nodes.QueryPropertyMarker;
-import datawave.query.jexl.visitors.JexlStringBuildingVisitor;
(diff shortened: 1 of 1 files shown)


apmoriarty (approved):

warehouse/query-core/src/test/java/datawave/query/planner/rules/NodeTransformVisitorTest.java
```diff
@@ -65,18 +66,37 @@ private void testSimplify(String query, String expected) throws Exception {
...
+        assertScriptEquality(originalScript, original);
+        assertLineage(originalScript);
+        
+    }
+    
+    private void assertScriptEquality(ASTJexlScript actualScript, String expected) throws ParseException {
+        ASTJexlScript expectedScript = JexlASTHelper.parseJexlQuery(expected);
+        TreeEqualityVisitor.Comparison comparison = TreeEqualityVisitor.checkEquality(expectedScript, actualScript);
+        if (!comparison.isEqual()) {
+            log.error("Expected " + PrintingVisitor.formattedQueryString(expectedScript));
```
jschmidt10, line 93: I almost feel like I'd rather see this in as the unit test failure reason instead of `comparison.getReason()`. I'm not too in-tune with the query code but it looks like the `comparison.getReason()` will be the local difference. Is that more helpful for debugging/fixing than the Expected/Actual message here?

jschmidt10 (approved):

Issue #1272, opened 2021-09-01
ParallelIndexExpansion produces invalid query trees


apmoriarty:
ParallelIndexExpansion, FixUnfieldedTermsVisitor, and a method in the JexlNodeFactory all work together to produce an invalid query tree with respect to proper parentage.

lbschanno:
Just ran into a similar issue when testing RangeConjunctionRebuildingVisitor, which uses `JexlNodeFactory.createNodeTreeFromFieldsToValues()`.

apmoriarty:
This was addressed through multiple code changes

PR #1273, opened 2021-09-01
Address parentage in ParallelIndexExpansion and JexlNodeFactory.

apmoriarty:
Addresses https://github.com/NationalSecurityAgency/datawave/issues/1272

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
@@ -190,7 +190,7 @@ public class JexlNodeFactory {
                         childNode.jjtSetParent(wrappedChildNode);
                         
                         parentNode.jjtAddChild(wrappedChildNode, parentNodeChildCount);
-                        childNode.jjtSetParent(childNode);
+                        wrappedChildNode.jjtSetParent(parentNode);
                         
                         parentNodeChildCount++;
                     }
@@ -225,7 +225,7 @@ public class JexlNodeFactory {
                     childNode.jjtSetParent(wrappedChildNode);
                     
                     parentNode.jjtAddChild(wrappedChildNode, parentNodeChildCount);
-                    childNode.jjtSetParent(childNode);
+                    wrappedChildNode.jjtSetParent(parentNode);
                     
                     parentNodeChildCount++;
                 }
@@ -241,9 +241,7 @@ public class JexlNodeFactory {
                 JexlNodes.promote(parentNode, child);
                 return child;
             default:
-                JexlNode wrappedParentNode = wrap(parentNode);
-                parentNode.jjtSetParent(wrappedParentNode);
-                return wrappedParentNode;
+                return wrap(parentNode);
         }
     }
     
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
@@ -80,7 +80,7 @@ public class JexlNodeFactory {
      * @param containerType
      *            should we create OR nodes or AND nodes
      * @param isNegated
-     *            is this node negated per {@see ParallelIndexExpansion#isNegativeNode(JexlNode)}
+     *            is this node negated per ParallelIndexExpansion.isNegativeNode
      * @param original
      *            the original JexlNode
      * @param fieldsToValues
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
@@ -198,7 +198,7 @@ public class JexlNodeFactory {
             // create a set of nodes wrapping each pattern
             List<String> patterns = new ArrayList<>(fieldsToValues.getPatterns() == null ? new ArrayList<>() : fieldsToValues.getPatterns());
             if (patterns.isEmpty()) {
-                 return new ExceededValueThresholdMarkerJexlNode(buildUntypedNode(copy(original), field));
+                return new ExceededValueThresholdMarkerJexlNode(buildUntypedNode(copy(original), field));
             } else if (patterns.size() == 1) {
                 return new ExceededValueThresholdMarkerJexlNode(buildUntypedNode(copy(original), field, patterns.get(0)));
             } else {
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
@@ -125,17 +125,15 @@ public class JexlNodeFactory {
             JexlNode child = copy(original);
             children.add(child);
             // remove this entry from the fieldsToValues to avoid duplication
-            for (String identifier : JexlASTHelper.getIdentifierNames(original)) {
-                for (Object value : JexlASTHelper.getLiteralValues(original)) {
-                    fieldsToValues.remove(identifier, value);
-                }
+            Set<String> identifiers = JexlASTHelper.getIdentifierNames(original);
+            for (Object value : JexlASTHelper.getLiteralValues(original)) {
+                identifiers.forEach((id) -> fieldsToValues.remove(id, value));
             }
         }
         
         for (String field : fields) {
             JexlNode child = createNodeFromValuesForField(field, fieldsToValues, original, containerType, expandValues, isNegated);
             if (child != null) {
-                JexlASTHelper.validateLineage(child, true);
                 children.add(child);
             }
         }
@@ -165,6 +163,7 @@ public class JexlNodeFactory {
      * @param field
      *            create a node for this field
      * @param fieldsToValues
+     *            an {@link datawave.query.jexl.lookups.IndexLookup}
      * @param original
      *            the original JexlNode
      * @param type
@@ -172,6 +171,7 @@ public class JexlNodeFactory {
      * @param expandValues
      *            should we expand values
      * @param isNegated
+     *            boolean indicating if the original node is negated
      * @return a JexlNode
      */
     public static JexlNode createNodeFromValuesForField(String field, IndexLookupMap fieldsToValues, JexlNode original, ContainerType type,
@@ -181,7 +181,7 @@ public class JexlNodeFactory {
         
         if (!expandValues) {
             
-            // If not expanding values set the original node's field name to '_ANYFIELD_'
+            // If not expanding values replace the original node's '_ANYFIELD_' with the specified field
             JexlNode child = copy(original);
             for (ASTIdentifier identifier : JexlASTHelper.getIdentifiers(child)) {
                 if (identifier.image.equals(Constants.ANY_FIELD)) {
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
@@ -4,7 +4,6 @@ import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import datawave.query.Constants;
-import datawave.query.jexl.visitors.RebuildingVisitor;
 import datawave.query.exceptions.DatawaveFatalQueryException;
 import datawave.query.jexl.lookups.IndexLookupMap;
 import datawave.query.jexl.lookups.ValueSet;
@@ -58,6 +57,8 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Set;
 
+import static datawave.query.jexl.visitors.RebuildingVisitor.copy;
+
 /**
  * Factory methods that can create JexlNodes
  * 
@@ -101,7 +102,7 @@ public class JexlNodeFactory {
         
         // no expansions needed if the field name threshold is exceeded
         if (fieldsToValues.isKeyThresholdExceeded()) {
-            return new ExceededTermThresholdMarkerJexlNode(original);
+            return new ExceededTermThresholdMarkerJexlNode(copy(original));
         }
         
         // collapse the value sets if not expanding fields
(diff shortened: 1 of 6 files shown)


FineAndDandy (changes requested):
tests?

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
```diff
@@ -225,7 +225,7 @@ else if (1 == valuesForField.size()) {
                     childNode.jjtSetParent(wrappedChildNode);
```
lbschanno: An initial test I made

```java
   @Test
    public void testNoFieldExpansion() throws ParseException {
        IndexLookupMap indexLookupMap = new IndexLookupMap(10, 10);
        indexLookupMap.putAll("FOO", Sets.newHashSet("6", "9"));
        indexLookupMap.putAll("BAR", Sets.newHashSet("7", "8"));
    
        JexlNode originalNode = JexlASTHelper.parseJexlQuery("((_Bounded_ = true) && (FOO > 5 && FOO < 10))");
        JexlNode node = JexlNodeFactory.createNodeTreeFromFieldsToValues(JexlNodeFactory.ContainerType.OR_NODE, new ASTEQNode(ParserTreeConstants.JJTEQNODE),
                        originalNode, indexLookupMap, false, true, false);
        
        assertEquals("(FOO == '6' || FOO == '7' || FOO == '8' || FOO == '9')", JexlStringBuildingVisitor.buildQuery(node));
        assertTrue(JexlASTHelper.validateLineage(node, false));
    }
```
will still fail unless the line `childNode.jjtSetParent(wrappedChildNode);` here is deleted. Effectively, this line is setting the parent to the reference rather than the correct referenceexpression.

lbschanno (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
apmoriarty: Good catch, I'll take another pass through this

apmoriarty:
Did some refactoring, added some tests, found another source that produced invalid jexl nodes.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
```diff
@@ -108,142 +112,151 @@ public static JexlNode createNodeTreeFromFieldsToValues(ContainerType containerT
...
-            parentNode.jjtAddChild(child, parentNodeChildCount);
-            child.jjtSetParent(parentNode);
-            parentNodeChildCount++;
+            JexlNode child = copy(original);
+            children.add(child);
             // remove this entry from the fieldsToValues to avoid duplication
-            for (String identifier : JexlASTHelper.getIdentifierNames(orgNode)) {
-                for (Object value : JexlASTHelper.getLiteralValues(orgNode)) {
+            for (String identifier : JexlASTHelper.getIdentifierNames(original)) {
+                for (Object value : JexlASTHelper.getLiteralValues(original)) {
```
lbschanno: We can retrieve the values just once instead of for each identifier, like below:
```java
Set<String> identifiers = JexlASTHelper.getIdentifierNames(original);
for (Object value : JexlASTHelper.getLiteralValues(original)) {
    identifiers.forEach((id) -> fieldsToValues.remove(id, value));
}
```

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
```diff
@@ -108,142 +112,151 @@ public static JexlNode createNodeTreeFromFieldsToValues(ContainerType containerT
...
                 }
             }
-        } else {
-            JexlNodes.ensureCapacity(parentNode, fields.size());
         }
         
         for (String field : fields) {
-            ValueSet valuesForField = fieldsToValues.get(field);
+            JexlNode child = createNodeFromValuesForField(field, fieldsToValues, original, containerType, expandValues, isNegated);
+            if (child != null) {
```
lbschanno, line 136: Could we instead have a unit test for `JexlNodeFactory.createNodeFromValuesForField()` to verify that it always returns a node with a valid lineage (and fix it if it doesn't) rather than making this check in runtime here?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FixUnfieldedTermsVisitor.java
```diff
@@ -206,21 +206,23 @@ protected void visit(JexlNode node, JexlNode newNode, Object data) {
...
                         // Otherwise, we want to add the child regardless
                         newNode.jjtAddChild(newChild, newIndex);
+                        newChild.jjtSetParent(newNode);
                         newIndex++;
                     }
                 }
             } else {
                 
                 newNode.jjtAddChild(node.jjtGetChild(i), newIndex);
+                node.jjtSetParent(newNode);
```
lbschanno: Is this supposed to be `node.jjtGetChild(i).jjtSetParent(newNode);`?

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
apmoriarty: Nifty, I like it

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
```diff
@@ -89,16 +93,16 @@
...
         if (!expandFields && !expandValues) {
-            return orgNode;
+            return original;
         }
         
-        // no expansions needed if the fieldname threshold is exceeded
+        // no expansions needed if the field name threshold is exceeded
         if (fieldsToValues.isKeyThresholdExceeded()) {
-            return new ExceededTermThresholdMarkerJexlNode(orgNode);
+            return new ExceededTermThresholdMarkerJexlNode(copy(original));
```
FineAndDandy, line 105: why a copy vs the original here?

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
apmoriarty: Removing the lineage validation, that was leftover from testing. There is also a case where a null child would be returned, if the inputs are bad. I can work on improving code coverage.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
```diff
@@ -108,142 +112,151 @@ public static JexlNode createNodeTreeFromFieldsToValues(ContainerType containerT
...
             }
-        } else {
-            JexlNodes.ensureCapacity(parentNode, fields.size());
         }
         
         for (String field : fields) {
-            ValueSet valuesForField = fieldsToValues.get(field);
+            JexlNode child = createNodeFromValuesForField(field, fieldsToValues, original, containerType, expandValues, isNegated);
+            if (child != null) {
+                JexlASTHelper.validateLineage(child, true);
```
FineAndDandy: this is the only place I see we are failing hard in non-test code

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
```diff
@@ -108,142 +112,151 @@ public static JexlNode createNodeTreeFromFieldsToValues(ContainerType containerT
...
+        if (!expandValues) {
             
-            // if not expanding values, then reuse the original node with simply a new field name (anyfield only)
-            if (!expandValues) {
-                JexlNode child = RebuildingVisitor.copy(orgNode);
-                for (ASTIdentifier identifier : JexlASTHelper.getIdentifiers(child)) {
-                    if (identifier.image.equals(Constants.ANY_FIELD)) {
-                        identifier.image = field;
-                    }
+            // If not expanding values set the original node's field name to '_ANYFIELD_'
```
FineAndDandy: this looks like it replaced ANY_FIELD rather than setting it

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FixUnfieldedTermsVisitor.java
apmoriarty: I suspect your are correct. Interestingly enough, after dropping a System.exit(0) here the build still passes...so yay for test coverage

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
```diff
@@ -108,142 +112,151 @@ public static JexlNode createNodeTreeFromFieldsToValues(ContainerType containerT
...
+     * @param original
+     *            the original JexlNode
+     * @param type
+     *            create an ASTOrNode or an ASTAndNode
+     * @param expandValues
+     *            should we expand values
+     * @param isNegated
+     * @return a JexlNode
+     */
+    public static JexlNode createNodeFromValuesForField(String field, IndexLookupMap fieldsToValues, JexlNode original, ContainerType type,
```
FineAndDandy, line 177: can you add some info to this javadoc?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RangeConjunctionRebuildingVisitor.java
```diff
@@ -138,8 +138,8 @@ protected JexlNode expandIndexBoundedRange(LiteralRange range, ASTReference curr
             return RebuildingVisitor.copy(currentNode);
         }
         
-        JexlNode orNode = JexlNodeFactory.createNodeTreeFromFieldsToValues(JexlNodeFactory.ContainerType.OR_NODE, new ASTEQNode(ParserTreeConstants.JJTEQNODE),
-                        currentNode, fieldsToTerms, expandFields, expandValues, false);
+        JexlNode orNode = JexlNodeFactory.createNodeTreeFromFieldsToValues(JexlNodeFactory.ContainerType.OR_NODE, false, currentNode, fieldsToTerms,
```
FineAndDandy, line 139: Is there a need to track the negated state or is false always acceptable here?

warehouse/query-core/src/test/java/datawave/query/jexl/JexlNodeFactoryTest.java
```diff
@@ -0,0 +1,179 @@
...
+        IndexLookupMap indexLookupMap = new IndexLookupMap(10, 10);
+        indexLookupMap.putAll("FOO", Sets.newHashSet("6", "9"));
+        indexLookupMap.putAll("BAR", Sets.newHashSet("7", "8"));
+        
+        JexlNode originalNode = JexlASTHelper.parseJexlQuery("((_Bounded_ = true) && (FOO > 5 && FOO < 10))");
+        JexlNode node = JexlNodeFactory.createNodeTreeFromFieldsToValues(JexlNodeFactory.ContainerType.OR_NODE, false, originalNode, indexLookupMap, false,
+                        true, false);
+        
+        assertEquals("(FOO == '6' || FOO == '7' || FOO == '8' || FOO == '9')", JexlStringBuildingVisitor.buildQuery(node));
+        assertTrue(JexlASTHelper.validateLineage(node, false));
```
FineAndDandy, line 37: should invalid lineage fail hard here and in other new tests?

FineAndDandy (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
apmoriarty: I've removed this in the next commit, I was building and looking for failure cases.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
apmoriarty: can do

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RangeConjunctionRebuildingVisitor.java
apmoriarty: If we're expanding a negated node it determines how the resulting expansion is generated, so yes

warehouse/query-core/src/test/java/datawave/query/jexl/JexlNodeFactoryTest.java
apmoriarty: Failing hard throws an exception, validating a lineage with failHard=false let's us say "this assertion failed" not "JexlASTHelper threw an exception"

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
apmoriarty: Thanks! Fixed in next commit.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
apmoriarty: We can create multiple exceeded tern threshold markers from the original, and not copying screws up the node lineage

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
```diff
@@ -108,142 +112,151 @@ public static JexlNode createNodeTreeFromFieldsToValues(ContainerType containerT
...
+            JexlNode child = createNodeFromValuesForField(field, fieldsToValues, original, containerType, expandValues, isNegated);
+            if (child != null) {
+                children.add(child);
+            }
+        }
+        
+        switch (children.size()) {
+            case 0:
+                // in this case we had no matches for the range, so this expression gets replaced with a FALSE node.
+                return new ASTFalseNode(ParserTreeConstants.JJTFALSENODE);
```
ivakegg, line 144: Is this correct if isNegated?  So I think we would be called if we have something like _ANYFIELD != bla.  If there are no matches for bla then this expression should result in true.  Please validate my logic here however.

ivakegg (changes requested):

apmoriarty:
Given the work done in #1307, I'm inclined to close this merge request and redo any work with respect to node parentage

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
apmoriarty: You are correct, a negated regex with zero hits should not be pruned from the query. I'm taking another pass through this work and validated your logic.

PR #1274, opened 2021-09-03
Refactor DefaultQueryPlanner into discrete methods

apmoriarty:
This is a proof of concept merge request. The goal is to eliminate boilerplate code in the DefaultQueryPlanner while also providing more options to override and extend specific operations.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
+++ b/warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
@@ -675,25 +675,10 @@ public class DefaultQueryPlanner extends QueryPlanner implements Cloneable {
         
         queryTree = timedFixNotNullIntent(timers, queryTree);
         
-        stopwatch = timers.newStartedStopwatch("DefaultQueryPlanner - include date filters");
+        queryTree = timedIncludeDateFilters(timers, queryTree, config, metadataHelper, scannerFactory, dateIndexHelper, settings);
         
-        try {
-            queryTree = addDateFilters(queryTree, scannerFactory, metadataHelper, dateIndexHelper, config, settings);
-        } catch (TableNotFoundException e1) {
-            throw new DatawaveQueryException("Unable to resolve date index", e1);
-        }
-        if (log.isDebugEnabled()) {
-            logQuery(queryTree, "Query after adding date filters:");
-        }
-        
-        stopwatch.stop();
-        
-        stopwatch = timers.newStartedStopwatch("DefaultQueryPlanner - cap date range");
-        
-        // note this must be called after we do the date adjustements per the query date type in addDateFilters
-        capDateRange(config);
-        
-        stopwatch.stop();
+        // note this must be called after we do the date adjustments per the query date type in addDateFilters
+        timedCapDateRange(timers, config);
         
         // Find unmarked bounded ranges
         if (UnmarkedBoundedRangeDetectionVisitor.findUnmarkedBoundedRanges(queryTree)) {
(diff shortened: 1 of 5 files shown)


FineAndDandy (dismissed):

jwomeara:
I think you're on to something good here.  You should push forward with the other visitors and add in configuration which allows us to conditionally run validity checks after each visitor is run (e.g. like validate lineage).  I see that the roots of that are already there.  keep going!

jwomeara (changes requested):
I would encapsulate the stuff after the processTree call as well.  I know that it's a little bit heavier weight, but in my opinion, that's even more of a reason to pull it into a separate function.  One of the goals of this update in my mind would be to make it easier to create an extended version of this class.  Encapsulating those other blocks of code would help with that.  

Also, consider making TimedVisitorManager a standalone, package-private class in the same package as the DefaultQueryPlanner.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -961,142 +859,43 @@ protected ASTJexlScript processTree(final ASTJexlScript originalQueryTree, Shard
...
+                        getTransformRules());
         
         // Find unfielded terms, and fully qualify them with an OR of all fields
         // found in the index
         // If the max term expansion is reached, then the original query tree is
         // returned.
         // If the max regex expansion is reached for a term, then it will be
         // left as a regex
-        Set<String> expansionFields = null;
+        Set<String> expansionFields = loadExpansionFields(config);
```
FineAndDandy: this should only be executed if !disabledAnyFieldLookup

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -961,142 +859,43 @@ protected ASTJexlScript processTree(final ASTJexlScript originalQueryTree, Shard
...
+            timedTestForNonExistentFields(timers, queryTree, config, metadataHelper, queryModel, settings);
         }
         
         // apply the node transform rules
         // running it here before any regex or range expansions to enable potentially pushing down terms before index lookups
-        queryTree = applyNodeTransformRules(queryTree, getTransformRules(), config, metadataHelper, "Pre regex/range expansions");
+        queryTree = timedApplyNodeTransformRules(timers, "Apply Node Transform Rules - Pre Regex/Range Expansions", queryTree, config, metadataHelper,
+                        getTransformRules());
         
         stopwatch = timers.newStartedStopwatch("DefaultQueryPlanner - Fetch required dataTypes");
```
FineAndDandy: add timer encapsulation for this and other steps that aren't getting the tree updates

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -1280,35 +1007,14 @@ protected ASTJexlScript processTree(final ASTJexlScript originalQueryTree, Shard
...
-                    innerStopwatch = timers.newStartedStopwatch("DefaultQueryPlanner - Expand pushing functions into exceeded value ranges");
-                    queryTree = PushFunctionsIntoExceededValueRanges.pushFunctions(queryTree, metadataHelper, config.getDatatypeFilter());
-                    if (log.isDebugEnabled()) {
-                        logQuery(queryTree, "Query after expanding pushing functions into exceeded value ranges:");
-                    }
-                    innerStopwatch.stop();
+                    queryTree = timedPushFunctions(timers, queryTree, config, metadataHelper);
                 }
                 
                 innerStopwatch = timers.newStartedStopwatch("DefaultQueryPlanner - Remove delayed predicates");
```
FineAndDandy: hit the other tree expansions below

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -2272,6 +2281,93 @@ private RangeStream initializeRangeStream(ShardQueryConfiguration config, Scanne
         }
     }
     
+    /**
+     * Handles boilerplate code execution for operations like timing a visit call, logging the query tree, and eventualy validating the resulting query tree.
+     */
+    public class TimedVisitorManager {
```
FineAndDandy: can this be pulled into its own class?

FineAndDandy (changes requested):

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
apmoriarty: done

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
apmoriarty: done.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
apmoriarty: done

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
apmoriarty: done

apmoriarty:
Rebased to address merge conflict introduced by https://github.com/NationalSecurityAgency/datawave/pull/1244, extracted that code into a timed method call.

jwomeara (dismissed):
Thank you for doing this!  Good work!

FineAndDandy (approved):

ivakegg:
I think this is great.  My only comment is that naming each method starting with "timed" seems excessive.  Other than that this appears good.

ivakegg (approved):

Issue #1285, opened 2021-09-20
Add user check for start/stop scripts


sbgift-dev:
Adding user check to prevent running start/stop scripts as root

PR #1287, opened 2021-09-21
#1285 Add user checks to start/stop-all scripts

sbgift-dev:


sbgift-dev:
This PR is showing all the commits I fetched from upstream - only the last commit from me is relevant

keith-ratcliffe:
> This PR is showing all the commits I fetched from upstream - only the last commit from me is relevant

Not sure what happened, but we wouldn't want to merge this in its current state. Is `integration` the intended target for your changes? If so, I'd recommend renaming your local branch, recreate the local `fix/issue-1285` by branching off of the latest integration...then cherry-pick your one commit and force push that your existing remote

warehouse/ingest-scripts/src/main/resources/bin/system/start-all.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+if [[ "$USER" != "datawave" ]]; then
```
keith-ratcliffe, line 3: Generally speaking, I don't think we want the DW code to require a specific `$USER`, since it limits portability and represents a breaking change for any existing DW deployment that doesn't adhere to the requirement.

I could maybe see disallowing all scripts to execute as `root`, since that's portable, and it most likely represents an inadvertent error on the part of the user. But, if we do that, it should really apply to all user scripts, not just _start-all.sh_/_stop-all.sh_

keith-ratcliffe (changes requested):

PR #1290, opened 2021-09-22
Fixes issue 678: attempt to record query plan in metrics when create fails

matthpeterson:


warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -2248,6 +2271,19 @@ else if (IvaratorRequiredVisitor.isIvaratorRequired(queryTree) && !config.canHan
         return new Tuple2<>(ranges, needsFullTable);
     }
     
+    private void attemptToRecordPlan(ShardQueryConfiguration config, JexlNode queryTree) {
+        if (null == this.plannedScript) {
```
ivakegg: with this mechanism, the query tree from only the first call with actually set the query string.  Don't we want the query string updated each time?

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
matthpeterson: I considered updating the query string each time it's updated but thought it might an expense we do not want to incur.  That's why I opted for setting it only if there was an error.  I'm open to changing the approach if the reviewers agree.

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
```diff
@@ -173,6 +174,9 @@ public void setConnection(Connector connection) throws Exception {
             // TODO: applyPrediction("Plan");
         } catch (Exception e) {
             log.error(e.getMessage(), e);
+            if (this.logic instanceof BaseQueryLogic) {
```
jwomeara:
```suggestion
            if (this.logic instanceof BaseQueryLogic && ((BaseQueryLogic)this.logic).getCollectQueryMetrics()) {
```

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
```diff
@@ -173,6 +174,9 @@ public void setConnection(Connector connection) throws Exception {
...
+                this.getMetric().setPlan(((BaseQueryLogic) this.logic).getConfig().getQueryString());
```
jwomeara: I would add a null check on the getMetric call

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -2248,6 +2271,19 @@ else if (IvaratorRequiredVisitor.isIvaratorRequired(queryTree) && !config.canHan
...
+            try {
+                this.plannedScript = JexlStringBuildingVisitor.buildQuery(queryTree);
+            } catch (Exception e) {
+                if (log.isTraceEnabled()) {
+                    log.trace("Unable to set plannedScript.");
+                }
+            }
+        }
+        config.setQueryString(this.plannedScript);
```
jwomeara: Shouldn't we update the plan with each call?  This is only going to update the planned script if it is null, which it won't be after the first time it's set, right?  Am I missing something?

jwomeara: Personally, I would simplify this and just call config.setQueryString(JexlStringBuildingVisitor.buildQuery(queryTree)) after each visitor visits the tree.  You might also want to add some kind of indicator which makes it clear that this is the partial query plan, and not the final query plan.

warehouse/query-core/src/test/java/datawave/query/testframework/AbstractFunctionalQuery.java
```diff
@@ -398,11 +403,15 @@ protected void runTestQuery(Collection<String> expected, String queryStr, Date s
...
-            log.debug("Plan: " + config.getQueryString());
+        if (useRunningQuery) {
+            QueryMetricFactory queryMetricFactory = (metricFactory == null) ? new QueryMetricFactoryImpl() : metricFactory;
+            new RunningQuery(connector, AccumuloConnectionFactory.Priority.NORMAL, this.logic, q, "", principal, queryMetricFactory);
+        } else {
+            GenericQueryConfiguration config = this.logic.initialize(connector, q, this.authSet);
+            this.logic.setupQuery(config);
+            if (log.isDebugEnabled()) {
+                log.debug("Plan: " + config.getQueryString());
+            }
```
jwomeara, line 414: I need to dig into this a little deeper, but at first glance I don't love the idea of modifying AbstractFunctionalQuery to support testing, assuming that is what's going on here.

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
jwomeara: Not all BaseQueryLogics support metrics.

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
jwomeara: Maybe not... given that we don't seem to be doing that anywhere else.  Seems error-prone though.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
jwomeara: Looking some more I see that you only update the query plan on a known failure.  Instead of doing that I would just update it every step of the way, and know that you will be covered if an exception is thrown at any point in the planner.

hlgp: Sounds reasonable

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
ivakegg: Basically what we will get here is merely a jexl translation of the original query.  What I think we want to see is the query plan just prior to the failure.  I think the approach needs to be changed here.

matthpeterson: RE: original query
This method only gets called if there's an error.  If the create is successful, plannedScript is initialized at the end of planning.  If an error occurs, this method will be called and it'll use the current state of queryTree.  The test cases demonstrate that this often reflects changes to the original query.
The null check probably isn't necessary here as it is almost always going to be null.

Do you want me to remove the null check?  Or would you rather plannedScript get updated after each modification to queryTree?

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
matthpeterson: I considered updating it on each change to the queryTree but thought it might be too expensive.  If you both think that's acceptable, I'll change it to work that way.

warehouse/query-core/src/test/java/datawave/query/testframework/AbstractFunctionalQuery.java
matthpeterson: Despite its name, AbstractFunctionalQuery exists solely for testing.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
matthpeterson: @jwomeara and @hlgp agreed that it would be worthwhile to update the planned script at each step along the way, which I'm understanding to mean every time the queryTree is modified.  I'll change it to do that.

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
matthpeterson: Good suggestion, thanks.

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
matthpeterson: Good catch!

matthpeterson:
Thanks for the feedback, @ivakegg, @jwomeara and @hlgp!  I think I've addressed all the comments.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
jwomeara: printing the tree is orders of magnitude faster than looking things up in the index.  i doubt we will notice.  however, perhaps it is worth creating a ticket to have each visitor return a boolean saying whether or not the tree was modified by the visitor.  if it wasn't modified, then you could skip updating the plan string.  that's a bigger task though, obviously.  worth creating a ticket?

warehouse/query-core/src/test/java/datawave/query/testframework/AbstractFunctionalQuery.java
jwomeara: ah, got it.  my mistake.  the name definitely threw me.  proceed as-is then.  thanks!

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -1422,6 +1467,11 @@ protected ASTJexlScript processTree(final ASTJexlScript originalQueryTree, Shard
         return queryTree;
     }
     
+    private void recordPlan(ASTJexlScript queryTree, ShardQueryConfiguration config) {
+        this.plannedScript = JexlStringBuildingVisitor.buildQuery(queryTree);
```
drewfarris: Is there an understanding of how expensive it might be to call buildQuery multiple times in the query planning process? As we're calling `recordPlan` many places throughout the code, and I'm assuming we ignore the result of the previous `buildQuery` call each time `recordPlan` is updated, we might consider a strategy where we call `buildQuery` in exception handling code only when it's needed? If `buildQuery` is cheap, no worries.

apmoriarty: I like the idea of only recording the query when the exception is thrown. Doing so will be far easier once https://github.com/NationalSecurityAgency/datawave/pull/1274 is merged in.

matthpeterson: @drewfarris and @apmoriarty, I agree with that concern and that's why I originally implemented the bugfix to avoid making so many calls to buildQuery.  All three of yesterday's reviewers disagreed so I changed the implementation.  

@ivakegg, @jwomeara and @hlgp do you happen to have an understanding of the relative expense of calling buildQuery after each step of the planning?  If not, perhaps I could measure it to better inform everyone.

jwomeara: buildQuery is already being called multiple times throughout the planning process.  The cost of calling that method will vary depending on the number of nodes in the tree (which will vary from query to query), but I think it's generally a fast operation.

Here's a thought...  Why not add a queryTree variable to ShardQueryConfiguration instead of having a local variable within the DefaultQueryPlanner:process method.  Update the variable in ShardQueryConfiguration after each phase of the  planning process, and then if an exception gets thrown anywhere in the process, you can build the query when the exception is caught in RunningQuery using the last updated queryTree in ShardQueryConfiguration.  Everybody wins!

ivakegg: @jwomeara has a good idea there.  This seems reasonable to me.

matthpeterson: Done.

Note that RunningQuery is in datawave-ws-query which does not depend on the datawave-query-core so doesn't have access to JexlStringBuildingVisitor.  As such, I put the call to JexlStringBuildingVisitor.buildQuery in ShardQueryConfiguration.  I'm open to suggestions on this approach.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -707,7 +708,7 @@ protected ASTJexlScript updateQueryTree(ScannerFactory scannerFactory, MetadataH
         stopwatch.stop();
         
         // Find unmarked bounded ranges
-        if (UnmarkedBoundedRangeDetectionVisitor.findUnmarkedBoundedRanges(queryTree)) {
+        if (UnmarkedBoundedRangeDetectionVisitor.findUnmarkedBoundedRanges(config.getQueryTree())) {
             throw new DatawaveFatalQueryException("Found incorrectly marked bounded ranges");
         }
```
hlgp, line 703: need a config.getQueyrTree on line 718 instead of old reference to queryTree, right?

matthpeterson: Good catch, thanks.  The github "resolve conflicts" brought these new lines in.  I just updated them and don't see any other additions that need updating.

hlgp (dismissed):

ivakegg (dismissed):

hlgp (dismissed):

jwomeara (dismissed):
I like it!  Nice work!

matthpeterson:
There were a lot of conflicts with https://github.com/NationalSecurityAgency/datawave/pull/1274.  I took a first shot at resolving them.

jwomeara (approved):
Phew...  Lots of getting and setting in there.  You probably could have just kept the local queryTree and called ShardQueryConfiguration.setQueryTree() after each method.  Not a big enough gripe for me not to approve though.

matthpeterson:
@jwomeara, @hlgp, and @ivakegg can you re-review?  The merge conflicts dismissed your prior approvals.

hlgp (approved):

ivakegg (approved):

PR #1294, opened 2021-09-23
Fixes #1293 Update StatsJob to read from NumShards and use round robi…

### LINKED ISSUE #1293: StatsJob reading incorrect shard setting and using suboptimal partitioning
The StatsJob is currently only utilizing about 1/3 of its reducers due to partitioning. We should switch to the RR partitioner that we use for normal ingest jobs.

Additionally, StatsJob is reading the legacy num.shards config item when determining the number of shards. This is leads to only processing a subset of the daily stats.

jschmidt10:


warehouse/ingest-configuration/src/main/resources/config/shard-stats-config.xml
```diff
@@ -65,4 +65,10 @@
     <value>24</value>
     <description>hyperlog value for sparse precision</description>
   </property>
+
+  <property>
+    <name>partitioner.default.delegate</name>
```
matthpeterson: This is duplicative with ingest-config.xml and I think that ingest jobs will load both.  That could lead to confusion if either of them changes in the future.  Perhaps the stats job could be changed to load the configs differently.

jschmidt10: Cool. We talked offline but I'll put the notes here for others.

We're going to separate out the shard-stats-config.xml into a subdirectory. That'll exclude it from normal ingest since ingest doesn't care about any of these properties. Then we'll update buildIndexStats.sh to load it's configs from the subdir, which for now, will just contain the one config file.

ivakegg (dismissed):

matthpeterson (dismissed):

jschmidt10:
After testing this out - the separate config directory is not picked up by generateSplitsFile.sh leading to the StatsJob failing. Need to go back to the drawing board. I had tested it before but it was while the previous splits file was still in place.

matthpeterson (dismissed):

jschmidt10:
Finally picked this back up. I ended up putting a partitioner override in the start script for the process and having the job pull in shard-stats-config.xml and shard-ingest-config.xml. This has been working for us.

avgAGB (approved):

apmoriarty (approved):

PR #1271, opened 2021-09-01
Better handling of ASTDelayedPredicate

apmoriarty:
Addresses https://github.com/NationalSecurityAgency/datawave/issues/1270

PR #1297, opened 2021-09-24
Fix reparented query property marker bug

### LINKED ISSUE #603: Getting the source node of a marked node should not reparent those nodes
Discovered a bug where getting the source node for a marked node was actually reparenting those child nodes to an orphaned, and node.  So, while the original parent still has knowledge of his children, the children think they have a new parent.

lbschanno:
Currently when the sources are retrieved for a query property marker via
the QueryPropertyMarkerVisitor, if the marker has multiple sources, they
are returned as children of a new AND node and are subsequently
reparented.

Modify the source retrieval code so that the source nodes are not
reparented as children of the new AND node. Additionally, create
visitors to fix query property marker nodes with multiple sources and
to validate query trees to ensure that they do not have multiple
sources.

Fixes #603

NOTE: per the original ticket, I also create a visitor that will validate a query tree to make sure it doesn't have any query markers with multiple sources, but wasn't sure where we would want this used. I would appreciate input on this.


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidQueryPropertyMarkerVisitor.java
```diff
@@ -0,0 +1,38 @@
...
+
+import datawave.query.exceptions.DatawaveFatalQueryException;
+import datawave.query.jexl.nodes.QueryPropertyMarker;
+import org.apache.commons.jexl2.parser.ASTAndNode;
+import org.apache.commons.jexl2.parser.JexlNode;
+
+/**
+ * This visitor validates that all query property marker nodes present have a singular root source node.
+ */
+public class ValidQueryPropertyMarkerVisitor extends BaseVisitor {
```
apmoriarty: I'm going to make a few requests for this one.
- create a package under datawave/query/jexl/visitors called 'validate'
- move this class to the validate package
- instead of throwing an exception return a boolean true/false indicating if the script is valid

Once https://github.com/NationalSecurityAgency/datawave/pull/1274 is merged we can add this into the chain of validating visitors that are called

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -729,6 +730,18 @@ protected ASTJexlScript updateQueryTree(ScannerFactory scannerFactory, MetadataH
...
+        try {
+            queryTree = QueryPropertyMarkerSourceConsolidator.consolidate(queryTree);
+        } catch (Exception e) {
+            throw new DatawaveQueryException("Failed to fix query property markers with multiple sources", e);
+        }
+        if (log.isDebugEnabled()) {
+            logQuery(queryTree, "Query after fixing query property markers with multiple sources");
+        }
+        stopwatch.stop();
+
```
apmoriarty: If #1274 is merged before this one, extract this to a `timedFixQueryPropertyMarkers()` method call

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java
```diff
@@ -3,28 +3,80 @@
...
+     * @return a new marker instance
+     */
+    public static <MARKER extends QueryPropertyMarker> MARKER create(JexlNode source, Function<JexlNode,MARKER> constructor) {
+        JexlNode parent = source.jjtGetParent();
+        MARKER marker = constructor.apply(source);
+        if (parent != null) {
+            JexlNodes.replaceChild(parent, source, marker);
+        }
+        return marker;
+    }
```
apmoriarty, line 59: It's currently possible to apply the same marker multiple times to a node and this is a case we need to handle.  See https://github.com/NationalSecurityAgency/datawave/pull/1271 for example unit tests. I think we can add an `isSourceAlreadyMarked()` method here.

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java
```diff
@@ -287,12 +332,57 @@ public boolean isIvarator() {
...
+                return sources.get(0);
+            } else if (totalSources > 1) {
+                // If multiple source nodes are found, return them as children of a new AND node.
+                ASTAndNode andNode = new ASTAndNode(ParserTreeConstants.JJTANDNODE);
+                JexlNodes.ensureCapacity(andNode, totalSources);
+                for (int i = 0; i < totalSources; i++) {
+                    andNode.jjtAddChild(sources.get(i), i);
+                    // The parents of the source nodes are intentionally not updated to the new AND node.
+                }
+                return andNode;
```
apmoriarty, line 396: Not for this merge request, but something to consider in the future. If I have a query like `(A || B)` and both terms end up getting delayed, it might be neat to represent that as
 `( _delayed_ && (A or B))` 
instead of 
`( _delayed_ && A) || ( _delayed_ && B)`

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPropertyMarkerSourceConsolidator.java
```diff
@@ -0,0 +1,42 @@
...
+import datawave.query.jexl.JexlNodeFactory;
+import datawave.query.jexl.nodes.QueryPropertyMarker;
+import org.apache.commons.jexl2.parser.ASTAndNode;
+import org.apache.commons.jexl2.parser.JexlNode;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * This visitor will ensure that query marker nodes always have a singular, root source node.
```
apmoriarty: Can we add something to indicate what is happening? "If a marker node has multiple sources they will be bundled within an ASTAndNode" or something like that

warehouse/query-core/src/test/java/datawave/query/jexl/nodes/QueryPropertyMarkerTest.java
```diff
@@ -29,11 +30,11 @@ public void testFindInstance() throws ParseException {
         assertEmptyInstance(findInstance("((_Bounded_ = true) && (FOO > 1 && FOO < 2)) && ((_EvalOnly_ = true) && (FOO == 1 && BAR == 2))"));
         
         // Test unwrapped marker and unwrapped sources
-        assertInstance(findInstance("(_Delayed_ = true) && FOO == 1 && BAR == 2"), ASTDelayedPredicate.class, "(BAR == 2 && FOO == 1)");
+        assertInstance(findInstance("(_Delayed_ = true) && FOO == 1 && BAR == 2"), ASTDelayedPredicate.class, "(FOO == 1 && BAR == 2)");
         // Test unwrapped marker and wrapped sources
         assertInstance(findInstance("(_Delayed_ = true) && (FOO == 1 && BAR == 2)"), ASTDelayedPredicate.class, "FOO == 1 && BAR == 2");
         // Test wrapped marker and unwrapped sources
-        assertInstance(findInstance("((_Delayed_ = true) && FOO == 1 && BAR == 2)"), ASTDelayedPredicate.class, "(BAR == 2 && FOO == 1)");
+        assertInstance(findInstance("((_Delayed_ = true) && FOO == 1 && BAR == 2)"), ASTDelayedPredicate.class, "(FOO == 1 && BAR == 2)");
```
apmoriarty: In the spirit of "you broke it you bought it" and "you touched it last", do you mind putting single quotes around the values in these queries? e.g., `FOO == '1'`

(sorry)

warehouse/query-core/src/test/java/datawave/query/jexl/nodes/QueryPropertyMarkerTest.java
```diff
@@ -57,13 +58,13 @@ public void testInstance_of() throws ParseException {
     public void testInstance_isAnyType() throws ParseException {
         assertFalse(QueryPropertyMarker.Instance.of().isAnyType());
         
-        QueryPropertyMarker.Instance instance = QueryPropertyMarker.Instance.of(ASTDelayedPredicate.class, parseJexlQuery("FOO == 'a'"));
+        QueryPropertyMarker.Instance instance = QueryPropertyMarker.Instance.of(ASTDelayedPredicate.class, Lists.newArrayList(parseJexlQuery("FOO == 'a'")));
```
apmoriarty: This implies we could have a chained constructor for Instance.of()
```
Instance.of(JexlNode node){
    return of(Lists.newArrayList(node))
}
...
Instance.of(Collection<JexlNode> nodes)
```

apmoriarty (changes requested):

warehouse/query-core/src/test/java/datawave/query/jexl/nodes/QueryPropertyMarkerTest.java
lbschanno: Haha, no problem!

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPropertyMarkerSourceConsolidator.java
lbschanno: Sure thing.

warehouse/query-core/src/test/java/datawave/query/jexl/nodes/QueryPropertyMarkerTest.java
lbschanno: Added!

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidQueryPropertyMarkerVisitor.java
lbschanno: Can do.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
lbschanno: Okay, no problem.

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java
lbschanno: I'm happy to work on this if you make a ticket for it.

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java
lbschanno: I can do this, but if we want to return the original source node, the return type for the `QueryPropertyMarker.create()` methods will have to be `JexlNode` rather than a `MARKER` specific type, since I can anticipate problems if we try to cast the source to a `MARKER` type. So instead of 
```java
BoundedRange node = BoundedRange.create(source);
```
we would have
```java
JexlNode node = BoundedRange.create(source);
```
If you're okay with that, I can make this change.

apmoriarty: The more I think about it the more I dislike my original suggestion. I'll work the 'already marked' case in 1271 after this is merged.

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java
apmoriarty: For future consideration, not something to be worked in the short term. https://github.com/NationalSecurityAgency/datawave/issues/1304

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
lbschanno: This has been extracted to `timedFixQueryPropertyMarkers()`.

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPropertyMarkerSourceConsolidator.java
```diff
@@ -0,0 +1,43 @@
...
+    public Object visit(ASTAndNode node, Object data) {
+        QueryPropertyMarker.Instance instance = QueryPropertyMarkerVisitor.getInstance(node);
+        // Only modify the node if it is a query property marker with multiple source nodes.
+        if (instance.isAnyType() && instance.hasMutipleSources()) {
+            // Make safe copies of the sources and wrap them in an AND node.
+            List<JexlNode> copiedSources = instance.getSources().stream().map(RebuildingVisitor::copy).collect(Collectors.toList());
+            JexlNode source = JexlNodeFactory.createAndNode(copiedSources);
+            // Create a new query property marker of the same instance.
+            JexlNode marker = QueryPropertyMarker.create(source, instance.getType());
+            return JexlASTHelper.dereference(marker);
```
ivakegg, line 38: Seems a waste to create a referenced marker just to dereference it.  Just saying.  This works of course.

ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPropertyMarkerSourceConsolidator.java
lbschanno: It's kind of hard to avoid creating a referenced marker given that `QueryPropertyMarker` extends `ASTReference`. I don't see a way to avoid this without duplicating the constructor logic in here.

ivakegg (approved):

apmoriarty (approved):

PR #1298, opened 2021-09-27
Added visitor to validate junction nodes

apmoriarty:
For use when https://github.com/NationalSecurityAgency/datawave/pull/1274 is merged

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/JunctionValidatingVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/JunctionValidatingVisitor.java
@@ -27,7 +27,6 @@ import org.apache.commons.jexl2.parser.ASTGTNode;
 import org.apache.commons.jexl2.parser.ASTIdentifier;
 import org.apache.commons.jexl2.parser.ASTIfStatement;
 import org.apache.commons.jexl2.parser.ASTIntegerLiteral;
-import org.apache.commons.jexl2.parser.ASTJexlScript;
 import org.apache.commons.jexl2.parser.ASTLENode;
 import org.apache.commons.jexl2.parser.ASTLTNode;
 import org.apache.commons.jexl2.parser.ASTMapEntry;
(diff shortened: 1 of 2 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/JunctionValidatingVisitor.java
```diff
@@ -0,0 +1,387 @@
...
+    
+    private boolean isValid = true;
+    
+    public static boolean validate(JexlNode script) {
+        JunctionValidatingVisitor visitor = new JunctionValidatingVisitor();
+        script.jjtAccept(visitor, null);
+        return visitor.getIsValid();
+    }
+    
+    public boolean getIsValid() {
```
jschmidt10: Convention here would be `isValid()`

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/JunctionValidatingVisitor.java
```diff
@@ -0,0 +1,387 @@
...
+        return this.isValid;
+    }
+    
+    /*
+     * Pass through
+     */
+    
+    @Override
+    public Object visit(ASTJexlScript node, Object data) {
```
jschmidt10: Same as parent impl.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/JunctionValidatingVisitor.java
```diff
@@ -0,0 +1,387 @@
...
+        if (!isJunctionValid(node)) {
+            return data;
+        }
+        node.childrenAccept(this, data);
+        return data;
+    }
+    
+    @Override
+    public Object visit(ASTAndNode node, Object data) {
+        if (!isJunctionValid(node)) {
```
jschmidt10: Just a thought here, let me know what you think.

If I read this method, it's not clear that the isValid flag is getting set unless I jump into `isJunctionValid()`. The `isJunctionValid()` method is doing double-duty. It both checks if the node is valid and has a side-effect. If you broke those two apart, this method could be:

```
  if (!isValid)
    return data;
  else if (!isJunctionValid(node)) {
    isValid = false;
    return data;
  }
  // recurse down tree
}
```

Then `isJunctionValid()` only has a single job of checking a node's validity. It's cool if you wanna skip this but wanted to mention it.

FineAndDandy (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/JunctionValidatingVisitor.java
apmoriarty: Unlike #1299 the validate methods are the same for and/or nodes, going to extract this code to a common method. Good call!

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/JunctionValidatingVisitor.java
apmoriarty: Removed in next commit

jschmidt10 (approved):

alerman (approved):
LGTM

PR #1299, opened 2021-09-27
Validate minimal reference expressions

apmoriarty:
Added visitor to validate minimal reference expressions within a query tree.

For use when https://github.com/NationalSecurityAgency/datawave/pull/1274 is merged.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/MinimalReferenceExpressionsVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/MinimalReferenceExpressionsVisitor.java
@@ -1,0 +1,431 @@
+package datawave.query.jexl.visitors.validate;
+
+import datawave.query.jexl.JexlASTHelper;
+import datawave.query.jexl.nodes.QueryPropertyMarker;
+import datawave.query.jexl.visitors.BaseVisitor;
+import datawave.query.jexl.visitors.JexlStringBuildingVisitor;
+import org.apache.commons.jexl2.parser.ASTAdditiveNode;
+import org.apache.commons.jexl2.parser.ASTAdditiveOperator;
+import org.apache.commons.jexl2.parser.ASTAmbiguous;
+import org.apache.commons.jexl2.parser.ASTAndNode;
+import org.apache.commons.jexl2.parser.ASTArrayAccess;
+import org.apache.commons.jexl2.parser.ASTArrayLiteral;
+import org.apache.commons.jexl2.parser.ASTAssignment;
+import org.apache.commons.jexl2.parser.ASTBitwiseAndNode;
+import org.apache.commons.jexl2.parser.ASTBitwiseComplNode;
+import org.apache.commons.jexl2.parser.ASTBitwiseOrNode;
+import org.apache.commons.jexl2.parser.ASTBitwiseXorNode;
+import org.apache.commons.jexl2.parser.ASTBlock;
+import org.apache.commons.jexl2.parser.ASTConstructorNode;
+import org.apache.commons.jexl2.parser.ASTDivNode;
+import org.apache.commons.jexl2.parser.ASTEQNode;
+import org.apache.commons.jexl2.parser.ASTERNode;
+import org.apache.commons.jexl2.parser.ASTEmptyFunction;
+import org.apache.commons.jexl2.parser.ASTFalseNode;
+import org.apache.commons.jexl2.parser.ASTFloatLiteral;
+import org.apache.commons.jexl2.parser.ASTForeachStatement;
+import org.apache.commons.jexl2.parser.ASTFunctionNode;
+import org.apache.commons.jexl2.parser.ASTGENode;
+import org.apache.commons.jexl2.parser.ASTGTNode;
+import org.apache.commons.jexl2.parser.ASTIdentifier;
+import org.apache.commons.jexl2.parser.ASTIfStatement;
+import org.apache.commons.jexl2.parser.ASTIntegerLiteral;
+import org.apache.commons.jexl2.parser.ASTJexlScript;
+import org.apache.commons.jexl2.parser.ASTLENode;
+import org.apache.commons.jexl2.parser.ASTLTNode;
+import org.apache.commons.jexl2.parser.ASTMapEntry;
+import org.apache.commons.jexl2.parser.ASTMapLiteral;
+import org.apache.commons.jexl2.parser.ASTMethodNode;
+import org.apache.commons.jexl2.parser.ASTModNode;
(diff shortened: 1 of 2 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/MinimalReferenceExpressionsVisitor.java
```diff
@@ -0,0 +1,431 @@
...
+        isValid = false;
+        log.info("Invalid node (" + cause + "): " + JexlStringBuildingVisitor.buildQueryWithoutParse(node));
+    }
+    
+    /*
+     * Pass through
+     */
+    
+    @Override
+    public Object visit(ASTJexlScript node, Object data) {
```
jschmidt10: These are the same implementation as the parent class. These can just be removed I think.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/MinimalReferenceExpressionsVisitor.java
```diff
@@ -0,0 +1,431 @@
...
+        return data;
+    }
+    
+    /*
+     * Short circuits
+     */
+    
+    @Override
+    public Object visit(ASTBlock node, Object data) {
+        return data;
```
jschmidt10, line 232: It looks like we could let these delegate to the parent implementations as well. Though here we're short circuiting I guess. Is that the intent? Are these parts of JEXL we know we don't utilize? 

Not saying we do it here, but if there is a subset of the language we're not supporting, we might bubble these "no-ops" up into our BaseVisitor.

jschmidt10: Oh, just saw the comment above. Missed that initially.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/MinimalReferenceExpressionsVisitor.java
```diff
@@ -0,0 +1,431 @@
...
+    }
+    
+    @Override
+    public Object visit(ASTOrNode node, Object data) {
+        if (isValid && isAnyChildAWrappedSingleTerm(node)) {
+            // found a single term child that is wrapped
+            invalidate(node, "wrapped single child");
+            return data;
+        }
+        node.childrenAccept(this, data);
```
jschmidt10: Do you want to short circuit here if `isValid` is false?

jschmidt10: Actually, I like your explicit short-circuit in the And method below. I'd recommend you use that same pattern here. It makes it clear that we are short-circuiting. I'd even say do it the same way in your Reference method above. That way it's the same across all of them.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/MinimalReferenceExpressionsVisitor.java
```diff
@@ -0,0 +1,431 @@
...
+    /*
+     * Operations
+     */
+    
+    @Override
+    public Object visit(ASTReference node, Object data) {
+        
+        if (isValid && isParen(node) && isParen(grandChild(node))) {
+            // found a double paren
+            invalidate(node, "double paren");
```
jschmidt10: The invalidate messages probably deserve an Enum or minimally to be constants.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/MinimalReferenceExpressionsVisitor.java
```diff
@@ -0,0 +1,431 @@
...
+     * Helper methods
+     */
+    
+    // is this node a parentheses
+    private boolean isParen(JexlNode node) {
+        return node instanceof ASTReference && node.jjtGetNumChildren() == 1 && node.jjtGetChild(0) instanceof ASTReferenceExpression;
+    }
+    
+    // fetch the grandchild
+    private JexlNode grandChild(JexlNode node) {
```
jschmidt10: Might be worth clarifying `getFirstGrandchild` or `getLeftmostGrandchild` or something? Just to specify that you are targeting a specific one out of possibly many.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/MinimalReferenceExpressionsVisitor.java
apmoriarty: Works for me. I'll leave the comment in for which nodes are explicitly pass-through

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/MinimalReferenceExpressionsVisitor.java
apmoriarty: Good call, I'll clean this up

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/MinimalReferenceExpressionsVisitor.java
apmoriarty: Really like this idea!

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/MinimalReferenceExpressionsVisitor.java
apmoriarty: Worth doing, thanks for pointing this out!

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/MinimalReferenceExpressionsVisitor.java
apmoriarty: Actually, I'll remove the visit for ASTJexlScript and modify the others to only continue recursing through the tree if it's still valid. Opportunity to short circuit on reference expressions, simple nodes, and not nodes

jschmidt10: Cool. Sounds good.

FineAndDandy (commented):
I'm ready to approve once @jschmidt10's comments are addressed

jschmidt10 (approved):

FineAndDandy (approved):

PR #1308, opened 2021-10-08
add locality groups to config cache

hlgp:


contrib/datawave-quickstart/bin/services/datawave/install-ingest.sh
```diff
@@ -141,7 +141,7 @@ fi
 function initializeDatawaveTables() {
     # create tables
     if [[ ${DW_DATAWAVE_SKIP_CREATE_TABLES} != true ]]; then
-        info "creating datawave tables and splits ..."
+        info "creating datawave tables, splits, and caches ..."
```
ivakegg, line 144: oh my

warehouse/ingest-core/src/main/java/datawave/ingest/config/TableConfigCache.java
```diff
@@ -48,17 +49,55 @@ public void update() {
...
+            if (table == null || !table.equals(parts[0])) {
+                table = parts[0];
+                tempMap = new HashMap<>();
+                this.configMap.put(table, tempMap);
+            }
+            if (parts.length == 3) {
+                propName = parts[1];
+                propVal = parts[2];
+                tempMap.put(propName, propVal);
+            }
```
ivakegg, line 84: Perhaps we should be failing if the file does not match our format as that would imply its the wrong file or something failed silently while writing.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/MultiRFileOutputFormatter.java
```diff
@@ -97,7 +91,7 @@
     protected String extension;
     protected Configuration conf;
     protected Map<String,ConfigurationCopy> tableConfigs;
-    protected Map<String,String> tableIds = null;
+    protected String[] tableIds = null;
```
ivakegg: Can we make this a Set so that containership is O(1)

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/TableConfigurationUtil.java
```diff
@@ -338,51 +351,94 @@ private void configureTablesIfNecessary(String[] tableNames, TableOperations top
...
                 return;
             } catch (Exception e) {
-                log.error("Unable to read accumulo config cache at " + cache.getCacheFilePath() + ". Proceeding to read directly from Accumulo.");
+                log.error("Unable to read accumulo config cache at " + tableConfigCache.getCacheFilePath() + ". Proceeding to read directly from Accumulo.");
             }
         }
-        Map<String,String> configMap = getTableAggregatorConfigs(accumuloHelper, log, tableNames);
+        Map<String,Map<String,String>> configMap = getTableConfigs(accumuloHelper, log, tableNames);
         for (Map.Entry entry : configMap.entrySet()) {
             conf.set(entry.getKey().toString(), entry.getValue().toString());
```
ivakegg: The value here is now a Map<String,String> instead of just a String.  Where is the code that appropriately can parse the configuration properties?  I cannot seem to find it in here.

ivakegg (changes requested):

warehouse/ingest-core/src/main/java/datawave/ingest/config/TableConfigCache.java
hlgp: Sure

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/MultiRFileOutputFormatter.java
hlgp: Sure

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/TableConfigurationUtil.java
hlgp: Ah, yes.  I wanted to talk about ripping this bandaid off.  Thank you for the reminder.
Previously, the aggregators and other table configs were read from Accumulo every time, then written to the Configuration object and then, in other classes, we would loop through the entire conf object later for properties matching the accumulo configs regexes of interest to pick them out.  So, I think to answer your question, the Conf was always parsed in other classes which needed those properties (and rather inefficiently at that).  With this cache mechanism, there is no need for that anymore, but I left the artifact of that here, but perhaps I should not have...

hlgp: It is also possible that there were some system-level properties set in the conf that may still be needed later, but, again, those are read out of the conf by the classes that need them.

contrib/datawave-quickstart/bin/services/datawave/install-ingest.sh
hlgp: I resisted putting that in here with all my might

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/MultiRFileOutputFormatter.java
hlgp: I see why I did it this way now.  I'd forgotten. IngestJob, partitioners, and PartitionerCache (maybe more?) accept the string array of tableIds as an argument to several methods (and have for a long time, it seems) which would cause me to touch more classes that I thought I should for this effort.  I can still update it if you want.

ivakegg (dismissed):

apmoriarty (dismissed):

warehouse/ingest-core/src/main/java/datawave/ingest/config/TableConfigCache.java
```diff
@@ -1,64 +1,116 @@
...
+                    out.println(table.getKey() + DELIMITER + tableProp.getKey() + DELIMITER + tableProp.getValue());
+                }
             }
         } catch (IOException e) {
             log.error("Unable to write cache file " + tmpCacheFile, e);
         }
     }
     
+    @Override
+    protected void readCache(BufferedReader in, String delimiter) throws IOException {
```
ivakegg: curious: why pass in the delimiter here if it is hardcoded in the writeCacheFile routine?

warehouse/ingest-core/src/main/java/datawave/ingest/config/TableConfigCache.java
```diff
@@ -1,64 +1,116 @@
...
+            }
+            if (parts.length == 3) {
+                propName = parts[1];
+                propVal = parts[2];
+                tempMap.put(propName, propVal);
+            } else {
+                throw new IOException("Invalid Table Config Cache at " + this.cacheFilePath + " . Please verify its contents.");
+            }
+        }
+        in.close();
```
ivakegg: If an open stream is being passed in, then closing the stream is the responsibility of the caller.  No need to call in.close() here.

ivakegg (changes requested):

warehouse/ingest-core/src/main/java/datawave/ingest/config/TableConfigCache.java
hlgp: Ah, yes.  I forgot I'd done a try with resources

ivakegg (approved):

apmoriarty (approved):
LGTM

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reduce/AggregatingReducer.java
```diff
@@ -120,61 +121,37 @@ public void setup(Configuration conf) throws IOException, InterruptedException {
...
                     Map<String,String> options = Maps.newHashMap();
                     
                     options.putAll(priorityOptions.get(priority));
                     
-                    String clazz = options.get("iterClazz");
+                    String clazz = options.get(TableConfigurationUtil.ITERATOR_CLASS_MARKER);
                     
                     if (null == clazz) {
-                        throw new RuntimeException("Unable to instantiate combiner class. Config item 'iterClazz' not present ");
+                        throw new RuntimeException("Unable to instantiate combiner class. Config item 'iterclass' not present " + priority + " "
```
ivakegg, line 150: Why did we change from iterClazz to iterclass ?

PR #1316, opened 2021-10-15
fixes #1315: Updated to allow specifying a scan limit when using the …

### LINKED ISSUE #1315: Need to be able to specify a scan limit on the edge query logics
We can have issues with the edge query logics timing out when we have way too many edges to scan over.  There needs to be a way to specify a scan limit on these logics.

ivakegg:
…edge query logic.


jwomeara (approved):
LGTM

jwomeara:
Approved, but wondering if this is something that is unit-testable...

apmoriarty (approved):
LGTM...no test?

PR #1317, opened 2021-10-18
Fixes 1312: Make SafeFileCommitter tolerant of residue from failed attempts

matthpeterson:


warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/SafeFileOutputCommitter.java
```diff
@@ -16,36 +18,32 @@
...
  * This extension of the FileOutputCommitter does a safety check to ensure no files were left around in the temporary directories before terminating. If there
  * were, then this will fail the job.
  */
 public class SafeFileOutputCommitter extends FileOutputCommitter {
+    private static final Logger LOG = LoggerFactory.getLogger(SafeFileOutputCommitter.class);
+    
+    public static final String ALLOW_EQUIVALENT_FILENAME = "mapreduce.safefileoutputcommitter.allow.equivalent.filename";
+    
+    private static final boolean DEFAULT_ALLOW_EQUIVALENT_FILENAME = false;
+    private final boolean shouldFindEquivalentFilename;
```
jschmidt10: So now we have two modes:

1) (Legacy) Every task's output files must've been imported
2) (New) Every unique file must've been imported

I'm wondering if there's a clearer name for #2. `uniqueFileMode` maybe? And then you could describe that mode in the javadoc? I'm not totally sure. What do you think?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/SafeFileOutputCommitter.java
```diff
@@ -70,19 +68,48 @@ private Path getPendingJobAttemptsPath() {
...
+            Path pendingJobAttemptsPath = getPendingJobAttemptsPath();
+            FileSystem fs = pendingJobAttemptsPath.getFileSystem(context.getConfiguration());
+            // now verify we do not have any files left in the temporary directory structure
+            List<Path> fileList = new ArrayList<>();
+            boolean containsPendingFiles = containsFiles(fs, pendingJobAttemptsPath, fileList);
+            
+            if (containsPendingFiles && !shouldFindEquivalentFilename) {
+                throw new FileExistsException("Found files still left in the temporary job attempts path: " + fileList);
+            }
+            if (containsPendingFiles) {
```
jschmidt10: This is implicitly `containsPendingFiles && shouldFindEquivalentFilename` - which is what we want, but I don't think we want to rely on that. For instance, if someone changed the condition of the previous `if`, it could break this code. I think we want this to be explicit:

```
if (containsPendingFiles) {
  if (shouldFindEquivalentFilename)
    // verify
  else
    // throw error
}
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/SafeFileOutputCommitter.java
```diff
@@ -70,19 +68,48 @@ private Path getPendingJobAttemptsPath() {
...
     }
+
+    /**
+     * Ensure that for each of the pending files in the provided list there is also a successful file with a matching filename.  If not, throw an exception.
+     * @param fs FileSystem to use
+     * @param pendingFileList List of pending files that need to be verified
+     * @throws IOException
+     * @throws FileExistsException a pending file exists and there is no successful file with the same name
+     */
+    private void verifyRemainingTemporaryFiles(FileSystem fs, List<Path> pendingFileList) throws IOException {
```
jschmidt10: How do you feel about `verifyByFilename` or `verifyUniqueFiles` or something to highlight that we're using filenames to do the verification?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/SafeFileOutputCommitter.java
```diff
@@ -70,19 +68,48 @@ private Path getPendingJobAttemptsPath() {
...
+        List<Path> allFilesInOutputPath = new ArrayList<>();
+        containsFiles(fs, super.getOutputPath(), allFilesInOutputPath);
+        
+        LOG.trace("Number of files in output path: {} and in pending: {}", allFilesInOutputPath.size(), pendingFileList.size());
+        // Exclude the pending files so that allFilesInOutputPath only contains the successful files
+        allFilesInOutputPath.removeAll(pendingFileList);
+        LOG.trace("Number of non-pending files: {}", allFilesInOutputPath.size());
+
+        // Retrieve just the filenames
+        Collection<String> successFileNamesOnly = allFilesInOutputPath.stream().map(Path::getName).collect(Collectors.toList());
```
jschmidt10: If we're looking for unique filenames here, I'm wondering if we should make these `Set`s.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/SafeFileOutputCommitter.java
```diff
@@ -70,19 +68,48 @@ private Path getPendingJobAttemptsPath() {
...
+        Collection<String> pendingFileNamesOnly = pendingFileList.stream().map(Path::getName).collect(Collectors.toList());
```
jschmidt10: Not a big deal but this stream sequence is the same as the one above. We could use a method `pendingFileNamesOnly = getNames(pendingFileList)`.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/SafeFileOutputCommitter.java
matthpeterson: Are you ok with this instead?
```
if (containsPendingFiles && shouldFindEquivalentFilename)
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/SafeFileOutputCommitter.java
matthpeterson: Ok.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/SafeFileOutputCommitter.java
matthpeterson: Good idea, thanks.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/SafeFileOutputCommitter.java
matthpeterson: I agree javadoc for the the constant is warranted.

I think there are merits to each name.  It seems that the most important name to discuss is the outward-facing property name, so maybe we should start there.  

It's currently set to this:
```
mapreduce.safefileoutputcommitter.allow.equivalent.filename
```

Do you think it should be changed to this?
```
mapreduce.safefileoutputcommitter.unique.file.mode
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/SafeFileOutputCommitter.java
matthpeterson: verifyRemainingTemporaryFilesByName?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/SafeFileOutputCommitter.java
matthpeterson: We agreed in a side chat to do this:
```
if (containsPendingFiles && shouldFindEquivalentFilename) {
    // verify
} else if (containsPendingFiles)
    // throw error
}
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/SafeFileOutputCommitter.java
matthpeterson: Perhaps it should be:
```
mapreduce.safefileoutputcommitter.lenient.mode
```

jschmidt10 (dismissed):

hlgp (dismissed):

jschmidt10 (approved):

ivakegg (approved):

Issue #1108, opened 2021-03-09
occurrence function not working as expected


ivakegg:
The occurence query function (#OCCURANCE) is not correct because of a couple of reasons:
1) Values are added into the JexlContext from the index as well as those from the event.  Only the values from the event should be considered in this case unless there are none.
2) The loop that tries to count the occurences is invalid in that it tries to create a map based on type, but then increments the count anyway for those that do not have a type.  Basically a somewhat random function based on the order of the values in the set.

lbschanno:
@ivakegg I'm interested in working on this. Where is the logic for this taking place? I didn't see it in the `OccurrenceFunction` class.

ivakegg:
Query functions are found in several classes.  For this function look in datawave.query.jexl.functions.  The occurance methods are at the top of the class.  In a lucene query the #OCCURANCE function will be turned into a filter:occurance() jexl function.  When a query is evaluated in the JexlEvaluation class, Jexl will pass in the field values from the jexl context as the first parameter of that function.  The issue is that we currently do not have a way to distinguish values in the Jexl context that were derived from the index vs the event fields themselves.  Also the getSizeOf(Iterable<?> iterable) method in the EvaluationPhaseFilterFunctions class is kinda hokey to start with.

The JexlContext is built when calling the visit method on the Document.  Attributes that are added to the Document need to be marked as being from the index vs the event as well as the ValueTuple objects which are put into the Jexl context.

PR #1320, opened 2021-10-19
Improve test coverage of EvaluationPhaseFilterFunctions

lbschanno:
Add a wide range of unit tests to improve test coverage for `EvaluationPhaseFilterFunctions`. In addition, improve documentation, and fix a few small bugs.

I began this work with the intent of familiarizing myself with `EvaluationPhaseFilterFunctions` in order to address issues with the occurrence function as described in #1108. However, the intended behavior of the occurrence function is still not clear, and potentially buggier than thought. See the test suite OccurrenceIterableValueTests.testsToShowOddBehavior() to see examples. It appears that the total count will reset whenever a new ValueTuple with a Type as its second property is seen in an Iterable argument.


apmoriarty (dismissed):
LGTM, someone with more knowledge of this area of the code should take a look

warehouse/query-core/src/test/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctionsTest.java
```diff
@@ -1,765 +1,1965 @@
...
+            
+            // If the list contains a mixture of objects and ValueTuples with a Type as its second, the count of the last type will be returned, and will not
+            // include the count for objects seen before it.
+            values.add(new Object());
+            values.add(new Object());
+            values.add(new Object());
+            values.add(toValueTuple("FOO.1,BAR,bar", NoOpType::new));
+            values.add(toValueTuple("FOO.1,BAR,bar", NoOpType::new));
+            values.add(toValueTuple("FOO.1,BAR,bar", LcNoDiacriticsType::new));
+            assertTrue(resultForOperator(1)); // The total number of LcNoDiacriticsType instances seen (1).
```
ivakegg, line 430: I don't think the type should matter.  This should probably be 3.

warehouse/query-core/src/test/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctionsTest.java
```diff
@@ -1,765 +1,1965 @@
...
+            // Similar to the case above, but this time there are objects in the list after the last type seen. The count for these objects will be added to the
+            // count of the last type seen.
+            values.clear();
+            values.add(toValueTuple("FOO.1,BAR,bar", LcNoDiacriticsType::new));
+            values.add(toValueTuple("FOO.1,BAR,bar", NoOpType::new));
+            values.add(toValueTuple("FOO.1,BAR,bar", NoOpType::new));
+            values.add(new Object());
+            values.add(new Object());
+            values.add(new Object());
+            assertTrue(resultForOperator(5)); // The count of the last type, NoOpType, seen (2) plus the number of non-typed objects seen after that (3).
```
ivakegg, line 441: This one does not make any sense to me.  This should be 3 as well I would expected (typed values only).

warehouse/query-core/src/test/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctionsTest.java
```diff
@@ -1,765 +1,1965 @@
...
+            values.clear();
+            values.add(new Object());
+            values.add(new Object());
+            values.add(toNonTypedValueTuple("Foo.1,BAR,bar"));
+            values.add(toNonTypedValueTuple("Foo.2,BAT,bat"));
+            values.add(toNonTypedValueTuple("Foo.3,FOO,foo"));
+            values.add(toNonTypedValueTuple("Foo.4,GRUMP,grump"));
+            values.add(new Object());
+            values.add(new Object());
+            assertTrue(resultForOperator(8)); // Returns the total number of objects seen (8).
```
ivakegg, line 453: I am thinking this should be 4 as well.

warehouse/query-core/src/test/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctionsTest.java
```diff
@@ -1,765 +1,1965 @@
...
+            givenFieldValue(listOfThreeValues);
+            assertFalse(resultForDefaultOperator(2));
+            assertTrue(resultForDefaultOperator(3));
+            assertFalse(resultForDefaultOperator(4));
+        }
+        
+        // These tests show examples of odd behavior that is not clearly documented, and is intended to be reviewed in order to determine what the correct
+        // behavior should be.
+        @Test
+        public void testsToShowOddBehavior() {
```
ivakegg, line 417: So I think I may have gotten clarification on what this function should do from a user standpoint.  The function needs to count all of the instances of a field in an event/document.  What this means is that fields that are in the events AND are found in the index because the query was searching on it should only count the fields that came from the event (the indexed values are added separately to the context).  However if the field is index-only, then and only then should it count the fields added to the context from the index.

ivakegg (changes requested):

ivakegg (approved):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
```diff
@@ -1612,97 +1749,228 @@ public static String getMatchToRightOfPeriod(String input, int pos) {
...
+     * @param set
+     *            the set
+     * @return the normalized values
+     */
+    private static Set<Object> getNormalizedValues(FunctionalSet<ValueTuple> set) {
+        return set.stream().map((value) -> value.getNormalizedValue()).collect(Collectors.toSet());
+    }
+    
+    // Do not allow this class to be instantiated.
+    private EvaluationPhaseFilterFunctions() {
```
ivakegg: Please remove this private constructor.  There is a test case on the other side that extends this class and this breaks it.

ivakegg (changes requested):
see comments

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
lbschanno: Fixed.

apmoriarty (approved):

ivakegg (approved):

PR #1324, opened 2021-10-20
removed word-wrap and replaced with word-break to wrap long strings

tynyttie:
Removed word-wrap and replaced with word-break to wrap long strings in DataWave/Query/Metrics/id/.


ivakegg (approved):

hlgp (approved):

hlgp:
This class has been moved to the query metrics microservice

PR #1327, opened 2021-10-22
fixes #1326: Changed the traversal algorithm to avoid recursion when …

### LINKED ISSUE #1326: Content functions can produce stack overflows
The content functions (e.g. phrase) is now throwing stack overflow exceptions from time to time.  The reason we are only now seeing this is that the jexl evaluation was changed to be strict to ensure that function exceptions actually failed the query.  The content function phrase is currently recursive in how it does its computation.  This needs to be changed to avoid such errors when very large term frequency lists are pulled in for evaluation.

ivakegg:
Changed the traversal algorithm to avoid recursion when re-starting a traversal farther down the offset list.


apmoriarty (approved):

jwomeara (approved):
Looks good to me.  In the comparator, i think it sounds reasonable to remove that last term comparison.

ivakegg:
Verified this builds with the other repo.  Integration testing required.

PR #1328, opened 2021-10-25
Adding version field to query metrics

### LINKED ISSUE #1322: Adding version to query metrics

avgAGB:
Resolves #1322


web-services/query/src/test/java/datawave/webservice/query/runner/QueryExecutorBeanTest.java
```diff
@@ -477,7 +480,8 @@ public boolean equals(Object o) {
         
         Object cachedRunningQuery = cache.get(q.getId().toString());
         Assert.assertNull(cachedRunningQuery);
-        
+        // System.out.println("*** testMetric ***");
+        // System.out.println(testMetric.toString());
```
billoley: Should these lines be deleted?

web-services/query/src/test/java/datawave/webservice/query/runner/QueryExecutorBeanTest.java
```diff
@@ -153,6 +154,7 @@
     public void setup() throws Exception {
         System.setProperty(NpeUtils.NPE_OU_PROPERTY, "iamnotaperson");
         System.setProperty("dw.metadatahelper.all.auths", "A,B,C,D");
+        System.setProperty(DnUtils.SUBJECT_DN_PATTERN_PROPERTY, "(?:^|,)\\\\s*OU\\\\s*=\\\\s*Subject DN OU\\\\s*(?:,|$)");
```
billoley: Is this related to the stated issue?

web-services/client/src/main/java/datawave/webservice/query/metric/QueryMetric.java
```diff
@@ -631,11 +642,25 @@ public int getFieldNumber(String name) {
             fieldMap.put("docRanges", 32);
             fieldMap.put("fiRanges", 33);
             fieldMap.put("plan", 34);
-            fieldMap.put("loginTime", 35);
-            fieldMap.put("predictions", 36);
+            fieldMap.put("version", 35);
+            fieldMap.put("loginTime", 36);
+            fieldMap.put("predictions", 37);
```
billoley: I believe that normal practice when adding a field to the schema is to use the next field number.  Otherwise, previousy serialized messages will have field numbers that are inconsistent with the new numbers.  (backward-compatibility)

billoley (commented):
There is still work to be done in adding this field to the code that persists a query metric to Accumulo (ShardTableQueryMetricHandler and related classes).

web-services/client/src/main/java/datawave/webservice/query/metric/QueryMetric.java
avgAGB: Understood, I'll adjust this in the next commit.

web-services/query/src/test/java/datawave/webservice/query/runner/QueryExecutorBeanTest.java
avgAGB: I found that the unit test was unable to run properly without the pattern being defined in the setup here, but it may have been a IDE specific problem. I'll confirm this and remove it if this is the case.

web-services/query/src/test/java/datawave/webservice/query/runner/QueryExecutorBeanTest.java
avgAGB: Yes, will trim them out upon next commit, thanks

avgAGB:
> There is still work to be done in adding this field to the code that persists a query metric to Accumulo (ShardTableQueryMetricHandler and related classes).

I've added the metric to ShardTableQueryMetricHandler and ContentQueryMetricsIngestHelper in the latest commit.

web-services/client/src/main/java/datawave/webservice/query/metric/QueryMetric.java
```diff
@@ -370,6 +366,14 @@ public void writeTo(Output output, QueryMetric message) throws IOException {
                     }
                 }
             }
+            
+            // I'm open to move this logic elsewhere, with the intent being that the version will always be presented
+            // in the QueryMetric object for display.
```
matthpeterson: I agree with the idea in this comment.
I think in writeTo, the logic should be simply:
```
            if (message.version != null) {
                output.writeString(37, message.getVersion(), false);
            }  
```
This code should probably be moved to exist wherever message is normally initialized. 
```
                String propertyVersion = getVersionFromProperties();
                message.setVersion(propertyVersion);
```

matthpeterson: Maybe the constructor would be a reasonable place to put it, near where host is initialized.

web-services/client/src/main/java/datawave/webservice/query/metric/QueryMetric.java
```diff
@@ -633,9 +642,23 @@ public int getFieldNumber(String name) {
...
             fieldMap.put("predictions", 36);
+            fieldMap.put("version", 37);
         }
     };
     
+    private static String getVersionFromProperties() {
+        String returnStr = "";
+        try {
+            final Properties props = new Properties();
+            String inputPath = QueryMetric.class.getClassLoader().getResource("version.properties").getPath();
```
matthpeterson: If version.properties is not found, this will result in a Null Pointer Exception.  I suggest handling that possibility

web-services/client/src/main/java/datawave/webservice/query/metric/QueryMetric.java
```diff
@@ -633,9 +642,23 @@ public int getFieldNumber(String name) {
...
+            props.load(new FileInputStream(inputPath));
+            returnStr = props.getProperty("currentVersion");
+        } catch (IOException e) {
+            e.printStackTrace();
```
matthpeterson: This seems like a good opportunity to use a logger instead of calling e.printStackTrace.

web-services/client/src/test/java/datawave/webservice/query/metric/QueryMetricTest.java
```diff
@@ -71,6 +67,77 @@ public void testSetError() {
...
+        assertEquals("proxyServer1", metric.getProxyServers().iterator().next());
+        assertEquals("query", metric.getQuery());
+        assertEquals("auths", metric.getQueryAuthorizations());
+        assertEquals("queryId", metric.getQueryId());
+        assertEquals("queryLogic", metric.getQueryLogic());
+        assertEquals("queryType", metric.getQueryType());
+        assertEquals(0, metric.getSetupTime());
+        assertEquals("user", metric.getUser());
+        assertEquals("userDN", metric.getUserDN());
+        assertEquals("3.5.1-TEST", metric.getVersion());
```
matthpeterson: This test verifies that the setters and getters work.  I'd like to see tests that call the other code that was changed including getVersionFromProperties, writeTo, and mergeFrom.

web-services/client/src/main/resources/version.properties
```diff
@@ -0,0 +1 @@
+currentVersion=${project.version}
```
matthpeterson, line 1: I'd like to see a test that verifies a currentVersion is successfully loaded via version.properties and is both non-empty as well as not equal to "${project.version}".  This will prove that the property injection and resource loading is working correctly.

matthpeterson (changes requested):
Looks really good.  I suggest adding more tests and some minor cleanup, as noted in comments.

web-services/client/src/main/resources/version.properties
avgAGB: I do have a matching version.properties in the test package that aims to simulate this with an assignment (currentVersion=3.5.1-TEST) but I agree with you on the point that some of the other functions involving the version should have accompanied unit tests in some manner. Will work on putting that together.

web-services/client/src/main/java/datawave/webservice/query/metric/QueryMetric.java
avgAGB: Cool, I'll make this update in the following commits.

web-services/client/src/main/java/datawave/webservice/query/metric/QueryMetric.java
avgAGB: Adding a logger into the next commit.

web-services/client/src/main/java/datawave/webservice/query/metric/QueryMetric.java
avgAGB: Adding an additional catch in a follow-up commit

web-services/client/src/test/java/datawave/webservice/query/metric/QueryMetricTest.java
avgAGB: I'll include a serialization unit test next commit which will cover code referencing all the above mentioned methods.

web-services/client/src/main/java/datawave/webservice/query/metric/QueryMetric.java
```diff
@@ -633,9 +652,27 @@ public int getFieldNumber(String name) {
...
+        try {
+            final Properties props = new Properties();
+            String inputPath = QueryMetric.class.getClassLoader().getResource("version.properties").getPath();
+            props.load(new FileInputStream(inputPath));
+            returnStr = props.getProperty("currentVersion");
+        } catch (IOException e) {
+            log.error("Unable to find a version from resources properties file: ", e);
+            returnStr = "";
+        } catch (NullPointerException e) {
+            log.error("Null pointer encountered attempting to draw version.properties resource file: ", e);
```
matthpeterson: See https://stackoverflow.com/a/15146399.  It'd be best to do a check for null rather than catch this.

avgAGB: Tweaked this in the latest commit, thanks

billoley (approved):
Looks good

matthpeterson (approved):

ivakegg (approved):

PR #1329, opened 2021-10-25
Add #NOEXPANSION function

apmoriarty:
Add #NOEXPANSION function, allows user to specify fields that are excluded from query model expansion.


ivakegg:
Brilliant.  I would also like to be able to specify the noExpansionFields as a query parameter as well like we do with unique.fields or grouped.fields ....

apmoriarty:
@ivakegg opened an issue for specifying #NOEXPANSION fields as a query parameter (see #1331).

FineAndDandy (approved):

alerman (approved):

Issue #1332, opened 2021-10-26
Add Lucene Parsing for #COMPARE Function


apmoriarty:
Add support for lucene to jexl parsing for the #COMPARE function.

apmoriarty:
Resolved by https://github.com/NationalSecurityAgency/datawave/pull/1333

PR #1333, opened 2021-10-26
Add support for parsing #COMPARE function from lucene query

apmoriarty:
Addresses #1332


d-hwang:
Would you be able to update web-services/deploy/docs/docs/query_help.html with some examples for both syntax? I missed updating it when I worked on it. Thanks!

d-hwang (dismissed):

Gara312:
> Сможете ли вы обновить web-services / deploy / docs / docs / query_help.html некоторыми примерами для обоих синтаксисов? Я пропустил обновление, когда работал над этим. Спасибо!

FineAndDandy (dismissed):

keith-ratcliffe (commented):
LGTM

keith-ratcliffe (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
```diff
@@ -1643,7 +1643,9 @@ public static boolean compare(Object field, String operator, String compareMode,
             }
         }
         
-        return compareFields(fields, operator, CompareFunctionValidator.Mode.valueOf(compareMode) == CompareFunctionValidator.Mode.ANY, fields2);
+        boolean matchesAny = compareMode.equalsIgnoreCase("any");
```
ivakegg: why can't we use the enum here for comparison?

warehouse/query-core/src/main/java/datawave/query/language/functions/jexl/Compare.java
```diff
@@ -0,0 +1,60 @@
...
+ * Function args are a field name, an operator, a mode, and another field name.
+ * <p>
+ * Lucene Example: #COMPARE(FIELD_A, &lt;, ANY, FIELD_B)
+ * <p>
+ * Jexl Example: filter:compare(FIELD_A, &lt;, ANY, FIELD_B)
+ */
+public class Compare extends AbstractEvaluationPhaseFunction {
+    
+    private final Set<String> validOpArgs = Sets.newHashSet("<", "<=", ">", ">=", "==", "=", "!=");
+    private final Set<String> validModeArgs = Sets.newHashSet("ANY", "ALL");
```
ivakegg: Why can't we use the enum names here?

warehouse/query-core/src/main/java/datawave/query/language/functions/jexl/Compare.java
```diff
@@ -0,0 +1,60 @@
...
+    
+    public Compare() {
+        super("compare");
```
ivakegg, line 20: How about a constant for "compare".  I see it used below as well.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
apmoriarty: fixed in next commit

warehouse/query-core/src/main/java/datawave/query/language/functions/jexl/Compare.java
apmoriarty: As-is this follows the pattern of the include/exclude functions. I can extract that to a constant if you prefer

warehouse/query-core/src/main/java/datawave/query/language/functions/jexl/Compare.java
apmoriarty: good call, that will be in the next commit

warehouse/query-core/src/main/java/datawave/query/language/functions/jexl/Compare.java
ivakegg: its fine as is

ivakegg (approved):

d-hwang (approved):

PR #1341, opened 2021-11-05
Merge filter matchPatterns across classes

matthpeterson:
This PR aims to verify through additional tests that two different filter classes can share matchPatterns through the existing parent/child merge paradigm, as long as the parent's matchPattern content is in an acceptable format for each of the two filter classes.  In verifying this, some adjustments were made to the TokenSpecParser.  It was also noticed that the use of floats in the TokenTtlTrie caused imprecision for the ttls with large numbers containing enough significant digits.


warehouse/core/src/test/java/datawave/ingest/util/cache/watch/DifferentClassesMergeTest.java
```diff
@@ -0,0 +1,216 @@
...
+        assertChildAccepts(colQual, TIMESTAMP_IN_FUTURE);
+        assertChildAccepts(colQual, 370 * DAYS_AGO);
+        assertChildAccepts(colQual, 547 * DAYS_AGO);
+        assertChildRejects(colQual, 548 * DAYS_AGO);
+        assertChildRejects(colQual, 570 * DAYS_AGO);
+        assertChildRejects(colQual, VERY_OLD_TIMESTAMP);
+    }
+    
+    @Test
+    public void inspectDriedBeans() {
```
hlgp, line 72: I love these tests

hlgp (approved):

avgAGB (approved):

PR #1342, opened 2021-11-09
JexlNodes negate builds proper reference expression

apmoriarty:
General clean up of how negations are handled
- JexlNodes negate call now builds proper reference-referenceExpression chain, to include marker nodes
- PushdownNegations dropNot call now drops all relevant nodes to the negation


ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PullupUnexecutableNodesVisitor.java
```diff
@@ -245,6 +245,10 @@ public Object visit(ASTReference node, Object data) {
         QueryPropertyMarker.Instance instance = QueryPropertyMarker.findInstance(node);
         if (instance.isType(ASTDelayedPredicate.class)) {
             JexlNode source = ASTDelayedPredicate.unwrapFully(node, ASTDelayedPredicate.class);
+            // when pulling up a delayed marker that is negated, a reference expression must be persisted
+            if (!(source instanceof ASTReferenceExpression) && JexlNodes.findNegatedParent(node)) {
```
lbschanno: It is possible for the source to be an instance of ASTReference, do you need to account for this here?

lbschanno (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PullupUnexecutableNodesVisitor.java
apmoriarty: Addressed this by always dereferencing the source and then creating an `ASTReferenceExpression` when the source is an `ASTOrNode`, an `ASTAndNode`, or if the source is negated.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PullupUnexecutableNodesVisitor.java
```diff
@@ -245,7 +250,12 @@ public Object visit(ASTReference node, Object data) {
         QueryPropertyMarker.Instance instance = QueryPropertyMarker.findInstance(node);
         if (instance.isType(ASTDelayedPredicate.class)) {
             JexlNode source = ASTDelayedPredicate.unwrapFully(node, ASTDelayedPredicate.class);
-            JexlNodes.swap(node.jjtGetParent(), node, source);
+            source = JexlASTHelper.dereference(source);
+            // when pulling up a delayed marker that is negated, a reference expression must be persisted
```
ivakegg, line 253: I am wondering if this dereference and wrap if an and/or should be a JexlASTHelper method: JexlASTHelper.wrapIfNeeded(node)

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownNegationVisitor.java
```diff
@@ -231,19 +234,47 @@ public static JexlNode applyDeMorgans(JexlNode root, boolean negateRoot) {
...
+     */
     private JexlNode dropNot(ASTNotNode toDrop, NegationState state) {
-        JexlNode newNode = toDrop.jjtGetChild(0);
+        JexlNode child = toDrop.jjtGetChild(0);
         
         // no need to swap nodes if toDrop is the root
         if (toDrop.jjtGetParent() != null) {
+            
+            // check for adjacent not nodes
+            if (child instanceof ASTNotNode || QueryPropertyMarker.findInstance(child).isAnyType()) {
```
ivakegg, line 253: What if the child of the child is a Not Node?  What if the child of the child of the child is a NotNode?  Doesn't the visitor pattern do the appropriate thing as it moves down the tree?

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/JexlNodes.java
```diff
@@ -170,8 +171,19 @@ public static <T> void setLiteral(ASTStringLiteral literal, String value) {
...
+    /**
+     * Negate the provided JexlNode
+     *
+     * @param node
+     *            an arbitrary JexlNode
+     * @return a negated version of the provided JexlNode
+     */
     public static ASTNotNode negate(JexlNode node) {
-        return children(new ASTNotNode(ParserTreeConstants.JJTNOTNODE), wrap(node));
+        if (QueryPropertyMarker.findInstance(node).isAnyType()) {
```
ivakegg, line 182: We could use that wrapIfNeeded call I suggested above.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownNegationVisitor.java
apmoriarty: This is handled by the calling method. After dropping the negation it immediately visits the returned node. I believe several of the tests for this visitor cover the exact case you're thinking of.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PullupUnexecutableNodesVisitor.java
apmoriarty: I can extract that logic to a local method. I'm not thinking of any broader use cases for such a method outside of this visitor. If you can think of any I'm happy to move that logic to the JexlASTHelper.

FineAndDandy (approved):

ivakegg (approved):

lbschanno (approved):

PR #1346, opened 2021-11-15
Apply UID maximum to removal list

matthpeterson:
The purpose of the UID maximum list is to improve storage and latency, at the expense of accuracy.  The removal list is currently exempt from the UID maximum limit.  This PR changes that.  If there are more removal UIDs than the maximum, it's likely that the same key has already exceeded its UID maximum (in another file) and so the removal UIDs are likely to get truncated at a later time anyway.

### DIFF
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
@@ -174,8 +174,8 @@ public class GlobalIndexUidAggregator extends PropogatingCombiner {
                     // set because that will take care of de-duping any duplicate UIDs.
                     long prevCount = uids.size() - uidsToRemove.size();
                     
-                    boolean wasSuccessful = processRemovalUids(v) && processAddedUids(v);
-                    if (!wasSuccessful) {
+                    boolean isUnderMaxThreshold = processRemovalUids(v) && processAddedUids(v);
+                    if (!isUnderMaxThreshold) {
                         enterCountOnlyMode(prevCount);
                     }
                 }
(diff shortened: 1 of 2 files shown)


warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
```diff
@@ -174,59 +174,9 @@ public Value reduce(Key key, Iterator<Value> iter) {
...
-                                // Set the count to the number of collected UIDs before we went
-                                // over the max, and then we'll add this PB's count below.
-                                seenIgnore = true;
-                                count = prevCount;
-                                uids.clear();
-                                uidsToRemove.clear();
-                                break;
-                            }
-                        }
+                    boolean wasSuccessful = processRemovalUids(v) && processAddedUids(v);
```
ivakegg: nice, I like the simplification here.

ivakegg (dismissed):

warehouse/ingest-core/src/main/java/datawave/ingest/table/aggregator/GlobalIndexUidAggregator.java
jschmidt10: Instead of `wasSuccessful`, maybe we can call this `underMaxThreshold` or something like that? `wasSuccessful` sorta makes you think the operation can fail.

matthpeterson: Good idea, thanks.

ivakegg (approved):

jschmidt10 (approved):

brianloss (approved):

PR #1348, opened 2021-11-16
DatawaveInterpreter now handles a flattened query.

apmoriarty:
This is definitely on the experimental side, feedback on the implementation is welcome. 

Ran a number of queries through this code and found no result differences, so I'm reasonably confident that the code is functionally correct.


warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
```diff
@@ -439,6 +439,7 @@ private void addHitsForFunction(String field, ASTFunctionNode node, HitListArith
     }
     
     public Object visit(ASTAndNode node, Object data) {
+
```
ivakegg, line 442: I love the simplification of this method.  awesome.  I wonder if the same thing can be done to the interpretOr method.  Just a thought, but not required of course.

warehouse/query-core/src/main/java/org/apache/commons/jexl2/DatawaveJexlScript.java
```diff
@@ -0,0 +1,171 @@
...
+import org.apache.commons.jexl2.parser.ASTJexlScript;
+
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.Callable;
+
+/**
+ * Lifted from {@link ExpressionImpl}. Modified to flatten the resulting {@link ASTJexlScript}.
+ */
+public class DatawaveJexlScript implements Expression, Script {
```
ivakegg, line 13: I see a lot of the methods are cut and paste from the Jexl implementations.  Can we extend instead or implement using aggregation instead?  Just wondering.

ivakegg (changes requested):

warehouse/query-core/src/main/java/org/apache/commons/jexl2/DatawaveJexlScript.java
apmoriarty: I tried extending/implementing and didn't find that to be particularly successful. I could take another look, this was written pretty quickly.

hlgp (approved):

FineAndDandy (approved):

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
apmoriarty: We handle some special cases in the `interpretOr` method related to function args and I would rather not expand the scope of this pull request, if possible. It's probably worth looking at in some follow on work.

ivakegg (approved):

Issue #1359, opened 2021-11-24
Query metrics /summary /summaryCounts endpoints broken


billoley:
These two endpoints use the following query which results in an exception. -> “Found incorrectly marked bounded ranges”

`USER > ‘A’ && USER < ‘ZZZZZZZ’`

The fix is to use:

`((_Bounded_ = true) && (USER > ‘A’ && USER < ‘ZZZZZZZ’))`

Additionaly, there are a bunch of inconsistencies with the three summary methods 

1. default begin time is either midnight of the current day (1) OR now - 90 days (2)
2. the same response object can support all three calls; use the Accept HTTPHeader 
3. /summary is mapped to getTotalQueriesSummaryCounts but /summaryCounts is mapped to getTotalQueriesSummary.  This is more than a little confusing and the confusion is carried through to the calls in STQMH
4. if the user asks for an end Date of one month ago, the 1 hr / 6 hr / 12 hr / etc calculations are still made using an end date of now

- /summary
  - mapped to QMB method getTotalQueriesSummaryCounts
  - returns a QueryMetricsSummaryResponse
  - any supported response type, but no Html
  - default begin time is now - 90 days
  - default end time is now
- /summaryCounts
  - mapped to QMB method getTotalQueriesSummary
  - returns a QueryMetricsSummaryHtmlResponse, which extends QueryMetricsSummaryResponse
  - any supprted response type including a custom HTML table
  - default begin time is midnight (the beginning) of the current day
  - default end time is now
  - histogram uses current time instead of the requested end time
- /summaryCounts/user 
  - mapped to QMB method getUserQueriesSummary
  - returns a QueryMetricsSummaryHtmlResponse, which extends QueryMetricsSummaryResponse
  - any supprted response type including a custom HTML table
  - default begin time is now - 90 days
  - default end time is now
  - histogram uses current time instead of the requested end time

billoley:
Changes:

use end time - 90 days as the default begin time
base time distributions on end date, not 'now'
check that begin < end, otherwise send an exception and display in the html response
change API from /summary and /summaryCounts to /summary/all (difference was response type and default date range)
change API from /summaryCounts/user to /summary/user 
use the same response object to support both calls -- just rely on the 'Accept' http header 
refactor duplicative code

PR #1363, opened 2021-11-30
re #1359 Query metrics /summary /summaryCounts endpoints broken

billoley:
use end time - 90 days as the default begin time
base time distributions on end date, not 'now'
check that begin < end, otherwise send an exception and display in the html response
change API from /summary and /summaryCounts to /summary/all (difference was response type and default date range)
change API from /summaryCounts/user to /summary/user 
use the same response object to support both calls -- just rely on the 'Accept' http header 
refactor duplicative code


warehouse/query-core/src/main/java/datawave/query/metrics/BaseQueryMetricHandler.java
```diff
@@ -79,61 +53,32 @@ public QueryMetricsSummaryHtmlResponse processQueryMetricsHtmlSummary(List<T> qu
     public void binSummary(T metric, QueryMetricsSummaryResponse summary, Date hour1, Date hour6, Date hour12, Date day1, Date day7, Date day30, Date day60,
                     Date day90) {
         Date d = metric.getCreateDate();
-        // Find out which bucket this query belongs to
+        // Find out which buckets this query belongs to based on query create date.
+        // If a query's create date is within one hour of the endDate then it is also within the other ranges.
         if (d.after(hour1)) {
             populateSummary(metric, summary.getHour1());
+        }
+        if (d.after(hour6)) {
```
ivakegg, line 61: much cleaner... thankyou

web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsBean.java
```diff
@@ -161,29 +163,17 @@ public QueryGeometryResponse map(@PathParam("id") @Required("id") String id) {
      * @HTTP 500 internal server error
      */
     @GET
-    @Path("/summaryCounts")
```
ivakegg, line 164: is there anyway to at least preserve these endpoints, even if they become synonyms for you new endpoints?

web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsBean.java
```diff
@@ -198,66 +188,40 @@ public QueryMetricsSummaryHtmlResponse getTotalQueriesSummary(@QueryParam("begin
      * @HTTP 500 internal server error
      */
     @GET
-    @Path("/summary")
```
ivakegg, line 201: See comment above

ivakegg (changes requested):

web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsBean.java
billoley: Added back the previous API endpoints and marked them as deprecated in the javadoc

web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsBean.java
billoley: Added back the previous API endpoints and marked them as deprecated in the javadoc

web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsBean.java
ivakegg: cool

PR #1366, opened 2021-12-06
Add method to JexlASTHelper to validate children of junction nodes

### LINKED ISSUE #1233: Add a method to JexlASTHelper which checks for AND/OR nodes with a single child
Once the method is added, we should add unit tests for each visitor to have some basic level of assurance that we are not intentionally doing this.  

Additionally, we should come up with a paradigm where this check can be enabled on a visitor via configuration so that we can test this in production if we suspect the issue is occurring.  We should do this for validateLineage as well.

lbschanno:
Add a method that will validate that a specified query tree does not
have any junction (AND/OR) nodes with less than 2 children.

Fixes #1233


warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -1611,6 +1612,23 @@ public String getFormattedViolations() {
...
+    /**
+     * Checks to see if the tree contains any AND/OR nodes with less than 2 children.
+     * 
+     * @param node
+     *            the tree to validate
+     * @param failHard
+     *            if true, throw a {@link RuntimeException} if a violation was found
+     * @return true if valid, or false otherwise
+     */
+    public static boolean validateJunctionChildren(JexlNode node, boolean failHard) {
```
apmoriarty, line 1635: Consider adding a default `valiateJunctionChildren(node)` with a reasonable default value for `failHard`?

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
lbschanno: Easy enough to do, and added.

apmoriarty (approved):

ivakegg (approved):

PR #1368, opened 2021-12-06
Re: #887 - Avoid NPEs for EvaluationPhaseFilterFunctions.timeFunction()

### LINKED ISSUE #887: Exceptions thrown when evaluating functions are silently swallowed

codebyneesh:
If null field values are passed into timeFunction(), that a NPE is not thrown, and an empty matches set is returned. getMaxTime(), getMinTime(), getMaxValue(), and getMinValue() are private to ensure null values cannot be passed to them.

Resolves #887


ivakegg (approved):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
```diff
@@ -1592,7 +1596,7 @@ static long getNextTime(long time, int granularity) {
      * @throws ParseException
      *             if the value failed to be parsed using the supplied format
      */
-    protected static long getTime(Object value, DateFormat format) throws ParseException {
+    static long getTime(Object value, DateFormat format) throws ParseException {
```
apmoriarty: It's a nitpick but I would prefer these be `public static...`

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
ivakegg: Create pr/1368 branch in datawave repo to allow integration testing to start.

ivakegg: I finally got to getting this built with our internal project.  This failed because this function is no longer available by an extending class.  This needs to be protected or public, but not package private.  This comment is the same for the methods below which you changed.  Thanks.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
```diff
@@ -1681,7 +1685,7 @@ static boolean betweenInclusive(long value, long left, long right) {
      * @return a {@link ValueTuple}
      * @see ValueTuple#toValueTuple(Object) documentation on conversion details
      */
-    protected static ValueTuple getHitTerm(Object valueTuple) {
+    static ValueTuple getHitTerm(Object valueTuple) {
```
ivakegg: You forgot this one

ivakegg (changes requested):

ivakegg (dismissed):

apmoriarty (dismissed):

ivakegg:
branch pr/1368 updated with latest changes

ivakegg (approved):

PR #1369, opened 2021-12-06
Fix #1063: Skip TimingMetadata entries when applying Unique transform…

### LINKED ISSUE #1063: UNIQUE functions appears to be dropping results with missing fields

friedlou:
…. (#1162)

* Fix #1063: Skip TimingMetadata entries when applying Unique transform.

* Removed left over commented out tests for debugging.

* Used isFinalDocumentKey to test if document is TIMING_METADATA, as suggested in PR comment.

* Removed unused method.

* Formatting

Co-authored-by: Eluzor Friedman efriedm <efriedm@ip-10-113-15-250.evoforge.org>
Co-authored-by: Moriarty <22225248+apmoriarty@users.noreply.github.com>
Co-authored-by: Ivan Bella <ivan@bella.name>


warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -96,6 +97,12 @@ public UniqueTransform(BaseQueryLogic<Entry<Key,Value>> logic, UniqueFields uniq
     @Override
     public Entry<Key,Document> apply(@Nullable Entry<Key,Document> keyDocumentEntry) {
         if (keyDocumentEntry != null) {
+            log.info("UniqueTransform.apply " + keyDocumentEntry.getKey() + " " + keyDocumentEntry.getValue().get("UUID.0"));
```
apmoriarty: Even after filtering down the documents logged to those that match the query, this will still generate a tremendous amount of logging. Can you wrap this in a `isTraceEnabled` statement, similar to this?

```java
if(log.isTraceEnabled()) {
   log.trace(....
}
```

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -118,10 +125,16 @@ private boolean isDuplicate(Document document) throws IOException {
         byte[] bytes = getBytes(document);
         synchronized (bloom) {
             if (bloom.mightContain(bytes)) {
+                if (bytes.length == 0) {
+                    log.info("Dup blank field. Fields: " + document.getDictionary().keySet() + " " + document.get("UUID.0"));
```
apmoriarty: Same comment, re:downgrading log level to trace

apmoriarty (changes requested):
Downgrade log levels to trace and I think this is good to go

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
friedlou: Just used for a debugging attempt. Removed in latest commit,

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
friedlou: Removed , as above.

alerman (approved):

ivakegg (approved):

apmoriarty (approved):

PR #1371, opened 2021-12-14
Fix occurrence query function

### LINKED ISSUE #1108: occurrence function not working as expected

lbschanno:
The occurrence function currently does not count occurrences correctly.
Add the ability to track when a field value originated from the index
vs. the event via its Attribute. Modify the occurrence function to only
return the total number of values from the event, unless there are none.
In this case, return the total number of values from the index.

Fixes #1108


warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
```diff
@@ -31,6 +33,7 @@
      */
     protected Key metadata = null;
     protected boolean toKeep = true; // a flag denoting whether this attribute is to be kept in the returned results (transient or not)
+    protected boolean fromIndex; // Assume attributes are from the index unless specified otherwise.
```
ivakegg: If this comment is correct, then should it be initialized to true?

warehouse/query-core/src/main/java/datawave/query/iterator/aggregation/DocumentData.java
```diff
@@ -52,6 +55,14 @@ public void setData(List<Entry<Key,Value>> data) {
         this.data = data;
     }
     
+    public boolean isFromIndex() {
+        return fromIndex;
+    }
+    
+    public void setFromIndex(boolean fromIndex) {
```
ivakegg: Should this actually be mutable?

warehouse/query-core/src/main/java/datawave/query/predicate/ValueToAttributes.java
```diff
@@ -52,14 +52,19 @@
     
     private EventDataQueryFilter attrFilter;
     
-    public ValueToAttributes(CompositeMetadata compositeMetadata, TypeMetadata typeMetadata, EventDataQueryFilter attrFilter, MarkingFunctions markingFunctions) {
+    // Whether the value is from the index
+    private boolean fromIndex;
```
ivakegg: final?

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
lbschanno: It should be. Fixed.

warehouse/query-core/src/main/java/datawave/query/iterator/aggregation/DocumentData.java
lbschanno: It should not be. Good catch. Fixed.

warehouse/query-core/src/main/java/datawave/query/predicate/ValueToAttributes.java
lbschanno: Fixed.

ivakegg (approved):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
```diff
@@ -58,32 +56,30 @@ public static boolean occurrence(Iterable<?> values, int count) {
...
     
     private static int getSizeOf(Iterable<?> iterable) {
-        Map<String,Integer> countMap = Maps.newHashMap();
-        int count = 0;
         if (iterable != null) {
-            // additional datatypes may have been added to this field to assist with evaluation and for use as
-            // a 'marker' for some other condition. We care only about how many were in the original so
-            // bin them all up to count what was really there in the beginning
+            int sourcedFromEvent = 0;
+            int sourcedFromIndex = 0;
```
ivakegg, line 62: I think I need a third category here for anything that is not a ValueTuple.  I am not sure this is needed but it seems odd to pass in an iterable of non ValueTuple objects and come back with a size of 0.  So if sourcedFromEvent is non-zero return that, else if sourceFromIndex is non-zero then return that , else return the number in the 3rd category..

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
lbschanno: This is now added.

ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
```diff
@@ -15,13 +15,15 @@
 import org.apache.hadoop.io.WritableComparable;
 import org.apache.hadoop.io.WritableUtils;
 import org.apache.log4j.Logger;
+import org.w3c.dom.Attr;
```
apmoriarty: I don't think is is needed

apmoriarty (dismissed):
LGTM

warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
lbschanno: Good catch, fixed!

apmoriarty (approved):
LGTM

ivakegg (approved):

PR #1373, opened 2021-12-17
fixes #1372: handle evaluation only fields

### LINKED ISSUE #1372: evaluation only fields not used correctly in RangeStream
The evaluation only fields are fields that may not have 'e' entries in the datawave metadata table but are still valid fields to query against because they are generated automatically when ever a document is aggregated for evaluation.  This is used appropriately in the DefaultQueryPlanner when ensuring that all the fields in the query exist.
However the RangeStream will mark a node as having an UNKNOWN_FIELD by use the metadata helper getAllFields.  This results in an ABSENT stream which can screw up the executability of the query.
The best solution may be to instrument the metadata helpers to include these evaluation only fields.

ivakegg:
* Updated metadata utils to 1.9
    * Updated the shard query logic to set the evaluation only fields on the metadata helper
    * Updated the test case to appropriately test evaluation only fields

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,7 +84,7 @@
         <version.microservice.base-rest-responses>1.2</version.microservice.base-rest-responses>
         <version.microservice.common-utils>1.0</version.microservice.common-utils>
         <version.microservice.dictionary-api>1.0</version.microservice.dictionary-api>
-        <version.microservice.metadata-utils>1.8.2</version.microservice.metadata-utils>
+        <version.microservice.metadata-utils>1.9</version.microservice.metadata-utils>
         <version.microservice.metrics-reporter>1.2</version.microservice.metrics-reporter>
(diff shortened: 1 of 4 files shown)


hlgp:
LGTM.  Only question is whether the new MetadataHelper should live in datawave-metadata-utils instead.

apmoriarty:
> LGTM. Only question is whether the new MetadataHelper should live in datawave-metadata-utils instead.

I don't think it's unreasonable to add evaluation only fields to the default implementation of the MetadataHelper.

jwomeara (dismissed):

ivakegg:
Working on moving the functionality into the base metadata helper class.

ivakegg:
Moved the use of evaluation only fields into the metadata-utils.  Please re-approve.

billoley (approved):

hlgp (approved):

PR #1374, opened 2021-12-20
fixes #1364 Process all eligible IndexStreams in Union.advanceQueue()

### LINKED ISSUE #1364: Range stream misses shards edge case
query && ((function && (SHARDS_AND_DAYS = 'X,Y,Z')) || (function && (SHARDS_AND_DAYS = 'X,Y,Z')))
where X,Y, and Z are days
All of the actually documents are in Z, but it only returns ranges for the first shard in Z.

Interestingly the following variants succeed in finding all of the ranges:
query && ((function && (SHARDS_AND_DAYS = 'Z')) || (function && (SHARDS_AND_DAYS = 'Z')))
query && ((function && (SHARDS_AND_DAYS = 'X,Z')) || (function && (SHARDS_AND_DAYS = 'X,Z')))

FineAndDandy:


warehouse/query-core/src/main/java/datawave/query/index/lookup/Union.java
```diff
@@ -198,13 +211,21 @@ public boolean hasNext() {
             pointers = pointers.union(itr.peek().second(), Lists.newArrayList(delayedNodes.getNodes()));
             itr.next();
             if (itr.hasNext()) {
-                children.add(itr);
+                // add this to the processed list so other iterators have a chance to be first even if this one comes back earlier in the stack
+                processedChildren.add(itr);
+                // children.add(itr);
```
apmoriarty: remove this comment

apmoriarty (dismissed):

alerman (approved):

warehouse/query-core/src/main/java/datawave/query/index/lookup/Union.java
```diff
@@ -198,13 +211,20 @@ public boolean hasNext() {
             pointers = pointers.union(itr.peek().second(), Lists.newArrayList(delayedNodes.getNodes()));
             itr.next();
             if (itr.hasNext()) {
-                children.add(itr);
+                // add this to the processed list so other iterators have a chance to be first even if this one comes back earlier in the stack
+                processedChildren.add(itr);
```
ivakegg, line 215: Do we always need to add it back in or only if it is a day?

warehouse/query-core/src/main/java/datawave/query/index/lookup/Union.java
```diff
@@ -198,13 +211,20 @@ public boolean hasNext() {
...
             } else {
                 if (log.isTraceEnabled()) {
                     log.trace("IndexStream exhausted for " + itr.getContextDebug());
                 }
             }
             childrenAdded = true;
+            
+            // repopulate children with all processed children
+            if (children.isEmpty() && !processedChildren.isEmpty()) {
```
ivakegg, line 224: Do the children need to be empty to add them in here?  And then if that is the case perhaps the code from 198-205 is not needed?

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/index/lookup/Union.java
ivakegg: ah I see.... just adding it into a new list.

warehouse/query-core/src/main/java/datawave/query/index/lookup/Union.java
ivakegg: Ah I see.... need to allow children to get empty first.

ivakegg (approved):

PR #1380, opened 2021-12-28
Create JexlNodeBuilder for creating JexlNodes

lbschanno:
Create an initial builder library for creating JexlNodes. This library
is aimed at reducing the verbosity involved in creating new JexlNode
instances, and at reducing bugs introduced by either incorrectly setting
or forgetting to set the correct parentage for child nodes.

This pull request also contains a proof-of-concept of the builder library
being put to use in JexlNodeFactory for several methods.


ivakegg (dismissed):

ivakegg:
nice work!

FineAndDandy (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeBuilder.java
```diff
@@ -0,0 +1,637 @@
...
+        T node = constructor.get();
+        node.image = this.image;
+        node.jjtSetValue(this.value);
+        
+        // Create the finalized children.
+        // @formatter:off
+        JexlNode[] childArray = children.stream()
+                        .filter(Objects::nonNull) // Filter out any possible null suppliers.
+                        .map(Supplier::get)       // Get the actual child nodes.
+                        .filter(Objects::nonNull) // Filter out any null children.
```
jwomeara: Should we consider throwing an exception if a child is null?  Or potentially having this be a configurable thing (whether or not to throw an exception when a null child is encountered.)

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeBuilder.java
```diff
@@ -0,0 +1,637 @@
...
+                        .toArray(JexlNode[]::new);
+        // @formatter:on
+        
+        // Set the parent for each child.
+        for (JexlNode child : childArray) {
+            child.jjtSetParent(node);
+        }
+        // Add each child to the root node.
+        for (int i = 0; i < childArray.length; i++) {
+            node.jjtAddChild(childArray[i], i);
```
jwomeara: This is a really bad call to make.  The underlying implementation creates a copy of the array with one additional entry, and then copies all of the elements each time you call jjtAddChild.  A better way of adding children to a parent node is to call JexlNodes.children(T node, JexlNode... children).  If you change the code to use this then you no longer need code to set the parent for each child - that is already handled by the call.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeBuilder.java
```diff
@@ -0,0 +1,637 @@
...
+
+/**
+ * A builder class that supports fluent-style building of {@link JexlNode} instances.
+ * 
+ * @param <T>
+ *            the {@link JexlNode} type
+ */
+public class JexlNodeBuilder<T extends JexlNode> {
+    
+    private final Supplier<T> constructor;
```
jwomeara: This is not so much a constructor supplier as it is an instance supplier.  I would consider renaming.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeBuilder.java
```diff
@@ -0,0 +1,637 @@
...
+     * Set the value for the node this builder will create.
+     * 
+     * @param value
+     *            the value
+     * @return this builder
+     */
+    public JexlNodeBuilder<T> withValue(Object value) {
+        this.value = value;
+        return this;
+    }
```
jwomeara, line 111: Should we consider throwing an exception on these methods if they are called when creating a JexlNode type which doesn't have an image or value?

warehouse/query-core/src/test/java/datawave/test/JexlNodeAssert.java
```diff
@@ -189,11 +189,101 @@ public JexlNodeAssert hasNumChildren(int numChildren) {
...
         isNotNull();
         return assertThat(actual.jjtGetChild(child));
     }
     
+    /**
+     * Returns a new {@link JexlNodeAssert} that will perform assertions on the child at index 0.
+     *
+     * @return the new {@link JexlNodeAssert} for the child
+     */
+    public JexlNodeAssert assertFirstChild() {
```
jwomeara, line 202: I guess these methods are convenient, but I'd be in favor of just having assertChild(int child) as that seems pretty clear and straightforward to me.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeBuilder.java
```diff
@@ -0,0 +1,637 @@
...
+     * Add a child node to this builder.
+     * 
+     * @param node
+     *            the child node
+     * @return this builder
+     */
+    public JexlNodeBuilder<T> withChild(JexlNode node) {
+        children.add(() -> node);
+        return this;
+    }
```
jwomeara: I would consider removing this.  IMO whoever uses the builder should have their children sorted out before they invoke the builder.

jwomeara (changes requested):
Have you done any performance testing comparing the original code with this code?  Given how fundamental building jexl nodes is to some of our code, we should get a feel for how these changes perform.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeBuilder.java
jwomeara: For example, and node, or node, etc.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeBuilder.java
lbschanno: I'm leaning towards making it configurable whether or not to throw an exception when a child is null. Do you have any preference for the default behavior?

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeBuilder.java
lbschanno: Good catch, I will update this to use JexlNodes.children instead.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeBuilder.java
lbschanno: Valid, I will rename it.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeBuilder.java
lbschanno: An interesting problem, and I'm inclined to say yes. I will also experiment with different design approaches to see if we can make it so that the methods `withImage()` and `withValue` are non-existent when creating nodes for which they're not supported.

Can you provide a list of what node types can/cannot have images/values, whichever one is shorter?

lbschanno:
@jwomeara I haven't done any extensive performance testing, but you make a good point of its importance, I'll try to get some performance metrics for you.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeBuilder.java
jwomeara: I would throw an exception by default.  That way we will hopefully find mistakes like that during development and not when things are deployed in production.  In the event that throwing an exception is disabled, I would log an error.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeBuilder.java
jwomeara: I don't know if I can make a comprehensive list, but the first few that come to mind that have neither an image or value are the AND/OR nodes.

lbschanno: Okay, understood.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeBuilder.java
lbschanno: Okay, I'll implement this behavior.

warehouse/query-core/src/test/java/datawave/test/JexlNodeAssert.java
lbschanno: I don't feel super strongly one way or the other, so I'm fine with removing the convenience methods.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeBuilder.java
lbschanno: I'm going to push back on this, I think this is a valid method for a builder to have so that we don't have to create a new array or collection instance before adding children to the builder. There are a number of places where we may just be adding one, or a few children (see some of the changes to JexlNodeFactory). I can add more detail to the documentation to make it clear that children added via this method will be in the order they're added, and possibly rename the method itself to try to make that clearer.

jwomeara: When the JexlNode is actually created, the children are stored as an array, regardless of the number of children you add.  Also, if we update the build method to use JexlNodes.children() to add children to a node, that method assigns the children parameter that is passed in to the children variable within the underlying SimpleNode object.  So I would argue that it's less wasteful to pass an array from the start since ultimately that array will be assigned directly to the children variable in the node.

Syntactically, to achieve a similar effect to 
```
builder
  .withChild(child1)
  .withChild(child2)
  .withChild(child3)
  .build()
```

you could easily change that to
```
builder
  .withChildren(new JexlNode[] {
    child1,
    child2,
    child3})
  .build()
```

And in the second case the array that you just created is the one that will be saved in the resulting JexlNode.  Unless you're thinking that someone might call `withChild` in a loop, I don't know that the `withChild` method is all that necessary.  This is just my 2 cents - the choice is ultimately yours.  If you decide to keep it anyway, then I would definitely update the docs as you described.

lbschanno: Children are stored as an array in a JexlNode, yes, but not in a JexlNodeBuilder. They're added to the list variable `JexlNodeBuilder.children`, and a finalized array of children is created when `JexlNodeBuilder.build()` is called. Your example
```
builder
  .withChildren(new JexlNode[] {
    child1,
    child2,
    child3})
  .build()
```
would result in each child in the array being added to the underlying list, not in it being directly saved in the resulting JexlNode. 

Although only supporting passing in an array of children would be very efficient, I would argue the goal of utilizing the builder pattern here is not just efficiency, but also convenience, and that there's a balance to be found. I prefer to retain the `withChild(JexlNode child)` method for now, and if it turns out not to be used in practice, remove it in a future optimization update.

lbschanno:
Closing due to significant changes needed to make compatible with jexl updates.

Issue #1270, opened 2021-09-01
ASTDelayedPredicate needs better handling


apmoriarty:
Right now the ASTDelayedPredicate marker can be applied multiple times to a query node resulting in multiple layers of wrapping. This is a problem when delayed markers are pulled up because only the outer wrapping layer is removed.

apmoriarty:
closing https://github.com/NationalSecurityAgency/datawave/pull/1271 in favor of https://github.com/NationalSecurityAgency/datawave/pull/1386 which was merged into 3.26

PR #1386, opened 2022-01-04
QueryPropertyMarker will not apply multiple times

apmoriarty:
One of the QueryPropertyMarker's create methods now checks to see if the source node or one of its ancestors is already marked. This prevents the same marker from being applied multiple times.

This handles the general case of the problem described in https://github.com/NationalSecurityAgency/datawave/issues/1270


warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java
```diff
@@ -49,7 +49,7 @@ public static String label() {
      *            the {@link QueryPropertyMarker} type
      * @return a new marker instance
      */
-    public static <MARKER extends QueryPropertyMarker> MARKER create(JexlNode source, Function<JexlNode,MARKER> constructor) {
+    public static <MARKER extends QueryPropertyMarker> JexlNode create(JexlNode source, Function<JexlNode,MARKER> constructor) {
```
lbschanno: I would like this to return a MARKER rather than a JexlNode. In this method we're able to be specific about the type of node we're returning (since we're not potentially returning the original source node), so why not be? Alternatively, if we want to verify the source node isn't already marked for every marker type we currently have, then perhaps this method should be deleted in favor of `create(source, class)`.

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/BoundedRange.java
```diff
@@ -30,7 +29,7 @@ public static String label() {
      * @return the new marker node
      * @see QueryPropertyMarker#create(JexlNode, Function)
      */
-    public static BoundedRange create(JexlNode node) {
+    public static JexlNode create(JexlNode node) {
```
lbschanno: This can return a BoundedRange if we return a MARKER from `create(source, constructor)`.

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/IndexHoleMarkerJexlNode.java
```diff
@@ -29,7 +28,7 @@ public static String label() {
      * @return the new marker node
      * @see QueryPropertyMarker#create(JexlNode, Function)
      */
-    public static IndexHoleMarkerJexlNode create(JexlNode node) {
+    public static JexlNode create(JexlNode node) {
```
lbschanno: This can return IndexHoleMarkerJexlNode if we return a MARKER from `create(source, constructor)`.

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java
```diff
@@ -161,6 +165,66 @@ protected void setupSource(JexlNode source) {
...
+    /**
+     * Determines if this source node is already marked with the provided marker class
+     *
+     * @param source
+     *            a JexlNode
+     * @param markerClass
+     *            an instance of a {@link QueryPropertyMarker}
+     * @return true if this node is already marked with the provided marker class
+     */
+    public static boolean isSourceAlreadyMarked(JexlNode source, Class<? extends QueryPropertyMarker> markerClass) {
```
lbschanno: I think this can be deleted in favor of just using `isSubTreeAlreadyMarked()`. See comment for that method.

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java
```diff
@@ -161,6 +165,66 @@ protected void setupSource(JexlNode source) {
...
+    }
+    
+    /**
+     * Unwrap a delayed predicate, fully. Intended to handle the odd edge case when multiple delayed predicate markers are applied to the same node
+     *
+     * @param node
+     *            an arbitrary jexl node
+     * @return the source node, or the original node if this node is not delayed
+     */
+    public static JexlNode unwrapFully(JexlNode node) {
```
lbschanno: I would like to see this method moved to ASTDelayedPredicate to further imply that this is an operation that specifically revolves around delayed predicate markers. Additionally, this can be simplified to
```java
public static JexlNode unwrapFully(JexlNode node) {
    QueryPropertyMarker.Instance instance = QueryPropertyMarker.findInstance(node);
    if (instance.isType(ASTDelayedPredicate.class)) {
        return unwrapFully(instance.getSource());
    }
    return node;
}
```

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java
```diff
@@ -161,6 +165,66 @@ protected void setupSource(JexlNode source) {
...
+    
+    /**
+     * Determine if one the source node's ancestors is already marked. For example, is this subtree already marked as delayed?
+     *
+     *
+     * @param node
+     *            a JexlNode
+     * @return true
+     */
+    public static boolean isSubTreeAlreadyMarked(JexlNode node, Class<? extends QueryPropertyMarker> markerClass) {
```
lbschanno: The original source node is passed into this, which is then subsequently checked to see if it's marked before checking for a marked ancestor. Do we really need isSourceMarked() given this?

Additionally, I don't love `isSubTreeAlreadyMarked()` as a name, to my mind that implies that we're descending into the source's children rather than ascending through the source's parentage. How do you feel about `isPartOfMarkedSubTree()` or something similar? Alternatively, we could pass in the source's parent, check for null first, and rename this to `hasMarkedAncestor()`.

Additionally pt. 2, this can be a private method.

lbschanno (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java
apmoriarty: I'll revert this change along with any supporting changes necessary. While I would like to check the source node for every create, that may be overkill in which case having an alternate create method that is faster would be nice.

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java
apmoriarty: I'm looking at refactoring this to `unwrapFully(node, clazz)` so the caller can specify a marker class. I think that case would address your comment?

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java
apmoriarty: I like renaming the method to `hasMarkedAncestor` or `isAncestorMarked` and making the related changes you suggested.

Ideally, both `hasMarkedAncestor` and `isSourceMarked` would be methods available to any caller. I can imagine a situation where a visitor would need to know if an ancestor is marked or not before modifying the query tree.

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java
lbschanno: Yes it would. I'm fine with leaving this in QueryPropertyMarker in that case.

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java
lbschanno: You make a good point about leaving it public for outside callers to use. I would add a note to the documentation for `isSubTreeAlreadyMarked` to stress that it will traverse up the node's entire ancestry, and not just the immediate ancestor to make a note of comparative expense.

apmoriarty: I'll update the docs

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java
apmoriarty: Resolving per our other conversation

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java
```diff
@@ -59,18 +59,22 @@ public static String label() {
...
      * @param <MARKER>
      *            the marker type
-     * @return the new query property marker instance
+     * @return the new query property marker instance, or the original source if the marker has already been applied
      */
-    public static <MARKER extends QueryPropertyMarker> MARKER create(JexlNode source, Class<MARKER> markerType) {
+    public static <MARKER extends QueryPropertyMarker> JexlNode create(JexlNode source, Class<MARKER> markerType) {
         try {
+            if (isSourceMarked(source, markerType) || isAncestorMarked(source, markerType)) {
+                return source;
```
FineAndDandy, line 76: should this return the passed source or the source of the QueryPropertyMarker? They may not be the same.

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java
```diff
@@ -161,6 +165,55 @@ protected void setupSource(JexlNode source) {
...
+    }
+    
+    /**
+     * Determine if one the source node's ancestors is already marked. For example, is this subtree already marked as delayed?
+     *
+     * @param node
+     *            a JexlNode
+     * @return true
+     */
+    public static boolean isAncestorMarked(JexlNode node, Class<? extends QueryPropertyMarker> markerClass) {
```
FineAndDandy, line 198: This method could be really expensive depending on how many steps up the tree we have to go. Maybe note this in the comments.

FineAndDandy: alternatively add a cut off depth to the search for the marker in the visitor? So we don't do the same work multiple times.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnexecutableNodesVisitor.java
```diff
@@ -180,7 +181,8 @@ private void addDelays(JexlNode node, Object data) {
                 ExecutableDeterminationVisitor.STATE state = ExecutableDeterminationVisitor.getState(child, data, config, indexedFields, indexOnlyFields,
                                 nonEventFields, forFieldIndex, null, helper);
                 if (state == ExecutableDeterminationVisitor.STATE.PARTIAL) {
-                    ASTDelayedPredicate.create(child);
+                    JexlNode delayed = ASTDelayedPredicate.create(child);
+                    JexlNodes.replaceChild(node, child, delayed);
```
FineAndDandy: Is this fixing another bug? Or was this functionality moved out by this PR? I didn't see it when I looked so assume this was another bug?

FineAndDandy (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnexecutableNodesVisitor.java
apmoriarty: One of the changes in this pull request was to not automatically reparent a source when a marker node is applied, letting the caller handle that operation. So this addition here maintains the initial functionality.

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java
apmoriarty: It should return the original node that was passed in. This keeps things simple for the caller.

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java
apmoriarty: Updated docs forthcoming

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/ASTDelayedPredicate.java
```diff
@@ -28,10 +26,10 @@ public static String label() {
      * @param node
      *            the source node
      * @return the new marker node
-     * @see QueryPropertyMarker#create(JexlNode, Function)
+     * @see QueryPropertyMarker#create(JexlNode, Class)
```
FineAndDandy, line 30: add comment describing difference in behavior

FineAndDandy (changes requested):

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/ASTDelayedPredicate.java
apmoriarty: done!

FineAndDandy (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java
```diff
@@ -161,6 +165,59 @@ protected void setupSource(JexlNode source) {
...
+    }
+    
+    /**
+     * Unwrap a marker node, fully. Intended to handle the odd edge case when multiple instances of the same marker is applied to the same node
+     *
+     * @param node
+     *            an arbitrary jexl node
+     * @return the source node, or the original node if the source node is not marked
+     */
+    public static JexlNode unwrapFully(JexlNode node, Class<? extends QueryPropertyMarker> marker) {
```
ivakegg, line 219: can we change this to getSource to be consistent with the original code?

ivakegg: In fact I am wondering if the instance.getSource() call should be doing this work instead.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownLowSelectivityNodesVisitor.java
```diff
@@ -58,7 +59,10 @@ public Object visit(ASTReference node, Object data) {
     public Object visit(ASTEQNode node, Object data) {
         // if this node represents a field/value that has poor selectability, then push it down
         if (hasLowSelectability(node)) {
-            return ASTDelayedPredicate.create(node);
+            JexlNode parent = node.jjtGetParent();
+            JexlNode delayed = ASTDelayedPredicate.create(node);
```
ivakegg: the create call should already be reparenting

ivakegg (changes requested):
The create call should be doing all of the necessary reparenting.  If not then that is the actually bug.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownLowSelectivityNodesVisitor.java
apmoriarty: I'll update per our conversation

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java
apmoriarty: This is sort of a placeholder method. It will unwrap a node tree for a specific marker type, for example, multiple delayed predicate markers. Once this feature is merged this method becomes obsolete since the same marker cannot be applied multiple times. 

If you are talking about unwrapping EVERY marker node from a given subtree, say unwrapping the source of a delayed bounded range, that would be something else entirely.

ivakegg (approved):

FineAndDandy (approved):

lbschanno (approved):

PR #1388, opened 2022-01-06
Add visibility/authorization validation to unit test framework

drewfarris:
Currently the unit test framework doesn't do anything to validate the operations of authorizations and visibilities. This PR changes the `CitiesDataType` to use its [ACCESS](https://github.com/NationalSecurityAgency/datawave/blob/afbdc0546c23562dfa5e66a193a094f574200609/warehouse/query-core/src/test/java/datawave/query/testframework/CitiesDataType.java#L87) field to populate the `columnVisibility` security marking in events and then adds a test to validate that they are operating  correctly.

Minor modifications are made elsewhere, such as other tests in datawave-query-core, in order to work with this new approach.

As a part of this work, it was necessary to produce an `ExtendedTestCSVIngestHelper` in order to properly propagate security markings to the `NormalizedContentInterface` instances generated by the `ExtendedCSVIngestHelper` produced when reading events from a sequence file in the [CSVTestFileLoader](https://github.com/NationalSecurityAgency/datawave/blob/afbdc0546c23562dfa5e66a193a094f574200609/warehouse/query-core/src/test/java/datawave/query/testframework/CSVTestFileLoader.java#L33)

### DIFF
--- a/warehouse/query-core/src/test/java/datawave/query/QueryAuthsTest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/QueryAuthsTest.java
@@ -26,7 +26,6 @@ import org.junit.ClassRule;
 import org.junit.Test;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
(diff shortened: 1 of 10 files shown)


warehouse/query-core/src/test/java/datawave/query/QueryAuthsTest.java
```diff
@@ -0,0 +1,225 @@
...
+    @BeforeClass
+    public static void filterSetup() throws Exception {
+        Collection<DataTypeHadoopConfig> dataTypes = new ArrayList<>();
+        FieldConfig generic = new GenericCityFields();
+        // Set<String> virt = new HashSet<>(Arrays.asList(CityField.CITY.name(), CityField.CONTINENT.name()));
+        // generic.removeVirtualField(virt);
+        // generic.addIndexField(CityField.CODE.name());
+        // for (String idx : generic.getIndexFields()) {
+        // generic.addReverseIndexField(idx);
+        // }
```
keith-ratcliffe: No longer needed?

keith-ratcliffe (changes requested):
Aside from the one comment, looks good. Thanks for adding this!

apmoriarty (dismissed):
Agree with Keith's comment, otherwise LGTM!

keith-ratcliffe (approved):

apmoriarty (approved):

PR #1390, opened 2022-01-07
Address #419 PushdownQueryTest test filter include regex with expansion

hlgp:


apmoriarty (dismissed):

warehouse/query-core/src/test/java/datawave/query/PushdownQueryTest.java
```diff
@@ -138,6 +138,21 @@ public void testDelayedFilterIncludeRegex() throws Exception {
         }
     }
     
+    @Test
+    public void testDelayedFilterIncludeRegexExpansion() throws Exception {
+        log.info("------  testErrorFilterIncludeRegexExpansion  ------");
```
ivakegg: wrong test name in log message?

ivakegg (dismissed):

warehouse/query-core/src/test/java/datawave/query/PushdownQueryTest.java
hlgp: This is a silly reason, but the ticket explicitly says to "Duplicate testErrorFilterIncludeRegex but do not disable expansion." in PushdownQueryTest, which only matches the info statement and not the actual method, so I tried to make it clear how the two were related by maintaining the same level of wrongness between method name and info statement.  I'll fix them both.

apmoriarty (approved):

jwomeara (approved):

PR #1396, opened 2022-01-13
Event results should maintain an overall colvis when groupby and redu…

jschmidt10:
We have a bug where using a group by query along with a reduced.response parameter results in the overall Event columnVisibility being empty.

The bug fix I did was the path of least risk/resistance. Now, with reduced.response, we will still serialize the field visibilities from the tserver back to the web tier. The transformer on the web side will eliminate them in the actual HTTP responses. 

There's an optimization where we change the Document to always serialize it's top level visibility so that we don't waste cycles serializing the individual field visibilities when using reduced.response. However, that'll take more work and testing. I can open up a follow on ticket to do that if we're interested. For now, this fixes the immediate issue.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
+++ b/warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
@@ -155,64 +155,48 @@ public abstract class Attribute<T extends Comparable<T>> implements WritableComp
     }
     
     protected void writeMetadata(DataOutput out, Boolean reducedResponse) throws IOException {
-        out.writeBoolean(reducedResponse);
-        
-        if (!reducedResponse) {
-            out.writeBoolean(isMetadataSet());
-            if (isMetadataSet()) {
-                byte[] cvBytes = getColumnVisibility().getExpression();
-                
-                WritableUtils.writeVInt(out, cvBytes.length);
-                
-                out.write(cvBytes);
-                out.writeLong(getTimestamp());
-            }
+        out.writeBoolean(isMetadataSet());
+        if (isMetadataSet()) {
+            byte[] cvBytes = getColumnVisibility().getExpression();
+            
+            WritableUtils.writeVInt(out, cvBytes.length);
+            
+            out.write(cvBytes);
+            out.writeLong(getTimestamp());
         }
     }
     
     protected void writeMetadata(Kryo kryo, Output output, Boolean reducedResponse) {
-        output.writeBoolean(reducedResponse);
-        
-        if (!reducedResponse) {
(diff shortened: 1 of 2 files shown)


apmoriarty (approved):

warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
```diff
@@ -155,64 +155,48 @@ protected void clearMetadata() {
...
-            out.writeBoolean(isMetadataSet());
-            if (isMetadataSet()) {
-                byte[] cvBytes = getColumnVisibility().getExpression();
-                
-                WritableUtils.writeVInt(out, cvBytes.length);
-                
-                out.write(cvBytes);
-                out.writeLong(getTimestamp());
-            }
+        out.writeBoolean(isMetadataSet());
```
ivakegg, line 158: This is no longer based on the reducedResponse boolean being sent in.  Is that what you meant to do?  I think the metadata will basically always be set on the attribute.
The reduced response was an effort to reduce the amount of data being sent back to the client (webserver/microservice).  If you are removing that concept altogether, then we shouldn't have a reduced response boolean here nor should that option be sent from the DefaultQueryPlanner over to the QueryIterator. via the scan options.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
jschmidt10: Yes, that is what I meant to do. I can work on ripping it all the way out. I was initially targeting a quick bug fix to get the events marked correctly. I'll start working on fully taking it out though.

ivakegg (approved):

warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
jschmidt10: Had a discussion with Ivan. We're going to merge this as-is and I opened up a follow on ticket for a better long-term fix, which will be serializing the document level colvis always and then reverting this change.

keith-ratcliffe:
The download.jboss.org endpoint for the Wildfly tarball was temporarily down, causing jobs to fail, but seems to be back online now. Rerunning jobs...

keith-ratcliffe:
> The download.jboss.org endpoint for the Wildfly tarball was temporarily down, causing jobs to fail, but seems to be back online now. Rerunning jobs...

Scratch that....the URL needs to be changed to https:// in order for it to work. Update forthcoming

PR #1399, opened 2022-01-21
DATAWAVE-1395: Updated the virtual field generation to more effecient…

### LINKED ISSUE #1395: Virtual field performance issue
The virtual field logic can cause performance issues in the following manner.  In general a pattern of field1-field2 will have a performance of N*M where N is the number of field1 values and M is the number of field2 values.  This can of course be problematic when N and M are large.  In general we cannot get past that however when it comes to grouping we can reduce M to 1.  In one case we had a set of grouped fields that were in the 10's of thousands, and we had a virtual field that was combining two of the fields therein.  The result was that it would take up to a full hour to actually compute the virtual fields when it could have been done in under a second.  The adjustment to the algorithm is to create a hash map of group to value when the grouping policy is SAME_GROUP.

ivakegg:
…ly process fields when grouping is to be considered.

### DIFF
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/VirtualIngest.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/VirtualIngest.java
@@ -1,31 +1,31 @@
 package datawave.ingest.data.config.ingest;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.Map.Entry;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Multimap;
 import datawave.data.type.NoOpType;
 import datawave.ingest.data.Type;
 import datawave.ingest.data.config.GroupedNormalizedContentInterface;
 import datawave.ingest.data.config.NormalizedContentInterface;
 import datawave.ingest.data.config.NormalizedFieldAndValue;
-
 import datawave.marking.MarkingFunctions;
-import org.apache.commons.lang.NotImplementedException;
 import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang.builder.HashCodeBuilder;
+import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.hadoop.conf.Configuration;
 
-import com.google.common.base.Objects;
-import com.google.common.collect.HashMultimap;
-import com.google.common.collect.Multimap;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public interface VirtualIngest {
     
@@ -339,6 +339,12 @@ public interface VirtualIngest {
             return markings1;
         }
         
+        /**
+         * Create the normalized form of a map of fields, and add the virtual fields as configured above.
+         * 
+         * @param fields
+         * @return The multimap of normalized fields including virtual fields.
+         */
         public Multimap<String,NormalizedContentInterface> normalize(Multimap<String,String> fields) {
             Multimap<String,NormalizedContentInterface> eventFields = HashMultimap.create();
             for (Entry<String,String> field : fields.entries()) {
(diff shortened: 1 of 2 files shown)


warehouse/ingest-core/src/test/java/datawave/ingest/data/config/ingest/VirtualFieldTest.java
```diff
@@ -0,0 +1,108 @@
...
+        eventFields.put("PARTIAL_1", new NormalizedFieldAndValue("PARTIAL_1", "value5"));
+        
+        eventFields.put("PARTIAL_2", new NormalizedFieldAndValue("PARTIAL_2", "value1"));
+        eventFields.put("PARTIAL_2", new NormalizedFieldAndValue("PARTIAL_2", "value2"));
+        eventFields.put("PARTIAL_2", new NormalizedFieldAndValue("PARTIAL_2", "value3", "group3", "subgroup1"));
+        eventFields.put("PARTIAL_2", new NormalizedFieldAndValue("PARTIAL_2", "value4", "group4", "subgroup1"));
+        eventFields.put("PARTIAL_2", new NormalizedFieldAndValue("PARTIAL_2", "value5", "group5", "subgroup1"));
+    }
+    
+    // config.set("test" + VirtualIngest.VIRTUAL_FIELD_GROUPING_POLICY, "SAME_GROUP_ONLY,GROUPED_WITH_NON_GROUPED,IGNORE_GROUPS")
```
hlgp: testing leftovers?

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/VirtualIngest.java
```diff
@@ -383,11 +405,58 @@ private void compilePatterns() {
...
+            public VirtualFieldGrouping(String group0, String group1) {
+                this.group0 = group0;
+                this.group1 = group1;
+            }
+            
+            public String getGroup1() {
+                return group1;
+            }
+            
+            public String getGroup0() {
```
hlgp: Tiny pet peeve. Can we call these something more like getGroup and getSubGroup rather than maintaining an homage to the now defunct String array position in the method name?

warehouse/ingest-core/src/test/java/datawave/ingest/data/config/ingest/VirtualFieldTest.java
```diff
@@ -0,0 +1,108 @@
...
+        eventFields.put("GROUPED_2", new NormalizedFieldAndValue("GROUPED_2", "value4", "group4", "subgroup1"));
+        eventFields.put("GROUPED_2", new NormalizedFieldAndValue("GROUPED_2", "value5", "group5", "subgroup1"));
+        
+        eventFields.put("UNGROUPED_1", new NormalizedFieldAndValue("UNGROUPED_1", "value1"));
+        eventFields.put("UNGROUPED_1", new NormalizedFieldAndValue("UNGROUPED_1", "value2"));
+        eventFields.put("UNGROUPED_1", new NormalizedFieldAndValue("UNGROUPED_1", "value3"));
+        eventFields.put("UNGROUPED_1", new NormalizedFieldAndValue("UNGROUPED_1", "value4"));
+        eventFields.put("UNGROUPED_1", new NormalizedFieldAndValue("UNGROUPED_1", "value5"));
+        
+        eventFields.put("UNGROUPED_2", new NormalizedFieldAndValue("UNGROUPED_1", "value1"));
```
javease: Should this set of puts be using "UNGROUPED_2" for the first parameter of NormalizedFieldAndValue rather than "UNGROUPED_1"? (It doesn't impact your results since you're not comparing fields, but for correctness sake.)

ivakegg: you are correct that it really does not mapper... but I will fix that

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/VirtualIngest.java
ivakegg: sure, why not

warehouse/ingest-core/src/test/java/datawave/ingest/data/config/ingest/VirtualFieldTest.java
ivakegg: yes

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/VirtualIngest.java
```diff
@@ -383,11 +405,58 @@ private void compilePatterns() {
...
+            }
+            
+            @Override
+            public String toString() {
+                return group0 + '.' + group1;
+            }
+        }
+        
         // Create the virtual field from the event Fields members
         public void addVirtualFields(List<NormalizedContentInterface> virtualFields, Multimap<String,NormalizedContentInterface> eventFields,
```
javease, line 500: It might help (me, at least) if there were some comments for this method that describe how we're going about adding the appropriate fields.

matthpeterson (dismissed):

dtspence (dismissed):

matthpeterson (approved):

dtspence (approved):

PR #1404, opened 2022-01-28
Add SHARDS_AND_DAYS via parameter

### LINKED ISSUE #1405: Support setting SHARDS_AND_DAYS hint via query option
The SHARDS_AND_DAYS hint is typically set via the query string. However, certain syntax parsers do not support this. In these cases the user must specify the hint as a query parameter, but the default query planner does not extract the hint.

The query options map is the only place the default query planner looks for the shards and days hint. This map is only populated under certain circumstances, and only the query string itself is examined for the hint. The default query planner should fall back to searching the query settings for the hint.

Add support for setting the shards and days hint from the query settings.

apmoriarty:
Fixes https://github.com/NationalSecurityAgency/datawave/issues/1405

DefaultQueryPlanner will look in query settings for SHARDS_AND_DAYS if it is not found in the options map

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
+++ b/warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
@@ -704,6 +704,13 @@ public class DefaultQueryPlanner extends QueryPlanner implements Cloneable {
         
         if (optionsMap.containsKey(QueryParameters.SHARDS_AND_DAYS)) {
             config.setQueryTree(timedAddShardsAndDaysFromOptions(timers, config.getQueryTree(), optionsMap));
+        } else {
+            // look for the shards and days hint in the query settings
+            Parameter parameter = settings.findParameter(QueryParameters.SHARDS_AND_DAYS);
+            if (StringUtils.isNotBlank(parameter.getParameterValue())) {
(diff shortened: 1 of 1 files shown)


warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -704,6 +704,13 @@ protected ASTJexlScript updateQueryTree(ScannerFactory scannerFactory, MetadataH
         
         if (optionsMap.containsKey(QueryParameters.SHARDS_AND_DAYS)) {
             config.setQueryTree(timedAddShardsAndDaysFromOptions(timers, config.getQueryTree(), optionsMap));
+        } else {
+            // look for the shards and days hint in the query settings
```
FineAndDandy, line 708: I would like to see a comment here explaining why the query parameter is insufficient

FineAndDandy (changes requested):

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
apmoriarty: updated

avgAGB (approved):

jschmidt10 (approved):

FineAndDandy (approved):

PR #1409, opened 2022-02-01
Feature/one splits file to rule them all

hlgp:
Merge NonShardedSplitsFile and ShardedTableMapFile.  Use TableSplitsCache instead of reaching out to Accumulo in every reducer.

Splits File generation now takes configured Partitioner needs into account.
-  If the table's partitioner doesn't need splits, e.g. HashPartitioner, don't write them
-  If the table's partitioner only needs splits, only write the splits
-  If the table's partitioner needs splits and locations, write them
-  If the table's partitioner needs splits in a specific order, write them in that order, so we do not need to sort them at the beginning of every mapper.


warehouse/ingest-core/src/main/java/datawave/ingest/config/BaseHdfsFileCacheUtil.java
```diff
@@ -34,15 +34,22 @@ public Path getCacheFilePath() {
...
-            readCache(in, delimiter);
-        } catch (IOException ex) {
-            if (shouldRefreshCache(this.conf)) {
-                update();
-            } else {
-                throw new IOException("Unable to read cache file at " + this.cacheFilePath, ex);
-            }
+        int attempts = 0;
+        boolean retry = true;
+        while (retry && attempts <= MAX_RETRIES) {
```
ivakegg, line 48: no need to check for attempts here as you blow out with an exception when reaching this limit below.

hlgp: true!

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/TableSplitsCache.java
```diff
@@ -211,21 +268,36 @@ private boolean exceedsMaxSplitsDeviation(Map<String,Integer> tmpSplitsPerTable)
     @Override
     protected void readCache(BufferedReader in, String delimiter) throws IOException {
```
ivakegg: Same comment about the delimiter arg as for the other pull request

hlgp: I planned to address this once the other MR is merged

warehouse/ingest-core/src/main/java/datawave/ingest/config/BaseHdfsFileCacheUtil.java
```diff
@@ -17,7 +17,9 @@ public abstract class BaseHdfsFileCacheUtil {
     protected Path cacheFilePath;
     protected final Configuration conf;
     protected AccumuloHelper accumuloHelper;
+    
     protected String delimiter = "\t";
+    private static final int MAX_RETRIES = 3;
```
ivakegg, line 23: Can we override this with configuration?  Is it worth it?  I don't know the right answer here, just asking.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/SplitsFile.java
```diff
@@ -0,0 +1,184 @@
...
+        boolean doValidation = conf.getBoolean(SHARD_VALIDATION_ENABLED, false);
+        
+        try {
+            log.info("Base splits: " + baseSplitsPath);
+            
+            Path destSplits = new Path(conf.get(SPLIT_WORK_DIR) + "/"
+                            + conf.get(TableSplitsCache.SPLITS_CACHE_FILE, TableSplitsCache.DEFAULT_SPLITS_CACHE_FILE));
+            log.info("Dest splits: " + destSplits);
+            
+            FileUtil.copy(sourceFs, baseSplitsPath, destFs, destSplits, false, conf);
```
ivakegg, line 54: I am confused here.  Why are we copying this file if the TableSplitsCache is already handling reading the original file?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/TableSplitsCache.java
```diff
@@ -35,25 +40,68 @@ public class TableSplitsCache extends BaseHdfsFileCacheUtil {
...
+    
     private Path splitsPath = null;
-    private Map<String,List<Text>> splits = null;
+    private Map<String,List<Text>> splits = new HashMap<>();
+    private Map<String,Map<Text,String>> splitLocations = new HashMap<>();
+    
+    private Map<Text,String> getSplitsWithLocation(String table) throws AccumuloException, AccumuloSecurityException, TableNotFoundException {
+        SortedMap<KeyExtent,String> tabletLocations = new TreeMap<>();
+        SortedMap<Text,String> tabletLocationsByEndRow = new TreeMap<>();
+
```
ivakegg:
```
        tabletLocationsByEndRow = tabletLocations.entrySet().stream().
                collect(Collectors.toMap(e -> e.getKey().getEndRow(),
                        e -> e.getValue() == null ? NO_LOCATION : e.getValue(),
                        (o1, o2) -> o1,
                        TreeMap::new));
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/TableSplitsCache.java
```diff
@@ -141,18 +193,23 @@ public void writeCacheFile(FileSystem fs, Path tmpSplitsFile) throws IOException
...
         Set<String> tableNames = getIngestTableNames();
         Map<String,Integer> splitsPerTable = new HashMap<>();
+        Map<String,Map<Text,String>> tmpSplitLocations = new HashMap<>();
         if (tops != null) {
             try (PrintStream out = new PrintStream(new BufferedOutputStream(fs.create(tmpSplitsFile)))) {
                 this.splits = new HashMap<>();
                 // gather the splits and write to PrintStream
                 for (String table : tableNames) {
                     log.info("Retrieving splits for " + table);
+                    Map<Text,String> splitLocations = getSplitsWithLocation(table);
```
ivakegg: If we already have all of the split locations, why get the splits again in the next line?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/partition/MultiTableRangePartitioner.java
```diff
@@ -115,35 +117,38 @@ public static void setContext(TaskInputOutputContext<?,?,?,?> context) {
...
-        writer.createFile(isTrimmed());
-        return writer.getUri();
-    }
-    
-    protected boolean isTrimmed() {
-        return true;
-    }
-    
-    @Override
+    // <<<<<<< HEAD
```
ivakegg: Did you forget to resolve this?

ivakegg (changes requested):

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/SplitsFile.java
hlgp: This is preserving the current workflow.  setupFile is called by IngestJob, where it uses or updates the base splits file, then generates copies in the work dir.  The mappers reference the splits files in their job's work dir to avoid a thundering herd against the base file only.  I opted to keep this behavior the same for now, but we could change it and use distributedcache or some other mechanism.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/TableSplitsCache.java
hlgp: oversight.  good catch

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/partition/MultiTableRangePartitioner.java
hlgp: yep, this commented out code can be deleted now

warehouse/ingest-core/src/main/java/datawave/ingest/config/BaseHdfsFileCacheUtil.java
hlgp: Not sure if it's worth it, honestly.  I could go either way.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/TableSplitsCache.java
hlgp: throws an NPE

hlgp: Added a filter for endRow not null.  1ms slower on average than the regular for loop.  Not a big deal, but not sure I see the appeal.

ivakegg (dismissed):

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/partition/TestShardGenerator.java
```diff
@@ -39,17 +42,49 @@ public TestShardGenerator(Configuration conf, File tmpDir, int numDays, int shar
...
+
+            for (String table : tableNames) {
+
+                for (Map.Entry<Text,String> entry : sortedLocations.entrySet()) {
+                    out.println(table + "\t" + new String(Base64.encodeBase64(entry.getKey().toString().getBytes())).trim() + "\t" + entry.getValue());
+                }
+            }
+        }
+
+        // writeSplits(locations, tmpDirectory, BALANCEDISH_SHARDS_LST);
```
apmoriarty: remove?

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/partition/TestShardGenerator.java
```diff
@@ -39,17 +42,49 @@ public TestShardGenerator(Configuration conf, File tmpDir, int numDays, int shar
...
+            for (String table : tableNames) {
+
+                for (Map.Entry<Text,String> entry : sortedLocations.entrySet()) {
+                    out.println(table + "\t" + new String(Base64.encodeBase64(entry.getKey().toString().getBytes())) + "\t" + entry.getValue());
+                }
+            }
+        }
+        conf.set(TableSplitsCache.SPLITS_CACHE_DIR, tmpDir.getAbsolutePath());
+
+        // writeSplits(locations, tmpDirectory, BALANCEDISH_SHARDS_LST);
```
apmoriarty: remove?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
```diff
@@ -249,7 +249,7 @@ protected void printUsage() {
 
     @Override
     public int run(String[] args) throws Exception {
-        long setupStart = System.currentTimeMillis();
+        long setupstart = System.currentTimeMillis();
```
alerman: Why change? Do you not like Camel Case?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
```diff
@@ -826,12 +826,16 @@ protected Configuration parseArguments(String[] args, Configuration conf) throws
...
-            conf.set("split.work.dir", conf.get("ingest.work.dir.qualified"));
+        if (null == conf.get(SplitsFile.SPLIT_WORK_DIR)) {
+            conf.set(SplitsFile.SPLIT_WORK_DIR, conf.get("ingest.work.dir.qualified"));
         }
         conf.setInt("splits.num.reduce", this.reduceTasks);
         // used by the output formatter and the sharded partitioner
-        ShardedTableMapFile.setupFile(conf);
+        long before = System.currentTimeMillis();
+        SplitsFile.setupFile(job, conf);
+        long after = System.currentTimeMillis();
```
alerman, line 836: This could be a follow on task, but perhaps we should make use of org.springframework.util.StopWatch class to make this prettier. StopWatch has a nice prettyPrint method that will show the percentage time for each stage too.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/TableSplitsCache.java
```diff
@@ -2,30 +2,51 @@
...
+import java.util.stream.Collectors;
 
-import org.apache.accumulo.core.client.AccumuloClient;
+import org.apache.accumulo.core.client.Accumulo;
 import org.apache.accumulo.core.client.AccumuloException;
 import org.apache.accumulo.core.client.AccumuloSecurityException;
 import org.apache.accumulo.core.client.TableNotFoundException;
-import org.apache.accumulo.core.client.admin.TableOperations;
+import org.apache.accumulo.core.client.security.tokens.PasswordToken;
+import org.apache.accumulo.core.clientImpl.ClientConfConverter;
```
alerman: Note that these classes are marked as being in the private Accumulo API. We have been trying to reduce/remove references to private API where possible. I had already gone through and removed Threads references before (and in fact in newer versions of 2.1.3 the api has changed and this will likely break).

You can see the build warnings for import control with a regular maven build:

```
[WARN] warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/TableSplitsCache.java:27:1: Disallowed import - org.apache.accumulo.core.clientImpl.ClientConfConverter. [ImportControl]
[WARN] warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/TableSplitsCache.java:28:1: Disallowed import - org.apache.accumulo.core.clientImpl.ClientContext. [ImportControl]
[WARN] warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/TableSplitsCache.java:29:1: Disallowed import - org.apache.accumulo.core.clientImpl.ClientInfo. [ImportControl]
[WARN] warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/TableSplitsCache.java:31:1: Disallowed import - org.apache.accumulo.core.dataImpl.KeyExtent. [ImportControl]
[WARN] warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/TableSplitsCache.java:32:1: Disallowed import - org.apache.accumulo.core.metadata.MetadataServicer. [ImportControl]
[WARN] warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/TableSplitsCache.java:33:1: Disallowed import - org.apache.accumulo.core.singletons.SingletonReservation. [ImportControl]
[WARN] warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/TableSplitsCache.java:34:1: Disallowed import - org.apache.accumulo.core.util.threads.Threads. [ImportControl]
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
ivakegg: I was being consistent with the setupstop variable.  I had to choose 1 due to a merge conflict.  I can change them all to camel case if you care that much.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
ivakegg: agreed

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/TableSplitsCache.java
ivakegg: oh, let me take a look at that

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/partition/TestShardGenerator.java
ivakegg: ok

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/partition/TestShardGenerator.java
ivakegg: ok

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
ivakegg: I will change them

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
ivakegg: not changing that right now...

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/TableSplitsCache.java
alerman: Looks like it at least doesnt break after building with new Accumulo jars. The breaking change was in ThreadPools not Threads, but it would be good to remove what we can. If there are things here that cannot be removed or that should be public in accumulo we should start that discussion with the accumulo team and add the class to import-control-accumulo.xml

ivakegg: I don't think there is a public API to do this currently.  I posted a question in ASF slack to see what I can do.  I suppose I could simply scan the metadata table but I would really prefer a better way.

ivakegg: I should be able to remove the use of Threads at least.

ivakegg: Actually the accumulo client is a client context, so I can remove a bunch of it.  However I cannot get rid of the metadata servicer.

warehouse/ingest-core/src/test/resources/datawave/ingest/mapreduce/job/all-splits.txt
austin007008, line 1: While you're in here you might want to fix the test data in this and the corresponding unit tests. those "splits" are actually memory locations from something (I can't remember what) and they should be some kind of base64. The only reason they don't fail on the encode/decode is because apache commons base64 will happily let you input and output bad data . you can see https://github.com/NationalSecurityAgency/datawave/pull/2480/files for an example of how they should look

alerman (approved):

apmoriarty (approved):

PR #1410, opened 2022-02-01
WIP - #877 - Adding the ability to track if field has already been ex…

### LINKED ISSUE #877: Avoid normalizing unfielded lookup results
When we do an unfielded lookup, it may not be necessary to subsequently expand those terms using different normalizations as we have already retrieved the actual values from the index.

codebyneesh:
…panded and provide a cache to map the expanded field to.


foster33:
See #2218

PR #1411, opened 2022-02-01
Task/locking

ejrgilbert:
Added locking mechanism for easy reuse. Added lock to `load-job-cache.sh`


warehouse/assemble/datawave/pom.xml
```diff
@@ -267,7 +267,7 @@
                                             <artifactId>datawave-ingest-scripts</artifactId>
                                         </artifactItem>
                                     </artifactItems>
-                                    <excludes>META-INF/**,bin/ingest/ingest-versions.sh,bin/ingest/ingest-env.sh,bin/system/install.sh</excludes>
+                                    <excludes>META-INF/**,bin/ingest/ingest-versions.sh,bin/ingest/ingest-env.sh,bin/system/install.sh,bin/util/file_locker.sh</excludes>
```
javease, line 304: I realize you're just updating what was present, but I don't understand why we have two blocks that appear to do the same thing. I'd say the second block, with "filter" in its id name, is the "correct" one as we want these scripts "filtered" to insert properties. Do you know if/why we need this block as well as the lower one? (Note that this block contains META-INF/** which isn't in the lower one.)

ejrgilbert: I'm unsure what the logic is behind it. All I know is that this is the structure that "works." If we want to investigate if they're both necessary, a new ticket can be opened to address it.

hlgp (approved):

hlgp (approved):

alerman (approved):
Looks good

PR #1414, opened 2022-02-03
Set certs in props

alerman:
Set certificate paths and passwords in  properties

### DIFF
--- a/web-services/deploy/application/src/main/wildfly/add-datawave-configuration.cli
+++ b/web-services/deploy/application/src/main/wildfly/add-datawave-configuration.cli
@@ -94,10 +94,10 @@ module add --name=com.mysql.driver --dependencies=javax.api,javax.transaction.ap
 
 # Keystore and Truststore settings for export into MapReduce jobs run via the MapReduce API
 /system-property=dw.mapreduce.securitydomain.useJobCache:add(value=${mapreduce.securitydomain.useJobCache})
-/system-property=dw.mapreduce.securitydomain.keyStoreURL:add(value=file:///${KEYSTORE})
+/system-property=dw.mapreduce.securitydomain.keyStoreURL:add(value=file://${KEYSTORE})
 /system-property=dw.mapreduce.securitydomain.keyStoreType:add(value="${KEYSTORE_TYPE}")
 /system-property=dw.mapreduce.securitydomain.keyStorePassword:add(value="${KEYSTORE_PASSWORD}")
-/system-property=dw.mapreduce.securitydomain.trustStoreURL:add(value=file:///${TRUSTSTORE})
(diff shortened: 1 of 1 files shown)


web-services/deploy/application/src/main/wildfly/add-datawave-configuration.cli
```diff
@@ -94,12 +94,12 @@ module add --name=com.mysql.driver --dependencies=javax.api,javax.transaction.ap
 
 # Keystore and Truststore settings for export into MapReduce jobs run via the MapReduce API
 /system-property=dw.mapreduce.securitydomain.useJobCache:add(value=${mapreduce.securitydomain.useJobCache})
-/system-property=dw.mapreduce.securitydomain.keyStoreURL:add(value=file:/${jboss.server.config.dir}/certificates/testServer.p12)
-/system-property=dw.mapreduce.securitydomain.keyStoreType:add(value="PKCS12")
-/system-property=dw.mapreduce.securitydomain.keyStorePassword:add(value=ChangeIt)
-/system-property=dw.mapreduce.securitydomain.trustStoreURL:add(value=file:/${jboss.server.config.dir}/certificates/ca.jks)
-/system-property=dw.mapreduce.securitydomain.trustStoreType:add(value="JKS")
-/system-property=dw.mapreduce.securitydomain.trustStorePassword:add(value=ChangeIt)
+/system-property=dw.mapreduce.securitydomain.keyStoreURL:add(value=file:///${KEYSTORE})
```
keith-ratcliffe: First of all, thanks for doing this. Swapping out certs will be much less of a pain now

On L97 and L100, don't we want `file://` rather than `file:///` ? Otherwise the 1st slash on the interpolated KEYSTORE/TRUSTSTORE gives us `file:////` in the end. Could be that URL resolver does the right thing in the case of too many/too few, so it might not really matter

Either way, would be good to be consistent with L150 below

keith-ratcliffe (dismissed):

web-services/deploy/application/src/main/wildfly/add-datawave-configuration.cli
keith-ratcliffe: Or maybe there's something peculiar about the property loader here that I'm not seeing(?)

alerman: Hmm -- it worked, so it must handle it, but I agree we should be consistent. Ill take it out and double check it still works.

matthpeterson (dismissed):

apmoriarty (dismissed):

apmoriarty (approved):

matthpeterson (approved):

Issue #1413, opened 2022-02-03
Bounded query range total in DQP are summarized in total timing


phrocker:
The following line is summarized with the total timing throwing off the total time. 

stopwatch = timers.newStartedStopwatch("DefaultQueryPlanner - Expand bounded query ranges (total)");

I propose changing it to 

stopwatch = timers.newNonSummarizedStopwatch("DefaultQueryPlanner - Expand bounded query ranges (total)");

to clearly identify that this should not be summarized in the final, tabulated, total elapsed timing.

phrocker:
I have the code change ready for this so I can take work item. 

https://github.com/NationalSecurityAgency/datawave/compare/integration...phrocker:Issue1413?expand=1

Will add a unit test for QueryStopWatch before submitting the PR.

PR #1415, opened 2022-02-04
Create a non summarized stop watch that is not tabulated into the tot…

phrocker:
…al elapsed time

Resolves issues #1413


warehouse/query-core/src/main/java/datawave/query/util/QueryStopwatch.java
```diff
@@ -45,13 +45,47 @@ private TraceStopwatch newStopwatch(String header) {
...
+    /**
+     * Creates a new stopwatch that will be tabulated into the total time during summarization iff nonSummarized is false
+     * 
+     * @param header
+     *            header header string to print during summarization
+     * @param nonSummarized
+     *            boolean to determine if this stopwatch's time is tabulated during summarization
+     * @return newly created instance
+     */
+    public TraceStopwatch newStartedStopwatch(String header, boolean nonSummarized) {
```
phrocker, line 72: I thought about adding a simple modifier to TraceStopWatch that made it "non summarized" so I'd be open to doing that if it is preferred. 

ala TraceStopwatch.makeNonSummarized()  if that is preferred (i.e. encapsulate the modifier versus using the class type to differentiate ). 

Doing so could simplify this code change and remove the inner class -- open to feed back.

keith-ratcliffe (dismissed):
The current implementation looks fine to me, although I'll defer to others, if there's a strong preference for the alternate approach that you mentioned

matthpeterson (dismissed):

keith-ratcliffe (dismissed):

hlgp:
Already incorporated.

PR #1416, opened 2022-02-08
Swap Log4j12 for Reload4j

alerman:
This PR swaps out log4j12 (no longer updated) for reload4j. Reload4j is a drop in replacement that fixes several security issues. This PR is blocked waiting on Accumulo 1.10.2. Without that, there are 2 unit tests which fail as an option needs to be added to zookeeper on startup of the mini accumulo cluster.


keith-ratcliffe (changes requested):
Assuming we keep the bash variable names the same, the `*/bin/ingest/findJars.sh` needs the following update at minimum:
```
LOG4J_JAR=$(findJar reload4j)
SLF4J_BRIDGE_JAR=$(findJar slf4j-reload4j)
```
There may be other scripts that are doing the same thing

keith-ratcliffe:
FWIW, last week I tried out the experiment of building a quickstart from the integration branch, then replaced all `log4j-1*` jars and all `slf4j-log4j12*` jars with the reload4j versions after the fact and across the board (i.e., in DW, Accumulo 1.9.3, Apache Hadoop 3.1.2, and ZK 3.4.5). 

The only showstoppers I ran into were certain scripts (our `findJars.sh` as noted, and Accumulo's `bin/accumulo` script) that do an fs lookup to find a log4j jar to put on the classpath.

After fixing that, I ran it under load for almost 20hrs with no apparent ill-effects

P.S., I tested with reload4j `v1.2.18.3` and slf4j-reload4j `v1.7.35`

alerman:
@keith-ratcliffe I updated the requested DWV scripts and ensured checks are passing now. Should I also update Accumulo to 1.10.2 for quickstart in this PR, or leave that for later?

keith-ratcliffe (approved):

ivakegg (approved):

drewfarris:
Any idea what `log4j-extras` is used for?

PR #1423, opened 2022-02-11
Push content functions into each component piece of an index query

apmoriarty:


FineAndDandy (approved):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionsDescriptor.java
```diff
@@ -387,6 +397,169 @@ public void addFilters(AttributeFactory attributeFactory, Map<String,EventDataQu
...
+            if (deref instanceof ASTAndNode) {
+                // single component
+                components.add(deref);
+            } else if (deref instanceof ASTOrNode) {
+                // multiple components
+                for (JexlNode child : JexlNodes.children(deref)) {
+                    components.add(JexlASTHelper.dereference(child));
+                }
+            } else {
+                throw new IllegalStateException("Expected a dereferenced IndexQuery node to be an AND or OR node, but was " + deref.getClass().getSimpleName());
```
alerman, line 436: I feel that this should be "to be an ASTAndNode or ASTOrNode" -- since the classname is what will print.

alerman (approved):

hlgp (approved):

PR #1428, opened 2022-02-16
Fixes #1426 Updated geowave pruning visitor to replace pruned nodes with false no…

### LINKED ISSUE #1426: GeoWavePruningVisitor should replace pruned nodes with a false node
Previously, assumptions had been made as to how the query tree would be structured at the time of geowave pruning.  It was expected that the query would consist of a geowave function anded with a large OR node of geowave terms and ranges.  However, in some cases the executable expansion visitor, which runs between geowave function index expansion and geowave pruning, may alter the structure of the tree, such that each ORd term/range becomes anded with additional terms/functions/ranges.  In this case, when we run geowave pruning we remove the geowave component from some of these terms, but not the whole term, which is wrong.  Instead, the safer option would be to replace the term/range with a false node to ensure that the query is still valid.

jwomeara:
…des.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/GeoWavePruningVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/GeoWavePruningVisitor.java
@@ -114,7 +114,7 @@ public class GeoWavePruningVisitor extends RebuildingVisitor {
                         if (prunedTerms != null) {
                             prunedTerms.put(field, value);
                         }
-                        return null;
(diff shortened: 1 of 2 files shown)


apmoriarty (approved):

FineAndDandy (approved):

matthpeterson (approved):
Looks good to me.  I wonder if it's worthwhile and reasonably straightforward to add another test case that demonstrates the issue described in 1426 (now corrected).

avgAGB (approved):

PR #1429, opened 2022-02-16
Fixes #1427 Moved executable expansion visitor to run after index exp…

### LINKED ISSUE #1427: ExecutableExpansionVisitor should be run after bounded range & regex expansion
Currently, we run an executability check on the query tree before any bounded ranges or regex terms have had a chance to be expanded, and if the query is not executable at that point, we attempt to distribute terms until it is.  Instead of doing this after function index expansion, we should instead do this after all of the ranges and regex terms have had a chance to be expanded.  If they are all expanded (and they are against indexed terms) then there would be no need to distribute terms.

jwomeara:
…ansion.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownLargeFieldedListsVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownLargeFieldedListsVisitor.java
@@ -296,7 +296,12 @@ public class PushdownLargeFieldedListsVisitor extends RebuildingVisitor {
                     List<JexlNode> otherNodes) {
         QueryPropertyMarker.Instance instance = QueryPropertyMarker.findInstance(subNode);
         if (subNode instanceof ASTEQNode) {
-            eqNodes.put(JexlASTHelper.getIdentifier(subNode, false), origNode);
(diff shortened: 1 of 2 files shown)


FineAndDandy (approved):

apmoriarty (approved):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownLargeFieldedListsVisitor.java
```diff
@@ -296,7 +296,12 @@ protected void assignNodeByField(JexlNode origNode, JexlNode subNode, Multimap<S
                     List<JexlNode> otherNodes) {
         QueryPropertyMarker.Instance instance = QueryPropertyMarker.findInstance(subNode);
         if (subNode instanceof ASTEQNode) {
-            eqNodes.put(JexlASTHelper.getIdentifier(subNode, false), origNode);
+            String identifier = JexlASTHelper.getIdentifier(subNode, false);
+            if (identifier != null) {
+                eqNodes.put(JexlASTHelper.getIdentifier(subNode, false), origNode);
```
ivakegg, line 301: Why not reuse the 'identifier' variable here?

ivakegg (approved):

PR #1432, opened 2022-02-18
Adding documentation for unique_by_x visitor functions

avgAGB:


web-services/deploy/docs/docs/query_help.html
```diff
@@ -209,5 +209,17 @@ <h3>LUCENE Query Syntax</h3>
 		</ul>
 		<li>A special date format of 'e' can be supplied to mean milliseconds since epoch.</li>
 	</ol>
+	<h3>Unique Functions</h3>
+	<p class="left">In addition, there are functions that are available to visit the query tree, and perform operations on the nodes and their
```
jschmidt10: I may be off base but here's what I'm thinking:

This document's audience will be end users. There is a "query lingo" we can assume knowledge of - like "index-only" fields or "events". But in this paragraph you're talking about the query parse tree and visitors - that's going to be out of their realm. This belongs more in a javadoc I think.

web-services/deploy/docs/docs/query_help.html
```diff
@@ -209,5 +209,17 @@ <h3>LUCENE Query Syntax</h3>
...
+		children. The extracted parameters will be added to the provided data. Any options function nodes will be subsequently deleted and remaining AND/OR nodes will be optimized to account for the potential removal of their children.
+		Potential function options that the user may use are listed below.</p>
+	<ul>
+		<li>f:options(): Expects a comma-delimited list of key/value pairs, e.g. f:options('hit.list','true','limit.fields','FOO_1_BAR=3)</li>
```
jschmidt10: This part I think is more appropriate. I like the example usages. You might want to add a follow on paragraph with an example dataset or something for how "unique_by_x" impacts the results.

For instance, let's say we have

```
event1, 01/01/2022, icecream=vanilla
event2, 01/01/2022, icecream=vanilla
event3, 01/01/2022, icecream=chocolate

event4, 01/02/2022, icecream=vanilla
```
Unique_by_day, I'm assuming could either yield (event1, event3, event4) OR (event1, event2, event4). You could also add minutes into that example.

web-services/deploy/docs/docs/query_help.html
avgAGB: Makes sense. So maybe replace the excerpt there with "Checks the fields to be unique over the interval of a day/minute/second", or something along those lines?

jschmidt10: Yeah, the wording is a bit tough. Might be best to snag a query SME for the phrasing but I'd guess something like "Returns one result per unique value in the given timeframe" or something like that?

web-services/deploy/docs/docs/query_help.html
avgAGB: Thanks. I'll look to build off of this and put in some more concrete examples.

web-services/deploy/docs/docs/query_help.html
avgAGB: I've updated the description in the latest commit, and provided a few more contextual examples.

jschmidt10 (approved):

apmoriarty (approved):

PR #1433, opened 2022-02-18
Persist toKeep flag when serializing most attributes

apmoriarty:


warehouse/query-core/src/main/java/datawave/query/attributes/GeoPoint.java
```diff
@@ -66,16 +66,16 @@ public void write(DataOutput out) throws IOException {
...
+        WritableUtils.writeVInt(out, toKeep ? 1 : 0);
     }
     
     @Override
     public void readFields(DataInput in) throws IOException {
         readMetadata(in);
-        
         this.point = WritableUtils.readString(in);
         validate();
+        this.toKeep = WritableUtils.readVInt(in) != 0;
```
FineAndDandy, line 77: should validate() be after this?

warehouse/query-core/src/main/java/datawave/query/attributes/IpAddress.java
```diff
@@ -78,6 +78,7 @@ public void readFields(DataInput in) throws IOException {
         setValue(ipString);
         setNormalizedValue(ipString);
         validate();
+        this.toKeep = WritableUtils.readVInt(in) != 0;
```
FineAndDandy, line 80: move this in front of validate?

warehouse/query-core/src/main/java/datawave/query/attributes/IpAddress.java
```diff
@@ -138,6 +139,7 @@ public void read(Kryo kryo, Input input) {
         setValue(ipAddressString);
         setNormalizedValue(ipAddressString);
         validate();
+        this.toKeep = input.readBoolean();
```
FineAndDandy, line 141: validate?

FineAndDandy (dismissed):
I like the tests. Would prefer to see validate() called after reading the new fields

warehouse/query-core/src/main/java/datawave/query/attributes/GeoPoint.java
apmoriarty: Seems like `validate()` is called after the payload/content is read, it has no bearing on the toKeep flag.

jschmidt10: I know what you mean - the toKeep flag is just a boolean and requires no validation. But from a high level, we want to deserialize the entire state of the object, then validate it. If we establish that pattern in the code now, it's less likely someone might accidentally violate it with another new field in the future.

warehouse/query-core/src/test/java/datawave/query/attributes/AttributeTest.java
```diff
@@ -0,0 +1,56 @@
...
+            byte[] data = baos.toByteArray();
+            
+            Document next = new Document();
+            next.readFields(new DataInputStream(new ByteArrayInputStream(data)));
+            
+            Attribute<?> nextAttr = next.get("KEY");
+            assertEquals(expected, nextAttr.isToKeep());
+            
+        } catch (Exception e) {
+            fail("Test failed with exception: " + e.getMessage());
```
jschmidt10, line 54: This will swallow the stacktrace. It might be worth just throwing and letting junit print the stacktrace.

warehouse/query-core/src/test/java/datawave/query/attributes/AttributeTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.DataInputStream;
+import java.io.DataOutput;
+import java.io.DataOutputStream;
+
+import static org.junit.Assert.*;
+
+public class AttributeTest {
```
jschmidt10, line 15: Slick tests!

warehouse/query-core/src/main/java/datawave/query/attributes/GeoPoint.java
apmoriarty: Fair point, I will move `validate()` to the end and make that the standard

FineAndDandy (approved):

alerman (approved):

PR #1437, opened 2022-02-23
Issue1435 - fixes and unit tests for SequentialScheduler

tomnelson:
Modifications (by phrocker) to SequentialScheduler. Unit tests from query-code duplicated and changed to use SequentialScheduler to assure that the results are the same with either Scheduler


warehouse/query-core/src/main/java/datawave/query/scheduler/SequentialScheduler.java
```diff
@@ -129,9 +134,25 @@ public boolean hasNext() {
                     if (null != qd.getRanges())
                         rangesSeen += qd.getRanges().size();
                     count.incrementAndGet();
-                    if (null == newQueryData)
+                    if (null == newQueryData) {
                         newQueryData = new QueryData(qd);
-                    else {
+                        String transformedQuery = JexlStringBuildingVisitor.buildQuery(config.getQueryTree());
```
ivakegg, line 139: Why are we doing this?  The QueryData object should have the query that is appropriate to run for the referenced range.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/scheduler/SequentialScheduler.java
phrocker: @ivakegg You are correct that the QueryData object has the correct range for the referenced range; however, the SequentialScheduler runs ranges with a vanilla Accumulo Batch Scanner: one IteratorSetting per group of ranges. The provided QueryData object can have pruned queries, meaning that any given QueryData range may have an insufficient query for all ranges since RangeStream's process reduces the query.

One way to solve the problem was to take the query as it entered RangeStream (  expanded, flattened, and validated ), but there are other possible solutions, such as combining queries from all QueryData objects, and disabling RangeStream functions when a different scheduler is used. Do those seem more palatable?

apmoriarty: Fundamentally we're pitting the time savings of the sequential scheduler against the time savings of pruning the query tree in the range stream. Has either cost been quantified? At what point would it make sense to send a query down either path?

I suspect that for a subset of sufficiently simple queries the sequential scheduler will indeed be faster, but there still exists a large number of queries where pruning via range stream is still desirable. 

Combining all queries from all QueryData objects sounds attractive, but difficult to get right the first time.

tomnelson: If the question becomes: "Should the SequentialScheduler exist?" then I would say "Maybe not, but if it does, it should return the same results as the PushdownScheduler for any given query"
I found no unit tests here for SequentialScheduler so I moved a pile of them over and got many failures prior to phrocker's changes. 
His changes are a minimally intrusive way to address that.
You may also want to consider whether running all these additional tests is worth the build time. A subset of them would probably suffice.

phrocker: I have another PR that I'm prepping for RangeStream ( currently resolving a few unit tests) that may help on the performance front, but this PR is simply focusing on the bug fix, which is that we can miss results if the sequential scheduler is used. 

You are correct @apmoriarty , this was tested as a means of comparing and contrasting schedulers and trying to determine when and where either might be more beneficial. In that testing missing results were identified. The choice of using the pre-RangeStream query is out of simplicity since it already exists within the query planner. Tom added some tests as you can see, which I think is a huge help.

keith-ratcliffe: I can confirm that both sequential and pushdown get the same number of hits for the previously affected queries on our test system after applying this patch. 
@ivakegg , @apmoriarty , any reservations with moving forward with this as-is?

apmoriarty: Nope, let's get this in!

apmoriarty (dismissed):
There are some star imports but that won't hold up an approval for this one

keith-ratcliffe (dismissed):

PR #1439, opened 2022-02-24
Fix #1438 - optionally gather fields after ingest exception

### LINKED ISSUE #1438: Generate minimal subset of fields when EventMapper.getFields throws Exception
Summary:
Exceptions in the IngestHelper's getFields method cause ingest jobs to produce error table entries with very few fields, making troubleshooting difficult compared to records that arrive marked with a fatal error.  Adjust the ingest code to minimize this difference for improved consistency in which fields exist in the error tables.

Background:
In the existing baseline, there are primarily two code paths for errors output by ingest jobs:
1. the incoming rawRecordContainer is already flagged as having a fatal error.
2. the incoming rawRecordContainer is not flagged with a fatal error, but experiences an error in the mapper.

Both code paths put the record into the error tables, but the fields that are written vary greatly.

That's because in path 1, when an fatal error flag is noticed, the handlers are all cleared and only the error handler is added.  Then EventMapper.processEvent is called with the error handler.  This results in a call to getFields which gathers fields (using IngestHelper) as well as supplemental fields including the LOAD_DATE, ORIG_FILE, virtual fields, and more.

Conversely, in path 2 the call to EventMapper.processEvent is made with all of the handlers but throws an exception, potentially in getFields.  This means that the fields are not gathered as they were in path 1.  The exception is caught, the error handler provided to executeHandler (which doesn't attempt to gather fields using IngestHelper).

The differences in the error entries between these two code paths makes error investigation, including queries, more difficult.

Desired behavior for path 2:
- Some of the additional fields, missing from path 2's error table entries, are helpful such as the ORIG_FILE and LOAD_DATE.  Ensure they are added even if there's an error in getFields.
- When the call to IngestHelper's getFields fails, optionally make another call to gather a minimal subset of fields.
- If that minimal subset of fields is configured to generate a virtual field, ensure it is created and included in the error tables.
- If the gathering of the minimal subset of fields causes an error, this should be tolerated and not fail the job.

matthpeterson:


matthpeterson:
Multiple test cases were conducted using the quickstart json data:
1. EVENT_DATE_MISSING added in JsonReader to every other record.
2. getFields in JsonIngestHelper was made to throw an exception.  JsonIngestHelper was also modified to implement SalvagedFields with a hardcoded multimap of mocked salvaged fields.  
Note both of the two test cases were done with throw-away code changes, made only for testing this PR.
3. Regression case (no modifications, no errors)

Case 1 operated as it does today, including all the normally-ingested fields and the supplemental fields in the error table writes.  The other half of the data was written to the shard table, as expected.

Case 2 operated with the new behavior: adding LOAD_DATE, ORIG_FILE, and salvaged fields to the error tables along with what it previously written.

Case 3 operated as expected, ingesting the data into the shard table.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/EventMapper.java
```diff
@@ -546,10 +484,60 @@ public void map(K1 key, V1 value, Context context) throws IOException, Interrupt
             context.progress();
         }
         
+        incrementEventCount(value, context);
```
jschmidt10, line 558: I dig the method breakout. This function used to be huge.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/EventMapper.java
```diff
@@ -546,10 +484,60 @@ public void map(K1 key, V1 value, Context context) throws IOException, Interrupt
...
+        
+        updateMetrics(value, eventMapperTimer, fields);
+    }
+    
+    private void failJobOnConstraintViolations(Exception e) {
+        if (e instanceof ConstraintChecker.ConstraintViolationException) {
+            throw ((RuntimeException) e);
```
jschmidt10, line 565: Not your code but can this cast be dropped? Doesn't seem like it's doing anything.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/FieldHarvester.java
```diff
@@ -0,0 +1,271 @@
...
+    private Exception originalException;
+    
+    public FieldHarvester(Configuration configuration) {
+        this.createSequenceFileName = configuration.getBoolean(LOAD_SEQUENCE_FILE_NAME, true);
+        this.trimSequenceFileName = configuration.getBoolean(TRIM_SEQUENCE_FILE_NAME, true);
+        this.createRawFileName = configuration.getBoolean(LOAD_RAW_FILE_NAME, true);
+    }
+    
+    /**
+     * Updates "fields" with extracted, derived, and automatically generated fields. Will capture
```
jschmidt10: Incomplete sentence.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/EventMapper.java
matthpeterson: Thanks!

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/EventMapper.java
matthpeterson: I think Exception is considered a checked Exception while RuntimeException is considered unchecked, so the method signature wouldn't require a "throws RuntimeException" while without the cast a "throws Exception" would be needed.  The current signature of map, as defined in org.apache.hadoop.mapreduce.Mapper is 
```
public void map(K1 key, V1 value, Context context) throws IOException, Interrupt
```
I think it makes more sense to cast it to ConstraintViolationException.  What do you think?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/FieldHarvester.java
```diff
@@ -0,0 +1,271 @@
...
+        
+        try {
+            // try adding supplemental fields to candidateFields, whether or not they were salvaged
+            addSupplementalFields(value, offset, splitStart, ingestHelper, candidateFields);
+        } catch (Exception exception) {
+            if (null == this.originalException) {
+                this.originalException = exception;
+            } else {
+                // preserve original exception and log the latest exception
+                log.error(exception);
```
jschmidt10: Might be helpful here to add a logline that this exception occurred while adding supplemental fields. It'll just add a little bit of context maybe? Instead of just a stacktrace popping up in the log.

```
logger.error("Encountered another exception while adding supplemental fields", exception);
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/FieldHarvester.java
```diff
@@ -0,0 +1,271 @@
...
+                fieldError = entry.getValue().getError();
+            }
+            fields.put(entry.getKey(), entry.getValue());
+        }
+        if (null != fieldError) {
+            if (null == this.originalException) {
+                this.originalException = new FieldNormalizationError("Failed getting all fields", fieldError);
+            } else {
+                // preserve original exception
+                log.error(originalException);
```
matthpeterson: @dtspence thanks for pointing out that this should be logging fieldError, not originalException.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/EventMapper.java
jschmidt10: Ahh, we're catching it as a vanilla Exception above, I missed that - gotcha. Nope, this is fine then.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/FieldHarvester.java
```diff
@@ -0,0 +1,271 @@
...
+    
+    /**
+     * If IngestHelper implements FieldSalvager, get the salvageable fields from value. Otherwise, return an empty Multimap.
+     */
+    private Multimap<String,NormalizedContentInterface> attemptToSalvageFields(RawRecordContainer value, IngestHelperInterface ingestHelper) {
+        // If this helper is able, attempt to salvage a subset of the fields
+        if (null != ingestHelper && ingestHelper instanceof FieldSalvager) {
+            FieldSalvager salvager = (FieldSalvager) ingestHelper;
+            try {
+                Multimap<String,NormalizedContentInterface> salvagedFields = salvager.getSalvageableEventFields(value);
```
ivakegg: I am wondering whether a better approach may be to pass the multimap into the IngestHelper.getFields(value) method such that if an exception is thrown, then the map will already contain the fields that could be "salvaged".  That way we are forced to allow allow ingest helpers to "salvage fields" and not have to implement a separate interface.

matthpeterson: I'll make the suggested change.  Thanks for the review.

ivakegg (dismissed):

dtspence (dismissed):

hlgp (dismissed):

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/EventMapperSalvageFieldsOnErrorTest.java
```diff
@@ -0,0 +1,243 @@
...
+import datawave.ingest.data.config.BaseNormalizedContent;
+import datawave.ingest.data.config.NormalizedContentInterface;
+import datawave.ingest.data.config.NormalizedFieldAndValue;
+import datawave.ingest.data.config.ingest.IngestHelperInterface;
+import datawave.ingest.data.config.ingest.MinimalistIngestHelperImpl;
+import datawave.ingest.mapreduce.job.BulkIngestKey;
+import datawave.ingest.mapreduce.job.writer.ContextWriter;
+
+/**
+ * This test class aims to test only the scenarios in which EventMapper encounters an error, to verify its behavior with the FieldSalvager interface. It's
```
hlgp, line 37: Seeing references to a FieldSalvager but not finding that interface.  Old name for the Harvester?  Am I just missing it?

ivakegg:
Added branch name to ticket if anybody decides to pick this up again.

PR #1442, opened 2022-02-25
Datawave 1400

cmitchell:
Queries marked as "allowLongRunningQuery" can exceed the 55 minute  timeout to return results. This ticket is explicitely for groupBy queries, but it was implemented in a way that can be extending to other types of queries as well. Key changes:

  - If a query that's allowed to be long running exceeds 50 minutes to return (so there's no risk of hitting the 55 minute timeout) , an empty ResultsPage is returned with a status of "PARTIAL", and the query continues to execute on the tserver.
  - When the query does complete, all results are returned at once with a status of "COMPLETE" .
  - The call to getTransformer() in the query logic now returns a singleton instance so that transformers in the list can be updated with the query start time necessary for the logic of this ticket 
  - The client side and tserver side of the grouping transform were broken out into two separate instances for clarity and maintainability (GroupingTransform and GroupingIterator respectively)
  - LongRunningQueryTest tests this logic


properties/default.properties
```diff
@@ -202,6 +202,10 @@ disable.index.only.documents=false
 # Indicates whether index-only filter functions should be enabled, such as filter:includeRegex()
 enable.index.only.filter.functions=false
 query.tld.collapse.uids=false
+# Time after which a blank page is returned to indicate to the client that the query is still running (in milliseconds)
+query.execution.for.page.timeout=3000000
+# Whehter or not the query is allowed to exceed the overall query execution time out. If true, the query.execution.for.page.timeout should also be set
```
hlgp: Whether

warehouse/query-core/src/main/java/datawave/query/common/grouping/GroupingUtil.java
```diff
@@ -0,0 +1,316 @@
...
+import static org.slf4j.LoggerFactory.getLogger;
+
+/**
+ * Provides functionality commonly needed to group documents (regardless if done server or client side).
+ *
+ * This class and its methods aren't static so that we don't run into concurrency issues, although all required state should be passed into the individual
+ * methods and not kept in this class. Calling classes could extend this class to inherit the methods, but the state still shouldn't be inherited because not
+ * all callers will be able to easily extend this class if they already/need to extend other parents.
+ */
+public class GroupingUtil {
```
hlgp, line 37: Nice!  I like this refactor a lot.

warehouse/query-core/src/main/java/datawave/query/common/grouping/GroupingUtil.java
```diff
@@ -0,0 +1,316 @@
+package datawave.query.common.grouping;
+
+import com.google.common.collect.*;
```
hlgp: No star imports, please

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -1936,4 +1967,20 @@ public void setYieldThresholdMs(long yieldThresholdMs) {
...
     
+    public void setQueryExecutionTimeout(long queryExecutionTimeout) {
+        this.queryExecutionTimeout = queryExecutionTimeout;
+    }
+    
+    public long getQueryExecutionTimeout() {
+        return queryExecutionTimeout;
+    }
+    
+    public void setAallowLongRunningQuery(boolean allowLongRunningQuery) {
```
hlgp: Nitpick, tiny typo

warehouse/query-core/src/main/java/datawave/query/transformer/ContentQueryTransformer.java
```diff
@@ -16,16 +12,21 @@
...
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map.Entry;
+
 public class ContentQueryTransformer extends BaseQueryLogicTransformer<Entry<Key,Value>,DefaultEvent> {
     
     private Authorizations auths = null;
     private Logger log = Logger.getLogger(ContentQueryTransformer.class);
+    // long queryExecutionForCurrentPageStartTime;
```
hlgp: no longer needed?

warehouse/query-core/src/main/java/datawave/query/tables/ScannerSession.java
```diff
@@ -28,12 +20,9 @@
 import org.apache.commons.lang.builder.EqualsBuilder;
 import org.apache.log4j.Logger;
 
-import com.google.common.base.Preconditions;
-import com.google.common.collect.Iterables;
-import com.google.common.collect.Lists;
-import com.google.common.collect.Queues;
-import com.google.common.util.concurrent.AbstractExecutionThreadService;
-import com.google.common.util.concurrent.MoreExecutors;
+import java.util.*;
```
hlgp: More star imports, probably a setting in your IDE

web-services/deploy/configuration/src/main/resources/datawave/query/QueryLogicFactory.xml
```diff
@@ -133,6 +133,7 @@
         <property name="indexTableName" value="${index.table.name}" />
         <property name="reverseIndexTableName" value="${rindex.table.name}" />
         <property name="maxResults" value="${event.query.max.results}" />
+        <property name="executionTimeout" value="${query.execution.for.page.timeout}"/>
```
hlgp: The name of the property has to match what's in the class for spring to do its autowiring thing, so this should be queryExecutionForPageTimeout

warehouse/query-core/src/test/java/datawave/query/LongRunningQueryTest.java
```diff
@@ -0,0 +1,140 @@
...
+        query.setEndDate(endDate);
+        query.setQueryAuthorizations(auths.serialize());
+        query.setColumnVisibility("A&E&I");
+        query.setPagesize(Integer.MAX_VALUE);
+        query.setParameters(extraParameters);
+        query.setId(UUID.randomUUID());
+        
+        // this parameter is what makes the query long running. Failing to set this will let it default to 50 minutes
+        // (and not the 1000 milliseconds that it is set to) which will return only 1 page of 8 results, thereby failing this test.
+        logic.setQueryExecutionForPageTimeout(1000);
```
hlgp: I'm probably missing something.  Isn't there a chance this can execute so quickly that we'll still get results in the first page?

warehouse/query-core/src/test/java/datawave/query/LongRunningQueryTest.java
```diff
@@ -0,0 +1,140 @@
...
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.util.*;
+
+import static org.junit.Assert.assertEquals;
+
+public class LongRunningQueryTest {
```
hlgp, line 42: Good stuff!  How about a few more tests:
1. query for something legitimately nonexistent and make sure it returns the expected status
2. set the pageSize to something less than the number of results and make sure nothing goes awry there

properties/default.properties
cmitchell: Fixed

warehouse/query-core/src/main/java/datawave/query/common/grouping/GroupingUtil.java
cmitchell: Fixed and changed the setting in intellij so it won't happen again.

warehouse/query-core/src/main/java/datawave/query/transformer/ContentQueryTransformer.java
cmitchell: Removed

warehouse/query-core/src/main/java/datawave/query/tables/ScannerSession.java
cmitchell: Fixed and changed the setting in intellij so it won't happen again.

web-services/deploy/configuration/src/main/resources/datawave/query/QueryLogicFactory.xml
cmitchell: Fixed. Thanks for catching this. I decided to rename this to be more descriptive later in development and forgot to update the xml configuration.

warehouse/query-core/src/test/java/datawave/query/LongRunningQueryTest.java
cmitchell: Ha, yes, you're correct. In fact, that's the reason this test passes locally, but fails on github. I will put in a forced sleep or otherwise guarantee that the timeout is hit.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
cmitchell: thanks, fixed

warehouse/query-core/src/test/java/datawave/query/LongRunningQueryTest.java
cmitchell: Added `testAllowLongRunningQueryWithSmallPageSize()` and `testAllowLongRunningQueryOnNoResultsc()`

warehouse/query-core/src/main/java/datawave/query/iterator/GroupingIterator.java
```diff
@@ -0,0 +1,240 @@
...
+import datawave.query.common.grouping.GroupingUtil;
+import datawave.query.jexl.JexlASTHelper;
+import org.apache.accumulo.core.data.Key;
+import org.apache.accumulo.core.iterators.YieldCallback;
+import org.apache.accumulo.core.security.ColumnVisibility;
+import org.slf4j.Logger;
+import org.springframework.util.Assert;
+
+import java.math.BigDecimal;
+import java.util.*;
```
alerman: Missed one start import

warehouse/query-core/src/main/java/datawave/query/iterator/GroupingIterator.java
```diff
@@ -0,0 +1,240 @@
...
+        this.countingMap = new GroupCountingHashMap(this.markingFunctions);
+    }
+    
+    @Override
+    public boolean hasNext() {
+        for (int i = 0; i < groupFieldsBatchSize; i++) {
+            if (previousIterators.hasNext()) {
+                Map.Entry<Key,Document> entry = previousIterators.next();
+                if (entry != null) {
+                    log.trace("{} get list key counts for: {}", "t-server", entry);
```
alerman: Why use substitution for first hardcoded value?

alerman (changes requested):
Couple of start imports still need fixed.

Also, This may be my naivete, but does there need to be an overall maximum? For instance, if the user has called next 10 times and still gotten no results, should we allow that query to keep running?

warehouse/query-core/src/main/java/datawave/query/iterator/GroupingIterator.java
cmitchell: Fixed

warehouse/query-core/src/main/java/datawave/query/iterator/GroupingIterator.java
cmitchell: Didn't mean to. This was a copy/paste and then modified from the code before it was separated into GroupingIterator and GroupingTransform (it used to be a ternary statement to substitute either t-server or webclient) -- changed to use "t-server get list...." instead of substitution.

alerman (dismissed):

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -394,10 +394,10 @@ private DateIndexHelper getDateIndexHelper(final ShardQueryConfiguration config)
         // lets mark the query as started (used by ivarators at a minimum)
         try {
             markQueryStarted(config, settings);
+            settings.setQueryExecutionForCurrentPageStartTime(System.currentTimeMillis());
```
jwomeara: Does it make sense to move this into the previous method call?

warehouse/query-core/src/main/java/datawave/query/transformer/DocumentTransformerSupport.java
```diff
@@ -497,13 +499,26 @@ public void setEventQueryDataDecoratorTransformer(EventQueryDataDecoratorTransfo
     }
     
     /**
-     * Add a document transformer
+     * ln Add a document transformer. If the type of document transformer is already in the list of transformers, that instance is replaced. This works under
```
jwomeara: ln?

warehouse/query-core/src/main/java/datawave/query/transformer/DocumentTransformerSupport.java
```diff
@@ -497,13 +499,26 @@ public void setEventQueryDataDecoratorTransformer(EventQueryDataDecoratorTransfo
...
+     * hte assumption that there is only one instance (singleton) of Transformer per query logic
```
jwomeara: typo: hte

warehouse/query-core/src/test/java/datawave/query/transformer/GroupingTestWithModel.java
```diff
@@ -169,6 +170,8 @@ public void setup() {
         
         logic.setFullTableScanEnabled(true);
         logic.setMaxEvaluationPipelines(1);
+        // logic.setQueryExecutionForPageTimeout(1000);
```
jwomeara: Remove commented line if not needed

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
```diff
@@ -261,7 +265,6 @@ public ResultsPage next() throws Exception {
                 if (timing != null && currentPageCount > 0 && timing.shouldReturnPartialResults(currentPageCount, maxPageSize, pageTimeInCall)) {
                     log.info("Query logic max expire before page is full, returning existing results " + currentPageCount + " " + maxPageSize + " "
                                     + pageTimeInCall + " " + timing);
-                    hitPageTimeTrigger = true;
```
jwomeara: With this removed, hitPageTimeTrigger is always false now.  Is that expected?  If so, we can remove it from the condition below.

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
```diff
@@ -337,10 +348,14 @@ public ResultsPage next() throws Exception {
...
         }
-        if (resultList.isEmpty()) {
+        
+        if (hitIntermediateResult) {
+            return new ResultsPage(new ArrayList<>(), ResultsPage.Status.PARTIAL);
+        } else if (resultList.isEmpty()) {
             return new ResultsPage();
         } else {
-            return new ResultsPage(resultList, ((hitPageByteTrigger || hitPageTimeTrigger) ? ResultsPage.Status.PARTIAL : ResultsPage.Status.COMPLETE));
+            return new ResultsPage(resultList, ((hitPageByteTrigger || hitPageTimeTrigger || hitPageSizeTrigger) ? ResultsPage.Status.PARTIAL
```
jwomeara: hitPageTimeTrigger is used here but is always false.

web-services/query/src/test/java/datawave/webservice/query/configuration/TestBaseQueryLogic.java
```diff
@@ -58,6 +58,8 @@ public void testConstructor_Copy() throws Exception {
         expect(this.copy.getRoleManager()).andReturn(roleManager);
         expect(this.copy.getSelectorExtractor()).andReturn(null);
         expect(this.copy.getBypassAccumulo()).andReturn(false);
+        expect(this.copy.getQueryExecutionForPageTimeout()).andReturn(500000L);
+        expect(this.copy.isLongRunningQuery()).andReturn(false);
```
jwomeara: Oof, powermock!  Getting a glimpse of the skeletons in our closet.  :)

warehouse/query-core/src/test/java/datawave/query/LongRunningQueryTest.java
```diff
@@ -0,0 +1,306 @@
...
+        RunningQuery runningQuery = new RunningQuery(connector, AccumuloConnectionFactory.Priority.NORMAL, logic, query, "", datawavePrincipal,
+                        new QueryMetricFactoryImpl());
+        List<ResultsPage> pages = new ArrayList<>();
+        
+        ResultsPage page = runningQuery.next();
+        pages.add(page);
+        // guarantee the need for at least a second page. (make the wait slightly longer than the page timeout is set to)
+        try {
+            Thread.sleep(250);
+        } catch (InterruptedException e) {
```
jwomeara: The test throws Exception, so you don't need this try/catch block

warehouse/query-core/src/test/java/datawave/query/LongRunningQueryTest.java
```diff
@@ -0,0 +1,306 @@
...
+                        new QueryMetricFactoryImpl());
+        List<ResultsPage> pages = new ArrayList<>();
+        
+        ResultsPage page = runningQuery.next();
+        pages.add(page);
+        
+        // guarantee the need for at least a second page.
+        try {
+            Thread.sleep(250);
+        } catch (InterruptedException e) {
```
jwomeara: same comment as above.

warehouse/query-core/src/test/java/datawave/query/LongRunningQueryTest.java
```diff
@@ -0,0 +1,306 @@
...
+                        new QueryMetricFactoryImpl());
+        List<ResultsPage> pages = new ArrayList<>();
+        
+        ResultsPage page = runningQuery.next();
+        pages.add(page);
+        
+        // guarantee the need for at least a second page (make the wait slightly longer than the page timeout is set to)
+        try {
+            Thread.sleep(15);
+        } catch (InterruptedException e) {
```
jwomeara: same comment as above.

warehouse/query-core/src/test/java/datawave/query/LongRunningQueryTest.java
```diff
@@ -0,0 +1,306 @@
...
+        ResultsPage page = runningQuery.next();
+        pages.add(page);
+        // guarantee the need for at least a second page. (make the wait slightly longer than the page timeout is set to)
+        try {
+            Thread.sleep(250);
+        } catch (InterruptedException e) {
+            
+        }
+        
+        while (page.getStatus() != ResultsPage.Status.COMPLETE) {
```
jwomeara, line 128: Is there any chance that this could run infinitely?  Maybe fail after a few failed attempts?  Or you could incorporate the sleep here and put an overall time limit on the loop.  Or do none of those things if you're not worried about it.

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -454,6 +456,10 @@ public void initialize(ShardQueryConfiguration config, Connector connection, Que
         
         config.setQueryString(getQueryPlanner().getPlannedScript());
         
+        // Now that the query has been started, update the transformers used by the client with the start time that was set on the settings
+        // by the call to process()
+        getTransformer(settings).setQueryExecutionForPageStartTime(settings.getQueryExecutionForCurrentPageStartTime());
```
jwomeara: Should we just add the `setQueryExecutionForPageStartTime` call to the `getTransformer` method since the settings are passed?

warehouse/query-core/src/test/java/datawave/query/LongRunningQueryTest.java
```diff
@@ -0,0 +1,306 @@
...
+    /**
+     * Tests that having results returned over multiple pages (by setting the page size to be less than the number of results that will be returned) behaves
+     * even with long running queries exceeding their execution timeout for a page. Should get at least 1 page of 0 results and a PARTIAL status, but could be
+     * more depending on cpu speed due hitting the execution timeout. The last two pages should have 4 results each, but the last page should have a status of
+     * COMPLETE, and the next to last page should have a status of PARTIAL.
+     *
+     * @throws Exception
+     */
+    @Test
+    public void testAllowLongRunningQueryWithSmallPageSize() throws Exception {
```
jwomeara, line 224: When I run this test by itself in my IDE it hangs indefinitely.  It runs fine when running all of the tests in the suite though, which is odd.  Debugging it, it looks like it's getting hung up in the next page while loop.  Probably worth figuring out what's going on here.

jwomeara (changes requested):

warehouse/query-core/src/test/java/datawave/query/LongRunningQueryTest.java
cmitchell: Try again - a new version has been pushed

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
cmitchell: The `getTransformer` method is called by/at other times aside from the start of the query and we don't want to re-set/override the startTime.

warehouse/query-core/src/test/java/datawave/query/LongRunningQueryTest.java
cmitchell: Haha, I'm not worried about it. This is being done to account for different cpu speeds allowing an unknown number of pages to be returned. if this causes the test to run indefinitely, then the underlying code is messed up.

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
cmitchell: Added it back. Not sure why I removed it.

warehouse/query-core/src/test/java/datawave/query/transformer/GroupingTestWithModel.java
cmitchell: Removed

warehouse/query-core/src/main/java/datawave/query/transformer/DocumentTransformerSupport.java
cmitchell: Fixed

warehouse/query-core/src/main/java/datawave/query/transformer/DocumentTransformerSupport.java
cmitchell: Removed "In"

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
cmitchell: Maybe?  My thinking behind putting it here is that there's already something that "marks the query as started" here, so was trying to keep all of the "query started" marking code in one spot.

warehouse/query-core/src/test/java/datawave/query/LongRunningQueryTest.java
cmitchell: I don't see a related comment above.

warehouse/query-core/src/test/java/datawave/query/LongRunningQueryTest.java
cmitchell: I don't see a related comment above.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryTransformer.java
```diff
@@ -17,22 +13,25 @@
...
+import java.util.List;
+import java.util.Map;
 
 public class DiscoveryTransformer extends BaseQueryLogicTransformer<DiscoveredThing,EventBase> implements CacheableLogic {
     private List<String> variableFieldList = null;
     private BaseQueryLogic<DiscoveredThing> logic = null;
     private QueryModel myQueryModel = null;
     private MarkingFunctions markingFunctions;
     private ResponseObjectFactory responseObjectFactory;
+    private long queryExecutionForCurrentPageStartTime;
```
ivakegg: Should this actually be located in the BaseQueryLogicTransformer?

ivakegg: same comment on all other extensions of BaseQueryLogicTransformer below....

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
jwomeara: If you consider the setQueryExecutionForCurrentPageStartTime call to be part of "marking the query as started" then I would say that's even more of an argument for moving it into the markQueryStarted method.  I'll leave it up to you...

warehouse/query-core/src/test/java/datawave/query/LongRunningQueryTest.java
jwomeara: You have to click on the button that says "3 hidden conversations.  Load more..."

warehouse/query-core/src/test/java/datawave/query/LongRunningQueryTest.java
jwomeara: fair enough

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
jwomeara: ok.  sounds good

jwomeara (dismissed):
My outstanding comments are just minor nits.  If you decide to change anything for those, let me know and I'll reapprove.

warehouse/query-core/src/test/java/datawave/query/LongRunningQueryTest.java
cmitchell: Ah, removed try/catch (but kept sleep)

warehouse/query-core/src/test/java/datawave/query/LongRunningQueryTest.java
cmitchell: Removed try/catch

warehouse/query-core/src/test/java/datawave/query/LongRunningQueryTest.java
cmitchell: Removed try/catch

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryTransformer.java
cmitchell: Not sure what I was thinking.... After another look, I found that I could actually put it in `AbstractQueryLogicTransformer` (BaseQueryLogicTransformer extends AbstractQueryLogicTransformer and there are a few Transformers that extend AbstractQueryLogicTransformer directly) and I removed it from the children classes.

warehouse/query-core/src/main/java/datawave/query/dashboard/DashboardQueryLogic.java
```diff
@@ -20,6 +20,8 @@
  */
 public class DashboardQueryLogic extends ShardQueryLogic implements QueryLogicTransformer {
     
+    protected long queryExecutionForCurrentPageStartTime;
```
ivakegg, line 23: I don't think this is needed here.

warehouse/query-core/src/main/java/datawave/query/tables/ScannerSession.java
```diff
@@ -1,24 +1,16 @@
...
-import java.util.Map.Entry;
-import java.util.Set;
-import java.util.concurrent.ArrayBlockingQueue;
-import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.Executor;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
-
+import com.google.common.base.Preconditions;
```
ivakegg, line 3: Reorg of the imports is fine, but inevitably causes merge conflict with others.  Try and avoid unless really needed.

warehouse/query-core/src/main/java/datawave/query/tables/shard/IndexStatsQueryLogic.java
```diff
@@ -63,11 +63,18 @@
     public QueryLogicTransformer getTransformer(Query settings) {
         return new AbstractQueryLogicTransformer<Object,Object>() {
             
+            long queryExecutionForCurrentPageStartTime;
```
ivakegg: Is this needed in this class?

warehouse/query-core/src/main/java/datawave/query/transformer/ContentQueryTransformer.java
```diff
@@ -16,12 +12,15 @@
 import datawave.webservice.query.result.event.Metadata;
 import datawave.webservice.result.BaseQueryResponse;
 import datawave.webservice.result.DefaultEventQueryResponse;
-
 import org.apache.accumulo.core.data.Key;
 import org.apache.accumulo.core.data.Value;
 import org.apache.accumulo.core.security.Authorizations;
 import org.apache.log4j.Logger;
 
+import java.util.ArrayList;
```
ivakegg: Same comment about import reorg.  I don't think any changes are required to this class.

warehouse/query-core/src/main/java/datawave/query/transformer/EdgeQueryTransformer.java
```diff
@@ -121,4 +121,5 @@ public EdgeBase transform(Entry<Key,Value> entry) {
         }
         return edge;
     }
+
```
ivakegg: ??

warehouse/query-core/src/main/java/datawave/query/transformer/EventQueryTransformer.java
```diff
@@ -148,4 +148,5 @@ public EventBase transform(Entry<Key,Value> entry) {
         
         return event;
     }
+
```
ivakegg, line 151: ??

warehouse/query-core/src/main/java/datawave/query/transformer/FieldIndexCountQueryTransformer.java
```diff
@@ -18,17 +13,21 @@
...
 import datawave.webservice.result.BaseQueryResponse;
 import datawave.webservice.result.EventQueryResponseBase;
-
 import org.apache.accumulo.core.data.Key;
 import org.apache.accumulo.core.data.Value;
 import org.apache.accumulo.core.security.Authorizations;
 import org.apache.hadoop.io.Text;
 import org.apache.log4j.Logger;
 
+import java.util.ArrayList;
```
ivakegg: No changes needed to this class

warehouse/query-core/src/main/java/datawave/query/transformer/ShardQueryCountTableTransformer.java
```diff
@@ -16,15 +11,19 @@
...
 import datawave.webservice.query.result.event.Metadata;
+import datawave.webservice.query.result.event.ResponseObjectFactory;
 import datawave.webservice.result.BaseQueryResponse;
-
 import datawave.webservice.result.EventQueryResponseBase;
 import org.apache.accumulo.core.security.Authorizations;
 import org.apache.accumulo.core.security.ColumnVisibility;
 import org.apache.log4j.Logger;
 
+import java.util.ArrayList;
```
ivakegg: No changes required to this class.

warehouse/query-core/src/main/java/datawave/query/transformer/TermFrequencyQueryTransformer.java
```diff
@@ -1,27 +1,26 @@
...
 import org.apache.commons.lang3.StringUtils;
 import org.apache.curator.shaded.com.google.common.collect.ImmutableList;
 
-import datawave.marking.MarkingFunctions;
-import datawave.query.table.parser.TermFrequencyKeyValueFactory.TermFrequencyKeyValue;
-import datawave.webservice.query.Query;
-import datawave.webservice.query.exception.EmptyObjectException;
-import datawave.webservice.query.logic.BaseQueryLogicTransformer;
-import datawave.webservice.result.BaseQueryResponse;
+import java.util.ArrayList;
```
ivakegg: No changes required to this class

web-services/client/src/main/java/datawave/webservice/query/Query.java
```diff
@@ -61,6 +61,8 @@ public abstract void initialize(String userDN, List<String> dnList, String query
     
     public abstract Set<Parameter> getParameters();
     
+    public abstract long getQueryExecutionForCurrentPageStartTime();
```
ivakegg: I am wondering if we can shorten this name somehow.  Perhaps getCurrentPageStartTime ?  Of course this applies to the naming of this variable throughout.

web-services/query/src/main/java/datawave/webservice/query/logic/BaseQueryLogicTransformer.java
```diff
@@ -12,4 +12,5 @@ public BaseQueryLogicTransformer(MarkingFunctions markingFunctions) {
         }
         this.markingFunctions = markingFunctions;
     }
+
```
ivakegg: ??

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogicTransformer.java
```diff
@@ -1,17 +1,16 @@
...
 import datawave.webservice.query.cachedresults.CacheableLogic;
 import datawave.webservice.query.cachedresults.CacheableQueryRow;
 import datawave.webservice.query.exception.QueryException;
 import datawave.webservice.query.logic.AbstractQueryLogicTransformer;
 import datawave.webservice.query.logic.QueryLogicTransformer;
 import datawave.webservice.result.BaseQueryResponse;
-
 import org.apache.log4j.Logger;
 
+import java.util.List;
```
ivakegg: ??

warehouse/query-core/src/main/java/datawave/query/dashboard/DashboardQueryLogic.java
cmitchell: this class implements QueryLogicTransformer, which  now has `void setQueryExecutionForPageStartTime(long queryExecutionForCurrentPageStartTime);`, so this variable is used in the implementation of that method.

warehouse/query-core/src/main/java/datawave/query/transformer/DocumentTransform.java
```diff
@@ -16,9 +16,18 @@
     // called after the last document is passed through to get any remaining aggregated results.
     Map.Entry<Key,Document> flush();
     
+    /**
+     * Some transformers (GroupingTrnasform) have logic that is predicated on knowing when a page of results is starting to be processed. This sets the time
```
ivakegg: minor: GroupingTransform spelling

warehouse/query-core/src/main/java/datawave/query/tables/ScannerSession.java
cmitchell: I think this, as well as most of your follow on comments, was caused by me not originally having * imports ignored in IntelliJ.  Then, when I went through getting rid of the * imports per PR comments, the order of the imports got changed.  This shouldn't happen in the future now that I have proper settings for * imports.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryTransformer.java
```diff
@@ -17,15 +13,17 @@
 import datawave.webservice.query.logic.BaseQueryLogicTransformer;
 import datawave.webservice.query.result.event.EventBase;
 import datawave.webservice.query.result.event.FieldBase;
-import datawave.webservice.query.result.event.ResponseObjectFactory;
 import datawave.webservice.query.result.event.Metadata;
+import datawave.webservice.query.result.event.ResponseObjectFactory;
```
ivakegg: No changes required to this class.

warehouse/query-core/src/main/java/datawave/query/transformer/ContentQueryTransformer.java
cmitchell: Please see above.

warehouse/query-core/src/main/java/datawave/query/transformer/EdgeQueryTransformer.java
cmitchell: Removed -- it was leftover from moving the method into the AbstractQueryTransformer (and thus removed from child classes)

warehouse/query-core/src/main/java/datawave/query/transformer/EventQueryTransformer.java
cmitchell: Same as above

warehouse/query-core/src/main/java/datawave/query/transformer/ShardQueryCountTableTransformer.java
cmitchell: Please see above

warehouse/query-core/src/main/java/datawave/query/transformer/TermFrequencyQueryTransformer.java
cmitchell: Please see above

web-services/query/src/test/java/datawave/webservice/query/logic/composite/CompositeQueryLogicTest.java
```diff
@@ -187,6 +186,8 @@ public DifferentTestQueryLogicTransformer(MarkingFunctions markingFunctions) {
             super(markingFunctions);
         }
         
+        long queryExecutionForCurrentPageStartTime;
```
ivakegg: This is not required here as it is in the AbstractQueryLogicTransformer which BaseQueryLogicTransformer extends

web-services/client/src/main/java/datawave/webservice/query/Query.java
cmitchell: Verbosity adds clarity and self documentation. In the age of IDEs and auto completion, we're not actually typing the entire thing anyways :)

web-services/query/src/test/java/datawave/webservice/query/logic/composite/CompositeQueryLogicTest.java
ivakegg: Hence no changes to the CompositeQueryLogicTest are required.

web-services/query/src/test/java/datawave/webservice/query/logic/TestLegacyBaseQueryLogicTransformer.java
```diff
@@ -1,25 +1,24 @@
...
 import datawave.webservice.query.result.event.EventBase;
 import datawave.webservice.result.BaseQueryResponse;
-
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.powermock.api.easymock.PowerMock;
 import org.powermock.api.easymock.annotation.Mock;
 import org.powermock.modules.junit4.PowerMockRunner;
 
+import java.util.Arrays;
```
ivakegg, line 13: No changes required in this class

web-services/query/src/main/java/datawave/webservice/query/logic/BaseQueryLogicTransformer.java
cmitchell: Removed, see above

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
```diff
@@ -203,22 +204,22 @@ public ResultsPage next() throws Exception {
...
+        boolean hitIntermediateResult = false;
         try {
             addNDC();
             int currentPageCount = 0;
             long currentPageBytes = 0;
             
             // test for any exceptions prior to loop as hasNext() would likely be false;
             testForUncaughtException(resultList.size());
-            
             while (!this.finished && ((future != null) || this.iter.hasNext())) {
```
ivakegg, line 216: I am missing something here.  As this is where the beginning of a page actually starts, should there not be a call to set the page start time here?

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogicTransformer.java
cmitchell: Please see above about imports/order

warehouse/query-core/src/main/java/datawave/query/transformer/DocumentTransform.java
cmitchell: Fixed

web-services/query/src/test/java/datawave/webservice/query/logic/composite/CompositeQueryLogicTest.java
cmitchell: Yep, missed this one when I moved it into AbstractQueryLogicTrnasformer removed.

warehouse/query-core/src/main/java/datawave/query/transformer/FieldIndexCountQueryTransformer.java
cmitchell: Please see above about the imports

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryTransformer.java
cmitchell: Please see above about imports/order

web-services/query/src/test/java/datawave/webservice/query/logic/TestLegacyBaseQueryLogicTransformer.java
cmitchell: Please see above about imports/order

warehouse/query-core/src/main/java/datawave/query/tables/shard/IndexStatsQueryLogic.java
cmitchell: Removed this, as well as the setter since it's not needed.

warehouse/query-core/src/main/java/datawave/query/transformer/ContentQueryTransformer.java
cmitchell: Removed this, as well as the setter since it's not needed.

warehouse/query-core/src/main/java/datawave/query/dashboard/DashboardQueryLogic.java
ivakegg: oh right, of course.  I missed that.

warehouse/query-core/src/main/java/datawave/query/tables/ScannerSession.java
ivakegg: right, np

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryTransformer.java
ivakegg: My point on this one is that the ONLY changes you have to this class are import reordering.  Changes to this class are completely irrelevant to the pull request and can be reverted.

warehouse/query-core/src/main/java/datawave/query/transformer/ContentQueryTransformer.java
ivakegg: My point on this one is that the ONLY changes you have to this class are import reordering.  Changes to this class are completely irrelevant to the pull request and can be reverted.

warehouse/query-core/src/main/java/datawave/query/transformer/FieldIndexCountQueryTransformer.java
ivakegg: My point on this one is that the ONLY changes you have to this class are import reordering.  Changes to this class are completely irrelevant to the pull request and can be reverted.

warehouse/query-core/src/main/java/datawave/query/transformer/ShardQueryCountTableTransformer.java
ivakegg: My point on this one is that the ONLY changes you have to this class are import reordering.  Changes to this class are completely irrelevant to the pull request and can be reverted.

warehouse/query-core/src/main/java/datawave/query/transformer/TermFrequencyQueryTransformer.java
ivakegg: My point on this one is that the ONLY changes you have to this class are import reordering.  Changes to this class are completely irrelevant to the pull request and can be reverted.

web-services/client/src/main/java/datawave/webservice/query/Query.java
ivakegg: fair enough

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogicTransformer.java
ivakegg: My point on this one is that the ONLY changes you have to this class are import reordering.  Changes to this class are completely irrelevant to the pull request and can be reverted.

web-services/query/src/test/java/datawave/webservice/query/logic/TestLegacyBaseQueryLogicTransformer.java
ivakegg: My point on this one is that the ONLY changes you have to this class are import reordering.  Changes to this class are completely irrelevant to the pull request and can be reverted.

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
cmitchell: The timeout is reset in the GroupingTransform at the same time that the intermediate result flag is set.(client side) since that's what actually knows both if long running queries are allowed and if the time has elapsed.  (all the transformers depend on knowing this information -- in the case of grouping, this has to be tested before dciding to pop a document from the key list)  Since the RunningQuery detects that `hitIntermediateResult` flag, the RunningQuery  knows that a new page should be started (when next() is called again)

The running query doesn't know when the first query start time was set (which gets set on the transformers in markQueryStarted())  I understand that kicking off the query is not the exact same as when the results for a page will start to be returned, but to add access to the transformers would requiring adding a dependency to datawave-query-core, AND keeping that variable in sync between the transformers (which will still and to know it) and the RuningQuery.

web-services/query/src/test/java/datawave/webservice/query/logic/TestLegacyBaseQueryLogicTransformer.java
cmitchell: Yep, understood, was just explaining that there had been changes, and then due to PR comments, those changes were removed, but the import reordering stayed.

performed a `git checkout integration warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryTransformer.java`

warehouse/query-core/src/main/java/datawave/query/transformer/TermFrequencyQueryTransformer.java
cmitchell: Performed `it checkout integration warehouse/query-core/src/main/java/datawave/query/transformer/TermFrequencyQueryTransformer.java`

warehouse/query-core/src/main/java/datawave/query/transformer/ShardQueryCountTableTransformer.java
cmitchell: performed `git checkout integration warehouse/query-core/src/main/java/datawave/query/transformer/ShardQueryCountTableTransformer.java`

warehouse/query-core/src/main/java/datawave/query/transformer/FieldIndexCountQueryTransformer.java
cmitchell: performed ` git checkout integration warehouse/query-core/src/main/java/datawave/query/transformer/FieldIndexCountQueryTransformer.java`

warehouse/query-core/src/main/java/datawave/query/transformer/ContentQueryTransformer.java
cmitchell: performed 'git checkout integration warehouse/query-core/src/main/java/datawave/query/transformer/ContentQueryTransformer.java'

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogicTransformer.java
cmitchell: performed `git checkout integration ./web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogicTransformer.java`

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryTransformer.java
cmitchell: performed `git checkout integration warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryTransformer.java`

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
ivakegg: Ah, ok, I see it now.  Not very obvious but that will work.

warehouse/query-core/src/main/java/datawave/query/transformer/EventQueryTransformer.java
ivakegg: Please revert this class to what it was before.

web-services/query/src/main/java/datawave/webservice/query/iterator/DatawaveTransformIterator.java
```diff
@@ -1,13 +1,13 @@
 package datawave.webservice.query.iterator;
 
+import datawave.webservice.query.exception.EmptyObjectException;
 import datawave.webservice.query.logic.Flushable;
-import java.util.Iterator;
-
 import org.apache.commons.collections4.Transformer;
```
ivakegg, line 5: Please revert this class to what is in integration.

cmitchell: performed `git checkout integration ./web-services/query/src/main/java/datawave/webservice/query/iterator/DatawaveTransformIterator.java`

warehouse/query-core/src/main/java/datawave/query/transformer/EventQueryTransformer.java
cmitchell: performed `git checkout integration warehouse/query-core/src/main/java/datawave/query/transformer/EventQueryTransformer.java`

ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
```diff
@@ -168,17 +155,15 @@ public GroupingTransform(BaseQueryLogic<Entry<Key,Value>> logic, Collection<Stri
...
-            if (elapsedExecutionTimeForCurrentPage > this.queryExecutionForPageTimeout) {
-                // Reset the queryExecutionForPageStartTime and clear the documents list so that it doesn't contain
-                // duplicates. Then return an empty document with the intermediate result flag set to true
-                this.queryExecutionForPageStartTime = System.currentTimeMillis();
-                documents.clear();
-                Document intermediateResult = new Document();
-                intermediateResult.setIsIntermediateResult(true);
-                return Maps.immutableEntry(new Key(), intermediateResult);
-            }
+        // The first page queryExecutionForPageStartTime will be 0
```
jwomeara: The first queryExecutionForPageStartTime isn't 0, is it?  It's set to System.currentTimeMillis() at the beginning of each next call, right?  Guessing this is a holdover from a previous implementation?

jwomeara (changes requested):
Looks good.  Just had a minor nit.  Happy to approve if I got it wrong.

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
cmitchell: Yes, you're right. I meant to remove that comment.  Will remove it.

jwomeara (dismissed):

properties/default.properties
```diff
@@ -222,6 +222,10 @@ disable.index.only.documents=false
 # Indicates whether index-only filter functions should be enabled, such as filter:includeRegex()
 enable.index.only.filter.functions=false
 query.tld.collapse.uids=false
+# Time after which a blank page is returned to indicate to the client that the query is still running (in milliseconds)
+query.execution.for.page.timeout=3000000
+# Whether or not the query is allowed to exceed the overall query execution time out. If true, the query.execution.for.page.timeout should also be set
+query.is.long.running=false
```
billoley: Seems like this should be something more like:
query.long.running.enabled
query.long.running.allowed
query.allow.long.running
since the properties files are for configuring a build and deployment, not a specific query

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -182,6 +183,8 @@
     private Set<String> mandatoryQuerySyntax = null;
     private QueryPlanner planner = null;
     private QueryParser parser = null;
+    private QueryLogicTransformer transformerInstance = null;
+    private GroupingTransform groupingTransformerInstance = null;
```
billoley, line 187: The names transformerInstance and groupingTransformerInstance read as if they are part of a static singleton pattern, which ititially caused me a great deal of concern -- because DocumentTransformerSupport extended by DocumentTransformer is stateful on a per-query basis.  I suggest using transformer and groupingTransformer

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -577,6 +580,11 @@ protected String getStopwatchHeader(ShardQueryConfiguration config) {
     
     @Override
     public QueryLogicTransformer getTransformer(Query settings) {
+        if (this.transformerInstance != null) {
+            addConfigBasedTransformers();
+            return this.transformerInstance;
+        }
```
billoley, line 586: could move the creation and configuration of the transformer to a createTransformer() method which would leave the getTransformer method as 


```
    @Override
    public QueryLogicTransformer getTransformer(Query settings) {
        if (this.transformer == null) {
            this.transformer = createTransformer(settings)
         } else {
            addConfigBasedTransformers();
         }
         return this.transformer;
     }

```

or something like that

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1598,16 +1597,16 @@ protected UniqueTransform getUniqueTransform() {
         return uniqueTransform;
     }
     
-    protected GroupingTransform getGroupingTransform() {
-        if (groupingTransform == null && getGroupFields() != null && !getGroupFields().isEmpty()) {
+    protected GroupingIterator getGroupingIteratorInstance(Iterator<Entry<Key,Document>> in) {
```
billoley, line 1600: Maybe getGroupingIterator instead of getGroupingIteratorInstance ?

web-services/client/src/main/java/datawave/webservice/query/QueryImpl.java
```diff
@@ -281,6 +284,11 @@ public boolean isMaxResultsOverridden() {
         return parameters == null ? null : Collections.unmodifiableSet(parameters);
     }
     
+    @Override
+    public long getQueryExecutionForCurrentPageStartTime() {
+        return this.queryExecutionForCurrentPageStartTime;
+    }
+    
     public void setQueryLogicName(String name) {
```
billoley: Is this method ever called?  It looks like queryExecutionForCurrentPageStartTime is tracked through the  transformer chain.

billoley (changes requested):

properties/default.properties
cmitchell: Actually, the query.is.long.running can be removed -- the code no longer uses it.

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
cmitchell: It is meant to imply the singleton pattern on a per query. i.e. the QueryLogicTransformer and GroupingTransformer are singletons per instance of ShardQueryLogic, so the names are in keeping with what it's intended ti imply.

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
cmitchell: The creation code was always in the `getTransformer()` method -- I just added the add the null condition. Also in keeping with the singleton patter, the getInstnace() usually does the creation if the instance is null, but it could be delegated as you suggest.  Kind of a half a dozen here half a dozen there style thing.

web-services/client/src/main/java/datawave/webservice/query/QueryImpl.java
cmitchell: You're right.  I've removed both the getter and setter from here, and `Query.java` and removed the unnecessary call to the setter in `DefaultQueryPlanner` since the currentPageStartTime  is tracked from RunningQuery -> logic -> transformer.  Thanks!

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
cmitchell: I renamed that method because it is effectively returning a singleton -- before me, it already checked if the grouping iterator was null and only created a new one if so, so effectively it was a singleton before me.  The re-naming makes it clear that the getter returns a singleton.

jwomeara (dismissed):

ivakegg (dismissed):
Still not thrilled that we are not starting the clock via an obvious call in the RunningQuery, but perhaps we can change that when attacking the UniqueTransform.

jwomeara (dismissed):

ivakegg (dismissed):

ivakegg (dismissed):

billoley (dismissed):

PR #1443, opened 2022-02-25
fixes #1444 Updated the Whindex Visitor to add evaluation only key/value terms fo…

### LINKED ISSUE #1444: Whindex can produce invalid queries for one to many term mappings
For one to many term mappings, after the whindex visitor runs, it is unclear which key/value pair caused a whindex to be created.  This can cause a query to return ALL values associated with a whindex term instead of the one which created the whindex in the first place.

We should add evaluation only terms to the query with the original key/value pair which caused the whindex to be created to preserve the intent of the query.

jwomeara:
…r one to many whindex terms.

Also updated the Whindex Visitor to attempt to reduce subtrees after they have been expanded and whindex terms have been applied.


apmoriarty (dismissed):

FineAndDandy (dismissed):
tests really helped understanding here. LGTM

FineAndDandy (approved):

ivakegg (approved):

apmoriarty (approved):

PR #1451, opened 2022-03-04
Issue1421

phrocker:


phrocker:
@keith-ratcliffe I wasn't able to build this locally to be able to run the formatter. Note that I initially created this and built this in a separate branch ( branched from 2.0 I believe ), where tests passed; however, I copied these changes into a branch from feature/accumulo-2.1 where I had some conflicts I resolved through github but wasn't able to build and validate locally. Any guidance would help so I can make sure this branch works

PR #1452, opened 2022-03-04
Issue1421

phrocker:


phrocker:
@keith-ratcliffe I wasn't able to build this locally to be able to run the formatter ( I submitted from windows, hence the line break changes, which should go away when I can successfully build on linux ). Note that I initially created this and built this in a separate branch ( branched from 2.0 I believe ), where tests passed; however, I copied these changes into a branch from feature/accumulo-2.1 where I had some conflicts I resolved through github but wasn't able to build and validate locally. Any guidance would help so I can make sure this branch works

tomnelson:
I don't think that the automated tests will work until the accumulo-2.1-SNAPSHOT jars are in an available repo, and until the datawave-in-memory-accumulo 'feature/accumulo-2.1' branch is updated to fix breaking changes in accumulo-2.1-SNAPSHOT, and until the datawave-in-memory-accumulo 2.1 snapshot jars are also in some repo somewhere.

keith-ratcliffe:
@phrocker sorry for the hassle. There's still quite a bit of churn on the upstream Accumulo 2.1.x, which is why I haven't released  the dependencies you were lacking in your build. For now, if you clone these...

https://github.com/NationalSecurityAgency/datawave-in-memory-accumulo
https://github.com/NationalSecurityAgency/datawave-metadata-utils
https://github.com/NationalSecurityAgency/datawave-accumulo-utils

...and `mvn clean install` the `feature/accumulo-2.1` branch on each repo, you should be able to build the entire project.

Note that you'll probably also need to clone/build the latest Accumulo 2.1.0-SNAPSHOT as a prerequisite, as I just merged @tomnelson's `datawave-in-memory-accumulo` PR, which fixes breaking changes stemming from very recent Accumulo changes.

phrocker:
> 

Thanks @keith-ratcliffe . I had to make similar changes; however, I still have some build failures in webservice. I'll investigate those.

phrocker:
@hlgp I will update this PR. The value add of this PR would be that it changes the data wait mechanism in the scanner. What I was encountering through testing was the case where some logic operation ( like intersection) would wait on the future get to complete a bulk of pulling data ( in RangeStreamScanner::scannerInvarant ). In these cases the index stream comparator would perceive the relative ordering as the same since those many terms would be "PRESENT". 

What the changes in this PR would do would be to enable continuance and allow the stream's node to be placed into a data structure for further evaluation ( in the case of an intersection, the nodeset and children sets ). I also simplified some of the logic of adding keys to the queues by simplify8ing the event per day and shard per day thresholds, which I see after rebase was committed via #2447 . That will simplify the PR to only the timing related change, which I will rebase and update after doing some additional benchmarking. Past data showed  certain query trees seeing a 15% improvement in the range stream code for specific queries ( many terms, many of which are on a lot of shards ), but the real benefit to those queries was on the order of 2-5% median improvement of the index lookup phase -- with some much higher. -- but those which are below the median are "negligible improvement" versus a negative response. 

What I would like to do, if it's okay, is update this branch and do some testing with the current branch as I don't purport that my previous experience still holds. One of my concerns is that a change in the blocking mechanism could cause more polling and have unintended consequences so I'd like to validate that.

ivakegg:
Closing as we don't believe this is actually fixing anything of note and don't have cycles to dig into it.

PR #1457, opened 2022-03-08
Task/bloated jexl 1447

mineralntl:
At present, the values/data of the attributes are compared directly but depending on how strict/lenient we want to be, we could do actual fuzzy string matching using something like JavaWuzzy.


warehouse/query-core/src/main/java/datawave/query/function/JexlContextCreator.java
```diff
@@ -64,7 +65,47 @@ public JexlContextCreator() {
...
-        return Tuples.tuple(from.first(), from.second(), context);
+        return Tuples.tuple(dedupedFrom.first(), dedupedFrom.second(), context);
+    }
+    
+    private Tuple3 dedupeAttributes(Tuple3<Key,Document,Map<String,Object>> rawTuple) {
+        Map<String,Attribute<? extends Comparable<?>>> dict = rawTuple.second().getDictionary();
+        Set<Attribute<? extends Comparable<?>>> dedupedAttributeSet = new HashSet<Attribute<? extends Comparable<?>>>();
+        
+        for (Entry<String,Attribute<? extends Comparable<?>>> entry : dict.entrySet()) {
+            if (entry.getKey().equals(BODY_KEY)) {
```
ivakegg: Why are we only doing this for the BODY_KEY?  The BODY values are in fact index only so there will be no values from the event.  Dedupping those values are reasonable but why not do that for all fields?

warehouse/query-core/src/main/java/datawave/query/function/JexlContextCreator.java
```diff
@@ -64,7 +65,47 @@ public JexlContextCreator() {
...
+                Attributes rawButes = (Attributes) entry.getValue();
+                Set<Attribute<? extends Comparable<?>>> attributeSet = rawButes.getAttributes();
+                List<String> valueList = new ArrayList<String>();
+                
+                // get list of any non-preNormalizedAttributes since those are the ones
+                // we actually care about and want to keep
+                attributeSet.forEach(attribute -> {
```
ivakegg: I think we should use the Attribute.isFromIndex() instead of checking for PreNormalizedAttribute.  That would be much less fragile.

ivakegg (changes requested):

mineralntl:
Belay this PR until further notice - spoke with @apmoriarty and need to rework the logic.

mineralntl:
Completely changed direction based on team feedback.

warehouse/query-core/src/main/java/datawave/query/util/FuzzyAttributeComparator.java
```diff
@@ -0,0 +1,91 @@
+package datawave.query.util;
+
+import com.google.common.collect.Sets;
+import datawave.query.attributes.Attribute;
+import datawave.query.attributes.Attributes;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public final class FuzzyAttributeComparator {
```
apmoriarty: I know I wrote the original ticket with the fuzzy wording, but let's change this class name. Fuzzy matching is a mathematical operation and in the end we're just doing a comparison of the backing data values. 

Rename to `AttributeComparator` or something similar. Add a class level comment that we're identifying attributes with equivalent data values across attribute types.

warehouse/query-core/src/main/java/datawave/query/util/FuzzyAttributeComparator.java
```diff
@@ -0,0 +1,91 @@
...
+    
+    // this is a utility and should never be instantiated
+    private FuzzyAttributeComparator() {
+        throw new UnsupportedOperationException();
+    }
+    
+    public static boolean singleToSingle(final Attribute existingAttribute, final Attribute newAttribute) {
```
apmoriarty: Let's standardize the argument naming convention here. The method names are keeping to `single` and `multiple` but the args are `existing`, `new`, `single`, and `multiple`. You could make this a simple `left`/`right` or `leftSingle` and `rightSingle`. Or perhaps `attr` for single Attribute and `attrs` for an instance of Attributes. Just make the args consistent among themselves.

warehouse/query-core/src/main/java/datawave/query/util/FuzzyAttributeComparator.java
```diff
@@ -0,0 +1,91 @@
...
+        return (existingAttribute.getData().equals(newAttribute.getData())
+                        && existingAttribute.getColumnVisibility().equals(newAttribute.getColumnVisibility()) && existingAttribute.getTimestamp() == newAttribute
+                        .getTimestamp());
+    }
+    
+    public static boolean singleToMultiple(final Attributes multipleAttributes, final Attribute singleAttribute) {
```
apmoriarty: Method args should follow the method name. Single attribute on the left, Multiple attributes on the right.

warehouse/query-core/src/main/java/datawave/query/util/FuzzyAttributeComparator.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+        return containsMatch;
+    }
+    
+    /**
+     * Performs a basic fuzzy match on the content/data of two given Attributes. If such a match is found, combine the metadata from the two Attributes and
+     * create a new set of deduplicated Attributes to return to the Document.
+     *
+     * @param existingAttributes
+     *            The set of Attributes against which we want to compare
```
apmoriarty: Strictly speaking, we do not have a set of attributes (`Set<Attributes>`). We have a single instance of `Attributes`, which is itself a set of type attribute (`Set<Attribute>`)

warehouse/query-core/src/test/java/datawave/query/util/FuzzyAttributeComparatorTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+import java.util.HashSet;
+import java.util.Set;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class FuzzyAttributeComparatorTest {
```
apmoriarty: This test looks fine overall, but let's add a `DocumentTest` and roll through a number of expected cases.
1. add two instances of `Attribute` to document
2. add two instances of `Attributes` to document
3. add a mix of `Attribute` and `Attributes` to document

For each case outlined above hit the following variants
- test case without duplicate values
- test case with duplicate value
- test case with duplicate value in different Attribute types (Content vs. PreNormalizedAttribute)

Also be sure to hit a few edge cases
- merge attribute with document key vs. no document key
- merge attribute with toKeep flag set/not set
- merge attribute with different visibilities (test will act as documentation of expected behavior)

apmoriarty (changes requested):
Overall this is looking much better, would like to see a few more test cases. Give some consideration to putting this work behind a feature flag.

warehouse/query-core/src/main/java/datawave/query/util/FuzzyAttributeComparator.java
mineralntl: Done and done

warehouse/query-core/src/main/java/datawave/query/util/FuzzyAttributeComparator.java
mineralntl: Fair enough - will change the phrasing.

warehouse/query-core/src/main/java/datawave/query/util/FuzzyAttributeComparator.java
mineralntl: Done and done.

warehouse/query-core/src/main/java/datawave/query/function/JexlContextCreator.java
mineralntl: OBE

warehouse/query-core/src/main/java/datawave/query/function/JexlContextCreator.java
mineralntl: Updated to check all keys

warehouse/query-core/src/main/java/datawave/query/util/FuzzyAttributeComparator.java
mineralntl: Done and done.

warehouse/query-core/src/test/java/datawave/query/util/FuzzyAttributeComparatorTest.java
mineralntl: Updated merging logic and added new edge cases as requested

warehouse/query-core/src/main/java/datawave/query/attributes/Attributes.java
```diff
@@ -1,5 +1,15 @@
...
+import com.esotericsoftware.kryo.Kryo;
+import com.esotericsoftware.kryo.io.Input;
+import com.esotericsoftware.kryo.io.Output;
+import datawave.marking.MarkingFunctions;
+import datawave.query.collections.FunctionalSet;
+import datawave.query.jexl.DatawaveJexlContext;
+import org.apache.commons.lang.builder.HashCodeBuilder;
+import org.apache.hadoop.io.WritableUtils;
+import org.apache.log4j.Logger;
+
```
ivakegg: I think you can remove this change from the set of changes....

warehouse/query-core/src/main/java/datawave/query/util/AttributeComparator.java
```diff
@@ -0,0 +1,131 @@
...
+        List<Attribute> trueDuplicates = new ArrayList<Attribute>();
+        
+        attrs1.getAttributes().forEach(attr1 -> {
+            AtomicBoolean containsMatch = new AtomicBoolean(false);
+            attrs2.getAttributes().forEach(attr2 -> {
+                if (singleToSingle(attr1, attr2) && (attr1.isToKeep() || attr2.isToKeep())) {
+                    if (attr1.getMetadata().getSize() == attr2.getMetadata().getSize()) {
+                        containsMatch.set(true);
+                        trueDuplicates.add(attr1);
+                        combinedAttrMap.put(attr1, 1); // found true match so return pre-existing Attribute
```
ivakegg, line 93: If !attr1.isToKeep and attr2.isToKeep, but we are only retaining attr1 then I think we need to set attr1.keep flag.

ivakegg: same comment goes for below when retaining attr2 instead

warehouse/query-core/src/main/java/datawave/query/util/AttributeComparator.java
```diff
@@ -0,0 +1,131 @@
...
+                containsMatch = true;
+            }
+        }
+        return containsMatch;
+    }
+    
+    /**
+     * Performs a basic match on the content/data of two given Attributes. If such a match is found, combine the metadata from the two Attributes and create new
+     * deduplicated Attributes to return to the Document.
+     *
```
ivakegg, line 74: document how you are choosing which attr to keep if two values are the same.

warehouse/query-core/src/main/java/datawave/query/util/AttributeComparator.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+    
+    public static Set<Attribute<? extends Comparable<?>>> combineMultipleAttributes(final Attributes attrs, final Attribute attr, final boolean trackSizes) {
+        HashSet<Attribute<? extends Comparable<?>>> attrSet = Sets.newHashSet();
+        attrSet.add(attr);
+        
+        return combineMultipleAttributes(attrs, new Attributes(attrSet, attrs.isToKeep(), trackSizes));
+    }
+    
+    public static Attribute combineSingleAttributes(final Attribute attr1, final Attribute attr2) {
```
ivakegg, line 149: method comments

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/util/AttributeComparator.java
```diff
@@ -0,0 +1,131 @@
...
+        int attr1mdSize = attr1.getMetadata().getSize();
+        int attr2mdSize = attr2.getMetadata().getSize();
+        
+        Attribute mergedAttr = null;
```
ivakegg, line 153: We also need to look at the isFromIndex flag in here.  We need to prefer holding onto things from the event over things from the index.  This has a huge impact when using the #OCCURANCE function.

warehouse/query-core/src/main/java/datawave/query/attributes/Attributes.java
mineralntl: Noted - will undo this in the next commit

warehouse/query-core/src/main/java/datawave/query/util/AttributeComparator.java
mineralntl: Will do

warehouse/query-core/src/main/java/datawave/query/util/AttributeComparator.java
mineralntl: Ah good catch - will update in the next commit

warehouse/query-core/src/main/java/datawave/query/util/AttributeComparator.java
mineralntl: Will add

warehouse/query-core/src/main/java/datawave/query/util/AttributeComparator.java
mineralntl: Added as per team request

mineralntl:
Closing as outdated - see #1612 for updated PR

PR #1461, opened 2022-03-10
Hittermtype

tomnelson:
I will also make this same PR against the integration branch. This fixes a problem where Content was probably not the best choice as a type for the hit terms, and also the accumulo key added to the value of the Content type was not making thru the serialization / deserialization stages, so it was never seen.
The new HitTermType could/should go in the datawave-type-utils project at some point, but that module may be harder to make a new release of.


ivakegg:
I think this will get into the accumulo 2.1 branch when we merge integration into it which I believe Keith does periodically.

PR #1462, opened 2022-03-10
Hittermtype2

tomnelson:
Added HitTermType and changed hit list code to use it instead of Content for hit terms.
The motivation for this was as follows:
- I saw that the document after the serialization and deserialization did not match the original for the HIT_TERM key/value.
The existing code puts the accumulo key into the value of the HIT_TERM element, but that is lost during serialization/deserialization. This was evident while comparing to another effort to use json serialization.
I felt that the use of Content may not have been a good choice, however changing to just use TypeAttribute with a StringType was a problem because it was difficult to differentiate between the hit term items and the actual query return items. 
I added a HitTermType that extends StringType. This class should probably live in datawave-type-utils in the future, when that project is ready for a release. At that time, the imports can be changed here to use that version.


warehouse/query-core/src/main/java/datawave/query/attributes/HitTermType.java
```diff
@@ -0,0 +1,11 @@
+package datawave.query.attributes;
+
+import datawave.data.type.StringType;
+
+import java.util.function.Predicate;
+
+public class HitTermType extends StringType {
+    
+    public static final Predicate<Attribute> matches = attr -> attr instanceof TypeAttribute && ((TypeAttribute) attr).getType() instanceof HitTermType;
```
phrocker: Would a static method that returns a boolean simplify the code leveraging this predicate? Is there other code that uses the predicate in a filter function?

tomnelson: it sure would!

warehouse/query-core/src/main/java/datawave/query/attributes/HitTermType.java
```diff
@@ -0,0 +1,11 @@
+package datawave.query.attributes;
+
+import datawave.data.type.StringType;
+
+import java.util.function.Predicate;
+
+public class HitTermType extends StringType {
```
ivakegg: Let go ahead and move this to Type-utils now.  Releasing a new version of that and updating dependencies in the datawave repo is not hard to do.  The problem with putting here is that it will likely never be moved.

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
```diff
@@ -231,10 +233,9 @@ private void fillHitTermMap(Attribute<?> attr, Map<String,String> attrMap) {
                 for (Attribute<?> at : attrs.getAttributes()) {
                     fillHitTermMap(at, attrMap);
                 }
-            } else if (attr instanceof Content) {
-                Content content = (Content) attr;
+            } else if (HitTermType.matches.test(attr)) {
```
ivakegg: Don't we need to leave the Content section in here as well?>

tomnelson: I want to document the difference here to make sure.
In a unit test (HitsAreAlwaysIncludedTest)
Here's the Content attribute:
image


Here's what went into the attributeMap:

image

Here's what is in the Document's dictionary to be serialized:

image



Here's what comes out after deserializing:
image


With my changes, here is the HitTermType attribute:
image

Here is what goes into the attributeMap:
image

Here's what is in the Document's dictionary to be serialized:
image


Here's what comes out after deserializing:
image

warehouse/query-core/src/main/java/datawave/query/attributes/HitTermType.java
tomnelson: I opened a PR in datawave-type-utils

warehouse/query-core/src/main/java/datawave/query/attributes/HitTermType.java
tomnelson: I removed this convenience method/predicate when I put HitTermType into datawave-type-utils

ivakegg (dismissed):

keith-ratcliffe (dismissed):

warehouse/query-core/src/test/java/datawave/query/function/JexlEvaluationTest.java
```diff
@@ -159,7 +159,7 @@ public void testContentPhraseFunction() {
         boolean foundPhrase = false;
         Attributes attrs = (Attributes) d.get("HIT_TERM");
         for (Attribute<?> attr : attrs.getAttributes()) {
-            if (attr.getData().equals("TOKFIELD:big red dog")) {
+            if (attr.toString().equals("TOKFIELD:big red dog")) {
```
ivakegg: We should not be counting on toString() here.  I think we still need to verify that getData() returns this value.

warehouse/query-core/src/test/java/datawave/query/function/HitsAreAlwaysIncludedTest.java
```diff
@@ -187,14 +188,14 @@ protected void runTestQuery(Connector connector, String queryString, Date startD
             if (hitAttribute instanceof Attributes) {
                 Attributes attributes = (Attributes) hitAttribute;
                 for (Attribute attr : attributes.getAttributes()) {
-                    if (attr instanceof Content) {
-                        Content content = (Content) attr;
-                        Assert.assertTrue(expectedHits.remove(content.getContent()));
+                    if (attr instanceof TypeAttribute && ((TypeAttribute) attr).getType() instanceof HitTermType) {
+                        TypeAttribute content = (TypeAttribute) attr;
+                        Assert.assertTrue(expectedHits.remove(content.toString()));
```
ivakegg: content.getData() instead of content.toString() ?

warehouse/query-core/src/test/java/datawave/query/function/HitsAreAlwaysIncludedTest.java
```diff
@@ -187,14 +188,14 @@ protected void runTestQuery(Connector connector, String queryString, Date startD
...
                     }
                 }
-            } else if (hitAttribute instanceof Content) {
-                Content content = (Content) hitAttribute;
-                Assert.assertTrue(content.getContent() + " is not an expected hit", expectedHits.remove(content.getContent()));
+            } else if (hitAttribute instanceof TypeAttribute) {
+                TypeAttribute attr = (TypeAttribute) hitAttribute;
+                Assert.assertTrue(attr + " is not an expected hit", expectedHits.remove(attr.toString()));
```
ivakegg: attr.getData() instenad of attr.toString() ?

ivakegg (changes requested):
see comments

warehouse/query-core/src/test/java/datawave/query/HitsAreAlwaysIncludedCommonalityTokenTest.java
```diff
@@ -178,14 +176,13 @@ protected void runTestQuery(Connector connector, String queryString, Date startD
             if (hitAttribute instanceof Attributes) {
                 Attributes attributes = (Attributes) hitAttribute;
                 for (Attribute attr : attributes.getAttributes()) {
-                    if (attr instanceof Content) {
-                        Content content = (Content) attr;
-                        Assert.assertTrue(goodResults.contains(content.getContent()));
+                    if (TypeAttribute.matches(attr, HitTermType.class)) {
+                        String content = attr.toString();
```
keith-ratcliffe: Ivan's comment about relying on `getData()` instead of `toString()` should apply here too, I think

keith-ratcliffe (changes requested):

warehouse/query-core/src/test/java/datawave/query/function/JexlEvaluationTest.java
tomnelson: getData() returns an Object, the delegate, which is actually the underlying Type (inside the TypeAttribute)
In this case, it returns a HitTermType which extends StringType which extends BaseType&lt;String&gt;...
I could call attr.getData().toString() and fail if getData() returns null, 
but our TypeAttribute has a toString() that checks for a null delegate and then either returns getDelegateAsString, or it returns the class name and other stuff. That null case return won't match what's in the expected results so the test would still work.
Thoughts?

tomnelson: indeed, Content getData() returned the Content's data, which is always a String but returned as an Object.

tomnelson: also, these are only unit tests, not production code, that get the string values to compare against expected string values.

ivakegg: If one would happen to change what toString() returns, that should not break test cases.  If one changes for getData() returns, one could expect that to break test cases.

tomnelson: I'll avoid the toString and just do something like this:

// if its a TypeAttribute
((TypeAttribute)attr).getType().getDelegateAsString()

tomnelson: I made changes and pushed to the branch.

keith-ratcliffe (approved):

keith-ratcliffe:
@ivakegg , if you're ok with @tomnelson 's updates I think we can get this merged

PR #1463, opened 2022-03-10
fix mapreduce cert uris in add-datawave-configuration.cli

drewfarris:
* they were missing a slash which caused the uris to be treated like ftp fetches


drewfarris:
Fixed in #1414 already.

PR #1468, opened 2022-03-21
WIP: Upgrade spring to 5.3.7

hlgp:


jwomeara:
If it's not too much trouble, could you target at least 5.3.15?  That's the version included by the latest version of spring boot that we're using in the query microservices - 2.6.3.  Looks like the absolute latest is 2.6.4, so we may bump up to that version shortly.

jwomeara:
We don't want to make this change anymore?

PR #1472, opened 2022-03-24
fixes #1471: Allow normalization in all cases.

### LINKED ISSUE #1471: Evaluation Only and other delayed predicates incorrectly skip term normalization
The normalization visitor invoked with the DefaultQueryPlanner will skip over "delayed" predicates.  This include evaluation only, exceeded value, ... query property markers.  This is incorrect as it will result in missed results as only the normalized values are being compared against in the JexlContext except for a very few exceptions (see #TEXT or includeText in the evaluation only phase functions).

ivakegg:


apmoriarty (dismissed):

ivakegg:
Working on adding test cases....

hlgp (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
```diff
@@ -136,10 +136,7 @@ public Object visit(ASTFunctionNode node, Object data) {
     
     @Override
     public Object visit(ASTReference node, Object data) {
-        /**
-         * If we have a delayed predicate we can safely assume that expansion has occurred in the unfielded expansion along with all types
-         */
-        if (QueryPropertyMarker.findInstance(node).isDelayedPredicate() || this.expandedNodes.contains(node)) {
```
jwomeara, line 142: Originally, evaluation only was added for composite nodes, which were generated post-normalization.  Now that the whindex exists, it is possible that something will be marked evaluation only prior to normalization.  (Also, the query string can include evaluation only terms, so I guess it's possible there as well).  Allowing evaluation only terms to be normalized seems safe enough to me.  Are we sure we want to do that for EVEYRYTHING that is considered delayed though?  If the answer is yes, then feel free to ignore this comment.  Otherwise, consider just removing evaluation only from the list of 'delayed' nodes.

jwomeara (dismissed):
Approved, but please check out my comment before merging.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
ivakegg: I saw no harm in normalizing any of those constructs.  Normalizing something that is normalized will result in the same value.

jwomeara: at this point in the query nothing has been normalized, so i'm not entirely sure what you're getting at with the second part of your response.  if you see no harm in normalizing index hole, delayed, exceeded or, exceeded term, or exceeded value nodes, then the approval stands.

jwomeara (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
ivakegg: Well I actually need to keep the ExceededValue and ExceededTerm threshold markers per the original comment.

ivakegg:
OK @hlgp @jwomeara @apmoriarty .  I think this should be good to go.  I setup a pipeline to test this.

jwomeara (approved):

hlgp (approved):

PR #1481, opened 2022-03-31
Add test cases for Projection, KeyProjection and DocumentProjection

apmoriarty:
- Added test cases for Projection, KeyProjection, and DocumentProjection
- Simplified Projection class by removing useless booleans
- Updated whitelist/blacklist language to include/exclude

I haven't put this through integration testing yet, but feel free to review in the meantime.


warehouse/query-core/src/test/java/datawave/query/predicate/KeyProjectionTest.java
```diff
@@ -0,0 +1,125 @@
...
+        
+        Iterator<Entry<Key,String>> iter = fiData.iterator();
+        assertTrue(projection.apply(iter.next()));
+    }
+    
+    @Test(expected = RuntimeException.class)
+    public void testTooMuchOfTheSameConfiguration() {
+        KeyProjection projection = new KeyProjection();
+        projection.setExcludes(Sets.newHashSet("FIELD_X", "FIELD_Y"));
+        projection.setExcludes(Sets.newHashSet("FIELD_X", "FIELD_Y"));
```
jwomeara, line 63: Looking at the code I understand why this fails, but why do we want it to fail?  I get that this is not a new concept - just trying to understand.  I'm guessing in practice it's never even set twice, but if it were, what would be the harm in just resetting to use whatever the last setter said to use (be it setInclude or setExclude)?

jwomeara (approved):
Added a comment with a question.  Not a ding at all, just curious.

warehouse/query-core/src/test/java/datawave/query/predicate/KeyProjectionTest.java
cmitchell: I second this question.

cmitchell (approved):
Pretty straight forward and clean.

warehouse/query-core/src/test/java/datawave/query/predicate/KeyProjectionTest.java
apmoriarty: As far as I can tell, these projections were meant to be created once and this design decision was enforced by throwing an exception. Presumably this was to prevent state change in how the query handles includes and excludes?

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1245,10 +1244,10 @@ protected DocumentProjection getProjection() {
         DocumentProjection projection = new DocumentProjection(this.isIncludeGroupingContext(), this.isReducedResponse(), isTrackSizes());
         
         if (this.useWhiteListedFields) {
```
jwomeara, line 1246: Do we want to update this language as well?

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1245,10 +1244,10 @@ protected DocumentProjection getProjection() {
...
-            projection.initializeWhitelist(this.whiteListedFields);
+            projection.setIncludes(this.whiteListedFields);
             return projection;
         } else if (this.useBlackListedFields) {
```
jwomeara, line 1249: and here

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1245,10 +1244,10 @@ protected DocumentProjection getProjection() {
...
-            projection.initializeBlacklist(this.blackListedFields);
+            projection.setExcludes(this.blackListedFields);
             return projection;
         } else {
             String msg = "Configured to use projection, but no whitelist or blacklist was provided";
```
jwomeara, line 1253: and here too?

jwomeara (approved):

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
jwomeara: sorry.  should have added these comments on the first pass.  approval still stands.

warehouse/query-core/src/test/java/datawave/query/predicate/KeyProjectionTest.java
jwomeara: The includes/excludes should probably be immutable if that's the case and passed in as constructor arguments vs. using a setter.  Again, not really related to this PR, just an observation.  The approval still stands for me.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
apmoriarty: No problem. I'm limiting the scope to just the Projection classes, updating the QueryIterator, various querylogics, and configs will be another unit of work (with a higher cost to integrate)

warehouse/query-core/src/test/java/datawave/query/predicate/KeyProjectionTest.java
FineAndDandy: Think I'd just like to see a new issue created for this. The scope of this ticket is really just updating the methods and adding unit tests for the existing behavior.

FineAndDandy (approved):

warehouse/query-core/src/test/java/datawave/query/predicate/KeyProjectionTest.java
apmoriarty: Created https://github.com/NationalSecurityAgency/datawave/issues/1512 to update the contract for the Projection classes

PR #1482, opened 2022-03-31
Add term frequency excerpts to results

### LINKED ISSUE #1387: Incorporate excerpts into the results
A TermFrequencyExcerptIterator was created that demonstrates how to pull an excerpt out of the term frequencies.  What is needed is the ability to specify in the query that excerpts for a specified field and a specified number of surrounding terms should be returned in the query.  Initially this would be a query parameter such as excerpt_fields:BODY/10,...  but then we could also add a lucene function such as #EXCERPT_FIELDS(BODY/10,...).
The TermFrequencyExcerptIterator (or something like that) will need to be integrated into the QueryIterator.  The excepts could be returned in the HIT_TERMS field or in a separate field like HIT_EXCERPTS or maybe even separate fields such as BODY_EXCERPT.

lbschanno:
Support retrieving excerpts from term frequencies and adding them to the
results. Excerpts can be specified by the new #EXCERPT_FIELDS function,
and will be added to the HIT_EXCERPTS field in the results.

Resolves #1387

I will continue adding tests and documentations to this body of work, but it is ready for review.


warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -0,0 +1,197 @@
...
+        if (entry != null) {
+            Document document = entry.getValue();
+            // Do not bother adding excerpts to transient documents.
+            if (document.isToKeep()) {
+                PhraseIndexes phraseIndexes = getPhraseIndexes(document);
+                if (phraseIndexes != null) {
+                    Set<String> excerpts = getExcerpts(phraseIndexes, document);
+                    addExcerptsToDocument(excerpts, document);
+                } else {
+                    log.debug("Phrase indexes were not added to document " + document.getMetadata() + ", skipping");
```
ivakegg: This might be better as trace since the majority of the time there will be no phrase indexes.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermOffsetPopulator.java
```diff
@@ -79,41 +79,6 @@ public Document document() {
         return termFrequencyFieldValues;
     }
     
-    // merge two maps presuming both came from getContextMap()
-    @SuppressWarnings("unchecked")
-    public static Map<String,Object> mergeContextMap(Map<String,Object> map1, Map<String,Object> map2) {
```
ivakegg, line 84: I double checked this is not used in the other codebase.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
ivakegg: And then only if log.isTraceEnabled()

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -0,0 +1,197 @@
...
+        }
+        return excerpts;
+    }
+    
+    private String getExcerpt(String field, int start, int end, Range range) {
+        Map<String,String> options = new HashMap<>();
+        options.put(TermFrequencyExcerptIterator.FIELD_NAME, field);
+        options.put(TermFrequencyExcerptIterator.START_OFFSET, String.valueOf(start));
+        options.put(TermFrequencyExcerptIterator.END_OFFSET, String.valueOf(end));
+        TermFrequencyExcerptIterator iterator = new TermFrequencyExcerptIterator();
```
ivakegg: Do we need to construct a new one each time?

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -0,0 +1,197 @@
...
+     */
+    private Set<String> getExcerpts(PhraseIndexes phraseIndexes, Document document) {
+        phraseIndexes = getOffsetPhraseIndexes(phraseIndexes);
+        if (phraseIndexes.isEmpty()) {
+            return Collections.emptySet();
+        }
+        
+        // Construct the required range for this document.
+        Key metadata = document.getMetadata();
+        Key startKey = new Key(metadata.getRow(), new Text("datatype\0uid"));
```
ivakegg: What I meant by the datatype\0uid is to use the actually datatype of the document and the uid of the document separated by a null byte.  I believe the column family of the metadata Key is exactly that.  So this should be metadata.getColumnFamily().

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -0,0 +1,197 @@
...
+        Map<String,String> options = new HashMap<>();
+        options.put(TermFrequencyExcerptIterator.FIELD_NAME, field);
+        options.put(TermFrequencyExcerptIterator.START_OFFSET, String.valueOf(start));
+        options.put(TermFrequencyExcerptIterator.END_OFFSET, String.valueOf(end));
+        TermFrequencyExcerptIterator iterator = new TermFrequencyExcerptIterator();
+        try {
+            iterator.init(source, options, env);
+            iterator.seek(range, Collections.emptyList(), false);
+            if (iterator.hasTop()) {
+                iterator.getTopKey();
```
ivakegg: Line 300 of the TermFrequencyExceptIterator class shows what the key will look like that is returned here.  The column qualifier will contain the actually excerpt preceeded by the field name and a null character.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -0,0 +1,197 @@
...
+            }
+            // TODO - seek to phrase
```
ivakegg: nothing to TODO here.  If the iterator does not have a top, then there is no except to return.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -0,0 +1,197 @@
...
+        Key endKey = startKey.followingKey(PartialKey.ROW_COLFAM);
+        Range range = new Range(startKey, true, endKey, false);
+        
+        // Fetch the excerpts.
+        Set<String> excerpts = new HashSet<>();
+        for (String field : phraseIndexes.getFields()) {
+            Collection<Pair<Integer,Integer>> indexes = phraseIndexes.getIndices(field);
+            for (Pair<Integer,Integer> indexPair : indexes) {
+                String excerpt = getExcerpt(field, indexPair.getValue0(), indexPair.getValue1(), range);
+                excerpts.add(excerpt);
```
ivakegg: For safety, check if excerpt is null here.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
ivakegg: might want to log a warning if this is the case.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
lbschanno: Fixed.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermOffsetPopulator.java
lbschanno: Awesome, thanks for checking.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
lbschanno: Understood, thank you for clarifying. Fixed.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
lbschanno: I don't believe so. I will modify this to reuse the same iterator instance.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
lbschanno: Good catch, I had meant to finish this part. Fixed.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -0,0 +1,197 @@
...
+        TermFrequencyExcerptIterator iterator = new TermFrequencyExcerptIterator();
+        try {
+            iterator.init(source, options, env);
+            iterator.seek(range, Collections.emptyList(), false);
+            if (iterator.hasTop()) {
+                iterator.getTopKey();
+            }
+            // TODO - seek to phrase
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to scan for excerpts", e);
```
lbschanno: @ivakegg do you want an exception thrown here, or swallowed and logged instead? If an exception, is there a type you prefer?

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
lbschanno: Done.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
lbschanno: Will do.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
ivakegg: This is probably fine as is.  I do not want it swallowed.

lbschanno: Okay.

ivakegg (dismissed):

avgAGB (dismissed):

ivakegg (dismissed):

ivakegg (approved):

warehouse/query-core/src/main/java/datawave/query/attributes/ExcerptFields.java
```diff
@@ -0,0 +1,200 @@
...
+        if (string.isEmpty()) {
+            return new ExcerptFields();
+        }
+        
+        ExcerptFields excerptFields = new ExcerptFields();
+        
+        String[] fieldParts = string.split(Constants.COMMA);
+        for (String fieldPart : fieldParts) {
+            String[] parts = StringUtils.split(fieldPart, Constants.FORWARD_SLASH);
+            excerptFields.put(parts[0], Integer.valueOf(parts[1]));
```
FineAndDandy, line 60: validate formatting?

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermOffsetMap.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.StringJoiner;
+
+/**
+ * This encapsulates term frequencies information that is used when evaluating content:phrase and content:within functions in a JEXL query. It is also used to
+ * store phrase indexes that have been found for matching hits so that we can retrieve excerpts when requested.
+ */
+public class TermOffsetMap {
```
apmoriarty, line 16: Does this class support merging with another `TermOffsetMap`?

apmoriarty (approved):
LGTM -- one question that doesn't affect approval

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermOffsetMap.java
lbschanno: Currently no, there was code in TermOffsetPopulator that merged two term frequency lists, but Ivan verified that it was not in use anywhere, so I deleted it and didn't implement support for it for TermOffsetMap. If need be, I can implement it.

apmoriarty: Let's add merging TermOffsetMaps as an issue to be done at some point in the future. There's currently no use case, but at some point we may want to shard documents which would require merging them.

lbschanno: Sounds good to me.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1615,4 +1625,15 @@ protected ActiveQueryLog getActiveQueryLog() {
         }
         return this.activeQueryLog;
     }
+    
+    protected ExcerptTransform getExcerptTransform() {
+        if (excerptTransform == null && getExcerptFields() != null && !getExcerptFields().isEmpty()) {
+            synchronized (getExcerptFields()) {
```
FineAndDandy, line 1631: why a synchronized block here?

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
```diff
@@ -0,0 +1,452 @@
...
+        
+        if (log.isTraceEnabled()) {
+            log.trace(source.hasTop() + " nexting on " + scanRange);
+        }
+        
+        // find a valid dt/uid (depends on initial column families set in seek call)
+        String dtUid = null;
+        while (source.hasTop() && dtUid == null) {
+            Key top = source.getTopKey();
+            String thisDtUid = getDtUidFromTfKey(top);
```
FineAndDandy, line 245: is it possible with a multi-row range that next() doesn't return a TF key?

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermOffsetMap.java
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * Put the {@link TermFrequencyList} for the specified field
+     * 
+     * @param field
+     *            the field
+     * @param termFrequencyList
+     *            the term frequency list
+     */
+    public void putTermFrequencyList(String field, TermFrequencyList termFrequencyList) {
+        termFrequencies.put(field, termFrequencyList);
```
FineAndDandy, line 45: this will replace any existing value rather than merge if it already exists

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/PhraseIndexes.java
```diff
@@ -0,0 +1,166 @@
...
+        
+        // Strip whitespaces.
+        string = StringUtils.deleteWhitespace(string);
+        
+        PhraseIndexes phraseIndexes = new PhraseIndexes();
+        String[] fieldParts = string.split(Constants.FORWARD_SLASH);
+        for (String fieldPart : fieldParts) {
+            String[] parts = fieldPart.split(Constants.COLON);
+            String field = parts[0];
+            for (int i = 1; i < parts.length; i++) {
```
FineAndDandy, line 49: check length first to avoid npe

PR #1483, opened 2022-04-01
Fix bug in content function index expansion

apmoriarty:
Fix bug in content function index expansion when a phrase is a repeated term. Addresses #1484

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionsDescriptor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionsDescriptor.java
@@ -426,8 +426,8 @@ public class ContentFunctionsDescriptor implements JexlFunctionArgumentDescripto
             JexlNode deref = JexlASTHelper.dereference(indexQuery);
             if (deref instanceof ASTAndNode || deref instanceof ASTEQNode) {
                 
-                // it is possible that an index query contains only a single equality node
-                // i.e., "phrase('foo', 'foo')" would produce a single equality node "FOO == 'bar'"
+                // the index query may be a single equality node when the phrase is composed of repeated terms
+                // i.e., "phrase(termOffsetMap, 'bar', 'bar')" will produce an index query of "FOO == 'bar'"
                 components.add(deref);
                 
             } else if (deref instanceof ASTOrNode) {
@@ -474,8 +474,10 @@ public class ContentFunctionsDescriptor implements JexlFunctionArgumentDescripto
                 children.add(updateFunctionWithField(function, field));
                 if (component instanceof ASTEQNode) {
                     children.add(component);
-                } else {
+                } else if (component instanceof ASTAndNode) {
                     children.addAll(Arrays.asList(JexlNodes.children(component)));
+                } else {
+                    throw new IllegalStateException("Unexpected component. Expected ASTAndNode or ASTEqNode but was: " + component.getClass().getSimpleName());
                 }
                 return JexlNodeFactory.createAndNode(children);
             } else {
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionsDescriptor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionsDescriptor.java
@@ -424,9 +424,12 @@ public class ContentFunctionsDescriptor implements JexlFunctionArgumentDescripto
             List<JexlNode> components = new LinkedList<>();
             
             JexlNode deref = JexlASTHelper.dereference(indexQuery);
-            if (deref instanceof ASTAndNode) {
-                // single component
+            if (deref instanceof ASTAndNode || deref instanceof ASTEQNode) {
+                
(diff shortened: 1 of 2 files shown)


apmoriarty:
@FineAndDandy @hlgp @jwomeara If you all could provide a review that would be much appreciated

warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionsDescriptor.java
```diff
@@ -461,15 +464,19 @@ public static JexlNode distributeFunctionIntoIndexQuery(JexlNode function, JexlN
...
          */
         public static JexlNode distributeFunctionIntoComponent(JexlNode function, JexlNode component) {
             List<JexlNode> children = new LinkedList<>();
             String field = findCommonField(component);
             if (field != null) {
                 children.add(updateFunctionWithField(function, field));
-                children.addAll(Arrays.asList(JexlNodes.children(component)));
+                if (component instanceof ASTEQNode) {
+                    children.add(component);
+                } else {
```
FineAndDandy: `else if { ASTAndNode } else { //warn|error}` ?

FineAndDandy (changes requested):

jwomeara:
What issue is this tied to?

warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionsDescriptor.java
jwomeara: i agree

warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionsDescriptor.java
```diff
@@ -424,9 +424,12 @@ public static JexlNode distributeFunctionIntoIndexQuery(JexlNode function, JexlN
             List<JexlNode> components = new LinkedList<>();
             
             JexlNode deref = JexlASTHelper.dereference(indexQuery);
-            if (deref instanceof ASTAndNode) {
-                // single component
+            if (deref instanceof ASTAndNode || deref instanceof ASTEQNode) {
+                
+                // it is possible that an index query contains only a single equality node
+                // i.e., "phrase('foo', 'foo')" would produce a single equality node "FOO == 'bar'"
```
jwomeara: Is this comment correct?  `phrase(foo, foo)` becomes `FOO == 'bar'` ?

warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionsDescriptor.java
apmoriarty: Added in next commit

apmoriarty:
> What issue is this tied to?

Wrote up issue https://github.com/NationalSecurityAgency/datawave/issues/1484

warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionsDescriptor.java
apmoriarty: I'll make this language more precise. The index expansion for that function will produce a single equality node.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionsDescriptor.java
```diff
@@ -424,9 +424,12 @@ public static JexlNode distributeFunctionIntoIndexQuery(JexlNode function, JexlN
...
                 components.add(deref);
+                
             } else if (deref instanceof ASTOrNode) {
                 // multiple components
                 for (JexlNode child : JexlNodes.children(deref)) {
```
jwomeara, line 435: On line 439 theres an exception whose message should be updated to include EQ nodes.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionsDescriptor.java
jwomeara: my question was more about it going from foo, foo, to foo, bar.  is that right?

jwomeara (changes requested):
See comments.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionsDescriptor.java
apmoriarty: As written, that is incorrect. I updated the comment to be " `phrase(termOffsetMap,'bar','bar'` produces index query `"FOO == 'bar'"` "

warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionsDescriptor.java
apmoriarty: ahh, good catch

jwomeara (approved):

FineAndDandy (approved):

Issue #1485, opened 2022-04-04
Use ArrayList instead of LinkedHashSet in DatawavePrincipal, maintain chronological order


billoley:
DatawavePrincipal currently maintains the list of users as a LinkedHashSet which hides any duplcate users or call chain cycles from our logs and monitoring. Change the LinkedHashSet to an ArrayList to maintain a more accurate representation and record of the calling and proxied users.

Currently, the collection retrieved by DatawavePrincipal.getProxiedUsers (which is all of the users) has the final caller as the first entry followed by the entities in the call X-ProxiedEntitiesChain in the order that the calls occurred. This creates some minor complexity in finding the primaryUser and especially in explaining this order to others.

Modify the places where DatawavePrincipal is created such that the proxied entities are first (already in chronological order) and the final caller is last ..... making it also in chronological order

billoley:
Merged in https://github.com/NationalSecurityAgency/datawave/pull/1486

PR #1486, opened 2022-04-04
re #1485 Use ArrayList instead of LinkedHashSet in DatawavePrincipal, maintain chronological order

billoley:
Backward compatibility notes relating to microservices:

Datawave calls many microservice operation via Remote*Service with a JWT.  Both datawave-spring-boot-starter <= 1.6.5 (in use) and 1.6.7 select the primaryUser as the first USER if present or the first entity in proxiedUsers.... so no change when sending JWTs to a microservice.  

RemoteDatawaveUserService.lookup constructs it's X-ProxiedEntitiesChain/X-ProxiedIssuersChain in the order that it gets from DatawaveCredential which it then passes on to DatawavePrincipal.  Any ordering issues are internal to Datawave.

DatawavePrincipal currently maintains the list of users as a LinkedHashSet which hides any duplcate users or call chain cycles from our logs and monitoring. Change the LinkedHashSet to an ArrayList to maintain a more accurate representation and record of the calling and proxied users.

Currently, the collection retrieved by DatawavePrincipal.getProxiedUsers (which is all of the users) has the final caller as the first entry followed by the entities in the call X-ProxiedEntitiesChain in the order that the calls occurred. This creates some minor complexity in finding the primaryUser and especially in explaining this order to others.

Modify the places where DatawavePrincipal is created such that the proxied entities are first (already in chronological order) and the final caller is last ..... making it also in chronological order


ivakegg (dismissed):

keith-ratcliffe (dismissed):

jwomeara (dismissed):
Hmmmm, ok.

jwomeara (approved):
Hmmmmmmmmmmmmm, still ok.

keith-ratcliffe (approved):

ivakegg (approved):

billoley:
Integration tested

PR #1488, opened 2022-04-05
Add visitor to prune IsNotNull terms when possible.

apmoriarty:
The issue as written says to prune IsNotNull terms when there's another inclusive node. The strictest definition of this is 'is not null' and an equality node, but this code also attempts to look for regex equals nodes. Bounded ranges and other marker nodes are not considered.


warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IsNotNullPruningVisitorTest.java
```diff
@@ -0,0 +1,112 @@
...
+import datawave.query.jexl.visitors.validate.ASTValidator;
+import org.apache.commons.jexl2.parser.ASTJexlScript;
+import org.apache.commons.jexl2.parser.ParseException;
+import org.junit.Test;
+
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public class IsNotNullPruningVisitorTest {
+
```
avgAGB, line 14: More of a question for my curiosity, but does it make sense to have a test case to see what occurs if the IsNotNull queries are the only ones in the string, to see if the scenario is handled as would be expected?

apmoriarty: Sounds worthwhile to me, I'll add a few more test cases to handle what you describe

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/JexlNodes.java
```diff
@@ -252,4 +252,31 @@ public static boolean removeFromParent(JexlNode parent, JexlNode childToRemove)
...
     }
+    
+    /**
+     * Ascends the Jexl tree searching for a negation. Search defaults to four nodes
+     *
+     * @param node
+     * @return
+     */
+    public static boolean isNodeNegated(JexlNode node) {
+        return isNodeNegated(node, 4);
```
jwomeara: Can we please avoid hard coding the number of nodes to ascend in the tree?  This is not a very robust approach, and assumptions made about the parentage of a node often come back to bite us.  If the intent here is just to see if there is a nearby or immediate negation, could I suggest ascending until you either reach a parent AND, OR or root node and stopping there instead?

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/JexlNodes.java
```diff
@@ -252,4 +252,31 @@ public static boolean removeFromParent(JexlNode parent, JexlNode childToRemove)
...
+    }
+    
+    /**
+     * Ascends the JexlTree searching for negations according to the specified maximum number of
+     *
+     * @param node
+     * @return
+     */
+    public static boolean isNodeNegated(JexlNode node, int max) {
```
jwomeara: Similar comment here with allowing a 'max' parameter.  This has bit us in the past.  Maybe consider allowing an oldest ancestor to be passed, and stopping there at a known point?  Setting a numerical limit leaves too much to chance IMO.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IsNotNullPruningVisitorTest.java
```diff
@@ -0,0 +1,173 @@
...
+import datawave.query.jexl.JexlASTHelper;
+import datawave.query.jexl.visitors.validate.ASTValidator;
+import org.apache.commons.jexl2.parser.ASTJexlScript;
+import org.apache.commons.jexl2.parser.ParseException;
+import org.junit.Test;
+
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public class IsNotNullPruningVisitorTest {
```
jwomeara, line 13: Can you add a test case for the following to ensure that the visitor doesn't act in this case:
`!(FOO == null) && (FOO == 'bar' || FOO2 == 'baz')`

jwomeara (changes requested):
See comments.

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/JexlNodes.java
apmoriarty: I'll remove this.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IsNotNullPruningVisitorTest.java
apmoriarty: Added this to the 'no op' test case and it passes just fine.

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/JexlNodes.java
apmoriarty: Fair point. I'm refactoring the method so it ascends the entire tree looking for an `ASTNotNode`, and making that clear in the comment. At this point the query tree has been flattened, so we shouldn't be in danger of a stack overflow even with relatively complex queries. 

Any thoughts as to the method name? I realize there's a private static `isNodeNegated` method in the `JexlASTHelper` that determines if a node is a negated based on *every* negation in the ancestry....I only care about if the node is under *at least one* negation.

jwomeara: Since we're traversing the parent nodes iteratively, and not recursively, I don't think there's a risk of a stack overflow here.

As far as the name, I'm not sure.  The method attempts to find a negated parent, so maybe 'findNegatedParent'?

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IsNotNullPruningVisitorTest.java
jwomeara: I figured it would.  This is to make sure that if someone changes things in the future that they don't break this contract.

jwomeara (dismissed):
Comments look good to me.  Happy to reapprove if you decide to rename the isNodeNegated method.

jwomeara (approved):

avgAGB (approved):

PR #1489, opened 2022-04-05
Adding new role for max query result override

avgAGB:


web-services/query/src/main/java/datawave/webservice/query/runner/BasicQueryBean.java
```diff
@@ -69,6 +69,7 @@
 public class BasicQueryBean {
     
     private static final String PRIVILEGED_USER = "PrivilegedUser";
+    private static final String UNLIMITED_QUERY_RESULTS_USER = "UnlimitedQueryResultsUser";
```
jwomeara: FWIW, I don't think anyone uses this query wizard stuff.  It's not going to be carried forward to the query microservices.

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -168,8 +168,8 @@
 import static datawave.webservice.query.cache.QueryTraceCache.PatternWrapper;
 
 @Path("/Query")
-@RolesAllowed({"AuthorizedUser", "AuthorizedQueryServer", "PrivilegedUser", "InternalUser", "Administrator", "JBossAdministrator"})
-@DeclareRoles({"AuthorizedUser", "AuthorizedQueryServer", "PrivilegedUser", "InternalUser", "Administrator", "JBossAdministrator"})
+@RolesAllowed({"AuthorizedUser", "AuthorizedQueryServer", "PrivilegedUser", "UnlimitedQueryResultsUser", "InternalUser", "Administrator", "JBossAdministrator"})
+@DeclareRoles({"AuthorizedUser", "AuthorizedQueryServer", "PrivilegedUser", "UnlimitedQueryResultsUser", "InternalUser", "Administrator", "JBossAdministrator"})
```
jwomeara: Do we really want to add the UnlimitedQueryResultsUser role here?  This is an OR'd list of users allowed to call this method.  By doing this, a user wouldn't have to be an AuthorizedUser to run queries.

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -512,9 +513,9 @@ private QueryData validateQuery(String queryLogicName, MultivaluedMap<String,Str
         }
         
         // validate the max results override relative to the max results on a query logic
-        // privileged users however can set whatever they want
+        // privileged users however can set whatever they want, or unlimited max results users.
         if (qp.isMaxResultsOverridden() && qd.logic.getMaxResults() >= 0) {
-            if (!ctx.isCallerInRole(PRIVILEGED_USER)) {
+            if (!ctx.isCallerInRole(PRIVILEGED_USER) || !ctx.isCallerInRole(UNLIMITED_QUERY_RESULTS_USER)) {
```
jwomeara, line 518: I think this line is all you need to support having an unlimited query results role.  Also, it would be nice if this could be configurable, but if that's not something you think you'll need, then disregard.

jwomeara (changes requested):
Looks good aside from the role additions to the QueryExecutorBean.  I realize that this is WIP, so I apologize if you were already intending to make some of these changes.

web-services/query/src/main/java/datawave/webservice/query/runner/BasicQueryBean.java
avgAGB: Okay, will remove

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
avgAGB: Are PrivilegedUser/UnlimitedQueryResultsUser permutations of the AuthorizedUser by default? If it doesn't need to be added to the RolesAllowed would that also mean that PrivilegedUser could be removed from this list as well?

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
avgAGB: I don't think I need it in this case, but I'll keep it in mind when I follow up with the people that requested the change.

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
jwomeara: I don't think it's possible to have a PrivilegedUser that is not an AuthorizedUser, so PrivilegedUser probably doesn't need to be in this list.  The same probably goes for Administrator and JBossAdministrator to be honest.

avgAGB: Got it. Updated the beans to remove the extra roles

alerman (dismissed):

jwomeara (dismissed):

jwomeara (approved):
Nice work!

alerman (approved):

PR #1491, opened 2022-04-06
Issue1445 - Support injection of Scheduler choice and remove 'check for test mode' from PushdownScheduler

tomnelson:
Makes the choice of Pushdown or Sequential Scheduler configurable in the QueryLogic.
The 'test' version of the PushdownScheduler is configured to use InMemoryAccumulo so that the actual production class no longer needs to check whether it is in a 'test' mode or not.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
+++ b/warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
@@ -93,7 +93,6 @@ public class ShardQueryConfiguration extends GenericQueryConfiguration implement
     private int collapseUidsThreshold = -1;
     // Should this query dedupe terms within ANDs and ORs
     private boolean enforceUniqueTermsWithinExpressions = false;
-    private boolean sequentialScheduler = false;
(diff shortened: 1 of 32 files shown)


tomnelson:
what does the no_update label mean for me?

keith-ratcliffe:
> what does the no_update label mean for me?

@alerman was likely testing out the label-based filtering associated with the new `autoupdate` GitHub Action defined [here](https://github.com/NationalSecurityAgency/datawave/blame/integration/.github/workflows/autoupdate.yml#L15) , committed the same day the label was added to this PR

Can't think of a reason we'd want to keep it labeled as such, unless merge conflicts are problematic for `autoupdate` for whatever reason(?)

@alerman?

keith-ratcliffe:
@tomnelson once the merge conflicts are resolved, I'll approve

tomnelson:
Is the only conflict is the meaningless 'expectedObjectCount' magic number?

hlgp:
The SequentialScheduler was removed in https://github.com/NationalSecurityAgency/datawave/pull/2631
If we ever add it back, this can be reopened.

PR #1497, opened 2022-04-09
fixes the ingest test

tomnelson:
I think this fixes that failing ingest test (ShardedTableMapFileTest). I used the approach in an accumulo unit test.
There is more work to do in other modules.


tomnelson:
I fixed (perhaps) the ScannerSessionTest.
SharedCacheCoordinatorTest is still failing (if @Test is uncommented)

jwomeara:
I think we already fixed this on integration.

PR #1501, opened 2022-04-12
Bugfix/query pruning visitor

jwomeara:


apmoriarty (approved):

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -1383,28 +1383,14 @@ protected ASTJexlScript timedExecutableExpansion(QueryStopwatch timers, final AS
...
     
-    protected ASTJexlScript timedPushdownLowSelectiveTerms(QueryStopwatch timers, final ASTJexlScript script, ShardQueryConfiguration config,
-                    Set<String> indexedFields, Set<String> indexOnlyFields, Set<String> nonEventFields) {
+    protected ASTJexlScript timedPushdownLowSelectiveTerms(QueryStopwatch timers, ShardQueryConfiguration config) {
         TraceStopwatch stopwatch = timers.newStartedStopwatch("DefaultQueryPlanner - Pushdown Low-Selective Terms");
         
         config.setQueryTree(PushdownLowSelectivityNodesVisitor.pushdownLowSelectiveTerms(config.getQueryTree(), config, metadataHelper));
         if (log.isDebugEnabled()) {
             logQuery(config.getQueryTree(), "Query after pushing down low-selective terms:");
         }
```
FineAndDandy, line 1393: are we guaranteed now to get an executable tree from the visitor? Why do we no longer need to pushdownPredicates?

ivakegg: This is the wrong place for this code.  We do the executability pushdown/pullup later.  The fact that this delayed an entire tree of geo terms is why the pruning visitor missed a ton of pruning.

ivakegg (approved):

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
jwomeara: To answer a question you didn't exactly ask... on the following lines when we check executability and run the pushdown visitor, THAT visitor is NOT guaranteed to produce an executable tree.  Because of that, SOMETIMES a geo query is able to make it through this code unchanged - if the entire query is unexecutable, it doesn't bother pushing anything down.  So, not producing an executable tree at this point is actually beneficial for geo queries.

FineAndDandy: @ivakegg That was part of the concern, as long as it is still guaranteed to run later

@jwomeara got it. That was the other part of the question I meant to ask :)

FineAndDandy (approved):

Issue #1503, opened 2022-04-13
Partial Document Evaluation


apmoriarty:
There exist certain cases where a document cannot be fully evaluated on the tablet server. In these cases a partial evaluation must be done and a final document evaluation occurs in the webservice.

PR #1504, opened 2022-04-13
feature/PartialEvaluation

apmoriarty:
This is a draft merge request for allowing queries to perform a partial evaluation on the tablet server and a final evaluation in the webservice. 

See https://github.com/NationalSecurityAgency/datawave/issues/1503 for details on the motivation for this feature. 

High level changes
- documents now have a field for termOffsetMaps
- termOffsetMaps and related classes are now serialized
- document post-processing on the query iterator can be disabled
- webservice implementations now exist for query iterator document post-processing
- documents can be evaluated with the DatawavePartialInterpreter on the tablet server


warehouse/query-core/src/main/java/datawave/query/attributes/Document.java
```diff
@@ -855,4 +917,35 @@ public boolean isIntermediateResult() {
...
     
+    private byte[] getOffsetMapToBytes() {
+        try (ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
+            try (ObjectOutputStream oos = new ObjectOutputStream(baos)) {
+                oos.writeObject(offsetMap);
+                oos.flush();
+                return baos.toByteArray();
+            }
+        } catch (Exception e) {
+            log.error("something went wrong trying to serialize the offset map to a byte array");
```
FineAndDandy: don't swallow this exception

warehouse/query-core/src/main/java/datawave/query/attributes/Document.java
```diff
@@ -855,4 +917,35 @@ public boolean isIntermediateResult() {
...
+        return null;
+    }
+    
+    private TermOffsetMap getOffsetMapFromBytes(byte[] data) {
+        try (ByteArrayInputStream bais = new ByteArrayInputStream(data)) {
+            try (ObjectInputStream ois = new ObjectInputStream(bais)) {
+                return (TermOffsetMap) ois.readObject();
+            }
+        } catch (Exception e) {
+            log.error("failed to deserialize offset map from byte array");
```
FineAndDandy: don't swallow this exception

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
```diff
@@ -283,6 +284,12 @@ public class ShardQueryConfiguration extends GenericQueryConfiguration implement
      */
     private boolean speculativeScanning = false;
     private boolean disableEvaluation = false;
+    
+    // a partial interpreter will evaluate 'incomplete fields' as true when more work is required for a full evaluation
```
FineAndDandy, line 298: true here depends on context? Presumably returns false when inside a negation?

warehouse/query-core/src/main/java/datawave/query/jexl/ArithmeticJexlEngines.java
```diff
@@ -51,6 +76,26 @@ public static DatawaveJexlEngine getEngine(JexlArithmetic arithmetic) {
...
+     * @return a JexlEngine
+     */
+    public static DatawaveJexlEngine getEngine(JexlArithmetic arithmetic, Set<String> incompleteFields) {
+        if (arithmetic == null) {
+            return null;
+        }
+        
+        DatawaveJexlEngine engine = createEngine(arithmetic);
+        engine.setUsePartialInterpreter(true);
+        engine.setIncompleteFields(incompleteFields);
```
FineAndDandy, line 97: should this be configurable or only when incomplete fields are specified?

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -1002,6 +1033,71 @@ protected void loadQueryParameters(ShardQueryConfiguration config, Query setting
...
+        try {
+            if (scannerFactory.config.getConnector() != null) {
+                MetadataHelper metadataHelper = prepareMetadataHelper(scannerFactory.config.getConnector(), getConfig().getMetadataTableName(), getConfig()
+                                .getAuthorizations());
+                CompositeMetadata compositeMetadata = metadataHelper.getCompositeMetadata().filter(config.getQueryFieldsDatatypes().keySet());
+                Collection<Multimap<String,String>> compositeMaps = compositeMetadata.getCompositeFieldMapByType().values();
+                documentEvaluation.addFunction(new CompositeProjectionFunction(includeGroupingContext, isReducedResponse(), compositeMaps));
+            }
+        } catch (TableNotFoundException e) {
+            log.error("Could not add CompositeProjectionTransform, error was: ", e);
```
FineAndDandy: throw this

warehouse/query-core/src/test/java/datawave/query/CompositeFunctionsTest.java
```diff
@@ -622,6 +629,7 @@ public void testWithLucene() throws Exception {
                 Arrays.asList("CORLEONE", "CAPONE"), // family has child MICHAEL
                 Collections.singletonList("CORLEONE"), Collections.singletonList("CORLEONE")};
         for (int i = 0; i < queryStrings.length; i++) {
+            System.out.println("Running query for index: " + i);
```
FineAndDandy: make log statement or remove

warehouse/query-core/src/test/java/datawave/query/LongRunningQueryTest.java
```diff
@@ -151,6 +152,7 @@ public void testLongRunningGroupByQuery() throws Exception {
      * 
      * @throws Exception
      */
+    @Ignore
```
FineAndDandy: this part of WIP?

warehouse/query-core/src/main/java/datawave/query/jexl/DatawavePartialInterpreter.java
```diff
@@ -0,0 +1,887 @@
...
+ * </ul>
+ * <h2>Evaluation Logic (Unions)</h2>
+ * <ul>
+ * <li>(TRUE or TRUE) evaluates to TRUE and is a FULL evaluation</li>
+ * <li>(TRUE or FALSE) evaluates to TRUE and is a FULL evaluation</li>
+ * <li>(FALSE or FALSE) evaluates to FALSE and is a FULL evaluation</li>
+ * <li>(TRUE or UNKNOWN) evaluates to TRUE and is a FULL evaluation</li>
+ * <li>(FALSE or UNKNOWN) evaluates to TRUE and is a PARTIAL evaluation</li>
+ * <li>(UNKNOWN or UNKNOWN) evaluates to TRUE and is a PARTIAL evaluation</li>
+ * </ul>
```
FineAndDandy, line 78: can you include negations in this as well?

warehouse/query-core/src/main/java/datawave/query/jexl/DatawavePartialInterpreter.java
```diff
@@ -0,0 +1,887 @@
...
+     *
+     * @param node
+     *            the JexlNode
+     * @return true if incomplete fieldname found, false otherwise
+     */
+    private boolean isIncomplete(JexlNode node) {
+        for (String field : JexlASTHelper.getIdentifierNames(node)) {
+            field = JexlASTHelper.deconstructIdentifier(field);
+            // if the field is not in the context, we cannot evaluate it.
+            return incompleteFields.contains(field) && context.has(field);
```
FineAndDandy: is it possible that the context wouldn't have the field?

warehouse/query-core/src/main/java/datawave/query/jexl/DatawavePartialInterpreter.java
```diff
@@ -0,0 +1,887 @@
...
+                } else if (state.hasNumeric()) {
+                    right = state.getNumeric();
+                } else if (state.getSet() != null) {
+                    right = state.getSet().size();
+                }
+            }
+            
+            if (arithmetic.lessThan(left, right)) {
+                if (state != null) {
+                    return state;
```
FineAndDandy: should this have its set and numeric cleared?

warehouse/query-core/src/main/java/datawave/query/jexl/DatawavePartialInterpreter.java
```diff
@@ -0,0 +1,887 @@
...
+                } else if (state.hasNumeric()) {
+                    right = state.getNumeric();
+                } else if (state.getSet() != null) {
+                    right = state.getSet().size();
+                }
+            }
+            
+            if (arithmetic.lessThanOrEqual(left, right)) {
+                if (state != null) {
+                    return state;
```
FineAndDandy: same question about clearing state as below

warehouse/query-core/src/main/java/datawave/query/jexl/DatawavePartialInterpreter.java
```diff
@@ -0,0 +1,887 @@
...
+                } else if (state.hasNumeric()) {
+                    right = state.getNumeric();
+                } else if (state.getSet() != null) {
+                    right = state.getSet().size();
+                }
+            }
+            
+            if (arithmetic.greaterThan(left, right)) {
+                if (state != null) {
+                    return state;
```
FineAndDandy: same question

warehouse/query-core/src/main/java/datawave/query/jexl/DatawavePartialInterpreter.java
```diff
@@ -0,0 +1,887 @@
...
+                throw new IllegalStateException("expected STATE but was " + o.getClass().getSimpleName());
+            }
+        }
+        
+        if (!functionalSet.isEmpty()) {
+            return new State(true, anyFromUnknown, functionalSet);
+        } else {
+            return new State(true, anyFromUnknown);
+        }
+    }
```
FineAndDandy, line 559: should these be `new State(o.matched, ...)`?

warehouse/query-core/src/main/java/datawave/query/jexl/DatawavePartialInterpreter.java
```diff
@@ -0,0 +1,887 @@
...
+            leftState.mergeOr(rightState);
+            return leftState;
+        } else if (leftState != null) {
+            return leftState;
+        } else if (rightState != null) {
+            return rightState;
+        } else {
+            // if we got here then we are likely dealing with a function that takes a union of null fields as arguments
+            // for example, filter:notNull(NULL1 || NULL2). return false with an initialized empty functional set
+            return new State(true);
```
FineAndDandy: This would require both states to be null, but null or null is surely null?

warehouse/query-core/src/main/java/datawave/query/jexl/DatawavePartialInterpreter.java
```diff
@@ -0,0 +1,887 @@
...
+            }
+        } else {
+            
+            // We are likely within a normal union and can short circuit
+            while (!children.isEmpty()) {
+                // Child nodes were put onto the stack left to right. PollLast to evaluate left to right.
+                result = (State) interpretOr(children.pollLast().jjtAccept(this, data), result);
+                
+                if (result.matched && !result.fromUnknown) {
+                    break;
```
FineAndDandy: if even one component of the children is fromUnknown this will break won't it? You may not have processed the child that has this state yet...

warehouse/query-core/src/main/java/datawave/query/jexl/DatawavePartialInterpreter.java
```diff
@@ -0,0 +1,887 @@
...
+        }
+        return getState(node, result);
+    }
+    
+    @Override
+    public Object visit(ASTEQNode node, Object data) {
+        String nodeString = JexlStringBuildingVisitor.buildQueryWithoutParse(node);
+        Object result = resultMap.get(nodeString);
+        if (null != result)
+            return new State(true); // TODO -- this might be from an unknown field
```
FineAndDandy: Maybe check instanceof State and cast, or otherwise pass along all relevant State to the new object

warehouse/query-core/src/main/java/datawave/query/jexl/DatawavePartialInterpreter.java
```diff
@@ -0,0 +1,887 @@
...
+                    left = state.getNumeric();
+                }
+            }
+            
+            result = arithmetic.equals(left, right) ? Boolean.TRUE : Boolean.FALSE;
+        } catch (ArithmeticException xrt) {
+            throw new JexlException(node, "== error", xrt);
+        }
+        
+        resultMap.put(nodeString, result);
```
FineAndDandy: store State into the resultsMap

warehouse/query-core/src/main/java/datawave/query/jexl/DatawavePartialInterpreter.java
```diff
@@ -0,0 +1,887 @@
...
+        // get the namespace
+        FunctionJexlNodeVisitor visitor = new FunctionJexlNodeVisitor();
+        visitor.visit(node, null);
+        
+        // content functions should always have the field as the first argument
+        if (visitor.namespace().equals(ContentFunctions.CONTENT_FUNCTION_NAMESPACE)) {
+            JexlNode first = visitor.args().get(0);
+            if (first instanceof ASTIdentifier) {
+                String field = JexlASTHelper.deconstructIdentifier((ASTIdentifier) first);
+                if (context.has(field) && incompleteFields.contains(field)) {
```
FineAndDandy: does it matter if the context has the field? Is it enough that the incomplete fields contains it?

warehouse/query-core/src/main/java/datawave/query/jexl/DatawavePartialInterpreter.java
```diff
@@ -0,0 +1,887 @@
...
+                    return new State(true, true);
+                }
+            }
+        } else {
+            // fall back to descriptor for non-content functions
+            JexlArgumentDescriptor descriptor = JexlFunctionArgumentDescriptorFactory.F.getArgumentDescriptor(node);
+            Set<String> fields = descriptor.fields(null, null);
+            
+            for (String field : fields) {
+                if (context.has(field) && incompleteFields.contains(field)) {
```
FineAndDandy: same question as above

warehouse/query-core/src/main/java/datawave/query/jexl/DatawavePartialInterpreter.java
```diff
@@ -0,0 +1,887 @@
...
+    }
+    
+    /**
+     * Used to track matched state on a per-node basis.
+     */
+    public class State {
+        boolean matched;
+        boolean fromUnknown;
+        FunctionalSet set;
+        Object numeric;
```
FineAndDandy: could these be combined to just Object value? Then provide convenience methods to getValue() which attempt to cast to Numeric and FunctionalSet, String, and Boolean? Might make this cleaner in the rest of the impl. It can be a bit messy to follow all the logic.

FineAndDandy (changes requested):
this is looking great so far

warehouse/query-core/src/main/java/datawave/query/attributes/Document.java
apmoriarty: fixed in next commit

warehouse/query-core/src/main/java/datawave/query/attributes/Document.java
apmoriarty: fixed in next commit

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
apmoriarty: fixed in next commit

warehouse/query-core/src/test/java/datawave/query/CompositeFunctionsTest.java
apmoriarty: that's left over from testing, it will be removed in the next commit

warehouse/query-core/src/main/java/datawave/query/jexl/ArithmeticJexlEngines.java
apmoriarty: The code is correct. This constructor is only called if you are running a partial evaluation incomplete fields.

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
apmoriarty: An incomplete field is always evaluated as true on the tablet server. The partial interpreter knows to interpret a result from an incomplete field as true, so `!(INCOMPLETE_FIELD == 'value')` evaluates to true.

warehouse/query-core/src/test/java/datawave/query/LongRunningQueryTest.java
apmoriarty: This was from an older iteration, it's removed in the next commit.

warehouse/query-core/src/main/java/datawave/query/jexl/DatawavePartialInterpreter.java
apmoriarty: Sure thing

warehouse/query-core/src/main/java/datawave/query/jexl/DatawavePartialInterpreter.java
apmoriarty: Yes, it is possible. This composite check of both the incompleteFields and the context allows us to appropriately defeat more documents on the tablet server. Consider the following case;

A query like `FOO == 'anchor' && filter:includeRegex(INCOMPLETE, 'value.*')` means we could have a document of just `{FOO=anchor}`. In this case the absence of an incomplete field in the document means it will never evaluate to true, so we shouldn't even bother returning it to the webservice for a full evaluation.

The presence of the field in both the query and the set of incomplete fields is not enough to determine the state of the node in the interpreter.

apmoriarty: Another possible case would be `FIELD_A && (FIELD_B || INCOMPLETE)`. In this case of a document like `{FIELD_A, FIELD_B}` we can fully evaluate the query without having to run a secondary evaluation in the webservice. This is only possible if we check the context for the incomplete field.

warehouse/query-core/src/main/java/datawave/query/jexl/DatawavePartialInterpreter.java
apmoriarty: Yes, the field's presence in the context matters. See comment above.

warehouse/query-core/src/main/java/datawave/query/jexl/DatawavePartialInterpreter.java
apmoriarty: Yes, see above.

warehouse/query-core/src/main/java/datawave/query/jexl/DatawavePartialInterpreter.java
apmoriarty: No. At this point in the method the intersection has evaluated to true. The only remaining questions are a) is there a functional set of hits and b) was any part of this from an unknown/incomplete field

warehouse/query-core/src/main/java/datawave/query/jexl/DatawavePartialInterpreter.java
apmoriarty: I don't think State objects belong in the resultMap, but I could be persuaded otherwise. 

The unspoken contract for the partial interpreter is to make a decision based on the field and the result of a normal interpreter visit.

warehouse/query-core/src/main/java/datawave/query/jexl/DatawavePartialInterpreter.java
apmoriarty: See above comment

warehouse/query-core/src/main/java/datawave/query/jexl/DatawavePartialInterpreter.java
apmoriarty: This behavior is intended. If the choice exists between `(FIELD || INCOMPLETE_FIELD)` then the interpreter uses the least-intensive option.

warehouse/query-core/src/main/java/datawave/query/jexl/DatawavePartialInterpreter.java
apmoriarty: This code is a bit old, it returns a false state.

apmoriarty:
OBE..for now

PR #1508, opened 2022-04-13
Datawave 707

### LINKED ISSUE #707: result uniqueness (#UNIQUE) should make use of intermediate results
The a unique filter is used in the QueryIterator, this can result is long delays between returned results.  This results is scans that eat up readahead slots.  The accumulo yielding mechanism used in the QueryIterator should allow us to yield in this case.

cmitchell:
This ticket implemented the same intermediate results functionality that was implemented for DATAWAVE-1400 and builds off it, but this time for unique queries. Just like in that ticket, the following was done:

* Separated the UniqueTransform into `UniqueTransform` and `UniqueIterator` to make it explicitly clear when the code was executing on the client (transform) vs the server (iterator)
* Moved the common code from UniqueTransform into `UniqueUtil`
* Updated existing unique query test setups (but did not need to modify the actual tests)
* Added a `testLongRunningUniqueQuery` to the `LongRunningQueryTest`


warehouse/query-core/src/main/java/datawave/query/iterator/UniqueIterator.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Map;
+
+public class UniqueIterator {
+    
+    private static final Logger log = Logger.getLogger(UniqueTransform.class);
+    
+    private UniqueFields uniqueFields;
+    private UniqueUtil uniqueUtil = new UniqueUtil();
+    
+    // private Multimap<String,String> modelMapping;
```
apmoriarty: unused comment?

warehouse/query-core/src/main/java/datawave/query/iterator/UniqueIterator.java
```diff
@@ -0,0 +1,54 @@
...
+import datawave.query.iterator.profile.FinalDocumentTrackingIterator;
+import datawave.query.transformer.UniqueTransform;
+import org.apache.accumulo.core.data.Key;
+import org.apache.log4j.Logger;
+
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.util.Map;
+
+public class UniqueIterator {
```
apmoriarty, line 20: Add a short description, even if it's a simple as "delegates unique operations to UniqueUtil"

warehouse/query-core/src/test/java/datawave/query/LongRunningQueryTest.java
```diff
@@ -143,15 +201,16 @@ public void testAllowLongRunningQueryWithShardQueryLogic() throws Exception {
     }
     
     /**
+     * ALWAYS ENSURE THAT HTE NAME OF THIS TEST IS SUCH THAT IT COMES ALPHABETICALLY FIRST IN THIS CLASS.
```
apmoriarty:
```suggestion
     * ALWAYS ENSURE THAT THE NAME OF THIS TEST IS SUCH THAT IT COMES ALPHABETICALLY FIRST IN THIS CLASS.
```

warehouse/query-core/src/test/java/datawave/query/transformer/UniqueTransformTest.java
```diff
@@ -507,16 +503,18 @@ private void assertUniqueDocuments() {
...
+        return docs.stream().map(d -> uniqueTransform.flush()).filter(Objects::nonNull).map(d -> d.getValue()).collect(Collectors.toList());
     }
     
     private void assertOrderedFieldSets() {
-        UniqueTransform uniqueTransform = getUniqueTransform();
-        List<UniqueTransform.FieldSet> actual = inputDocuments.stream().map(uniqueTransform::getOrderedFieldSets).flatMap(List::stream).sorted()
-                        .collect(Collectors.toList());
+        UniqueUtil uniqueUtil = new UniqueUtil();
+        List<UniqueUtil.FieldSet> actual = inputDocuments.stream().map(doc -> uniqueUtil.getOrderedFieldSets(doc, uniqueFields, null)) // .map(uniqueUtil::getOrderedFieldSets)
+                        .flatMap(List::stream).sorted().collect(Collectors.toList());
```
apmoriarty: remove comment?

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -42,13 +26,17 @@ public class UniqueTransform extends DocumentTransform.DefaultDocumentTransform
...
     private Multimap<String,String> modelMapping;
+    private LinkedList<Entry<Key,Document>> documents;
     
-    public UniqueTransform(UniqueFields uniqueFields) {
-        this(null, uniqueFields);
-    }
+    /**
+     * how long (in milliseconds) to let a page of results to collect before signaling to return a blank page to the client (which indicates the request is
+     * still in process and all results will be returned at once at the end)
+     */
```
apmoriarty: I'm not in love with this wording, but I also don't have a great alternative. Took a stab below, feel free to improve it or just ignore this comment.

"Length in milliseconds that a client will wait while results are collected. If a full page is not collected before the timeout, a blank page will be returned to signal the request is still in progress."

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -97,261 +78,33 @@ public Predicate<Entry<Key,Document>> getUniquePredicate() {
...
-                Set<String> values = getValues(document.get(documentField));
-                Set<String> transformedValues = uniqueFields.transformValues(field, values);
-                Multimap<String,String> groupedValues = mapGroupingContextToField.get(groupingContext);
-                if (groupedValues == null) {
-                    groupedValues = HashMultimap.create();
-                    mapGroupingContextToField.put(groupingContext, groupedValues);
-                }
-                groupedValues.putAll(field, transformedValues);
-            }
+    @Override
```
apmoriarty: Add method doc

apmoriarty (changes requested):

warehouse/query-core/src/main/java/datawave/query/iterator/UniqueIterator.java
cmitchell: Removed.

warehouse/query-core/src/main/java/datawave/query/iterator/UniqueIterator.java
cmitchell: Added class level javadoc

warehouse/query-core/src/test/java/datawave/query/LongRunningQueryTest.java
cmitchell: Fixed

warehouse/query-core/src/test/java/datawave/query/transformer/UniqueTransformTest.java
cmitchell: Removed.  Also re-named this test class to be `UniqueIteratorTest` since it's testing the UniqueIterator, not the UniqueTransform.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
cmitchell: I took your suggestion and just changed it to "Length of time....". I also changed the comment in GroupingTransform since it was the same wording for the same variable declaration.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
cmitchell: Added

apmoriarty (dismissed):
LGTM

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -79,279 +69,50 @@ public void updateConfig(UniqueFields uniqueFields, QueryModel model) {
...
+    public Entry<Key,Document> flush() {
+        long elapsedExecutionTimeForCurrentPage = System.currentTimeMillis() - this.queryExecutionForPageStartTime;
+        if (elapsedExecutionTimeForCurrentPage > this.queryExecutionForPageTimeout) {
```
ivakegg: We should only ever return "intermediate results" if there are potentially more results to return.  How do we know if the underlying scanners are finished or not?  I am missing something here.

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
```diff
@@ -201,74 +206,76 @@ public ResultsPage next() throws Exception {
...
-                    hitPageByteTrigger = true;
-                    break;
-                }
-                // if the logic had a max num results (across all pages) and we have reached that (or the maxResultsOverride if set), then break out
-                if (this.settings.isMaxResultsOverridden()) {
-                    if (this.settings.getMaxResultsOverride() >= 0 && numResults >= this.settings.getMaxResultsOverride()) {
-                        log.info("Max results override has been reached, aborting query.next call");
+            hasNextFuture = executor.submit(() -> this.iter.hasNext());
+            try {
+                while ((!this.finished && (future != null)) || (Boolean) hasNextFuture.get(3000000, TimeUnit.MILLISECONDS)) {
```
ivakegg: do we really want 50 minutes hard coded here?

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
```diff
@@ -285,41 +292,42 @@ public ResultsPage next() throws Exception {
...
+                    testForUncaughtException(resultList.size());
+                    hasNextFuture = executor.submit(() -> this.iter.hasNext());
                 }
-                
-                testForUncaughtException(resultList.size());
-                
+            } catch (TimeoutException te) {
+                // This means the iter.hasNext() call didn't return within the allotted time. Manually return an
+                // intermediate result -- don't have a count here to know how many times we've returned an intermediate result...
+                hitIntermediateResult = true;
```
ivakegg: I think we only want to do this if "allowLongRunning" is true?

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
```diff
@@ -119,6 +119,14 @@ public Entry<Key,Document> apply(@Nullable Entry<Key,Document> keyDocumentEntry)
             this.countingMap = groupingInfo.getCountsMap();
```
ivakegg: Up above we may want to check for a FinalDocument before adding it into the fold.  See the UniqueTransform for the code into apply method.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
cmitchell: Ignoring this comment in favor of our in person conversation, and subsequent changes made.

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
cmitchell: Do we already have a configurable variable for this?  I was attempting to use the known system default. If that's something that can be injected and changed system wide, then I'll use the variable.

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
cmitchell: This will only get triggered by that check you were asking if we wanted hard-coded -- As I understood it, if hasNext hasn't returned in 50 minutes (regardless of whether or not this is a long running query) then we're hung.  Maybe this should just return no results instead of manually running an intermediate result since it has nothing to do with long running queries

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
ivakegg: See the QueryExpirationProperties

ivakegg: shortCircuitTimeout.  Access via the "timing" class.

cmitchell: Done

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
cmitchell: Done

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
cmitchell: Done

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -95,279 +69,51 @@ public void updateConfig(UniqueFields uniqueFields, QueryModel model) {
...
-            if (FinalDocumentTrackingIterator.isFinalDocumentKey(keyDocumentEntry.getKey())) {
-                return keyDocumentEntry;
-            }
-            
             try {
-                if (isDuplicate(keyDocumentEntry.getValue())) {
-                    keyDocumentEntry = null;
+                if (!FinalDocumentTrackingIterator.isFinalDocumentKey(keyDocumentEntry.getKey())
+                                && uniqueUtil.isDuplicate(keyDocumentEntry.getValue(), uniqueFields, modelMapping)) {
+                    return null;
```
ivakegg, line 87: I think we need to return the keyDocumentEntry here instead of null.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -95,279 +69,51 @@ public void updateConfig(UniqueFields uniqueFields, QueryModel model) {
...
                 }
             } catch (IOException ioe) {
                 log.error("Failed to convert document to bytes.  Returning document as unique.", ioe);
             }
+            documents.add(keyDocumentEntry);
```
ivakegg, line 92: I think we missed checking whether the keyDocumentEntry is a duplicate before adding it to the list.  Also we can simply return the document here if it is unique.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -1,62 +1,42 @@
...
  */
 public class UniqueTransform extends DocumentTransform.DefaultDocumentTransform {
     
     private static final Logger log = Logger.getLogger(UniqueTransform.class);
     
+    private final UniqueUtil uniqueUtil = new UniqueUtil();
     private BloomFilter<byte[]> bloom;
     private UniqueFields uniqueFields;
     private Multimap<String,String> modelMapping;
+    private LinkedList<Entry<Key,Document>> documents;
```
ivakegg, line 33: Why are we caching documents in this transform?  As long as the document is unique, we can simply pass it along.  This is not like the grouping paradigm where we need to keep all of the info before we can return anything.  In this case we can simply stream the results back.  The problem becomes towards the end of the query where we may have to go through many results before we find another unique one.

web-services/query/src/main/java/datawave/webservice/query/iterator/DatawaveTransformIterator.java
```diff
@@ -30,6 +32,17 @@ public DatawaveTransformIterator(Iterator iterator, Transformer transformer) {
     public boolean hasNext() {
         if (next == null) {
             next = getNext();
+            if (next != null) {
```
ivakegg: No need to check for 3 intermediate results here.  The RunningQuery takes care of counting how many empty result pages are returned.

web-services/query/src/main/java/datawave/webservice/query/logic/BaseQueryLogic.java
```diff
@@ -35,6 +35,7 @@
     protected Iterator<T> iterator = (Iterator<T>) Collections.emptyList().iterator();
     private int maxPageSize = 0;
     private long pageByteTrigger = 0;
+    private long pageProcessingStartTime = 0;
```
ivakegg: not used

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
```diff
@@ -214,6 +217,7 @@ public ResultsPage next() throws Exception {
         boolean hitPageByteTrigger = false;
         boolean hitPageTimeTrigger = false;
         boolean hitIntermediateResult = false;
+        boolean hitMaxIntermediateResult = false;
```
ivakegg: not used

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
```diff
@@ -110,7 +110,10 @@ public RunningQuery(QueryMetricsBean queryMetrics, Connector connection, Accumul
         this.connectionPriority = priority;
         this.settings = settings;
         this.calculatedAuths = AuthorizationsUtil.getDowngradedAuthorizations(methodAuths, principal);
-        this.timing = timing;
+        this.timing = (RunningQueryTimingImpl) timing;
+        if (this.timing == null) {
```
ivakegg: We check for timing being null below... no need for this

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
cmitchell: Not using a list (and the flush() method) will cause one result to be dropped each time an intermediate result is hit.  This is because an intermediate result is "replaced" with an empty document, and without a backing list, that result that was "replaced" is lost.  The `RunnigQuery` call itself does not have a way to persist that result across calls to the next()  -- with our current implementation, a list needs to be kept somewhere, or else we will lose results each time we hit an intermmediate result. 

This is reproducible by commenting out line 112 and lines 31 - 136 of  the `testLongRunningUniqueQuery`.  Commenting out those lines turns off the intermediate result functionality, and the relevant assertions.  You will see that you get 3 results.  

Changing the code in `UniqueTransform` to what is suggested, specifically the following, and uncommenting out the above lines and re-running the test, you will see that you get an assertion error, and only 2 results. 
```
            try {
                if (!FinalDocumentTrackingIterator.isFinalDocumentKey(keyDocumentEntry.getKey())
                                && uniqueUtil.isDuplicate(keyDocumentEntry.getValue(), uniqueFields, modelMapping)) {
                    return null;
                }
            } catch (IOException ioe) {
                log.error("Failed to convert document to bytes.  Returning document as unique.", ioe);
            }
//            documents.add(keyDocumentEntry);
        }
        
        long elapsedExecutionTimeForCurrentPage = System.currentTimeMillis() - this.queryExecutionForPageStartTime;
        if (elapsedExecutionTimeForCurrentPage > this.queryExecutionForPageTimeout) {
            Document document = new Document();
            document.setIntermediateResult(true);
            return Maps.immutableEntry(new Key(), document);
        }
        
        return keyDocumentEntry;
    }
    
    /**
     * Returns the next unique document in the pipeline, if there is one. Also checks to see if the timeout for the page has exceeded and interjects an
     * intermediate result instead of hte next unique document if so.
     *
     * @return The next unique document per the configured fields, or an intermediate result (blank document) if the page execution time has elapsed, or null if
     *         there isn't one and the timeout hasn't elapsed.
     */
//    @Override
//    public Entry<Key,Document> flush() {
//        Entry<Key,Document> entry = null;
//        if (!documents.isEmpty()) {
//            entry = documents.pop();
//        }
//
//        return entry;
//    }
```

While I understand this is less than ideal because the apply() and flush() mechanism waits until all results are ready, before sending them to the client as opposed to streaming them,  I don't think changing this to a true streaming paradigm can currently be done without major modifications, chiefly to the `RunningQuery` to make it aware of result state across calls to `next()`   Therefore, I did not make this change, nor the one above.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
cmitchell: Please see previous comment.

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
cmitchell: Removed.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
cmitchell: See the next comment.

web-services/query/src/main/java/datawave/webservice/query/iterator/DatawaveTransformIterator.java
cmitchell: Removed the condition.

web-services/query/src/main/java/datawave/webservice/query/logic/BaseQueryLogic.java
cmitchell: Removed

web-services/query/src/main/java/datawave/webservice/query/runner/RunningQuery.java
cmitchell: Removed

hlgp:
Stale and OBE

PR #1509, opened 2022-04-14
Disable LAZY_SET mechanism

### LINKED ISSUE #1295: Remove the LAZY_SET_EVALUATION mechanism or restrict the fields.
The @LAZY_SET_EVALUATION (or something close to that) is being added on index-only fields that are used in the filter: functions.  Firstly this breaks the filter:includeText mechanism which is already added the index query to ensure the jexl context is pre-populated.  That function should probably be moved to a separate function class as it is not "EvaluationOnly" but that is beside the point.  The main problem is a performance issue when the index only field is a tokenized field generated from large pieces of text and hence can result in a large number of items being added to the jexl context and subsequently can have performance issues.  It is easier to teach that index-only fields cannot be used in filter: functions.

lbschanno:
Add the config flag lazySetMechanismEnabled to enabled/disable the
LAZY_SET mechanism for index-only fields. Enforce this flag within
SetMembershipVisitor.

Resolves #1295

Additionally, the following behavior of SetMembershipVisitor has been changed:
- `SetMembershipVisitor.contains()` will return true if there is a match found against a previously-tagged field.
- `SetMembershipVisitor.getMembers()` will return matches that include any that were previously tagged.


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/SetMembershipVisitor.java
```diff
@@ -41,404 +40,291 @@ public class SetMembershipVisitor extends BaseVisitor {
...
      */
     public static final String INDEX_ONLY_FUNCTION_SUFFIX = "@LAZY_SET_FOR_INDEX_ONLY_FUNCTION_EVALUATION";
-    public static final String FILTER = "filter";
-    protected final Set<String> expectedFields;
-    protected final MetadataHelper metadataHelper;
-    protected final DateIndexHelper dateIndexHelper;
-    protected final ShardQueryConfiguration config;
-    protected final Set<String> discoveredFields;
-    protected final boolean fullTraversal;
+    public static final String FILTER_FUNCTION_NAME = "filter";
```
apmoriarty: Discard this constant for `EvaluationPhaseFilterFunctions.EVAL_PHASE_FUNTION_NAMESPACE`

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/SetMembershipVisitor.java
```diff
@@ -29,10 +26,12 @@
 import org.apache.commons.jexl2.parser.ASTReferenceExpression;
 import org.apache.commons.jexl2.parser.JexlNode;
 
-/** 
- *
+/**
+ * This visitor provides methods for determining whether a query tree contains any fields from a given query tree, and retrieving those matching fields if
```
apmoriarty: I think the first sentence should read "This visitor provides methods for determining if a query tree contains any of the provided fields"

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/SetMembershipVisitorTest.java
```diff
@@ -0,0 +1,284 @@
...
+import org.apache.commons.jexl2.parser.JexlNode;
+import org.apache.commons.jexl2.parser.ParseException;
+import org.assertj.core.api.Assertions;
+import org.junit.Test;
+
+import java.util.Set;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class SetMembershipVisitorTest {
```
apmoriarty, line 19: Excellent test coverage. While not required for approval, I would like to see common code and common test infrastructure extracted to common methods. For example, 
- `testContains(query, fields, expectedContains)`
- `testMembership(query, fields, expectedFields)`
- `testMembershipWithTagging(query, fields, expectedQuery, expectedFields)`

apmoriarty (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/SetMembershipVisitor.java
lbschanno: Fixed.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/SetMembershipVisitor.java
lbschanno: Good catch, fixed.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/SetMembershipVisitorTest.java
lbschanno: I took a slightly different approach than suggested, but I think you will still like it. Let me know either way.

apmoriarty: I like it!

apmoriarty (dismissed):

ivakegg (dismissed):

hlgp (dismissed):

ivakegg:
I was going to attempt integration testing but I need some conflicts resolved please.

ivakegg (dismissed):

ivakegg:
I will try to integration test this today

apmoriarty (dismissed):

PR #1523, opened 2022-04-20
Fixes #1522 - Upgrade to log4j 2.x

### LINKED ISSUE #1522: Upgrade to log4j 2x

keith-ratcliffe:
This PR represents the minimum number of changes required to eliminate the log4j 1.2.x dependency, as a first step toward full 2.x adoption for DW 3.x. That is, `log4j-1.2-api-2.x.jar` and `-Dlog4j1.compatibility=true` are leveraged to allow our existing loggers and configs to work as-is.

I'm not certain this upgrade is something we want to pursue in the near term (or at all) for DW 3.x, but since this work has already been done on the `feature/accumulo-2.1` branch, I thought it would be useful to apply the upgrade to DW 3.x just to gauge overall feasibility, level of effort, etc

Also worth noting that this PR demonstrates (via automated quickstart install) in-place replacement of log4j 1.x jars in both  Accumulo and Hadoop with the same log4j 2.17.2 jars that DW uses here. Testing so far has revealed no major issues and seems to indicate log4j 2x viability for patching Accumulo 1.x and Hadoop 3.x systems


warehouse/core/src/test/java/datawave/iterators/filter/ageoff/FieldAgeOffFilterTest.java
```diff
@@ -423,11 +425,11 @@ public void handlesFieldsStartingWithNumber() {
         ageOffFilter.init(filterOptions, iterEnv);
         // field_y is a match, but its ttl was not defined, so it will use the default one
         Key key = new Key("1234", "myDataType\\x00my-uuid", "field_y\u0000value", VISIBILITY_PATTERN, tenSecondsAgo);
-        Assert.assertFalse(ageOffFilter.accept(filterOptions.getAgeOffPeriod(System.currentTimeMillis()), key, new Value()));
+        Assert.assertFalse(ageOffFilter.accept(filterOptions.getAgeOffPeriod(currentTime), key, new Value()));
         Assert.assertTrue(ageOffFilter.isFilterRuleApplied());
         // field_z is a match, but the key is more recent than its age off period
         Key keyNumeric = new Key("1234", "myDataType\\x00my-uuid", "12_3_4\u0000value", VISIBILITY_PATTERN, oneSecondAgo);
-        Assert.assertTrue(ageOffFilter.accept(filterOptions.getAgeOffPeriod(System.currentTimeMillis()), keyNumeric, new Value()));
+        Assert.assertTrue(ageOffFilter.accept(filterOptions.getAgeOffPeriod(currentTime), keyNumeric, new Value()));
```
keith-ratcliffe, line 432: This assertion seems to fail sporadically due to the interplay between `oneSecondAgo`, the 5sec TTL, and the amount of time that it takes for this test to run, which is subject to the whims of the test runner JVM. E.g., you can force the failure simply by placing a breakpoint at the top of this test and slowly stepping through.

Moving the `FieldAgeOffFilter` creation, which seems to be fairly expensive, to the top and pinning assertions to the `currentTime` definition seems to help, in my testing at least

ivakegg: +1

ivakegg (dismissed):

hlgp (dismissed):

keith-ratcliffe:
Closing, as it is OBE

PR #1535, opened 2022-04-25
Create matchRegex function for index-only fields

### LINKED ISSUE #1296: non evaluation-phase function equivalents required
Users can currently apply a regex  to a field (e.g. FIELD =~ 'regex' or FIELD:/regex/) .  Users can also apply a regex with an evaluation phase filter function (filter:includeRegex(FIELD, 'regex') or #INCLUDE(FIELD,regex)).  On functions we can apply other operations such as size (see FunctionalSet) (e.g. filter:getAllMatches(FIELD, 'regex').size() == 2).  However we do not have a regex matching function that can be applied efficiently to a index-only field such (given we implement #1295) such that additional operations can be applied.
Recommend creating a new bucket of functions like those in EvaluationPhaseFilterFunctions that so the same things except that the index queries are generated correctly (via the function descriptor).

lbschanno:
Create the function f:matchRegex, which will act as the equivalent
of filter:includeRegex that is able to be applied efficiently to an
index-only field.

Resolves #1296


lbschanno:
Force push performed to correct information in top-level commit message.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/QueryFunctionsDescriptor.java
```diff
@@ -43,24 +48,44 @@ public QueryJexlArgumentDescriptor(ASTFunctionNode node, String namespace, Strin
...
+                    return TRUE_NODE;
+            }
+        }
+        
+        private JexlNode getIndexQuery() {
+            JexlNode node0 = args.get(0);
+            final String value = args.get(1).image;
+            if (node0 instanceof ASTIdentifier) {
+                final String field = JexlASTHelper.deconstructIdentifier(node0.image);
+                return JexlNodeFactory.buildNode((ASTEQNode) null, field, value);
```
hlgp: make these RE nodes instead of EQ, please
perhaps add a more end to end unit test running a whole query

hlgp: might also need to tweak the planner since it looks like we try to push down regular expressions before attempting to expand against the index which obviously won't work for index-only fields

hlgp: No planner work needed.  Turns out my test regex was hitting some preconfigured test regex that was forcing the attempted push down

lbschanno: Glad to hear it.

hlgp: I would still like a more end-to-end test though, please.

ivakegg (changes requested):
You need to create a datawave.query.language.functions.jexl.matches or matchesRegex class and add the new function to the QueryLogicFactory.xml

warehouse/query-core/src/main/java/datawave/query/jexl/functions/QueryFunctionsDescriptor.java
lbschanno: Will do, currently writing one up.

hlgp (approved):

ivakegg (approved):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/QueryFunctionsDescriptor.java
```diff
@@ -16,15 +16,20 @@
...
+import org.apache.commons.jexl2.parser.ASTEQNode;
 import org.apache.commons.jexl2.parser.ASTERNode;
 import org.apache.commons.jexl2.parser.ASTFunctionNode;
 import org.apache.commons.jexl2.parser.ASTGENode;
+import org.apache.commons.jexl2.parser.ASTIdentifier;
 import org.apache.commons.jexl2.parser.ASTLENode;
 import org.apache.commons.jexl2.parser.JexlNode;
 import org.apache.commons.jexl2.parser.ParserTreeConstants;
 
 public class QueryFunctionsDescriptor implements JexlFunctionArgumentDescriptorFactory {
```
apmoriarty, line 28: Great refactor for this class

warehouse/query-core/src/main/java/datawave/query/jexl/functions/QueryFunctionsDescriptor.java
```diff
@@ -43,24 +48,44 @@ public QueryJexlArgumentDescriptor(ASTFunctionNode node, String namespace, Strin
...
+                    JexlNode leNode = JexlNodeFactory.buildNode(new ASTLENode(ParserTreeConstants.JJTLENODE), args.get(0), args.get(2).image);
+                    // Return a bounded range.
+                    return BoundedRange.create(JexlNodeFactory.createAndNode(Arrays.asList(geNode, leNode)));
+                case LENGTH:
+                    // Return a regex node with the appropriate number of matching characters
+                    return JexlNodeFactory.buildNode(new ASTERNode(ParserTreeConstants.JJTERNODE), args.get(0), ".{" + args.get(1).image + ','
+                                    + args.get(2).image + '}');
+                case QueryFunctions.MATCH_REGEX:
+                    // Return an index query.
+                    return getIndexQuery();
```
apmoriarty, line 63: Adding in the index query is new -- can we document the behavior in a unit test? 

If you elect to add a test case to the MatchRegexTest, consider using raw query strings and the `getPlan` execution path.

apmoriarty (approved):

PR #1539, opened 2022-04-27
WIP: issue-1530: add metrics to the edgelogic/transformer classes

wgtong:


warehouse/query-core/src/main/java/datawave/query/transformer/EdgeQueryTransformerSupport.java
```diff
@@ -241,4 +255,21 @@ public String generateEventId(EdgeBase edge) {
         
         return Integer.toString(hashCode);
     }
+
+    public void setLogTimingDetails(Boolean logTimingDetails) {
+        this.logTimingDetails = logTimingDetails;
+    }
+
+    protected void extractMetrics(EdgeValue edgeValue, EdgeKey edgeKey) {
+        //TODO add a LogTiming.TIMING_METADATA to the EdgeValue to get source/next/seek counts
```
wgtong, line 264: seems like this data could be populated is added in an iterator? the EdgeValue does not have a flexible dict like Document.. not sure if it's ok to add those properties

wgtong:
The scope of the ticket is getting out of hand. we should revisit it as an epic

PR #1542, opened 2022-04-28
fixes #1541: Updated the excerpt mechanism to handle phrases against …

### LINKED ISSUE #1541: Excerpt fields does not work correctly for the TLD event query logic
The TLD event query logic treats the document tree as if it were one document and returns only the top level document (TLD) for those document trees that match the query.  If the excerpt fields mechanism is used in this case, the excerpts are pulled from the top level document only even if the hit was in a child document.  This results in an excerpt that has absolutely nothin to do with the actual phrase that was matched.

ivakegg:
…multiple documents, in particular for the TLD query logic.


lbschanno:
LGTM.

hlgp (approved):

PR #1556, opened 2022-05-11
Fixes #1517 - Adds new function signature for JexlNodes.children

### LINKED ISSUE #1517: Add a JexlNodes.children(JexlNode, Collection) method
The `JexlNodes` class offers the method ` children(T node, JexlNode... children)` for general purpose setting of children. 

Add a new method with signature `children(T node, Collection<JexlNode> children)`

magrable:


magrable:
Should any of the existing calls to this function use the new signature?

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/JexlNodes.java
```diff
@@ -78,6 +79,20 @@ public static <T extends JexlNode> T children(T node, JexlNode... children) {
...
     
+    /**
+     * Sets the supplied child array as the children member of {node} and sets the parent reference of each element in {children} to {node}.
+     *
+     * @param node
+     * @param children
+     * @return
+     */
+    public static <T extends JexlNode> T children(T node, Collection<JexlNode> children) {
+        node.children = children.toArray(new JexlNode[0]);
```
lbschanno: You can delegate to the existing JexlNode.children() method by shortening this to 
```java
return children(node, children.toArray(new JexlNode[0]));
```
and thereby avoid repeating some code.

lbschanno (changes requested):

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/JexlNodes.java
```diff
@@ -78,6 +79,20 @@ public static <T extends JexlNode> T children(T node, JexlNode... children) {
         return node;
     }
     
+    /**
+     * Sets the supplied child array as the children member of {node} and sets the parent reference of each element in {children} to {node}.
```
lbschanno: The documentation here is not quite accurate, as a child array is not actually supplied. Please update this to reflect that a child array will be created from the supplied collection, which is then set as the children of the node.

lbschanno (changes requested):

jack-gitdev:
Taking over ticket and made changes to address comments.

apmoriarty (approved):
LGTM. I believe you also addressed both of @lbschanno's comments

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/JexlNodes.java
jack-gitdev: Updated new children method to create a children array and pass it to the existing children method.

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/JexlNodes.java
jack-gitdev: Updated documentation to reflect the new implementation.

ivakegg (approved):

ivakegg:
I think this is OBE with jexl3.  Also the issue is already closed.

PR #1559, opened 2022-05-17
Fix expansion logic for RewriteNullFunctionsVisitor

apmoriarty:


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RewriteNullFunctionsVisitor.java
```diff
@@ -217,24 +191,70 @@ private JexlNode buildIsNullNode(ASTIdentifier identifier) {
...
+        }
+        
         return data;
     }
     
+    private void flattenJunction(JexlNode node, Class junctionType) {
+        List<JexlNode> children = new ArrayList<>();
+        for (JexlNode child : JexlNodes.children(node)) {
+            JexlNode deref = JexlASTHelper.dereference(child);
+            if (deref.getClass().getSimpleName().equals(junctionType.getSimpleName())) {
```
ivakegg: Why not use the instanceOf mechanism here?

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RewriteNullFunctionsVisitor.java
apmoriarty: Done

ivakegg (approved):

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/RewriteNullFunctionsVisitorTest.java
```diff
@@ -39,33 +39,33 @@ public void testIntersectionWithSingleFieldedIsNull() throws ParseException {
     @Test
     public void testMultiFieldedIsNull() throws ParseException {
         String query = "filter:isNull(FOO || FOO2)";
-        String expected = "FOO == null || FOO2 == null";
+        String expected = "FOO == null && FOO2 == null";
```
javease, line 42: I'm confused here and it might be my ignorance, but I don't know that this is what we want.

To me "isNull(F1 || F2)" should equate to "F1 == null || F2 == null" as you originally had.
However, if we have a multi-field expansion, i.e.
"isNull(F)" where "F" maps to "F1" and "F2", then we want:
"F1 == null && F2 == null".

I believe these are two distinct usages with different expected outcomes.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RewriteNullFunctionsVisitor.java
```diff
@@ -190,7 +160,11 @@ private JexlNode rewriteFilterFunction(FunctionJexlNodeVisitor visitor) {
             default:
                 // multi fielded case
                 rebuiltMultiFieldedFunction = true;
-                return JexlNodeFactory.createOrNode(children);
+                if (visitor.name().equals(IS_NULL)) {
```
javease, line 163: this seems odd to be doing OR for everything other than IS_NULL and AND for IS_NULL. Hmmm....

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/RewriteNullFunctionsVisitorTest.java
javease: Maybe the issue is that at this point we can't distinguish the two usages, i.e. we only see "isNull(F1 || F2)" regardless of whether that was the entered query or due to expansion. If that's the case, then it seems like at expansion time we should be doing something different (and maybe that's what you're doing and I just don't understand it).

javease (approved):

PR #1563, opened 2022-05-18
fixes #1557: Added the gathering of excerpts for query terms in addit…

### LINKED ISSUE #1557: Allow excerpts to work with non-content function terms
The next addition to excerpts is to be able to handle (FIELD op 'value') terms where the field has term offsets, and the term is in the excerpt.field list.  This should be able to handle at least ==, =~, and ranges.  The basic design plan is as follows:

1) Create a simple iterator that will take a term (e.g. FIELD op 'value') and will find the first offset that matches in the term frequencies for a specified document.
2) For documents that evaluate to true, and for those terms that meet the above critera AND are not already part of an index query tied to a content function (may have to use a marker of some sort in the query), use the iterator in (1) to find the offset.
3) Then use the existing term frequency excerpt mechanism to find an excerpt.

ivakegg:
…ion to the previously added phrase excerpts.

* This means that match from any term (e.g. FIELD == 'value' or FIELD =~ 'value.*') will also result in an hit excerpt being added to the document if requested.


warehouse/query-core/src/main/java/datawave/query/function/JexlEvaluation.java
```diff
@@ -96,25 +97,21 @@ public boolean apply(Tuple3<Key,Document,DatawaveJexlContext> input) {
                 
                 for (ValueTuple hitTuple : hitListArithmetic.getHitTuples()) {
                     
-                    ColumnVisibility cv = null;
                     String term = hitTuple.getFieldName() + ':' + hitTuple.getValue();
+                    Attribute a = hitTuple.getSource();
```
lbschanno: I'd like to see a more descriptive variable name here instead of 'a', 'attr' would suffice.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -84,12 +91,113 @@ public Entry<Key,Document> apply(@Nullable Entry<Key,Document> entry) {
...
+            allPhraseIndexes.addAll(phraseIndexes);
+        }
+        // now lets find all of the terms for excerpt fields and add them to the list
+        if (document.containsKey(JexlEvaluation.HIT_TERM_FIELD)) {
+            Attributes hitList = (Attributes) document.get(JexlEvaluation.HIT_TERM_FIELD);
+            // for each hit term
+            for (Attribute a : hitList.getAttributes()) {
+                ValueTuple hitTuple = attributeToHitTuple(a);
+                // if this is for a requested excerpt field
+                if (excerptFields.getFields().contains(hitTuple.getFieldName())) {
```
lbschanno: It would be nice to add a `containsField(field)` method to the ExcerptFields class and use that instead of retrieving the field set to do this for each attribute.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/PhraseIndexes.java
```diff
@@ -73,9 +90,38 @@ public static PhraseIndexes from(String string) {
      *            the end index of the phrase
      */
     public void addIndexTriplet(String field, String eventId, int start, int end) {
+        // first remove any overlapping phrases and extend the start/end appropriately
+        if (map.containsKey(field)) {
+            Iterator<Triplet<String,Integer,Integer>> indices = map.get(field).iterator();
+            while (indices.hasNext()) {
+                Triplet<String,Integer,Integer> indice = indices.next();
+                // if we have gone past the end, then no more possiblility of overlapping
```
lbschanno: Spelling typo here for the word possibility.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/PhraseIndexes.java
```diff
@@ -170,4 +216,69 @@ public String toString() {
...
+     * Return a map of fieldname to a map of eventId to its subset of phrase indexes.
+     * 
+     * @return a map of fieldname to eventId to phraseIndexes
+     */
+    public Map<String,Map<String,PhraseIndexes>> toMap() {
+        Map<String,Map<String,PhraseIndexes>> map = new HashMap<>();
+        for (String field : getFields()) {
+            Map<String,PhraseIndexes> fieldMap = new HashMap<>();
+            map.put(field, fieldMap);
+            for (String eventId : getIndices(field).stream().map(t -> t.getValue0()).collect(Collectors.toSet())) {
```
lbschanno: The lambda `t -> t.getValue0()` can be replaced with the method reference `Triplet::getValue0`.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -84,12 +91,113 @@ public Entry<Key,Document> apply(@Nullable Entry<Key,Document> entry) {
...
-        } else {
+            PhraseIndexes phraseIndexes = PhraseIndexes.from(content.getContent());
+            phraseIndexMap = phraseIndexes.toMap();
+            allPhraseIndexes.addAll(phraseIndexes);
+        }
+        // now lets find all of the terms for excerpt fields and add them to the list
+        if (document.containsKey(JexlEvaluation.HIT_TERM_FIELD)) {
+            Attributes hitList = (Attributes) document.get(JexlEvaluation.HIT_TERM_FIELD);
+            // for each hit term
+            for (Attribute a : hitList.getAttributes()) {
```
lbschanno: Raw usage of the parameterized class Attribute here, please use `Attribute<?>`.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -84,12 +91,113 @@ public Entry<Key,Document> apply(@Nullable Entry<Key,Document> entry) {
...
+                        // add the term as a phrase as defined in the term weight position. Note that this will collapse with any overlapping phrases already in
+                        // the list.
+                        allPhraseIndexes.addIndexTriplet(String.valueOf(hitTuple.getFieldName()), keyToEventId(a.getMetadata()), pos.getLowOffset(),
+                                        pos.getOffset());
+                    }
+                }
+            }
+        }
+        // return the file set of phrase indexes if any
+        return (allPhraseIndexes.isEmpty() ? null : allPhraseIndexes);
```
lbschanno: Instead of returning null here, I would prefer returning allPhraseIndexes as is and change the check for null on line 70 to `if (!phraseIndexes.isEmpty())`.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -84,12 +91,113 @@ public Entry<Key,Document> apply(@Nullable Entry<Key,Document> entry) {
...
+        // if we do not know the source document key, then we cannot find the term offset
+        if (docKey == null) {
+            log.warn("Unable to find the source document for a hit term, skipping excerpt generation");
             return null;
         }
+        String fieldName = hitTuple.getFieldName();
+        String eventId = keyToEventId(docKey);
+        
+        // pull out the appropriate phrase indexes for this field/event id
+        PhraseIndexes phraseIndexes = null;
```
lbschanno: This variable is not used anywhere outside of the try block, and doesn't need to be initialized here. You can perform a check for a pre-existing PhraseIndex where it's relevant at line 165, i.e.

```java
if(phraseIndexMap.containsKey(fieldName)) {
    PhraseIndexes phraseIndexes = phraseIndexMap.get(fieldName).get(eventId);
    for (int i = 0; i < twInfo.getTermOffsetCount(); i++) {
    ...
```

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -84,12 +91,113 @@ public Entry<Key,Document> apply(@Nullable Entry<Key,Document> entry) {
...
+        return null;
+    }
+    
+    /**
+     * Given a hit term attribute, return a ValueTuple representation which will give us the field and value parsed out.
+     * 
+     * @param source
+     * @return A ValueTuple representation of the document hit-term attribute
+     */
+    private ValueTuple attributeToHitTuple(Attribute source) {
```
lbschanno: Raw usage of Attribute, please use `Attribute<?>`.

lbschanno (changes requested):

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
ivakegg: not quite right because I still need to see of phraseIndexes is null as well, but since this is a hashmap lookup I can move it inside the loop.

lbschanno (dismissed):
LGTM

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/PhraseIndexes.java
```diff
@@ -54,14 +57,28 @@ public static PhraseIndexes from(String string) {
                 String eventId = indexParts[0];
                 int start = Integer.parseInt(indexParts[1]);
                 int end = Integer.parseInt(indexParts[2]);
-                phraseIndexes.addIndexTriplet(field, eventId, start, end);
+                phraseIndexes.addIndexTriplet(field, eventId.isEmpty() ? null : eventId, start, end);
```
drewfarris: consider pushing this empty check into the `addIndexTriplet` method, it seems a little odd here.

warehouse/query-core/src/main/java/datawave/query/jexl/HitListArithmetic.java
```diff
@@ -560,6 +560,14 @@ private void addTheCorrectHitSetValue(Set<ValueTuple> hitSet, Object left, Objec
...
     public static ColumnVisibility getColumnVisibilityForHit(Document document, String hitTerm) {
+        Attribute attr = getAttributeForHit(document, hitTerm);
+        if (attr != null) {
+            return attr.getColumnVisibility();
+        }
+        return null;
+    }
+    
+    public static Attribute getAttributeForHit(Document document, String hitTerm) {
         // get the visibility for the record with this hit
```
drewfarris: Consider updating this comment because we are no longer getting the visibiltiy only

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/PhraseIndexes.java
```diff
@@ -158,8 +204,8 @@ public String toString() {
             Iterator<Triplet<String,Integer,Integer>> indexIterator = map.get(field).iterator();
             while (indexIterator.hasNext()) {
                 Triplet<String,Integer,Integer> indexTriplet = indexIterator.next();
-                sb.append(indexTriplet.getValue0()).append(Constants.COMMA).append(indexTriplet.getValue1()).append(Constants.COMMA)
-                                .append(indexTriplet.getValue2());
+                sb.append(indexTriplet.getValue0() == null ? "" : indexTriplet.getValue0()).append(Constants.COMMA).append(indexTriplet.getValue1())
+                                .append(Constants.COMMA).append(indexTriplet.getValue2());
```
drewfarris: Consider disabling auto-formatting here to make the `append` chain more readable (e.g., one append per line)

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -44,8 +51,8 @@ public ExcerptTransform(ExcerptFields excerptFields, IteratorEnvironment env, So
         this(excerptFields, env, source, new TermFrequencyExcerptIterator());
     }
     
-    protected ExcerptTransform(ExcerptFields excerptFields, IteratorEnvironment env, SortedKeyValueIterator<Key,Value> source,
-                    TermFrequencyExcerptIterator excerptIterator) {
+    public ExcerptTransform(ExcerptFields excerptFields, IteratorEnvironment env, SortedKeyValueIterator<Key,Value> source,
+                    SortedKeyValueIterator<Key,Value> excerptIterator) {
         this.excerptFields = excerptFields;
```
drewfarris, line 58: Consider adding a null check for `excerptFields` here?

warehouse/query-core/src/test/java/datawave/query/transformer/ExcerptTransformTest.java
```diff
@@ -120,8 +126,38 @@ public void testExcerpts() throws IOException {
         givenExcerptField("BODY", 2);
         givenPhraseIndex("BODY", 10, 14);
         
-        givenMockDocument();
+        givenMockDocumentWithHitTerm("BODY", "word");
```
drewfarris, line 134: This test threw me of a bit because I expected the phrase with the term `word` to match and the one without the term `word` to not match, but I suspect we're not really doing that sort of evaluation in the transformation, so it is fine. Consider clarifying/commenting?

warehouse/query-core/src/test/java/datawave/query/postprocessing/tf/PhraseIndexesTest.java
```diff
@@ -64,4 +67,65 @@ public void testParsingFromNonBlankString() {
                         + ",12,17");
         assertEquals(expected, actual);
     }
+    
+    @Test
```
drewfarris, line 85: Consider adding some pathalogical cases, like phrases that have a zero length, phases where the end appears before the start, cases where the numbers are negative or the ranges are really really big. Maybe even cases where the field and/or event are empty or null?

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -100,30 +206,54 @@ private PhraseIndexes getPhraseIndexes(Document document) {
...
     }
     
+    /**
+     * Given an event ID, return the document Key
+     * 
+     * @param eventId
+     * @return the document Key
+     */
+    private Key eventIdToKey(String eventId) {
+        int split = eventId.indexOf('\u0000');
```
drewfarris: consider doing a bit more checking here? (e.g., null/empty string, unexpected split result?)

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -100,30 +206,54 @@ private PhraseIndexes getPhraseIndexes(Document document) {
      * @param document
      *            the document
      */
-    private void addExcerptsToDocument(Set<String> excerpts, Document document) {
+    private void addExcerptsToDocument(Set<Excerpt> excerpts, Document document) {
         Attributes attributes = new Attributes(true);
```
drewfarris, line 206: Consider adding an empty check to the `excerpts` argument here? -or- do we want to add an empty HIT_EXCERPT attriute if the `excerpts` argument is empty? Do we want to always overwrite the contents of the HIT_EXCERPT attribute? (e.g., Would it be possible for HIT_TERM to exist in the Document already?)

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -84,12 +91,111 @@ public Entry<Key,Document> apply(@Nullable Entry<Key,Document> entry) {
...
+                // if no overlapping phrases, then return the first position
+                position.setTermWeightOffsetInfo(twInfo, 0);
+                return position.build();
+            }
+            
+        } catch (InvalidProtocolBufferException e) {
+            log.error("Value passed to aggregator was not of type TermWeight.Info for " + tfKey, e);
+        } catch (IOException e) {
+            log.error("Failed to scan for term frequencies at " + tfKey, e);
+        }
```
drewfarris, line 179: Consider whether we want to swallow and log these -or- whether we should throw them updwards and fail the query? It seems that they might be sufficiently bad to fail upwards, but also, other things would probably fail if either of these happened, so the extra complexity might not be necessary.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -84,12 +91,111 @@ public Entry<Key,Document> apply(@Nullable Entry<Key,Document> entry) {
...
+        String eventId = keyToEventId(docKey);
+        
+        // get the key at which we would find the term frequencies
+        Key tfKey = new Key(docKey.getRow().toString(), Constants.TERM_FREQUENCY_COLUMN_FAMILY.toString(), docKey.getColumnFamily().toString() + '\u0000'
+                        + hitTuple.getValue() + '\u0000' + hitTuple.getFieldName());
+        Range range = new Range(tfKey, tfKey.followingKey(PartialKey.ROW_COLFAM_COLQUAL));
+        try {
+            // seek directly to that key
+            source.seek(range, Collections.emptyList(), false);
+            if (source.hasTop()) {
```
drewfarris, line 154: I'm wondering if we need to build in some limits somewhere. I'm thinking of a case where we have an incredibly long document with many term positions and we've searched that appears incredibly frequently in that document. Perhaps the phrase index map provides sufficient limiting here or since we are returning only the first offset of the term is is OK. Do we create a circuit breaker that will bail after we've visited some number of tf's (This is probably not the right place to do this just thought of it while looking at this code).

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/PhraseIndexes.java
ivakegg: I disagree on this one.  having an empty string for the event id is simply because of the way it is being parsed out of the string.

ivakegg: I moved the check out of the call and added an appropriate comment

warehouse/query-core/src/main/java/datawave/query/jexl/HitListArithmetic.java
ivakegg: done

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/PhraseIndexes.java
ivakegg: ok

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
ivakegg: ok

warehouse/query-core/src/test/java/datawave/query/transformer/ExcerptTransformTest.java
ivakegg: comments added

warehouse/query-core/src/test/java/datawave/query/postprocessing/tf/PhraseIndexesTest.java
ivakegg: ok, I will work this.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
ivakegg: ok

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
ivakegg: document.put already merges the attributes.  Given this is a private method and this class controls how it is called, I don'e think we need to check for null.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
ivakegg: I was considering adding excerpts as a best effort sort of activity.  I want to see the errors in the tserver logs, but I don't want to fail queries in these cases.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
ivakegg: This is going directly to the key and pulling it.  There is no looping through TF's here.  We do loop through the offsets deserialized from the single key's value IFF we also have phrase indexes (i.e. there are phrase or within clauses in the query as well).  I think the offsets are sorted so perhaps a binary search would be more prudent here (O(Nlog2) vs O(N)).  At a minimum I could short circuit the loop once the offset is past that maximum phrase end offset.

lbschanno (approved):
LGTM.

drewfarris (approved):

hlgp (approved):

PR #1566, opened 2022-05-19
fixes #1558: Wrap terms with evaluation only if we have a regex that fail normalization

### LINKED ISSUE #1558: Regexes against numeric fields cannot use the field index
For fields that are types as being numeric, regexes against said terms cannot work against the field index given the nature of our numberic normalization.  Hence those fields should be automatically marked as being evaluation only.

ivakegg:


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
```diff
@@ -259,6 +261,14 @@ protected JexlNode expandNodeForNormalizers(JexlNode node, Object data) {
                             if (log.isTraceEnabled()) {
                                 log.trace("Could not normalize " + term + " using " + normalizer.getClass());
                             }
+                            // if this was a regex, then lets assume that the regex could not be decoded enough to determine how to match against the index.
+                            // in this case we need to force this term to be evaluation only (i.e. cannot be done against an index)
+                            if (node instanceof ASTNRNode || node instanceof ASTERNode) {
```
apmoriarty, line 266: Can we expand the test case below to also hit the ASTNRNode?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
```diff
@@ -274,6 +284,13 @@ protected JexlNode expandNodeForNormalizers(JexlNode node, Object data) {
                         nodeToReturn = JexlNodeFactory.buildUntypedNewLiteralNode(node, fieldName, literal);
                     }
                     
+                    // if we have an unnormalizable regex in the mix, then wrap with evaluation only
+                    // this is ok even if we have a mix of normalizations that worked vs not because
+                    // if one of a union is evaluation only, then they all must be (see executability
+                    // visitor for more understanding)
```
apmoriarty, line 290: Does this imply we need to push whole  unions into EvalOnly? `(A or B or (eval & C)) -> (eval & (A or B or C))`

apmoriarty, line 294: Does this imply we need to push whole  unions into EvalOnly? `(A or B or (eval & C)) -> (eval & (A or B or C))`

apmoriarty (changes requested):

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTermsTest.java
```diff
@@ -490,6 +488,27 @@ public void testDelayedPredicates() throws ParseException {
...
+    public void testFailedRegexNormalizers() throws ParseException {
+        Multimap<String,Type<?>> dataTypes = HashMultimap.create();
+        dataTypes.putAll("FOO", Sets.newHashSet(new LcNoDiacriticsType(), new LcType(), new NumberType()));
+        
+        helper.setIndexedFields(dataTypes.keySet());
+        helper.addTermFrequencyFields(dataTypes.keySet());
+        
+        config.setQueryFieldsDatatypes(dataTypes);
+        
+        // this tests for the successful normalization as a simply number can be normalized as a regex
```
hlgp: simple number

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTermsTest.java
```diff
@@ -490,6 +488,27 @@ public void testDelayedPredicates() throws ParseException {
         }
     }
     
+    @Test
+    public void testFailedRegexNormalizers() throws ParseException {
```
hlgp: Maybe a few more test cases? 
Consider the case when the field is index only.
Are we going to update the HitListArithmetic where it uses the normalized form out of the tuple? Or is that out of the scope of this issue?  Currently if there's an event with FRUIT:BANANA and WEIGHT:40, the query 
FRUIT == 'banana' AND WEIGHT =~ '4.\*'
will result in the correct plan, FRUIT == 'banana' && (( _Eval_=true) && WEIGHT =~ '4.\*')
but will try to compare the pattern, 4.* to +bE4, which doesn't work.  Maybe I'm missing something.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
ivakegg: So this is the case were we have one field and multiple normalizations there of in which one or more of them failed.  In this case I thought it reasonable to send the whole expression to be eval only.  For example:
```
FIELD =~ '.*AE3.*'
where FIELD is a String and a Number type
becomes:  ((_Eval_ = true ) && ( ( FIELD =~ '.*ae3.*' ) || ( FIELD =! '.*AE3.*' ) ) )
```
If I cannot run one of them against the index, then the entire union cannot be run against the index.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTermsTest.java
ivakegg: ok

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTermsTest.java
ivakegg: looking at fixing the HitListArithmetic to be consistent with the DefaultArithmetic.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
ivakegg: I did, and I also realized that we were not handling them correctly.  Note that the NR and NE cases are not actually ever exercised  because all NE and NR nodes are turned into EQ and ER nodes surrounded by a negation before we get here.

apmoriarty (approved):

hlgp (approved):

PR #1572, opened 2022-05-23
Additional cases for IsNotNullPruningVisitor

apmoriarty:
IsNotNullPruningVisitor handles additional cases, add isNull/isNotNull tests to DatawaveInterpreterTest


avgAGB (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IsNotNullPruningVisitor.java
```diff
@@ -196,22 +298,28 @@ protected String fieldForChild(JexlNode node) {
...
+                    field = names.iterator().next();
                 } else {
                     return null; // EQ or ER node had more than one identifier...this is not correct
                 }
+            } else if (child instanceof ASTFunctionNode) {
+                field = fieldForFunction(child);
+            }
+            
+            if (field == null) {
+                return null; // encountered an unexpected node type, the union cannot have a single shared field
```
hlgp, line 328: Worth logging maybe?

warehouse/query-core/src/test/java/datawave/query/jexl/DatawaveInterpreterTest.java
```diff
@@ -186,6 +186,83 @@ public void testBoundedRangeOrTerm() {
...
+        query = "FOO == 'bar' && filter:isNull(ABSENT)";
+        test(query, buildDefaultContext(), true);
+        
+        // multi field, all present
+        query = "FOO == 'bar' && filter:isNull(FOO || FOO)";
+        test(query, buildDefaultContext(), false);
+        
+        // multi field, (present || absent)
+        query = "FOO == 'bar' && filter:isNull(FOO || ABSENT)";
+        // test(query, buildDefaultContext(), true); // this is wrong
```
hlgp: If these are wrong, maybe correct or remove them

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IsNotNullPruningVisitor.java
```diff
@@ -188,6 +270,26 @@ protected String fieldForChild(JexlNode node) {
...
+     *
+     * @param node
+     *            an ASTFunction node
+     * @return a single field if it exists, otherwise null
+     */
+    protected String fieldForFunction(JexlNode node) {
+        FunctionJexlNodeVisitor visitor = new FunctionJexlNodeVisitor();
+        node.jjtAccept(visitor, null);
+        
+        if (visitor.namespace().equals("filter") && visitor.name().equals("includeRegex")) {
```
ivakegg: This seems like a really specific case.  Can't we do something generic for functions in general?  I think the function descriptor classes have a method to return the fields.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IsNotNullPruningVisitor.java
```diff
@@ -196,22 +298,28 @@ protected String fieldForChild(JexlNode node) {
      * @return a singular field if it exists, or null
      */
     protected String fieldForUnion(JexlNode node) {
+        String field;
         Set<String> fields = new HashSet<>(node.jjtGetNumChildren());
         for (JexlNode child : JexlNodes.children(node)) {
             child = JexlASTHelper.dereference(child);
+            field = null;
             if (child instanceof ASTEQNode || child instanceof ASTERNode) {
```
ivakegg, line 313: why not a function node as well?

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IsNotNullPruningVisitor.java
apmoriarty: Agreed, added in next commit.

warehouse/query-core/src/test/java/datawave/query/jexl/DatawaveInterpreterTest.java
apmoriarty: Moved the multi-fielded instances of `isNull` and `isNotNull` to their own test methods and put them behind an `@Ignore` annotation until #1604 can be worked.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IsNotNullPruningVisitor.java
apmoriarty: To limit scope I would like to only add the ability to prune if an `isNotNull` term is intersected with a `filter:includeRegex` term against the same field (for example, `filter:isNotNull(FOO) && filter:includeRegex(FOO, 'ba.*')`.

I'm open to adding more cases in the future, did you have any specific functions in mind? An initial list is below
- content function: primarily index-only
- f:includeText once that move is complete
- f:geo functions?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IsNotNullPruningVisitor.java
apmoriarty: Complexity and scope. I'll write the test case for this and leave it as future work if you find that agreeable.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IsNotNullPruningVisitor.java
ivakegg: I think you can do it for all functions.  The descriptor mechanism gives you a way to do that without having to enumerate them.

apmoriarty: The descriptor `fields()` method requires the MetadataHelper and the datatype filter for the query. I was hoping to avoid adding those classes to this visitor to keep it lightweight. 

Even so, `filter:isNotNull` is not allowed to have to index-only fields as arguments. Logically, we can't prune against content functions because content functions are (almost?) always run against index-only fields.

apmoriarty: Wrote up an issue for this https://github.com/NationalSecurityAgency/datawave/issues/1605

ivakegg (approved):

warehouse/query-core/src/test/java/datawave/query/jexl/DatawaveInterpreterTest.java
```diff
@@ -186,6 +187,98 @@ public void testBoundedRangeOrTerm() {
...
+        // single field, present
+        String query = "FOO == 'bar' && filter:isNull(FOO)";
+        test(query, buildDefaultContext(), false);
+        
+        // single field, absent
+        query = "FOO == 'bar' && filter:isNull(ABSENT)";
+        test(query, buildDefaultContext(), true);
+    }
+    
+    @Ignore
```
FineAndDandy, line 202: should this be ignored?

warehouse/query-core/src/test/java/datawave/query/jexl/DatawaveInterpreterTest.java
```diff
@@ -186,6 +187,98 @@ public void testBoundedRangeOrTerm() {
...
+        
+        // single field, absent
+        query = "FOO == 'bar' && filter:isNotNull(ABSENT)";
+        test(query, buildDefaultContext(), false);
+        
+        query = "FOO == 'bar' && !(filter:isNull(ABSENT))";
+        test(query, buildDefaultContext(), false);
+    }
+    
+    @Ignore
```
FineAndDandy, line 245: and this

FineAndDandy (changes requested):
please resolve the ignored tests

FineAndDandy (approved):

Issue #1579, opened 2022-05-25
ColumnVisibilityFilters should cache last decision


apmoriarty:
The ColumnVisibilityOrFilter and ColumnVisibilityAndFilter should cache the last decision made.

If a column visibility matches the filter (or does NOT match) and an identical column visibility is seen immediately after, there is no need to check the patterns again -- just return the cached decision.

```
# for CV filter VIZ_A

shard datatype\0uid FIELD_A:value_a VIZ_A -- matches, record decision
shard datatype\0uid FIELD_B:value_b VIZ_A -- visibility matches last, return last decision without checking
```

apmoriarty:
Closed in favor of a different approach (accumulo access)

PR #1580, opened 2022-05-25
Cache decisions for ColumnVisibility filters

apmoriarty:
- The TokenFilterBase now has a column visibility cache to reduce the time spent parsing visibilities. 
- The cache has a default size of 50 but the size can be overridden using the COLUMN_VISIBILITY_CACHE_SIZE age off parameter. 
- added unit tests for the ColumnVisibilityOrFilter and ColumnVisibilityAndFilter

Addresses https://github.com/NationalSecurityAgency/datawave/issues/1579

### DIFF
--- a/warehouse/age-off/src/main/java/datawave/iterators/filter/ColumnVisibilityOrFilter.java
+++ b/warehouse/age-off/src/main/java/datawave/iterators/filter/ColumnVisibilityOrFilter.java
@@ -6,8 +6,6 @@ package datawave.iterators.filter;
 import org.apache.accumulo.core.data.Key;
 import org.apache.accumulo.core.data.Value;
 
-import java.util.Arrays;
-
 /**
  * This subclass of the {@code RegexFilterBase} class is used to filter based on the column visibility of the {@code Key} object {@code k}.
  */
@@ -27,16 +25,6 @@ public class ColumnVisibilityOrFilter extends TokenFilterBase {
         boolean found = false;
 
         byte[] cv = k.getColumnVisibilityData().getBackingArray();
-
-        if (prevCVBytes == null) {
-            prevCVBytes = cv; // first time, record cv
-        } else if (Arrays.equals(prevCVBytes, cv)) {
-            // ruleApplied was reset before the call to hasToken
-            setRuleApplied(true);
-            return prevDecision; // return cached decision
-        } else {
-            prevCVBytes = cv; // new cv found, record it
-        }
 
         // find the start of the first token to test
         int start = findNextNonDelimiter(cv, 0);
@@ -64,8 +52,7 @@ public class ColumnVisibilityOrFilter extends TokenFilterBase {
             }
         }
 
-        prevDecision = found;
-        return prevDecision;
+        return found;
     }
 
 }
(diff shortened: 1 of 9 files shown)


avgAGB:
You probably already have something like it in the works, but would it also be possible to test the functionality of the ColumnVisibilityAndFilter? It may not even need it's own test class per se, an instance of it in the existing unit test with a few examples tested might do the trick.

warehouse/core/src/main/java/datawave/iterators/filter/ColumnVisibilityAndFilter.java
```diff
@@ -6,19 +6,21 @@
...
+     *
      * @param k
      *            {@code Key} object containing the row, column family, and column qualifier.
      * @param v
      *            {@code Value} object containing the value corresponding to the {@code Key: k}
-     * @return {@code String} object containing the column visibility of {@code k}
+     * @return true if the provided key's column visibility matches the configured {@link TokenFilterBase}'s pattern
      */
     @Override
     public boolean hasToken(Key k, Value v, byte[][] testTokens) {
```
FineAndDandy: most of this method can be moved to an abstract method in TokenFilterBase

FineAndDandy (changes requested):

warehouse/core/src/test/java/datawave/iterators/filter/ColumnVisibilityOrFilterTest.java
```diff
@@ -0,0 +1,102 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Test for the {@link ColumnVisibilityOrFilter}
+ */
+public class ColumnVisibilityOrFilterTest {
```
hlgp: How about an AndFilterTest, too?

warehouse/core/src/main/java/datawave/iterators/filter/ColumnVisibilityAndFilter.java
apmoriarty: done

warehouse/core/src/test/java/datawave/iterators/filter/ColumnVisibilityOrFilterTest.java
apmoriarty: done

ivakegg (dismissed):

warehouse/age-off/src/main/java/datawave/iterators/filter/ColumnVisibilityOrFilter.java
```diff
@@ -12,25 +12,23 @@
 public class ColumnVisibilityOrFilter extends TokenFilterBase {
```
matthpeterson, line 10: We may want to deprecate this class.  I think we tested https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/age-off/src/main/java/datawave/iterators/filter/ColumnVisibilityTokenizingFilter.java and found it to be much faster.  Perhaps this caching change could be made to it.

apmoriarty: I can deprecate the ColumnVisibilityOrFilter and ColumnVisibilityAndFilter.

~~It looks like the ColumnVisibilityTokenizingFilter ultimately delegates to the `accept()` method found in the TokenFilterBase.~~

The TokenizingFilterBase could also use a cache. I can add that if you like.

avgAGB (approved):

apmoriarty:
Local testing showed some improvements in some cases, on-cluster testing showed degradation in the common cases.

PR #1581, opened 2022-05-25
Task/short circuit ivarator required 1476

### LINKED ISSUE #1476: Short Circuit Visitors - IvaratorRequired

mineralntl:


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ShortCircuitBaseVisitor.java
```diff
@@ -0,0 +1,216 @@
...
+        return data;
+    }
+    
+    @Override
+    public Object visit(ASTFalseNode node, Object data) {
+        return data;
+    }
+    
+    @Override
+    public Object visit(ASTIntegerLiteral node, Object data) {
```
avgAGB, line 147: The Integer/Float literals appear to be deprecated when I viewed this branch in IDE, would it still be necessary to override? Just wondering.

avgAGB, line 147: The Integer/Float literals appear to be deprecated when I viewed this branch in IDE, would it still be necessary to override? Just wondering.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IvaratorRequiredVisitorTest.java
```diff
@@ -0,0 +1,40 @@
...
+    
+    @Test
+    public void testExceededValueThresholdTrue() throws ParseException {
+        String query = "((_Value_ = true) && (FOO == 1 && FOO == 3))";
+        ASTJexlScript script = JexlASTHelper.parseJexlQuery(query);
+        assertTrue(IvaratorRequiredVisitor.isIvaratorRequired(script));
+    }
+    
+    @Test
+    public void testExceededValueThresholdFalse() throws ParseException {
```
avgAGB, line 28: Is this meant to be testing the same query as the one in line 21? If not, maybe adding some extra layer of complexity to the query to set it apart could be interesting.

mineralntl: That's an extraneous oversight on my part - any query without the _Value_ or _List_ marker should return true so there isn't really a need for this test.

avgAGB (dismissed):

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IvaratorRequiredVisitorTest.java
```diff
@@ -0,0 +1,34 @@
...
+    @Test
+    public void testExceededThresholdFalse() throws ParseException {
+        String query = "(FOO == 1 && FOO == 3)";
+        ASTJexlScript script = JexlASTHelper.parseJexlQuery(query);
+        assertFalse(IvaratorRequiredVisitor.isIvaratorRequired(script));
+    }
+    
+    @Test
+    public void testExceededValueThresholdTrue() throws ParseException {
+        String query = "((_Value_ = true) && (FOO == 1 || FOO == 3))";
```
apmoriarty: Value markers are for regexes

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ShortCircuitBaseVisitor.java
```diff
@@ -0,0 +1,216 @@
...
+import org.apache.commons.jexl2.parser.ASTSizeFunction;
+import org.apache.commons.jexl2.parser.ASTSizeMethod;
+import org.apache.commons.jexl2.parser.ASTStringLiteral;
+import org.apache.commons.jexl2.parser.ASTTernaryNode;
+import org.apache.commons.jexl2.parser.ASTTrueNode;
+import org.apache.commons.jexl2.parser.ASTUnaryMinusNode;
+import org.apache.commons.jexl2.parser.ASTWhileStatement;
+import org.apache.commons.jexl2.parser.SimpleNode;
+
+public class ShortCircuitBaseVisitor extends BaseVisitor {
```
apmoriarty, line 51: Class level documentation explaining what this is and why it's necessary. 

For my benefits, why are some methods overridden but not others?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ShortCircuitBaseVisitor.java
```diff
@@ -0,0 +1,216 @@
...
+    
+    @Override
+    public Object visit(ASTDivNode node, Object data) {
+        return data;
+    }
+    
+    @Override
+    public Object visit(ASTModNode node, Object data) {
+        return data;
+    }
```
apmoriarty: A visitor that needs to visit queries like the ones found in FunctionalSetTest.testFunctionsAsArguments() will not visit the whole tree. I think in most cases that is fine but if someone is doing something fancy with their query we could run into problems. Perhaps note that above in the class level documentation.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IvaratorRequiredVisitor.java
```diff
@@ -4,13 +4,17 @@
...
 import org.apache.commons.jexl2.parser.JexlNode;
 
 /**
  * A visitor that checks the query tree to determine if the query requires an ivarator (ExceededValue or ExceededOr)
- * 
  */
-public class IvaratorRequiredVisitor extends BaseVisitor {
+public class IvaratorRequiredVisitor extends ShortCircuitBaseVisitor {
     
     private boolean ivaratorRequired = false;
```
apmoriarty: There is a way to short circuit this visitor even faster. If the boolean `ivaratorRequired` is set to true there is no need to continue visiting the rest of the tree.

For take the method that visits an AND node for example
```
    @Override
        node.childrenAccept(this, data);
        return data;
    }


# becomes

    @Override
        if (!ivaratorRequired) {
            node.childrenAccept(this, data);
        }
        return data;
    }
```

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ShortCircuitBaseVisitor.java
apmoriarty: Perhaps we could also group nodes into logical groups. 
- There are nodes we always want to pass through or descend into (AND, OR, NOT, REF, REF_EXPR)
- There are nodes we generally consider as leaf nodes (EQ, NE, ER, NR, GT, LT, GE, LE)
- There are arithmetic nodes (Add, Mult, Div, Mod)
- The remaining nodes could be grouped into an "other" category, or further distinguished.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IvaratorRequiredVisitorTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+import datawave.query.jexl.JexlASTHelper;
+import org.apache.commons.jexl2.parser.ASTJexlScript;
+import org.apache.commons.jexl2.parser.ParseException;
+import org.junit.Test;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class IvaratorRequiredVisitorTest {
```
apmoriarty, line 10: This is a good start. At a minimum I would like to see some additional test cases for when the ivarator node is the child of an intersection and the child of a union
```
# base case
((_Value_ = true) && (FOO =~ 'ba.*'))

# with intersection
FOO == 'bar' && ((_Value_ = true) && (FOO =~ 'ba.*'))

# with union
FOO == 'bar' || ((_Value_ = true) && (FOO =~ 'ba.*'))
```

At that point it makes sense to refactor your test to have a common test method that accepts a query and the expected result of a visit. 
```
private void test(String query, boolean expected){
// common code here
```

apmoriarty (changes requested):

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IvaratorRequiredVisitorTest.java
mineralntl: Noted and changed.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IvaratorRequiredVisitorTest.java
mineralntl: Good idea - will make those changes

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IvaratorRequiredVisitor.java
mineralntl: Ah fair enough - I can make those changes to the nodes we always want to pass thru

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ShortCircuitBaseVisitor.java
mineralntl: Noted and will add a note in the docs

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ShortCircuitBaseVisitor.java
mineralntl: Notes and brief overview added

mineralntl:
Closing as OBE

PR #1582, opened 2022-05-25
Task/short circuit valid pattern 1498

### LINKED ISSUE #1498: Short Circuit Visitors - ValidPAttern

mineralntl:


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidPatternVisitor.java
```diff
@@ -126,4 +128,24 @@ public void parseAndPutPattern(JexlNode node) {
             }
         }
     }
+    
+    // Descend through these nodes
+    @Override
+    public Object visit(ASTJexlScript node, Object data) {
```
ivakegg: I don't think you need to override these methods as the short circuit base visitor does not block descending through these.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidPatternVisitor.java
mineralntl: For some reason I was thinking BaseVisitor blocked but must have ben an oversight on my part.

jschmidt10 (dismissed):

jschmidt10 (approved):

apmoriarty (approved):

ivakegg (approved):

PR #1587, opened 2022-05-27
fixes #1586: Updated the whindex visitor to correctly handle function…

### LINKED ISSUE #1586: Whindex visitor isn't handling geo functions with multiple fields correctly
The Whindex visitor is producing invalid trees when presented with a geo function with both a whindes field and a non-whindex field.

jwomeara:
…s with multiple fields


ivakegg (approved):

keith-ratcliffe (approved):
LGTM

Issue #1425, opened 2022-02-15
Create a method to pretty-print a Datawave query


jwomeara:
It would be really useful for debugging purposes to have an algorithm that could pretty-print a Datawave query such that the parens, terms, ranges, and functions are all aligned in a way that makes it easy to tell which terms are grouped, and where.

jwomeara:
Just to give an idea of what we're thinking of, if we were to have a query like this:
```
((FIELD == 'some value') && ((geowave:intersects(GEO, 'POLYGON((0.5 2, 1.5 2, 1.5 10, 0.5 10, 0.5 2))') && ((_Bounded_ = true) && (GEO >= '2.0|0.5' && GEO <= '10.0|1.5'))) || (geowave:intersects(POINT, 'POLYGON((0.5 2, 1.5 2, 1.5 10, 0.5 10, 0.5 2))') && (((_Bounded_ = true) && (POINT >= '1f20004a1400000000' && POINT <= '1f20004a1fffffffff')) || ((_Bounded_ = true) && (POINT >= '1f20004a2000000000' && POINT <= '1f20004a2fffffffff')) || ((_Bounded_ = true) && (POINT >= '1f20004a6000000000' && POINT <= '1f20004b0fffffffff')) || ((_Bounded_ = true) && (POINT >= '1f20004b3000000000' && POINT <= '1f20004b3fffffffff')) || ((_Bounded_ = true) && (POINT >= '1f20004b4000000000' && POINT <= '1f20004b47ffffffff')) || POINT == '1f20005b4000000000'))) && ((_Eval_ = true) && (FOO == 'bar' && (BAR == 'foo' || BAR == 'blah'))) && ((_Value_ = true) && ((_Bounded_ = true) && (NUM >= 0 && NUM <= 10))))
```

We would like to be able to pretty-print it to look something like this:
```
(
    (FIELD == 'some value') &&
    (
        (
            geowave:intersects(GEO, 'POLYGON((0.5 2, 1.5 2, 1.5 10, 0.5 10, 0.5 2))') && 
            ((_Bounded_ = true) && (GEO >= '2.0|0.5' && GEO <= '10.0|1.5'))
        ) || 
        (
            geowave:intersects(POINT, 'POLYGON((0.5 2, 1.5 2, 1.5 10, 0.5 10, 0.5 2))') && 
            (
[...]

PR #1588, opened 2022-06-01
Pretty-print a Jexl Query (#1425)

kevinrr888:
Created a new file JexlFormattedStringBuildingVisitor.java which functions the same as JexlStringBuildingVisitor.java except the queries are built in a formatted manner (split into multiple lines, indented). Also created a test file JexlFormattedStringBuildingVisitorTest.java with unit tests.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlFormattedStringBuildingVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlFormattedStringBuildingVisitor.java
@@ -1,74 +1,35 @@
 package datawave.query.jexl.visitors;
 
-import com.google.common.collect.Sets;
 import datawave.query.exceptions.DatawaveFatalQueryException;
 import datawave.query.jexl.JexlASTHelper;
 import datawave.query.jexl.nodes.BoundedRange;
 import datawave.query.jexl.nodes.QueryPropertyMarker;
 import datawave.webservice.query.exception.DatawaveErrorCode;
 import datawave.webservice.query.exception.QueryException;
-import org.apache.commons.jexl2.parser.ASTAdditiveNode;
-import org.apache.commons.jexl2.parser.ASTAdditiveOperator;
 import org.apache.commons.jexl2.parser.ASTAndNode;
-import org.apache.commons.jexl2.parser.ASTAssignment;
-import org.apache.commons.jexl2.parser.ASTDivNode;
-import org.apache.commons.jexl2.parser.ASTEQNode;
-import org.apache.commons.jexl2.parser.ASTERNode;
-import org.apache.commons.jexl2.parser.ASTFalseNode;
-import org.apache.commons.jexl2.parser.ASTFunctionNode;
-import org.apache.commons.jexl2.parser.ASTGENode;
-import org.apache.commons.jexl2.parser.ASTGTNode;
-import org.apache.commons.jexl2.parser.ASTIdentifier;
 import org.apache.commons.jexl2.parser.ASTJexlScript;
-import org.apache.commons.jexl2.parser.ASTLENode;
-import org.apache.commons.jexl2.parser.ASTLTNode;
-import org.apache.commons.jexl2.parser.ASTMethodNode;
-import org.apache.commons.jexl2.parser.ASTModNode;
-import org.apache.commons.jexl2.parser.ASTMulNode;
-import org.apache.commons.jexl2.parser.ASTNENode;
-import org.apache.commons.jexl2.parser.ASTNRNode;
-import org.apache.commons.jexl2.parser.ASTNotNode;
-import org.apache.commons.jexl2.parser.ASTNullLiteral;
-import org.apache.commons.jexl2.parser.ASTNumberLiteral;
 import org.apache.commons.jexl2.parser.ASTOrNode;
 import org.apache.commons.jexl2.parser.ASTReference;
 import org.apache.commons.jexl2.parser.ASTReferenceExpression;
-import org.apache.commons.jexl2.parser.ASTSizeMethod;
-import org.apache.commons.jexl2.parser.ASTStringLiteral;
-import org.apache.commons.jexl2.parser.ASTTrueNode;
-import org.apache.commons.jexl2.parser.ASTUnaryMinusNode;
 import org.apache.commons.jexl2.parser.JexlNode;
 import org.apache.commons.jexl2.parser.ParseException;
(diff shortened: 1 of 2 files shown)


warehouse/query-core/src/test/java/datawave/query/jexl/visitors/JexlFormattedStringBuildingVisitorTest.java
```diff
@@ -0,0 +1,349 @@
...
+	public void testNestedMarkerNodes() throws ParseException {
+		// Nested marker nodes anded with a single term or range, should be printed as a single line
+		String query = "((_Value_ = true) && ((_Bounded_ = true) && (NUM >= 0 && NUM <= 10)))";
+		String expected = "((_Value_ = true) && ((_Bounded_ = true) && (NUM >= 0 && NUM <= 10)))\n";
+		JexlNode node = JexlASTHelper.parseJexlQuery(query);
+        String builtQuery = JexlFormattedStringBuildingVisitor.buildQuery(node);
+        
+        Assert.assertEquals(expected, builtQuery);
+	}
+}
```
jwomeara, line 224: I haven't looked at the implementation yet, but I ran a few test queries against this.  For the most part, it works really well, but can you add support for this test case as well?  

```suggestion
    @Test
    public void testMultiParensIndentation() throws ParseException {
        String query = "((((FOO == 'BAR' && ((_Eval_ = true) && (THIS == 'THAT'))))))";
        String expected = "((((\n\tFOO == 'BAR' && \n\t((_Eval_ = true) && (THIS == 'THAT'))\n))))\n";
        JexlNode node = JexlASTHelper.parseJexlQuery(query);
        String builtQuery = JexlFormattedStringBuildingVisitor.buildQuery(node);

        Assert.assertEquals(expected, builtQuery);
    }
}
```

jwomeara (changes requested):
I haven't looked at the implementation yet, but I found an additional test case we should cover (see additional comments).

A few other notes: I would prefer that we use spaces in lieu of tabs.  4 spaces seems like a good default, but this could be configurable.  I also would not put a newline at the end of each query string.  

Again, I haven't checked out the implementation yet, but this looks very well done at first glance.  I will provide more feedback once I inspect the code.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/JexlFormattedStringBuildingVisitorTest.java
kevinrr888: Thank you for your comments and extra test case! Currently working on fixing the check-code-formatting test failure. I will work to fix my code to use 4 spaces instead of tabs. I will also fix it so there is no new line at the end of the query, and will add support for your given test case.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlFormattedStringBuildingVisitor.java
```diff
@@ -0,0 +1,762 @@
...
+        }
+        sb.append(String.join(" && " + (needNewLines ? NEWLINE : ""), childStringsFormatted));
+        
+        if (wrapIt)
+            sb.append(")");
+        
+        return data;
+    }
+    
+    public Object visit(ASTEQNode node, Object data) {
```
ivakegg: I believe almost all of the methods below this point are unchanged from the JexlStringBuildingVisitor.  It would be better to extend that class and then we would not need all of these copied methods.

ivakegg (changes requested):

ivakegg (dismissed):

ivakegg:
I realize this is a little bit of a scope creep, but could you add a main method such that this can be invoked from the command line?

ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlFormattedStringBuildingVisitor.java
```diff
@@ -0,0 +1,321 @@
...
+        }
+        return sb;
+    }
+    
+    public static void main(String args[]) {
+        if (args.length != 1) {
+            System.out.println("Invalid args. Must give 1 argument (the query).");
+        } else {
+            try {
+                System.out.println("Given Query:");
```
ivakegg: Remove these three lines.  Just print out the formatted query.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlFormattedStringBuildingVisitor.java
```diff
@@ -0,0 +1,321 @@
...
+            sb.setLength(sb.length() - 1);
+        } else {
+            sb.append((needNewLines ? NEWLINE : "") + ")");
+        }
+        return sb;
+    }
+    
+    public static void main(String args[]) {
+        if (args.length != 1) {
+            System.out.println("Invalid args. Must give 1 argument (the query).");
```
ivakegg: If there are no arguments, then read the query from stdin.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlFormattedStringBuildingVisitor.java
```diff
@@ -0,0 +1,326 @@
...
+        if (args.length != 1) {
+            Scanner scanner = new Scanner(System.in);
+            System.out.println("Enter Query:");
+            query = scanner.nextLine();
+            scanner.close();
+        } else {
+            query = args[0];
+        }
+        try {
+            System.out.println("Formatted Query:");
```
ivakegg: Please remove this "Formatted Query:" line.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlFormattedStringBuildingVisitor.java
```diff
@@ -0,0 +1,326 @@
...
+        }
+        return sb;
+    }
+    
+    public static void main(String args[]) {
+        String query;
+        
+        if (args.length != 1) {
+            Scanner scanner = new Scanner(System.in);
+            System.out.println("Enter Query:");
```
ivakegg: Please remove this "Enter Query:" line.  I want to be able to do something like:
```
cat some.file.txt | java JexlFormattedStringBuildingVisitor
```
and have the formatted query come out the other side.

ivakegg (changes requested):

Issue #1589, opened 2022-06-01
RewriteNullFunctionsVisitor destroys QueryPropertyMarkers


apmoriarty:
The RewriteNullFunctionsVisitor will destroy query property markers when flattening a junction.

PR #1590, opened 2022-06-01
RewriteNullFunctionsVisitor should preserve query property marker nodes

apmoriarty:
Addresses https://github.com/NationalSecurityAgency/datawave/issues/1589


FineAndDandy (dismissed):

jwomeara (approved):
Looks good.  I appreciate the addition of the query string comparison as that should catch those cases where parens are dropped when they shouldn't be.

FineAndDandy (approved):

apmoriarty:
> Looks good. I appreciate the addition of the query string comparison as that should catch those cases where parens are dropped when they shouldn't be.

Of course, I appreciate you pointing that out!

PR #1595, opened 2022-06-02
fixes #1594: Updated unique transform to avoid resetting bloom filter

### LINKED ISSUE #1594: UNIQUE mechanism appears to only work within pages
The bloom filter required for the unique transform appears to be resetting in between pages.  This needs to be fixed

ivakegg:
when update config is called if nothing actually changed.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
+++ b/warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
@@ -79,11 +79,16 @@ public class UniqueTransform extends DocumentTransform.DefaultDocumentTransform 
     }
     
     public void updateConfig(UniqueFields uniqueFields, QueryModel model) {
-        this.uniqueFields = uniqueFields;
-        this.uniqueFields.deconstructIdentifierFields();
-        this.bloom = BloomFilter.create(new ByteFunnel(), 500000, 1e-15);
-        if (log.isTraceEnabled()) {
-            log.trace("unique fields: " + this.uniqueFields.getFields());
+        if (this.uniqueFields != uniqueFields) {
+            uniqueFields.deconstructIdentifierFields();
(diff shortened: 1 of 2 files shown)


jschmidt10 (approved):

hlgp (approved):

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -79,11 +79,16 @@ public UniqueTransform(BaseQueryLogic<Entry<Key,Value>> logic, UniqueFields uniq
     }
     
     public void updateConfig(UniqueFields uniqueFields, QueryModel model) {
-        this.uniqueFields = uniqueFields;
-        this.uniqueFields.deconstructIdentifierFields();
-        this.bloom = BloomFilter.create(new ByteFunnel(), 500000, 1e-15);
-        if (log.isTraceEnabled()) {
-            log.trace("unique fields: " + this.uniqueFields.getFields());
+        if (this.uniqueFields != uniqueFields) {
+            uniqueFields.deconstructIdentifierFields();
```
alerman, line 83: In looking at this PR i was investigating this method call. It looks like in https://github.com/NationalSecurityAgency/datawave/blob/2a8be3f1b0d542ec68f50115964498d6d8cec9d3/warehouse/query-core/src/main/java/datawave/query/attributes/UniqueFields.java#L203

we could simplify that for loop. Not necessarily part of this PR, but could that for loop not be simplified to 

`for (String field : fieldMap.keySet()) {
            String newField = JexlASTHelper.deconstructIdentifier(field);
            newFieldMap.putAll(newField, fieldMap.get(field));
        }`

ivakegg: +1

PR #1600, opened 2022-06-03
WIP: Feature/update iterator builder 1578

### LINKED ISSUE #1578: Iterator Builder should handle exceeded term thresholds
Currently iterator building does not handle exceeded term markers and will fail queries. We should implement a way to convert query terms that have exceeded the system limit into Ivarators. When determining whether or not a query is executable, we should continue to mark as NON_EXEC when comparing against the global index, but EXEC when comparing against the field index. 

Note: This is the long-term fix to #1560 and reverts the query marking to ERROR

mineralntl:


warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -2416,10 +2415,11 @@ public Tuple2<CloseableIterable<QueryPlan>,Boolean> getQueryRanges(ScannerFactor
             }
             
             // if a term threshold is exceeded and we cannot handle that, then
-            // throw unsupported
+            // force a full table scan
             boolean thresholdExceeded = StreamContext.EXCEEDED_TERM_THRESHOLD.equals(stream.context());
-            if (thresholdExceeded && !config.canHandleExceededTermThreshold()) {
-                throw new UnsupportedOperationException(EXCEED_TERM_EXPANSION_ERROR);
+            if (IvaratorRequiredVisitor.isIvaratorRequired(queryTree) && thresholdExceeded && !config.canHandleExceededTermThreshold()) {
+                log.debug("Needs full table scan because we exceeded the term threshold and config.canHandleExceededTermThreshold() is false");
```
wgtong: seems like there's a `fullTableScanReason` variable you can set as well

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
```diff
@@ -236,7 +234,32 @@ public Object visit(ASTJexlScript node, Object data) {
     @Override
     public Object visit(ASTAndNode and, Object data) {
         QueryPropertyMarker.Instance instance = QueryPropertyMarker.findInstance(and);
-        if (instance.isType(ExceededOrThresholdMarkerJexlNode.class)) {
+        if (instance.isType(ExceededTermThresholdMarkerJexlNode.class)) {
```
wgtong, line 237: will the `ExceededTermThresholdMarkerJexlNode` class need to be handled differently by the `BoundedRangeIndexExpansionVisitor` and other usages such as `RangeStreamInfo`, `EntryParser` now? or it is fine since it is still delayed

PR #1603, opened 2022-06-07
Add visitor to ensure filter functions do not have index-only fields as arguments

### LINKED ISSUE #1602: Reject Index-Only Fields for Filter Functions
Per the 'query_help.html' document in reference to filter functions

```
None of these filter functions can be applied against index-only fields except for filter:text (#TEXT).
```
https://github.com/NationalSecurityAgency/datawave/pull/1534 already handles the #TEXT function. 

This work handles failing a query if a filter function has an index-only field as an argument.

For example the query `FOO == 'bar' && filter:includeRegex(BODY, 'ba.*')` will never evaluate correctly as no term will aggregate values for the BODY field into the document. It would be better to fail up front in the planning stage than let malformed queries execute.

apmoriarty:
Addresses https://github.com/NationalSecurityAgency/datawave/issues/1602


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidateFilterFunctionVisitor.java
```diff
@@ -0,0 +1,412 @@
...
+                        throw new DatawaveFatalQueryException(qe);
+                    }
+                }
+            }
+        }
+    }
+    
+    // pass-through
+    
+    @Override
```
lbschanno, line 142: I'm curious about why you're overriding these functions when pass-through is already the default behavior in BaseVisitor. Is it mainly to document which nodes are pass-through vs. not?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidateFilterFunctionVisitor.java
```diff
@@ -0,0 +1,412 @@
...
+                identifiers = JexlASTHelper.getIdentifierNames(arg);
+                for (String identifier : identifiers) {
+                    if (indexOnlyFields.contains(identifier)) {
+                        
+                        if (log.isDebugEnabled()) {
+                            log.debug("Filter function contained index-only field [" + identifier + "] for node ["
+                                            + JexlStringBuildingVisitor.buildQueryWithoutParse(node) + "]");
+                        }
+                        
+                        String msg = "Filter function cannot evaluate against index-only field [" + identifier + "]";
```
lbschanno, line 131: Can we include the filter function name in this message make it easier for client-side debugging?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidateFilterFunctionVisitor.java
```diff
@@ -0,0 +1,412 @@
...
+    private void validateFunction(ASTFunctionNode node) {
+        FunctionJexlNodeVisitor visitor = new FunctionJexlNodeVisitor();
+        node.jjtAccept(visitor, null);
+        
+        Set<String> identifiers;
+        if (visitor.namespace().equals(EVAL_PHASE_FUNCTION_NAMESPACE) && !excludedFunctions.contains(visitor.name())) {
+            for (JexlNode arg : visitor.args()) {
+                identifiers = JexlASTHelper.getIdentifierNames(arg);
+                for (String identifier : identifiers) {
+                    if (indexOnlyFields.contains(identifier)) {
```
lbschanno, line 124: Possibly a silly question, but is it ever possible for indexOnlyFields to be empty? If so, then the entire visitor could simply return early without traversing the query tree since there's nothing to check against.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidateFilterFunctionVisitor.java
```diff
@@ -0,0 +1,412 @@
...
+ * The exceptions to this rule are the #NO_EXPANSION and #TEXT functions.
+ * <p>
+ * The #NO_EXPANSION term will be pruned from the query prior to this visitor's execution, and the #TEXT query will have an index query added into the query.
+ */
+public class ValidateFilterFunctionVisitor extends BaseVisitor {
+    
+    private static final Logger log = ThreadConfigurableLogger.getLogger(ValidateFilterFunctionVisitor.class);
+    
+    private final Set<String> indexOnlyFields;
+    private final Set<String> excludedFunctions = Sets.newHashSet(NO_EXPANSION, INCLUDE_TEXT);
```
lbschanno: This can be a static final immutable set instantiated via `Collections.unmodifiableSet()`. Additionally, I'd like to see a comment indicating that these are filter functions that support the usage of index-only fields to make it clear why these are excluded from the validation.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/ValidateFilterFunctionVisitorTest.java
```diff
@@ -0,0 +1,348 @@
...
+import org.apache.commons.jexl2.parser.ASTJexlScript;
+import org.apache.commons.jexl2.parser.ParseException;
+import org.junit.Test;
+
+import java.util.Set;
+
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public class ValidateFilterFunctionVisitorTest {
```
lbschanno, line 15: Nice test coverage.

lbschanno (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidateFilterFunctionVisitor.java
apmoriarty: That is the precisely the case -- these overrides exist to serve as documentation for which nodes the visitor will pass through/descend/visit.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidateFilterFunctionVisitor.java
apmoriarty: Can do

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidateFilterFunctionVisitor.java
apmoriarty: After some thought I think that's a great change to make. Even though we expect index only fields to be present in nearly every case, here are few exceptions
- no index only fields generated for the date range of the query
- index only fields may be explicitly configured to be an empty set (i.e., turn off support for querying index only fields)

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidateFilterFunctionVisitor.java
apmoriarty: I'll make the set immutable and add comments. Also created https://github.com/NationalSecurityAgency/datawave/issues/1615 to move NOEXPANSION to query functions

apmoriarty: These functions are excluded because they are NOT actually supposed to be filter functions. The class level docs hint at why, but I can make it explicit for why these functions are excluded from validation. The short summary is
- #INCLUDE is explicitly allowed to function against index only fields
- #NOEXPANSION controls query model expansion, to include index only fields

lbschanno: Ah, gotcha, thanks for clarifying.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidateFilterFunctionVisitor.java
lbschanno: Okay, good to know.

hlgp (approved):

FineAndDandy (approved):

lbschanno (approved):

PR #1606, opened 2022-06-07
fixes #1601: Term excerpt performance

### LINKED ISSUE #1601: Hit Excerpts are slowing down queries too much even when not requested
The hit excerpts are causing a significant delay in processing phrases even when they are not being requested.  We believe this is because we are going through the trouble of adding phrase excerpts even when they are not needed during the phrase/within function evaluation.

ivakegg:
* Updated to avoid gathering phrase indexes/offsets when excerpts are not being requested.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
+++ b/warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
@@ -249,7 +249,7 @@ public class QueryIterator extends QueryOptions implements YieldingKeyValueItera
         // Parse the query
         try {
             this.script = JexlASTHelper.parseAndFlattenJexlQuery(this.getQuery());
-            this.myEvaluationFunction = new JexlEvaluation(this.getQuery(), arithmetic);
+            this.myEvaluationFunction = getJexlEvaluation(this.getQuery(), arithmetic);
             
         } catch (Exception e) {
             throw new IOException("Could not parse the JEXL query: '" + this.getQuery() + "'", e);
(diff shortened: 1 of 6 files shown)


lbschanno (dismissed):
LGTM.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1096,20 +1096,47 @@ private Range getDocumentRange(NestedQueryIterator<Key> documentSource) {
...
+                if (null == jexlEvaluationFunction) {
+                    jexlEvaluationFunction = new JexlEvaluation(query, arithmetic);
+                }
             }
         }
+        
+        // update the jexl evaluation to gather phrase offsets if required for excerpts
+        if (getExcerptFields() != null && !getExcerptFields().isEmpty()) {
+            jexlEvaluationFunction.setGatherPhraseOffsets(true);
+        }
```
apmoriarty, line 1138: Might want to create an issue to diff the excerpt fields against the fields in the query's content function. It's possibly (though unlikely) that the content function fields in the query do not match the fields specified in the excerpt fields -- as the result of query model expansion and/or range stream pruning.

apmoriarty (dismissed):
LGTM -- commented for additional follow on work.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
ivakegg: #1607

PR #1612, opened 2022-06-09
Attribute comparator and supporting classes

### LINKED ISSUE #1447: Bloated JexlContext
When using the IndexOnlyContextCreator, the JexlContext is unnecessarily bloated. Setting a breakpoint at line 67 of the JexlContextCreator will show multiple entries under the context map. Half of these are from the FieldIndex with a datatype, and half are from the Content and do not include a datatype. Given that the attributes are functionally the same, there is no reason to store both. 

We should look into adding a method of fuzzy matching Attributes that prefers ones containing a datatype and uid.

mineralntl:
Replacing the previous branch with this one as the previous PR had languished to the point of rebasing becoming more work than it was worth.


warehouse/query-core/src/main/java/datawave/query/attributes/Document.java
```diff
@@ -280,77 +281,110 @@ public void put(String key, Attribute<?> value, Boolean includeGroupingContext,
...
+                        _count += attrs.size();
+                        if (trackSizes) {
+                            _bytes += attrs.sizeInBytes();
+                        }
+                    } else {
+                        // fuzzy matches found, attempt to combine attributes
+                        Set<Attribute<? extends Comparable<?>>> combinedSet = AttributeComparator.combineMultipleAttributes((Attributes) existingAttr,
+                                        (Attributes) value);
+                        Attributes mergedAttributes = new Attributes(combinedSet, this.isToKeep(), trackSizes);
+                        dict.put(key, mergedAttributes);
```
apmoriarty, line 311: `_count` and `_bytes` is not being updated

warehouse/query-core/src/main/java/datawave/query/attributes/Document.java
```diff
@@ -280,77 +281,110 @@ public void put(String key, Attribute<?> value, Boolean includeGroupingContext,
...
+                        _count += attrs.size();
+                        if (trackSizes) {
+                            _bytes += attrs.sizeInBytes();
+                        }
+                    } else {
+                        // fuzzy matches found, attempt to combine attributes
+                        Set<Attribute<? extends Comparable<?>>> combinedSet = AttributeComparator.combineMultipleAttributes((Attribute) existingAttr,
+                                        (Attributes) value, trackSizes);
+                        Attributes mergedAttributes = new Attributes(combinedSet, this.isToKeep(), trackSizes);
+                        dict.put(key, mergedAttributes);
```
apmoriarty, line 341: `_count` and `_bytes` are not updated

warehouse/query-core/src/main/java/datawave/query/util/AttributeComparator.java
```diff
@@ -0,0 +1,195 @@
...
+     */
+    public static boolean multipleToMultiple(final Attributes attrs1, final Attributes attrs2) {
+        boolean containsMatch = false;
+        for (Attribute<? extends Comparable<?>> newAttr : attrs2.getAttributes()) {
+            if (singleToMultiple(newAttr, attrs1)) {
+                containsMatch = true;
+            }
+        }
+        return containsMatch;
+    }
```
apmoriarty: Once you know a match exists there is no reason to continue looking for additional matches
```
    public static boolean multipleToMultiple(final Attributes attrs1, final Attributes attrs2) {
        for (Attribute<? extends Comparable<?>> newAttr : attrs2.getAttributes()) {
            if (singleToMultiple(newAttr, attrs1)) {
                return true;
            }
        }
        return false;
    }
```

warehouse/query-core/src/main/java/datawave/query/util/AttributeComparator.java
```diff
@@ -0,0 +1,195 @@
...
+    
+    /**
+     * Performs a basic match on the content/data of two given Attributes. If such a match is found, combine the metadata from the two Attributes and create new
+     * deduplicated Attributes to return to the Document. Preference is given to the Attribute found in the event, NOT the index. In the case that true
+     * duplicates are discovered we keep the Attribute that already exists within the Document, set the Attribute's toKeep() flag to true (if not already), and
+     * ignore the potential new Attribute being checked.
```
apmoriarty: If everything else is equal, is priority given to first collection?

apmoriarty (changes requested):
If you run ContentFunctionQueryTest.phraseTest and set your debugger to JexlEvaluation on line 81 and open up the document you will see that the BODY field has four attributes -- it should only have two.

I bring this case up because this is the exact case that prompted this work -- a prenormalized attribute exists in the field index and a content attribute exists in the term frequency column family. Both attributes are from the index but we should only keep one.

warehouse/query-core/src/main/java/datawave/query/attributes/Document.java
mineralntl: Noted and updated

warehouse/query-core/src/main/java/datawave/query/attributes/Document.java
mineralntl: Noted and updated

warehouse/query-core/src/main/java/datawave/query/util/AttributeComparator.java
mineralntl: Noted and simplified - thanks for the suggestion

warehouse/query-core/src/main/java/datawave/query/util/AttributeComparator.java
mineralntl: Yes, that's correct - shall I add a note about that in the docs?

mineralntl: Comment added to docs.

mineralntl:
Updated and simplified comparison to ensure original "unique" attributes are not reused and added to Context upon merging.

warehouse/query-core/src/test/java/datawave/query/util/AttributeComparatorTest.java
```diff
@@ -0,0 +1,142 @@
...
+    private final Key testKey1 = new Key(new Text("20000101_69"), new Text("foo%00;d8zay2.-3pnndm.-anolok"), new Text(), new ColumnVisibility("ALL"),
+                    946684800000L);
+    private final Key testKey2 = new Key(new Text("20000101_69"), new Text("bar%00;d8zay2.-3pnndm.-anolok"), new Text(), new ColumnVisibility("ALL"),
+                    946684800000L);
+    private final Key testKey3 = new Key(new Text("20000101_69"), new Text("baz%00;d8zay2.-3pnndm.-anolok"), new Text(), new ColumnVisibility("ALL"),
+                    946684800000L);
+    private final Key testKey4 = new Key(new Text("20000101_69"), new Text("foo"), new Text(), new ColumnVisibility("ALL"), 946684800000L);
+    private final Key testKey5 = new Key(new Text("20000101_69"), new Text("bar"), new Text(), new ColumnVisibility("ALL"), 946684800000L);
+    private final Key testKey6 = new Key(new Text("20000101_69"), new Text("baz"), new Text(), new ColumnVisibility("ALL"), 946684800000L);
+    private final Key testKey7 = new Key(new Text("192.168.1.1"), new Text("barf"), new Text(), new ColumnVisibility("ALL"), 946684800000L);
```
apmoriarty: Extract common parts to variables. This will make it much easier to tell at a glance what differences are being compared
```
private final Text row = new Text("20000101_69");
private final Text cq = new Text(); // empty column qualifier
private final ColumnVisibility cv = new ColumnVisibility("ALL");
private final long ts = 946684800000L;

private final Key k1= new Key(row, new Text("foo%00;d8zay2.-3pnndm.-anolok"), cq, cv, ts);
private final Key k2= new Key(row, new Text("bar%00;d8zay2.-3pnndm.-anolok"), cq, cv, ts);
```

warehouse/query-core/src/test/java/datawave/query/attributes/DocumentTest.java
```diff
@@ -0,0 +1,307 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.HashSet;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class DocumentTest {
+    private final Key testKey1 = new Key(new Text("20000101_69"), new Text("foo%00;d8zay2.-3pnndm.-anolok"), new Text(), new ColumnVisibility("ALL"),
```
apmoriarty: Same comment as before, extract common parts to variables

warehouse/query-core/src/test/java/datawave/query/util/AttributeComparatorTest.java
```diff
@@ -0,0 +1,142 @@
...
+    
+    @Test
+    public void testCombineAttributesToAttributes() {
+        Set<Attribute<? extends Comparable<?>>> mergedSet = AttributeComparator.combineMultipleAttributes(preNormAttributes, contentAttributes);
+        assertThat(mergedSet, CoreMatchers.hasItem(combinedContent1));
+        assertThat(mergedSet, CoreMatchers.hasItem(combinedContent2));
+        assertThat(mergedSet, CoreMatchers.hasItem(combinedContent3));
+        assertEquals(3, mergedSet.size());
+    }
+
```
apmoriarty: If two keys have different metadata they should not be considered equivalent. Consider the TLD case where you have a parent and child, or two different children. The two attributes for the common value 'foo' have different metadata keys, but are considered equivalent and will reduce to a single attribute

```
    @Test
    public void testTldCase(){
        Text row = new Text("20000101_69");
        ColumnVisibility cv = new ColumnVisibility("ALL");
        Text cq = new Text();   //  empty column qualifier for content attrs
        long ts = 946684800000L;

        // parent / child
        Key k1 = new Key(row, new Text("datatype%00;d8zay2.-3pnndm.-anolok"), cq, cv, ts);
        Key k2 = new Key(row, new Text("datatype%00;d8zay2.-3pnndm.-anolok.1"), cq, cv, ts);

        // different children
        //Key k1 = new Key(row, new Text("datatype%00;d8zay2.-3pnndm.-anolok.1"), cq, cv, ts);
        //Key k2 = new Key(row, new Text("datatype%00;d8zay2.-3pnndm.-anolok.2"), cq, cv, ts);

        Content content1 = new Content("foo", k1, true);
        Content content2 = new Content("foo", k2, true);

        //  the value of Content.getData() matches, along with the column visibility and timestamp
        assertTrue(AttributeComparator.singleToSingle(content1, content2));

        Attribute combined = AttributeComparator.combineSingleAttributes(content1, content2);
        int i = 0; // breakpoint here to see the combined attributed
    }
```

This also raises up an interesting case when a parent PreNormalizedAttr is compared and combined with a child Content attr. The PreNormalizedAttr is kept when both should be returned in an instance of `Attributes`
```
        Key k1 = new Key(row, new Text("datatype%00;d8zay2.-3pnndm.-anolok"), cq, cv, ts);
        Key k2 = new Key(row, new Text("datatype%00;d8zay2.-3pnndm.-anolok.1"), cq, cv, ts);

        PreNormalizedAttribute content1 = new PreNormalizedAttribute ("foo", k1, true);
        Content content2 = new Content("foo", k2, true);

        // still considered equal, when combined the prenormalized attribute is kept
```

warehouse/query-core/src/main/java/datawave/query/util/AttributeComparator.java
```diff
@@ -0,0 +1,203 @@
...
+     * Compares two individual Attribute's flags and metadata, preferring to keep the Attribute from the event rather than the index. If the origin of the
+     * Attributes being compared is identical, prefer to keep the one with the toKeep() flag set.
+     *
+     * @param attr1
+     *            An Attribute against which we want to compare
+     * @param attr2
+     *            Another Attribute against which we want to check for likeness
+     * @return
+     */
+    public static Attribute combineSingleAttributes(final Attribute attr1, final Attribute attr2) {
```
apmoriarty: Is this method ever called with two logically different attributes? In that case should this return an instance of Attributes that contains both of the args?

warehouse/query-core/src/main/java/datawave/query/util/AttributeComparator.java
```diff
@@ -0,0 +1,203 @@
...
+        int attr1mdSize = attr1.getMetadata().getSize();
+        int attr2mdSize = attr2.getMetadata().getSize();
```
apmoriarty: Why compare on metadata size? The metadata key's column family should contain the datatype and uid which we absolutely want to compare.

warehouse/query-core/src/main/java/datawave/query/util/AttributeComparator.java
```diff
@@ -0,0 +1,203 @@
...
+
+public final class AttributeComparator {
+    
+    // this is a utility and should never be instantiated
+    private AttributeComparator() {
+        throw new UnsupportedOperationException();
+    }
+    
+    /**
+     * Performs a simple comparison on a set of Attribute's data and some parts of their metadata
```
apmoriarty: Be explicit about what is considered for  equality, this could be rewritten to `Compares an Attribute's data, column visibility, and timestamp`. Realistically we should also compare the metadata for reasons detailed below

warehouse/query-core/src/main/java/datawave/query/util/AttributeComparator.java
```diff
@@ -0,0 +1,203 @@
...
+import datawave.query.attributes.Attributes;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public final class AttributeComparator {
```
apmoriarty, line 25: Would like to see class level docs that explain the following..
- why are we comparing attributes in the first place?
- what determines when two attributes equivalent
- perhaps add a simple example

warehouse/query-core/src/main/java/datawave/query/util/AttributeComparator.java
```diff
@@ -0,0 +1,203 @@
...
+     *
+     * @param attr1
+     *            An Attribute against which we want to compare
+     * @param attr2
+     *            Another Attribute against which we want to check for likeness
+     * @return
+     */
+    public static boolean singleToSingle(final Attribute attr1, final Attribute attr2) {
+        return (attr1.getData().equals(attr2.getData()) && attr1.getColumnVisibility().equals(attr2.getColumnVisibility()) && attr1.getTimestamp() == attr2
+                        .getTimestamp());
```
apmoriarty:
```suggestion
        //  @formatter:off
        return attr1.getData().equals(attr2.getData()) &&
                attr1.getColumnVisibility().equals(attr2.getColumnVisibility()) &&
                attr1.getTimestamp() == attr2.getTimestamp();
        //  @formatter:on
```

apmoriarty (changes requested):

warehouse/query-core/src/test/java/datawave/query/util/AttributeComparatorTest.java
mineralntl: Updated as requested

warehouse/query-core/src/test/java/datawave/query/attributes/DocumentTest.java
mineralntl: Updated as requested

warehouse/query-core/src/main/java/datawave/query/util/AttributeComparator.java
mineralntl: Updated in new commit

warehouse/query-core/src/main/java/datawave/query/util/AttributeComparator.java
mineralntl: Removed in new commit

warehouse/query-core/src/main/java/datawave/query/util/AttributeComparator.java
mineralntl: Pathway into this method has been updated to ensure it is only ever called when we know we have distinctly different metadata - specifically differently populated column family information.

warehouse/query-core/src/test/java/datawave/query/util/AttributeComparatorTest.java
mineralntl: Updated algorithm to account for these cases and will add examples to unit tests.

warehouse/query-core/src/main/java/datawave/query/util/AttributeComparator.java
mineralntl: Added in recent commit

warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
```diff
@@ -45,6 +45,14 @@ public boolean isMetadataSet() {
         return (metadata != null);
     }
     
+    public void setColumnFamily(Text columnFamily) {
```
apmoriarty, line 54: Calling this method always changes the backing metadata for an attribute -- this deserves a mention in the method level docs that are forthcoming.

warehouse/query-core/src/test/java/datawave/query/attributes/DocumentTest.java
```diff
@@ -0,0 +1,284 @@
...
+import org.apache.hadoop.io.Text;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.HashSet;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class DocumentTest {
```
apmoriarty, line 18: Remove values like 'lulz' and 'barf'.

warehouse/query-core/src/test/java/datawave/query/attributes/DocumentTest.java
```diff
@@ -0,0 +1,284 @@
...
+    private final Text row = new Text("20000101_69");
```
apmoriarty: Almost expected to see shard 420

apmoriarty (changes requested):
Almost there! All the requested changes are miner in scope

warehouse/query-core/src/test/java/datawave/query/attributes/DocumentTest.java
mineralntl: Done and done

warehouse/query-core/src/test/java/datawave/query/attributes/DocumentTest.java
mineralntl: Changed

warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
mineralntl: Noted

hlgp:
OBE

Issue #1624, opened 2022-06-14
Query metric query formatting - colors, themes, and parens matching


ivakegg:
I would be nice to be able to hover over a paren in the query or query plan and to see where the matching parenthesis is.  Both by highlighting and then if you click on the paren it takes you to the matching one.
The thought is that this can be done with a decorator on the jexl formatted string visitor such that we can apply other decorators for other circumstances.  For example a terminal (bash) decorator that applies coloring based on depth of the node or other such decorations.

jwomeara:
Parens matching would be great.  You could probably get away with doing that solely in javascript on the metrics html page.

To broaden this idea a bit, it might be nice to identify the various elements of a query (e.g. fields, numeric values, quoted values, geohashes, functions, query property markers, etc.) and come up with a color scheme that could be used to format each of these elements.  Ideally, we would be able to easily swap between different themes and color schemes (some people might find one theme more readable than another).

kevinrr888:
Created a commit for the interactive matching parenthesis functionality. See: https://github.com/NationalSecurityAgency/datawave-query-metric-service/pull/5

kevinrr888:
Updated https://github.com/NationalSecurityAgency/datawave-query-metric-service/pull/5 and created https://github.com/NationalSecurityAgency/datawave/pull/1631 for completed functionality. The updated JexlFormattedStringBuildingVisitor.java will have to be added to feature/queryMicroservices for it to work with the microservices Query Metrics page. Let me know any comments/changes.

PR #1631, opened 2022-06-24
Support for further query, query plan formatting (issue #1624)

kevinrr888:
Changed JexlFormattedStringBuildingVisitor.java to have the option of including HTML for the various query elements. Updated the screen.css file to have support for the HTML elements added in JexlFormattedStringBuildingVisitor (this screen.css file is used in the webserver's Query Metrics page). Moved the call of formatting the metrics to a more appropriate location.


ivakegg:
I was hoping that you would use a decorator pattern instead of having if (includeHtml) scattered throughout the code.
What that means is that you would have a call to decorator.apply(String text, TYPE type) at the end of each one of those visitor functions.
The decorator would implement your own interface that has an apply(String, TYPE) method.
Then all of the html stuff would be implemented within an HtmlDecorator.
And then the next step would be to create a BashDecorator that would also do the coloring stuff but instead for a bash shell.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlFormattedStringBuildingVisitor.java
```diff
@@ -25,17 +57,31 @@
...
- * new, indented line. Same functionality as JexlStringBuildingVisitor.java except the query is built in a formatted manner.
+ * new, indented line. Additionally, there is the option to decorate the query with added color for easier readability in terminal or on web pages. Same
+ * functionality as JexlStringBuildingVisitor.java except the query is built in a formatted manner.
  */
 public class JexlFormattedStringBuildingVisitor extends JexlStringBuildingVisitor {
     protected static final String NEWLINE = System.getProperty("line.separator");
+    protected static final String DOUBLE_QUOTE = "\"";
+    
+    // allowed methods for composition. Nothing that mutates the collection is allowed, thus we have:
+    private Set<String> allowedMethods = Sets.newHashSet("contains", "retainAll", "containsAll", "isEmpty", "size", "equals", "hashCode", "getValueForGroup",
```
ivakegg: This list of methods seems a little fragile to me.  I wonder if this list should be maintained in the FunctionSet class instead such that if we add methods there we might have a chance at remembering to update this list.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlFormattedStringBuildingVisitor.java
```diff
@@ -88,6 +134,9 @@ private static String formatBuiltQuery(String query) {
      * @return
      */
     private static boolean containsOnly(String str, char ch) {
+        str = str.replaceAll("<[^>]*>", ""); // remove all existing html from the string (if present)
```
ivakegg: Can we move the "remove all existing html" and "remove bash color" stuff into the decorator classes?  That way these two lines would reduce to something like decorator.removeDecorations(str).

ivakegg: Alternatively we could build the decorated and undecorated variants at the same time which would avoid this somewhat fragile decoration removal stuff.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlFormattedStringBuildingVisitor.java
```diff
@@ -281,14 +388,372 @@ public Object visit(ASTAndNode node, Object data) {
...
+        
+        int numChildren = node.jjtGetNumChildren();
+        
+        if (2 != numChildren) {
+            throw new IllegalArgumentException("An ASTEQNode has more than two children");
+        }
+        
+        node.jjtGetChild(0).jjtAccept(this, sb);
+        
+        decorator.apply(sb, node);
```
ivakegg: beautiful!

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlFormattedStringBuildingVisitor.java
```diff
@@ -281,14 +388,372 @@ public Object visit(ASTAndNode node, Object data) {
...
+            } else if (2 < i) {
+                sb.append(", ");
+            }
+            
+            decorator.apply(sb, node, i);
+            
+            node.jjtGetChild(i).jjtAccept(this, sb);
+            
+            if (i == 0 || i == 1) {
+                if (decorator instanceof HtmlDecorator) {
```
ivakegg: These pieces of code should be encoded as a separate decorator function.  There should be nothing in this class that does something different based on which decorator is being used.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlFormattedStringBuildingVisitor.java
```diff
@@ -367,9 +926,18 @@ public static void main(String args[]) {
             query = args[0];
         }
         try {
-            System.out.println(JexlFormattedStringBuildingVisitor.buildQuery(JexlASTHelper.parseJexlQuery(query)));
+            String echoCommand = "echo " + DOUBLE_QUOTE + buildQueryWithBashFormatting(JexlASTHelper.parseJexlQuery(query)) + DOUBLE_QUOTE;
```
ivakegg: OK, it would be nice if the user could supply a parameter to choose html formatting or plain text formatting.  Additionally perhaps the user could supply their own class to be used for decorating.

ivakegg (changes requested):
What we also need are unit test cases for all of this great work please.

ivakegg (dismissed):

kevinrr888:
Added the script to run JexlFormattedStringBuildingVisitor from the command line. 

Script removes line 4 of ingest-libs.sh ("`. ../ingest/ingest-env.sh`") before running it and then adds it back after to avoid error of not finding ingest-passwd.sh. This is probably not the best way to go about this, any ideas of a better way would be great. 

Also, the cd command could probably be changed. How can I get the local datawave directory of the user to use that instead?

warehouse/ingest-scripts/src/main/resources/bin/util/prettyPrintQuery.sh
```diff
@@ -0,0 +1,21 @@
...
+echo "Decorator can be EmptyDecorator, HtmlDecorator, or BashDecorator"
+
+cd ../../../../../../../contrib/datawave-quickstart/bin/services/datawave/datawave-ingest-install/bin/ingest
+
+findJars () {
+    ls -1 ../../lib/$1-*.jar | sort | paste -sd ':' -
+}
+
+# Delete line 4 of ./ingest-libs.sh: ". ../ingest/ingest-env.sh"
+sed -i '4d' ./ingest-libs.sh
```
ivakegg: I see you are trying to be fancy here.  However normally this would not be an issue.  Please remove the sed line here and below.

ivakegg (changes requested):

warehouse/ingest-scripts/src/main/resources/bin/util/prettyPrintQuery.sh
kevinrr888: Okay, it was an issue for me but if it's only on my end then its no problem. I'll delete those and you can let me know if it poses a problem on your end

ivakegg (dismissed):

ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/BashDecorator.java
```diff
@@ -0,0 +1,221 @@
...
+    protected static final String STRING_COLOR = "\\e[38;5;33m"; // light blue
+    
+    @Override
+    public void apply(StringBuilder sb, ASTAdditiveOperator node) {
+        sb.append(ARITHMETIC_OP_COLOR + node.image + NC);
+    }
+    
+    @Override
+    public void apply(StringBuilder sb, ASTAndNode node, Collection<String> childStrings, boolean needNewLines) {
+        sb.append(String.join(AND_COLOR + " && " + NC + (needNewLines ? NEWLINE : ""), childStrings));
```
jwomeara: Probably not a big deal, but I would have only colored the '&&', not the spaces.  This doesn't matter when you're changing the font color, but it might look odd if we're coloring the background.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/HtmlDecorator.java
```diff
@@ -0,0 +1,197 @@
...
+import datawave.query.jexl.nodes.QueryPropertyMarker;
+
+/**
+ * Decorates a Jexl Query with HTML coloring. To be used in viewing Jexl Queries on web pages. Specifically, the Query Metrics webpage.
+ */
+public class HtmlDecorator implements JexlQueryDecorator {
+    
+    @Override
+    public void apply(StringBuilder sb, ASTAdditiveOperator node) {
+        sb.append(String.format("<span class=\"add-op\">%s</span>", node.image));
```
jwomeara: Probably could have saved yourself a few keystrokes by creating a method to create the CSS tag.

e.g.
```
String createSpan(String cssClass, String content) {
   return String.format("<span class=\"%s\">%s</span>", cssClass, content);
}
```

Then you just call `createSpan(css, content)` everywhere. It also might help if we ever decide to add additional formatting that affects every span.  Adding in one place vs. many.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/HtmlDecorator.java
```diff
@@ -0,0 +1,197 @@
...
+        if (i == 0)
+            sb.append("<span class=\"function-namespace\">");
+        else if (i == 1)
+            sb.append("<span class=\"function\">");
+        // the span elements are closed on visiting the ASTIdentifier children
+    }
+    
+    @Override
+    public void apply(StringBuilder sb, ASTGENode node) {
+        sb.append("<span class=\"greater-than-equal-op\"> >= </span>");
```
jwomeara: Same comment as before about applying formatting to whitespace.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlFormattedStringBuildingVisitor.java
```diff
@@ -1,79 +1,136 @@
...
 import java.util.Set;
 import java.util.TreeSet;
 
 /**
  * A Jexl visitor which builds an equivalent Jexl query in a formatted manner. Formatted meaning parenthesis are added on one line, and children are added on a
- * new, indented line. Same functionality as JexlStringBuildingVisitor.java except the query is built in a formatted manner.
+ * new, indented line. Additionally, there is the option to decorate the query with added color for easier readability in terminal or on web pages. Same
+ * functionality as JexlStringBuildingVisitor.java except the query is built in a formatted manner.
  */
 public class JexlFormattedStringBuildingVisitor extends JexlStringBuildingVisitor {
```
jwomeara: It looks like there's a lot of code in here copied directly from JexlStringBuildingVisitor.  Did you ever look into keeping all of this functionality in the original JexlStringBuildingVisitor, and adding configuration to turn on/off different formatting features?  My preference would be to avoid duplicating code if we can.

In the past, I have had to make changes to JexlStringBuildingVisitor to fix query issues that we run into (see all of the special handling throughout for backslashes in string literals).  If/when those kinds of issues arise in the future, my concern is that we'll make the updated in one class but not the other, and the logic will eventually drift between the two StringBuildingVisitors.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlFormattedStringBuildingVisitor.java
```diff
@@ -310,21 +708,139 @@ public Object visit(ASTReferenceExpression node, Object data) {
...
+        return sb;
+    }
+    
+    /**
+     * Returns a new list of QueryMetrics which is identical to the given list except the query and query plan for each metric are formatted according to
+     * {@link #buildDecoratedQuery(JexlNode, boolean, JexlQueryDecorator) buildDecoratedQuery} with an {@link HtmlDecorator}
+     * 
+     * @param metrics
+     */
+    public static void formatMetrics(List<? extends BaseQueryMetric> metrics) {
```
jwomeara: This seems like a weird home for this method.  Should this not just be a method in the ShardTableQueryMetricHandler?  The StringBuilder shouldn't have to concern itself with query metrics.

jwomeara (changes requested):
Looks pretty good.  Just a few comments.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/BashDecorator.java
kevinrr888: Yes, good idea. I'll change that and in other places where the white space is colored

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/HtmlDecorator.java
kevinrr888: Great idea. I'll fix that

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlFormattedStringBuildingVisitor.java
kevinrr888: Okay, yes I see the concern with this. At first JexlFormattedStringBuildingVisitor didn't have much code duplication but with how much I've had to change about it, its come to be similar to JexlStringBuildingVisitor. Do you think I should just try to move all the functionality into one class (JexlStringBuildingVisitor)?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlFormattedStringBuildingVisitor.java
kevinrr888: Yes I agree. It was meant to just be a temporary spot but I never got around to changing it. Thanks for pointing this out. I'll move it to ShardTableQueryMetricHandler.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlFormattedStringBuildingVisitor.java
jwomeara: Barring any obstacles I'm not seeing - yes, that is what I would do.

warehouse/ingest-scripts/src/main/resources/bin/util/prettyPrintQuery.sh
```diff
@@ -0,0 +1,19 @@
...
+
+findJacksonJars () {
+    ls -1 ../../lib/jackson-*.jar | sort | paste -sd ':' -
+}
+
+. ./ingest-libs.sh
+
+JACKSON_JARS=$(findJacksonJars)
+
+java -cp $CLASSPATH:$JACKSON_JARS datawave.query.jexl.visitors.JexlStringBuildingVisitor
```
ivakegg: I don't see the arguments being passed to the JexlStringBuildingVisitor.  I think adding $@ at the end of this line would do it.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlStringBuildingVisitor.java
```diff
@@ -637,8 +868,76 @@ public Object visit(ASTAssignment node, Object data) {
...
+    
+    /**
+     * Usage: "JexlStringBuildingVisitor query decorator-class-name" or "JexlStringBuildingVisitor query" or "JexlStringBuildingVisitor"
+     * 
+     * @param args
+     */
+    public static void main(String args[]) {
+        String query = null;
+        JexlQueryDecorator decorator = null;
+
```
ivakegg: This is looking pretty good.  I ask for a little change in how the parameters can be specified.  I am thinking a little closer to how xmllint may be used:
```
JexlStringBuildingVisitor [ --html --bash --help ] {JEXL-FILE(s)}

The JexlStringBuildingVisitor program parses one or more JEXL files, specified on the command
line as JEXL-FILE (or the standard input if no filenames are provided).  It prints various types of
output, depending upon the options selected.  It is useful for detecting errors in the JEXL query
and providing a formatted output.
```

warehouse/ingest-scripts/src/main/resources/bin/util/prettyPrintQuery.sh
ivakegg: Note that I believe unless overridden by the user, the default should be to format using the bash decorator.

warehouse/query-core/src/main/java/datawave/query/collections/FunctionalSet.java
```diff
@@ -22,6 +22,10 @@
     
     private static final long serialVersionUID = 8759102050325550844L;
     protected Set<T> delegate;
+    // allowed methods for composition. Nothing that mutates the collection is allowed, thus we have:
+    public static final Set<String> allowedMethods = Collections.unmodifiableSet(Sets.newHashSet("contains", "retainAll", "containsAll", "isEmpty", "size",
```
ivakegg: Thank you for moving there here.  This makes a lot more sense.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlStringBuildingVisitor.java
```diff
@@ -185,24 +383,30 @@ public Object visit(ASTOrNode node, Object data) {
...
         Collection<String> childStrings = (sortDedupeChildren) ? new TreeSet<>() : new ArrayList<>(numChildren);
+        Collection<String> childStringsFormatted = (sortDedupeChildren) ? new TreeSet<>() : new ArrayList<>(numChildren);
         StringBuilder childSB = new StringBuilder();
         for (int i = 0; i < numChildren; i++) {
             node.jjtGetChild(i).jjtAccept(this, childSB);
             childStrings.add(childSB.toString());
             childSB.setLength(0);
         }
-        sb.append(String.join(" && ", childStrings));
+        // If needNewLines is false, we should remove the new lines added to the child strings
```
ivakegg: I will have to say that this one scares me a little.  Since we have mergde the whole formatting thing into this one class (which is not a bad idea), we have to be very careful that the unformatted query path does not change the query.  If one of the nodes contains a term that actually has a newline in it, then I think that needs to be preserved.  So if buildMultipleLines is false, then we should never be adding any new lines in the first place.

ivakegg (changes requested):

warehouse/ingest-scripts/src/main/resources/bin/util/prettyPrintQuery.sh
kevinrr888: Ah I had it so the user types the arguments. I'll change that and I'll make the default use the bash decorator.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlStringBuildingVisitor.java
kevinrr888: Okay, I will work on changing that.

warehouse/query-core/src/main/java/datawave/query/collections/FunctionalSet.java
kevinrr888: Yup!

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlStringBuildingVisitor.java
kevinrr888: Good catch! I overlooked this. I will fix it

ivakegg (dismissed):

hlgp:
Closing for now.  Reopen when revived and deconflicted.

PR #1639, opened 2022-07-11
jsonNULLS-7914 Added default value for jsonNULLS to all-config as wel…

ndhudso:
…l as implemented logic to check for nulls and use the default value


hlgp:
OBE

PR #1653, opened 2022-07-20
fixes #1650: Updated the executability visitor

### LINKED ISSUE #1650: Full Table scan errors need to provide a reason why
When a query is deemed non-executable by the executability visitor, then the user will get a "full table scan required" message which does not help in determining how to modify the query.  Given the fix for #1558  fixed in #1566 , this may be occurring more often especially when model expansions are involved.

ivakegg:
* executability visitor provides a better query debug output
* Also provides a summary of the reason for a executability state
* Query planner will return the summary with the exception

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExecutableDeterminationVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExecutableDeterminationVisitor.java
@@ -188,9 +188,18 @@ public class ExecutableDeterminationVisitor extends BaseVisitor {
         Multimap<STATE,String> pop(Set<STATE> states);
         
         /**
+         * Move contributors from one state to another
+         * 
+         * @param state
+         * @param newState
+         */
+        void move(STATE state, STATE newState);
+        
+        /**
          * Take the top contributers map and summarize it adding that to the output.
          */
         void summarize();
+        
     }
     
     /**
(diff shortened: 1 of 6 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExecutableDeterminationVisitor.java
```diff
@@ -369,51 +516,24 @@ else if (states.contains(STATE.EXECUTABLE)) {
...
     }
     
     @Override
     public Object visit(ASTEQNode node, Object data) {
+        if (output != null) {
+            output.disable();
+        }
+        
         STATE state;
+        Set<STATE> states = new HashSet<>();
```
jwomeara: Where is this used?  Looks like it isn't.

jwomeara (changes requested):
Overall, looks good.  Some comments for the Output class and methods, plus some test cases would be appreciated.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExecutableDeterminationVisitor.java
ivakegg: will remove

jwomeara (dismissed):

apmoriarty (approved):
LGTM

FineAndDandy (approved):

PR #1654, opened 2022-07-26
Porting over unit tests for SetMembershipVisitor

avgAGB:


warehouse/query-core/src/test/java/datawave/query/jexl/visitors/SetMembershipVisitorTest.java
```diff
@@ -18,17 +25,152 @@
...
         config.setLazySetMechanismEnabled(false);
         query = null;
         fields.clear();
     }
     
+    @Test
+    public void simpleTestMembers() throws Exception {
+        Assert.assertEquals(Sets.newHashSet("INDEX_ONLY"),
+                        SetMembershipVisitor.getMembers(indexOnly, config, JexlASTHelper.parseJexlQuery("INDEX_ONLY == 'foobar'")));
+    }
```
apmoriarty: What do you think about having the tests delegate to a test method, something like this?
```java
public void simpleTestMembers(){
    test("INDEX_ONLY == 'foobar'", Sets.newHashSet("INDEX_ONLY"));
}

private void test(String query, Set<String> expectedMembers){
    // query parsing and asserts
}
```

avgAGB: Sure, I can make that adjustment in the next commit

apmoriarty (approved):
Great work on the unit test!

ivakegg (approved):

PR #1655, opened 2022-07-28
Create new Ingest Job to handle files from new services

adro5:
Work in progress


warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
```diff
@@ -247,6 +247,7 @@ protected void printUsage() {
     
     @Override
     public int run(String[] args) throws Exception {
+        inputFormat = EventSequenceFileInputFormat.class;
```
ivakegg, line 250: This is already initialized to the EventSequenceFileInputFormat in the construction of the class.  Is this needed?

adro5:
Closing this PR as this work is no longer needed.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
adro5: This was just a placeholder.

PR #1668, opened 2022-08-11
Query ID Input Checks

avgAGB:


web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -1938,6 +1938,11 @@ private BaseQueryResponse next(final String id, boolean checkForContentLookup) {
         RunningQuery query = null;
         Query contentLookupSettings = null;
         try {
+            if (!id.matches("[a-z\\d-]+")) {
```
apmoriarty: Extract common pattern to a class level variable. Additionally, do we need 'a-z' or just 'a-f' since we're expecting uuids?

avgAGB: a-f seems fine to me, I'll make that adjustment and pull it out

web-services/query/src/test/java/datawave/webservice/query/runner/ExtendedQueryExecutorBeanTest.java
```diff
@@ -864,6 +865,161 @@ public void testCreateQueryAndNext_HappyPath() throws Exception {
...
+        subject.createQuery(queryLogicName, queryParameters);
+        
+        Throwable result1 = null;
+        try {
+            subject.next("{id}");
+        } catch (Exception e) {
+            result1 = e.getCause();
+        }
+        
+        assertNotNull("Expected a non-null response", result1);
```
matthpeterson, line 1025: Maybe it's worth verifying that the exception is of type BadRequestQueryException with a getErrorCode equal to DatawaveErrorCode.INVALID_QUERY_ID

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -1938,6 +1938,11 @@ private BaseQueryResponse next(final String id, boolean checkForContentLookup) {
...
+                log.error("Invalid query id: " + id);
```
matthpeterson, line 1943: I don't know if the ID endpoint is currently case sensitive.  If it isn't, this change will make it so, which may or may not be the intention.

avgAGB: Good point, I'll add upper cases in as well just in case.

web-services/query/src/test/java/datawave/webservice/query/runner/ExtendedQueryExecutorBeanTest.java
avgAGB: Okay, will adjust this on next commit

apmoriarty (approved):

matthpeterson (approved):

PR #1672, opened 2022-08-15
1671: add query.syntax param to edgequery

wgtong:


warehouse/query-core/src/main/java/datawave/query/tables/edge/EdgeQueryLogic.java
```diff
@@ -148,6 +162,78 @@ public GenericQueryConfiguration initialize(Connector connection, Query settings
...
+        if (null == originalQuery) {
+            throw new IllegalArgumentException("Query cannot be null");
+        } else {
+            if ("JEXL".equals(querySyntax) || "LIST".equals(querySyntax) || "LIMITED_JEXL".equals(querySyntax)) {
+                queryString = originalQuery;
+            } else {
+                QueryNode node = querySyntaxParser.parse(originalQuery);
+                queryString = node.getOriginalQuery();
+                if (log.isTraceEnabled()) {
+                    log.trace("luceneQueryString: " + originalQuery + " --> jexlQueryString: " + queryString);
```
wgtong: Shouldn't necessarily assume lucene

warehouse/query-core/src/main/java/datawave/query/tables/edge/EdgeQueryLogic.java
```diff
@@ -148,6 +162,78 @@ public GenericQueryConfiguration initialize(Connector connection, Query settings
...
+                if (null == querySyntaxParser) {
+                    throw new IllegalArgumentException("QueryParser not configured for syntax: " + querySyntax);
+                }
+            }
+        }
+        
+        if (null == originalQuery) {
+            throw new IllegalArgumentException("Query cannot be null");
+        } else {
+            if ("JEXL".equals(querySyntax) || "LIST".equals(querySyntax) || "LIMITED_JEXL".equals(querySyntax)) {
```
wgtong: need a better way to handle null allowed syntax

warehouse/query-core/src/main/java/datawave/query/tables/edge/EdgeQueryLogic.java
```diff
@@ -148,6 +162,78 @@ public GenericQueryConfiguration initialize(Connector connection, Query settings
         return cfg;
     }
     
+    public String getJexlQueryString(Query settings) throws datawave.query.language.parser.ParseException {
+        // queryString should be JEXl after all query parsers are applied
+        String queryString;
+        String originalQuery = settings.getQuery();
+        
+        originalQuery = this.expandQueryMacros(originalQuery);
```
wgtong, line 169: add this property to edgequeries?

wgtong: seems like this logic takes a string from the query like foo(123,1234) and replaces foo with some transformation of the selector.
regardless I dont see why we should exclude this functionality from edgequery aside from this falling out of scope of the ticket

warehouse/query-core/src/main/java/datawave/query/config/EdgeExtendedSummaryConfiguration.java
```diff
@@ -52,7 +52,7 @@ public EdgeQueryConfiguration parseParameters(Query settings) {
                 String paramVal = p.getParameterValue();
                 if (paramVal.equals(LIST)) {
                     summaryInputType = true;
-                } else if (paramVal.equals(LIMITED_JEXL)) {
+                } else { // Lucene or JEXEL or LIMITED_JEXL
```
wgtong: probably don't need this comment

warehouse/query-core/src/main/java/datawave/query/tables/edge/EdgeQueryLogic.java
```diff
@@ -148,6 +161,83 @@ public GenericQueryConfiguration initialize(Connector connection, Query settings
...
+        QueryParser querySyntaxParser = getParser();
+        
+        if (org.apache.commons.lang.StringUtils.isBlank(querySyntax)) {
+            // Default to the class's query parser when one is not provided
+            // Falling back to Jexl when one is not set on this class
+            if (null == querySyntaxParser) {
+                querySyntax = "JEXL";
+            }
+        }
+        if (querySyntax.equals("JEXL")) {
```
wgtong, line 199: it is important to note that if we are NOT using JEXL then a parser is required to be defined in the LogicFactory.xml (can be null)
the behavior before accepted LIMITED_JEXL/LIST parms without the parser because it didn't actually have any parsing functionallity

warehouse/query-core/src/main/java/datawave/query/tables/edge/EdgeQueryLogic.java
wgtong: fyi i changed the logic in `getJexlQueryString` from shardquerylogic instead of copying it because i think it's more readable this way. One could argue that we should pop the logic out of shardquerylogic and share it between the two. they might be right. but i am not sure.

warehouse/query-core/src/main/java/datawave/query/tables/edge/EdgeQueryLogic.java
avgAGB: if LUCENE is passed as a syntax how would you want to deal with that?

warehouse/query-core/src/test/java/datawave/query/edge/ExtendedEdgeQueryLogicTest.java
```diff
@@ -69,6 +69,22 @@ public void testEdgeQuerySyntax_WithQueryModel() throws Exception {
         compareResults(logic, expected);
     }
     
+    @Test
+    public void testEdgeQuerySyntaxLuceneWithQueryModel() throws Exception {
```
avgAGB, line 74: I like this test. I'd be interested in seeing a JEXL syntax query added in here as well for parity, just to see how the logic handles both syntaxes if that seems reasonable.

warehouse/query-core/src/main/java/datawave/query/tables/edge/EdgeQueryLogic.java
wgtong: If LUCENE is passed then the setting must include the parser or it will error out

warehouse/query-core/src/test/java/datawave/query/edge/ExtendedEdgeQueryLogicTest.java
wgtong: since jexl is the default param, other test cases are essentially testing jexl syntax. I feel like adding another test to test that would be overkill. instead I think adding the param to one of the existing one should cover that.

avgAGB: Sounds good to me, thanks

hlgp (dismissed):

warehouse/query-core/src/main/java/datawave/query/tables/edge/EdgeQueryLogic.java
hlgp: Worth opening some tickets for these.  I'd say they're out of the scope of this ticket.

hlgp (approved):

apmoriarty (approved):

PR #1674, opened 2022-08-16
fixes #1673: Updates the whindex visitor to use the whindex creation …

### LINKED ISSUE #1673: Add Whindex Creation Date metadata
Instead of using the F column to determine whether a whindex field existed on a particular day, we should instead manually record the date that the whindex field was created and use that to determine whether or not to use that particular whindex.

jwomeara:
…date to determine whether a whindex should be used.

This depends on the following change in the metadata-utils repo:
https://github.com/NationalSecurityAgency/datawave-metadata-utils/pull/12


warehouse/query-core/src/test/java/datawave/query/jexl/visitors/WhindexVisitorTest.java
```diff
@@ -74,6 +71,10 @@ public static void beforeClass() {
         
         allMappings.putAll(multipleFieldMapping);
         allMappings.putAll(manyToOneFieldMapping);
+        
+        creationDateMap.put("HOT_FUDGE_SUNDAE", new Date(0));
+        creationDateMap.put("BANANA_SPLIT", new Date(0));
+        creationDateMap.put("NUT_SUNDAE", new Date(0));
```
FineAndDandy, line 77: can you add one test case which verifies that the dates are applied correctly? -1/0/+1

FineAndDandy (dismissed):

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/WhindexVisitorTest.java
jwomeara: I think I already covered this.  Check line 89 where I set the date for HOT_FUDGE_SUNDAE to something other than new Date(0)

ivakegg (approved):

keith-ratcliffe (approved):
LGTM

PR #1679, opened 2022-08-30
Fixes #1678 fix LimitFields group commonality bug

### LINKED ISSUE #1678: LimitFields hit term commonality check bug
https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java#L183 compares groupings with .endsWith instead of .equals causing us to return extra fields.  We will return group 8 if 18 is in the HIT_TERMS, 2 if 12, 3 if 33, etc.

hlgp:


warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
```diff
@@ -180,7 +180,8 @@ private void manageHitsAndMisses(String keyWithGrouping, String keyNoGrouping, A
                         String[] commonalityAndGroupingContext = LimitFields.getCommonalityAndGroupingContext(key);
                         if (commonalityAndGroupingContext != null) {
                             String hitTermKeyCommonality = commonalityAndGroupingContext[0];
-                            if (hitTermKeyCommonality.equals(keyWithGroupingCommonality) && key.endsWith(keyWithGroupingSuffix)) {
+                            String hitTermGroup = commonalityAndGroupingContext[1];
+                            if (hitTermKeyCommonality.equals(keyWithGroupingCommonality) && keyWithGroupingSuffix.equals(hitTermGroup)) {
```
apmoriarty, line 184: Can we add an issue to refactor and clean up some of the extra object creation here?

apmoriarty (approved):

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
hlgp: feel free to open one

apmoriarty: Opened https://github.com/NationalSecurityAgency/datawave/issues/1680

ivakegg (approved):
Nice job!

Issue #1658, opened 2022-08-01
Remote query service query logic and query filters


ivakegg:
We need a query logic (BaseQueryLogic extension) that can be configured with a remote URL through which its activities are delegated using the standard create, next, and close rest calls.

ivakegg:
In addition to the remote query service logic, we need to be able to create composite logics that would invoke a local and potentially remote query logic.  This already exists in the form of the CompositeQueryLogic.  However in addition we need to be able to restrict calling one of the query logics based on some filter.  Current filters require include one that matches the auths be requested, and another that matches a query parameter.

PR #1681, opened 2022-08-30
re #1658: Added a remote query service

ivakegg:
Added a remote query service to allow the webserver to call through to the query microservices
Added a remote query logic that can invoke a remote query service using an injected RemoteQueryService.

### DIFF
--- a/contrib/datawave-quickstart/docker/datawave-bootstrap.sh
+++ b/contrib/datawave-quickstart/docker/datawave-bootstrap.sh
@@ -16,6 +16,13 @@ do
           --web)
              START_WEB=true
              ;;
+          --webdebug)
+             START_WEB_DEBUG=true
+             ;;
+          --test)
+             START_TEST=true
+             START_AS_DAEMON=false
+             ;;
           --ingest)
              START_INGEST=true
              ;;
@@ -28,6 +35,30 @@ done
 [ "${START_INGEST}" == true ] && datawaveIngestStart
 
 [ "${START_WEB}" == true ] && datawaveWebStart
+
+[ "${START_WEB_DEBUG}" == true ] && datawaveWebStart --debug
+
+if [ "${START_TEST}" == true ] ; then
+    datawaveWebStart
+    status=$?
+
+    if [ "$status" != "0" ] ; then
+        echo "datawaveWebStart Failed"
+        cat ${WILDFLY_HOME}/standalone/log/server.log
+        exit $status
+    else
+        datawaveWebTest --blacklist-files QueryMetrics
+        status=$?
+
+        if [ "$status" != "0" ] ; then
+            echo "datawaveWebTest Failed"
+            cat ${WILDFLY_HOME}/standalone/log/server.log
+        fi
+
+        allStop
+        exit $status
+    fi
+fi
 
 if [ "${START_AS_DAEMON}" == true ] ; then
     while true; do sleep 1000; done
--- a/pom.xml
+++ b/pom.xml
@@ -69,8 +69,8 @@
         <version.jcommander>1.72</version.jcommander>
         <version.jetty>6.1.26</version.jetty>
         <version.jgroups>4.0.19.Final</version.jgroups>
-        <version.jjwt>0.9.1</version.jjwt>
-        <version.junit>4.13.1</version.junit>
+        <version.jjwt>0.11.2</version.jjwt>
+        <version.junit>4.13.2</version.junit>
         <version.kryo>2.20</version.kryo>
         <version.kryonet>2.20</version.kryonet>
         <version.log4j-extras>1.0</version.log4j-extras>
@@ -79,7 +79,7 @@
         <version.microservice.accumulo-api>1.1</version.microservice.accumulo-api>
         <version.microservice.accumulo-utils>1.4</version.microservice.accumulo-utils>
         <version.microservice.audit-api>1.4</version.microservice.audit-api>
-        <version.microservice.authorization-api>1.4</version.microservice.authorization-api>
+        <version.microservice.authorization-api>1.14</version.microservice.authorization-api>
         <version.microservice.base-rest-responses>1.2</version.microservice.base-rest-responses>
         <version.microservice.common-utils>1.0</version.microservice.common-utils>
         <version.microservice.dictionary-api>1.0</version.microservice.dictionary-api>
@@ -377,7 +377,12 @@
             </dependency>
             <dependency>
                 <groupId>io.jsonwebtoken</groupId>
-                <artifactId>jjwt</artifactId>
+                <artifactId>jjwt-impl</artifactId>
+                <version>${version.jjwt}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.jsonwebtoken</groupId>
+                <artifactId>jjwt-jackson</artifactId>
                 <version>${version.jjwt}</version>
             </dependency>
             <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -79,7 +79,7 @@
         <version.microservice.accumulo-api>1.1</version.microservice.accumulo-api>
         <version.microservice.accumulo-utils>1.4</version.microservice.accumulo-utils>
         <version.microservice.audit-api>1.4</version.microservice.audit-api>
-        <version.microservice.authorization-api>1.14</version.microservice.authorization-api>
+        <version.microservice.authorization-api>1.13</version.microservice.authorization-api>
         <version.microservice.base-rest-responses>1.2</version.microservice.base-rest-responses>
         <version.microservice.common-utils>1.0</version.microservice.common-utils>
         <version.microservice.dictionary-api>1.0</version.microservice.dictionary-api>
--- a/warehouse/query-core/src/main/java/datawave/query/config/RemoteQueryConfiguration.java
+++ b/warehouse/query-core/src/main/java/datawave/query/config/RemoteQueryConfiguration.java
@@ -1,0 +1,153 @@
+package datawave.query.config;
+
+import datawave.query.tables.RemoteEventQueryLogic;
+import datawave.webservice.query.Query;
+import datawave.webservice.query.configuration.GenericQueryConfiguration;
+
+import java.io.ObjectStreamException;
+import java.io.Serializable;
+import java.util.Objects;
+
+/**
+ * <p>
+ * A GenericQueryConfiguration implementation that provides the additional logic on top of the traditional query that is needed to run a remote query logikc
+ *
+ */
+public class RemoteQueryConfiguration extends GenericQueryConfiguration implements Serializable {
+    
+    private static final long serialVersionUID = -4354990715046146110L;
+    
+    // the id of the remote query
+    private String remoteId;
+    
+    private String remoteQueryLogic;
+    
+    private Query query;
+    
+    /**
+     * Default constructor
+     */
+    public RemoteQueryConfiguration() {
+        super();
+    }
+    
+    /**
+     * Performs a deep copy of the provided RemoteQueryConfiguration into a new instance
+     *
+     * @param other
+     *            - another RemoteQueryConfiguration instance
+     */
+    public RemoteQueryConfiguration(RemoteQueryConfiguration other) {
+        
+        // GenericQueryConfiguration copy first
+        super(other);
+        
+        // RemoteQueryConfiguration copy
+        this.remoteId = other.getRemoteId();
+        this.remoteQueryLogic = other.getRemoteQueryLogic();
+        this.query = other.getQuery();
+    }
+    
+    /**
+     * Delegates deep copy work to appropriate constructor, sets additional values specific to the provided RemoteRemoteQueryLogic
+     *
+     * @param logic
+     *            - a RemoteQueryLogic instance or subclass
+     */
+    public RemoteQueryConfiguration(RemoteEventQueryLogic logic) {
+        this(logic.getConfig());
+    }
+    
+    /**
+     * Factory method that instantiates an fresh RemoteQueryConfiguration
+     *
+     * @return - a clean RemoteQueryConfiguration
+     */
+    public static RemoteQueryConfiguration create() {
+        return new RemoteQueryConfiguration();
+    }
+    
+    /**
+     * Factory method that returns a deep copy of the provided RemoteQueryConfiguration
+     *
+     * @param other
+     *            - another instance of a RemoteQueryConfiguration
+     * @return - copy of provided RemoteQueryConfiguration
+     */
+    public static RemoteQueryConfiguration create(RemoteQueryConfiguration other) {
+        return new RemoteQueryConfiguration(other);
+    }
+    
+    /**
+     * Factory method that creates a RemoteQueryConfiguration deep copy from a RemoteQueryLogic
+     *
+     * @param remoteQueryLogic
+     *            - a configured RemoteQueryLogic
+     * @return - a RemoteQueryConfiguration
+     */
+    public static RemoteQueryConfiguration create(RemoteEventQueryLogic remoteQueryLogic) {
+        return create(remoteQueryLogic.getConfig());
+    }
+    
+    /**
+     * Factory method that creates a RemoteQueryConfiguration from a RemoteQueryLogic and a Query
+     *
+     * @param remoteQueryLogic
+     *            - a configured RemoteQueryLogic
+     * @param query
+     *            - a configured Query object
+     * @return - a RemoteQueryConfiguration
+     */
+    public static RemoteQueryConfiguration create(RemoteEventQueryLogic remoteQueryLogic, Query query) {
+        RemoteQueryConfiguration config = create(remoteQueryLogic);
+        config.setQuery(query);
+        return config;
+    }
+    
+    public String getRemoteId() {
+        return remoteId;
+    }
+    
+    public void setRemoteId(String remoteId) {
+        this.remoteId = remoteId;
+    }
+    
+    public String getRemoteQueryLogic() {
+        return remoteQueryLogic;
+    }
+    
+    public void setRemoteQueryLogic(String remoteQueryLogic) {
+        this.remoteQueryLogic = remoteQueryLogic;
+    }
+    
+    public Query getQuery() {
+        return query;
+    }
+    
+    public void setQuery(Query query) {
+        this.query = query;
+    }
+    
+    @Override
+    public boolean equals(Object o) {
+        if (this == o)
+            return true;
+        if (o == null || getClass() != o.getClass())
+            return false;
+        if (!super.equals(o))
+            return false;
+        RemoteQueryConfiguration that = (RemoteQueryConfiguration) o;
+        return Objects.equals(getRemoteId(), that.getRemoteId()) && Objects.equals(getRemoteQueryLogic(), that.getRemoteQueryLogic())
+                        && Objects.equals(getQuery(), that.getQuery());
+    }
+    
+    @Override
+    public int hashCode() {
+        return Objects.hash(super.hashCode(), getRemoteId(), getRemoteQueryLogic(), getQuery());
+    }
+    
+    // Part of the Serializable interface used to initialize any transient members during deserialization
+    protected Object readResolve() throws ObjectStreamException {
+        return this;
+    }
+}
--- a/warehouse/query-core/src/main/java/datawave/query/config/RemoteQueryConfiguration.java
+++ b/warehouse/query-core/src/main/java/datawave/query/config/RemoteQueryConfiguration.java
@@ -10,7 +10,7 @@ import java.util.Objects;
 
 /**
  * <p>
- * A GenericQueryConfiguration implementation that provides the additional logic on top of the traditional query that is needed to run a remote query logikc
+ * A GenericQueryConfiguration implementation that provides the additional logic on top of the traditional query that is needed to run a remote query logic
  *
  */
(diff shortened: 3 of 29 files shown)


contrib/datawave-quickstart/docker/datawave-bootstrap.sh
```diff
@@ -16,6 +16,13 @@ do
           --web)
              START_WEB=true
              ;;
+          --webdebug)
```
ivakegg, line 19: Added some flags to allow debugging the web server or running the tests therein

pom.xml
```diff
@@ -69,8 +69,8 @@
         <version.jcommander>1.72</version.jcommander>
         <version.jetty>6.1.26</version.jetty>
         <version.jgroups>4.0.19.Final</version.jgroups>
-        <version.jjwt>0.9.1</version.jjwt>
-        <version.junit>4.13.1</version.junit>
+        <version.jjwt>0.11.2</version.jjwt>
```
ivakegg, line 72: as it turns out the authorization service upped the jwt version requiring me to update them here to allow using it in remote query service.

pom.xml
```diff
@@ -79,7 +79,7 @@
         <version.microservice.accumulo-api>1.1</version.microservice.accumulo-api>
         <version.microservice.accumulo-utils>1.4</version.microservice.accumulo-utils>
         <version.microservice.audit-api>1.4</version.microservice.audit-api>
-        <version.microservice.authorization-api>1.4</version.microservice.authorization-api>
+        <version.microservice.authorization-api>1.14</version.microservice.authorization-api>
```
ivakegg: This dependency was way behind.

web-services/client/src/main/java/datawave/webservice/query/QueryImpl.java
```diff
@@ -817,6 +817,8 @@ public Map<String,List<String>> toMap() {
                 throw new RuntimeException("Error formatting date", e);
             }
         }
+        p.set("pageTimeout", Integer.toString(this.pageTimeout));
```
ivakegg: Found a parameter in the map

web-services/common/src/main/java/datawave/webservice/common/json/DefaultMapperDecorator.java
```diff
@@ -21,6 +22,7 @@ public ObjectMapper decorate(ObjectMapper mapper) {
         mapper.enable(MapperFeature.USE_WRAPPER_NAME_AS_PROPERTY_NAME);
         mapper.registerModule(new GuavaModule());
         mapper.registerModule(new JaxbAnnotationModule());
+        mapper.disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
```
ivakegg, line 25: used to enable invoking the new query microservice remotely

web-services/common/src/main/java/datawave/webservice/common/remote/RemoteHttpService.java
```diff
@@ -86,12 +86,13 @@ public abstract class RemoteHttpService {
...
     
     protected <T> T execute(HttpRequestBase request, IOFunction<T> resultConverter, Supplier<String> errorSupplier) throws IOException {
+        log.info("Executing " + request.getClass().getSimpleName() + " against " + request.getURI());
         try {
             activeExecutions.incrementAndGet();
             return client.execute(
                             request,
                             r -> {
-                                if (r.getStatusLine().getStatusCode() != 200) {
+                                if (r.getStatusLine().getStatusCode() >= 300) {
```
ivakegg, line 95: only fail if we actually have a failure code.  a 204 should not be an error for example.

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogicTransformer.java
```diff
@@ -75,33 +63,41 @@ public List<Object> readFromCache(List<CacheableQueryRow> row) {
     
     @Override
     public BaseQueryResponse createResponse(ResultsPage resultList) {
-        BaseQueryResponse result = null;
+        Exception lastFailure = null;
```
ivakegg, line 66: It appears that we were calling all of the delegates which did not make sense.  Only the first one is required.  Same for the method below.

web-services/security/src/main/java/datawave/security/system/ServerSecurityProducer.java
```diff
@@ -31,17 +27,9 @@
 @ApplicationScoped
 @Exclude(ifProjectStage = DatawaveEmbeddedProjectStageHolder.DatawaveEmbedded.class)
 public class ServerSecurityProducer {
-    // Allow injection of JSSESecurityDomain without having to specify the JNDI name at each injection point.
```
ivakegg, line 34: I had to split some of the beans/resource into a separate class.  Otherwise there ends up being issues creating the JSSESecurityDomain for a remote query logic when loading the QLF because the DatawaveUserService bean (included in this same class) cannot be created at that time.

warehouse/query-core/src/main/java/datawave/query/config/RemoteQueryConfiguration.java
```diff
@@ -0,0 +1,153 @@
...
+import datawave.webservice.query.Query;
+import datawave.webservice.query.configuration.GenericQueryConfiguration;
+
+import java.io.ObjectStreamException;
+import java.io.Serializable;
+import java.util.Objects;
+
+/**
+ * <p>
+ * A GenericQueryConfiguration implementation that provides the additional logic on top of the traditional query that is needed to run a remote query logikc
```
drewfarris:
```suggestion
 * A GenericQueryConfiguration implementation that provides the additional logic on top of the traditional query that is needed to run a remote query logic
```

warehouse/query-core/src/main/java/datawave/query/config/RemoteQueryConfiguration.java
```diff
@@ -0,0 +1,153 @@
...
+    
+    @Override
+    public int hashCode() {
+        return Objects.hash(super.hashCode(), getRemoteId(), getRemoteQueryLogic(), getQuery());
+    }
+    
+    // Part of the Serializable interface used to initialize any transient members during deserialization
+    protected Object readResolve() throws ObjectStreamException {
+        return this;
+    }
```
drewfarris: Is this strictly necessary since there are no transient members _or_ should this have the `@Override` annotation because you are overriding behavior from a superclass?

warehouse/query-core/src/main/java/datawave/query/tables/RemoteEventQueryLogic.java
```diff
@@ -0,0 +1,276 @@
...
+            throw new QueryException("Did not receive a RemoteQueryConfiguration instance!!");
+        }
+        
+        config = (RemoteQueryConfiguration) genericConfig;
+        
+        // Create an iterator that returns a stream of EventBase objects
+        this.iterator = new RemoteQueryLogicIterator();
+    }
+    
+    private QueryLogicTransformer transformerInstance = null;
```
drewfarris: Consider moving this to the top of the class? Generally I try to keep member variables all in one place, but opinions differ on this.

warehouse/query-core/src/main/java/datawave/query/tables/RemoteEventQueryLogic.java
```diff
@@ -0,0 +1,276 @@
...
+            addConfigBasedTransformers();
+            return this.transformerInstance;
+        }
+        
+        MarkingFunctions markingFunctions = this.getMarkingFunctions();
+        EventBaseTransformer transformer = new EventBaseTransformer(settings, markingFunctions, getResponseObjectFactory());
+        
+        this.transformerInstance = transformer;
+        addConfigBasedTransformers();
+        return this.transformerInstance;
```
drewfarris: could theoretically drop `this` throughout this method, seems potentially unnecessary and inconsistent with `this` usage thorughout the rest of the class.

warehouse/query-core/src/main/java/datawave/query/tables/RemoteEventQueryLogic.java
```diff
@@ -0,0 +1,276 @@
...
+    }
+    
+    /**
+     * If the configuration didn't exist, OR IT CHANGED, we need to create or update the transformers that have been added.
+     */
+    private void addConfigBasedTransformers() {
+        if (getConfig() != null) {
+            // @TODO update transformer config
```
drewfarris: Fix TODO or leave for now?

web-services/query/src/main/java/datawave/webservice/query/logic/filtered/FilteredQueryLogic.java
```diff
@@ -0,0 +1,383 @@
...
+    }
+    
+    @Override
+    public String getPlan(Connector connection, Query settings, Set<Authorizations> runtimeQueryAuthorizations, boolean expandFields, boolean expandValues)
+                    throws Exception {
+        if (!filtered && filter.canRunQuery(settings, runtimeQueryAuthorizations)) {
+            return delegate.getPlan(connection, settings, runtimeQueryAuthorizations, expandFields, expandValues);
+        } else {
+            filtered = true;
+            return "";
```
drewfarris, line 50: Wondering if an empty plan is the correct thing to do here. I guess it comes down to the question: Is there some way to record the fact that the query was not run because is was 'filtered' elsewhere in the metrics?

web-services/query/src/main/java/datawave/webservice/query/logic/filtered/FilteredQueryLogic.java
```diff
@@ -0,0 +1,383 @@
...
+        if (!filtered) {
+            return delegate.iterator();
+        } else {
+            return Collections.emptyIterator();
+        }
+    }
+    
+    public List<String> getSelectors(Query settings) {
+        return delegate.getSelectors(settings);
+    }
```
drewfarris: Consider introducing a class called `DelegatingQueryLogic` or something that simply delegates to another query logic and would include most of the code in `FilteredQueryLogic`. Then `FilteredQueryLogic` can extend that and implement only those methods required for implementing filtering. This would simplify the implementation of `FilteredQueryLogic` and provide a clearer separation of concerns (at the expense of a more complex class hierarchy) - so it's may not be a new win to do this.

web-services/query/src/main/java/datawave/webservice/query/logic/filtered/QueryLogicFilterByAuth.java
```diff
@@ -0,0 +1,101 @@
...
+
+import datawave.webservice.query.Query;
+import org.apache.accumulo.core.security.Authorizations;
+import org.apache.accumulo.core.security.ColumnVisibility;
+import org.apache.accumulo.core.security.VisibilityEvaluator;
+import org.apache.accumulo.core.security.VisibilityParseException;
+
+import java.util.Set;
+
+public class QueryLogicFilterByAuth implements FilteredQueryLogic.QueryLogicFilter {
```
drewfarris: Please consider adding some javadocs here that describe the intent of the class.

web-services/query/src/main/java/datawave/webservice/query/logic/filtered/QueryLogicFilterByAuth.java
```diff
@@ -0,0 +1,101 @@
...
+    @Override
+    public boolean canRunQuery(Query settings, Set<Authorizations> auths) {
+        boolean canRunQuery = matches(auths);
+        if (negated) {
+            canRunQuery = !canRunQuery;
+        }
+        return canRunQuery;
+    }
+    
+    private boolean matches(Set<Authorizations> auths) {
```
drewfarris: ooh, I could really use the ability to do this 'generically', outside of the context of a `QueryLogicFilter`, can we implement the auth/visibility matching logic in a utility class that could be reused?

My use case is - looking at events that are being returned to the end-user and determining whether they need to be encrypted based on the presence of certain elements in the ColumnVisibility.

web-services/query/src/main/java/datawave/webservice/query/logic/filtered/QueryLogicFilterByParameter.java
```diff
@@ -0,0 +1,82 @@
...
+public class QueryLogicFilterByParameter implements FilteredQueryLogic.QueryLogicFilter {
+    
+    // The parameter to match against
+    private String parameter;
+    
+    // The value to match against. If empty then boolean value of parameter is used.
+    private String value;
+    
+    // if negated than the negation of the match is returned
+    private boolean negated = false;
```
drewfarris: Consider whether any of these can be `final`?

web-services/query/src/main/java/datawave/webservice/query/remote/RemoteQueryServiceImpl.java
```diff
@@ -0,0 +1,370 @@
...
+    
+    private String queryServiceURI = "/query/v1/";
+    
+    private int maxConnections = 100;
+    
+    private int retryCount = 5;
+    
+    private int unavailableRetryCount = 15;
+    
+    private int unavailableRetryDelay = 2000;
```
drewfarris: Consider a separate configuration object to hold these fields?

web-services/client/src/main/java/datawave/webservice/query/QueryImpl.java
apmoriarty: Could replace this with a reference to `QueryParameters.QUERY_PAGETIMEOUT`

web-services/common/src/main/java/datawave/webservice/common/remote/RemoteQueryService.java
```diff
@@ -0,0 +1,21 @@
+package datawave.webservice.common.remote;
+
+import datawave.webservice.result.BaseQueryResponse;
+import datawave.webservice.result.GenericResponse;
+import datawave.webservice.result.VoidResponse;
+
+import java.util.List;
+import java.util.Map;
+
+public interface RemoteQueryService {
```
apmoriarty, line 13: Add class level docs

warehouse/query-core/src/main/java/datawave/query/config/RemoteQueryConfiguration.java
ivakegg: got it

warehouse/query-core/src/main/java/datawave/query/config/RemoteQueryConfiguration.java
ivakegg: you are correct, not needed.

warehouse/query-core/src/main/java/datawave/query/tables/RemoteEventQueryLogic.java
ivakegg: I usually do as well.... this was copied as such from another class.

warehouse/query-core/src/main/java/datawave/query/tables/RemoteEventQueryLogic.java
ivakegg: hmm... maybe

warehouse/query-core/src/main/java/datawave/query/tables/RemoteEventQueryLogic.java
ivakegg: ah... forgot this.  cleaning up.

web-services/client/src/main/java/datawave/webservice/query/QueryImpl.java
ivakegg: ok

web-services/common/src/main/java/datawave/webservice/common/remote/RemoteQueryService.java
ivakegg: got it

web-services/query/src/main/java/datawave/webservice/query/logic/filtered/FilteredQueryLogic.java
ivakegg: If this query logic was filtered out, then there was no plan generated.  Are you suggesting that I should return something like "<< filtered >>" here?  The lack of plan is a pretty good indication that nothing was run.

web-services/query/src/main/java/datawave/webservice/query/logic/filtered/FilteredQueryLogic.java
ivakegg: Actually, I like that idea.  Doing it.

web-services/query/src/main/java/datawave/webservice/query/logic/filtered/QueryLogicFilterByAuth.java
ivakegg: right, of course.  Sorry.

web-services/query/src/main/java/datawave/webservice/query/logic/filtered/QueryLogicFilterByAuth.java
ivakegg: Let me see what I can do....

web-services/query/src/main/java/datawave/webservice/query/logic/filtered/QueryLogicFilterByParameter.java
ivakegg: No they cannot.  I need to be able to set them via setters to allow easier construction via spring (e.g. QueryLogicFactory.xml)

web-services/query/src/main/java/datawave/webservice/query/remote/RemoteQueryServiceImpl.java
ivakegg: perhaps.  considering.

web-services/query/src/main/java/datawave/webservice/query/logic/filtered/QueryLogicFilterByAuth.java
ivakegg: Creating java.util.Predicate implementations that you can reuse.

web-services/query/src/main/java/datawave/webservice/query/remote/RemoteQueryServiceImpl.java
ivakegg: done

web-services/query/src/main/java/datawave/webservice/query/logic/filtered/FilteredQueryLogic.java
drewfarris: Good point. I was thinking it would be helpful to know that the query wasn't run because this specific query logic decided that it should not run, might be helpful for folks troubleshooting that aren't as familiar with the system. Nevertheless, it's not clear how valuable that might be in the long run.

web-services/query/src/main/java/datawave/webservice/query/predicate/AuthorizationsPredicate.java
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.accumulo.core.security.ColumnVisibility;
+import org.apache.accumulo.core.security.VisibilityEvaluator;
+import org.apache.accumulo.core.security.VisibilityParseException;
+
+import java.util.function.Predicate;
+
+/**
+ * This is a predicate that will test the auths against a specified visibility (as defined by accumulo's ColumnVisibility). In addition to the visibility, one
+ * can specify that only the first of the authorizations is matched (presumably the user).
+ */
```
drewfarris, line 13: Thanks @ivakegg! 🥇

web-services/query/src/main/java/datawave/webservice/query/predicate/ProxiedAuthorizationsPredicate.java
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.accumulo.core.security.Authorizations;
+import org.apache.accumulo.core.security.ColumnVisibility;
+
+import java.util.List;
+import java.util.function.Predicate;
+
+/**
+ * This is a predicate that will test the auths against a specified visibility (as defined by accumulo's ColumnVisibility). In addition to the visibility, one
+ * can specify that only the first of the authorizations is matched (presumably the user).
+ */
```
drewfarris, line 12: Thanks for this one too! 🐩

web-services/query/src/main/java/datawave/webservice/query/remote/RemoteQueryServiceImpl.java
```diff
@@ -0,0 +1,347 @@
...
+            config.getFailureCounter().inc();
+            throw new RuntimeException(e.getMessage(), e);
+        }
+    }
+    
+    @Override
+    protected String serviceHost() {
+        return config.getServiceHost();
+    }
+
```
drewfarris, line 250: Are all these delegate methods really needed? Can't we just get an reference to the configuration object and interact with that directly? (or inject a configuration object where necessary?)

ivakegg: I can only get rid of the setter methods because the other methods are required by the parent class.  Creating yet another bean to be configured via the QLF.xml seems more tedious.  I feel the setter methods here are not all that overbearing.  The Idea generated them for me!

apmoriarty (dismissed):

hlgp (dismissed):

drewfarris (dismissed):

PR #1683, opened 2022-08-31
Short circuit pattern node

### LINKED ISSUE #1505: Short Circuit Visitors - PatternNode

mineralntl:


avgAGB (dismissed):

alerman (dismissed):

apmoriarty (dismissed):

wgtong (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PatternNodeVisitor.java
```diff
@@ -34,4 +36,30 @@ public Object visit(ASTERNode node, Object data) {
         
         return patterns;
     }
+    
+    // Ensure we short circuit these nodes
```
ivakegg, line 40: Is this not exactly what the ShortCircuitBaseVisitor does for you?  Why have these methods here?

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PatternNodeVisitor.java
mineralntl: No, the ShortCircuitBaseVisitor does not include these specific functions. It originally contained them, but it was discovered that other visitors required specific logic in those methods so it was easier to pull them out and implement them for each extending visitor.

apmoriarty (dismissed):

apmoriarty (dismissed):

ivakegg (dismissed):

avgAGB (approved):

apmoriarty (approved):

PR #1695, opened 2022-09-15
Fixing cache trace cast in DateIndexHelper

avgAGB:


hlgp (approved):

warehouse/query-core/src/main/java/datawave/query/util/DateIndexHelper.java
```diff
@@ -22,6 +21,7 @@
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.stereotype.Component;
+import com.github.benmanes.caffeine.cache.Cache;
```
matthpeterson, line 24: This seems like a reasonable approach because of the current class name in use by https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/core/src/main/resources/CacheContext.xml#L20.

Another possible approach for this would be to wrap the block of code that does the casting and logging with a conditional that uses instanceof to check the class type.

matthpeterson (approved):

Issue #1696, opened 2022-09-15
Optionally Pull Accumulo Password from the Environment


apmoriarty:
Optionally pull the accumulo password from the environment, if configured.

PR #1697, opened 2022-09-15
QueryLogics can optionally pull accumulo passwords from the environment

apmoriarty:
Addresses https://github.com/NationalSecurityAgency/datawave/issues/1696


web-services/query/src/main/java/datawave/webservice/query/configuration/GenericQueryConfiguration.java
```diff
@@ -167,6 +170,48 @@ public void setBypassAccumulo(boolean bypassAccumulo) {
...
+     *
+     * @param accumuloPasswordEnv
+     *            the environment variable where the accumulo password is stored
+     */
+    public void setAccumuloPasswordEnv(String accumuloPasswordEnv) {
+        this.accumuloPasswordEnv = accumuloPasswordEnv;
+        if (StringUtils.isNotBlank(accumuloPasswordEnv)) {
+            String password = System.getenv(accumuloPasswordEnv);
+            if (StringUtils.isNotBlank(password)) {
+                this.accumuloPassword = password;
```
avgAGB: splitting hairs, but would it make sense to use the setter for the accumulo password rather then directly modding it here?

apmoriarty: works for me. I'll add it in the next commit

avgAGB (dismissed):

hlgp (dismissed):

warehouse/query-core/src/main/java/datawave/query/metrics/AccumuloRecordWriter.java
```diff
@@ -298,7 +299,8 @@ protected static String getUsername(Configuration conf) {
      * string, and is not intended to be secure.
      */
     protected static byte[] getPassword(Configuration conf) {
-        return Base64.decodeBase64(conf.get(PASSWORD, "").getBytes());
+        String pw = EnvProvider.resolve(PASSWORD);
```
FineAndDandy: will this be base64 encoded?

FineAndDandy (changes requested):

warehouse/query-core/src/main/java/datawave/query/metrics/AccumuloRecordWriter.java
apmoriarty: This whole code block was wrong, good catch.

ivakegg (approved):

FineAndDandy (approved):

PR #1700, opened 2022-09-15
fixes #1693: Added the ability to restrict the query logics

### LINKED ISSUE #1693: The QueryLogicFactory needs to be able to restrict query logics
Currently the QueryLogicFactory will allow all query logic beans within the spring configuration to be used.  However depending on the deployed system, we may want to reduce the set of allowed query logic beans.  The suggested mechanism is to configure the query logic factory with a set of "allowedQueryTypes".  These types would be used to control query logics in that the query logic names would be prefaced with the type.
So for example of the allowed query types are "TEST,DEFAULT", then when requesting a query logic by the name of "EventQuery", then QueryLogicFactory would first look for a bean named "TEST-EventQuery".  If that is found then it would be returned.  If not, then just "EventQuery" will be searched for.  If that is found then it would be returned.
If DEFAULT is not in the list of allowed query types, then the defaults could never be returned.

ivakegg:
* Added a query logics map to the query logic factory config.
* That map, if populated, will restrict the query logics
*  that can be loaded, and will allow using a different
*  name than the bean name.

### DIFF
--- a/web-services/query/src/main/java/datawave/webservice/query/logic/QueryLogicFactoryImpl.java
+++ b/web-services/query/src/main/java/datawave/webservice/query/logic/QueryLogicFactoryImpl.java
@@ -2,6 +2,8 @@ package datawave.webservice.query.logic;
 
 import java.security.Principal;
 import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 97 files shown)


ivakegg:
I also updated the composite query logic to allow configuring a list of QueryLogic instead of BaseQueryLogic instances.  This will allow using RemoteEventQueryLogic or FilteredQueryLogic instances in a CompositeQueryLogic.

FineAndDandy (approved):

drewfarris (approved):

web-services/query/src/main/java/datawave/webservice/query/logic/QueryLogicFactoryImpl.java
```diff
@@ -54,6 +68,15 @@ public QueryLogic<?> getQueryLogic(String name, Principal principal) throws Ille
     @Override
     public List<QueryLogic<?>> getQueryLogicList() {
         Map<String,QueryLogic> logicMap = applicationContext.getBeansOfType(QueryLogic.class);
+        if (queryLogicFactoryConfiguration.hasLogicMap()) {
+            Map<String,QueryLogic> renamedLogicMap = new HashMap<>();
+            for (Map.Entry<String,String> entry : queryLogicFactoryConfiguration.getLogicMap().entrySet()) {
+                if (logicMap.containsKey(entry.getValue())) {
+                    renamedLogicMap.put(entry.getKey(), logicMap.get(entry.getValue()));
+                }
```
drewfarris, line 76: Consider logging a warning or throwing some sort of fatal exception if the `logicMap` doesn't contain the expected entry?

ivakegg: Failure to get the logic will result in appropriate errors that get logged and passed back to the user outside of this class.

PR #1706, opened 2022-09-26
fixes #1705: LimitFields performance

### LINKED ISSUE #1705: LimitFields will create unnecessarily large lists of attributes
The LimitFields algorithm is designed to limit the number of values returned for a field.  If the number of matched values exceeds that limit, then all of the hits values are returned.  If the number matches is less than the limit requested, then the values are padded up to the limit with values that did not match in the query.  The problem is that the current approach creates a full list of hit values and a full list of unmatched values before creating the final list of attributes to return.  When a document has an extreme number of values for one field, this is a large waste and can start causing GC issues.

ivakegg:
* Updated to avoid collecting all misses for any limited field
* Updated to use the toKeep mechanism instead of a collection of removals

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
+++ b/warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
@@ -125,8 +125,9 @@ public class LimitFields implements Function<Entry<Key,Document>,Entry<Key,Docum
                                 value.setToKeep(true);
                                 keepers++;
                                 missesRemaining--;
-                                missesToSet--;
                                 attributesToDrop--;
+                                foundMiss = true;
+                                missesToSet--;
                                 if (missesToSet == 0) {
                                     break;
                                 }
@@ -138,7 +139,6 @@ public class LimitFields implements Function<Entry<Key,Document>,Entry<Key,Docum
                             attr.setToKeep(true);
                             keepers++;
                             missesRemaining--;
-                            missesToSet--;
                             attributesToDrop--;
                             foundMiss = true;
                         }
@@ -163,6 +163,16 @@ public class LimitFields implements Function<Entry<Key,Document>,Entry<Key,Docum
                 int originalCount = countForFieldMap.get(keyNoGrouping);
                 if (originalCount > keepers) {
                     document.put(keyNoGrouping + ORIGINAL_COUNT_SUFFIX, new Numeric(originalCount, document.getMetadata(), document.isToKeep()), true, false);
+                    
+                    // some sanity checks
+                    int missesRemaining = countMissesRemainingForFieldMap.get(keyNoGrouping);
+                    int limit = this.limitFieldsMap.get(keyNoGrouping);
+                    int missesToSet = Math.min(limit - keepers, missesRemaining);
+                    if (missesToSet > 0) {
+                        log.error("Failed to limit fields correctly, " + missesToSet + " attributes failed to be included");
+                        throw new RuntimeException("Failed to limit fields correctly, " + missesToSet + ' ' + keyNoGrouping
+                                        + " attributes failed to be included");
+                    }
                 }
             }
         }
--- a/warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
+++ b/warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
@@ -2,7 +2,6 @@ package datawave.query.function;
 
 import com.google.common.base.Function;
 import com.google.common.collect.HashMultimap;
-import com.google.common.collect.Maps;
 import com.google.common.collect.Multimap;
 import datawave.data.type.Type;
 import datawave.query.attributes.Attribute;
@@ -35,15 +34,14 @@ public class LimitFields implements Function<Entry<Key,Document>,Entry<Key,Docum
     
     @Override
     public Entry<Key,Document> apply(Entry<Key,Document> entry) {
-        // key is the limited field name with _ORIGINAL_COUNT appended,
-        // value will be set to the original count of that field in the document
-        Map<String,Integer> limitedFieldCounts = new HashMap<>();
         Document document = entry.getValue();
         Multimap<String,String> hitTermMap = this.getHitTermMap(document);
         
-        Map<String,Integer> countForFieldMap = Maps.newHashMap();
-        Map<String,Integer> countMissesRemainingForFieldMap = Maps.newHashMap();
-        Map<String,Integer> countKeepersForFieldMap = Maps.newHashMap();
+        CountMap countForFieldMap = new CountMap();
+        CountMap countMissesRemainingForFieldMap = new CountMap();
+        CountMap countKeepersForFieldMap = new CountMap();
+        
+        int attributesToDrop = 0;
         
         // first pass is to set all of the hits to be kept, the misses to drop, and count em all
         for (Map.Entry<String,Attribute<? extends Comparable<?>>> de : document.entrySet()) {
@@ -62,9 +60,9 @@ public class LimitFields implements Function<Entry<Key,Document>,Entry<Key,Docum
                     log.trace("limitFieldsMap contains " + keyNoGrouping);
                 
                 Attribute<?> attr = de.getValue();
-                int keepers = get(keyNoGrouping, countKeepersForFieldMap);
-                int missesRemaining = get(keyNoGrouping, countMissesRemainingForFieldMap);
-                int total = get(keyNoGrouping, countForFieldMap);
+                int keepers = countKeepersForFieldMap.get(keyNoGrouping);
+                int missesRemaining = countMissesRemainingForFieldMap.get(keyNoGrouping);
+                int total = countForFieldMap.get(keyNoGrouping);
                 if (attr instanceof Attributes) {
                     Attributes attrs = (Attributes) attr;
                     Set<Attribute<? extends Comparable<?>>> attrSet = attrs.getAttributes();
@@ -74,6 +72,7 @@ public class LimitFields implements Function<Entry<Key,Document>,Entry<Key,Docum
                         } else {
                             value.setToKeep(false);
                             missesRemaining++;
+                            attributesToDrop++;
                         }
                         total++;
                     }
@@ -83,12 +82,13 @@ public class LimitFields implements Function<Entry<Key,Document>,Entry<Key,Docum
                     } else {
                         attr.setToKeep(false);
                         missesRemaining++;
+                        attributesToDrop++;
                     }
                     total++;
                 }
-                set(keyNoGrouping, countKeepersForFieldMap, keepers);
-                set(keyNoGrouping, countMissesRemainingForFieldMap, missesRemaining);
-                set(keyNoGrouping, countForFieldMap, total);
+                countKeepersForFieldMap.put(keyNoGrouping, keepers);
+                countMissesRemainingForFieldMap.put(keyNoGrouping, missesRemaining);
+                countForFieldMap.put(keyNoGrouping, total);
             }
         }
         
@@ -107,8 +107,8 @@ public class LimitFields implements Function<Entry<Key,Document>,Entry<Key,Docum
                     continue;
                 }
                 
-                int keepers = get(keyNoGrouping, countKeepersForFieldMap);
-                int missesRemaining = get(keyNoGrouping, countMissesRemainingForFieldMap);
+                int keepers = countKeepersForFieldMap.get(keyNoGrouping);
+                int missesRemaining = countMissesRemainingForFieldMap.get(keyNoGrouping);
                 int missesToSet = Math.min(limit - keepers, missesRemaining);
                 
                 // if we have misses yet to keep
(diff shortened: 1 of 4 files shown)


alerman (dismissed):

ivakegg:
Now that I think about it, I could avoid the hits and misses lists altogether.  I am already making two passes through the attributes.  Why not simply take one pass marking all hits to keep, and all misses to not keep.  Then based on whether more are needed for each attribute make another passes simply flipping the no-keep misses to keeps.  Working this again.

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
```diff
@@ -39,6 +32,31 @@ public LimitFields(Map<String,Integer> limitFieldsMap) {
             log.trace("limitFieldsMap set to:" + limitFieldsMap);
     }
     
+    protected void increment(String key, Map<String,Integer> map) {
```
jschmidt10: These 3 methods are all basically behavior on the map. Thoughts on a top level class (CountMap) or an inner class (FieldCountMap) that encapsulates this a little bit? 

```
fieldCountMap.increment(field);

fieldCountMap.decrement(field);
```

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
```diff
@@ -87,89 +88,127 @@ public Multimap<String,Attribute<? extends Comparable<?>>> load(String key) {
...
                     log.trace("limitFieldsMap contains " + keyNoGrouping);
+                
                 Attribute<?> attr = de.getValue();
-                int limit = this.limitFieldsMap.get(keyNoGrouping); // used below if you un-comment to get all hits
                 if (attr instanceof Attributes) {
                     Attributes attrs = (Attributes) attr;
                     Set<Attribute<? extends Comparable<?>>> attrSet = attrs.getAttributes();
                     for (Attribute<? extends Comparable<?>> value : attrSet) {
-                        manageHitsAndMisses(keyWithGrouping, keyNoGrouping, value, hitTermMap, hits, misses, countForFieldMap);
+                        if (isHit(keyWithGrouping, value, hitTermMap)) {
```
jschmidt10, line 70: Nice clean up. I dig it.

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
```diff
@@ -87,89 +88,127 @@ public Multimap<String,Attribute<? extends Comparable<?>>> load(String key) {
...
-                    reducedMap.put(keyWithGrouping, value);
-                    limit--;
-                }
+        
+        // second pass is to set any misses back to be kept if the limit allows
+        for (Map.Entry<String,Attribute<? extends Comparable<?>>> de : document.entrySet()) {
+            String keyWithGrouping = de.getKey();
+            String keyNoGrouping = keyWithGrouping;
+            // if we have grouping context on, remove the grouping context
+            if (keyNoGrouping.indexOf('.') != -1) {
```
jschmidt10: The computing of the key with no grouping also appears above. We could make a function and reuse:

```
String keyNoGrouping = removeGrouping(keyWithGrouping);
```

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
```diff
@@ -87,89 +88,127 @@ public Multimap<String,Attribute<? extends Comparable<?>>> load(String key) {
...
-            Multimap<String,Attribute<? extends Comparable<?>>> missMap = misses.getUnchecked(keyNoGrouping);
-            for (String keyWithGrouping : missMap.keySet()) {
-                for (Attribute<? extends Comparable<?>> value : missMap.get(keyWithGrouping)) {
-                    if (limit <= 0)
-                        break;
-                    reducedMap.put(keyWithGrouping, value);
-                    limit--;
+            
+            if (this.limitFieldsMap.containsKey(keyNoGrouping)) { // look for the key without the grouping context
+                int limit = this.limitFieldsMap.get(keyNoGrouping); // used below if you un-comment to get all hits
```
jschmidt10: The comment here is OBE'd I think. We are using the limit below.

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
```diff
@@ -87,89 +88,127 @@ public Multimap<String,Attribute<? extends Comparable<?>>> load(String key) {
...
+                // if we have misses yet to keep
+                if (missesToSet > 0) {
+                    
+                    Attribute<?> attr = de.getValue();
+                    if (attr instanceof Attributes) {
+                        Attributes attrs = (Attributes) attr;
+                        Set<Attribute<? extends Comparable<?>>> attrSet = attrs.getAttributes();
+                        
+                        for (Attribute<? extends Comparable<?>> value : attrSet) {
+                            // if this was a miss previously set to not keep
```
jschmidt10:
```suggestion
                            // if this was a miss previously set to keep now
```

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
```diff
@@ -87,89 +88,127 @@ public Multimap<String,Attribute<? extends Comparable<?>>> load(String key) {
...
+                                increment(keyNoGrouping, countKeepersForFieldMap);
+                                decrement(keyNoGrouping, countMissesRemainingForFieldMap);
+                                missesToSet--;
+                                if (missesToSet == 0) {
+                                    break;
+                                }
+                            }
+                        }
+                    } else {
+                        // if this was a miss previously set to not keep
```
jschmidt10:
```suggestion
                        // if this was a miss previously set to keep now
```

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
```diff
@@ -87,89 +88,127 @@ public Multimap<String,Attribute<? extends Comparable<?>>> load(String key) {
...
+        }
+        
+        // reduce the document to those to keep
+        document.reduceToKeep();
+        
+        for (String keyNoGrouping : countForFieldMap.keySet()) {
             // only generate an original count if a field was reduced
-            if (countForFieldMap.get(keyNoGrouping) > this.limitFieldsMap.get(keyNoGrouping)) {
-                limitedFieldCounts.put(keyNoGrouping + ORIGINAL_COUNT_SUFFIX, countForFieldMap.get(keyNoGrouping));
+            if (get(keyNoGrouping, countForFieldMap) > get(keyNoGrouping, countKeepersForFieldMap)) {
```
jschmidt10: Slight readability improvement would be to assign these two values to variables. You could reuse the first one:

```
int origFieldCount = get(keyNoGrouping, countForFieldMap);
int keepFieldCount = get(keyNoGrouping, countKeepersForFieldMap);

if (origFieldCount > keepFieldCount) 
  // ... use origFieldCount
```

warehouse/query-core/src/test/java/datawave/query/TestLimitReturnedGroupsToHitTermGroups.java
```diff
@@ -286,4 +286,36 @@ public void testMultipleIncludeGroupingFalse() throws Exception {
...
+    public void testLimitingToZero() throws Exception {
+        Map<String,String> extraParameters = new HashMap<>();
+        extraParameters.put("include.grouping.context", "false");
+        extraParameters.put("hit.list", "true");
+        extraParameters.put("limit.fields", "CANINE=0,BIRD=0");
+        extraParameters.put("return.fields", "CANINE,BIRD");
+        
+        String queryString = "filter:getAllMatches(CANINE,'.*e.*')";
+        
+        // CANINE is the hit, so that stays in. Only 1 bird is returned.
```
jschmidt10:
```suggestion
        // CANINE is the hit, so that stays in. No birds are returned.
```

jschmidt10:
Code looks good. All of my comments are small comment corrections or suggestions. Let me know your thoughts and I'll approve.

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
ivakegg: Ayye

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
ivakegg: good one.  will do

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
ivakegg: cool

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
ivakegg: The sentence was actually correct.  Rewording to me more clear however.

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
ivakegg: good idea.  I will apply that suggestion throughout the class.

warehouse/query-core/src/test/java/datawave/query/TestLimitReturnedGroupsToHitTermGroups.java
ivakegg: check.

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
```diff
@@ -87,89 +88,127 @@ public Multimap<String,Attribute<? extends Comparable<?>>> load(String key) {
...
-            if (log.isTraceEnabled()) {
-                log.trace("reducedMap:" + reducedMap);
-                log.trace("mapOfHits:" + hits.asMap());
-                log.trace("mapOfMisses:" + misses.asMap());
-            }
-            
+        }
+        
+        // reduce the document to those to keep
+        document.reduceToKeep();
```
FineAndDandy: only do this conditionally on having a reduction?

FineAndDandy (dismissed):

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
ivakegg: I actually removed those methods.  Only set and get methods now exist.  I can incapsolate those however

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
ivakegg: Not a bad idea.  Looking into it.

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
```diff
@@ -241,48 +250,26 @@ private void fillHitTermMap(Attribute<?> attr, Map<String,String> attrMap) {
...
-                if (log.isTraceEnabled())
-                    log.trace("removed " + entry.getKey() + " and its value " + entry.getValue() + " from document");
-            }
-        }
-        
-        for (Entry<String,Attribute<? extends Comparable<?>>> entry : reducedMap.entries()) {
-            document.put(entry, true);
-            if (log.isTraceEnabled())
-                log.trace("put " + entry + " into document");
+    protected int get(String key, Map<String,Integer> map) {
```
4a4f484e: Replace with map.getOrDefault(key, 0)?

ivakegg: Ah, good.  Thanks.

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
ivakegg: Creating a CountMap.  Thanks!

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
```diff
@@ -87,145 +58,186 @@ public Multimap<String,Attribute<? extends Comparable<?>>> load(String key) {
...
+                        Set<Attribute<? extends Comparable<?>>> attrSet = attrs.getAttributes();
+                        
+                        for (Attribute<? extends Comparable<?>> value : attrSet) {
+                            // if this was an attribute previously set to not keep, then it is one of the misses (not a hit)
+                            if (!value.isToKeep()) {
+                                value.setToKeep(true);
+                                keepers++;
+                                missesRemaining--;
+                                missesToSet--;
+                                attributesToDrop--;
```
jschmidt10, line 128: Do we need to set `foundMiss = true;` here?

ivakegg: oh crud... I did have that in there.  good catch!

jschmidt10 (approved):

alerman (approved):

jschmidt10 (approved):

PR #1710, opened 2022-09-30
Geo query optimization

jwomeara:
This optimization makes queries against geo fields behave somewhat similarly to geowave queries.  Multiple, tight ranges will be generated which should greatly improve query performance.


jwomeara:
This PR depends on the release of the following pr:
https://github.com/NationalSecurityAgency/datawave-type-utils/pull/12

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GeoFunctionsDescriptor.java
```diff
@@ -128,7 +144,7 @@ public JexlNode getIndexQuery(ShardQueryConfiguration config, MetadataHelper hel
                                         BoundedRange.create(JexlNodeFactory.createAndNode(Arrays.asList(geLatNode2, leLatNode2)))));
                         
                         // link em up
-                        returnNode = JexlNodeFactory.createAndNode(Arrays.asList(andNode1, andNode2));
+                        returnNode = JexlNodeFactory.createOrNode(Arrays.asList(andNode1, andNode2));
```
ivakegg, line 147: oh interesting.... I guess we would have been missing hits here.  However this situation is probably extremely rare.

ivakegg: Also can't this section reuse the code above (if args size == 3)?  As far as I can tell you are simply parsing the coordinates from the arguments slightly differently.  Isn't the subsequent jexl tree essentially the same except that you are not  "optimizing"?  Speaking of which, why aren't we optimizing the ranges in this case?

warehouse/query-core/src/main/java/datawave/query/util/GeoUtils.java
```diff
@@ -0,0 +1,475 @@
...
+            
+            double maxLat = Collections.max(latitudes);
+            double maxLon = Collections.max(longitudes);
+            
+            Coordinate[] coords = new Coordinate[5];
+            coords[0] = new Coordinate(minLon, minLat);
+            coords[1] = new Coordinate(maxLon, minLat);
+            coords[2] = new Coordinate(maxLon, maxLat);
+            coords[3] = new Coordinate(minLon, maxLat);
+            coords[4] = new Coordinate(minLon, minLat);
```
ivakegg: save an object creation: coords[4] = coords[0];

warehouse/query-core/src/main/java/datawave/query/util/GeoUtils.java
```diff
@@ -0,0 +1,475 @@
...
+        
+        double minLon = latLon[1];
+        double maxLon = minLon + 0.00001;
+        
+        Coordinate[] coords = new Coordinate[5];
+        coords[0] = new Coordinate(minLon, minLat);
+        coords[1] = new Coordinate(maxLon, minLat);
+        coords[2] = new Coordinate(maxLon, maxLat);
+        coords[3] = new Coordinate(minLon, maxLat);
+        coords[4] = new Coordinate(minLon, minLat);
```
ivakegg: save an object creation: coords[4] = coords[0];

ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GeoFunctionsDescriptor.java
jwomeara: Yeah, I was surprised by the use of and over or. Not sure how we missed that all these years. This case is different. This path doesn't use the Geo index at all. This sets up individual bounded range queries against the individual lat and long fields, so the optimizations don't apply at all.

warehouse/query-core/src/main/java/datawave/query/util/GeoUtils.java
jwomeara: Easy enough. I'll make the change throughout.

warehouse/query-core/src/main/java/datawave/query/util/GeoUtils.java
jwomeara: done

warehouse/query-core/src/main/java/datawave/query/util/GeoUtils.java
jwomeara: done

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GeoFunctionsDescriptor.java
alerman, line 147: Maybe rename andNode1 and andNode2 to orNodeN as well?

jwomeara: andNode1 and andNode2 are AND nodes though...  lol.  We are ORing together two AND nodes.

pom.xml
```diff
@@ -86,7 +87,7 @@
         <version.microservice.metadata-utils>1.10</version.microservice.metadata-utils>
         <version.microservice.metrics-reporter>1.2</version.microservice.metrics-reporter>
         <version.microservice.query-metric-api>1.5.7</version.microservice.query-metric-api>
-        <version.microservice.type-utils>1.8</version.microservice.type-utils>
+        <version.microservice.type-utils>1.11</version.microservice.type-utils>
```
apmoriarty, line 90: Recording the changeset here so people don't have to go looking
- added HitTermType
- added new date formats for normalization
- GeoNormalizer ignores already-normalized values

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GeoFunctionsDescriptor.java
apmoriarty: I don't think it needs to be handled in this merge request, but if the parent is an OR node this return point would create an unflattened tree.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GeoFunctionsDescriptor.java
```diff
@@ -173,26 +189,47 @@ public JexlNode getIndexQuery(ShardQueryConfiguration config, MetadataHelper hel
...
             return returnNode;
         }
         
-        public static JexlNode getIndexNode(JexlNode fieldsNode, double minLon, double maxLon, double minLat, double maxLat, String splitChar) {
-            JexlNode indexNode;
-            List<JexlNode> indexNodes = Lists.newArrayList();
-            if (fieldsNode.jjtGetNumChildren() > 1) {
-                for (int i = 0; i < fieldsNode.jjtGetNumChildren(); i++) {
-                    JexlNode kid = JexlASTHelper.dereference(fieldsNode.jjtGetChild(i));
+        public static List<String> getFieldNames(JexlNode node) {
```
apmoriarty, line 197: Why not use `JexlNodes.getIdentifiers()` ? Do the field names ever need to be deconstructed?

warehouse/query-core/src/main/java/datawave/query/util/GeoUtils.java
```diff
@@ -0,0 +1,487 @@
...
+     * 
+     * @param latitude
+     *            the latitude
+     * @param longitude
+     *            the longitude
+     * @return a geo index
+     */
+    public static String latLonToIndex(double latitude, double longitude) {
+        double latShift = latitude + 90.0;
+        double lonShift = longitude + 180.0;
```
apmoriarty, line 266: Consider adding constants for 90 degrees and 180 degrees.

warehouse/query-core/src/test/java/datawave/query/jexl/functions/GeoFunctionsDescriptorTest.java
```diff
@@ -16,10 +17,21 @@ public void testMultiFieldGeoFunction() throws Exception {
...
+        // @formatter:off
         Assert.assertEquals(
-                        "(((_Bounded_ = true) && (FIELD_1 >= '-10.0|-10.0' && FIELD_1 <= '10.0|10.0')) || ((_Bounded_ = true) && (FIELD_2 >= '-10.0|-10.0' && FIELD_2 <= '10.0|10.0')))",
+                "(" +
+                        "((_Bounded_ = true) && (FIELD_1 >= '018700..0000000000' && FIELD_1 <= '018900..0000000000')) || " +
+                        "((_Bounded_ = true) && (FIELD_1 >= '019000..0000000000' && FIELD_1 <= '020000..0000000000')) || " +
+                        "((_Bounded_ = true) && (FIELD_1 >= '110800..0000000000' && FIELD_1 <= '110900..0000000000')) || " +
+                        "((_Bounded_ = true) && (FIELD_2 >= '018700..0000000000' && FIELD_2 <= '018900..0000000000')) || " +
+                        "((_Bounded_ = true) && (FIELD_2 >= '019000..0000000000' && FIELD_2 <= '020000..0000000000')) || " +
+                        "((_Bounded_ = true) && (FIELD_2 >= '110800..0000000000' && FIELD_2 <= '110900..0000000000'))" +
```
apmoriarty, line 31: Much easier to read, this is great

warehouse/query-core/src/test/java/datawave/query/util/GeoUtilsTest.java
```diff
@@ -0,0 +1,99 @@
...
+public class GeoUtilsTest {
+    
+    @Test
+    public void optimizeSinglePointTest() throws Exception {
+        String wktString = "POINT(12.34567890123 9.87654321098)";
+        Geometry geom = new WKTReader().read(wktString);
+        
+        List<String[]> optimizedRanges = GeoUtils.generateOptimizedIndexRanges(geom, 32);
+        
+        Assert.assertEquals(1, optimizedRanges.size());
```
apmoriarty, line 20: static imports for junit methods?

apmoriarty (dismissed):
Comments are largely cosmetic, LGTM

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GeoFunctionsDescriptor.java
jwomeara: All the getIndexQuery method is doing is using the parameters from the ASTFunctionNode to generate the node that represents the index query.  It isn't manipulating the tree at all, so flattening isn't really this method's concern.

FunctionIndexQueryExpansionVisitor calls getIndexQuery and ANDs the resulting index query JexlNode with the original ASTFunctionNode.  The original ASTFunctionNode is then replaced by the ANDed ASTFunctionNode+IndexQueryNode combo.  

In short, I don't really see the need to flatten anything here as this is simple replacing a function node with the expanded function + index query node.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GeoFunctionsDescriptor.java
jwomeara: We're not modifying the ASTFunctionNode.  We just need the string field names in order to generate the index terms and ranges.

warehouse/query-core/src/main/java/datawave/query/util/GeoUtils.java
jwomeara: Fair enough.  These are only used here though, fwiw.

warehouse/query-core/src/test/java/datawave/query/util/GeoUtilsTest.java
jwomeara: tomato tomato?

jwomeara:
> Comments are largely cosmetic, LGTM

Thanks for the review.

apmoriarty (approved):

billoley (approved):

PR #1715, opened 2022-10-05
Add multi-fielded versions of the isNull and isNotNull functions to m…

### LINKED ISSUE #1604: Add multi-fielded isNull/isNotNull filter function
The EvaluationPhaseFilterFunctions class has methods for single fielded `isNull` and `isNotNull` functions. When a multi-fielded `isNull` or `isNotNull` is in a query it is evaluated using the single fielded method and produces an incorrect evaluation.

Add multi-fielded versions of the `isNull` and `isNotNull` functions.

jjonesy16:
Ticket Details:  "The EvaluationPhaseFilterFunctions class has methods for single fielded isNull and isNotNull functions. When a multi-fielded isNull or isNotNull is in a query it is evaluated using the single fielded method and produces an incorrect evaluation."

- Added method overloads for EvaluationPhaseFilterFunctions IsNull and isNotNull that will except an Iterable<?> parameter and performed original method logic for each object found in new parameter.
- Added unit tests to EvaluationPhaseFilterFunctionsTest IsNotNullTests class.
- Added unit tests to EvaluationPhaseFilterFunctionsTest IsNullTests class.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
@@ -1,7 +1,6 @@
 package datawave.query.jexl.functions;
 
 import com.google.common.base.CharMatcher;
-import com.google.common.collect.Iterables;
 import com.google.common.collect.Sets;
 import datawave.query.attributes.Attribute;
 import datawave.query.attributes.ValueTuple;
@@ -9,7 +8,6 @@ import datawave.query.jexl.JexlPatternCache;
 import datawave.query.collections.FunctionalSet;
 import datawave.util.OperationEvaluator;
 import org.apache.commons.collections4.SetUtils;
-import org.apache.hadoop.yarn.webapp.hamlet.HamletSpec;
 import org.apache.log4j.Logger;
 
 import java.text.DateFormat;
@@ -50,6 +48,18 @@ public class EvaluationPhaseFilterFunctions {
     public static final String CASE_INSENSITIVE = ".*\\(\\?[idmsux]*-[dmsux]*i[idmsux]*\\).*";
     
     private static final Logger log = Logger.getLogger(EvaluationPhaseFilterFunctions.class);
+    
+    /**
+     * This public property is required for unit testing purposes. Required for proper unit test data-context reset. Referenced in the DatawaveInterpreterTest
+     * class while testing EvaluationPhaseFilterFunctions.
+     *
+     * @param multiSetIsNotNullEmptySetFound
+     */
+    public static void setMultiSetIsNotNullEmptySetFound(boolean multiSetIsNotNullEmptySetFound) {
+        EvaluationPhaseFilterFunctions.multiSetIsNotNullEmptySetFound = multiSetIsNotNullEmptySetFound;
+    }
+    
+    private static boolean multiSetIsNotNullEmptySetFound = false;
     
     public static boolean occurrence(Iterable<?> fieldValues, String operator, int count) {
         return compareSizeToCount(fieldValues, operator, count);
@@ -157,21 +167,22 @@ public class EvaluationPhaseFilterFunctions {
      *
      * @param iterable
      *            Collection of objects to iterate and utilize to populate return list of FunctionalSet objects.
-     * @return List of {@link FunctionalSet} of hit terms found.
+     * @return the {@link FunctionalSet} of hit terms found
      */
     public static FunctionalSet<ValueTuple> isNotNull(Iterable<?> iterable) {
+        // Return variable containing 1st valid hit / result.
         FunctionalSet<ValueTuple> firstResultSet = null;
         
-        if (iterable != null) {
+        if (iterable != null && !multiSetIsNotNullEmptySetFound) {
             for (Object o : iterable) {
                 FunctionalSet<ValueTuple> currentResultSet = isNotNull(o);
                 
-                if (currentResultSet.equals(FunctionalSet.emptySet())) {
-                    return FunctionalSet.emptySet();
-                } else if (firstResultSet == null)
+                if (firstResultSet == null) {
                     firstResultSet = currentResultSet;
+                }
             }
         } else {
+            multiSetIsNotNullEmptySetFound = true;
             return FunctionalSet.emptySet();
         }
         
@@ -190,7 +201,7 @@ public class EvaluationPhaseFilterFunctions {
     }
     
     /**
-     * Returns whether {@code iterable} is considered an equivalently null field value.
+     * Returns whether {@code iterable} is contains an equivalently null field value.
      *
      * @param iterable
      *            a iterable collection of objects.
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
@@ -1,6 +1,7 @@
 package datawave.query.jexl.functions;
 
 import com.google.common.base.CharMatcher;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Sets;
 import datawave.query.attributes.Attribute;
 import datawave.query.attributes.ValueTuple;
@@ -8,6 +9,7 @@ import datawave.query.jexl.JexlPatternCache;
 import datawave.query.collections.FunctionalSet;
 import datawave.util.OperationEvaluator;
 import org.apache.commons.collections4.SetUtils;
+import org.apache.hadoop.yarn.webapp.hamlet.HamletSpec;
 import org.apache.log4j.Logger;
 
 import java.text.DateFormat;
@@ -140,7 +142,6 @@ public class EvaluationPhaseFilterFunctions {
                 if (!values.isEmpty()) {
                     return values.stream().map(EvaluationPhaseFilterFunctions::getHitTerm).collect(Collectors.toCollection(FunctionalSet::new));
                 }
-                
             } else {
                 return FunctionalSet.singleton(getHitTerm(fieldValue));
             }
@@ -151,25 +152,30 @@ public class EvaluationPhaseFilterFunctions {
     
     /**
      * Behaves similarly to isNotNull(Object) except with multiple parameters. When evaluating multiple parameters, every parameter must meet the criteria or an
-     * emptySet will be returned. Returns List of {@link FunctionalSet} hit terms found foreach object in {@code iterable} collection. If {@code iterable} is
-     * non-empty collection of value tuples, a List of {@link FunctionalSet} containing the hit terms from each value in {@code iterable} collection will be
-     * returned.
+     * emptySet will be returned. Returns 1st {@link FunctionalSet} hit term found in {@code iterable} collection. If {@code iterable} contains a null value, an
+     * empty FunctionalSet is returned.
      *
      * @param iterable
      *            Collection of objects to iterate and utilize to populate return list of FunctionalSet objects.
      * @return List of {@link FunctionalSet} of hit terms found.
      */
-    public static List<FunctionalSet<ValueTuple>> isNotNull(Iterable<?> iterable) {
-        List<FunctionalSet<ValueTuple>> resultSet = new ArrayList<>();
+    public static FunctionalSet<ValueTuple> isNotNull(Iterable<?> iterable) {
+        FunctionalSet<ValueTuple> firstResultSet = null;
+        
         if (iterable != null) {
             for (Object o : iterable) {
-                resultSet.add(isNotNull(o));
+                FunctionalSet<ValueTuple> currentResultSet = isNotNull(o);
+                
+                if (currentResultSet.equals(FunctionalSet.emptySet())) {
+                    return FunctionalSet.emptySet();
+                } else if (firstResultSet == null)
+                    firstResultSet = currentResultSet;
             }
         } else {
-            resultSet.add(FunctionalSet.emptySet());
+            return FunctionalSet.emptySet();
         }
         
-        return resultSet;
+        return firstResultSet;
     }
     
     /**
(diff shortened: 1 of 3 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
```diff
@@ -145,6 +146,29 @@ public static FunctionalSet<ValueTuple> isNotNull(Object fieldValue) {
...
+     * tuples, a {@link FunctionalSet} containing the hit terms from each value in the {@code fieldValue1} collection will be returned. Otherwise, an empty
+     * {@link FunctionalSet} will be returned.
+     *
+     * @param fieldValue1
+     *            1st field value to evaluate.
+     * @param fieldValue2
+     *            2nd field value to evaluate.
+     * @return the {@link FunctionalSet} of hit terms found
+     */
+    public static FunctionalSet<ValueTuple> isNotNull(Object fieldValue1, Object fieldValue2) {
```
apmoriarty: This method signature should look like `isNotNull(Iterable<?> iterable)`. 

For an example, see the getSizeOf methods
- `getSizeOf(Object fieldValue)` (single fielded)
- `getSizeOf(Iterable<?> iterable)` (multi fielded)

apmoriarty (changes requested):

warehouse/query-core/src/test/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctionsTest.java
```diff
@@ -476,46 +476,97 @@ private boolean resultForDefaultOperator(int count) {
...
         }
         
+        // Verify that a non-null value tuple returns a set with that value tuple with use of multiple filter fields.
+        @Test
+        public void testNonNullValuesWithMultipleFields() {
+            ValueTuple valueTuple = toValueTuple("F00.1,BAR,bar");
+            givenFieldValue1(valueTuple);
+            givenFieldValue2("F00.1");
+            
+            assertThat(resultForMultipleFieldValues()).containsExactly(valueTuple);
```
matthpeterson: I'm wondering about "F00.1" - should it have been in the resultForMultipleFieldValues?  I'm not very familiar with the use cases for these FilterFunctions so please forgive my ignorance.

matthpeterson: Is it that value1 and value2's contents overlap?

warehouse/query-core/src/test/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctionsTest.java
```diff
@@ -476,46 +476,97 @@ private boolean resultForDefaultOperator(int count) {
...
+            ValueTuple valueTuple2 = toValueTuple("FOO.1,ZOOM,zoom");
+            givenFieldValue1(Sets.newHashSet(valueTuple1, valueTuple2));
+            
+            ValueTuple valueTuple3 = toValueTuple("FOO.2,BAT,bat");
+            ValueTuple valueTuple4 = toValueTuple("FOO.2,ZAP,zap");
+            givenFieldValue2(Sets.newHashSet(valueTuple3, valueTuple4));
+            
+            assertThat(resultForMultipleFieldValues()).containsExactlyInAnyOrder(valueTuple1, valueTuple2);
+        }
+
```
matthpeterson: I'm reviewing the tests before the code and am thinking it might be worth adding more tests.  Here are a few scenarios to consider:
1. givenFieldValue1 is a collection containing multiple entries and givenFieldValue2 is empty
2. reverse of scenario 1, where value1 is empty and value2 is a collection
3. givenFieldValue1 is a collection containing multiple entries and givenFieldValue2 is null
4. reverse of scenario 3
5. reverse of testEmptyCollectionWithMultipleFields, where value2 is a non-empty collection and value1 is empty
6. reverse of testNullValueWithMultipleFields with null for 2, value 1 is a non-empty string
7. both value1 and value 2 are null
8. both value1 and value 2 are empty

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
```diff
@@ -1,6 +1,7 @@
 package datawave.query.jexl.functions;
 
 import com.google.common.base.CharMatcher;
+import com.google.common.base.Function;
```
matthpeterson: This might be an unused import.  I don't see Function referenced.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
```diff
@@ -145,6 +146,29 @@ public static FunctionalSet<ValueTuple> isNotNull(Object fieldValue) {
                 return FunctionalSet.singleton(getHitTerm(fieldValue));
             }
         }
+        
+        return FunctionalSet.emptySet();
+    }
+    
+    /**
+     * Returns a {@link FunctionalSet} of hit terms found for {@code fieldValue1}. If {@code fieldValue1} is a singular value tuple, a singleton
```
matthpeterson: The first sentence may need to be adjusted unless I'm misunderstanding the intended functionality.  Can this method's comment be made to reference the single param method's comments, i.e. "behaves similiarly to isNotNull(Object) except on multiple parameters.  When evaluating multiple parameters, every parameter must meet the criteria or an emptySet will be returned."  I'm not sure that's accurately describing the intention.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
```diff
@@ -145,6 +146,29 @@ public static FunctionalSet<ValueTuple> isNotNull(Object fieldValue) {
...
+     *            1st field value to evaluate.
+     * @param fieldValue2
+     *            2nd field value to evaluate.
+     * @return the {@link FunctionalSet} of hit terms found
+     */
+    public static FunctionalSet<ValueTuple> isNotNull(Object fieldValue1, Object fieldValue2) {
+        FunctionalSet<ValueTuple> matches = isNotNull(fieldValue1);
+        
+        if (matches != FunctionalSet.EMPTY_SET && !((FunctionalSet) isNotNull(fieldValue2)).isEmpty()) {
+            return matches;
```
matthpeterson: If it's a reasonable goal to include matching terms for fieldValue2, the results of isNotNull on each of the parameters could be merged together.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
```diff
@@ -145,6 +146,29 @@ public static FunctionalSet<ValueTuple> isNotNull(Object fieldValue) {
...
+     * @param fieldValue1
+     *            1st field value to evaluate.
+     * @param fieldValue2
+     *            2nd field value to evaluate.
+     * @return the {@link FunctionalSet} of hit terms found
+     */
+    public static FunctionalSet<ValueTuple> isNotNull(Object fieldValue1, Object fieldValue2) {
+        FunctionalSet<ValueTuple> matches = isNotNull(fieldValue1);
+        
+        if (matches != FunctionalSet.EMPTY_SET && !((FunctionalSet) isNotNull(fieldValue2)).isEmpty()) {
```
matthpeterson: It might be prudent to use ! FunctionalSet.EMPTY_SET.equals(matches) instead of !=.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
```diff
@@ -159,6 +183,25 @@ public static boolean isNull(Object fieldValue) {
...
+     * Returns whether {@code fieldValue1} or {@code fieldValue2} is considered an equivalently null field value.
+     *
+     * @param fieldValue1
+     *            the 1st fieldValue
+     * @param fieldValue2
+     *            the 2nd fieldValue
+     * @return true if {@code fieldValue1} or {@code fieldValue2} is a null {@link Object} or an empty {@link Collection}, or false otherwise
+     */
+    public static boolean isNull(Object fieldValue1, Object fieldValue2) {
+        if (isNull(fieldValue1)) {
```
matthpeterson: This is just a stylistic thought which doesn't require any change.  This method could be modified to be:
```
return isNull(fieldValue1) || isNull(fieldValue2);
```

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
jjonesy16: Hey Drew, thank you so much for the extra guidance!  This ticket makes a lot more sense now.  

I have one quick question...I'm struggling with the return type for the isNotNull methods.  Does the return type need to be of type, "FunctionalSet<ValueTuple>"?  To me, it seems that a True/False would be a better response from the isNotNull methods, but I also do not have a great understanding of how query side of the house either...  Any direction on this return type would be much appreciated.  The isNull is just a boolean, so that one makes good sense to me.

Thanks Buddy!

apmoriarty: I think for the sake of consistency, the iterable version of isNotNull method will have to return a functional set just like it's single-fielded counterpart.

jjonesy16: Would a List<FunctionalSet> for the return type work?  That way I could put the FuctionalSet returned by the original method in a List for each item in the Iterable parameter passed in.  Let me know when you can, thanks!

jjonesy16: Modified code to utilize more appropriate method signature with multi-fielded filter parameters for the isNull and isNotNull overloaded methods.  Updated method signatures now utilize "Iterable<?> iterable" parameters.  Updated unit tests to test new method signatures and logic.

warehouse/query-core/src/test/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctionsTest.java
jjonesy16: Significant updates to original branch logic.  Please re-review as time permits.  Code in this comment has been replaced.

warehouse/query-core/src/test/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctionsTest.java
jjonesy16: There have been significant changes from the initial MR, including the unit tests.  Can you please re-review when you have a chance.  Thanks!

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
jjonesy16: Modified method comments to have a better description.  Thank you for the suggestion!

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
jjonesy16: There have been significant changes from the initial MR, including the unit tests.  Can you please re-review when you have a chance.  Thanks!

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
jjonesy16: There have been significant changes from the initial MR, including the unit tests. Can you please re-review when you have a chance. Thanks!

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
jjonesy16: Un-used imports removed.  Thanks!

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
jjonesy16: There have been significant changes from the initial MR, including the unit tests. Can you please re-review when you have a chance. Thanks!

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
jjonesy16: Please review as time allows.  Thanks!

apmoriarty:
I was wrong when I wrote the original ticket. This is not as simple as adding iterable versions of the isNull/isNotNull functions. There is additional work required in the DatawaveInterpreter to detect these functions and then pass in the correct hits per field. Let's discuss later today if you have time.

warehouse/query-core/src/test/java/datawave/query/jexl/DatawaveInterpreterTest.java
```diff
@@ -297,32 +296,27 @@ public void testFilterFunctionIsNull() {
...
     }
     
-    @Ignore
     @Test
     public void testFilterFunctionMultiFieldedIsNull() {
         // Once #1604 is complete these tests will evaluate correctly
         
         // multi field, all present
-        String query = "FOO == 'bar' && filter:isNull(FOO || FOO)";
+        String query = "FOO == 'bar' && (filter:isNull(FOO) || filter:isNull(FOO))";
```
avgAGB, line 304: To clarify here, will the checks eventually be returned to the original logic once the ticket here is merged, or will it be dependent on #1604  still?

apmoriarty:
Close this PR per discussion comments

PR #1717, opened 2022-10-05
fixes #1716: Enrich query logic responses

### LINKED ISSUE #1716: Need the ability to enrich query logic responses in a general manner
Query logic responses may need to be enriched (e.g. enriched with the originating system, or whatever else).  This should be doable in a generic fashion such that all query logics can be easily configured as such.

ivakegg:
* Added a ResponseEnricher concept whichp providces a shim between the query logic responses and the client.

### DIFF
--- a/web-services/query/src/main/java/datawave/webservice/query/logic/BaseQueryLogic.java
+++ b/web-services/query/src/main/java/datawave/webservice/query/logic/BaseQueryLogic.java
@@ -196,7 +196,7 @@ public abstract class BaseQueryLogic<T> implements QueryLogic<T> {
     }
     
     @Override
-    public QueryLogicTransformer getEnrichedTransformer(Query settings) {
+    public final QueryLogicTransformer getEnrichedTransformer(Query settings) {
         QueryLogicTransformer transformer = this.getTransformer(settings);
         if (responseEnricherBuilder != null) {
             transformer.setResponseEnricher(responseEnricherBuilder.withConfig(getConfig()).withMarkingFunctions(getMarkingFunctions())
--- a/web-services/query/src/main/java/datawave/webservice/query/logic/BaseQueryLogic.java
+++ b/web-services/query/src/main/java/datawave/webservice/query/logic/BaseQueryLogic.java
@@ -199,8 +199,15 @@ public abstract class BaseQueryLogic<T> implements QueryLogic<T> {
     public final QueryLogicTransformer getEnrichedTransformer(Query settings) {
         QueryLogicTransformer transformer = this.getTransformer(settings);
         if (responseEnricherBuilder != null) {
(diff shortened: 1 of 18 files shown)


web-services/query/src/main/java/datawave/webservice/query/logic/BaseQueryLogic.java
```diff
@@ -193,6 +195,16 @@ public Iterator<T> iterator() {
         return iterator;
     }
     
+    @Override
+    public QueryLogicTransformer getEnrichedTransformer(Query settings) {
```
FineAndDandy: Is there a reason not to reuse getTransformer() instead of creating a new method? If an enriched transformer is configured we probably always want to use it?

ivakegg: The reason was that the underlying transformer still needed to be accessible in various places.  Also the getTransformer method is overridden almost everywhere.  I want to ensure that the response transform shim is applied no matter what underlying transformer is returned.

ivakegg: Now that I think about it, I should probably make this method final to ensure that nothing overrides it.

drewfarris: FWIW, I had the same question as @FineAndDandy as well here. 

Consider renaming `getTransformer` and its usages/implementations to `getBaseTransformer`, documenting it as an internal API (maybe making it protected) and renaming `getEnrichedTransformer` to `getTransformer` and making it `public final`?

(There's a good chance this could make things more "hairy" in the long run, I should maybe work through this exercise myself before suggesting it)

web-services/query/src/main/java/datawave/webservice/query/logic/BaseQueryLogic.java
```diff
@@ -193,6 +195,16 @@ public Iterator<T> iterator() {
...
+        QueryLogicTransformer transformer = this.getTransformer(settings);
+        if (responseEnricherBuilder != null) {
+            transformer.setResponseEnricher(responseEnricherBuilder.withConfig(getConfig()).withMarkingFunctions(getMarkingFunctions())
+                            .withResponseObjectFactory(responseObjectFactory).withPrincipal(getPrincipal()).build());
+        }
```
drewfarris: I like the builder pattern but really picky about the formatting...

```suggestion
        if (responseEnricherBuilder != null) {
            //@formatter:off
            ResponseEnricher enricher = responseEnricherBuilder
                    .withConfig(getConfig())
                    .withMarkingFunctions(getMarkingFunctions())
                    .withResponseObjectFactory(responseObjectFactory)
                    .withPrincipal(getPrincipal())
                    .build();
            //@formatter:on
            transformer.setResponseEnricher(enricher);
        }
```

drewfarris (dismissed):
Approved. Just added some food for thought and a minor suggestion.

PR #1729, opened 2022-10-17
Feature/version filter

### LINKED ISSUE #1740: Datatype Version Filter
Create new Accumulo filter which can accept either one timestamp version per datatype or any timestamp version greater than or equal to a datatype's configured datatype.

matthpeterson:
Adding a new filter which can accept either one timestamp version per datatype or any timestamp version greater than or equal to a datatype's configured datatype.

### DIFF
--- a/warehouse/core/src/main/java/datawave/iterators/filter/ageoff/DataTypeParser.java
+++ b/warehouse/core/src/main/java/datawave/iterators/filter/ageoff/DataTypeParser.java
@@ -1,0 +1,154 @@
+package datawave.iterators.filter.ageoff;
+
+import org.apache.accumulo.core.data.ArrayByteSequence;
+import org.apache.accumulo.core.data.ByteSequence;
+import org.apache.accumulo.core.data.Key;
+
+/**
+ * Originally extracted from DataTypeAgeOffFilter
+ *
+ * Contains a single public static method for parsing the datatype (as a ByteSequence) from a key.
+ *
+ * It can parse the various shard schema key structures including the d, fi, tf, and event schemas.
+ *
+ * It can also parse an index table key.
+ */
+public class DataTypeParser {
+    
+    private static final char FI_CF_CHAR_1 = 'f';
+    private static final char FI_CF_CHAR_2 = 'i';
+    
+    private static final char TF_CF_CHAR_2 = 'f';
+    private static final char TF_CF_CHAR_1 = 't';
+    
+    private static final char DOC_CF_CHAR = 'd';
+    
+    private static final int NULL_BYTE = 0x00;
+    
+    // shard ID's minimum length = sum of the minimum length of its components: year + month + day + underscore + single digit partition
+    private static final int MIN_LENGTH_SHARD_ID = 10;
+    
+    /**
+     * @param key
+     *            Accumulo key to parse
+     * @param isIndexTable
+     *            true if the key is from an index table, false if the shard table
+     * @return a ByteSequence containing the data type name, possibly returning null or empty for unexpected column data.
+     */
+    public static final ByteSequence parseKey(Key key, boolean isIndexTable) {
+        /**
+         * Supports the shard and index table. There should not be a failure, however if either one is used on the incorrect table
+         */
+        if (isIndexTable) {
+            return extractDatatypeForIndexTable(key);
+        } else {
+            return extractDatatypeForShardTable(key);
+        }
+    }
+    
+    private static final ByteSequence extractDatatypeForShardTable(Key k) {
+        // ASSUMES THAT THE KEY STARTS WITH A CORRECTLY SIZED BYTE ARRAY
+        byte[] cf = k.getColumnFamilyData().getBackingArray();
+        
+        if (cf.length >= 3 && cf[0] == FI_CF_CHAR_1 && cf[1] == FI_CF_CHAR_2 && cf[2] == NULL_BYTE) {
+            return parseDataTypeForFi(k);
+        } else if (cf.length == 2 && cf[0] == TF_CF_CHAR_1 && cf[1] == TF_CF_CHAR_2) {
+            return parseDataTypeForDocOrTf(k);
+        } else if (cf.length == 1 && cf[0] == DOC_CF_CHAR) {
+            return parseDataTypeForDocOrTf(k);
+        } else {
+            return parseDataTypeForEvent(cf);
+        }
+    }
+    
+    private static ArrayByteSequence parseDataTypeForEvent(byte[] cf) {
+        int nullIndex = -1;
+        
+        int cfLength = cf.length;
+        for (int i = 0; i < cfLength; i++) {
+            if (cf[i] == NULL_BYTE) {
+                nullIndex = i;
+                break;
+            }
+        }
+        // data column
+        if (nullIndex > 0) {
+            return new ArrayByteSequence(cf, 0, nullIndex);
+        }
+        return null;
+    }
+    
+    private static ArrayByteSequence parseDataTypeForDocOrTf(Key k) {
+        // get the column qualifier, so that we can use it throughout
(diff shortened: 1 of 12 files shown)


warehouse/core/src/main/java/datawave/iterators/filter/ageoff/DataTypeParser.java
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * @param key
+     *            Accumulo key to parse
+     * @param isIndexTable
+     *            true if the key is from an index table, false if the shard table
+     * @return a ByteSequence containing the data type name, possibly returning null or empty for unexpected column data.
+     */
+    public static final ByteSequence parseKey(Key key, boolean isIndexTable) {
+        /**
+         * Supports the shard and index table. There should not be a failure, however if either one is used on the incorrect table
```
jjonesy16: Help me understand this comment.  What happens if either one is used on the incorrect table?  "There should not be a failure, however if either one is used on the incorrect table" ________?

warehouse/core/src/main/java/datawave/iterators/filter/ageoff/LatestVersionFilter.java
```diff
@@ -0,0 +1,396 @@
...
+        // determine dataType's configured timestamp
+        long configuredTimestamp = filterConfiguration.timestampVersion;
+        
+        // apply logic for configured mode
+        switch (filterConfiguration.mode) {
+            case EQUAL:
+                return dataTimestamp == configuredTimestamp;
+            case GREATER_THAN_OR_EQUAL:
+                return dataTimestamp >= configuredTimestamp;
+            case UNDEFINED:
```
jjonesy16, line 114: What is the purpose of the UNDEFINED switch case?  Should an exception be raised, or maybe a warning logged?

warehouse/core/src/main/java/datawave/iterators/filter/ageoff/LatestVersionFilter.java
```diff
@@ -0,0 +1,396 @@
...
+        
+        String dataTypeListOptionValue = options.get(DATATYPE_LIST_OPTION_NAME);
+        if (Strings.isNullOrEmpty(dataTypeListOptionValue)) {
+            log.trace(DATATYPE_LIST_OPTION_NAME + " is null or empty.  Configure data type list.");
+            return configurationsMap;
+        }
+        
+        String[] dataTypes = StringUtils.split(dataTypeListOptionValue, ",");
+        for (String dataTypeStr : dataTypes) {
+            if (null == dataTypeStr || dataTypeStr.trim().length() == 0) {
```
jjonesy16, line 162: Would Strings.isNullOrEmpty() make more sense here?

warehouse/core/src/main/java/datawave/iterators/filter/ageoff/LatestVersionFilter.java
```diff
@@ -0,0 +1,396 @@
...
+        
+        // apply logic for configured mode
+        switch (filterConfiguration.mode) {
+            case EQUAL:
+                return dataTimestamp == configuredTimestamp;
+            case GREATER_THAN_OR_EQUAL:
+                return dataTimestamp >= configuredTimestamp;
+            case UNDEFINED:
+            default:
+                log.warn("DataType {} was not configured for 'eq' or 'gte'");
```
dtspence: Minor - it appeared that the data-type should be specified for the log-message, but was missing.

jjonesy16 (dismissed):
I left a couple minor comments, but all and all looks good to me.  Nice Job!

warehouse/core/src/main/java/datawave/iterators/filter/ageoff/DataTypeParser.java
```diff
@@ -0,0 +1,154 @@
...
+        if (isIndexTable) {
+            return extractDatatypeForIndexTable(key);
+        } else {
+            return extractDatatypeForShardTable(key);
+        }
+    }
+    
+    private static final ByteSequence extractDatatypeForShardTable(Key k) {
+        // ASSUMES THAT THE KEY STARTS WITH A CORRECTLY SIZED BYTE ARRAY
+        byte[] cf = k.getColumnFamilyData().getBackingArray();
```
avgAGB, line 50: for my own understanding, is there any worry about a situation where the byte array is of an incorrect size?

warehouse/core/src/main/java/datawave/iterators/filter/ageoff/DataTypeParser.java
matthpeterson: Thanks for pointing this out.  It's a comment from the old code that was extracted to this class.
https://github.com/NationalSecurityAgency/datawave/pull/1729/files#diff-8ae64ed5c5daf1aac02cb892e5fafbb8fdbdcc3498a8fff042f19465eafec5ebL126

I'll push changes to reword these old comments.

warehouse/core/src/main/java/datawave/iterators/filter/ageoff/LatestVersionFilter.java
matthpeterson: Sometimes table iterator options are modified and I'd rather not throw exceptions.  It's outputting a log.warn.  The code handles the situation where it's undefined and have sensible behavior: accept all data for compaction scopes (to avoid data loss) and hide all data for scan (to avoid accidentally accepting data intended to be rejected).

matthpeterson: I think this will flood the logs though so I'm thinking it makes more sense to only log when the datatype's configuration is read.  I'll make that change.

warehouse/core/src/main/java/datawave/iterators/filter/ageoff/LatestVersionFilter.java
matthpeterson: Strings.isNullOrEmpty doesn't do the trim

warehouse/core/src/main/java/datawave/iterators/filter/ageoff/LatestVersionFilter.java
matthpeterson: I'll push a change for this.

warehouse/core/src/main/java/datawave/iterators/filter/ageoff/DataTypeParser.java
matthpeterson: I left this comment intact from the legacy code that was moved here.  To me, it seems like this is a reasonable assumption.  If the key doesn't start with a correctly sized byte array, then I think it's reasonable if an exception were thrown.

avgAGB (approved):

jjonesy16 (approved):
Looks good.  Thank you for the comments / explanations returned.  Nice Job!

dtspence (approved):

warehouse/core/src/main/java/datawave/iterators/filter/ageoff/DataTypeParser.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+        return null;
+    }
+    
+    private static ArrayByteSequence parseDataTypeForDocOrTf(Key k) {
+        // Assumes that the key starts with a correctly sized byte array
+        final byte[] cq = k.getColumnQualifierData().getBackingArray();
+        
+        // don't need to check the last byte as we expect more than one null if formatted correctly
+        for (int i = 0; i < cq.length - 1; i++) {
```
ivakegg, line 78: so this code is identical to the parseDataTypeForEvent(byte[]).  Possible consolidation?

ivakegg (approved):

hlgp:
Ticket referenced in commits is closed citing a better workaround.

PR #1731, opened 2022-10-19
Bugfix/federated query

ivakegg:
This ensures that errors are passed through appropriately.  Also ensure that appropriate setters will pass through to the children query logics.


web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
```diff
@@ -109,14 +116,21 @@ public void run() {
                             results.add(last);
                         }
                     } catch (InterruptedException e) {
-                        log.warn("QueryLogic thread interrupted", e);
+                        // if this was on purpose, then just log and the loop will naturally exit
+                        if (interrupted) {
+                            log.warn("QueryLogic thread interrupted", e);
+                        }
+                        // otherwise something else unexpected happened
+                        throw new RuntimeException(e);
```
drewfarris: Based on the comment, it seems that you might want to put the `RuntimeException` in an `else` clause instead of always throwing a `RuntimeException` regardless of whether the `interrupted` flag is set.

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
```diff
@@ -109,14 +116,23 @@ public void run() {
...
                     resultCount++;
                 }
-                
+                success = true;
             } finally {
-                completionLatch.countDown();
-                log.trace("Finished thread: " + this.getName());
+                if (success) {
+                    completionLatch.countDown();
+                }
```
drewfarris, line 117: I'm trying to convince myself that we only want to call `countDown()` on `success`. Docs seems to suggest that until the latch reaches zero, an `await()` call elsewhere will block. If we don't `countDown()` on failure it stands to reason that the counter will never reach zero and thus something somewhere will fail to unblock. That said, I haven't completed an analysis to understand how this latch is truly used.

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
```diff
@@ -137,31 +160,45 @@ public CompositeQueryLogic() {}
...
-                                }
-                            });
-            
-            if (count > 1) {
-                throw new RuntimeException("More than one instance of query logic class configured with the same table: " + logic.getClass().getName());
+                                });
+                
+                if (count > 1) {
+                    throw new RuntimeException("More than one instance of query logic class configured with the same table: " + logic.getClass().getName());
+                }
```
drewfarris: This is an unusual way to perform this check. Not only that, `CollectionUtils.countMatches` is deprecated. 

How about we loop over the list of query logics putting a composite of their class/table name into a set. If the set's `add` method returns false, we can throw our `RuntimeException`. That way we don't have to traverse the `queryLogicsTable `n-squared times (once in the loop and then once for each item in the loop in `CollectionUtils.countMatches()`)

Would something like this achieve the same goal?
```java
        if (checkTables) {
            Set<String> dupeCheckSet = new HashSet<>();
            for (QueryLogic<?> logic : queryLogics) {
                final String key = logic.getClass().getName() + ":" + logic.getTableName();
                if (!dupeCheckSet.add(key)) {
                    throw new RuntimeException("More than one instance of query logic class configured with the same table: " + key);
                }
            }
        }
```

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
```diff
@@ -181,7 +218,12 @@ public GenericQueryConfiguration initialize(Connector connection, Query settings
                 holder.setMaxResults(logic.getMaxResults());
                 logicState.put(logic, holder);
             } catch (Exception e) {
-                log.info(e.getMessage() + " removing query logic " + logic.getClass().getName() + " from CompositeQuery");
+                // if all must initialize successfully, then pass up an exception
+                if (allMustInitialize) {
+                    throw new RuntimeException(e);
```
drewfarris, line 179: Consider adding a message about which class failed to initialize (this one) - the stack trace may not be clear if it is deep in the code (or may get trimmed unfortunately)

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
```diff
@@ -361,4 +403,82 @@ public Set<String> getExampleQueries() {
...
+        boolean canRun = false;
+        if (super.canRunQuery()) {
+            for (QueryLogic<?> logic : this.queryLogics) {
+                canRun = logic.canRunQuery();
+                if (canRun) {
+                    break;
+                }
+            }
+        }
+        return canRun;
```
drewfarris: Consider this slight simplification:
```suggestion
        if (super.canRunQuery()) {
            for (QueryLogic<?> logic : this.queryLogics) {
                if (logic.canRunQuery()) {
                    return true;
                }
            }
        }
        return false;
```

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogicResultsIterator.java
```diff
@@ -1,20 +1,21 @@
...
     protected static final Logger log = Logger.getLogger(CompositeQueryLogicResultsIterator.class);
     
-    private ArrayBlockingQueue<Object> results = null;
+    private final ArrayBlockingQueue<Object> results;
     private Object nextEntry = null;
-    private Object lock = new Object();
-    private CountDownLatch completionLatch = null;
+    private final Object lock = new Object();
+    private final CountDownLatch completionLatch;
+    private Throwable failure = null;
```
drewfarris: Will the `uncaughtException` method be called from a different thread from the `hasNext` call? If so, we need to make this field `volatile` or otherwise Atomic in some way so that the changes can be persisted across threads.

ivakegg: I believe volitile only is useful for the base types (long, short, boolean, etc).  I thought it did not make a difference for objects (i.e. pointers).  I will research.

ivakegg: Well, I am wrong.  I will make this volatile.

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
ivakegg: The latches are need to know when a query logic has completed.  If not successful, then an exception should have been logged which will also break the loop in the CompositeQueryLogicResultsIterator loop in the hasNext call.

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
ivakegg: I am removing this whole check tables thing entirely.  I realized that was only there to try and create uniqueness in the logicStates tree map (see the comparator that was at the top of the class).  Very hokey concept to start with.

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
ivakegg: I turned the info message into a warn after the allMustInitialize check.  Also includes the exception that was thrown.

ivakegg: oh I see, you would like a message before throwing... got it.

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
ivakegg: sure.

PR #1738, opened 2022-10-28
Block off param route for invalid page.size option

avgAGB:


web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -422,6 +423,11 @@ private QueryData validateQuery(String queryLogicName, MultivaluedMap<String,Str
         if (params != null) {
             for (Parameter pm : QueryUtil.parseParameters(params)) {
                 if (!queryParameters.containsKey(pm.getParameterName())) {
+                    if (pm.getParameterName().equals(INVALID_PAGESIZE)) {
+                        log.error("Invalid parameter: page.size");
```
matthpeterson: In the log and/or the error message, maybe it's worth specifying the name of the correct parameter to use.

avgAGB: Can do, I'll update this on next commit.

matthpeterson (approved):

dtspence (approved):

apmoriarty (approved):

PR #1741, opened 2022-10-30
in openjdk 11.0.17, there is a bug described here:

tomnelson:
https://stackoverflow.com/questions/55906161/weird-javadoc-error-with-jdk12-for-osgi-version-annotation Some non-javadoc annotations are causing the javadoc generation to fail. This patch makes the javadoc-plugin ignore the annotations, as it should.


keith-ratcliffe (approved):
Thanks for the fix!

PR #1742, opened 2022-11-01
Feature/upgrade guava

cmitchell:
Upgraded guava to v31.1-jre.  

Please ignore the first half of the commit history -- it was branched from a branch and then later not.

Every test was touched. Found/had issues with some of the tests that exercised guava, so upgraded to junit5 (jupiter) because I tracked it down to one of the new guava calls wanting/needing true async execution.  (This may have been IDE related if Keith didn't run into this)  I did not change any test logic, just apis. Tests that were @Ignored before this work are still (now) @Disabled.  Powermck is not a thing for junit5, in which case the equivalent Mockito or normal java reflection was used.   

The only test that is @Disabled (which replaced @Ignore) are 5 tests in eb-services/query/src/test/java/datawave/webservice/query/runner/ExtendedQueryExecutorBeanTest


hlgp:
OBE

PR #1750, opened 2022-11-04
fixes #1746: Updated the scanner factory to fix concurrency issues wi…

### LINKED ISSUE #1746: ConcurrentModificationException is thrown when closing ScannerFactory in ShardQueryLogic
In ShardQueryLogic.close() when creating a new array list from ScannerFactory.currentScanners() or ScannerFactory.currentSessions() it is possible that a ConcurrentModificationException will be thrown.  The creation of the new array list causes us to iterate over the elements of the set, which could (and is) still being modified when ScannerFactory.close() is called.  ScannerFactory.close() is not synchronized, so it is able to modify the set while other synchronized methods are acting on the set.  

Also, while ScannerFactory.currentSessions() does return an unmodifiable set, the underlying set is still modifiable, so this doesn't really do anything to protect against a race condition.  

We could solve this by making ScannerFactory.close() synchronized, and by updating ScannerFactory.currentScanners() and ScannerFactory.currentSessions() to return a NEW set when they are called, instead of the original set.

jwomeara:
…th the instances hash sets

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/tables/ScannerFactory.java
+++ b/warehouse/query-core/src/main/java/datawave/query/tables/ScannerFactory.java
@@ -214,11 +214,11 @@ public class ScannerFactory {
     }
     
     public synchronized Collection<ScannerBase> currentScanners() {
-        return new HashSet<>(instances);
+        return Collections.unmodifiableSet(instances);
     }
     
     public synchronized Collection<ScannerSession> currentSessions() {
-        return new HashSet<>(sessionInstances);
+        return Collections.unmodifiableSet(sessionInstances);
     }
     
     public synchronized boolean lockdown() {
--- a/warehouse/query-core/src/main/java/datawave/query/tables/ScannerFactory.java
+++ b/warehouse/query-core/src/main/java/datawave/query/tables/ScannerFactory.java
@@ -213,17 +213,19 @@ public class ScannerFactory {
         }
         return removed;
     }
-
+    
     /**
      * Returns a NEW collection of scanner instances to the caller.
+     * 
      * @return a NEW collection of scanners
      */
     public synchronized Collection<ScannerBase> currentScanners() {
         return new ArrayList<>(instances);
     }
-
+    
     /**
      * Returns a NEW collection of scanner session instances to the caller.
+     * 
      * @return a NEW collection of scanner session instances
      */
     public synchronized Collection<ScannerSession> currentSessions() {
--- a/warehouse/query-core/src/main/java/datawave/query/tables/ScannerFactory.java
+++ b/warehouse/query-core/src/main/java/datawave/query/tables/ScannerFactory.java
@@ -214,11 +214,11 @@ public class ScannerFactory {
     }
     
     public synchronized Collection<ScannerBase> currentScanners() {
-        return Collections.unmodifiableSet(instances);
+        return new HashSet<>(instances);
     }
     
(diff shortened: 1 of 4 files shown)


warehouse/query-core/src/main/java/datawave/query/tables/ScannerFactory.java
```diff
@@ -214,11 +214,11 @@ public synchronized boolean close(ScannerBase bs) {
     }
     
     public synchronized Collection<ScannerBase> currentScanners() {
-        return Collections.unmodifiableSet(instances);
+        return new HashSet<>(instances);
```
ivakegg: I understand synchronizing the close call below.  I do not understand why we removed the Collections.unmodifiableSet.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/tables/ScannerFactory.java
jwomeara: Why does it need to be unmodifiable?  We are creating a new set for the caller which ScannerFactory has no ownership of.  Why does the ScannerFactory need to control what the user does with that set?  What the caller does with that set has no impact on ScannerFactory, or the sets contained within.

matthpeterson (dismissed):

warehouse/query-core/src/main/java/datawave/query/tables/ScannerFactory.java
jwomeara: I changed this based on your feedback, but I am changing it back because simply returning an unmodifiable set without creating a new collection _creates_ the potential for a concurrent modification exception as I described in the original issue description, which is exactly what we're trying to avoid.  

As I explained in the issue description, we run into a concurrent modification exception _precisely because_ we are using the exact same set that is used internal to the ScannerFactory.  It _does not matter_ that the set returned is unmodifiable by the caller because it is _still_ modifiable by the ScannerFactory.  So, in ShardQueryLogic, if the ScannerFactory removes an element from either of those sets while ShardQueryLogic is iterating through those sets, a concurrent modification exception will be thrown.  This is why we need to return a copy of the scanners and sessions - so that the ScannerFactory can't modify the set while we are iterating through it.

As it pertains to performance, this implementation should actually produce less unnecessary references than the original implementation.  In the original implementation, we wrapped each set with an unmodifiable wrapper, and then in each of the places where this set was used, we created a new list from that collection.  In the current implementation, we have stopped creating the wrapper, and moved the creation of the new list to the call itself, so technically we are dropping one unnecessary reference - the wrapper.

If we are really concerned about the creation of a new list, then my recommendation would be to create synchronized shutdownCurrentScanners and shutdownCurrentSessions methods in the ScannerFactory that can be used to close all of the running scanners and instances in a synchronized fashion.  Note that if we go this route, then we can add back in the unmodifiable wrapper for the currentScanners and currentSessions methods, or just remove those methods _entirely_ since they are only ever used to close the running scanners and instances.

ivakegg (approved):

warehouse/query-core/src/main/java/datawave/query/tables/ScannerFactory.java
ivakegg: I see it now.  Sorry I was not understanding that previously.  Creating a new collection makes sense to me now.

drewfarris (approved):

apmoriarty (approved):

PR #1764, opened 2022-11-15
fixes #1763: Updates the QueryIterator to keep all valid ivarator cac…

### LINKED ISSUE #1763: Ivarator cache dir pruning was broken and excludes all but one cache dir
Previously, the QueryIterator was configured to remove entries from the list of ivarator cache dirs if they were not valid.  The last update made to that logic updated it to remove ALL ivarator cache directories except one, instead of just removing the ones that were not valid.   This can significantly reduce the number of available ivarator cache directories available to a query and adversely affect query performance.

jwomeara:
…he directories instead of only keeping one.


warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -286,29 +284,26 @@ public void init(SortedKeyValueIterator<Key,Value> source, Map<String,String> op
...
             try {
                 Path basePath = new Path(config.getBasePathURI());
                 FileSystem fileSystem = this.getFileSystemCache().getFileSystem(basePath.toUri());
-                return isWritablePath(basePath, fileSystem);
+                
+                // Note: The ivarator config base paths are used by ALL queries which run on the system, so there
+                // should be no harm in creating these directories if they do not already exist by this point.
+                // Also, since we are selecting these directories intentionally for use by the ivarators, it
+                // should be a given that we have write permissions.
+                return fileSystem.exists(basePath) || fileSystem.mkdirs(basePath);
```
ivakegg, line 309: I think the real problem may have been the cleanup of said directories.  We were creating way too many and subsequently if an ivarator directory was to be reused (iterator torn down and rebuilt) then it was unable to find the files from the previous iteration.  We need to find the other ticket that resulted in this change.

lbschanno: @ivakegg Not sure if this is the other ticket you wanted to find, but just in case, work was done in PR #1125 for to address lazily creating ivarator cache dirs to resolve issue #1059.

jwomeara: @ivakegg That's not what the ticket associated with any of the changes indicated.  If there's another one I'm missing, can you point me to it?

ivakegg: I will have to take a look tomorrow.   I am probably remembering incorrectly, but I need to verify.  Sorry for the delay.

jwomeara: That's fine.  if it turns out that there is a legitimate concern there, then I would prefer to take the time to update the code to handle that with multiple ivarator cache dirs available instead of just limiting the query iterator to one.

apmoriarty (approved):

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
ivakegg: For performance reasons, I am ok with reducing this check as you have done.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -286,29 +284,26 @@ public void init(SortedKeyValueIterator<Key,Value> source, Map<String,String> op
...
-        if (ivaratorCacheDirConfigs.isEmpty()) {
-            return;
-        }
-        IvaratorCacheDirConfig validConfig = null;
+        List<IvaratorCacheDirConfig> configsToRemove = new ArrayList<>();
         for (IvaratorCacheDirConfig config : ivaratorCacheDirConfigs) {
-            if (hasValidBasePath(config)) {
-                validConfig = config;
-                break;
+            if (!hasValidBasePath(config)) {
```
ivakegg, line 292: Now that we are sending potentially going to use multiple directories again, please verify that the rebuild logic will work appropriatly if the files are spread across multiple directories.

ivakegg (approved):

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
jwomeara: I refreshed my memory w.r.t. the multiple directory support tonight...

When an ivarator calls `fillSortedSets` it creates an HdfsBackedSortedSet which loops over each of the ivarator cache dirs to determine whether any files exist in any of the ivarator cache dirs.  If a file exists, it creates a FileSortedSet which is marked as already being persisted, and adds that to the list of sets maintained by the ivarator.  

This leads me to believe that we would handle reading from multiple ivarator cache dirs in the event of a teardown/rebuild (given that the same set of ivarator cache dirs should be set for the rebuilt ivarator).  Any sets that have already been written will be loaded, and any new sets that need to be written will be written to a valid ivarator cache dir.  

Does that sound right to you?

Issue #1708, opened 2022-09-29
Regex model expansions


ivakegg:
In issue #1558, we now appropriately push down terms for which the regex normalization cannot be applied to the index.  This has caused some issues when we have a regex against a model field which subsequently expands to a mess of database fields and only a few of those cannot be executed using the regex supplied.  The feature that we would like is as follows:
1) Create a query property marker called "ignored" that can wrap terms and will subsequently be ignored for index lookups and evaluation.
2) When expanding model fields for a regex term, if the regex cannot be normalized for one of the expanded fields which is indexed, then wrap that terms with the ignored query property marker.
3) If it turns out that all of the field regex terms that the model expanded to cannot be normalized as such, then do not wrap the fields with ignored, and instead wrap the entire expansion with the evaluation only query property marker.

PR #1765, opened 2022-11-16
Feature/regex model expansions

ivakegg:
#1708 is the ticket being handled here.  The idea is that when doing model expansions, if a regex cannot be normalized for one of the expanded fields then it can be dropped/ignored instead of being used in the query plan.

A couple things were cleaned up along the way:
1) The JexlASTHelper.getIdentifier was only checking the first child/grandchild within for loops.  Fixed to check em all.
2) Moved the NoExpansionFunctionVisitor functionality into the QueryOptionsFromQueryVisitor along with all of the rest of the functions that are simply used to specifiy query parameters.
3) Removed the deprecated index-only/unevaluated stuff that was encoded in the query model.  Index-only is determined from the datawave metadata.  This was a really old concept pre-rewrite of all of the query logics.
4) Several classes were moved into the metadata utils module to enable better reuse of the key parsing.


apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -369,11 +369,9 @@ public static String getIdentifier(JexlNode node, boolean deconstruct) throws No
                     for (int j = 0; j < child.jjtGetNumChildren(); j++) {
                         JexlNode grandChild = child.jjtGetChild(j);
                         
-                        // If the grandchild and its image is non-null and equal to the any-field identifier
+                        // if the grandChild is an identifier, then return its image/value
                         if (grandChild instanceof ASTIdentifier) {
                             return (deconstruct ? deconstructIdentifier(grandChild.image) : grandChild.image);
-                        } else if (grandChild instanceof ASTFunctionNode) {
-                            return null;
```
jwomeara, line 376: Intellij makes the valid argument that the top level for loop in this method doesn't loop at all.  If the child is a reference, we loop over it's children and either return an identifier or null.  If the child is anything else, we immediately return null.  Is this method doing what we expect?

ivakegg: oh crud... you are right.  I guess this has been working because the identifier is always the first child.  working on this.

ivakegg: Sometimes I wonder how this works at all....wow

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
```diff
@@ -242,36 +253,53 @@ protected JexlNode expandNodeForNormalizers(JexlNode node, Object data) {
...
+                    }
+                    
+                    // determine if we are marking this term as ignored or evaluation only
+                    boolean ignoreExpression = false;
+                    boolean evaluationOnly = false;
+                    if (failedNormalization) {
+                        // if we are not being lenient then add the original term into the mix
+                        if (!lenient) {
+                            normalizedTerms.add(term);
+                            // @TODO we should be able to do the same with all nodes but we need
```
ivakegg: @jwomeara we need to talk this one out.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryModelVisitor.java
```diff
@@ -367,18 +397,14 @@ protected JexlNode expandBinaryNodeFromModel(JexlNode node, Object data) {
             }
             if (log.isTraceEnabled())
                 log.trace("expanded:" + PrintingVisitor.formattedQueryString(expanded));
-            return expanded;
-        } else if (1 == product.size()) {
```
hlgp, line 374: curious why we had this in here in the first place

hlgp:
Nice!  Possible I missed it, but could we have a test for a lenient field with two normalizers: one that passes, one that fails, and ensure the plan is what we expect?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryModelVisitor.java
ivakegg: I am guessing that the JexlnodeFactory.createNodeTreeFromPairs initially did not handle the single case.

warehouse/query-core/src/test/java/datawave/query/LenientFieldsTest.java
```diff
@@ -0,0 +1,280 @@
...
+        }
+        // @formatter:off
+        String[] queryStrings = {
+                "AG > 'abc10'",
+                "AG == '40'",
+                "NAM > '40'",
+                "NAM == 'abc40'",
+        };
+        String[] expectedPlans = {
+                "(((_Drop_ = true) && ((_Reason_ = 'Normalizations failed and lenient') && (_Query_ = 'AGE > \\'abc10\\''))) || ((_Drop_ = true) && ((_Reason_ = 'Normalizations failed and lenient') && (_Query_ = 'ETA > \\'abc10\\''))))",
```
hlgp: I love the _Reason_!

ivakegg: +1

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/NoExpansionFunctionVisitor.java
```diff
@@ -1,98 +0,0 @@
...
-import org.apache.commons.jexl2.parser.JexlNode;
-
-import java.util.HashSet;
-import java.util.Set;
-import java.util.stream.Collectors;
-
-/**
- * Visitor that parses out a NO_EXPANSION function and args from a query tree
- */
-public class NoExpansionFunctionVisitor extends RebuildingVisitor {
```
apmoriarty, line 19: Query docs will need to be updated to reflect that this functionality is only available via query params

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPropertyMarkerVisitor.java
```diff
@@ -90,10 +92,12 @@ public static boolean registerMarker(Class<? extends QueryPropertyMarker> marker
             registerMarker(IndexHoleMarkerJexlNode.class);
             registerMarker(ASTDelayedPredicate.class);
             registerMarker(ASTEvaluationOnly.class);
+            registerMarker(DroppedExpression.class);
             registerMarker(ExceededOrThresholdMarkerJexlNode.class);
             registerMarker(ExceededTermThresholdMarkerJexlNode.class);
             registerMarker(ExceededValueThresholdMarkerJexlNode.class);
             registerMarker(BoundedRange.class);
+            registerMarker(LenientExpression.class);
```
apmoriarty, line 100: Missing Strict?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
```diff
@@ -221,6 +238,10 @@ protected JexlNode expandNodeForNormalizers(JexlNode node, Object data) {
             final String fieldName = op.deconstructIdentifier();
             final Object literal = op.getLiteralValue();
             
+            boolean lenient = !config.getStrictFields().contains(fieldName)
```
apmoriarty: If I'm reading this correctly, a field that is both lenient and strict will default to strict wrt expansion.

warehouse/query-core/src/test/java/datawave/query/config/ShardQueryConfigurationTest.java
```diff
@@ -376,7 +376,7 @@ public void testDeepCopyConstructor() {
         QueryModel expectedQueryModel = new QueryModel();
         Assert.assertEquals(expectedQueryModel.getForwardQueryMapping(), config.getQueryModel().getForwardQueryMapping());
         Assert.assertEquals(expectedQueryModel.getReverseQueryMapping(), config.getQueryModel().getReverseQueryMapping());
-        Assert.assertEquals(expectedQueryModel.getUnevaluatedFields(), config.getQueryModel().getUnevaluatedFields());
+        Assert.assertEquals(expectedQueryModel.getLenientForwardMappings(), config.getQueryModel().getLenientForwardMappings());
```
apmoriarty, line 399: missing strict? or do we assume expansions are strict by default

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/QueryModelVisitorTest.java
```diff
@@ -345,6 +351,41 @@ public void testCastesianProduct() throws ParseException {
...
+    
+    @Test
+    public void testStrictOptionRegexQuery() throws ParseException {
+        String original = "AG =~ '123' && f:strict(AG)";
+        String expected = "AGE =~ '123' || ETA =~ '123'";
+        testWithOptionFunction(original, expected, model, Collections.emptySet(), Collections.emptySet(), Sets.newHashSet("AG"));
+    }
+    
+    @Test
+    public void testStrictPrecidenceRegexQuery() throws ParseException {
```
apmoriarty, line 396: +1

apmoriarty (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
ivakegg: That is correct.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/NoExpansionFunctionVisitor.java
ivakegg: That is not true.  This functionality was moved into the QueryOptionsFromQueryVisitor.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPropertyMarkerVisitor.java
ivakegg: Nope, there is no strict property marker.

warehouse/query-core/src/test/java/datawave/query/config/ShardQueryConfigurationTest.java
ivakegg: The query model only lenient forward mappings.  Strict is presummed.

apmoriarty (dismissed):

hlgp:
I'm curious why you moved ModelKeyParser, FieldMapping, etc. to metadata-utils but not Model

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
```diff
@@ -483,11 +508,15 @@ public Object visit(ASTAndNode node, Object data) {
...
-            return functionalSet;
-        } else {
-            return Boolean.TRUE;
-        }
+        return functionalSet;
+    }
+    
+    @Override
+    public Object visit(ASTAssignment node, Object data) {
+        // evaluate the assignment to get the context updates
```
apmoriarty: This could be a bit clearer, perhaps something like
```
// evaluate the assignment to populate the context
```

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
```diff
@@ -254,41 +255,19 @@ public Object visit(ASTOrNode node, Object data) {
...
-            }
-        } else {
-            rightFunctionalSet = new FunctionalSet();
-            rightFunctionalSet.addAll((Collection) right);
-        }
+    /**
+     * Determine the result of or'ing left with right. null means no result, empty functional set is essentially a false, otherwise true.
+     * 
+     * @param left
+     * @param right
```
apmoriarty: docs!

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
```diff
@@ -483,11 +508,15 @@ public Object visit(ASTAndNode node, Object data) {
...
-            return Boolean.TRUE;
-        }
+        return functionalSet;
+    }
+    
+    @Override
+    public Object visit(ASTAssignment node, Object data) {
+        // evaluate the assignment to get the context updates
+        super.visit(node, data);
+        // return null because we do not want assignments to affect results
```
apmoriarty: Why would an assignment returning null affect the results?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
```diff
@@ -235,71 +284,102 @@ protected JexlNode expandNodeForNormalizers(JexlNode node, Object data) {
...
                                 normalizedTerms.add(term);
+                                normalizedNodes.add(JexlNodeFactory.buildUntypedNode(node, fieldName, term));
                             }
+                            droppedExpression = true;
                         }
                     }
                     
-                    if (normalizedTerms.size() > 1) {
+                    // build an expression from the normalized terms
+                    if (normalizedNodes.size() > 1) {
```
apmoriarty: My preference when creating a node from a variable length list is to do a switch statement on `normalizedNodes.size()` to get a clean logical progression like so
```
case 0:
case 1:
default:
```

jwomeara: Just adding my 2 cents...  I think it's fine as-is.

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/DroppedExpression.java
```diff
@@ -0,0 +1,131 @@
...
+
+import com.google.common.collect.Lists;
+import datawave.query.jexl.JexlNodeFactory;
+import datawave.query.jexl.nodes.QueryPropertyMarker;
+import datawave.query.jexl.visitors.JexlStringBuildingVisitor;
+
+import java.util.function.Function;
+
+/**
+ * Represents a node which captures an expression which will be dropped/ignored. The ignored express will be assigned to the _Query_ variable.
```
jwomeara:
```suggestion
 * Represents a node which captures an expression which will be dropped/ignored. The ignored expression will be assigned to the _Query_ variable.
```

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
```diff
@@ -298,8 +277,48 @@ public Object interpretOr(Object left, Object right) {
         } else if (rightFunctionalSet != null) {
             return rightFunctionalSet;
         } else {
-            return getBooleanOr(left, right);
+            return null;
```
jwomeara: getBooleanOr will return false with two null inputs, but now we're returning null.  What happens if null is returned as the final result of this visitor?  Looking at the isMatched method, it appears that that will log a debug message and return false in that case.  If intentional, then we should consider removing getBooleanOr (and getBooleanAnd) since it isn't used.  We could also clean up the if/else statements as such:

```
        if (leftFunctionalSet != null && rightFunctionalSet != null) { // add left and right
            FunctionalSet functionalSet = new FunctionalSet(leftFunctionalSet);
            functionalSet.addAll(rightFunctionalSet);
            return functionalSet;
        } else if (leftFunctionalSet != null) {
            return leftFunctionalSet;
        } else {
            return rightFunctionalSet;
        }
```

The reasoning being that if we get to the final `else if` and leftFunctionalSet is null, then rightFunctionalSet is either non-null (which we would want to return) or is null (which we also want to return).

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
```diff
@@ -483,11 +508,15 @@ public Object visit(ASTAndNode node, Object data) {
         }
         
         // the expression evaluated to true. either return the functional set of hits, or boolean true
```
jwomeara: Do we want to update this comment?

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/DroppedExpression.java
```diff
@@ -0,0 +1,131 @@
...
+    private static final String CLASS_NAME = DroppedExpression.class.getSimpleName();
+    
+    private static final String LABEL = "_Drop_";
+    
+    private static final String REASON_LABEL = "_Reason_";
+    
+    private static final String QUERY_LABEL = "_Query_";
+    
+    /**
+     * Return the label this marker type: {@value #LABEL}. Overrides {@link QueryPropertyMarker#label()}.
```
jwomeara:
```suggestion
     * Return the label for this marker type: {@value #LABEL}. Overrides {@link QueryPropertyMarker#label()}.
```

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/DroppedExpression.java
```diff
@@ -0,0 +1,131 @@
...
+     *
+     * @return the label
+     */
+    public static String label() {
+        return LABEL;
+    }
+    
+    /**
+     * Create and return a new {@link DroppedExpression}. In this case the source node is turned into a string which is assigned to a variable such that this
+     * expression avoiding its evaluation.
```
jwomeara:
```suggestion
     * expression avoids its evaluation.
```

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/DroppedExpression.java
```diff
@@ -0,0 +1,131 @@
...
+        super();
+    }
+    
+    public DroppedExpression(int id) {
+        super(id);
+    }
+    
+    /**
+     * Create and return a new {@link DroppedExpression}. In this case the source node is turned into a string which is assigned to a variable such that this
+     * expression avoiding its evaluation.
```
jwomeara:
```suggestion
     * expression avoids its evaluation.
```

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/DroppedExpression.java
```diff
@@ -0,0 +1,131 @@
...
+     * @return the new marker node
+     * @see QueryPropertyMarker#create(JexlNode, Function)
+     */
+    public static DroppedExpression create(JexlNode node) {
+        return create(node, DroppedExpression::new);
+    }
+    
+    /**
+     * Returns a new query property marker with the expression <code>(({@value #LABEL} = true) &amp;&amp; ( {@value #QUERY_LABEL} = '{source}' ))</code>. In
+     * this case the source node is turned into a string which is assigned to a variable such that this expression avoiding its evaluation.
```
jwomeara:
```suggestion
     * this case the source node is turned into a string which is assigned to a variable such that this expression avoids its evaluation.
```

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/DroppedExpression.java
```diff
@@ -0,0 +1,131 @@
...
+     * @see QueryPropertyMarker#QueryPropertyMarker(JexlNode) the super constructor for additional information on the tree structure
+     */
+    public DroppedExpression(JexlNode source, String reason) {
+        super(source);
+        addReason(reason);
+    }
+    
+    /**
+     * Returns a new query property marker with the expression <code>(({@value #LABEL} = true) &amp;&amp; ( {@value #QUERY_LABEL} = '{source}' ))</code>. In
+     * this case the source node is turned into a string which is assigned to a variable such that this expression avoiding its evaluation.
```
jwomeara:
```suggestion
     * this case the source node is turned into a string which is assigned to a variable such that this expression avoids its evaluation.
```

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/LenientExpression.java
```diff
@@ -0,0 +1,72 @@
...
+     *
+     * @return the label
+     */
+    public static String label() {
+        return LABEL;
+    }
+    
+    /**
+     * Create and return a new {@link LenientExpression}. In this case the source node is turned into a string which is assigned to a variable such that this
+     * expression avoiding its evaluation.
```
jwomeara:
```suggestion
     * expression avoids its evaluation.
```

web-services/client/src/main/java/datawave/webservice/model/Model.java
jwomeara, line 1: This seems a little half-baked.  We iterate through all of the fields, calling isLenient on each of them to make a list of the lenient fields.  Then we iterate through all of the fields again, only create rows if isLenient returns true for the field, and populate the lenient column of the row by checking the prerecorded list of lenient fields for this field name...

What am i missing?

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
```diff
@@ -284,7 +295,23 @@ public static JexlNode createNodeTreeFromPairs(ContainerType containerType, Jexl
...
+                leftNode = leftType.getSource();
+            }
+            if (rightType.isType(LenientExpression.class)) {
+                lenient = true;
+                rightNode = rightType.getSource();
+            }
+            JexlNode child = buildUntypedBinaryNode(node, leftNode, rightNode);
+            if (lenient) {
+                child = LenientExpression.create(child);
+            }
```
jwomeara, line 314: Can you create or do you have a test that exercises this code path?  I tried QueryModelVisitor but i never got a positive hit for any of these if statements.  I'm wondering if this code is needed.

jwomeara (changes requested):
Just a few comments.  Otherwise good.

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
ivakegg: that would be equivalent.  Agreed.

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
ivakegg: removed

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
ivakegg: fair

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
ivakegg: In other words, an assignment should not affect the boolean evaluation of the expression

ivakegg: I am adding some comments to the top to explain what the various return values mean.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
ivakegg: good catch.  Thank You.  I will work it.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
ivakegg: If it reads better for you, no problem switching it around.

web-services/client/src/main/java/datawave/webservice/model/Model.java
ivakegg: Hmm.... trying to figure out what I was thinking here.....

ivakegg: Ah right.  So to mark a model field as being lenient, we add a field mapping of MODEL_FIELD_NAME -> "lenient".  So the first loop is finding the lenient markers, and the second loop is displaying all the non-lenient markers which are the actual fields, exposing the leniency in a separate column.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
ivakegg: test added

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
```diff
@@ -254,52 +261,68 @@ public Object visit(ASTOrNode node, Object data) {
...
-                throw new RuntimeException(right.toString() + " boolean coercion error", xrt);
-            }
-        } else {
-            rightFunctionalSet = new FunctionalSet();
-            rightFunctionalSet.addAll((Collection) right);
-        }
+    /**
+     * Determine the result of or'ing left with right. null means no result, empty functional set is essentially a false, otherwise true.
+     * 
+     * @param left
```
avgAGB, line 266: to piggyback off the annotations comment, could the parameters be formatted to have a "description" line for each so the mvn warnings don't get tripped? something like:

     * @param left
     *                    left param
     * @param right
     *                   right param

web-services/client/src/main/java/datawave/webservice/model/Model.java
```diff
@@ -130,16 +134,30 @@ public String getMainContent() {
...
         
         for (FieldMapping f : this.getFields()) {
-            builder.append("<td>").append(f.getColumnVisibility()).append("</td>");
-            builder.append("<td>").append(f.getFieldName()).append("</td>");
-            builder.append("<td>").append(f.getDatatype()).append("</td>");
-            builder.append("<td>").append(f.getModelFieldName()).append("</td>");
-            builder.append("<td>").append(f.getDirection()).append("</td>");
-            builder.append("</tr>");
+            // don't include the lenient marker itself
+            if (!f.isLenient()) {
```
jwomeara: is lenient marker perhaps?

jwomeara (changes requested):
minor feedback

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
ivakegg: ok

web-services/client/src/main/java/datawave/webservice/model/Model.java
ivakegg: Awesome suggestion.  thanks.

jwomeara (dismissed):

hlgp (dismissed):

ivakegg:
This has been rebased on top of integration, and I verified all of the code changes are the same.  I will be merging this as I already had the approvals.

jwomeara (approved):

PR #1632, opened 2022-06-24
Task/reduce term frequency search space

apmoriarty:
I'm to redo this as a series of smaller PRs, and then feature flag which iterator is used to collect term offsets.

Reopen of https://github.com/NationalSecurityAgency/datawave/pull/1058, but this time the branch is from the correct repo

PR #1766, opened 2022-11-18
detect and log invalid tf keys

apmoriarty:
Part of https://github.com/NationalSecurityAgency/datawave/pull/1632

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/core/iterators/key/TFKey.java
+++ b/warehouse/query-core/src/main/java/datawave/core/iterators/key/TFKey.java
@@ -38,7 +38,7 @@ public class TFKey {
     }
     
     public String getDatatype() {
-        if (isValid() && datatype == null && splits.size() > 1) {
+        if (datatype == null && isValid()) {
             int stop = splits.get(0);
             datatype = new String(backing, 0, stop);
         }
@@ -46,7 +46,7 @@ public class TFKey {
     }
     
     public String getUid() {
-        if (isValid() && uid == null) {
+        if (uid == null && isValid()) {
             if (splits.size() == 1) {
                 int start = splits.get(0) + 1;
                 uid = new String(backing, start, (backing.length - start));
@@ -60,7 +60,7 @@ public class TFKey {
     }
     
     public String getValue() {
-        if (isValid() && value == null && splits.size() >= 3) {
+        if (value == null && isValid()) {
             int start = splits.get(1) + 1;
             int stop = splits.get(splits.size() - 1);
             value = new String(backing, start, (stop - start));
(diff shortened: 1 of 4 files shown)


warehouse/query-core/src/main/java/datawave/core/iterators/key/TFKey.java
```diff
@@ -18,7 +17,7 @@ public class TFKey {
     private String uidAndValue;
     
     private byte[] backing;
-    private ArrayList<Integer> nulls;
+    private final ArrayList<Integer> nulls = new ArrayList<>();
```
avgAGB: for readability would it be reasonable to change the naming here to nullValues or nullIntegers?

warehouse/query-core/src/main/java/datawave/core/iterators/key/TFKey.java
```diff
@@ -38,11 +37,9 @@ public void parse(Key k) {
     }
     
     public String getDatatype() {
-        if (datatype == null) {
-            if (nulls.size() > 1) {
-                int stop = nulls.get(0);
-                datatype = new String(backing, 0, stop);
-            }
+        if (datatype == null && nulls.size() > 1) {
```
jschmidt10: This isn't a deal breaker but right now, we could have an invalid tf key but still allow `getDatatype()` to return something which may or may not be valid.
What do you think about all of the getXXX() methods instead doing:

```
if (datatype == null && isValid()) 
  // parse out datatype
```
We could do the same for uid and value, etc. That way, if it's invalid, we just return null for everything.

warehouse/query-core/src/main/java/datawave/core/iterators/key/TFKey.java
apmoriarty: `nulls` seems to be the convention when parsing a key, but perhaps `nullIndices` would be more precise....however typing that out I see why the decision was made for `nulls`

warehouse/query-core/src/main/java/datawave/core/iterators/key/TFKey.java
apmoriarty: Simple enough -- done

warehouse/query-core/src/main/java/datawave/core/iterators/key/TFKey.java
apmoriarty: Updated this variable name to `splits` with a comment

avgAGB (dismissed):

avgAGB (approved):

jschmidt10 (approved):

PR #1771, opened 2022-11-21
Response object factory creation and use

ivakegg:
* Updated to set the response object factory via spring for non-CDI instantiated object (i.e. query logics)
* Updated to fix responseObjectFactory usage various classes

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/transformer/EventQueryDataDecorator.java
+++ b/warehouse/query-core/src/main/java/datawave/query/transformer/EventQueryDataDecorator.java
@@ -1,19 +1,14 @@
 package datawave.query.transformer;
 
+import com.google.common.collect.Multimap;
+import datawave.webservice.query.result.event.FieldBase;
+import datawave.webservice.query.result.event.ResponseObjectFactory;
+import org.apache.log4j.Logger;
+
 import java.util.Collection;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.Map;
-
-import javax.inject.Inject;
-
-import datawave.configuration.spring.SpringBean;
-import datawave.webservice.query.result.event.ResponseObjectFactory;
-import datawave.webservice.query.result.event.FieldBase;
-
-import org.apache.log4j.Logger;
-
-import com.google.common.collect.Multimap;
 
 public class EventQueryDataDecorator {
     private String fieldName = null;
--- a/warehouse/query-core/src/main/java/datawave/query/transformer/EventQueryDataDecorator.java
+++ b/warehouse/query-core/src/main/java/datawave/query/transformer/EventQueryDataDecorator.java
@@ -7,6 +7,7 @@ import java.util.Map;
 
 import javax.inject.Inject;
 
+import datawave.configuration.spring.SpringBean;
 import datawave.webservice.query.result.event.ResponseObjectFactory;
 import datawave.webservice.query.result.event.FieldBase;
 
@@ -18,7 +19,6 @@ public class EventQueryDataDecorator {
     private String fieldName = null;
     private Map<String,String> patternMap = new LinkedHashMap<>();
     private Logger log = Logger.getLogger(EventQueryDataDecorator.class);
-    @Inject
     private ResponseObjectFactory responseObjectFactory;
     
     private EventQueryDataDecorator() {}
--- a/warehouse/query-core/src/main/java/datawave/webservice/modification/MutableMetadataHandler.java
+++ b/warehouse/query-core/src/main/java/datawave/webservice/modification/MutableMetadataHandler.java
@@ -930,6 +930,8 @@ public class MutableMetadataHandler extends ModificationServiceConfiguration {
                 }
                 
                 e = eResponse.getEvents().get(0);
+            } else {
+                log.error("Received a " + response.getClass().getSimpleName() + " but expected a subclass of EventQueryResponseBase");
             }
         } catch (Exception ex) {
             log.error(ex);
--- a/warehouse/query-core/src/main/java/datawave/webservice/modification/MutableMetadataHandler.java
+++ b/warehouse/query-core/src/main/java/datawave/webservice/modification/MutableMetadataHandler.java
@@ -22,11 +22,10 @@ import datawave.webservice.modification.ModificationRequestBase.MODE;
 import datawave.webservice.modification.configuration.ModificationServiceConfiguration;
 import datawave.webservice.query.QueryParametersImpl;
 import datawave.webservice.query.QueryPersistence;
-import datawave.webservice.query.result.event.DefaultEvent;
 import datawave.webservice.query.result.event.EventBase;
 import datawave.webservice.query.runner.QueryExecutorBean;
 import datawave.webservice.result.BaseQueryResponse;
-import datawave.webservice.result.DefaultEventQueryResponse;
+import datawave.webservice.result.EventQueryResponseBase;
 import datawave.webservice.result.GenericResponse;
 import org.apache.accumulo.core.client.AccumuloException;
 import org.apache.accumulo.core.client.AccumuloSecurityException;
@@ -901,7 +900,7 @@ public class MutableMetadataHandler extends ModificationServiceConfiguration {
         
         String logicName = "LuceneUUIDEventQuery";
         
-        DefaultEvent e = null;
+        EventBase e = null;
         QueryExecutorBean queryService = this.getQueryService();
         
         String id = null;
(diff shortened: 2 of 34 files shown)


jwomeara:
I haven't looked at this yet but it's the same as what we did in the microwave branch?  I'll look closer tomorrow.

warehouse/query-core/src/main/java/datawave/webservice/modification/MutableMetadataHandler.java
```diff
@@ -921,16 +920,16 @@ columnVisibility, new Date(0), new Date(), StringUtils.join(auths, ','), expirat
...
                 if (eResponse.getEvents().size() > 1) {
                     throw new IllegalStateException("More than one event matched " + uuid + " (" + eResponse.getEvents().size() + " matched)");
                 }
                 if (eResponse.getEvents().isEmpty()) {
                     throw new IllegalStateException("No event matched " + uuid);
                 }
                 
-                e = (DefaultEvent) eResponse.getEvents().get(0);
+                e = eResponse.getEvents().get(0);
             }
```
drewfarris, line 935: Should we throw some exception in the case where the response is not an instance of `EventQueryResponseBase`?

warehouse/query-core/src/main/java/datawave/query/transformer/EventQueryDataDecorator.java
```diff
@@ -7,6 +7,7 @@
 
 import javax.inject.Inject;
 
+import datawave.configuration.spring.SpringBean;
```
drewfarris: This might be unused.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/ExecutableExpansionVisitorTest.java
```diff
@@ -134,14 +136,16 @@ protected void runTestQuery(List<String> expected, String querystr, Date startDa
     
     @Deployment
     public static JavaArchive createDeployment() throws Exception {
+        System.setProperty("cdi.bean.context", "queryBeanRefContext.xml");
```
drewfarris: Not super critical, but I don't know this stuff too well (and maybe other devs won't). Perhaps add a comment explaining why we need to add this property/file? Also, consider adding comments why we add/delete certain packages and classes in the call to `ShrinkWrap` below

web-services/client/src/main/java/datawave/webservice/result/QueryWizardResultResponse.java
```diff
@@ -100,8 +101,8 @@ public String getMainContent() {
         builder.append("<div id=\"myTable_wrapper\" class=\"dataTables_wrapper no-footer\">\n");
         builder.append("<table id=\"myTable\" class=\"dataTable no-footer\" role=\"grid\" aria-describedby=\"myTable_info\">\n");
         
-        if (response instanceof DefaultEventQueryResponse) {
-            DefaultEventQueryResponse tempResponse = (DefaultEventQueryResponse) response;
+        if (response instanceof EdgeQueryResponseBase) {
```
drewfarris: I suspect we want an `instanceof EventQueryResponseBase` here instead of an `instanceof EdgeQueryResponseBase`. Also check the end of this if statement and consider whether we want to throw an exception in the case where we don't know what to do with a specific event response type.

web-services/query/src/main/java/datawave/webservice/query/runner/BasicQueryBean.java
```diff
@@ -100,6 +102,10 @@ public class BasicQueryBean {
     @Resource
     private SessionContext sessionContext;
     
+    @Inject
+    @SpringBean(name = "ResponseObjectFactory")
```
drewfarris, line 106: Do we need `@SpringBean` here or can we follow the pattern established throughout this PR where we just use an `@Inject`?

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -233,6 +233,7 @@ public class QueryExecutorBean implements QueryExecutor {
     private SecurityMarking marking;
     
     @Inject
+    @SpringBean(name = "ResponseObjectFactory")
```
drewfarris, line 236: Is this `@SpringBean` required here?

web-services/security/src/main/java/datawave/security/user/UserOperationsBean.java
```diff
@@ -45,6 +46,7 @@ public class UserOperationsBean {
     private CredentialsCacheBean credentialsCache;
     
     @Inject
+    @SpringBean(name = "ResponseObjectFactory")
```
drewfarris, line 49: Likewise is `@SpringBean` required here as well?

web-services/cached-results/src/main/java/datawave/webservice/results/cached/CachedRunningQuery.java
```diff
@@ -842,7 +842,7 @@ private ResultsPage convert(CachedRowSet cachedRowSet, Integer rowBegin, Integer
             while (cachedRowSet.next() && cachedRowSet.getRow() <= rowEnd && !hitPageByteTrigger) {
                 if (log.isTraceEnabled())
                     log.trace("CRS.position: " + cachedRowSet.getRow() + ", size: " + cachedRowSet.size());
-                CacheableQueryRow row = CacheableQueryRowReader.createRow(cachedRowSet, this.fixedFieldsInEvent);
+                CacheableQueryRow row = CacheableQueryRowReader.createRow(cachedRowSet, this.fixedFieldsInEvent, this.queryLogic.getResponseObjectFactory());
```
drewfarris: Instead of `this.queryLogic.getResponseObjectFactory()` could/should we do `this.responseObjectFactory` instead? (This does bring up a good point, are there places where we should be getting the responseObjectFactory from the query logic instead of injecting it directly?)

drewfarris (changes requested):

warehouse/query-core/src/main/java/datawave/query/transformer/EventQueryDataDecorator.java
ivakegg: got it

warehouse/query-core/src/main/java/datawave/webservice/modification/MutableMetadataHandler.java
ivakegg: That does seem reasonable.  will do

ivakegg: but it appears that all exceptions are caught in this method denoting the event was not found (for whatever reason).  So I think I will leave this as is for now, but I will log something.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/ExecutableExpansionVisitorTest.java
ivakegg: I wish I knew.  I copied this from a different class to get it working.  I will try to get my head around it however.

web-services/cached-results/src/main/java/datawave/webservice/results/cached/CachedRunningQuery.java
ivakegg: ah, right, sure

web-services/client/src/main/java/datawave/webservice/result/QueryWizardResultResponse.java
ivakegg: doh... good catch

web-services/query/src/main/java/datawave/webservice/query/runner/BasicQueryBean.java
ivakegg: yes we need a spring bean here.  This is an EJB that is created via CDI, and the response object factory is a spring created object.

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
ivakegg: yes we need a spring bean here. This is an EJB that is created via CDI, and the response object factory is a spring created object.

web-services/security/src/main/java/datawave/security/user/UserOperationsBean.java
ivakegg: yes we need a spring bean here. This is an EJB that is created via CDI, and the response object factory is a spring created object.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/ExecutableExpansionVisitorTest.java
ivakegg: added what I could figure out.

drewfarris (dismissed):

PR #1774, opened 2022-11-22
WIP: issue-1107: Added sameAs function and unit tests for Uid.List

### LINKED ISSUE #1107: shardIndex UIDs not being updated correctly in the UID.LIST
There is one known issue related to UID counts not being updated correctly:

case 1: uid0 is added twice to an empty UID list: the count becomes 2 when it should be 1
case 2: uid0 is added twice to a maxed out UID list: the count is incremented by 2 when it should be incremented only by 1

In addition to this we have seen evidence where the UIDS in the UID_LIST are not being removed correctly either when delete mutations are added.

The UID.LIST needs to be updated to ensure counts and UID lists are updated correctly in all scenarios.

friedlou:
Fixes #1107

### DIFF
--- a/warehouse/core/src/main/java/datawave/ingest/protobuf/Uid.java
+++ b/warehouse/core/src/main/java/datawave/ingest/protobuf/Uid.java
@@ -3,6 +3,17 @@
 
 package datawave.ingest.protobuf;
 
+import com.google.protobuf.Descriptors;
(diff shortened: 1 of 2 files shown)


ivakegg:
Is this still a WIP ?

friedlou:
Yes

hlgp:
This has gotten stale.  Reopen when ready to revive.

PR #1780, opened 2022-12-05
fixes #1779 RemoteQueryMetricService.map using wrong URL, need to set…

### LINKED ISSUE #1779: RemoteQueryMetricService.map using wrong URL, need to set basemaps on remote response
`private static final String MAP_METRIC_SUFFIX = "id/map/%s";`
should be
`private static final String MAP_METRIC_SUFFIX = "id/%s/map";`

Because basemaps in QueryGeometryResponse is declared @XmlTransient (only used in the HTML response)
`    @XmlTransient
    protected String basemaps = null;` 
basemaps are not returned with the JSON when using RemoteQueryMetricService.  QueryMetricsBean should set basemaps on the remote response.

billoley:
RemoteQueryMetricService.map using wrong URL, need to set basemaps on remote response

### DIFF
--- a/web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsBean.java
+++ b/web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsBean.java
@@ -17,6 +17,7 @@ import datawave.webservice.query.exception.QueryException;
 import datawave.webservice.query.map.QueryGeometryHandler;
 import datawave.webservice.query.map.QueryGeometryResponse;
 import org.apache.commons.lang.time.DateUtils;
+import org.apache.deltaspike.core.api.config.ConfigProperty;
 import org.apache.deltaspike.core.api.exclude.Exclude;
 import org.apache.log4j.Logger;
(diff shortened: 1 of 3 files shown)


web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsBean.java
```diff
@@ -74,6 +75,9 @@ public class QueryMetricsBean {
     @Inject
     private QueryGeometryHandler queryGeometryHandler;
     @Inject
+    @ConfigProperty(name = "dw.basemaps", defaultValue = "{}")
```
keith-ratcliffe, line 78: Do we need to add entries for `dw.basemaps` in `[add|remove]-datawave-configuration.cli` and `default.properties` as we do for others?

billoley: It's already in there since it gets injected into SimpleQueryGeometryHandler

apmoriarty (approved):

keith-ratcliffe (approved):

PR #1784, opened 2022-12-06
Bugfix/issue 1275

friedlou:


warehouse/query-core/src/main/java/datawave/query/common/grouping/GroupingUtil.java
```diff
@@ -172,7 +172,9 @@ private Multimap<String,String> getFieldToFieldWithGroupingContextMap(Document d
                 log.trace("{} contains {}", groupFieldsSet, shorterName);
```
ivakegg: Looking at this, lines 158-159 are equivalent to what was done between 153-154 because shortName and shorterName are initialized to be the same value.  I am wondering if we are missing a statement at line 154 reassigning shorterName to shortName.  I wonder if this oversight was part of the problem.

ivakegg:
Please rebase this on top of integration.  There are way more commits in here than we need.

ivakegg (changes requested):

warehouse/query-core/src/test/java/datawave/query/transformer/GroupingTest.java
```diff
@@ -553,6 +554,49 @@ public void testGroupingWithFieldWithSparseGroupingEntries() throws Exception {
...
+        
+        extraParameters.put("group.fields", "MOTIVE");
+        
+        for (RebuildingScannerTestHelper.TEARDOWN teardown : TEARDOWNS) {
+            for (RebuildingScannerTestHelper.INTERRUPT interrupt : INTERRUPTS) {
+                runTestQueryWithGrouping(expectedMap, queryString, startDate, endDate, extraParameters, teardown, interrupt);
+            }
+        }
+    }
+
```
jwomeara, line 576: missing @Test here?

warehouse/query-core/src/main/java/datawave/query/common/grouping/GroupingUtil.java
jwomeara: Are shortName and shorterName ever different?  Doesn't look like it to me.

warehouse/query-core/src/main/java/datawave/query/common/grouping/GroupingUtil.java
```diff
@@ -185,7 +187,15 @@ private Multimap<String,String> getFieldToFieldWithGroupingContextMap(Document d
...
                     for (Map.Entry<Type<?>,GroupingTypeAttribute<?>> sortedEntry : attrSortedMap.entries()) {
-                        String fieldNameWithContext = fieldName + "." + i++;
+                        String fieldNameWithContext = null;
+                        if (containsContext) {
+                            // multivalued entries from separate entries, but the field names contain the same context number
+                            fieldNameWithContext = shortName + "." + i++;
+                        } else {
+                            // multivalued entries that do not have grouping context
+                            fieldNameWithContext = fieldName + "." + i++;
+                        }
```
jwomeara: If the field contains the context, we use the shortName - Ok.

If the field doesn't contain the context, we use the fieldName.  In that case aren't fieldName, shortName, and shorterName all equal?  Why not simplify this and just use shortName all the time?

Am I missing something?

jwomeara (changes requested):

warehouse/query-core/src/main/java/datawave/query/common/grouping/GroupingUtil.java
hlgp: Yes, they can be.  Cases where fieldname is of the form FIELD_NAME.GROUP.{other info}. GROUPING_CONTEXT.

The problem this is trying to address is for processor-generated fields where we didn't increase the grouping context counter (this should be fixed too!)  It causes a collision in the map so we only ended up returning one instance of the field. 

This class is overdue for an overhaul in any case.  Things could definitely be simplified, but I propose we address this bug first and refactor in a separate ticket.

friedlou: Unit test was added to cover case described by @hlgp.

PR #1785, opened 2022-12-08
Correct matchesInGroup bug where extra groups are returned

hlgp:


mineralntl (dismissed):
LGTM

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GroupingRequiredFilterFunctions.java
```diff
@@ -133,6 +135,17 @@ private static boolean manageGroupsForMatchesInGroupRemainingArgs(Object fieldVa
         }
     }
     
+    // move all these kinds of methods into a central utility once we refactor this
+    private static String getSubgroup(String fieldName) {
+        String[] splits = StringUtils.split(fieldName, '.');
```
ivakegg: So yes, this will do the job.  However a little more efficient might be:
```
int index = fieldName.lastIndexOf('.');
if (index > 0) {
  return fieldName.substring(index+1);
}
return null;
```

ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GroupingRequiredFilterFunctions.java
apmoriarty: +1

warehouse/query-core/src/test/java/datawave/query/TestLimitReturnedGroupsToHitTermGroups.java
```diff
@@ -273,6 +273,22 @@ public void testMultipleGroupsWithRegexAndReturnFields() throws Exception {
         runTestQuery(queryString, format.parse("20091231"), format.parse("20150101"), extraParameters, goodResults);
     }
     
+    @Test
+    public void testGroupWithRegexAndMatchesInGroup() throws Exception {
+        Map<String,String> extraParameters = new HashMap<>();
+        extraParameters.put("include.grouping.context", "true");
+        extraParameters.put("hit.list", "true");
+        extraParameters.put("limit.fields", "CANINE=-1,BIRD=-1");
+        extraParameters.put("return.fields", "CANINE,BIRD");
```
apmoriarty, line 282: For the keys, use the QueryOptions or QueryParameters constants instead of creating new strings

apmoriarty (changes requested):
Had one request, also agree with Ivan's suggestion

apmoriarty (dismissed):

warehouse/query-core/src/test/java/datawave/query/TestLimitReturnedGroupsToHitTermGroups.java
hlgp: Good idea.  Since we do this across a bunch of other test classes, I'll do them all at once under a follow-on MR.

ivakegg (approved):

mineralntl (approved):

apmoriarty (approved):

PR #1787, opened 2022-12-09
sonar lint updates for a number of visitors

apmoriarty:


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EdgeTableRangeBuildingVisitor.java
```diff
@@ -96,14 +95,17 @@ public class EdgeTableRangeBuildingVisitor extends BaseVisitor implements EdgeMo
...
     
-    private static final char BACKSLASH = '\\';
-    private static final char STRING_QUOTE = '\'';
     private Set<String> allowedFunctions = Sets.newHashSet("has_all");
     private int termCount = 0;
     private long maxTerms = 10000;
     private boolean sawEquivalenceRegexSource = false;
     private boolean sawEquivalenceRegexSink = false;
     
+    private static final String AND_ERR_MSG = "And node had unexpected return type";
```
avgAGB: nitpicky, but could this be changed to: "AND node had unexpected return type"

apmoriarty: that's a great catch. error messages should be precise

apmoriarty: Done in https://github.com/NationalSecurityAgency/datawave/pull/1787/commits/6e7a4779796ebc06ec15de3d37e4be52b0fa5cbb

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EdgeTableRangeBuildingVisitor.java
```diff
@@ -456,7 +454,7 @@ private Object visitExpresionNode(SimpleNode node, String operator) {
                         IdentityContext iContext = new IdentityContext(identifier, normalizedLiteral, operator);
                         contexts.add(iContext);
                     } catch (PatternSyntaxException e) {
-                        continue;
+                        // keep going
```
FineAndDandy, line 457: this is a behavior change

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
```diff
@@ -564,12 +566,12 @@ public Object visit(ASTNotNode not, Object data) {
...
-        builder = builder.replaceAll("OrIterator:", "\n\tOrIterator:");
-        builder = builder.replaceAll("Includes:", "\n\t\tIncludes:");
-        builder = builder.replaceAll("Excludes:", "\n\t\tExcludes:");
-        builder = builder.replaceAll("Bridge:", "\n\t\t\tBridge:");
-        return builder.toString();
+        String s = in.toString();
+        s = s.replace("OrIterator:", "\n\tOrIterator:");
+        s = s.replace("Includes:", "\n\t\tIncludes:");
+        s = s.replace("Excludes:", "\n\t\tExcludes:");
+        s = s.replace("Bridge:", "\n\t\t\tBridge:");
```
FineAndDandy, line 573: these are not equivalent

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
```diff
@@ -1309,7 +1308,7 @@ public FunctionFilter(List<ASTFunctionNode> nodes) {
             if (nodes.size() > 1) {
                 ASTAndNode andNode = new ASTAndNode(ParserTreeConstants.JJTANDNODE);
                 children(script, andNode);
-                children(andNode, nodes.toArray(new JexlNode[nodes.size()]));
+                children(andNode, nodes.toArray(new JexlNode[0]));
```
FineAndDandy, line 1311: previously was the last node, now is the first?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushFunctionsIntoExceededValueRanges.java
```diff
@@ -156,7 +154,7 @@ private Set<String> getFunctionFields(JexlNode child) {
         if (child instanceof ASTFunctionNode) {
             return JexlASTHelper.getFieldNames((ASTFunctionNode) child, helper, datatypeFilter);
         }
-        return null;
+        return Collections.emptySet();
```
FineAndDandy, line 157: api change?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPruningVisitor.java
```diff
@@ -573,7 +573,7 @@ private void replaceAndAssign(JexlNode toReplace, String queryString, JexlNode b
                         }
                     }
                     
-                    JexlNodes.children(parent, children.toArray(new JexlNode[children.size()]));
+                    JexlNodes.children(parent, children.toArray(new JexlNode[0]));
```
FineAndDandy, line 613: this is a change in node allocation size

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RegexFunctionVisitor.java
```diff
@@ -66,34 +66,34 @@ public Object visit(ASTFunctionNode node, Object data) {
                 JexlNode regexNode = buildRegexNode(identifier, functionMetadata.name(), arguments.get(1).image);
                 if (regexNode != null) {
                     children.add(regexNode);
+                } else {
+                    // failure to rewrite a regex means a copy of the original node is returned
+                    return returnNode;
```
FineAndDandy: this also is a change in behavior? otherwise is covered in the case 0 below

FineAndDandy (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EdgeTableRangeBuildingVisitor.java
apmoriarty: sonar flags the continue statement as a redundant jump because it's the last statement in the loop. This is safe to remove.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
apmoriarty: I'll add a unit test to prove that these are functionally equivalent. In this case because the first arg isn't an actual regex, it's better to use `String.replace()` [see the sonar explanation for more details](https://rules.sonarsource.com/java/RSPEC-5361)

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
apmoriarty: intellij flags toArray calls with pre-allocated sizes, [baeldung has a good deep dive on the topic](https://www.baeldung.com/java-collection-toarray-methods). 
Sonar also [has an adjacent rule that shows a compliant solution using zero-sized arrays](https://rules.sonarsource.com/java/RSPEC-3020).

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushFunctionsIntoExceededValueRanges.java
apmoriarty: Not really. The only place this is called is one line 82. Returning an empty collection is safer than returning a null
```
                for (String field : getFunctionFields(child)) {
                    functionNodesByField.put(field, child);
                }
```

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPruningVisitor.java
apmoriarty: See previous comment about zero sized toArray calls

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RegexFunctionVisitor.java
apmoriarty: The intent was to make this a fail-fast loop but realistically there will only be a single identifier, so this is covered in the case below. I'll return this to it's original form.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
ivakegg: The function name is misleading, however this does appear to be equivalent.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
ivakegg: I remember delving into the baeldung analysis many years back.  Its a good article.

ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EdgeTableRangeBuildingVisitor.java
```diff
@@ -579,13 +578,12 @@ private VisitationContext computeVisitaionContext(List<QueryContext> queryContex
...
                 }
-                if (firstColumn != null) {
-                    if (!(firstColumn.equals(queryContexts.get(i).getColumnContext()))) {
-                        columnFamilyAreDifferent = true;
-                        break;
-                    }
+                
+                if (columnFamilyAreDifferent) {
+                    break;
                 }
```
jwomeara: Couldn't this be simplified further as:
```
if ((firstColumn != null && (currentContext == null || !(firstColumn.equals(queryContexts.get(i).getColumnContext())))) || (firstColumn == null && currentContext != null)) {
    columnFamilyAreDifferent = true;
    break;
}
```

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RegexFunctionVisitor.java
jwomeara: You can absolutely have multiple fields in a function node - this is very common with geo function queries.  You are also skipping all of the trace statements in the subsequent switch statement by allowing the code to return at this point.  I think you should remove the else statement.  

Also, as a general comment, with this being advertised as a sonar lint PR, I think it's kind of bad practice to mix in functionality changes as well.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RootNegationCheckVisitor.java
```diff
@@ -46,8 +46,7 @@ private boolean hasLeafWithoutNegation(JexlNode node) {
             
             if (ASTNENode.class.equals(childClass) || ASTNotNode.class.equals(childClass) || ASTNRNode.class.equals(childClass)) {
                 // no need to inspect this path anymore
-                continue;
```
jwomeara, line 49: Why are we keeping the preceding if statement around?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RootNegationCheckVisitor.java
```diff
@@ -46,8 +46,7 @@ private boolean hasLeafWithoutNegation(JexlNode node) {
...
-            } else if (child.jjtGetNumChildren() == 0) {
+            } else if (child == null || child.jjtGetNumChildren() == 0) {
```
jwomeara: A null counts as a leaf node?  Is that really true?  Should we not just throw an exception if we encounter a null child?  This is a functionality change.

jwomeara (changes requested):
Looks ok for the most part, but there are some functionality changes mixed in which should not be in this PR.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RegexFunctionVisitor.java
apmoriarty: Strictly speaking this wasn't a functional change -- the original node would still be returned if the number of rebuilt regex nodes did not match the number of identifiers. I did misspeak (mistype?) when I said there would only be one identifier -- there is often more than one identifier.

Regardless, I'll revert this change so we keep the trace logging as part of the execution.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RootNegationCheckVisitor.java
apmoriarty: The continue is implicit as it's the last statement in the loop. We keep the if statement because we do not want to mark a found child, nor do we want to continue recursing down the tree.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RootNegationCheckVisitor.java
apmoriarty: I'm fine with letting this blow up on a null child. Will revert this line back

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RootNegationCheckVisitor.java
jwomeara: What I'm saying is that instead of keeping an empty if statement, why not just do something like this:

```
if (!(ASTNENode.class.equals(childClass) || ASTNotNode.class.equals(childClass) || ASTNRNode.class.equals(childClass)))  {
    if (child == null || child.jjtGetNumChildren() == 0) {
        // reached a leaf node without ever encountering a negation
        foundLeaf = true;
    } else {
        // neither a leaf nor negation, continue looking
        foundLeaf = hasLeafWithoutNegation(child);
    }
}
```

jwomeara: This could be simplified further as:
```
if (!(ASTNENode.class.equals(childClass) || ASTNotNode.class.equals(childClass) || ASTNRNode.class.equals(childClass)))  {
    foundLeaf = (child == null || child.jjtGetNumChildren() == 0) || hasLeafWithoutNegation(child);
}
```

apmoriarty: Oh that's much nicer.

jwomeara: Basically, instead of checking for what we don't want, and doing nothing, let's just check for the condition we do want, and do something.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RootNegationCheckVisitor.java
```diff
@@ -44,14 +44,9 @@ private boolean hasLeafWithoutNegation(JexlNode node) {
...
-                // no need to inspect this path anymore
-            } else if (child == null || child.jjtGetNumChildren() == 0) {
-                // reached a leaf node without ever encountering a negation
-                foundLeaf = true;
-            } else {
-                // neither a leaf nor negation, continue looking
-                foundLeaf = hasLeafWithoutNegation(child);
+            if (!(ASTNENode.class.equals(childClass) || ASTNotNode.class.equals(childClass) || ASTNRNode.class.equals(childClass))) {
+                // either reached a leaf node without encountering a negation, or we need to continue looking
+                foundLeaf = child.jjtGetNumChildren() == 0 || hasLeafWithoutNegation(child);
```
jwomeara, line 49: We don't want to keep the child null check?

apmoriarty: Adding the null check was an unintended functional change. Originally a null node would throw a NPE, whereas with the null check the visitor continues on.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RegexFunctionVisitor.java
```diff
@@ -69,9 +69,8 @@ public Object visit(ASTFunctionNode node, Object data) {
                 }
             }
             
-            // only re-parent if the same number of regex nodes were built
             if (identifiers.size() == children.size()) {
-                switch (identifiers.size()) {
+                switch (children.size()) {
```
jwomeara: So 6 months ago you thought we should switch on `identifiers.size()`, but now you're had a change of heart?  Does this need to be changed?  Also, I would keep the comment.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RootNegationCheckVisitor.java
jwomeara: ok.  i figured that was one of the things sonar lint complained about.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RegexFunctionVisitor.java
apmoriarty: Good catch, that's a failure to fully revert. I suppose it doesn't matter since the collections are the same size, but I'll keep it consistent with what was there before.

jwomeara (approved):

PR #1790, opened 2022-12-12
Add TermFrequencyOffsetIterator and test

apmoriarty:
Adding alternate iterator for collecting term frequency offsets. Not integrated yet.


apmoriarty:
This is just adding code, not integrating it.

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyOffsetIterator.java
```diff
@@ -0,0 +1,164 @@
...
+ */
+public class TermFrequencyOffsetIterator implements SortedKeyValueIterator<Key,Value> {
+    
+    private static final Logger log = Logger.getLogger(TermFrequencyOffsetIterator.class);
+    
+    private Key tk = null;
+    private Value tv = null;
+    
+    // A set of full column qualifiers that describe the offset's exact location (datatype\0uid\0value\0FIELD)
+    private final SortedSet<Text> searchSpace;
```
ivakegg: I am wondering why we need this iterator.  This is essentially equivalent to supplying a set of ranges with the underlying source that include only the searchSpace values.  If we truly need this as as separate iterator, then I would point you to consider extending accumulo's SeekingFilter instead.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyOffsetIterator.java
ivakegg: I added a TermFrequencyFilter offered up as an alternative implementation

apmoriarty: Great work -- I removed the TermFrequencyOffsetIter in the latest commit in favor of the TermOffsetFilter

ivakegg (approved):

avgAGB (approved):

PR #1791, opened 2022-12-13
Add test for KeyToFieldName and integration tests

apmoriarty:
Add test for KeyToFieldName and integration tests FieldNameFilterTest, DocumentKeyToFieldNameTest


warehouse/query-core/src/main/java/datawave/query/function/KeyToFieldName.java
```diff
@@ -24,43 +23,54 @@ public class KeyToFieldName implements Function<Entry<Key,Value>,Entry<Key,Strin
...
-        
-        byte[] arrayReference = sequence.getBackingArray();
-        
-        for (int i = 0; i < sequence.length(); i++) {
-            if (!includeGroupingContext && arrayReference[i] == '.') {
-                index = i;
-                break;
-            }
-            if (arrayReference[i] == 0x00) {
+        for (int i = 0; i < backing.length; i++) {
```
ivakegg: The whole point of using the length from the ByteSequence is that the length of the valid bytes in the backing array may not include the entire length of the array.  I think you need to revert this section of code to what it was previously.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/function/KeyToFieldName.java
apmoriarty: Reverted in https://github.com/NationalSecurityAgency/datawave/pull/1791/commits/0fa094f4c0ee175dfc8e4c4bf44e0e4877c87a57

warehouse/query-core/src/main/java/datawave/query/function/KeyToFieldName.java
```diff
@@ -24,43 +24,56 @@ public class KeyToFieldName implements Function<Entry<Key,Value>,Entry<Key,Strin
...
+     * @return the field name
+     */
     public String getFieldName(Key k) {
         
         int index = -1;
         
-        ByteSequence sequence = k.getColumnQualifierData();
-        
-        byte[] arrayReference = sequence.getBackingArray();
+        ByteSequence bs = k.getColumnQualifierData();
```
jwomeara, line 61: Aside from the javadoc, the changes here seem kind of pedantic...

jwomeara (commented):
The javadoc and tests are welcome, but the minor code changes in KeyToFieldName seem kind of unnecessary to me.

warehouse/query-core/src/main/java/datawave/query/function/KeyToFieldName.java
apmoriarty: When key parsing I tend to favor very short names. Range, Key, Row, ColumnFamily, ColumnQualifier, ByteSequence, etc become r, k, row, cf, cq, bs respectively.

I'm certainly willing to make the change if you have a strong preference to the contrary. Otherwise I'll leave it as-is.

jwomeara: I won't begrudge your preferences - we all have them.  Clearly the original author had a different set of preferences.  However, if you're going to make more PRs like this, then we should probably get the team leads together and come up with a set of agreed upon coding standards.

magrable (approved):

hlgp (approved):

alerman (approved):

ivakegg (approved):

PR #1792, opened 2022-12-14
Task/update datawave key test

apmoriarty:


ivakegg (approved):

apmoriarty:
Only unit test additions

alerman (approved):

warehouse/query-core/src/test/java/datawave/query/data/parsers/DatawaveKeyTest.java
```diff
@@ -1,88 +1,220 @@
...
-        Assert.assertEquals("uid", key.getUid());
-        Assert.assertEquals("fieldValue", key.getFieldValue());
-        Assert.assertEquals("fieldName", key.getFieldName());
+    public void testGlobalIndexKey() {
+        Key key = new Key("fieldValue", "fieldName", "row\0datatype");
+        parse(key);
+        
+        assertInvalid(false);
+        assertType(KeyType.INDEX);
+        assertRow(fieldValue);
```
matthpeterson, line 164: Is the value "row" retrievable from DatawaveKey for global index entries?  It seems odd that both assertRow and assertFieldValue are matching fieldValue.  I know you didn't change the implementation.  Just curious about your thoughts.

matthpeterson (approved):

warehouse/query-core/src/test/java/datawave/query/data/parsers/DatawaveKeyTest.java
apmoriarty: It is *technically* correct, even if it looks odd. 

It looks like the DatawaveKey was originally a intended to flexibly parse field index, event, and term frequency keys with a single object (all key types from the shard table). However, it also works just as well on global index keys and in that case the field value and row are the same.

PR #1795, opened 2022-12-16
QueryPruningVisitor fixes

apmoriarty:
Updated the QueryPruningVisitor to produce trees with no leftover reference expressions. Previously a junction that was pruned to a single field would leave the field wrapped in a reference expression.

Also updated the TreeFlatteningRebuilderTest to exercise both the `flatten` and `flattenAll` paths.

This issue was discovered by running a build with ASTValidation enabled via the test query logic factory.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPruningVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPruningVisitor.java
@@ -233,6 +233,9 @@ public class QueryPruningVisitor extends BaseVisitor {
                 return TruthState.TRUE;
             } else if (state == TruthState.FALSE) {
                 // drop this node from the OR it can never be satisfied
+                // if the child was a junction that was pruned, then the child object is a ref node but the child
+                // at index i is not. reassign for consistency.
+                child = node.jjtGetChild(i);
                 replaceAndAssign(child, null, null);
                 if (debugPrune) {
                     log.debug("Pruning " + originalChild + " from " + originalString);
@@ -272,6 +275,9 @@ public class QueryPruningVisitor extends BaseVisitor {
                 return TruthState.FALSE;
             } else if (state == TruthState.TRUE) {
                 // drop this node from the AND it will always be satisfied
+                // if the child was a junction that was pruned, then the child object is a ref node but the child
+                // at index i is not. reassign for consistency.
+                child = node.jjtGetChild(i);
                 replaceAndAssign(child, null, null);
                 if (debugPrune) {
                     log.debug("Pruning " + originalChild + " from " + originalString);
@@ -313,7 +319,41 @@ public class QueryPruningVisitor extends BaseVisitor {
         if (node.jjtGetNumChildren() != 1) {
             throw new IllegalStateException("ASTReference must only have one child: " + node);
         }
-        return node.jjtGetChild(0).jjtAccept(this, data);
+        Object o = node.jjtGetChild(0).jjtAccept(this, data);
+        
+        pruneJunctionWithSingleChild(node);
+        
+        return o;
+    }
+    
+    /**
+     * When a junction is pruned to a single child also prune that junction and the surrounding reference expression
+     * <p>
+     * <code>parent - ref - refExpr - and/or - single child</code>
+     * <p>
+     * prunes to
+     * <p>
+     * <code>parent - single child</code>
+     *
+     * @param node
+     *            a JexlNode
+     */
+    private void pruneJunctionWithSingleChild(JexlNode node) {
+        boolean isRef = node.getClass().isAssignableFrom(ASTReference.class);
+        if (isRef && node.jjtGetNumChildren() == 1 && node.jjtGetChild(0) instanceof ASTReferenceExpression) {
+            JexlNode refExpr = node.jjtGetChild(0);
+            if (refExpr.jjtGetNumChildren() > 0) {
+                JexlNode junction = refExpr.jjtGetChild(0);
+                if (junction.jjtGetNumChildren() == 1 && (junction instanceof ASTOrNode || junction instanceof ASTAndNode)) {
+                    JexlNode parent = node.jjtGetParent();
+                    JexlNode source = junction.jjtGetChild(0);
+                    JexlNodes.swap(parent, node, source);
+                    
+                    // blank out the junction, refExpr, and ref nodes
+                    JexlNodes.blank(junction, refExpr, node);
+                }
+            }
+        }
     }
     
     @Override
@@ -552,10 +592,8 @@ public class QueryPruningVisitor extends BaseVisitor {
         if (rewrite && toReplace != null) {
             JexlNode parent = toReplace.jjtGetParent();
             if (parent != null && parent != toReplace) {
-                if (queryString != null && log.isDebugEnabled()) {
-                    if (this.debugPrune && baseReplacement != null) {
-                        log.debug("Pruning " + queryString + " to " + (baseReplacement instanceof ASTTrueNode ? "true" : "false"));
-                    }
+                if (queryString != null && log.isDebugEnabled() && this.debugPrune && baseReplacement != null) {
+                    log.debug("Pruning " + queryString + " to " + (baseReplacement instanceof ASTTrueNode ? "true" : "false"));
                 }
                 
                 if (baseReplacement != null) {
@@ -574,6 +612,21 @@ public class QueryPruningVisitor extends BaseVisitor {
                     }
                     
                     JexlNodes.children(parent, children.toArray(new JexlNode[children.size()]));
+                }
+            }
+            
+            // if a junction was completely pruned it is replaced with the last child. In this case, also check
+            // for extra reference expressions and prune those.
+            if ((toReplace instanceof ASTOrNode || toReplace instanceof ASTAndNode) && (parent != null && parent.jjtGetNumChildren() == 1)) {
+                if (parent instanceof ASTReferenceExpression && parent.jjtGetParent() != null) {
+                    JexlNode grandparent = parent.jjtGetParent();
+                    if (grandparent instanceof ASTReference && grandparent.jjtGetParent() != null) {
+                        JexlNode greatGrandParent = grandparent.jjtGetParent();
+                        JexlNodes.swap(greatGrandParent, grandparent, parent.jjtGetChild(0));
+                        
+                        // blank out the parent and grandparent
+                        JexlNodes.blank(parent, grandparent);
+                    }
                 }
             }
         }
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPruningVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPruningVisitor.java
@@ -348,9 +348,6 @@ public class QueryPruningVisitor extends BaseVisitor {
                     JexlNode parent = node.jjtGetParent();
                     JexlNode source = junction.jjtGetChild(0);
                     JexlNodes.swap(parent, node, source);
-                    
-                    // blank out the junction, refExpr, and ref nodes
-                    JexlNodes.blank(junction, refExpr, node);
                 }
             }
         }
@@ -623,9 +620,6 @@ public class QueryPruningVisitor extends BaseVisitor {
                     if (grandparent instanceof ASTReference && grandparent.jjtGetParent() != null) {
                         JexlNode greatGrandParent = grandparent.jjtGetParent();
                         JexlNodes.swap(greatGrandParent, grandparent, parent.jjtGetChild(0));
-                        
-                        // blank out the parent and grandparent
-                        JexlNodes.blank(parent, grandparent);
                     }
                 }
             }
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/TreeFlatteningRebuilder.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/TreeFlatteningRebuilder.java
@@ -97,14 +97,12 @@ public class TreeFlatteningRebuilder {
         Deque<JexlNode> parentStack = new LinkedList<>();
         Deque<List<JexlNode>> childrenStack = new LinkedList<>();
         
-        boolean pruned;
         JexlNode newNode = null;
         
         // now that we have the post order traversal, we can operate on the nodes...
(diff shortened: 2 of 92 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/TreeFlatteningRebuilder.java
```diff
@@ -125,6 +142,15 @@ else if (hasChildren && node == parentStack.peek()) {
...
             if (!postOrderStack.isEmpty()) {
                 
+                if (pruned) {
+                    if (!parentStack.isEmpty() && parentStack.peek() == postOrderStack.peek()) {
+                        childrenStack.peek().add(newNode);
+                    } else if (!postOrderStack.isEmpty()) {
+                        parentStack.push(postOrderStack.peek());
+                        childrenStack.push(Lists.newArrayList(newNode));
+                    }
+                } else
```
jwomeara: Dangling else?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/TreeFlatteningRebuilder.java
```diff
@@ -112,6 +114,21 @@ private JexlNode flattenTree(Deque<JexlNode> postOrderStack) {
...
+                        postOrderStack.peek() instanceof ASTReferenceExpression) {
+                    //  formatter:on
+                    JexlNode t1 = postOrderStack.pop();
+                    if (postOrderStack.peek() instanceof ASTReference) {
+                        postOrderStack.pop();
+                        pruned = true;
+                    } else {
+                        postOrderStack.push(t1);
+                    }
+                }
```
jwomeara: The ability to remove superfluous references and reference expressions already exists, and is toggleable via the `removeReferences` boolean.  See `TreeFlatteningRebuildingVisitor.flattenAll`.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/QueryPruningVisitorTest.java
```diff
@@ -46,18 +49,48 @@ public void delayedGeowaveTest() throws ParseException {
         
         JexlNode rebuilt = QueryPruningVisitor.reduce(script, false);
         
-        Assert.assertEquals(
-                        "geowave:intersects(GEO_FIELD, 'POINT(0 0)') && ((_Delayed_ = true) && (GEO_FIELD == '00' || GEO_FIELD == '0100' || GEO_FIELD == '0103'))",
+        assertEquals("geowave:intersects(GEO_FIELD, 'POINT(0 0)') && ((_Delayed_ = true) && (GEO_FIELD == '00' || GEO_FIELD == '0100' || GEO_FIELD == '0103'))",
```
jwomeara, line 56: If I update the test method to call flattenAll before comparing the actual and expected query strings, all of your new tests pass, except `testGeoQueryProblem`, but that test fails because flattenAll removed _more_ reference expressions than your test expected (which seems desireable based on the changes you're making here).

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/TreeFlatteningRebuildingVisitorTest.java
```diff
@@ -166,19 +169,42 @@ public void multipleNestingMixedOpsTest() throws Exception {
     }
     
     @Test
-    public void singleChildAndOrTest() throws Exception {
-        JexlNode eqNode = JexlASTHelper.parseJexlQuery("FIELD == 'value'");
+    public void singleChildAndOrTest() {
+        JexlNode eqNode = JexlNodeFactory.buildEQNode("FIELD", "value");
```
jwomeara, line 187: This is a better test with this change.  Nice work.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/TreeFlatteningRebuildingVisitorTest.java
```diff
@@ -166,19 +169,42 @@ public void multipleNestingMixedOpsTest() throws Exception {
...
+        JexlNode union = new ASTOrNode(ParserTreeConstants.JJTORNODE);
+        JexlNodes.children(union, eq);
+        
+        JexlNode refExpr = JexlNodes.wrap(union);
+        JexlNode ref = JexlNodes.makeRef(refExpr);
+        ASTJexlScript script = JexlNodeFactory.createScript(ref);
+        
+        JexlNode flattened = TreeFlatteningRebuildingVisitor.flatten(script);
+        
+        Assert.assertEquals("FIELD == 'value'", JexlStringBuildingVisitor.buildQueryWithoutParse(flattened));
```
jwomeara: If I remove your changes to TreeFlatteningRebuilder, then the actual query string in this test becomes `(FIELD == 'value')`.  That is the expected behavior though when you call `flatten`.  If you call `flattenAll`, the reference expression is removed and the test passes.

jwomeara (changes requested):
This is a weird one.  I personally don't think TreeFlatteningRebuilder needs changing at all.  I think we should take a look at QueryPruningVisitor's usage of the TreeFlatteningRebuildingVisitor though. 

We already have the ability to prune excess references and reference expressions using the `flattenAll` method of the TreeFlatteningRebuildingVisitor.  If that functionality is desired in the QueryPruningVisitor, then I would just call `flattenAll` instead of `flatten`.  Or better yet, I would update the QueryPruningVisitor to do proper upkeep when it modifies the tree.  

If the QueryPruningVisitor prunes all but one equals node underneath a reference expression, then the QueryPruningVisitor should prune the reference expression.

A question also - are there actually dangling reference expressions (i.e. a reference expression node with no children) left behind in the tree?  If so, can you add a test specifically for that?  It didn't seem like that was covered here.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/TreeFlatteningRebuilder.java
apmoriarty: The `flattenAll` method has a warning that it may adversely affect the evaluation of a query. 

I redirected the `flatten` method to `flattenAll` and got a bunch of range and geo query failures. After some digging it looks like the `flattenAll` will unwrap the source of a marked bounded range as well as unmarked bounded ranges.

Adding a few lines to the `TreeFlatteningRebuildingVisitorTest` provided the following examples
```
# marked bounded range
((_Bounded_ = true) && (STATE >= 'e' && STATE <= 'r'))
((_Bounded_ = true) && STATE >= 'e' && STATE <= 'r')

# unmarked bounded range (maybe this is fine)
GEO == '1f36c71c71c71c71c7' && (WKT_BYTE_LENGTH >= '+AE0' && WKT_BYTE_LENGTH < '+bE8')
GEO == '1f36c71c71c71c71c7' && WKT_BYTE_LENGTH >= '+AE0' && WKT_BYTE_LENGTH < '+bE8'
```

There's also a few problems with the input queries of the TreeFlatteningRebuildingVisitorTest where the input query structure will change during the initial parse. I think we should have the inputs match the result of `JexlStringBuildingVisitor.buildQueryWithoutParse(JexlASTHelper.parseQuery("query string"))`, let me know what you think.

All that to say, I don't think `flattenAll` should be used to address the specific case of leftover parens after a junction node has been pruned. At least not in it's current state.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/TreeFlatteningRebuilder.java
apmoriarty: Removing this else will cause a bunch of test failures. The execution flow is expecting to loop back around before making any changes to the nodes (from what I could tell).

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/TreeFlatteningRebuilder.java
jwomeara: I'm not advocating for `flattenAll` being used in place of `flatten` in all cases.  In this case though (query pruning), that seems like the functionality you need.  I still question why we aren't updating the QueryPruningVisitor to have it properly prune these nodes instead.  

The comment stating that 'removing reference expressions and references may adversely affect the evaluation of a query' was carried over from the original implementation: (https://github.com/NationalSecurityAgency/datawave/blob/1.1.0/warehouse/query-core/src/main/java/datawave/query/rewrite/jexl/visitors/TreeFlatteningRebuildingVisitor.java#L42)  Maybe that was true at the time, but I don't think that's true today.

My understanding of how the tree flattening rebuilding visitor is supposed to work is that `flatten` leaves parens as-is, while `flattenAll` collapses nested parens, and removes unnecessary parens where applicable, but 1) doesn't affect the evaluation of the tree, and 2) doesn't touch any marked or negated nodes (see `TreeFlatteningRebuilder.flattenReference` to see the evidence of that).  So, by adding logic that conditionally removes parens in the `flatten` call, we would be breaking that contract.

I honestly think we already have the functionality you need in the `flattenAll` call, but I am still advocating for fixing QueryPruningVisitor instead.

apmoriarty: I'm good with backing out the changes in the TreeFlatteningRebuilder and fixing the QueryPruningVisitor to clean up junctions with single nodes leftover from a prune.

Per our discussion I'll look at cleaning up the TreeFlatteningRebuilderTest to exercise both flatten and flattenAll cases.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPruningVisitor.java
```diff
@@ -233,6 +233,9 @@ public Object visit(ASTOrNode node, Object data) {
                 return TruthState.TRUE;
             } else if (state == TruthState.FALSE) {
                 // drop this node from the OR it can never be satisfied
+                // if the child was a junction that was pruned, then the child object is a ref node but the child
+                // at index i is not. reassign for consistency.
+                child = node.jjtGetChild(i);
```
jwomeara, line 238: I'm just starting to look through this, but this caught my eye.  I ran all of the tests in QueryPruningVisitor with a conditional breakpoint here: `child != node.jjtGetChild(i)`

None of the tests hit on this breakpoint.  Is this really needed?

jwomeara: I take that back.  I hit on it - still reviewing.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPruningVisitor.java
```diff
@@ -576,6 +614,21 @@ private void replaceAndAssign(JexlNode toReplace, String queryString, JexlNode b
...
+            // for extra reference expressions and prune those.
+            if ((toReplace instanceof ASTOrNode || toReplace instanceof ASTAndNode) && (parent != null && parent.jjtGetNumChildren() == 1)) {
+                if (parent instanceof ASTReferenceExpression && parent.jjtGetParent() != null) {
+                    JexlNode grandparent = parent.jjtGetParent();
+                    if (grandparent instanceof ASTReference && grandparent.jjtGetParent() != null) {
+                        JexlNode greatGrandParent = grandparent.jjtGetParent();
+                        JexlNodes.swap(greatGrandParent, grandparent, parent.jjtGetChild(0));
+                        
+                        // blank out the parent and grandparent
+                        JexlNodes.blank(parent, grandparent);
```
jwomeara: Is this blanking really necessary?  The blank method creates an extra, unused JexlNode which eventually is just going to be garbage collected anyway.

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/JexlNodes.java
```diff
@@ -175,6 +175,29 @@ public static void setLiteral(ASTStringLiteral literal, String value) {
...
+    
+    /**
+     * Null out parent and child references
+     *
+     * @param node
+     *            a JexlNode
+     */
+    public static void blank(JexlNode node) {
+        node.parent = null;
+        node.children = new JexlNode[0];
```
jwomeara: This seems unnecessary to me, unless there's some functional reason to do this that I'm missing.

jwomeara (changes requested):
The changes to QueryPruningVisitor look good, but I don't think the blank call is necessary.

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/JexlNodes.java
apmoriarty: After some digging I don't think this is necessary. This could be a replaced with a null check if you think eliminating pointers is a good idea, otherwise I'll just remove these methods.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPruningVisitor.java
apmoriarty: The motivation here is to eliminate pointers from nodes that are pruned from the tree. Just because a node is pruned, it may still have references to valid nodes in the resulting query tree, and as the visitor pops back up the call stack, it will visit the pruned nodes. 

In the case that the visitor tries to do any secondary operations on a node after the initial visit, this blanking would prevent any accidental secondary visits.

This is defensive in nature, not because of any known bug.

warehouse/query-core/src/main/java/org/apache/commons/jexl2/parser/JexlNodes.java
jwomeara: Setting to null would be better than creating a new node.  My preference would be to just remove the methods.  It's extra work that we're doing that shouldn't be necessary.

apmoriarty: Sounds good, I'll remove the methods

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/TreeFlatteningRebuildingVisitorTest.java
apmoriarty: Thanks!

jwomeara (approved):

ivakegg (approved):

PR #1800, opened 2022-12-20
1797 - Fix the Blocker level issues found by SonarQube

cogross:
This fixes most of the Blocker level issues found by SonarQube. A few will require more substantial refactoring or redesigning of the code, so those will remain as they are.

Most of the changes involve wrapping calls to connectors/files with a try-resource block and fixing some synchronization blocks.

### DIFF
--- a/warehouse/core/src/main/java/datawave/iterators/filter/ConfigurableAgeOffFilter.java
+++ b/warehouse/core/src/main/java/datawave/iterators/filter/ConfigurableAgeOffFilter.java
@@ -461,28 +461,20 @@ public class ConfigurableAgeOffFilter extends Filter implements OptionDescriber 
                         }
                     }, 1, 10, TimeUnit.SECONDS);
                 }
-            }
-        }
 
-        Path filePath = new Path(filename);
-        if (null == fs) {
-            synchronized (ConfigurableAgeOffFilter.class) {
+                Path filePath = new Path(filename);
                 if (null == fs) {
-                    if (log.isTraceEnabled()) {
-                        log.trace("Setting FileSystem reference");
-                    }
                     fs = filePath.getFileSystem(new Configuration());
+                } else {
+                    if (log.isTraceEnabled()) {
+                        log.trace("Reusing file system reference.");
+                    }
                 }
-            }
-        } else {
-            if (log.isTraceEnabled()) {
-                log.trace("Reusing file system reference.");
+                FileRuleWatcher watcherKey = new FileRuleWatcher(fs, filePath, 1, myEnv);
+
+                copyRules(watcherKey);
             }
         }
-        FileRuleWatcher watcherKey = new FileRuleWatcher(fs, filePath, 1, myEnv);
-
-        copyRules(watcherKey);
-
     }
 
     private long getLongProperty(final String prop, final long defaultValue) {
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/data/TypeRegistry.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/data/TypeRegistry.java
@@ -55,7 +55,7 @@ public class TypeRegistry extends HashMap<String,Type> {
 
     private static final long serialVersionUID = 1L;
 
-    private static TypeRegistry registry = null;
+    private static volatile TypeRegistry registry = null;
 
     private static final Object lock = new Object();
 
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/time/Now.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/time/Now.java
@@ -32,7 +32,7 @@ public class Now {
 
     private AtomicLong now = new AtomicLong();
 
-    private static Now instance;
+    private static volatile Now instance;
 
     public static Now getInstance() {
         if (instance == null) {
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/util/NGramTokenizationStrategy.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/util/NGramTokenizationStrategy.java
@@ -117,10 +117,6 @@ public class NGramTokenizationStrategy extends AbstractNGramTokenizationStrategy
                     fieldValue = null;
                 }
 
-                // Tokenize
-                NGramTokenizer tokenizer = null;
-                TokenStream tokenStream = null;
-                TokenizationException exception = null;
                 try {
                     // Create a reader and determine, if applicable, the max number of allowed n-grams
                     final Reader reader;
@@ -131,48 +127,28 @@ public class NGramTokenizationStrategy extends AbstractNGramTokenizationStrategy
                     }
 
                     // Create the tokenizer and tokenizing stream
-                    tokenizer = new NGramTokenizer(2, maxNGramLength);
-                    tokenizer.setReader(reader);
-                    tokenizer.reset();
-                    tokenStream = new ClassicFilter(tokenizer);
-                    tokenStream = new LowerCaseFilter(tokenStream);
-                    tokenStream.addAttribute(CharTermAttribute.class);
+                    try (NGramTokenizer tokenizer = new NGramTokenizer(2, maxNGramLength)) {
+                        tokenizer.setReader(reader);
+                        tokenizer.reset();
+                        try (TokenStream tokenStream = new LowerCaseFilter(new ClassicFilter(tokenizer))) {
+                            tokenStream.addAttribute(CharTermAttribute.class);
 
-                    // Reset the n-gram count
-                    ngramCount = 0;
+                            // Reset the n-gram count
+                            ngramCount = 0;
 
-                    // Increment the tokenizer and applied any generated n-grams
-                    String ngram = null;
-                    while (null != (ngram = this.increment(tokenizer))) {
-                        if (this.updateFilter(ngram, content)) {
-                            ngramCount++;
-                        }
-                    }
-                } catch (final TokenizationException e) {
-                    exception = e;
-                    throw e;
-                } catch (final IOException e) {
-                    exception = new TokenizationException(e);
-                    throw exception;
-                } finally {
-                    // Close the tokenizer
-                    if (null != tokenizer) {
-                        try {
-                            tokenizer.end();
-                            tokenizer.close();
-                        } catch (final IOException ignore) {}
-                    }
-                    // Close the stream
-                    if (null != tokenStream) {
-                        try {
-                            tokenStream.close();
-                        } catch (final IOException e) {
-                            // Prevent this secondary problem from masking a more serious one
-                            if (null == exception) {
-                                throw new TokenizationException(e);
+                            // Increment the tokenizer and applied any generated n-grams
+                            String ngram = null;
+                            while (null != (ngram = this.increment(tokenizer))) {
+                                if (this.updateFilter(ngram, content)) {
+                                    ngramCount++;
+                                }
                             }
                         }
                     }
+                } catch (final TokenizationException e) {
+                    throw e;
+                } catch (final IOException e) {
+                    throw new TokenizationException(e);
                 }
             }
         }
--- a/warehouse/query-core/src/main/java/datawave/query/index/lookup/ShardRangeStream.java
+++ b/warehouse/query-core/src/main/java/datawave/query/index/lookup/ShardRangeStream.java
@@ -96,7 +96,6 @@ public class ShardRangeStream extends RangeStream {
                 this.context = StreamContext.ABSENT;
 
             }
-
         } catch (TableNotFoundException | DatawaveQueryException e) {
             throw new RuntimeException(e);
         } finally {
--- a/warehouse/query-core/src/main/java/datawave/query/index/stats/IndexStatsClient.java
+++ b/warehouse/query-core/src/main/java/datawave/query/index/stats/IndexStatsClient.java
@@ -109,24 +109,23 @@ public class IndexStatsClient {
             Authorizations auths = client.securityOperations().getUserAuthorizations(client.whoami());
             if (fields.isEmpty()) {
                 scanner = client.createScanner(table, auths);
+                return doScanResults(scanner, dataTypes, dates);
             } else {
-                BatchScanner bScanner = client.createBatchScanner(table, auths, fields.size());
-                bScanner.setRanges(buildRanges(fields));
-                scanner = bScanner;
+                try (BatchScanner bScanner = client.createBatchScanner(table, auths, fields.size())) {
+                    bScanner.setRanges(buildRanges(fields));
+                    scanner = bScanner;
+                    return doScanResults(scanner, dataTypes, dates);
+                }
             }
         } catch (Exception e) {
             log.error(e);
             throw new IOException(e);
         }
+    }
 
+    private Map<String,Double> doScanResults(ScannerBase scanner, Set<String> dataTypes, SortedSet<String> dates) throws IOException {
         configureScanIterators(scanner, dataTypes, dates);
-
         Map<String,Double> results = scanResults(scanner);
-
-        if (scanner instanceof BatchScanner) {
-            scanner.close();
-        }
-
         return results;
     }
 
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ShardIndexQueryTableStaticMethods.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ShardIndexQueryTableStaticMethods.java
@@ -453,17 +453,17 @@ public class ShardIndexQueryTableStaticMethods {
 
         bs.setRanges(ranges);
 
-        SessionOptions options = new SessionOptions();
-
-        IteratorSetting setting = configureDateRangeIterator(config);
-        options.addScanIterator(setting);
-
-        setting = configureGlobalIndexTermMatchingIterator(config, literals, patterns, reverseIndex, limitToUniqueTerms);
-        if (setting != null) {
+        try (SessionOptions options = new SessionOptions()) {
+            IteratorSetting setting = configureDateRangeIterator(config);
             options.addScanIterator(setting);
-        }
 
-        bs.setOptions(options);
+            setting = configureGlobalIndexTermMatchingIterator(config, literals, patterns, reverseIndex, limitToUniqueTerms);
+            if (setting != null) {
+                options.addScanIterator(setting);
+            }
+
+            bs.setOptions(options);
+        }
 
         return bs;
     }
@@ -481,18 +481,19 @@ public class ShardIndexQueryTableStaticMethods {
 
         bs.setRanges(ranges);
 
-        SessionOptions options = new SessionOptions();
-        options.addScanIterator(configureDateRangeIterator(config));
-        IteratorSetting setting = configureGlobalIndexDataTypeFilter(config, config.getDatatypeFilter());
-        if (setting != null) {
-            options.addScanIterator(setting);
-        }
-        setting = configureGlobalIndexTermMatchingIterator(config, literals, patterns, reverseIndex, limitToUniqueTerms);
-        if (setting != null) {
-            options.addScanIterator(setting);
-        }
+        try (SessionOptions options = new SessionOptions()) {
+            options.addScanIterator(configureDateRangeIterator(config));
+            IteratorSetting setting = configureGlobalIndexDataTypeFilter(config, config.getDatatypeFilter());
+            if (setting != null) {
+                options.addScanIterator(setting);
+            }
+            setting = configureGlobalIndexTermMatchingIterator(config, literals, patterns, reverseIndex, limitToUniqueTerms);
+            if (setting != null) {
+                options.addScanIterator(setting);
+            }
 
-        bs.setOptions(options);
+            bs.setOptions(options);
+        }
 
         return bs;
     }
--- a/warehouse/query-core/src/main/java/datawave/query/scheduler/PushdownFunction.java
+++ b/warehouse/query-core/src/main/java/datawave/query/scheduler/PushdownFunction.java
@@ -100,28 +100,28 @@ public class PushdownFunction implements Function<QueryData,List<ScannerChunk>> 
                     queryPlanSet.add(hashCode);
                     try {
 
-                        SessionOptions options = new SessionOptions();
+                        try (SessionOptions options = new SessionOptions()) {
 
-                        if (log.isTraceEnabled()) {
-                            log.trace("setting ranges" + plan.getRanges());
-                            log.trace("range set size" + plan.getSettings().size());
+                            if (log.isTraceEnabled()) {
+                                log.trace("setting ranges" + plan.getRanges());
+                                log.trace("range set size" + plan.getSettings().size());
+                            }
+                            for (IteratorSetting setting : plan.getSettings()) {
+                                options.addScanIterator(setting);
+                            }
+
+                            for (IteratorSetting setting : customSettings) {
+                                options.addScanIterator(setting);
+                            }
+
+                            for (String cf : plan.getColumnFamilies()) {
+                                options.fetchColumnFamily(new Text(cf));
+                            }
+
+                            options.setQueryConfig(this.config);
+
+                            chunks.add(new ScannerChunk(options, Lists.newArrayList(plan.getRanges()), server));
                         }
-                        for (IteratorSetting setting : plan.getSettings()) {
-                            options.addScanIterator(setting);
-                        }
-
-                        for (IteratorSetting setting : customSettings) {
-                            options.addScanIterator(setting);
-                        }
-
-                        for (String cf : plan.getColumnFamilies()) {
-                            options.fetchColumnFamily(new Text(cf));
-                        }
-
-                        options.setQueryConfig(this.config);
-
-                        chunks.add(new ScannerChunk(options, Lists.newArrayList(plan.getRanges()), server));
-
                     } catch (Exception e) {
                         log.error(e);
                         throw new AccumuloException(e);
--- a/warehouse/query-core/src/main/java/datawave/query/statsd/QueryStatsDClient.java
+++ b/warehouse/query-core/src/main/java/datawave/query/statsd/QueryStatsDClient.java
@@ -41,8 +41,9 @@ public class QueryStatsDClient extends ConvenienceMethodProvidingStatsDClient {
     private static final StatsDClientErrorHandler NO_OP_HANDLER = e -> { /* No-op */};
 
     // the client
-    private static NonBlockingUdpSender client = null;
+    private static volatile NonBlockingUdpSender client = null;
     // this monitor controls when the client is being used
+
     private static final Object clientMonitor = new Object();
 
     public QueryStatsDClient(String queryId, String host, int port, int maxCacheSize) {
--- a/warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
+++ b/warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
@@ -85,7 +85,7 @@ public class ActiveQueryLog {
      *            the associated name by which to look up the desired {@link ActiveQueryLog}. This will typically be the name of a table or query logic.
      * @return the existing or new {@link ActiveQueryLog} for the name
      */
-    public static ActiveQueryLog getInstance(String name) {
+    public static synchronized ActiveQueryLog getInstance(String name) {
         // Return the default instance if the name is blank.
         if (StringUtils.isBlank(name)) {
             name = DEFAULT_NAME;
--- a/web-services/security/src/main/java/datawave/security/cache/AccumuloCacheStore.java
+++ b/web-services/security/src/main/java/datawave/security/cache/AccumuloCacheStore.java
@@ -207,15 +207,16 @@ public class AccumuloCacheStore<K extends Serializable,V> implements AdvancedLoa
             scanner = accumuloClient.createScanner(tableName, authorizations);
             byte[] keyBytes = ctx.getMarshaller().objectToByteBuffer(key);
             scanner.setRange(new Range(new Text(keyBytes)));
+
+            Iterator<Map.Entry<Key,Value>> iterator = scanner.iterator();
+            Map.Entry<Key,Value> entry = iterator.hasNext() ? iterator.next() : null;
+            return decodeEntry(entry, key, loadValue, loadMetadata);
+
         } catch (TableNotFoundException e) {
             throw new PersistenceException(e);
         } catch (IOException | InterruptedException e) {
             throw new PersistenceException("Unable to serialize key " + key, e);
         }
-
-        Iterator<Map.Entry<Key,Value>> iterator = scanner.iterator();
-        Map.Entry<Key,Value> entry = iterator.hasNext() ? iterator.next() : null;
-        return decodeEntry(entry, key, loadValue, loadMetadata);
     }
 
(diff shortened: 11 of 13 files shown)


warehouse/core/src/main/java/datawave/iterators/filter/ConfigurableAgeOffFilter.java
```diff
@@ -327,56 +327,47 @@ public void init(SortedKeyValueIterator<Key,Value> source, Map<String,String> op
...
-     * This method instantiates the the necessary implementations of the {@code Filter} interface, as they are defined in the configuration file specified by
+     * This method instantiates the necessary implementations of the {@code Filter} interface, as they are defined in the configuration file specified by
      * {@code this.filename}.
      *
      * @throws IllegalArgumentException
      *             if there is an error in the configuration file
      * @throws IOException
      */
-    private void initFilterRules() throws IllegalArgumentException, IOException {
+    private synchronized void initFilterRules() throws IllegalArgumentException, IOException {
```
keith-ratcliffe: This changes the existing synchronization from class-level scope to instance-level. Class scope is still needed here. Maybe the method can be made static?

keith-ratcliffe (dismissed):

warehouse/core/src/main/java/datawave/iterators/filter/ConfigurableAgeOffFilter.java
cogross: Changing it to `private static synchronized ...` requires the `getLongProperty()` to also be static, which cascades a lot more changes.
I moved the `synchronized` block back down into the code and reworked it to remove it from the blocker level problems.

warehouse/core/src/main/java/datawave/iterators/filter/ConfigurableAgeOffFilter.java
```diff
@@ -336,46 +336,39 @@ public void init(SortedKeyValueIterator<Key,Value> source, Map<String,String> op
...
+                                        log.error(t, t);
+                                    }
+                                }, 1000, 10 * 1000);
             }
         }
         
         Path filePath = new Path(filename);
         if (null == fs) {
             synchronized (ConfigurableAgeOffFilter.class) {
-                if (null == fs) {
```
jwomeara, line 470: Shouldn't the null check be performed within the synchronized block?  If 10 threads see that fs == null before the synchronized block, then all 10 threads will assign a value to fs.  If you perform the null check within the synchronized block, you guarantee that only one thread will set the value.

warehouse/assemble/webservice/src/main/java/datawave/webservice/mr/input/SecureEventSequenceFileInputFormat.java
```diff
@@ -1,24 +0,0 @@
...
-
-import java.io.IOException;
-
-import datawave.ingest.data.RawRecordContainer;
-import datawave.ingest.input.reader.event.EventSequenceFileInputFormat;
-import org.apache.hadoop.mapreduce.InputSplit;
-import org.apache.hadoop.mapreduce.RecordReader;
-import org.apache.hadoop.mapreduce.TaskAttemptContext;
-
-public class SecureEventSequenceFileInputFormat<K> extends EventSequenceFileInputFormat<K> {
```
jwomeara: Just guessing here, but I doubt we actually want to remove this.  We might want to add a SuppressWarnings annotation instead so that SonarQube ignores this.  

e.g. https://stackoverflow.com/a/29400824

warehouse/core/src/main/java/datawave/iterators/filter/ConfigurableAgeOffFilter.java
```diff
@@ -336,46 +336,39 @@ public void init(SortedKeyValueIterator<Key,Value> source, Map<String,String> op
...
-                                                log.info("Changing " + EXPIRATION_INTERVAL_MS_PROP + " to " + expiration);
-                                                EXPIRATION_INTERVAL_MS = expiration;
-                                                ruleCache = CacheBuilder.newBuilder().refreshAfterWrite(UPDATE_INTERVAL_MS, TimeUnit.MILLISECONDS)
-                                                                .expireAfterAccess(EXPIRATION_INTERVAL_MS, TimeUnit.MILLISECONDS)
-                                                                .build(new ReloadableCacheBuilder());
-                                            }
-                                        } catch (Throwable t) {
-                                            log.error(t, t);
+        synchronized (ConfigurableAgeOffFilter.class) {
+            if (null == ruleCache) {
```
jwomeara: Null check within the synchronized block is good.

warehouse/core/src/main/java/datawave/util/AccumuloCounterSource.java
```diff
@@ -62,8 +62,7 @@ public String getNextIdentifier() {
     public boolean hasNext() {
         
         if (null == iterator) {
-            try {
-                BatchScanner scanner = connector.createBatchScanner(queryTable, connector.securityOperations().getUserAuthorizations(username), 100);
+            try (BatchScanner scanner = connector.createBatchScanner(queryTable, connector.securityOperations().getUserAuthorizations(username), 100)) {
```
jwomeara: Try-with-resources will close the batch scanner after we exit the try block, which is not functionally equivalent to the existing code.  If we run with this, then the iterator that is set at the end of the try block will be defunct when we call it since the underlying scanner has been closed.  I would suppress this warning as well, or handle it a different way.

warehouse/ingest-core/src/main/java/datawave/ingest/data/TypeRegistry.java
```diff
@@ -60,13 +60,9 @@ public class TypeRegistry extends HashMap<String,Type> {
...
         if (null == registry) {
-            synchronized (lock) {
-                if (null == registry) {
-                    registry = new TypeRegistry(config);
-                }
-            }
+            registry = new TypeRegistry(config);
         }
         return registry;
     }
```
jwomeara: It would probably be more efficient to leave this code as-is.  In the original code, if the registry is not null, then access to the registry resource will never be blocked since the synchronize call is skipped.  With these changes, only one thread will be able to access the registry resource at a time, blocking all other threads.

warehouse/core/src/main/java/datawave/iterators/filter/ConfigurableAgeOffFilter.java
jwomeara: Actually, it might make sense to keep both null checks.  

In the original code, if the fs is not null, then access to the fs resource will never be blocked since the synchronize call is skipped. With these changes, only one thread will be able to access the fs resource at a time, blocking all other threads.

warehouse/ingest-core/src/main/java/datawave/ingest/time/Now.java
```diff
@@ -34,13 +34,9 @@ public void run() {
...
         if (instance == null) {
-            synchronized (Now.class) {
-                if (instance == null) {
-                    instance = new Now(false);
-                }
-            }
+            instance = new Now(false);
         }
         return instance;
     }
```
jwomeara: Same feedback here as the TypeRegistry.

warehouse/ingest-core/src/main/java/datawave/ingest/util/NGramTokenizationStrategy.java
```diff
@@ -130,48 +126,30 @@ public int tokenize(final NormalizedContentInterface content, int maxNGramLength
...
-                    exception = e;
                     throw e;
                 } catch (final IOException e) {
-                    exception = new TokenizationException(e);
-                    throw exception;
-                } finally {
-                    // Close the tokenizer
-                    if (null != tokenizer) {
-                        try {
-                            tokenizer.end();
```
jwomeara, line 161: It appears that tokenizer.end() is no longer called with this change.  We might want to consider how best to handle that.  Could we add a call to the end method to the close method?

warehouse/query-core/src/main/java/datawave/mr/bulk/RecordIterator.java
```diff
@@ -530,7 +530,7 @@ protected synchronized void fail(Throwable t) {
             // do nothing, but clean up resources anyway.
         }
         try {
-            Thread.sleep(failureSleep);
+            t.wait(failureSleep);
```
jwomeara: If we call wait on the RecordIterator object, this will allow other threads to enter the fail method while we wait.  Calling wait on the throwable won't do that.  Calling Thread.sleep doesn't do that either.  Are we sure this is the behavior we want?

warehouse/query-core/src/main/java/datawave/query/index/lookup/ShardRangeStream.java
```diff
@@ -44,59 +44,59 @@ public CloseableIterable<QueryPlan> streamPlans(JexlNode node) {
...
-                        case UNINDEXED_FIELD:
-                            this.context = StreamContext.UNINDEXED;
-                            break;
-                        case UNKNOWN:
-                            this.context = StreamContext.ABSENT;
-                    }
-                } else {
-                    itr = Iterators.transform(peeking, new FieldIndexParser(node));
-                    this.context = StreamContext.PRESENT;
+            try (BatchScanner scanner = scanners.newScanner(config.getShardTableName(), config.getAuthorizations(), config.getNumQueryThreads(),
```
jwomeara: Similar to my previous comment, if we use try-with-resources then the scanner will be closed, invalidating the iterator which is used outside this method.  I don't think we want to do this.

warehouse/query-core/src/main/java/datawave/query/index/stats/IndexStatsClient.java
```diff
@@ -106,9 +106,10 @@ public Map<String,Double> getStat(Set<String> fields, Set<String> dataTypes, Sor
             if (fields.isEmpty()) {
                 scanner = con.createScanner(table, auths);
             } else {
-                BatchScanner bScanner = con.createBatchScanner(table, auths, fields.size());
-                bScanner.setRanges(buildRanges(fields));
-                scanner = bScanner;
+                try (BatchScanner bScanner = con.createBatchScanner(table, auths, fields.size())) {
```
jwomeara: try-with-resources will close the scanner at the end of the try block, but in the code, the scanner is used outside the try block.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ShardIndexQueryTableStaticMethods.java
```diff
@@ -381,18 +381,18 @@ public static ScannerSession configureTermMatchOnly(ShardQueryConfiguration conf
...
         bs.setRanges(ranges);
         
-        SessionOptions options = new SessionOptions();
-        
-        IteratorSetting setting = configureDateRangeIterator(config);
-        options.addScanIterator(setting);
-        
-        setting = configureGlobalIndexTermMatchingIterator(config, literals, patterns, reverseIndex, limitToUniqueTerms);
-        if (setting != null) {
+        try (SessionOptions options = new SessionOptions()) {
```
jwomeara, line 456: Interesting that SessionOptions.close is a noop.  I guess there's no harm in closing it even though it doesn't do anything.

warehouse/query-core/src/main/java/datawave/query/planner/ThreadedRangeBundlerIterator.java
```diff
@@ -139,7 +139,7 @@ public boolean hasNext() {
                     // amount of time to wait has elapsed OR we have processed all of our ranges before continuing
                     while (this.rangeQueue.size() < numRangesToBuffer && this.rangeQueue.remainingCapacity() > 0
                                     && (startTimeMillis + rangeBufferTimeoutMillis) > System.currentTimeMillis() && !rangeConsumer.isStopped()) {
-                        Thread.sleep(rangeBufferPollMillis);
```
jwomeara: wait on the producerLock releases the lock for this thread and allows other threads to enter the synchronized block.  Is that really what we want?

warehouse/query-core/src/main/java/datawave/query/scheduler/PushdownFunction.java
```diff
@@ -97,28 +97,28 @@ public List<ScannerChunk> apply(QueryData qd) {
...
                         
-                        SessionOptions options = new SessionOptions();
-                        
-                        if (log.isTraceEnabled()) {
-                            log.trace("setting ranges" + plan.getRanges());
-                            log.trace("range set size" + plan.getSettings().size());
-                        }
-                        for (IteratorSetting setting : plan.getSettings()) {
-                            options.addScanIterator(setting);
+                        try (SessionOptions options = new SessionOptions()) {
```
jwomeara, line 103: same comment about SessionOptions.close being a noop.

warehouse/query-core/src/main/java/datawave/query/statsd/QueryStatsDClient.java
```diff
@@ -58,19 +58,15 @@ public QueryStatsDClient(String queryId, String host, int port, int maxCacheSize
     /**
      * Get the client, creating it if needed. This should only be called when the clientMonitor has been acquired.
      */
-    private NonBlockingUdpSender client() {
+    private synchronized NonBlockingUdpSender client() {
         if (this.client == null) {
-            synchronized (clientMonitor) {
-                if (this.client == null) {
```
jwomeara: Same comment as above about null checks before and after the synchronized call potentially blocking less.

warehouse/query-core/src/main/java/datawave/query/tables/shard/IndexStatsQueryLogic.java
```diff
@@ -156,10 +156,11 @@ public List<FieldStat> getStat(Set<String> fields, Set<String> dataTypes, Sorted
                     dataSource = simpleScanner;
                     scanner = simpleScanner;
                 } else {
-                    BatchScanner bScanner = ScannerHelper.createBatchScanner(con, table, auths, fields.size());
-                    bScanner.setRanges(buildRanges(fields));
-                    scanner = bScanner;
-                    dataSource = bScanner;
+                    try (BatchScanner bScanner = ScannerHelper.createBatchScanner(con, table, auths, fields.size())) {
```
jwomeara: This is another case where try-with-resources would close the batch scanner before we're done with it.

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
```diff
@@ -84,19 +84,15 @@ public static ActiveQueryLog getInstance() {
...
-    public static ActiveQueryLog getInstance(String name) {
+    public static synchronized ActiveQueryLog getInstance(String name) {
         // Return the default instance if the name is blank.
         if (StringUtils.isBlank(name)) {
             name = DEFAULT_NAME;
         }
         
         // Initialize the log cache if necessary.
         if (ActiveQueryLog.logCache == null) {
-            synchronized (ActiveQueryLog.logCacheLock) {
```
jwomeara: Same comment about null check before and after synchronized potentially blocking less.

web-services/cached-results/src/main/java/datawave/webservice/results/cached/CachedResultsBean.java
```diff
@@ -1548,10 +1548,11 @@ public BaseQueryResponse previous(@PathParam("queryId") @Required("queryId") Str
                 synchronized (crq) {
                     if (crq.isActivated() == false) {
                         if (crq.getShouldAutoActivate()) {
-                            Connection connection = ds.getConnection();
-                            String logicName = crq.getQueryLogicName();
-                            QueryLogic<?> queryLogic = queryFactory.getQueryLogic(logicName, p);
-                            crq.activate(connection, queryLogic);
+                            try (Connection connection = ds.getConnection()) {
```
jwomeara: I think this is another case where we would be closing the object before we're done using it.

web-services/cached-results/src/main/java/datawave/webservice/results/cached/CachedResultsBean.java
```diff
@@ -1662,11 +1663,11 @@ public CachedResultsResponse reset(@PathParam("queryId") @Required("queryId") St
                         crq.closeConnection(log);
                     }
                     
-                    Connection connection = ds.getConnection();
-                    String logicName = crq.getQueryLogicName();
-                    QueryLogic<?> queryLogic = queryFactory.getQueryLogic(logicName, p);
-                    crq.activate(connection, queryLogic);
-                    
+                    try (Connection connection = ds.getConnection()) {
```
jwomeara: I think this is another case where we would be closing the object before we're done using it.

web-services/cached-results/src/main/java/datawave/webservice/results/cached/CachedResultsBean.java
```diff
@@ -1759,10 +1760,11 @@ public BaseQueryResponse next(@PathParam("queryId") @Required("queryId") String
                 synchronized (crq) {
                     if (crq.isActivated() == false) {
                         if (crq.getShouldAutoActivate()) {
-                            Connection connection = ds.getConnection();
-                            String logicName = crq.getQueryLogicName();
-                            QueryLogic<?> queryLogic = queryFactory.getQueryLogic(logicName, p);
-                            crq.activate(connection, queryLogic);
+                            try (Connection connection = ds.getConnection()) {
```
jwomeara: I think this is another case where we would be closing the object before we're done using it.

web-services/cached-results/src/main/java/datawave/webservice/results/cached/CachedResultsBean.java
```diff
@@ -1902,10 +1904,11 @@ public BaseQueryResponse getRows(@PathParam("queryId") @Required("queryId") Stri
                 synchronized (crq) {
                     
                     if (crq.isActivated() == false) {
-                        Connection connection = ds.getConnection();
-                        String logicName = crq.getQueryLogicName();
-                        QueryLogic<?> queryLogic = queryFactory.getQueryLogic(logicName, p);
-                        crq.activate(connection, queryLogic);
+                        try (Connection connection = ds.getConnection()) {
```
jwomeara: I think this is another case where we would be closing the object before we're done using it.

web-services/map-reduce/src/main/java/datawave/webservice/mr/MapReduceBean.java
```diff
@@ -752,7 +752,6 @@ public StreamingOutput getResultFile(@PathParam("jobId") String jobId, @PathPara
         final FileSystem fs = getFS(hdfs, response);
```
jwomeara: try-with-resources for fs?

web-services/security/src/main/java/datawave/security/cache/AccumuloCacheStore.java
```diff
@@ -207,20 +207,19 @@ public MarshalledEntry<K,V> load(Object key) {
     }
     
     public MarshalledEntry<K,V> _load(Object key, boolean loadValue, boolean loadMetadata) {
-        Scanner scanner;
-        try {
-            scanner = connector.createScanner(tableName, authorizations);
+        try (Scanner scanner = connector.createScanner(tableName, authorizations)) {
```
jwomeara: LGTM!

jwomeara (dismissed):

warehouse/assemble/webservice/src/main/java/datawave/webservice/mr/input/SecureEventSequenceFileInputFormat.java
cogross: It wasn't being used -- but I can restore it.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ShardIndexQueryTableStaticMethods.java
cogross: If you trace down the extended/implemented classes, `ScannerBase` extends `AutoCloseable` so it wants to see a `close()` call somewhere.

warehouse/query-core/src/main/java/datawave/query/scheduler/PushdownFunction.java
cogross: `SessionOptions` extends `AutoCloseable`

web-services/cached-results/src/main/java/datawave/webservice/results/cached/CachedResultsBean.java
cogross: reverted

web-services/cached-results/src/main/java/datawave/webservice/results/cached/CachedResultsBean.java
cogross: reverted

web-services/cached-results/src/main/java/datawave/webservice/results/cached/CachedResultsBean.java
cogross: reverted

web-services/cached-results/src/main/java/datawave/webservice/results/cached/CachedResultsBean.java
cogross: reverted

warehouse/query-core/src/main/java/datawave/query/tables/shard/IndexStatsQueryLogic.java
cogross: refactored this -- think it should be better.

warehouse/query-core/src/main/java/datawave/query/index/stats/IndexStatsClient.java
cogross: refactored this - should be better.

warehouse/query-core/src/main/java/datawave/query/index/lookup/ShardRangeStream.java
cogross: reverted

warehouse/ingest-core/src/main/java/datawave/ingest/util/NGramTokenizationStrategy.java
cogross: Ah.  Added the `end()` call back.

warehouse/core/src/main/java/datawave/util/AccumuloCounterSource.java
cogross: Reverted.

warehouse/core/src/main/java/datawave/iterators/filter/ConfigurableAgeOffFilter.java
cogross: moved this block into the `synchronized` block above - since this should only be happening once on load.

warehouse/ingest-core/src/main/java/datawave/ingest/data/TypeRegistry.java
cogross: Did some more reading on this.  Seems like we can just mark the resource `volatile` and it should behave appropriately with the existing code. (Same with other spots.)

warehouse/query-core/src/main/java/datawave/mr/bulk/RecordIterator.java
cogross: Changed the wait(), should be better.

warehouse/query-core/src/main/java/datawave/query/planner/ThreadedRangeBundlerIterator.java
cogross: Yeah -- not entirely sure.  Changed it back to `sleep()`.

web-services/map-reduce/src/main/java/datawave/webservice/mr/MapReduceBean.java
cogross: it really complicates what the catching will do -- I'm gonna leave it as is for now.

cogross:
Pushed up a new version, up to date w/ latest in integration and fixes for the things mentioned here.

warehouse/core/src/main/java/datawave/iterators/filter/ConfigurableAgeOffFilter.java
jschmidt10: I thought we liked keeping one null check outside of the synchronized block to eliminate majority of the lock contention. As it stands now, we contend the lock on every invocation. With a check both inside and outside, we only contend the lock if the instance isn't set.

Will SonarQube be happy if we mark ruleCache as volatile?

warehouse/ingest-core/src/main/java/datawave/ingest/util/NGramTokenizationStrategy.java
```diff
@@ -131,48 +127,33 @@ public int tokenize(final NormalizedContentInterface content, int maxNGramLength
...
+
+                                // Increment the tokenizer and applied any generated n-grams
+                                String ngram = null;
+                                while (null != (ngram = this.increment(tokenizer))) {
+                                    if (this.updateFilter(ngram, content)) {
+                                        ngramCount++;
+                                    }
+                                }
+                            }
+                            if (null != tokenizer) {
```
jschmidt10: Since we called `tokenizer.setReader` and `tokenizer.reset` above, it looks like this null check is not needed.

warehouse/ingest-core/src/main/java/datawave/ingest/util/NGramTokenizationStrategy.java
```diff
@@ -131,48 +127,33 @@ public int tokenize(final NormalizedContentInterface content, int maxNGramLength
...
-
-                    // Increment the tokenizer and applied any generated n-grams
-                    String ngram = null;
-                    while (null != (ngram = this.increment(tokenizer))) {
-                        if (this.updateFilter(ngram, content)) {
-                            ngramCount++;
+                    try (NGramTokenizer tokenizer = new NGramTokenizer(2, maxNGramLength)) {
+                        tokenizer.setReader(reader);
+                        tokenizer.reset();
+                        try (TokenStream tokenStream = new ClassicFilter(tokenizer)) {
```
jschmidt10: We're only using this tokenStream to form the next one. Thoughts on removing one layer of nesting like this:

```
try (TokenStream tokenStream = new LowerCaseFilter(new ClassicFilter(tokenizer))) {
  // ...
  ```

File: warehouse/query-core/src/main/java/datawave/query/planner/ThreadedRangeBundlerIterator.java (hunk 13 above)
Comment (jschmidt10) [reply warehouse/query-core/src/main/java/datawave/query/planner/ThreadedRangeBundlerIterator.java]: The diff is not reflecting this was changed back to the `sleep()` call. There's also a similar change above in `RecordIterator` so if we're not sure the wait() is appropriate, might be worth flipping that back for now too. We can always dig in with a more specific task for those later.

File: web-services/map-reduce/src/main/java/datawave/webservice/mr/MapReduceBean.java [hunk 26]
```diff
@@ -757,64 +747,58 @@ public StreamingOutput getResultFile(@PathParam("jobId") String jobId, @PathPara
...
                                 MessageFormat.format("{0} at path {1}", fileName, resultsDir));
                 response.addException(qe);
                 throw new NotFoundException(qe, response);
             }
 
-            fis = fs.open(resultFile);
+            try (FSDataInputStream fis = fs.open(resultFile)) {
+                // Make a final reference to the fis for referencing inside the inner class
+                final FSDataInputStream fiz = fis;
+                return new StreamingOutput() {
```
jschmidt10: This looks like it won't work to me. We're returning a StreamingOutput instance where we closed the `fis` (via the try-with-resources above).

When the caller attempts to use `streamingOutput.write(someOutput)`, that'll produce some kind of IOException due to the underlying input stream being closed.

warehouse/ingest-core/src/main/java/datawave/ingest/util/NGramTokenizationStrategy.java
cogross: true - removed.

warehouse/ingest-core/src/main/java/datawave/ingest/util/NGramTokenizationStrategy.java
cogross: done!

web-services/map-reduce/src/main/java/datawave/webservice/mr/MapReduceBean.java
cogross: gotcha.  sonarqube is overzealous about closing thing out.  reverted.

warehouse/core/src/main/java/datawave/iterators/filter/ConfigurableAgeOffFilter.java
ivakegg: I agree with @jschmidt10 's comment here and I do not see a response.  Synchronization can be expensive and it would be good to avoid it if possible.  I think the check outside and inside the synchronization block is desired here.

ivakegg (dismissed):
Looks like some comments still need to be addressed here.

warehouse/core/src/main/java/datawave/iterators/filter/ConfigurableAgeOffFilter.java
cogross: Not sure how I missed this message.  Been trying to get sonarqube up today - current issue is the docker image pull limit.  Hopefully tomorrow I can run this early and get a scan going.
I can put the first null check back - unsure if the volatile will make a difference, but we can just ignore its guidance and do the double null check.

cogross: OK - making it volatile doesn't help -- https://wiki.sei.cmu.edu/confluence/display/java/CON50-J.+Do+not+assume+that+declaring+a+reference+volatile+guarantees+safe+publication+of+the+members+of+the+referenced+object
I'll leave the double-check in place.

warehouse/query-core/src/main/java/datawave/query/tracking/ActiveQueryLog.java
cogross: put it back how it was.

warehouse/query-core/src/main/java/datawave/query/tables/shard/IndexStatsQueryLogic.java
cogross: reverted.

warehouse/query-core/src/main/java/datawave/query/planner/ThreadedRangeBundlerIterator.java
cogross: reverted

cogross:
Fixed the last few things -- a few comments were hidden.

warehouse/ingest-core/src/main/java/datawave/ingest/util/NGramTokenizationStrategy.java
cogross: So with the 'try-with-resources' it will automatically call close().  Since this object is just being made here -- what does the end() call do?
I can just revert the whole block if it is really needed.

MiguelRicardos (approved):
Looks good to me. Can you squash and rebase?

avgAGB (approved):

MiguelRicardos (approved):

hlgp:
This appears to have broken internal tests

PR #1803, opened 2022-12-30
Correct StreamContext within ConcurrentScannerInitializer

apmoriarty:


warehouse/query-core/src/main/java/datawave/query/index/lookup/ConcurrentScannerInitializer.java
```diff
@@ -35,6 +35,10 @@ public ConcurrentScannerInitializer(BaseIndexStream stream) {
     public BaseIndexStream call() throws Exception {
         if (stream.context() == StreamContext.INITIALIZED) {
             if (stream.hasNext()) {
+                if (stream instanceof ScannerStream) {
```
ivakegg: a comment as to what is going on here please.

apmoriarty: Added in https://github.com/NationalSecurityAgency/datawave/pull/1803/commits/5656aa2a8c6ec0f910ff6e83b3b55f043c798462

warehouse/query-core/src/test/java/datawave/query/index/lookup/RangeStreamTest.java
```diff
@@ -1579,7 +1578,7 @@ public void testIntersection_ofDayRangesAndShardRange() throws Exception {
         
         config.setQueryFieldsDatatypes(dataTypes);
         config.setIndexedFields(dataTypes);
-        config.setShardsPerDayThreshold(2);
+        config.setShardsPerDayThreshold(0); // set to zero to roll up to day ranges
```
ivakegg, line 1581: How is changing this from 2 to 0 actually testing anything new?

warehouse/query-core/src/main/java/datawave/query/index/lookup/ConcurrentScannerInitializer.java
```diff
@@ -35,6 +35,12 @@ public ConcurrentScannerInitializer(BaseIndexStream stream) {
...
+                    // the Intersection and Union classes are predicated on the accuracy of the StreamContext,
+                    // so update this stream's context from INITIALIZED to PRESENT
+                    stream.context = StreamContext.PRESENT;
+                    return stream;
+                }
                 return ScannerStream.withData(stream, stream.currentNode());
```
ivakegg: Does not return ScannerStream.withData(stream, stream.currentNode()) already return a stream with a context of PRESENT?

warehouse/query-core/src/test/java/datawave/query/index/lookup/RangeStreamTest.java
apmoriarty: Now that we actually do more seeking, we need to lower the shards per day threshold to hit the rollup.

warehouse/query-core/src/main/java/datawave/query/index/lookup/ConcurrentScannerInitializer.java
apmoriarty: The case addressed here is
```
RangeStream line 503 -> visit(EQ) -> return ScannerStream.initialized(scannerSession, entryParser, node);
ScannerStream line 109 -> initial context is INITIALIZED
```
There cannot be a double initialization of the scanner stream.

avgAGB (approved):

ivakegg (approved):

PR #1804, opened 2022-12-30
ContentFunctionEvaluator now short-circuits

apmoriarty:
- updated several docs
- added short-circuit logic for when the search space is empty

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionEvaluator.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionEvaluator.java
@@ -5,7 +5,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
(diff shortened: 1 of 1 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionEvaluator.java
```diff
@@ -100,17 +99,15 @@ protected boolean initialize() {
                 log.trace("Failing within() because of bad arguments");
                 log.trace(distance + " " + ((termOffsetMap == null) ? "null" : termOffsetMap.toString()) + " " + Arrays.toString(terms));
             }
-            
             return false;
         }
         
-        // generate an intersection of event ids that cover all of the terms
+        // generate an intersection of event ids that cover all the terms
```
jschmidt10, line 105: Not a huge deal but the `of` here and below I think are technically correct.

jschmidt10 (approved):

avgAGB (approved):

PR #1808, opened 2023-01-10
1798 - Fix the Ciritcal Issues Found by SonarQube

cogross:
Fixed items that SonarQube marked as "Critical."

Examples:
 - add equals/hashCode functions (if you have one you need both)
 -  add some null and /0 checks
 -  fixed assertNotEquals calls
 -  check isPresent() before using Optional
 -  static volatile int -> AtomicInteger for some counters
 -  refactor BulkIngestMapFileLoader - catch excess `catch` statements and overall simplification of test case generation
 -  remove() instead of clear() for ThreadLocal objects
 -  refactor repeated code/checks to variables/functions
 -  `Random` objects should be instantiated once and reused throughout class

### DIFF
--- a/warehouse/core/src/main/java/datawave/ingest/protobuf/TermWeightPosition.java
+++ b/warehouse/core/src/main/java/datawave/ingest/protobuf/TermWeightPosition.java
@@ -2,6 +2,7 @@ package datawave.ingest.protobuf;
 
 import java.util.Comparator;
 
+import org.apache.commons.lang.builder.HashCodeBuilder;
 import org.apache.log4j.Logger;
 
 /**
@@ -96,6 +97,11 @@ public class TermWeightPosition implements Comparable<TermWeightPosition> {
     public boolean equals(TermWeightPosition o) {
 
         return (compareTo(o) == 0);
+    }
+
+    @Override
+    public int hashCode() {
+        return new HashCodeBuilder().append(offset).append(prevSkips).append(score).append(zeroOffsetMatch).toHashCode();
     }
 
     /**
--- a/warehouse/core/src/main/java/datawave/ingest/util/cache/Loader.java
+++ b/warehouse/core/src/main/java/datawave/ingest/util/cache/Loader.java
@@ -5,6 +5,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.concurrent.Executors;
 
 import org.apache.log4j.Logger;
@@ -164,8 +165,18 @@ public abstract class Loader<K,V> extends CacheLoader<K,V> implements Runnable {
 
     }
 
+    public boolean equals(Object o) {
+        if (this == o)
+            return true;
+        if (o == null || getClass() != o.getClass())
+            return false;
+        Loader<?,?> loader = (Loader<?,?>) o;
+        return childHash == loader.childHash && lazy == loader.lazy && Objects.equals(entryCache, loader.entryCache)
+                        && Objects.equals(executor, loader.executor) && Objects.equals(children, loader.children);
+    }
+
     @Override
     public int hashCode() {
-        return System.identityHashCode(this) + childHash;
+        return Objects.hash(entryCache, executor, children, childHash, lazy);
     }
 }
--- a/warehouse/core/src/main/java/datawave/iterators/filter/ageoff/DataTypeAgeOffFilter.java
+++ b/warehouse/core/src/main/java/datawave/iterators/filter/ageoff/DataTypeAgeOffFilter.java
@@ -295,12 +295,16 @@ public class DataTypeAgeOffFilter extends AppliedRule {
 
                 final String dataTypeHasScanTime = options.getOption(dataType + ".hasScanTime");
                 if (Boolean.parseBoolean(dataTypeHasScanTime)) {
-                    final String scanTime = iterEnv.getConfig().get("table.custom.timestamp.current." + dataType);
-                    try {
-                        dataTypeScanTimes.put(dataType, Long.parseLong(scanTime, 10));
-                    } catch (final NumberFormatException e) {
-                        throw new NumberFormatException(dataType + " marked as hasScanTime but corresponding table.custom.timestamp.current." + dataType
-                                        + " is invalid: " + scanTime);
+                    if (iterEnv != null) {
+                        final String scanTime = iterEnv.getConfig().get("table.custom.timestamp.current." + dataType);
+                        try {
+                            dataTypeScanTimes.put(dataType, Long.parseLong(scanTime, 10));
+                        } catch (final NumberFormatException e) {
+                            throw new NumberFormatException(dataType + " marked as hasScanTime but corresponding table.custom.timestamp.current." + dataType
+                                            + " is invalid: " + scanTime);
+                        }
+                    } else {
+                        throw new NullPointerException("IteratorEnvironment is null");
                     }
                 }
             }
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/CompositeIngest.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/CompositeIngest.java
@@ -69,8 +69,9 @@ public interface CompositeIngest {
     }
 
     static boolean isOverloadedCompositeField(Collection<String> compFields, String compositeFieldName) {
-        if (compFields != null && !compFields.isEmpty())
+        if (compFields != null && !compFields.isEmpty()) {
             return compFields.stream().findFirst().get().equals(compositeFieldName);
+        }
         return false;
     }
 
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/data/tokenize/TokenSearchSynonymFilter.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/data/tokenize/TokenSearchSynonymFilter.java
@@ -2,9 +2,11 @@ package datawave.ingest.data.tokenize;
 
 import java.io.IOException;
 import java.util.Collection;
+import java.util.Objects;
 import java.util.PriorityQueue;
 import java.util.Queue;
 
+import org.apache.commons.lang.builder.HashCodeBuilder;
 import org.apache.lucene.analysis.TokenFilter;
 import org.apache.lucene.analysis.TokenStream;
 import org.apache.lucene.analysis.tokenattributes.CharTermAttribute;
@@ -12,6 +14,8 @@ import org.apache.lucene.analysis.tokenattributes.OffsetAttribute;
 import org.apache.lucene.analysis.tokenattributes.PositionIncrementAttribute;
 import org.apache.lucene.analysis.tokenattributes.PositionLengthAttribute;
 import org.apache.lucene.analysis.tokenattributes.TypeAttribute;
+
+import it.unimi.dsi.fastutil.Hash;
 
 /** Wraps DefaultTokenSearch and exposes it as a Lucene TokenFilter. Not thread-safe */
 public class TokenSearchSynonymFilter extends TokenFilter {
@@ -218,6 +222,11 @@ public class TokenSearchSynonymFilter extends TokenFilter {
             }
 
             return false;
+        }
+
+        @Override
+        public int hashCode() {
+            return new HashCodeBuilder().append(term).append(position).append(offset).append(length).hashCode();
         }
 
         @Override
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/tokenize/ContentIndexingColumnBasedHandler.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/tokenize/ContentIndexingColumnBasedHandler.java
@@ -376,9 +376,10 @@ public abstract class ContentIndexingColumnBasedHandler<KEYIN> extends AbstractC
             tokenizerTimeWarned = false;
 
             while (true) {
-                if (heartBeatCount != HeartBeatThread.counter) {
-                    tokenizerBeats += HeartBeatThread.counter - heartBeatCount;
-                    heartBeatCount = HeartBeatThread.counter;
+                int currentHeartBeatCount = HeartBeatThread.counter;
+                if (heartBeatCount != currentHeartBeatCount) {
+                    tokenizerBeats += currentHeartBeatCount - heartBeatCount;
+                    heartBeatCount = currentHeartBeatCount;
 
                     // warn once on exceeding the warn threshold
                     long elapsedEstimateMsec = tokenizerBeats * HeartBeatThread.INTERVAL;
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
@@ -457,7 +457,7 @@ public class IngestJob implements Tool {
                     numRecords += recordC.getValue();
                 }
                 // records that throw runtime exceptions are still counted as processed
-                float percentError = 100 * ((float) numExceptions / numRecords);
+                float percentError = (numRecords == 0) ? 0.0f : 100 * ((float) numExceptions / numRecords);
                 log.info(String.format("Percent Error: %.2f", percentError));
                 if (conf.getInt("job.percent.error.threshold", 101) <= percentError) {
                     return jobFailed(job, runningJob, outputFs, workDirPath);
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/util/WeightedValuePruningStrategy.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/util/WeightedValuePruningStrategy.java
@@ -135,7 +135,8 @@ public class WeightedValuePruningStrategy extends NGramTokenizationStrategy {
                 if (ratioOfTotalNgramCountsToMaxAllowed > 1.0f) {
                     reducedScaleNgramCount = Math.round(((float) reducedScaleNgramCount) * ratioOfTotalNgramCountsToMaxAllowed);
                 } else {
-                    reducedScaleNgramCount = Math.round(((float) reducedScaleNgramCount) / ratioOfTotalNgramCountsToMaxAllowed);
+                    reducedScaleNgramCount = (ratioOfTotalNgramCountsToMaxAllowed == 0.0) ? 0
+                                    : Math.round(((float) reducedScaleNgramCount) / ratioOfTotalNgramCountsToMaxAllowed);
                 }
 
                 totalNgramCounts += reducedScaleNgramCount;
--- a/warehouse/core/src/main/java/datawave/iterators/filter/ConfigurableAgeOffFilter.java
+++ b/warehouse/core/src/main/java/datawave/iterators/filter/ConfigurableAgeOffFilter.java
@@ -114,7 +114,7 @@ public class ConfigurableAgeOffFilter extends Filter implements OptionDescriber 
     protected static long UPDATE_INTERVAL_MS = DEFAULT_UPDATE_INTERVAL_MS;
 
     public static final String EXPIRATION_INTERVAL_MS_PROP = "tserver.datawave.ageoff.cache.expiration.interval.ms";
-    protected static final long DEFAULT_EXPIRATION_INTERVAL_MS = 60 * 60 * 1000; // default 1 hour
+    protected static final long DEFAULT_EXPIRATION_INTERVAL_MS = 60 * 60 * 1000L; // default 1 hour
     protected static long EXPIRATION_INTERVAL_MS = DEFAULT_EXPIRATION_INTERVAL_MS;
 
     /**
--- a/warehouse/core/src/main/java/datawave/iterators/filter/TokenSpecParser.java
+++ b/warehouse/core/src/main/java/datawave/iterators/filter/TokenSpecParser.java
@@ -142,7 +142,9 @@ public abstract class TokenSpecParser<B extends TokenSpecParser> {
          */
         protected String expect(ParseTokenType type) {
             ParseToken next = peek();
-            if (next == null || next.type != type) {
+            if (next == null) {
+                throw error("Found null for " + type, -1);
+            } else if (next.type != type) {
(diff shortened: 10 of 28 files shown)


warehouse/core/src/main/java/datawave/ingest/protobuf/TermWeightPosition.java
```diff
@@ -98,6 +100,11 @@ public boolean equals(TermWeightPosition o) {
         return (compareTo(o) == 0);
     }
     
+    @Override
+    public int hashCode() {
+        return System.identityHashCode(this);
```
ivakegg: This kinda makes the hash code not useful.
```suggestion
        return new HashCodeBuilder().append(offset).append(prevSkips).append(score).append(zeroOffsetMatch).toHashCode();
```

warehouse/core/src/main/java/datawave/ingest/util/cache/Loader.java
```diff
@@ -165,6 +165,10 @@ public void run() {
         
     }
     
+    public boolean equals(Loader loader) {
+        return (this.hashCode() == loader.hashCode());
+    }
+
```
ivakegg: This is not the equals method you are looking for.  Also if we are to actually match the intent of the hashCode method below, then this needs to actually be the same object.
```suggestion
    @Override
    public boolean equals(Object loader) {
        return this == loader;
    }
    
```

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/CompositeIngest.java
```diff
@@ -68,7 +68,9 @@ static boolean isOverloadedCompositeField(Multimap<String,String> compositeField
     
     static boolean isOverloadedCompositeField(Collection<String> compFields, String compositeFieldName) {
         if (compFields != null && !compFields.isEmpty())
-            return compFields.stream().findFirst().get().equals(compositeFieldName);
+            if (compFields.stream().findFirst().isPresent()) {
+                return compFields.stream().findFirst().get().equals(compositeFieldName);
+            }
```
ivakegg: maybe a little more efficient:
```suggestion
            String first = compFields.iterator().next();
            return first != null && first.equals(compositeFieldName);
```

warehouse/ingest-core/src/main/java/datawave/ingest/data/tokenize/TokenSearchSynonymFilter.java
```diff
@@ -220,6 +220,11 @@ public boolean equals(Object thatObject) {
             return false;
         }
         
+        @Override
+        public int hashCode() {
+            return System.identityHashCode(this);
```
ivakegg: Use HashBuilder()

warehouse/ingest-core/src/main/java/datawave/ingest/data/tokenize/TokenizationHelper.java
```diff
@@ -24,7 +25,7 @@ public static class HeartBeatThread extends Thread {
         private static final Logger log = Logger.getLogger(HeartBeatThread.class);
         
         public static final long INTERVAL = 500; // half second resolution
-        public static volatile int counter = 0;
+        private static AtomicInteger counter = new AtomicInteger(0);
```
ivakegg: Why change this to an object?  Seems overkill.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/EventMapper.java
```diff
@@ -699,7 +699,7 @@ public void processEvent(K1 key, RawRecordContainer value, List<DataTypeHandler<
             // Try to only parse the event once. Parse the event on the first pass and only if
             // the BaseIngestHelper class differs. The same class used by different handlers
             // *should* produce the same result.
-            if (null == previousHelper || !previousHelper.getClass().getName().equals(thisHelper.getClass().getName())) {
+            if (null == previousHelper || !thisHelper.getClass().isAssignableFrom(previousHelper.getClass())) {
```
ivakegg: This is NOT what we want here.  We want to determine if this is the same class.  Instead we can change this with previousHelper.getClass().equals(thisHelper.getClass()).

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/tokenize/ExtendedContentIndexingColumnBasedHandler.java
```diff
@@ -654,7 +655,7 @@ public static void closeOutputStreams(OutputStream... streams) {
      */
     protected static class HeartBeatThread extends Thread {
         public static final long INTERVAL = 500; // half second resolution
-        public static volatile int counter = 0;
+        public AtomicInteger counter = new AtomicInteger(0);
```
ivakegg: Why do this?

warehouse/ingest-core/src/main/java/datawave/ingest/util/WeightedValuePruningStrategy.java
```diff
@@ -134,7 +134,7 @@ private int calculateWeightings(final Integer[] originalStringLengths) {
                 int reducedScaleNgramCount = BloomFilterUtil.predictNGramCount((float) scalableOriginalStringLengths[i], this.maxNGramLength);
                 if (ratioOfTotalNgramCountsToMaxAllowed > 1.0f) {
                     reducedScaleNgramCount = Math.round(((float) reducedScaleNgramCount) * ratioOfTotalNgramCountsToMaxAllowed);
-                } else {
+                } else if (ratioOfTotalNgramCountsToMaxAllowed > 0.0f) {
```
ivakegg: Perhaps this should be != 0.0f, and then we should consider what to do if it is 0.0f.

web-services/map-reduce/src/main/java/datawave/webservice/mr/configuration/MapReduceJobConfiguration.java
```diff
@@ -414,7 +414,7 @@ private void addArchivedDirectory(String source, Pattern pattern, Path classpath
         }
         
         int fileCount = 0;
-        String outputName = source.replaceAll("/", "_").replaceFirst("$_", "") + ".jar";
+        String outputName = source.replaceAll("/", "_") + ".jar";
```
ivakegg: I think we still want to remove the '_' if it is the first character.

ivakegg (changes requested):

warehouse/core/src/main/java/datawave/ingest/protobuf/TermWeightPosition.java
```diff
@@ -1,8 +1,10 @@
 package datawave.ingest.protobuf;
 
+import org.apache.commons.lang.builder.HashCodeBuilder;
 import org.apache.log4j.Logger;
 
 import java.util.Comparator;
+import java.util.Objects;
```
jwomeara: HashCodeBuilder and Objects are unused.

warehouse/core/src/main/java/datawave/ingest/protobuf/TermWeightPosition.java
jwomeara: +1

warehouse/core/src/main/java/datawave/ingest/util/cache/Loader.java
```diff
@@ -165,6 +165,10 @@ public void run() {
         
     }
     
+    public boolean equals(Loader loader) {
+        return (this.hashCode() == loader.hashCode());
+    }
```
jwomeara: why not a more formal equals method, like this?
```suggestion
    public boolean equals(Object o) {
        if (this == o) return true;
        if (o == null || getClass() != o.getClass()) return false;
        Loader<?, ?> loader = (Loader<?, ?>) o;
        return childHash == loader.childHash && lazy == loader.lazy && Objects.equals(entryCache, loader.entryCache) && Objects.equals(executor, loader.executor) && Objects.equals(children, loader.children);
    }
```

warehouse/core/src/main/java/datawave/ingest/util/cache/Loader.java
```diff
@@ -165,6 +165,10 @@ public void run() {
...
+    
     @Override
     public int hashCode() {
         return System.identityHashCode(this) + childHash;
```
jwomeara: I know you didn't add this, but why not create a legit hashcode, like this:
```suggestion
        return Objects.hash(entryCache, executor, children, childHash, lazy);
```

warehouse/core/src/main/java/datawave/iterators/filter/ageoff/DataTypeAgeOffFilter.java
```diff
@@ -295,12 +295,14 @@ protected void init(FilterOptions options, final long scanStart, IteratorEnviron
                 
                 final String dataTypeHasScanTime = options.getOption(dataType + ".hasScanTime");
                 if (Boolean.parseBoolean(dataTypeHasScanTime)) {
-                    final String scanTime = iterEnv.getConfig().get("table.custom.timestamp.current." + dataType);
-                    try {
-                        dataTypeScanTimes.put(dataType, Long.parseLong(scanTime, 10));
-                    } catch (final NumberFormatException e) {
-                        throw new NumberFormatException(dataType + " marked as hasScanTime but corresponding table.custom.timestamp.current." + dataType
-                                        + " is invalid: " + scanTime);
+                    if (iterEnv != null) {
```
jwomeara, line 298: Should we throw an exception if iterEnv is null?  That's effectively what would have happened before this change.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/CompositeIngest.java
```diff
@@ -68,7 +68,9 @@ static boolean isOverloadedCompositeField(Multimap<String,String> compositeField
     
     static boolean isOverloadedCompositeField(Collection<String> compFields, String compositeFieldName) {
         if (compFields != null && !compFields.isEmpty())
```
jwomeara: Don't like the idea of an if statement without curly braces with multiple lines afterwards...

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/CompositeIngest.java
jwomeara: why not the following?

```suggestion
            return compositeFieldName.equals(compFields.stream().findFirst().orElse(null));
```

warehouse/ingest-core/src/main/java/datawave/ingest/data/tokenize/TokenSearchSynonymFilter.java
jwomeara: +1, or Objects.hashCode()

warehouse/ingest-core/src/main/java/datawave/ingest/data/tokenize/TokenizationHelper.java
jwomeara: I agree.  If a single thread is updating the value, then volatile should be fine.  If multiple threads are updating the value, then AtomicInteger would be better to ensure the update happens atomically. 

FWIW, counter++ is _not_ an atomic operation, which is why you would want to use AtomicInteger if more than one thread were incrementing, but that's not the case here.

https://stackoverflow.com/a/19744523

warehouse/ingest-core/src/main/java/datawave/ingest/data/tokenize/TokenizationHelper.java
```diff
@@ -41,7 +42,7 @@ public void run() {
                 try {
                     Thread.sleep(INTERVAL);
                 } catch (InterruptedException e) {
-                    throw new RuntimeException(e);
+                    Thread.currentThread().interrupt();
```
jwomeara: If sleep gets interrupted, that just means our sleep interval was shorter than expected.  Do we really want to throw a RuntimeException or interrupt the thread if that happens?  Questioning whether the original logic was even desired.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/EventMapper.java
jwomeara: +1, this is not equivalent to what we had before.  If thisHelper and previousHelper are different classes that extend the same abstract class, then the old code will say they are different while the new code will say that they match.  I agree with Ivan's suggestion.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/tokenize/ContentIndexingColumnBasedHandler.java
```diff
@@ -369,16 +369,16 @@ protected void tokenizeField(final Analyzer a, final NormalizedContentInterface
...
             
             tokenizerTimeWarned = false;
             
             while (true) {
-                if (heartBeatCount != HeartBeatThread.counter) {
-                    tokenizerBeats += HeartBeatThread.counter - heartBeatCount;
-                    heartBeatCount = HeartBeatThread.counter;
+                if (heartBeatCount != HeartBeatThread.getCounter()) {
+                    tokenizerBeats += HeartBeatThread.getCounter() - heartBeatCount;
+                    heartBeatCount = HeartBeatThread.getCounter();
```
jwomeara: I think this would be more accurate.  HeartBeatThread.getCounter() could return a different value every time it's called.

```suggestion
                int currentHeartBeatCount = HeartBeatThread.getCounter()
                if (heartBeatCount != currentHeartBeatCount) {
                    tokenizerBeats += currentHeartBeatCount - heartBeatCount;
                    heartBeatCount = currentHeartBeatCount;
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/tokenize/ExtendedContentIndexingColumnBasedHandler.java
jwomeara: +1 I agree.  Same reasoning as the previous instance.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
```diff
@@ -455,9 +455,14 @@ public int run(String[] args) throws Exception {
...
-                log.info(String.format("Percent Error: %.2f", percentError));
-                if (conf.getInt("job.percent.error.threshold", 101) <= percentError) {
+                if (numRecords > 0) {
+                    float percentError = 100 * ((float) numExceptions / numRecords);
+                    log.info(String.format("Percent Error: %.2f", percentError));
+                    if (conf.getInt("job.percent.error.threshold", 101) <= percentError) {
+                        return jobFailed(job, runningJob, outputFs, workDirPath);
+                    }
+                } else {
+                    log.info("Job ran with no records");
```
jwomeara: Is it really a failure if there are no records?  I would think if you saw 0 records, then your percent error would 0%.

```suggestion
                float percentError = (numRecords == 0) ? 0.0 : 100 * ((float) numExceptions / numRecords);
                log.info(String.format("Percent Error: %.2f", percentError));
                if (conf.getInt("job.percent.error.threshold", 101) <= percentError) {
```

warehouse/ingest-core/src/main/java/datawave/ingest/util/WeightedValuePruningStrategy.java
jwomeara: The comments in the code seem to indicate that it's possible that `ratioOfTotalNgramCountsToMaxAllowed` could be a negative number (i.e. if `totalAllowedNgramsCount` were set to a negative number).  See the comment about this on line 46.

If negative numbers are allowed/expected, then this code change would affect how reducedScaleNgramCount is calculated.

jwomeara (changes requested):
I still need to review the tests, but I have some comments.

warehouse/core/src/main/java/datawave/ingest/protobuf/TermWeightPosition.java
cogross: The hashcode I used was the SonarQube suggestion.  Switching it to this one.

warehouse/core/src/main/java/datawave/ingest/util/cache/Loader.java
cogross: :+1:

warehouse/core/src/main/java/datawave/ingest/util/cache/Loader.java
cogross: okay

warehouse/core/src/main/java/datawave/iterators/filter/ageoff/DataTypeAgeOffFilter.java
cogross: Sure -- not sure what the message should be, but I'll throw an NPE so it will work the same.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/CompositeIngest.java
cogross: cleaned up.

warehouse/ingest-core/src/main/java/datawave/ingest/data/tokenize/TokenSearchSynonymFilter.java
cogross: :+1:

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
cogross: SonarQube was trying to avoid the divide by zero, I like your solution.

warehouse/ingest-core/src/main/java/datawave/ingest/util/WeightedValuePruningStrategy.java
cogross: SonarQube is trying to prevent the divide by zero.  This would prevent the error and keep it in line with what it was doing:
```
if (ratioOfTotalNgramCountsToMaxAllowed > 1.0f) {
     reducedScaleNgramCount = Math.round(((float) reducedScaleNgramCount) * ratioOfTotalNgramCountsToMaxAllowed);
} else {
     reducedScaleNgramCount = (ratioOfTotalNgramCountsToMaxAllowed == 0.0) ? 0 : Math.round(((float) reducedScaleNgramCount) / ratioOfTotalNgramCountsToMaxAllowed);
}
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/tokenize/ContentIndexingColumnBasedHandler.java
cogross: fixed

warehouse/ingest-core/src/main/java/datawave/ingest/data/tokenize/TokenizationHelper.java
cogross: Fair.  I was following their suggestion.  I went back and re-ran the scan on integration to see their logic as to why, and it didn't even show up (I did upgrade since then).
I'll revert this (and the other volatile int -> AtomicInteger) back to what it was.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/EventMapper.java
cogross: Reverted.  I forget what SQ was saying about this, and it didn't appear on the most recent scan of integration.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/tokenize/ExtendedContentIndexingColumnBasedHandler.java
cogross: reverted (reasons as above)

web-services/map-reduce/src/main/java/datawave/webservice/mr/configuration/MapReduceJobConfiguration.java
cogross: reverted

alerman (dismissed):

mineralntl (dismissed):

ivakegg (dismissed):

warehouse/core/src/main/java/datawave/ingest/protobuf/TermWeightPosition.java
jwomeara: Objects is still unused.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/CompositeIngest.java
```diff
@@ -69,8 +69,11 @@ static boolean isOverloadedCompositeField(Multimap<String,String> compositeField
     }
 
     static boolean isOverloadedCompositeField(Collection<String> compFields, String compositeFieldName) {
-        if (compFields != null && !compFields.isEmpty())
-            return compFields.stream().findFirst().get().equals(compositeFieldName);
+        if (compFields != null && !compFields.isEmpty()) {
+            if (compFields.stream().findFirst().isPresent()) {
```
jwomeara: Intellij is pointing out that this if statement is always true due to the check in the previous if statement.  let's remove the redundant if statement.

jwomeara (changes requested):
A few minor things.  Other than that, looks good.

warehouse/core/src/main/java/datawave/ingest/protobuf/TermWeightPosition.java
cogross: removed.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/CompositeIngest.java
cogross: fixed.

jwomeara (approved):

alerman (approved):

PR #1815, opened 2023-01-18
Improve TLDQueryIterator's EventDataQueryFilter

apmoriarty:
- refactored logic to only parse keys that are part of the root document
- removed DatawaveKey as parse mechanism
- added FiKeyUtil and test for parsing logical components from field index keys

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/tld/TLDQueryIterator.java
+++ b/warehouse/query-core/src/main/java/datawave/query/tld/TLDQueryIterator.java
@@ -4,7 +4,7 @@ import com.google.common.base.Function;
 import com.google.common.base.Predicate;
 import com.google.common.base.Predicates;
 import datawave.query.attributes.Document;
-import datawave.query.data.parsers.DatawaveKey;
+import datawave.core.iterators.key.util.FiKeyUtil;
 import datawave.query.function.TLDEquality;
 import datawave.query.iterator.NestedIterator;
 import datawave.query.iterator.QueryIterator;
@@ -100,7 +100,7 @@ public class TLDQueryIterator extends QueryIterator {
     /**
      * Distinct from getEvaluation filter as the FI filter is used to prevent FI hits on nonEventFields that are not indexOnly fields
      * 
-     * @return
+     * @return an {@link EventDataQueryFilter}
      */
     protected EventDataQueryFilter getFIEvaluationFilter() {
         ChainableEventDataQueryFilter chainableEventDataQueryFilter = new ChainableEventDataQueryFilter();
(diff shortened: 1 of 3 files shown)


FineAndDandy (dismissed):

warehouse/query-core/src/test/java/datawave/core/iterators/key/util/FiKeyUtilTest.java
```diff
@@ -0,0 +1,60 @@
...
+    @Test(expected = IllegalArgumentException.class)
+    public void testParseNoDatatype() {
+        Key key = new Key("row", "fi\0FIELD", "value\0uid");
+        assertEquals("datatype", FiKeyUtil.getDatatypeString(key));
+    }
+    
+    @Test(expected = IllegalArgumentException.class)
+    public void testParseNoUid() {
+        Key key = new Key("row", "fi\0FIELD", "value\0datatype");
+        assertEquals("uid", FiKeyUtil.getDatatypeString(key));
```
magrable: Is it worth noting that getUidString() will not throw an exception, but incorrectly parse the datatype as the uid in this case?

warehouse/query-core/src/main/java/datawave/query/tld/TLDQueryIterator.java
```diff
@@ -128,13 +128,16 @@ public boolean peek(@Nullable Map.Entry<Key,String> var1) {
...
+             *            a field index key
+             * @return true if the key should be kept for evaluation
              */
             @Override
             public boolean keep(Key k) {
                 boolean root = TLDEventDataFilter.isRootPointer(k);
-                DatawaveKey datawaveKey = new DatawaveKey(k);
-                return (root && getIndexOnlyFields().contains(datawaveKey.getFieldName())) || !root;
+                if (root) {
+                    return getIndexOnlyFields().contains(FiKeyUtil.getFieldString(k));
```
magrable, line 138: Maybe this is obvious.. Is the parameter guaranteed to be a field index key?

apmoriarty: In this case you are guaranteed to only have field index keys.

warehouse/query-core/src/test/java/datawave/core/iterators/key/util/FiKeyUtilTest.java
apmoriarty: Good catch -- this test is wrong. I'll push up a fix shortly.

magrable (approved):

FineAndDandy (approved):

PR #1816, opened 2023-01-18
TLDEventDataFilter now builds a correct rollover range

apmoriarty:


FineAndDandy (approved):

warehouse/query-core/src/main/java/datawave/query/predicate/TLDEventDataFilter.java
```diff
@@ -491,31 +491,37 @@ private Range getWhitelistSeek(Key current, String fieldName, Key endKey, boolea
...
+     *
+     * @param current
+     *            the current key
+     * @param end
+     *            the end key
+     * @param endInclusive
+     *            is end key inclusive flag
+     * @return a rollover range, or an empty range if the rollover range would extend beyond the end key
+     */
+    private Range getRolloverRange(Key current, Key end, boolean endInclusive) {
```
magrable, line 512: It looks like the consequence of removing sortedWhitelist is that getRolloverRange will start at the next UID's 1st field, not the 1st field that appears in the sortedWhitelist - is that ok?

apmoriarty: Yes. The seek issued by the old code would only seek to the next child -- it ignored the field portion of the range.

alerman (approved):

PR #1818, opened 2023-01-20
Update uid-shard intersection logic

apmoriarty:
Changed intersection logic in IndexInfo to only produce a document range when uids exist on both sides

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
+++ b/warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
@@ -407,9 +407,7 @@ public class IndexInfo implements Writable, UidIntersector {
         if (o.getNode() != null) {
             nodes.add(o.getNode());
         }
-        if (!delayedNodes.isEmpty()) {
-            nodes.addAll(delayedNodes);
-        }
+        nodes.addAll(delayedNodes);
         merged.myNode = TreeFlatteningRebuildingVisitor.flatten(JexlNodeFactory.createAndNode(nodes.getNodes()));
         
         return merged;
--- a/warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
+++ b/warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
@@ -5,6 +5,7 @@ import com.google.common.collect.HashMultimap;
 import com.google.common.collect.ImmutableSortedSet;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Sets;
+import datawave.query.config.ShardQueryConfiguration;
 import datawave.query.jexl.JexlNodeFactory;
 import datawave.query.jexl.nodes.ExceededOrThresholdMarkerJexlNode;
 import datawave.query.jexl.nodes.ExceededTermThresholdMarkerJexlNode;
@@ -134,9 +135,12 @@ public class IndexInfo implements Writable, UidIntersector {
      * simply use that node as our node. Otherwise, we will need to create an or node manually.
      * 
      * @param first
+     *            an IndexInfo
      * @param o
+     *            another IndexInfo
      * @param delayedNodes
-     * @return
+     *            a list of delayed terms
+     * @return the union of two IndexInfo objects and the list of delayed nodes
      */
     public IndexInfo union(IndexInfo first, IndexInfo o, List<JexlNode> delayedNodes) {
         IndexInfo merged = new IndexInfo();
@@ -208,10 +212,10 @@ public class IndexInfo implements Writable, UidIntersector {
      * Return the union of an IndexInfo object and a list of delayed nodes.
      *
      * @param o
-     *            - an IndexInfo object
+     *            an IndexInfo object
      * @param delayedNodes
-     *            - a list of delayed nodes
-     * @return - the resulting union
+     *            a list of delayed nodes
+     * @return the resulting union
      */
     public IndexInfo union(IndexInfo o, List<JexlNode> delayedNodes) {
         
@@ -307,8 +311,8 @@ public class IndexInfo implements Writable, UidIntersector {
      * Find the intersection of a list of delayed nodes.
      * 
      * @param delayedNodes
-     *            - a list of delayed nodes.
-     * @return - true or false, if any child nodes were added to this IndexInfo object.
+     *            a list of delayed nodes.
+     * @return true if any child nodes were added to this IndexInfo object.
      */
     public boolean intersect(List<JexlNode> delayedNodes) {
         
@@ -347,18 +351,27 @@ public class IndexInfo implements Writable, UidIntersector {
     }
     
     /**
-     * Intersection is where the logic becomes a bit more interesting. We will handle this in stages. So let's evaluate this in terms of three types of nodes <br>
-     * 1) {@code UNKNOWN -- COUNT likely > 20, but we can't be certain}<br>
-     * 2) {@code LARGE -- COUNT > 20 AT SOME POINT}<br>
-     * 3) {@code small < 20}
-     * 
-     * The reason we differentiate 1 from 2 is that 1 originates from the Index or from an Ivarat'd nodes.
-     * 
-     * Note that delayed nodes are propagated at all points.
-     * 
+     * The intersection logic is relatively straightforward as we only have three types of nodes
+     * <ol>
+     * <li>{@code INFINITE -- assume a hit exists}</li>
+     * <li>{@code LARGE -- uids > 20, were pruned and a count remains}</li>
+     * <li>{@code SMALL -- uids < 20}</li>
+     * </ol>
+     * See {@link #isInfinite()} for the definition of an infinite range.
+     * <p>
+     * We retain uids if and only if uids exist on both sides of the intersection
+     * <p>
+     * A uid and shard range will always produce a shard range. Never assume the uid hit exists on the other side.
+     * <p>
+     * In the case of two infinite ranges, we persist the -1 count for posterity
+     *
      * @param o
+     *            another {@link IndexInfo}
      * @param delayedNodes
-     * @return
+     *            a list of delayed terms
+     * @param uidIntersector
+     *            a class that helps intersect uids
+     * @return the result of an intersection operation on this IndexInfo and the other IndexInfo
      */
     public IndexInfo intersect(IndexInfo o, List<JexlNode> delayedNodes, UidIntersector uidIntersector) {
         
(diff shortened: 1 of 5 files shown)


warehouse/query-core/src/test/java/datawave/query/index/lookup/RangeStreamTest.java
```diff
@@ -1225,64 +1210,21 @@ public void testDropTwoPredicates() throws Exception {
         helper.addFields(Arrays.asList("FOO", "LAUGH"));
         
         // Create expected ranges verbosely, so it is obvious which shards contribute to the results.
```
ivakegg: This comments may no longer be applicable

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
```diff
@@ -389,14 +392,31 @@ protected IndexInfo intersect(long maxPossibilities, Iterable<IndexMatch> matchI
         }
         
         JexlNode newNode;
-        if (matches.size() > 1) {
+        if (matches.isEmpty()) {
+            JexlNodeSet nodes = new JexlNodeSet();
+            if (matchNode != null) {
+                nodes.add(matchNode);
+            }
+            if (!otherInfiniteNodes.isEmpty()) {
```
ivakegg: The check for empty is unnecessary

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
```diff
@@ -389,14 +392,31 @@ protected IndexInfo intersect(long maxPossibilities, Iterable<IndexMatch> matchI
...
+                nodes.addAll(otherInfiniteNodes);
+            }
+            if (!delayedNodes.isEmpty()) {
```
ivakegg: The check for empty is unncessary

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
```diff
@@ -389,14 +392,31 @@ protected IndexInfo intersect(long maxPossibilities, Iterable<IndexMatch> matchI
...
+                nodes.addAll(delayedNodes);
```
ivakegg: interesting that we were missing these before.  I wonder what misses that was resulting in?

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
```diff
@@ -447,19 +459,28 @@ public IndexInfo intersect(IndexInfo o) {
...
-            /*
-             * A) we are intersecting UNKNOWN AND small
-             */
-            if (o.onlyEvents())
-                return intersect(Math.max(count, o.count), o.uids(), o.getNode(), Lists.newArrayList(getNode()), delayedNodes);
+        
+        // infinite = (count == -1)
+        // onlyEvents = (count == uids.size())
+        
+        // if either side is only events and the other is infinite, roll to a shard range
```
ivakegg: Please add comment explaining why

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
```diff
@@ -389,14 +392,31 @@ protected IndexInfo intersect(long maxPossibilities, Iterable<IndexMatch> matchI
...
+                nodes.addAll(otherInfiniteNodes);
+            }
+            if (!delayedNodes.isEmpty()) {
+                nodes.addAll(delayedNodes);
+            }
+            newNode = TreeFlatteningRebuildingVisitor.flatten(JexlNodeFactory.createAndNode(nodes.getNodes()));
+            
+        } else if (matches.size() == 1) {
+            
+            newNode = TreeFlatteningRebuildingVisitor.flatten(matches.iterator().next().getNode());
```
ivakegg: what happens to delayed node in this case?

ivakegg: and in the case below?

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
```diff
@@ -447,19 +459,28 @@ public IndexInfo intersect(IndexInfo o) {
...
-             */
-            if (o.onlyEvents())
-                return intersect(Math.max(count, o.count), o.uids(), o.getNode(), Lists.newArrayList(getNode()), delayedNodes);
+        
+        // infinite = (count == -1)
+        // onlyEvents = (count == uids.size())
+        
+        // if either side is only events and the other is infinite, roll to a shard range
+        if ((isInfinite() && o.onlyEvents()) || (o.isInfinite() && this.onlyEvents())) {
+            JexlNodeSet nodes = new JexlNodeSet();
```
ivakegg: This block is the same as at the end of the method.  Why have a special return here?  Also can this case simply be if (isInfinite() || o.isInfinite()) and would be handled already by the logic below at lines 503-

ivakegg (changes requested):

warehouse/query-core/src/test/java/datawave/query/index/lookup/RangeStreamTest.java
apmoriarty: They are not applicable anymore, will remove in the next commit.

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
apmoriarty: The method is now obsolete. I'm removing it in the next commit.

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
apmoriarty: same as above

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
apmoriarty: Persisting delayed terms in uid matches was handled elsewhere...I think.

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
apmoriarty: Will do

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
apmoriarty: This code was removed and is now obsolete.

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
apmoriarty: Done.

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
```diff
@@ -339,203 +346,72 @@ public boolean intersect(List<JexlNode> delayedNodes) {
...
      * @param o
+     *            another {@link IndexInfo}
      * @param delayedNodes
-     * @return
+     *            a list of delayed terms
+     * @param uidIntersector
+     *            a class that helps intersect uids
+     * @return the result of an intersection operation on this IndexInfo and the other IndexInfo
      */
     public IndexInfo intersect(IndexInfo o, List<JexlNode> delayedNodes, UidIntersector uidIntersector) {
```
ivakegg, line 376: We are changing the API?  Where is this being called from?  Where was this being called from?

apmoriarty: This isn't an API change, a few parameters were missing from the method level docs.

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
```diff
@@ -339,203 +346,72 @@ public boolean intersect(List<JexlNode> delayedNodes) {
...
+        
+        // now handle updating the top level node
+        JexlNodeSet nodes = new JexlNodeSet();
+        if (this.getNode() != null) {
+            nodes.add(this.getNode());
+        }
+        if (o.getNode() != null) {
+            nodes.add(o.getNode());
+        }
+        if (!delayedNodes.isEmpty()) {
```
ivakegg: The check for !delayedNodes.isEmpty() is redundant.  addAll already does that.

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
```diff
@@ -339,203 +346,72 @@ public boolean intersect(List<JexlNode> delayedNodes) {
...
-                    Set<IndexMatch> matches = buildNodeList(ids, IndexMatchType.AND, true, Lists.newArrayList(ourDelayedNodes.getNodes()));
-                    merged.uids = ImmutableSortedSet.copyOf(matches);
-                    merged.count = merged.uids.size();
-                } else {
-                    
-                    merged.count = Math.min(count, o.count);
-                    merged.uids = ImmutableSortedSet.of();
-                }
-            }
+            // else we are left countably infinite shard ranges. Take the minimum count.
```
ivakegg, line 397: Here is where we should explain why we are choosing the shard instead of the UIDs.

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
```diff
@@ -339,203 +346,72 @@ public boolean intersect(List<JexlNode> delayedNodes) {
         return true;
     }
     
-    /**
```
ivakegg, line 342: I Love that we are getting rid of all of this logic!

ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
apmoriarty: Added

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
apmoriarty: same!

ivakegg (approved):

FineAndDandy (approved):
nice refactor

PR #1825, opened 2023-01-25
Feature/federate user operations

ivakegg:
This PR is adding the capability to invoke a remote user operations endpoint (i.e. listEffectiveAuthorizations) to enable a client to determine the effective authorizations in total that can be used across the local and remote systems.  The following changes are in here:
1. Added a RemoteUserOperations interface and a RemoteUserOperationsImpl.  These work just like the existing RemoteQueryService and RemoteQueryServiceImpl.
2. Added a remoteUserOperations attribute on the QueryLogic and BaseQueryLogic which is expected to be null for all local query logics
3. The UserOperationsBean will aggregate any configured RemoteUserOperations into its responses by default.
4. The RunningQuery will downgrade user authorizations base on the combined auths of the local and remote systems.
5. The CompositeQueryLogic will subsequent subset the auths when calling the various underlying query logics.
6. Added a ConditionalRemoteUserOperations class that can be used to filter calls to the underlying RemoteUserOperations delegate.  This can be used to, for example, avoid calling the underlying listEffectiveAuthorizations is it appears that the local principal may not need any different auths.
7. Added a trusted dn mechanism that will allow the webserver to remove trusted servers from the proxied entity change.


warehouse/query-core/src/test/resources/datawave/query/QueryLogicFactory.xml
```diff
@@ -278,4 +278,7 @@
     
     <bean id="easyRoleManager" class="datawave.webservice.query.logic.EasyRoleManager" />
     
+    <util:list id="RemoteUserServices" value-type="datawave.webservice.common.remote.RemoteUserOperations">
```
ivakegg: Used to configure the UserOperationsBean with any additional remote user operations

web-services/client/src/main/java/datawave/webservice/query/QueryImpl.java
```diff
@@ -413,6 +414,12 @@ public QueryImpl duplicate(String newQueryName) {
         if (null != this.parameters && !this.parameters.isEmpty())
             query.setParameters(new HashSet<Parameter>(this.parameters));
         query.setDnList(this.dnList);
+        if (null != this.optionalQueryParameters && !this.optionalQueryParameters.isEmpty()) {
```
ivakegg: These were oversights in the duplicate method for a QueryImpl

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
```diff
@@ -135,7 +154,7 @@ public void run() {
     
     protected static final Logger log = Logger.getLogger(CompositeQueryLogic.class);
     
-    private List<QueryLogic<?>> queryLogics = null;
+    private Map<String,QueryLogic<?>> queryLogics = null;
```
ivakegg, line 157: Changed to a map for easier relabeling of underlying child query logics.

warehouse/query-core/src/main/java/datawave/query/tables/RemoteEventQueryLogic.java
```diff
@@ -255,4 +258,13 @@ public EventBase transform(EventBase input) throws EmptyObjectException {
         
     }
     
+    @Override
+    public void setRemoteUserOperations(RemoteUserOperations userService) {
+        this.remoteUserOperatons = userService;
+    }
+    
+    @Override
+    public RemoteUserOperations getUserOperations() {
```
drewfarris: Consider renaming the getter to `getRemoteUserOperations` be congruent with the setter
```suggestion
    public RemoteUserOperations getRemoteUserOperations() {
```

web-services/client/src/main/java/datawave/webservice/query/QueryImpl.java
drewfarris: We use `StringUtils` for a null-safe check in many places, why not use `MapUtils` too?
```suggestion
        if (!MapUtils.isEmpty(this.optionalQueryParameters)) {
```

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
```diff
@@ -86,8 +99,11 @@ public static Set<Authorizations> getDowngradedAuthorizations(String requestedAu
      * @return A set of {@link Authorizations}, one per entity represented in {@code principal}. The user's auths are replaced by {@code requestedAuths} so long
      *         as the user actually had all of the auths. If {@code requestedAuths} is {@code null}, then the user's auths are returned as-is.
      */
-    public static LinkedHashSet<Authorizations> getDowngradedAuthorizations(String requestedAuths, DatawavePrincipal principal) {
-        
+    public static LinkedHashSet<Authorizations> getDowngradedAuthorizations(String requestedAuths, DatawavePrincipal principal, RemoteUserService userService)
```
drewfarris: Please update the javadoc for this method to include a description of what role the `RemoteUserService` plays in this process.

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
```diff
@@ -105,11 +121,15 @@ public static LinkedHashSet<Authorizations> getDowngradedAuthorizations(String r
...
      */
-    public static String downgradeUserAuths(Principal principal, String requested) {
+    public static String downgradeUserAuths(Principal principal, String requested, RemoteUserService userService) throws AuthorizationException {
         if (requested == null || requested.trim().isEmpty()) {
             throw new IllegalArgumentException("Requested authorizations must not be empty");
         }
         
+        if (userService != null) {
+            principal = userService.getRemoteUser((DatawavePrincipal) principal);
+        }
```
drewfarris, line 174: If I'm reading this correctly, replacing the local principal with the remote principal makes the assumption that we always want the remote auths to replace the local auths wholesale. 

I think this means that we will always expect the remote auths to be a superset of the local auths. How much complexity would it add if we assume it would be possible to the systems to have differerent auths? For example, the local system might have auths the remote system doesn't have any vice versa - yet we want to allow the user to submit a single list of desired auths for both systems.

web-services/common-util/src/test/java/datawave/security/util/AuthorizationsUtilTest.java
```diff
@@ -64,34 +65,34 @@ public void testMergeAuthorizations() {
     }
     
     @Test
-    public void testDowngradeAuthorizations() {
+    public void testDowngradeAuthorizations() throws AuthorizationException {
```
drewfarris, line 83: Consider adding a few tests that include a mock `RemoteUserService`?

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
```diff
@@ -64,14 +67,24 @@ public static Set<Authorizations> mergeAuthorizations(String requestedAuths, Col
...
+                    throws AuthorizationException {
         if (principal instanceof DatawavePrincipal) {
-            return getDowngradedAuthorizations(requestedAuths, (DatawavePrincipal) principal);
+            return getDowngradedAuthorizations(requestedAuths, (DatawavePrincipal) principal, userService);
         } else {
             return Collections.singleton(new Authorizations());
         }
     }
     
+    public static interface RemoteUserService {
```
drewfarris: Consider making this a top-level interface as opposed a static nested interface in this class. It seems a little excessive, but generally should avoid confusion in the future. It seems pretty obvious as to what it does, but minimal javadoc could be helpful down the road.

drewfarris: Is this really necessary, considering the fact that `RemoteUserOperations` is an interface? Consider getting rid of either `RemoteUserService` or `RemoteUserOperations`

web-services/security/src/main/java/datawave/security/user/UserOperationsBean.java
```diff
@@ -109,10 +124,39 @@ public AuthorizationsListBase listEffectiveAuthorizations() {
             DatawavePrincipal datawavePrincipal = (DatawavePrincipal) p;
             name = datawavePrincipal.getShortName();
             
+            // if we have any remote services configured, merge those authorizations in here
+            if (includeRemoteServices && remoteServices != null && !remoteServices.isEmpty()) {
```
drewfarris: Consider `CollectionUtils`?
```suggestion
            if (includeRemoteServices && !CollectionUtils.isEmpty(remoteServices)) {
```

web-services/security/src/main/java/datawave/security/user/UserOperationsBean.java
```diff
@@ -98,8 +113,8 @@ public class UserOperationsBean {
     @Path("/listEffectiveAuthorizations")
     @Produces({"application/xml", "text/xml", "text/plain", "application/json", "text/yaml", "text/x-yaml", "application/x-yaml", "application/x-protobuf",
             "text/html"})
-    public AuthorizationsListBase listEffectiveAuthorizations() {
-        
+    public AuthorizationsListBase listEffectiveAuthorizations(@DefaultValue("true") @QueryParam("includeRemoteServices") boolean includeRemoteServices)
```
drewfarris: javadoc update needed to add both `@param` and `@throws`

web-services/security/src/main/java/datawave/security/user/UserOperationsBean.java
```diff
@@ -137,10 +182,22 @@ public AuthorizationsListBase listEffectiveAuthorizations() {
...
             "application/x-protostuff"})
     @PermitAll
-    public GenericResponse<String> flushCachedCredentials() {
+    public GenericResponse<String> flushCachedCredentials(@DefaultValue("true") @QueryParam("includeRemoteServices") boolean includeRemoteServices) {
         GenericResponse<String> response = new GenericResponse<>();
         Principal callerPrincipal = context.getCallerPrincipal();
         log.info("Flushing credentials for " + callerPrincipal + " from the cache.");
+        
+        // if we have any remote services configured, then flush those credentials as well
+        if (includeRemoteServices && remoteServices != null && !remoteServices.isEmpty()) {
```
drewfarris: Another opportunity to use `CollectionUtils.isEmpty(..)`

web-services/security/src/main/java/datawave/security/user/UserOperationsBean.java
```diff
@@ -137,10 +182,22 @@ public AuthorizationsListBase listEffectiveAuthorizations() {
     @Produces({"application/xml", "text/xml", "application/json", "text/yaml", "text/x-yaml", "application/x-yaml", "application/x-protobuf",
             "application/x-protostuff"})
     @PermitAll
-    public GenericResponse<String> flushCachedCredentials() {
+    public GenericResponse<String> flushCachedCredentials(@DefaultValue("true") @QueryParam("includeRemoteServices") boolean includeRemoteServices) {
```
drewfarris, line 175: Another opportunity to improve javadoc.

web-services/security/src/main/java/datawave/security/user/UserOperationsBean.java
```diff
@@ -137,10 +182,22 @@ public AuthorizationsListBase listEffectiveAuthorizations() {
...
         Principal callerPrincipal = context.getCallerPrincipal();
         log.info("Flushing credentials for " + callerPrincipal + " from the cache.");
+        
+        // if we have any remote services configured, then flush those credentials as well
+        if (includeRemoteServices && remoteServices != null && !remoteServices.isEmpty()) {
+            for (RemoteUserOperations remote : remoteServices) {
+                try {
+                    remote.flushCachedCredentials(callerPrincipal);
+                } catch (Exception e) {
+                    log.error("Failed to flush user from remote user service", e);
```
drewfarris, line 194: Consider how we might alert the caller that the remote operation failed. It would not be fun to be the person calling the cache flush to resolve an issue and having it silently fail on the remote with no indication other than the logs.

web-services/security/src/test/java/datawave/security/authorization/remote/RemoteUserOperationsImplHttpTest.java
```diff
@@ -0,0 +1,376 @@
...
+        // create a remote event query logic that has our own server behind it
+        remote = new RemoteUserOperationsImpl();
+        remote.setQueryServiceURI("/Security/User/");
+        remote.setQueryServiceScheme("http");
+        remote.setQueryServiceHost("localhost");
+        remote.setQueryServicePort(server.getAddress().getPort());
+        remote.setExecutorService(null);
+        remote.setObjectMapperDecorator(new DefaultMapperDecorator());
+        remote.setResponseObjectFactory(new MockResponseObjectFactory());
+        remote.setJsseSecurityDomain(new JSSESecurityDomain() {
```
drewfarris: Consider a static inner class instead of an anonymous class here. It would improve readability. Maybe even a top level class - it might be something that someone else could use?

web-services/security/src/test/java/datawave/security/authorization/remote/RemoteUserOperationsImplHttpTest.java
```diff
@@ -0,0 +1,376 @@
+package datawave.security.authorization.remote;
```
drewfarris, line 1: Nice Test. I could use something like this. I think we have a couple of these 'Http` tests now, are there common items we can extract from them?

web-services/security/src/main/java/datawave/security/authorization/remote/RemoteUserOperationsImpl.java
```diff
@@ -0,0 +1,279 @@
...
+            return executeGetMethod(uriSuffix, uriCustomizer, requestCustomizer, resultConverter, errorSupplier);
+        } catch (URISyntaxException e) {
+            throw new RuntimeException("Invalid URI: " + e.getMessage(), e);
+        } catch (IOException e) {
+            config.getFailureCounter().inc();
+            throw new RuntimeException(e.getMessage(), e);
+        }
+    }
+    
+    @Override
```
drewfarris: The level of delegation we need to implement here makes me sad. Do we have to do this everywhere? Could we do it in a base class.

web-services/security/src/main/java/datawave/security/authorization/remote/RemoteUserOperationsImpl.java
```diff
@@ -0,0 +1,279 @@
...
+        char[] buffer = new char[1024];
+        int chars = reader.read(buffer);
+        while (chars >= 0) {
+            builder.append(buffer, 0, chars);
+            chars = reader.read(buffer);
+        }
+        return builder.toString();
+    }
+    
+    public static String getEntities(Object callerObject) {
```
drewfarris: Consider `getProxiedEntities` instead of `getEntities` for clarity.

web-services/security/src/main/java/datawave/security/authorization/remote/RemoteUserOperationsImpl.java
```diff
@@ -0,0 +1,279 @@
...
+        if (callerObject instanceof DatawavePrincipal) {
+            DatawavePrincipal callerPrincipal = (DatawavePrincipal) callerObject;
+            return callerPrincipal.getProxiedUsers().stream().map(u -> new StringBuilder().append('<').append(u.getDn().subjectDN()).append('>'))
+                            .collect(Collectors.joining());
+        } else {
+            throw new RuntimeException("Cannot use " + callerObject.getClass().getName() + " as a caller object");
```
drewfarris: Consider using checked exceptions for `getEntities` and `getIssuers` or at least something less generic (e.g., `IllegalArgumentException`)

web-services/security/src/main/java/datawave/security/authorization/remote/RemoteUserOperationsImpl.java
```diff
@@ -0,0 +1,279 @@
...
+            return null;
+        } else {
+            String content = getContent(entity.getContent());
+            try {
+                return reader.readValue(content);
+            } catch (IOException ioe) {
+                log.error("Failed to read entity content.  Trying as a VoidResponse.", ioe);
+                log.error(content);
+                VoidResponse response = voidResponseReader.readValue(content);
+                throw new RuntimeException(response.getMessages().toString());
```
drewfarris: Since this method already throws an `IOEexception`, consider just re-throwing `ioe` here? 

If not, consider adding the `ioe` as the init cause for the `RuntimeException` so we don't lose stack informaton?

web-services/security/src/main/java/datawave/security/authorization/remote/ConditionalRemoteUserOperations.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+/**
+ * A conditional remote user operations will only invoke the delegate remote service base on a specified function of the local principal
+ */
```
drewfarris, line 22: An example of what you mean by `specified function of the local principal` could be helpful here.

web-services/security/src/main/java/datawave/security/authorization/remote/ConditionalRemoteUserOperations.java
```diff
@@ -0,0 +1,114 @@
...
+    @Override
+    public GenericResponse<String> flushCachedCredentials(Object callerObject) {
+        if (condition.apply((DatawavePrincipal) callerObject)) {
+            return delegate.flushCachedCredentials(callerObject);
+        } else {
+            return EMPTY_RESPONSE;
+        }
+    }
+    
+    private AuthorizationsListBase getAuthorizationsList(Object callerObject) {
```
drewfarris: A brief description of what this is doing and why might be helpful. It looks like we're just trying to get the auths for each proxied user into the authMap and then copy them them from there to the auths object, but I'm a little confused about the need to filter the authMap by the dn of the primary user.

Could we avoid putting the auths of the primary user in the authMap in the first place?

web-services/security/src/main/java/datawave/security/authorization/remote/ConditionalRemoteUserOperations.java
```diff
@@ -0,0 +1,114 @@
...
+        AuthorizationsListBase auths = responseObjectFactory.getAuthorizationsList();
+        if (callerObject instanceof DatawavePrincipal) {
+            DatawavePrincipal principal = (DatawavePrincipal) callerObject;
+            Map<AuthorizationsListBase.SubjectIssuerDNPair,Set<String>> authMap = new HashMap<>();
+            principal.getProxiedUsers().forEach(u -> authMap.put(dn(u.getDn()), new HashSet<>(u.getAuths())));
+            DatawaveUser primaryUser = principal.getPrimaryUser();
+            AuthorizationsListBase.SubjectIssuerDNPair primaryDn = dn(primaryUser.getDn());
+            auths.setUserAuths(primaryDn.subjectDN, primaryDn.issuerDN, authMap.get(dn(primaryUser.getDn())));
+            authMap.entrySet().stream().filter(e -> !e.getKey().equals(primaryDn))
+                            .forEach(e -> auths.addAuths(e.getKey().subjectDN, e.getKey().issuerDN, e.getValue()));
```
drewfarris: Using `stream` in one case and `forEach` in another (and then mixing them) is a little rough here and at `77` - could we use a streams and collectors instead? 

For readability, good old for loops might be preferred. 

Maybe we could simplify this method in general.

web-services/security/.gitignore
```diff
@@ -1 +1,2 @@
 /target/
+.keystore
```
drewfarris: Since `.keystore` is test output, perhaps we should put in in the `target` directory.

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
```diff
@@ -86,8 +99,11 @@ public static Set<Authorizations> getDowngradedAuthorizations(String requestedAu
...
      */
-    public static LinkedHashSet<Authorizations> getDowngradedAuthorizations(String requestedAuths, DatawavePrincipal principal) {
-        
+    public static LinkedHashSet<Authorizations> getDowngradedAuthorizations(String requestedAuths, DatawavePrincipal principal, RemoteUserService userService)
+                    throws AuthorizationException {
+        if (userService != null) {
+            principal = userService.getRemoteUser((DatawavePrincipal) principal);
+        }
         final DatawaveUser primaryUser = principal.getPrimaryUser();
         UserAuthFunctions uaf = UserAuthFunctions.getInstance();
```
drewfarris: Did you consider overriding the default implementation of `UserAuthFunctions` using the `datawave.user.auth.functions.class` property when thinking about this implementation? It seems to be a natural extension point, but is pretty low-level and might not provide the appropriate hooks to customize the behavior you need.

web-services/common/src/main/java/datawave/webservice/common/remote/RemoteUserOperations.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * A remote query service is one that can pass calls off to another external user operations endpoint
+ */
+public interface RemoteUserOperations extends AuthorizationsUtil.RemoteUserService {
```
drewfarris: Consider eliminating RemoteUserService or RemoteUserOperations - two levels of interfaces here are potentially overkill.

web-services/query/src/main/java/datawave/webservice/query/logic/QueryLogic.java
```diff
@@ -379,4 +380,12 @@ default long getResultLimit(Collection<String> dns) {
...
      */
     void setPageProcessingStartTime(long pageProcessingStartTime);
+    
+    /**
+     * Normally this simply returns null as the query logic simply uses the user within the local system. However sometimes we may have a query that goes off
+     * system and hence needs to use a different set of auths when downgrading.
+     *
+     * @return A remote user operations interface implementation. Null if NA.
+     */
+    RemoteUserOperations getUserOperations();
```
drewfarris: This may be a bridge too far, but It would be really cool if this always returned an implementation we would use for all user operations. Sometimes that implementation would be something that reaches out to a remote. , otherwise, it would just be for local operations. 

In other words, if we're going to add this to some very central Datawave APIs, it could be really nice if it was somewhat generic and not specifically designed around performing remote user operations.

web-services/query/src/main/java/datawave/webservice/query/util/LookupUUIDUtil.java
```diff
@@ -314,18 +324,31 @@ public <T> T createUUIDQueryAndNext(final AbstractUUIDLookupCriteria unvalidated
...
+            try {
+                QueryLogic<?> logic = queryLogicFactory.getQueryLogic(logicName, principal);
+                if (queryParameters.containsKey(QueryParameters.QUERY_AUTHORIZATIONS)) {
+                    userAuths = AuthorizationsUtil.downgradeUserAuths(principal, queryParameters.getFirst(QueryParameters.QUERY_AUTHORIZATIONS),
+                                    logic.getUserOperations());
+                } else {
+                    if (logic.getUserOperations() != null) {
+                        userAuths = AuthorizationsUtil.buildUserAuthorizationString(logic.getUserOperations().getRemoteUser((DatawavePrincipal) principal));
+                    } else {
+                        userAuths = AuthorizationsUtil.buildUserAuthorizationString(principal);
```
drewfarris: Consider creating an updated `buildUserAuthorizationString` that takes the user operations object just like `downgradeUserAuthorizations` does. Ideally we could hide details of things like `getRemoteUser` from most other code in datawave (and remove the conditional that checks for null from this class, because ideally this class shouldn't have to worry about it)

web-services/query/src/main/java/datawave/webservice/query/util/LookupUUIDUtil.java
```diff
@@ -363,12 +386,11 @@ public <T> T createUUIDQueryAndNext(final AbstractUUIDLookupCriteria unvalidated
...
             // If the headers are defined as part of a standard UUID lookup, execute the query for a streamed response
             if (!validatedCriteria.isContentLookup() && (null != headers)) {
-                response = (T) this.queryExecutor.execute(queryParameters.getFirst(QueryParameters.QUERY_LOGIC_NAME), queryParameters,
-                                validatedCriteria.getStreamingOutputHeaders());
+                response = (T) this.queryExecutor.execute(logicName, queryParameters, validatedCriteria.getStreamingOutputHeaders());
             }
             // Otherwise, execute the query for a standard paged response
             else {
-                response = (T) this.queryExecutor.createQueryAndNext(queryParameters.getFirst(QueryParameters.QUERY_LOGIC_NAME), queryParameters);
+                response = (T) this.queryExecutor.createQueryAndNext(logicName, queryParameters);
```
drewfarris, line 380: This is much nicer than it was before.

web-services/query/src/main/java/datawave/webservice/query/logic/BaseQueryLogic.java
```diff
@@ -395,4 +396,10 @@ public ResponseEnricherBuilder getResponseEnricherBuilder() {
     public void setResponseEnricherBuilder(ResponseEnricherBuilder responseEnricherBuilder) {
         this.responseEnricherBuilder = responseEnricherBuilder;
     }
+    
+    @Override
+    public RemoteUserOperations getUserOperations() {
```
drewfarris: Consider removing the distinction between remote and local operations and just providing a `UserOperations` class that gets plugged into the query logics. Different implementations of this would perform either local-only or local-remote operations. This might clean up the API a bit.

web-services/examples/client-login/src/main/java/datawave/webservice/examples/ClientLoginExampleBean.java
```diff
@@ -75,7 +75,7 @@ public void handle(Callback[] callbacks) {
             
             // Call secured EJBs
             try {
-                AuthorizationsListBase auths = userOps.listEffectiveAuthorizations();
+                AuthorizationsListBase auths = userOps.listEffectiveAuthorizations(false);
```
drewfarris, line 78: Consider whether there is a way we can avoid the need to add this flag in the API.

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeUserOperations.java
```diff
@@ -0,0 +1,105 @@
...
+    final boolean includeLocal;
+    
+    public CompositeUserOperations(List<RemoteUserOperations> remoteOperations, boolean includeLocal, ResponseObjectFactory responseObjectFactory) {
+        this.responseObjectFactory = responseObjectFactory;
+        this.remoteOperations = remoteOperations;
+        this.includeLocal = includeLocal;
+    }
+    
+    @Override
+    public AuthorizationsListBase listEffectiveAuthorizations(Object callerObject) throws AuthorizationException {
```
drewfarris, line 42: I feel like we have similar code elsewhere in this PR, I wonder if there are ways we can consolidate?

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
```diff
@@ -145,52 +164,80 @@ public void run() {
...
                 }
-                this.queryLogics.add(logic);
             }
             setPrincipal(other.getPrincipal());
         }
         this.allMustInitialize = other.allMustInitialize;
     }
     
+    public static Set<Authorizations> updateRuntimeAuthorizationsAndQueryAuths(QueryLogic<?> logic, Query settings) throws AuthorizationException {
+        Set<String> requestedAuths = new HashSet<>(AuthorizationsUtil.splitAuths(settings.getQueryAuthorizations()));
```
drewfarris, line 189: Consider whether it would be better to avoid calling `splitAuths` with each call to this method.

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
```diff
@@ -145,52 +164,80 @@ public void run() {
...
+        
+        // determine the valid authorizations for this call to be the user's auths for this logic
+        DatawavePrincipal principal = (DatawavePrincipal) logic.getPrincipal();
+        if (logic.getUserOperations() != null) {
+            principal = logic.getUserOperations().getRemoteUser(principal);
```
drewfarris: See question elsewhere regarding the implications of replacing the local principal with the remote principal.

warehouse/query-core/src/main/java/datawave/query/tables/RemoteEventQueryLogic.java
ivakegg: sure, ok

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
ivakegg: ok

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
ivakegg: We always want the remote auths to replace the local auths for a remote query logic.  The composite / federated query logic will combine the local principal with the remote principal to allow that query logic to work correctly.  The two sets of auths do not actually have to overlap at all.

web-services/common-util/src/test/java/datawave/security/util/AuthorizationsUtilTest.java
ivakegg: right, the code got covered in other test cases but some specific unit tests would be appropriate here.

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
ivakegg: I guess getting rid of would be reasonable and result in less confusion.

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
ivakegg: I did consider that but the whole remote user service seemed more to be couple with remote query logics and hence did not necessarily belong in the authorization service.

web-services/examples/client-login/src/main/java/datawave/webservice/examples/ClientLoginExampleBean.java
ivakegg: The flag is a query parameter and is optional in the http API.  I can add convience methods to the bean if you think that is better.  I think passing in an explicit flag works better from a maintainance perspective.  What you want thec call to do is more obvious.

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeUserOperations.java
ivakegg: I think this is the only place where we actually merge authorization, so this is actually unique code.

web-services/query/src/main/java/datawave/webservice/query/util/LookupUUIDUtil.java
ivakegg: sure, that makes sense.

ivakegg: And this actually finds a few other beans that need to be able to handle the remote user service..... sigh.

warehouse/query-core/src/main/java/datawave/query/tables/RemoteEventQueryLogic.java
```diff
@@ -255,4 +258,13 @@ public EventBase transform(EventBase input) throws EmptyObjectException {
         
     }
     
+    @Override
+    public void setRemoteUserOperations(RemoteUserOperations userService) {
```
apmoriarty: rename `userService` to either `userOperations` or `remoteUserOperations`

warehouse/query-core/src/main/java/datawave/query/tables/remote/RemoteQueryLogic.java
```diff
@@ -1,11 +1,14 @@
...
+import datawave.security.authorization.RemoteUserOperations;
 import datawave.webservice.query.logic.QueryLogic;
 
 /**
  * A remote query logic is is a query logic that uses a remote query service.
  */
 public interface RemoteQueryLogic<T> extends QueryLogic<T> {
     public void setRemoteQueryService(RemoteQueryService service);
+    
+    public void setRemoteUserOperations(RemoteUserOperations userService);
```
apmoriarty: rename `userService` to either `remoteUserOperations` or `remoteOperations`

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
```diff
@@ -64,9 +68,32 @@ public static Set<Authorizations> mergeAuthorizations(String requestedAuths, Col
...
+     *            The auths to use for the user's auths. If this list contains any that are not owned by the user, an {@link IllegalArgumentException} is
+     *            thrown.
+     * @param principal
+     *            The principal from which to retrieve entity authorizations.
+     * @param userService
+     *            The remote user service if any which will be used to determine the remote authorizations. If this is null, then the auths in the supplied
+     *            principal will be used. Otherwise this service will be used to retrieve the remote principal from which the auths will be determined.
+     * @return A set of {@link Authorizations}, one per entity represented in {@code principal}. The user's auths are replaced by {@code requestedAuths} so long
+     *         as the user actually had all of the auths. If {@code requestedAuths} is {@code null}, then the user's auths are returned as-is.
+     */
```
apmoriarty: Good docs

web-services/pom.xml
```diff
@@ -40,7 +40,7 @@
         <version.abdera>1.1.3</version.abdera>
         <version.asm>4.0</version.asm>
         <version.axiom>1.2.14</version.axiom>
-        <version.bcprov>1.46</version.bcprov>
+        <version.bcprov>1.67</version.bcprov>
```
apmoriarty: Can you provide a quick explanation for this version bump?

web-services/query/src/main/java/datawave/webservice/query/logic/BaseQueryLogic.java
```diff
@@ -395,4 +396,10 @@ public ResponseEnricherBuilder getResponseEnricherBuilder() {
     public void setResponseEnricherBuilder(ResponseEnricherBuilder responseEnricherBuilder) {
         this.responseEnricherBuilder = responseEnricherBuilder;
     }
+    
+    @Override
+    public RemoteUserOperations getRemoteUserOperations() {
+        // null implies that the local user operations/principal is to be used for auths.
```
apmoriarty, line 402: +1

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeLogicException.java
```diff
@@ -1,20 +1,17 @@
...
-        exceptions.stream().forEach(e -> addSuppressed(e.getValue()));
+        exceptions.values().stream().forEach(e -> addSuppressed(e));
     }
     
-    private static String getMessage(String message, List<Map.Entry<QueryLogic<?>,Exception>> exceptions) {
+    private static String getMessage(String message, Map<String,Exception> exceptions) {
         StringBuilder builder = new StringBuilder();
         builder.append(message).append(":");
-        exceptions.stream().forEach(e -> builder.append('\n').append(e.getKey().getLogicName()).append(": ").append(e.getValue().getMessage()));
+        exceptions.entrySet().stream().forEach(e -> builder.append('\n').append(e.getKey()).append(": ").append(e.getValue().getMessage()));
```
apmoriarty, line 14: Nice cleanup. This version is much easier to read.

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeUserOperations.java
```diff
@@ -0,0 +1,104 @@
...
+import datawave.webservice.result.GenericResponse;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class CompositeUserOperations implements RemoteUserOperations {
```
apmoriarty: Add a brief description, even if it's only a couple of sentences.

web-services/query/src/main/java/datawave/webservice/query/util/LookupUUIDUtil.java
apmoriarty: Nice catch

web-services/security/pom.xml
```diff
@@ -82,7 +82,7 @@
         </dependency>
         <dependency>
             <groupId>org.bouncycastle</groupId>
-            <artifactId>bcprov-jdk16</artifactId>
+            <artifactId>bcprov-jdk15on</artifactId>
```
apmoriarty: Was this a downgrade? Why was this change made?

web-services/security/src/main/java/datawave/security/authorization/remote/ConditionalRemoteUserOperations.java
apmoriarty: For a code block like this I also have a preference for readable for-loops

apmoriarty (changes requested):

web-services/security/src/main/java/datawave/security/authorization/remote/RemoteUserOperationsImpl.java
ivakegg: It makes me sad too.  We are fixing this in the query microservices to be a configuration object instead of a mess of setters.  Can we change this later?

web-services/security/src/test/java/datawave/security/authorization/remote/RemoteUserOperationsImplHttpTest.java
ivakegg: considering

web-services/common-util/src/main/java/datawave/security/authorization/RemoteUserOperations.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * A remote query service is one that can pass calls off to another external user operations endpoint
```
billoley:
```suggestion
 * A remote user operations service is one that can pass calls off to another external user operations endpoint
```

warehouse/query-core/src/test/resources/datawave/query/QueryLogicFactory.xml
```diff
@@ -278,4 +278,7 @@
     
     <bean id="easyRoleManager" class="datawave.webservice.query.logic.EasyRoleManager" />
     
+    <util:list id="RemoteUserServices" value-type="datawave.security.authorization.RemoteUserOperations">
```
billoley:
```suggestion
    <util:list id="RemoteUserOperationsList" value-type="datawave.security.authorization.RemoteUserOperations">
```

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
```diff
@@ -64,9 +66,32 @@ public static Set<Authorizations> mergeAuthorizations(String requestedAuths, Col
...
+    public static Set<Authorizations> getDowngradedAuthorizations(String requestedAuths, Principal principal, RemoteUserOperations userService)
+                    throws AuthorizationException {
         if (principal instanceof DatawavePrincipal) {
-            return getDowngradedAuthorizations(requestedAuths, (DatawavePrincipal) principal);
+            return getDowngradedAuthorizations(requestedAuths, (DatawavePrincipal) principal, userService);
```
billoley:
```suggestion
            return getDowngradedAuthorizations(requestedAuths, (DatawavePrincipal) principal, remoteUserOperations);
```

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
```diff
@@ -64,9 +66,32 @@ public static Set<Authorizations> mergeAuthorizations(String requestedAuths, Col
...
+     *            thrown.
+     * @param principal
+     *            The principal from which to retrieve entity authorizations.
+     * @param userService
+     *            The remote user service if any which will be used to determine the remote authorizations. If this is null, then the auths in the supplied
+     *            principal will be used. Otherwise this service will be used to retrieve the remote principal from which the auths will be determined.
+     * @return A set of {@link Authorizations}, one per entity represented in {@code principal}. The user's auths are replaced by {@code requestedAuths} so long
+     *         as the user actually had all of the auths. If {@code requestedAuths} is {@code null}, then the user's auths are returned as-is.
+     */
+    public static Set<Authorizations> getDowngradedAuthorizations(String requestedAuths, Principal principal, RemoteUserOperations userService)
```
billoley:
```suggestion
    public static Set<Authorizations> getDowngradedAuthorizations(String requestedAuths, Principal principal, RemoteUserOperations remoteUserOperations)
```

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
```diff
@@ -83,11 +108,17 @@ public static Set<Authorizations> getDowngradedAuthorizations(String requestedAu
...
+     *            principal will be used. Otherwise this service will be used to retrieve the remote principal from which the auths will be determined.
      * @return A set of {@link Authorizations}, one per entity represented in {@code principal}. The user's auths are replaced by {@code requestedAuths} so long
      *         as the user actually had all of the auths. If {@code requestedAuths} is {@code null}, then the user's auths are returned as-is.
      */
-    public static LinkedHashSet<Authorizations> getDowngradedAuthorizations(String requestedAuths, DatawavePrincipal principal) {
-        
+    public static LinkedHashSet<Authorizations> getDowngradedAuthorizations(String requestedAuths, DatawavePrincipal principal, RemoteUserOperations userService)
+                    throws AuthorizationException {
+        if (userService != null) {
+            principal = userService.getRemoteUser((DatawavePrincipal) principal);
```
billoley:
```suggestion
            principal = userService.getRemoteUser(principal);
```
principal already is a DatawavePrincipal

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
```diff
@@ -103,16 +134,19 @@ public static LinkedHashSet<Authorizations> getDowngradedAuthorizations(String r
      *            the principal representing the user to verify that {@code requested} are all valid authorizations
      * @param requested
      *            the requested downgrade authorizations
+     * @param userService
+     *            The remote user service if any which will be used to determine the remote authorizations. If this is null, then the auths in the supplied
+     *            principal will be used. Otherwise this service will be used to retrieve the remote principal from which the auths will be determined.
      * @return requested, unless the user represented by {@code principal} does not have one or more of the auths in {@code requested}
      */
-    public static String downgradeUserAuths(Principal principal, String requested) {
+    public static String downgradeUserAuths(Principal principal, String requested, RemoteUserOperations userService) throws AuthorizationException {
```
billoley:
```suggestion
    public static String downgradeUserAuths(Principal principal, String requested, RemoteUserOperations remoteUserOperations) throws AuthorizationException {
```

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
```diff
@@ -103,16 +134,19 @@ public static LinkedHashSet<Authorizations> getDowngradedAuthorizations(String r
...
         if (requested == null || requested.trim().isEmpty()) {
             throw new IllegalArgumentException("Requested authorizations must not be empty");
         }
         
         List<String> requestedList = AuthorizationsUtil.splitAuths(requested);
         // Find all authorizations the user has access to
-        String userAuths = AuthorizationsUtil.buildUserAuthorizationString(principal);
+        String userAuths = AuthorizationsUtil.buildUserAuthorizationString(principal, userService);
```
billoley:
```suggestion
        String userAuths = AuthorizationsUtil.buildUserAuthorizationString(principal, remoteUserOperations);
```

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
```diff
@@ -166,7 +200,20 @@ public static String buildAuthorizationString(Collection<? extends Collection<St
...
+     * Build the authorization string for a prinbcipal.
+     *
+     * @param principal
+     *            the principal representing the user to verify that {@code requested} are all valid authorizations
+     * @param userService
+     *            The remote user service if any which will be used to determine the remote authorizations. If this is null, then the auths in the supplied
+     *            principal will be used. Otherwise this service will be used to retrieve the remote principal from which the auths will be determined.
+     * @return user authorizations string
+     */
+    public static String buildUserAuthorizationString(Principal principal, RemoteUserOperations userService) throws AuthorizationException {
```
billoley:
```suggestion
    public static String buildUserAuthorizationString(Principal principal, RemoteUserOperations remoteUserOperations) throws AuthorizationException {
```

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
```diff
@@ -166,7 +200,20 @@ public static String buildAuthorizationString(Collection<? extends Collection<St
...
+        if (userService != null) {
+            principal = userService.getRemoteUser((DatawavePrincipal) principal);
+        }
```
billoley:
```suggestion
        if (remoteUserOperations != null) {
            principal = remoteUserOperations.getRemoteUser((DatawavePrincipal) principal);
        }
```

billoley (changes requested):

web-services/query/src/main/java/datawave/webservice/query/logic/BaseQueryLogic.java
ivakegg: Done.

web-services/security/src/main/java/datawave/security/user/UserOperationsBean.java
ivakegg: Added a message attribute to the response which now get populated.

web-services/security/src/test/java/datawave/security/authorization/remote/RemoteUserOperationsImplHttpTest.java
ivakegg: done

web-services/security/src/main/java/datawave/security/authorization/remote/RemoteUserOperationsImpl.java
ivakegg: done

web-services/security/src/main/java/datawave/security/authorization/remote/ConditionalRemoteUserOperations.java
ivakegg: You are right, we do not need to create this fake response.  An empty one will do.

web-services/security/src/main/java/datawave/security/authorization/remote/ConditionalRemoteUserOperations.java
ivakegg: removed code

web-services/common/src/main/java/datawave/webservice/common/remote/RemoteUserOperations.java
ivakegg: done

web-services/security/.gitignore
ivakegg: cleaned this up.

web-services/query/src/main/java/datawave/webservice/query/logic/QueryLogic.java
ivakegg: I made this a little more generic.  I am keeping separate implementations because there are several use cases involved surrounding this concept.

web-services/query/src/main/java/datawave/webservice/query/util/LookupUUIDUtil.java
ivakegg: The interface for these has now changed.  Please reevaluate whether I got to a better place.

web-services/query/src/main/java/datawave/webservice/query/logic/BaseQueryLogic.java
ivakegg: done

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeUserOperations.java
ivakegg: I did move some of the common logic into the authorizations util however to handle merging users and principals.

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
ivakegg: This method only gets called once per create or getPlan call, so I think this is good as it stands.

warehouse/query-core/src/main/java/datawave/query/tables/RemoteEventQueryLogic.java
ivakegg: done

warehouse/query-core/src/main/java/datawave/query/tables/remote/RemoteQueryLogic.java
ivakegg: done

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
ivakegg: thanks

web-services/pom.xml
ivakegg: Actually this bump messed up some integration, so reworking

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeUserOperations.java
ivakegg: done

web-services/common-util/src/main/java/datawave/security/authorization/RemoteUserOperations.java
ivakegg: +1

warehouse/query-core/src/test/resources/datawave/query/QueryLogicFactory.xml
ivakegg: +1

web-services/security/pom.xml
ivakegg: Reverted this change.

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
```diff
@@ -64,12 +71,9 @@ public static Set<Authorizations> mergeAuthorizations(String requestedAuths, Col
         return mergedAuths;
     }
     
-    public static Set<Authorizations> getDowngradedAuthorizations(String requestedAuths, Principal principal) {
-        if (principal instanceof DatawavePrincipal) {
-            return getDowngradedAuthorizations(requestedAuths, (DatawavePrincipal) principal);
-        } else {
-            return Collections.singleton(new Authorizations());
-        }
+    public static DatawaveUser mergeAuths(DatawaveUser user1, Set<String> auths) {
```
apmoriarty: `user1` implies more than one user, just call it `user`

web-services/query/src/test/java/datawave/webservice/query/logic/composite/CompositeQueryLogicTest.java
```diff
@@ -372,9 +442,9 @@ public void setup() {
     
     @Test
     public void testClone() throws Exception {
-        List<QueryLogic<?>> logics = new ArrayList<>();
-        logics.add(new TestQueryLogic());
-        logics.add(new TestQueryLogic());
+        Map<String,QueryLogic<?>> logics = new HashMap<>();
+        logics.put("TestQueryLogic", new TestQueryLogic());
+        logics.put("TestQueryLogic2", new TestQueryLogic());
```
apmoriarty, line 447: These names should probably be static constants

web-services/security/src/main/java/datawave/security/authorization/remote/RemoteUserOperationsImpl.java
```diff
@@ -0,0 +1,83 @@
...
+    @Override
+    public GenericResponse<String> flushCachedCredentials(Object callerObject) throws AuthorizationException {
+        if (!(callerObject instanceof DatawavePrincipal)) {
+            throw new AuthorizationException("Cannot handle a " + callerObject.getClass() + ". Only DatawavePrincipal is accepted");
+        }
+        final DatawavePrincipal principal = (DatawavePrincipal) callerObject;
+        final String suffix = FLUSH_CREDS;
+        // includeRemoteServices=false to avoid any loops
+        return executeGetMethodWithRuntimeException(suffix, uriBuilder -> {
+            uriBuilder.addParameter("includeRemoteServices", "false");
```
apmoriarty: `includeRemoteServices` should probably be a static value, similar to `FLUSH_CREDS` and `LIST_EFFECTIVE_AUTHS`

apmoriarty (dismissed):
Several cosmetic/stylistic comments. Otherwise looks good.

web-services/query/src/test/java/datawave/webservice/query/logic/composite/CompositeQueryLogicTest.java
ivakegg: no need as this is a test case.

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
ivakegg: ok, will do

web-services/security/src/main/java/datawave/security/authorization/remote/RemoteUserOperationsImpl.java
ivakegg: agreed

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
```diff
@@ -81,57 +85,99 @@ public static Set<Authorizations> getDowngradedAuthorizations(String requestedAu
...
+            return Collections.singleton(new Authorizations());
+        }
+        
+        final UserAuthFunctions uaf = UserAuthFunctions.getInstance();
+        final DatawaveUser primaryUser = overallPrincipal.getPrimaryUser();
+        final DatawaveUser queryUser = queryPrincipal.getPrimaryUser();
+        
+        // validate that the query user is actually a subset of the primary user
+        if (!primaryUser.getAuths().containsAll(queryUser.getAuths())) {
+            throw new IllegalArgumentException("The user's query auths are not a subset of the user's overall auths.");
```
billoley: "The user's query auths are not a subset of the user's overall auths." seems a little too "inside baseball" with the "overall auths" language --- can we just say that the user requested authorizations that they do not have?

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
```diff
@@ -183,4 +236,72 @@ public static Collection<Authorizations> minimize(Collection<Authorizations> aut
...
+                                .collect(Collectors.toMap(DatawaveUser::getDn, Function.identity(), (e1, e2) -> e1, LinkedHashMap::new));
+                // keep track of extras
+                LinkedHashMap<SubjectIssuerDNPair,DatawaveUser> extraProxies = new LinkedHashMap<>();
+                
+                // for each user, merge or add to extras
+                for (DatawaveUser user : principal.getProxiedUsers()) {
+                    if (users.containsKey(user.getDn())) {
+                        users.put(user.getDn(), mergeUsers(users.get(user.getDn()), user));
+                    } else {
+                        extraProxies.put(user.getDn(), user);
```
billoley, line 269: Not really maintaining order of the proxiedUsers if any extra entities just get added at the end of the chain.

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
ivakegg: hmm... good point.... let me try and reword that.

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
ivakegg: yea, I was not too sure how to mix that in.  OK, I will try to maintain the order a little better.  good catch.

ivakegg: Actually this is not a concern as the only real possibility is that we have extra proxies at the end of the list.  And even if we do have a different ordering of things (a proxy inserted in the middle), then this really does not matter as long as the primary user is at the front.

billoley (approved):

apmoriarty (approved):

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
```diff
@@ -81,57 +85,99 @@ public static Set<Authorizations> getDowngradedAuthorizations(String requestedAu
...
+        // validate that the query user is actually a subset of the primary user
+        if (!primaryUser.getAuths().containsAll(queryUser.getAuths())) {
+            throw new IllegalArgumentException("System Error.  Unexpected authorization mismatch.  Please try again.");
         }
+        
+        // validate that the requestedAuths do not include anything outside of the principal's auths
+        uaf.validateRequestedAuthorizations(requestedAuths, primaryUser);
+        
+        // now return auths that are a reduced by what the query can handle.
+        return uaf.getRequestedAuthorizations(requestedAuths, queryUser, false).toString();
```
FineAndDandy, line 180: consider merging the duplicate code from this method and the previous one?

web-services/deploy/configuration/src/main/resources/datawave/query/QueryLogicFactory.xml
```diff
@@ -132,11 +132,25 @@
...
+        <property name="queryServiceHost" value="localhost" />
+        <property name="queryServicePort" value="8443" />
+        <property name="queryServiceURI" value="/DataWave/Security/User/" />
+        <property name="responseObjectFactory" ref="ResponseObjectFactory" />
+    </bean>
+
+    <!-- Configure the remote operation here for use by the local UserOperationsBean -->
+    <util:list id="RemoteUserOperationsList" value-type="datawave.security.authorization.RemoteUserOperations">
+        <!--ref bean="MyRemoteUserOps"/-->
+    </util:list>
```
FineAndDandy, line 147: I can't tell if this was commented out and forgotten or intended to be an illustration

web-services/query/src/main/java/datawave/webservice/query/logic/QueryLogic.java
```diff
@@ -379,4 +380,12 @@ default long getResultLimit(Collection<String> dns) {
...
      */
     void setPageProcessingStartTime(long pageProcessingStartTime);
+    
+    /**
+     * Normally this simply returns null as the query logic simply uses the user within the local system. However sometimes we may have a query that goes off
+     * system and hence needs to use a different set of auths when downgrading.
+     *
+     * @return A user operations interface implementation. Null if NA (i.e. the local principal is sufficient)
+     */
+    UserOperations getUserOperations();
```
FineAndDandy, line 390: How would you feel about renaming this to remoteUserOperations()? The name feels misleading. If there are no remote user operations then the local principal is sufficient.

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeUserOperations.java
```diff
@@ -0,0 +1,125 @@
...
+                    AuthorizationsListBase.SubjectIssuerDNPair dn = entry.getKey();
+                    authMap.put(dn, Sets.union(authMap.containsKey(dn) ? authMap.get(dn) : Collections.emptySet(), entry.getValue()));
+                }
+            }
+            DatawaveUser primaryUser = principal.getPrimaryUser();
+            AuthorizationsListBase.SubjectIssuerDNPair primaryDn = dn(primaryUser.getDn());
+            auths.setUserAuths(primaryDn.subjectDN, primaryDn.issuerDN, authMap.get(dn(primaryUser.getDn())));
+            authMap.entrySet().stream().filter(e -> !e.getKey().equals(primaryDn))
+                            .forEach(e -> auths.addAuths(e.getKey().subjectDN, e.getKey().issuerDN, e.getValue()));
+        }
```
FineAndDandy, line 66: should there be a warning/error/exception thrown if callerObject is not a DatawavePrincipal?

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeUserOperations.java
```diff
@@ -0,0 +1,125 @@
...
+    public GenericResponse<String> flushCachedCredentials(Object callerObject) throws AuthorizationException {
+        GenericResponse<String> response = new GenericResponse<>();
+        response.setResult("");
+        String separator = "";
+        for (UserOperations ops : userOperations) {
+            GenericResponse<String> remoteResponse = ops.flushCachedCredentials(callerObject);
+            if (remoteResponse.getHasResults()) {
+                response.setHasResults(true);
+                response.setResult(response.getResult() + separator + remoteResponse.getResult());
+                separator = "\n";
```
FineAndDandy, line 84: It isn't clear to me what the intended use of separator is here.

web-services/query/src/main/java/datawave/webservice/query/remote/RemoteQueryServiceImpl.java
```diff
@@ -92,6 +74,11 @@ public GenericResponse<String> planQuery(String queryLogicName, Map<String,List<
     }
     
     private GenericResponse<String> query(String endPoint, String queryLogicName, Map<String,List<String>> queryParameters, Object callerObject) {
+        if (!(callerObject instanceof DatawavePrincipal)) {
+            throw new RuntimeException("Cannot handle a " + callerObject.getClass() + ". Only DatawavePrincipal is accepted");
+        }
+        final DatawavePrincipal principal = (DatawavePrincipal) callerObject;
```
FineAndDandy, line 80: consider private DatawavePrincipal getPrincipal(Object callerObject)

web-services/security/src/main/java/datawave/security/authorization/remote/ConditionalRemoteUserOperations.java
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * An example may be a composite query that call a local and a remote query logic. Perhaps we can already tell that the user will not be able to get any
+ * additional authorities from the remote system and hence the remote call will not be required.
+ */
+public class ConditionalRemoteUserOperations implements UserOperations {
+    private static final Logger log = LoggerFactory.getLogger(ConditionalRemoteUserOperations.class);
+    
+    private UserOperations delegate;
+    private Function<DatawavePrincipal,Boolean> condition;
+    private ResponseObjectFactory responseObjectFactory;
```
FineAndDandy, line 28: None of these objects are null checked in this class

web-services/security/src/main/java/datawave/security/authorization/remote/RemoteUserOperationsImpl.java
```diff
@@ -0,0 +1,84 @@
...
+            super.init();
+            genericResponseReader = objectMapper.readerFor(GenericResponse.class);
+            authResponseReader = objectMapper.readerFor(responseObjectFactory.getAuthorizationsList().getClass());
+            initialized = true;
+        }
+    }
+    
+    @Override
+    public AuthorizationsListBase listEffectiveAuthorizations(Object callerObject) throws AuthorizationException {
+        if (!(callerObject instanceof DatawavePrincipal)) {
```
FineAndDandy, line 49: check initialized?

web-services/security/src/main/java/datawave/security/authorization/remote/RemoteUserOperationsImpl.java
```diff
@@ -0,0 +1,84 @@
...
+        }, () -> suffix);
+    }
+    
+    @Override
+    public GenericResponse<String> flushCachedCredentials(Object callerObject) throws AuthorizationException {
+        if (!(callerObject instanceof DatawavePrincipal)) {
+            throw new AuthorizationException("Cannot handle a " + callerObject.getClass() + ". Only DatawavePrincipal is accepted");
+        }
+        final DatawavePrincipal principal = (DatawavePrincipal) callerObject;
+        final String suffix = FLUSH_CREDS;
```
FineAndDandy, line 72: check initialized?

FineAndDandy (changes requested):
I found myself having to keep going back to review principal, queryPrincipal, remotePrincipal, and  how they all worked together to create auths. You have comments scattered around, but maybe a readme or extended class javadoc to put all the information in one place could be helpful for future maintenance

web-services/deploy/configuration/src/main/resources/datawave/query/QueryLogicFactory.xml
ivakegg: intended illustration

web-services/common-util/src/main/java/datawave/security/util/AuthorizationsUtil.java
ivakegg: good idea.

web-services/query/src/main/java/datawave/webservice/query/logic/QueryLogic.java
ivakegg: That is funny, Drew had me rename it from remote user operations to this.

ivakegg: I am leaving it as is.  maybe someday this would be non-null and yet may be local somehow.

web-services/query/src/main/java/datawave/webservice/query/remote/RemoteQueryServiceImpl.java
ivakegg: +1

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeUserOperations.java
ivakegg: +1

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeUserOperations.java
ivakegg: To separate

web-services/security/src/main/java/datawave/security/authorization/remote/ConditionalRemoteUserOperations.java
ivakegg: +1

web-services/security/src/main/java/datawave/security/authorization/remote/RemoteUserOperationsImpl.java
ivakegg: Good catch!

web-services/security/src/main/java/datawave/security/authorization/remote/RemoteUserOperationsImpl.java
ivakegg: Good catch!

PR #1830, opened 2023-01-31
Task/refactor proto handler

hlgp:
Some efficiency gains by not performing JexlEvaluation for all edge definitions on an event that doesn't even contain the necessary fields, general cleanup, etc.


warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
```diff
@@ -741,6 +609,183 @@ public long process(KEYIN key, RawRecordContainer event, Multimap<String,Normali
...
+        
+        for (String subGroup : commonKeys) {
+            for (NormalizedContentInterface ifaceSource : mSource.get(subGroup)) {
+                for (NormalizedContentInterface ifaceSink : mSink.get(subGroup)) {
+                    EdgeDataBundle edgeValue = createEdge(edgeDef, event, ifaceSource, sourceGroup, subGroup, ifaceSink, sinkGroup, subGroup);
+                    if (edgeValue != null) {
+                        
+                        // have to write out the keys as the edge values are generated, so counters get updated
+                        // and the system doesn't timeout.
+                        edgesCreated += writeEdgesForDateType(edgeValue, context, contextWriter, date_type);
```
matthpeterson: Compared with the pre-refactored code, it looks like writeEdgesForDateType adds this code which was not here before.  I wonder if that was intentional.
```
         if (dateType.equals(EdgeKey.DATE_TYPE.ACTIVITY_ONLY)) {
            edgesCreated += writeEdges(value, context, contextWriter, EdgeKey.DATE_TYPE.EVENT_ONLY);
        }
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
```diff
@@ -741,6 +609,183 @@ public long process(KEYIN key, RawRecordContainer event, Multimap<String,Normali
...
+        }
+        if (mSource.isEmpty() || mSink.isEmpty()) {
+            return true;
+        }
+        
+        return false;
+    }
+    
+    private long getActivityDate() {
+        if (normalizedFields.containsKey(edgeDefConfigs.getActivityDateField())) {
```
matthpeterson: I like that this method was extracted.  I'm concerned that the parsing may be redone though for each edge that needs it.  It might be a good idea to store the activity date as a field to avoid repeated parsing.

hlgp: Sounds good.  I like that better.  This was to get around needing to pass the date as an argument everywhere.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
```diff
@@ -741,6 +609,183 @@ public long process(KEYIN key, RawRecordContainer event, Multimap<String,Normali
...
+            if (tmp == null) {
+                tmp = HashMultimap.create();
+            }
+            tmp.put(subGroup, value);
+            depthFirstList.put(fieldName, tmp);
+            
+            normalizedFields.put(fieldName, value);
+        }
+    }
+
```
matthpeterson: I like the refactor because it improves readability of the code.  Given the amount of refactoring, it seems prudent to ensure there are unit tests that provide sufficient regression testing - to guarantee that exactly the same number and kind of edges are created as before.  I don't know that the existing tests provide that type of testing.  I recommend adding "black box" edge tests for this purpose.

Along these lines, I recommend separating the changes to functionality from the refactor changes.  It seems that the vast majority of the changes are for refactoring.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
hlgp: Yes.  I was trying to cut down on the number of arguments per function since it seemed to be out of control here (15 in some cases!).   We were repeatedly passing the two dates around and repeatedly checking their type.  The strategy for writing edges was dependent on the type, so I thought it clearer to extract some of that and update the method name to highlight this.
In the old writeEdges, it has ACTIVITY_ONLY and EVENT_ONLY written together thus:
`else {
                    edgesCreated += writeEdges(value, context, contextWriter, EdgeKey.DATE_TYPE.ACTIVITY_ONLY);
                    edgesCreated += writeEdges(value, context, contextWriter, EdgeKey.DATE_TYPE.EVENT_ONLY);`
The new code preserves this relationship in the snippet you pasted above.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
```diff
@@ -741,6 +609,183 @@ public long process(KEYIN key, RawRecordContainer event, Multimap<String,Normali
...
+            try {
+                return DateNormalizer.parseDate(actDate, DateNormalizer.FORMAT_STRINGS).getTime();
+            } catch (ParseException e1) {
+                log.error("Parse exception when getting the activity date: " + actDate + " for edge creation " + e1.getMessage());
+            }
+        }
+        return 0L;
+    }
+    
+    private String getEdgeAttr3(EdgeDefinitionConfigurationHelper edgeDefConfigs) {
```
matthpeterson: Just to reiterate, I support what you're doing here.  It's a step in the right direction.

This datatype handler is doing a lot.  I wonder if some of the newly extracted methods could be gathered into new classes to help with encapsulation and readability.  

Here's a (partial) list of concerns for the DataTypeHandler that might help guide the design of new classes:
- Extract event characteristics, e.g. dates, to be used for edges
- Generate a single EdgeDataBundle for a given event and edge definition
- Generate all the EdgeDataBundle objects for an event, according to its edge definitions 
- Functions that operate on an EdgeDataBundle, like deny list application, masking, etc.
- Create Accumulo Key(s) for EdgeDataBundle
 
Ideas for classes based on the above concerns:
1. One such class might be responsible for generating an EdgeDataBundle object for a particular EdgeDefinition.  Maybe the builder pattern would be appropriate here.

2. Another class could take a new event, extract just once its activity date, determine its date type, etc., and then loop through the edge definitions, calling the builder (item 1) for each edge definition until they are all created.

3. writeEdges could be the basis for another class that takes an EdgeDataBundle and a context writer.  It would then write the appropriate keys to the writer.  It's already using an EdgeKeyBuilder, so that part is already abstracted out.

4. The mask and deny list capabilities seem to suggest creating an interface that operates on an EdgeDataBundle and an event.  Maybe item 2 could call these interfaces.

I think the approach outlined here could address some of the comments like the one about reparsing activity date.  

A lot of the methods have already been extracted through the existing refactor, so much of this would be grouping them together into new classes.  One goal for the proposed new classes would be to have unit tests for each with minimal dependencies / integration points.  Maybe a lot of this is already part of the plan as I see it's a WIP.  Let me know if you want a hand, even if it's just adding tests or something.

hlgp: Cool ideas.  I'd rather implement them in phases, which is why I was starting with breaking blocks up into their own methods first.

hlgp: I'm just going to go ahead and implement these anyway.  They're crawling out of the woodwork so naturally.... EdgeIngestConfiguration, EdgeEventFieldInfo, etc forthcoming

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
hlgp: I added several tests to integration in this [PR](https://github.com/NationalSecurityAgency/datawave/pull/2154) to ensure that every piece of the edge remains unchanged

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
hlgp: I've broken out the classes and am pretty happy with where it landed.  There is probably more we could do, but I think this moves the ball forward significantly

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/handler/edge/EdgeHandlerTestUtil.java
```diff
@@ -1,7 +1,10 @@
 package datawave.ingest.mapreduce.handler.edge;
 
 import java.io.IOException;
+import java.util.ArrayList;
```
matthpeterson: These new imports appear to be unused.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/define/EdgeDataBundle.java
```diff
@@ -97,6 +99,17 @@ public EdgeDataBundle(EdgeDefinition edgeDef, NormalizedContentInterface ifaceSo
         this.initMarkings(getSource().getMarkings(), getSink().getMarkings());
```
matthpeterson, line 100: The constructor ` public EdgeDataBundle(EdgeDefinition edgeDef...` can be deleted now that it is unused.  Its code now seems to exist in the data type handler's createEdge.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
```diff
@@ -983,28 +623,6 @@ private NormalizedContentInterface getNullKeyedNCI(String fieldValue, Multimap<S
         return null;
```
matthpeterson, line 644: `getNullKeyedNCI` in PEDTHandler.java is now unused and can be deleted.

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/handler/edge/EdgeHandlerTestUtil.java
matthpeterson: I put this and a few other suggested removals in https://github.com/NationalSecurityAgency/datawave/pull/2269/files

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/define/EdgeDataBundle.java
matthpeterson: I put this and a few other suggested removals in https://github.com/NationalSecurityAgency/datawave/pull/2269/files

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
matthpeterson: I put this and a few other suggested removals in https://github.com/NationalSecurityAgency/datawave/pull/2269/files

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/define/EdgeDataBundle.java
```diff
@@ -539,4 +552,23 @@ public static Value getStatsLinkValue(final String realmedIdentifier) {
...
     }
 
+    public void setDateType(EdgeKey.DATE_TYPE dateType) {
+        this.dateType = dateType;
+    }
+
+    public void setEdgeDefinition(EdgeDefinition edgeDef) {
+        this.edgeDefinition = edgeDef;
+        this.edgeDirection = edgeDef.getDirection();
+        this.edgeType = edgeDef.getEdgeType().toString();
```
matthpeterson:
```suggestion
        this.edgeType = edgeDef.getEdgeType();
```

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDeleteModeTest.java
```diff
@@ -243,141 +230,12 @@ public void setup(TaskAttemptContext context) {
         public long process(KEYIN key, RawRecordContainer event, Multimap<String,NormalizedContentInterface> fields,
                         TaskInputOutputContext<KEYIN,? extends RawRecordContainer,KEYOUT,VALUEOUT> context, ContextWriter<KEYOUT,VALUEOUT> contextWriter)
                         throws IOException, InterruptedException {
-            long edgesCreated = 0;
```
matthpeterson, line 246: `this.deleteMode` (above) is never used.  I'm wondering if MyDerivedProtobufEdgeDataTypeHandler could be removed.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
matthpeterson: Thanks for adding these tests!

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
```diff
@@ -421,141 +169,44 @@ public void funnel(Key from, PrimitiveSink into) {
...
 
+        if (shortCircuit(event, typeName, edgeConfig.getEdges())) {
+            return edgesCreated;
         }
-        edgeDefConfigs = edges.get(typeName);
-        edgeDefs = edgeDefConfigs.getEdges();
 
-        trimFieldGroup = context.getConfiguration().getBoolean(typeName + TRIM_FIELD_GROUP, false);
+        // Track metadata for this event
+        eventMetadataRegistry = new HashMap<>();
```
matthpeterson: I suggest initializing this at its declaration and clearing the Map each time process is called, to avoid having to create a new HashMap in each call to process.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
```diff
@@ -421,141 +169,44 @@ public void funnel(Key from, PrimitiveSink into) {
...
+        if (edgeConfig.evaluatePreconditions()) {
             setupPreconditionEvaluation(fields);
-
         }
 
-        // Get the load date of the event from the fields map
-        loadDateStr = getLoadDateString(fields);
-
+        boolean trimGroups = context.getConfiguration().getBoolean(typeName + EdgeIngestConfiguration.TRIM_FIELD_GROUP, false);
+        EdgeEventFieldUtil edgeEventFieldUtil = new EdgeEventFieldUtil(trimGroups);
```
matthpeterson, line 195: Also to avoid object creation and collection, I would consider creating a reusable EdgeEventFieldUtil and clearing it here instead of creating a new one.

matthpeterson: Perhaps this suggestion is overkill.  Either way, with the refactor it is probably worth observing performance impact.  It would be cool if we had a JMH test handy for the datatype handler that would give us microbenchmarking information for process, given that it's called so much.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
```diff
@@ -620,160 +259,194 @@ public long process(KEYIN key, RawRecordContainer event, Multimap<String,Normali
...
+
+        } // end edge defs
+
+        if (this.enableMetadata) {
+            writeMetadataMap(context, contextWriter, eventMetadataRegistry);
+        }
+        postProcessEdges(event, context, contextWriter, edgesCreated, baseEdgeBundle);
+
+        return edgesCreated;
+    }
```
matthpeterson, line 347: Much cleaner, thank you!

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/EdgeEventFieldUtil.java
```diff
@@ -0,0 +1,241 @@
...
+                // also need to write EVENT_ONLY for some reason
+            }
+        } else {
+            dateType = EdgeKey.DATE_TYPE.EVENT_ONLY;
+        }
+        bundle.setDateType(dateType);
+    }
+
+    public EdgeDataBundle setEdgeInfoFromEventFields(EdgeDataBundle bundle, EdgeDefinitionConfigurationHelper edgeDefConfigs, RawRecordContainer event,
+                    EdgeIngestConfiguration edgeConfig, long newFormatStartDate, Configuration conf, String typeName) {
```
matthpeterson: `typeName` is unused.  Perhaps it should be?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/EdgeEventFieldUtil.java
```diff
@@ -0,0 +1,241 @@
...
+                dateType = EdgeKey.DATE_TYPE.ACTIVITY_ONLY;
+                // also need to write EVENT_ONLY for some reason
+            }
+        } else {
+            dateType = EdgeKey.DATE_TYPE.EVENT_ONLY;
+        }
+        bundle.setDateType(dateType);
+    }
+
+    public EdgeDataBundle setEdgeInfoFromEventFields(EdgeDataBundle bundle, EdgeDefinitionConfigurationHelper edgeDefConfigs, RawRecordContainer event,
```
matthpeterson: This method returns the bundle it received by reference.  The return value itself is never used.  Perhaps the method should just create a new bundle from the provided event and drop the first parameter.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/EdgeEventFieldUtil.java
```diff
@@ -0,0 +1,241 @@
...
+        setEdgeKeyDateType(bundle, validActivityDate, activityEqualsEvent, event.getDate(), newFormatStartDate);
+
+        bundle.setLoadDate(loadDateStr);
+        bundle.setActivityDate(activityDate);
+        bundle.setEventDate(event.getDate());
+        bundle.setValidActivityDate(validActivityDate);
+        bundle.setEdgeAttribute2(getEdgeAttr2(edgeDefConfigs));
+        bundle.setEdgeAttribute3(getEdgeAttr3(edgeDefConfigs));
+        bundle.setRequiresMasking(event.isRequiresMasking());
+        bundle.setHelper(event.getDataType().getIngestHelper(conf));
```
matthpeterson: Looking for opportunities to drop more method parameters - it looks like the `conf` parameter is only used here.  I suggest updating EdgeIngestConfiguration to save its constructor argument (which is the same as this conf, retrieved from context.getConfiguration()), so that it's only necessary to use EdgeIngestConfiguration.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/EdgeEventFieldUtil.java
```diff
@@ -0,0 +1,241 @@
...
+import datawave.ingest.data.config.NormalizedContentInterface;
+import datawave.ingest.data.normalizer.SimpleGroupFieldNameParser;
+import datawave.ingest.mapreduce.EventMapper;
+import datawave.ingest.mapreduce.handler.edge.define.EdgeDataBundle;
+import datawave.ingest.mapreduce.handler.edge.define.EdgeDefinition;
+import datawave.ingest.mapreduce.handler.edge.define.EdgeDefinitionConfigurationHelper;
+import datawave.ingest.time.Now;
+import datawave.util.time.DateHelper;
+
+public class EdgeEventFieldUtil {
```
matthpeterson, line 29: I really like this class.  I wonder if the name could be improved, mostly because Util raises some flags in my mind - but naming is hard so honestly I'm not sure what's better.  Its primary behavior is to extract edge information from fields.  So maybe FieldToEdgeDataGenerator or EdgeFieldExtractor.  Feel free to ignore.

matthpeterson (approved):

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
```diff
@@ -805,134 +476,101 @@ private void setupPreconditionEvaluation(Multimap<String,NormalizedContentInterf
...
-                if (null != elnci) {
-                    edgeDataBundle.initDuration(elnci);
-                }
-            }
-        }
+        EdgeDefinition edgeDef = edgeDataBundle.getEdgeDefinition();
+        // Get the edge value
 
-        String typeName = getTypeName(event.getDataType());
+        VertexValue source = new VertexValue(edgeDef.isUseRealm(), edgeDef.getSourceIndexedFieldRealm(), edgeDef.getSourceEventFieldRealm(),
```
apmoriarty, line 489: For something like this consider just passing in the edge definition and allow the constructor to pull out what it needs.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/EdgeIngestConfiguration.java
```diff
@@ -0,0 +1,384 @@
...
+        try {
+            ctx = new ClassPathXmlApplicationContext(ProtobufEdgeDataTypeHandler.class.getClassLoader().getResource(springConfigFile).toString());
+
+            log.info("Got config on first try!");
+        } catch (Exception e) {
+            log.error("Problem getting config for ProtobufEdgeDataTypeHandler: {}", e);
+            throw e;
+        }
+
+        Assert.notNull(ctx);
```
apmoriarty, line 117: Precondition instead of spring util?

hlgp: This code was moved from ProtobufEdgeDataTypeHandler as is.  I am trying my best to just reorganize without changing functionality in this MR.

apmoriarty (dismissed):

hlgp:
@gemini-cli /review

PR #1836, opened 2023-02-02
Improve tests for TermFrequencyIndexIterator

apmoriarty:
Added several integration style tests to the TermFrequencyIndexIteratorTest

- 94c106746619e5a4f00b5d9c03fb7d9f30e1d860 -- clean up junit asserts
- 94469e876ae21505c3804666d796df81dead8c04 -- add new tests


mineralntl (dismissed):

warehouse/query-core/src/test/java/datawave/query/iterator/logic/TermFrequencyIndexIteratorTest.java
```diff
@@ -337,23 +349,132 @@ public void testScanFullRangeExclusiveEventDataQueryExpressionFilter() throws IO
...
+        iter.seek(new Range(), Collections.emptyList(), false);
+        
+        int i = 0;
+        while (iter.hasNext()) {
+            Key key = iter.next();
+            assertEquals("received " + key.toStringNoTime(), "datatype\u00001", key.getColumnFamily().toString());
+            i++;
+        }
+        
+        assertEquals("should have one entry but got " + i, 1, i);
```
magrable: Maybe change this to something that obviously fails the test?

warehouse/query-core/src/test/java/datawave/query/iterator/logic/TermFrequencyIndexIteratorTest.java
```diff
@@ -337,23 +349,132 @@ public void testScanFullRangeExclusiveEventDataQueryExpressionFilter() throws IO
...
```
magrable: Maybe change this to something that obviously fails the test?

warehouse/query-core/src/test/java/datawave/query/iterator/logic/TermFrequencyIndexIteratorTest.java
```diff
@@ -337,23 +349,132 @@ public void testScanFullRangeExclusiveEventDataQueryExpressionFilter() throws IO
...
```
magrable: Following up on our conversation - shouldn't this also return no entries because there is no "FIELD=value" entry in the source TF rows?  It looks like its just returning the first entry in source after the start key

magrable (changes requested):

warehouse/query-core/src/test/java/datawave/query/iterator/logic/TermFrequencyIndexIteratorTest.java
apmoriarty: Changed this to `fail()` in the next commit

warehouse/query-core/src/test/java/datawave/query/iterator/logic/TermFrequencyIndexIteratorTest.java
apmoriarty: Changed this to `fail()` in the next commit

warehouse/query-core/src/test/java/datawave/query/iterator/logic/TermFrequencyIndexIteratorTest.java
apmoriarty: This is aggregating all matching uids into a `datatype\0uid`, so for the given dataset we can expect one entry only.

alerman (approved):

magrable (approved):

PR #1838, opened 2023-02-02
Utility for generating test PKI materials for SSL testing

drewfarris:


jwomeara (dismissed):

keith-ratcliffe (dismissed):
Nice!

ivakegg (approved):

ivakegg:
My only comment is that "doit.sh" could be better named, but good otherwise.

jwomeara (approved):
I think I agree with @ivakegg about the `doit.sh` script, but otherwise looks good.  I'll mark as integration tested so it can be merged.

PR #1842, opened 2023-02-03
Updated the unique transform to create a simpler but much more efficient

ivakegg:
algorithm for determining uniqueness


warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -159,186 +153,55 @@ private boolean isDuplicate(Document document) throws IOException {
...
-            for (Map.Entry<String,String> entry : fieldSet.entrySet()) {
-                output.writeChars(separator);
-                output.writeChars(entry.getKey());
-                output.writeChar('=');
-                output.writeChars(entry.getValue());
-                separator = ",";
-            }
-        }
-        output.flush();
+        outputSortedFieldValues(document, output);
```
cawaring, line 160: At this point the baos that output is writing to contains the fieldnames and all values that are used for unique.  So this is on order || all values and their fieldnames ||, which I believe now is just 1 extra copy of all those values which looks like a ~2x reduction over the last impl.  

The byte array from baos will be used as input to a bloom filter for unique'n.  I believe the bloom is ~ 5MB in size.  This brought up a few thoughts.  Since this function is running in the QueryIterator on the tabletserver(s) it will get torn down with its iterator. Triggered by the scan memory limit. How many times do we think the bloom is actually being used if we are concerned about memory efficiency?  Maybe once, worst case?  If so do we even want to do all this past a certain point and just punt to when this gets run on the webserver side.   This avoid copying the values, sorting them, and populating the baos.  Alternatively, we could generate a sketch of the fields and use that as input to the bloom.   This would avoid the sorts and generating a large byte array.  It would provide an upper bound on the amount of memory required by this function.  This approach is already probabilistic in nature, so maybe worth considering?

ivakegg: I agree that a sketch of the fields would be a great additional performance enhancement here.  What I have done here already reduces the accuracy/efficacy of the baos being generated.  What I would need to understand is how more more error we would be introducing into the algorithm.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GroupingRequiredFilterFunctions.java
```diff
@@ -87,14 +85,14 @@ public static Collection<?> getGroupsForMatchesInGroup(Object... args) {
                     }
                 }
             }
-        }
-        // if there was a match found at all levels, then the matches.size will be equal to the
-        // number of field/regex pairs
-        if (args.length == 2) { // maybe they passed in only 2 args, in that case, get the groups for whatever was in the first (only) arg pair
-            allMatches.addAll(leftSideMatches);
```
ivakegg, line 94: allMatches is not returned so no need to update that set here.

PR #1843, opened 2023-02-06
Add FieldIndexKey parser and test, integrated into IteratorBuildingVisitor

apmoriarty:
This build will fail until https://github.com/NationalSecurityAgency/datawave/pull/1841 is merged

The FunctionFilter has two major improvements
- the JexlContext is now reused instead of creating a new one for each key traversed
- parsing the field and value is now done via a key parser built for FieldIndex keys


warehouse/query-core/src/main/java/datawave/query/data/parsers/FieldIndexKey.java
```diff
@@ -0,0 +1,106 @@
...
+ * A reusable, dedicated key parser for Field Index keys
+ * <p>
+ * A Field Index key's components are
+ * <ul>
+ * <li>row - the shard</li>
+ * <li>column family - "fi\0FIELD_NAME"</li>
+ * <li>column qualifier - "value\0datatype\0uid"</li>
+ * </ul>
+ */
+public class FieldIndexKey {
```
hlgp, line 16: I'd rather see schematic classes like this somewhere common to ingest and query, but we can always move it later.

apmoriarty: Sounds like a plan

FineAndDandy (dismissed):

warehouse/query-core/src/main/java/datawave/query/data/parsers/FieldIndexKey.java
```diff
@@ -0,0 +1,106 @@
+package datawave.query.data.parsers;
```
drewfarris, line 1: I like this - especially how it avoids doing unnecessary or redundant work. Consider how this will degrade in an state where the parse of the key fails in either `traverseColumnQualifier` or calls to any of the `get` methods. You could add a 'parseError' flag, for instance, that gets set when an error is encountered and used to guide the response of the various `get` methods. 

Depending on the scope of reuse, it also might be helpful to consider implementing a 'reset' method or 'clear' method to call when we are done so this doesn't hold onto a reference to the last key parsed indefinitely.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
```diff
@@ -1327,19 +1332,11 @@ public FunctionFilter(List<ASTFunctionNode> nodes) {
...
-            int index = fieldValue.lastIndexOf('\0');
-            index = fieldValue.lastIndexOf('\0', index - 1);
-            fieldValue = fieldValue.substring(0, index);
+            parser.parse(k);
             
-            // create a jexl context with this valud
-            JexlContext context = new DatawaveJexlContext();
-            context.set(fieldName, new ValueTuple(fieldName, fieldValue, fieldValue, null));
+            context.clear();
+            context.set(parser.getField(), new ValueTuple(parser.getField(), parser.getValue(), parser.getValue(), null));
```
drewfarris: It could be marginally faster to assign the results of `getField` and `getValue` to `final String` as opposed to calling these methods multiple times (e.g., skip the function lookup and the branch checks in the functions themselves)

warehouse/query-core/src/test/java/datawave/query/data/parsers/FieldIndexKeyTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.accumulo.core.data.Key;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class FieldIndexKeyTest {
+    
+    private final Key fiKey = new Key("row", "fi\0FIELD", "value\0datatype\0uid");
+    private final Key fiKeyWithNullsInValue = new Key("row", "fi\0FIELD", "v\0al\0ue\0datatype\0uid");
+    private final Key fiKeyWithChildUid = new Key("row", "fi\0FIELD", "value\0datatype\0uid.12.37");
```
drewfarris, line 14: Consider adding some just plain wrong keys and testing error conditions.

apmoriarty: Added several error cases

warehouse/query-core/src/main/java/datawave/query/data/parsers/FieldIndexKey.java
apmoriarty: Spent some time reworking the whole class. 
- the `traverseColumnQualifier()` method will only be called once
- getters will return null values if the backing array is invalid
- added a `clearState()` method

warehouse/query-core/src/main/java/datawave/query/data/parsers/FieldIndexKey.java
```diff
@@ -0,0 +1,116 @@
...
+            if (backing.length() > 3) {
+                field = backing.subSequence(3, backing.length()).toString();
+            }
+        }
+        return field;
+    }
+    
+    public String getValue() {
+        if (value == null) {
+            if (cqBytes == null) {
```
ivakegg, line 85: could make this check inside the traverse method instead

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
ivakegg: +1

FineAndDandy (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
apmoriarty: Done

ivakegg (approved):

warehouse/query-core/src/main/java/datawave/query/data/parsers/FieldIndexKey.java
```diff
@@ -0,0 +1,116 @@
...
+                traverseColumnQualifier();
+            }
+            if (firstNull != -1 && secondNull != -1) {
+                value = cqBytes.subSequence(0, firstNull).toString();
+            }
```
ivakegg, line 92: Is this an exception if we cannot pull out the value?

warehouse/query-core/src/main/java/datawave/query/data/parsers/FieldIndexKey.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+    }
+    
+    public String getField() {
+        if (field == null) {
+            ByteSequence backing = key.getColumnFamilyData();
+            if (backing.length() > 3) {
+                field = backing.subSequence(3, backing.length()).toString();
+            }
```
ivakegg, line 78: Exception case if we cannot pull out the field?

warehouse/query-core/src/main/java/datawave/query/data/parsers/FieldIndexKey.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+    
+    public String getDatatype() {
+        if (datatype == null) {
+            if (cqBytes == null) {
+                traverseColumnQualifier();
+            }
+            if (firstNull != -1 && secondNull != -1) {
+                datatype = cqBytes.subSequence(firstNull + 1, secondNull).toString();
+            }
```
ivakegg, line 106: Is this an exception if we cannout pull out the datatype

warehouse/query-core/src/main/java/datawave/query/data/parsers/FieldIndexKey.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+    
+    public String getUid() {
+        if (uid == null) {
+            if (cqBytes == null) {
+                traverseColumnQualifier();
+            }
+            if (firstNull != -1 && secondNull != -1) {
+                uid = cqBytes.subSequence(secondNull + 1, cqBytes.length()).toString();
+            }
```
ivakegg, line 120: Is this an exception case if we cannot pull out the uid?

warehouse/query-core/src/main/java/datawave/query/data/parsers/FieldIndexKey.java
```diff
@@ -0,0 +1,116 @@
...
+    
+    /**
+     * Sets the key and resets all supporting objects
+     *
+     * @param k
+     *            a field index key
+     */
+    public void parse(Key k) {
+        this.key = k;
+        clearState();
```
ivakegg, line 38: Should we be validating this is an fi key?

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/data/parsers/FieldIndexKey.java
apmoriarty: The FieldIndexKey contract assumes it is operating on FI keys, that is, it is up to the caller to verify all keys passing through the parser are formatted properly. Where the parser is integrated into the IteratorBuildingVisitor (and by extension the ivarators) all of those ranges are bounded to the FI. 

Minimal effort is made to validate proper key structure. If an invalid key structure is detected, a null value is returned. If the caller blindly trusts the parser, this will turn into a null pointer exception.

FineAndDandy (dismissed):

warehouse/query-core/src/main/java/datawave/query/data/parsers/FieldIndexKey.java
apmoriarty: Done

warehouse/query-core/src/main/java/datawave/query/data/parsers/FieldIndexKey.java
apmoriarty: Done

warehouse/query-core/src/main/java/datawave/query/data/parsers/FieldIndexKey.java
apmoriarty: Done

warehouse/query-core/src/main/java/datawave/query/data/parsers/FieldIndexKey.java
apmoriarty: Done

warehouse/query-core/src/main/java/datawave/query/data/parsers/FieldIndexKey.java
apmoriarty: Done

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
```diff
@@ -1327,19 +1329,14 @@ public FunctionFilter(List<ASTFunctionNode> nodes) {
...
+            parser.parse(k);
+            
+            final String field = parser.getField();
+            final String value = parser.getValue();
             
-            // create a jexl context with this valud
-            JexlContext context = new DatawaveJexlContext();
-            context.set(fieldName, new ValueTuple(fieldName, fieldValue, fieldValue, null));
+            context.clear();
+            context.set(field, new ValueTuple(field, value, value, null));
```
drewfarris, line 1339: Consider calling `parser.clearState()` here so we don't hang on to the `Key` inside the parser now that we're done with it`

drewfarris: Ahh, nevermind - parser.clearState() doesn't remove the reference to the key. Maybe we need something else that does that?

drewfarris (approved):

FineAndDandy (approved):

ivakegg (approved):

PR #1845, opened 2023-02-09
TLD Index Iterator no longer calls next 256 times

apmoriarty:
Summary of changes
- added TFKeyUtil, patterned after FIKeyUtil
- cleaned up sonar/intellij warnings for TermFrequencyIndexIterator
- removed loop that would call next 256 times from TermFrequencyIndexIterator
- replaced usage of DatawaveKey with dedicated TFKey


warehouse/query-core/src/main/java/datawave/core/iterators/key/util/TFKeyUtil.java
```diff
@@ -0,0 +1,124 @@
...
+     * 
+     * @param k
+     *            the key
+     * @return the field
+     */
+    public static String getFieldString(Key k) {
+        ByteSequence bytes = k.getColumnQualifierData();
+        for (int i = bytes.length() - 1; i >= 0; i--) {
+            if (bytes.byteAt(i) == 0x00) {
+                return bytes.subSequence(i + 1, bytes.length()).toString();
```
drewfarris, line 42: We have some funny error cases in this class that likely existed in the class you based this on, so it might be OK to leave them. That said it is worth considering whether something different is needed.

For example, if for some reason the offsets when we call `subSequence` are wrong, we'll throw some sort of exception likely, but if we just can't find anything we return `null`. Should we consider treating all failure cases the same?  Not certain whether returning null or throwing an exception is more ideal.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyIndexIterator.java
```diff
@@ -111,8 +121,6 @@ public TermFrequencyIndexIterator(Key fiStartKey, boolean startInclusive, Key fi
         
         this.includeColumnFamilies = true;
         
-        this.document = new Document();
```
drewfarris, line 114: I suspect that this instance of `Document` was created here so that the public `document()` method never returns `null`. After this change, the `document()` method will return `null` if called prior to the `next()` method. 

As a result, we might want to leave it in place unless you have a really clear idea about everywhere (and when) that method is being called. Although not ideal, the overhead of a single document object is likely negligible.

apmoriarty: What do you think about leaving this code as is and modifying the document method to look like this?
```
public Document document(){
  if(document == null){
    document = new Document()
  }
  return document;
}
```

FineAndDandy (approved):

drewfarris (approved):

PR #1846, opened 2023-02-09
Javadoc annotation fixes for metrics core

avgAGB:


apmoriarty (approved):

warehouse/metrics-core/src/main/java/datawave/metrics/analytic/MetricsCorrelatorReducer.java
```diff
@@ -106,6 +106,10 @@ public static LongArrayWritable createSummary(Iterable<LongArrayWritable> flows,
     /**
      * Utility method for finding the median in a sorted set.
      * 
+     * @param s
+     *            the set to search
+     * @param <T>
```
matthpeterson, line 111: I think these two param entries should be combined because otherwise it appears as though there are two distinct parameters.  See https://stackoverflow.com/questions/2015972/is-there-a-javadoc-tag-for-documenting-generic-type-parameters

matthpeterson (approved):

PR #1849, opened 2023-02-11
Updated the ivarator mechanisms to cleanup directories

ivakegg:
Updated the ivarator mechanisms to cleanup directories when they are done.  This has resulted in collapsing the directory structure.  However since they are now deleted when done we should be able to avoid creating too many subdirectories.


warehouse/query-core/src/main/java/datawave/query/iterator/ivarator/IvaratorCacheDirConfig.java
```diff
@@ -34,6 +34,9 @@ public class IvaratorCacheDirConfig {
     // the minimum percent of available storage required to use this filesystem
     final protected double minAvailableStoragePercent;
     
+    // To we require the ivarators to cleanup after completion (true by default)
```
jschmidt10:
```suggestion
    // Do we require the ivarators to cleanup after completion (true by default)
```

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -1070,6 +1099,9 @@ protected boolean addKey(Key topFiKey, Value value) throws IOException {
      * This method will asynchronously fill the set with matches from within the specified bounding FI range.
      * 
      * @param boundingFiRange
+     *            The range to fill from
```
apmoriarty, line 1257: "The field index scan range" ?

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -1208,7 +1240,9 @@ protected Future<?> fillSet(final Range boundingFiRange, final TotalResults tota
      * Get the unique directory for a specific row
      *
      * @param uniqueDir
+     *            The unique directory
```
apmoriarty, line 1294: Is uniqueness actually guaranteed? It seems like if multiple ivarator queries are run without a query id they will re-use integer-based directory names.

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -1217,22 +1251,39 @@ protected Path getRowDir(Path uniqueDir, String row) {
...
+            }
+        }
+    }
+    
     /**
      * This will setup the set for the specified range. This will attempt to reuse precomputed and persisted sets if we are allowed to.
      * 
      * @param row
-     * @throws IOException
+     *            The row for which to setup a backed set
```
apmoriarty, line 1431: "The field index scan range that requires a hdfs-backed set"?

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -1427,7 +1476,9 @@ protected String getOwnerId(Object owner) {
         
         public void takeOwnership(String row, Object owner) throws IOException {
             Path file = getOwnershipFile(row);
+            log.warn("Creating ownership file for " + controlDir);
             writeFile(file, getOwnerId(owner).getBytes());
+            log.warn("Created ownership file for " + controlDir);
```
apmoriarty: Do we have a sense for how noisy these log statements will be?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
```diff
@@ -1047,21 +1047,22 @@ public Object visit(ASTNumberLiteral node, Object o) {
...
         
-        // first lets increment the count for a unique subdirectory
-        String subdirectory = ivaratorCacheSubDirPrefix + "term" + Integer.toString(++ivaratorCount);
+        StringBuilder subDirectoryBuilder = new StringBuilder();
+        subDirectoryBuilder.append(queryId);
+        if (scanId == null) {
+            log.warn("Running query iterator for " + queryId + " without a scan id.  This could cause ivarator directory conflicts.");
+        } else {
+            subDirectoryBuilder.append('_').append(scanId);
+        }
```
apmoriarty, line 1062: This seems wrong to rely solely on an integer count for directory uniqueness. I don't have a sense for how frequent this is, but there might be a better way to guard against this case.

Maybe if the scanId is null, this could fall back to a date-based id? For example: yyyyMMddhhmmSS

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
ivakegg: yea, I should probably remove these

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
ivakegg: I cannot fall back to a date based thing.  The whole point of these directories is that they will persist between tearing down and rebuilding the iterators, so I need to be able to recreate the same directory name.  That being said, maybe a fallback could be the tserver host name or something like that.

apmoriarty: ah, I see. Then I have no issues with the code as is

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
ivakegg: See the IteratorBuildingVisitor which will set this up to be unique.

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
ivakegg: yes

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
ivakegg: This will be the shard id which is the row id.

warehouse/query-core/src/main/java/datawave/query/iterator/builder/IvaratorBuilder.java
```diff
@@ -44,10 +45,20 @@ protected void validateIvaratorControlDir(IvaratorCacheDir ivaratorCacheDir) {
...
         try {
-            hdfsCacheURI = new URI(ivaratorCacheDirURI);
-            hdfsFileSystem.mkdirs(new Path(hdfsCacheURI));
+            final Path hdfsCachePath = new Path(new URI(ivaratorCacheDirURI));
+            // get the parent directory
+            final Path parentCachePath = hdfsCachePath.getParent();
+            final URI parentURI = parentCachePath.toUri();
+            if (!hdfsFileSystem.exists(parentCachePath)) {
+                // being able to make the parent directory is proof enough
+                hdfsFileSystem.mkdirs(parentCachePath);
```
drewfarris: Consider checking the return value to verify that the directory was successfully created?

warehouse/query-core/src/main/java/datawave/query/iterator/builder/IvaratorBuilder.java
```diff
@@ -44,10 +45,20 @@ protected void validateIvaratorControlDir(IvaratorCacheDir ivaratorCacheDir) {
...
+            } else if (parentURI.getScheme().equals("file")) {
+                File parent = new File(parentURI.getPath());
+                if (!parent.canWrite() || !parent.canRead()) {
+                    throw new IllegalStateException("Invalid permissions to directory " + parentCachePath);
+                }
+            }
```
drewfarris, line 65: There's the case where the object exists on hdfs, but is not a directory. It's unlikely, but consider double-checking here?

nonessentialprototype (dismissed):

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -1236,25 +1260,39 @@ protected Path getRowDir(Path uniqueDir, String row) {
...
+     *
+     */
+    protected void cleanup() {
+        // ensure all threads know we are done
+        this.setControl.setCancelled();
+        clearRowBasedHdfsBackedSet();
+        if (this.requiresCleanup) {
+            try {
+                log.warn("Deleting " + this.controlDir);
+                this.controlFs.delete(this.controlDir, true);
```
ivakegg: This will not work if we have multiple ivarator cache dirs.... probably need to clean up this directory on each of them.

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -506,6 +512,10 @@ public void init(SortedKeyValueIterator<Key,Value> source, Map<String,String> op
     
     @Override
     protected void finalize() throws Throwable {
+        // NOTE: We cannot actually delete the directory here because it could be the case
+        // that this instance if being finalized but the scan may be rebuilt which would
```
jwomeara:
```suggestion
        // that this instance is being finalized but the scan may be rebuilt which would
```

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
jwomeara: The control directory is only created on the first ivarator cache dir, so maybe you're fine?

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -1274,15 +1312,6 @@ protected void setupRowBasedHdfsBackedSet(String row) throws IOException {
...
             
-            // ensure the control directory is created
-            Path controlRowDir = getRowDir(this.controlDir, row);
-            if (!this.controlFs.exists(controlRowDir)) {
-                this.controlFs.mkdirs(controlRowDir);
-                this.createdRowDir = true;
-            } else {
-                this.createdRowDir = false;
-            }
-
```
jwomeara: Uhhh... if we're not creating the control dir anymore, do we have anything to cleanup?  Are we creating the control dir in multiple places?

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -1503,7 +1538,7 @@ public boolean isCancelledQuery() {
                     synchronized (this) {
                         // now recheck the cancelled flag and timeout to ensure we really need to make the hdfs calls
                         if (!cancelled && ((now - lastCancelledCheck) > CANCELLED_CHECK_INTERVAL)) {
-                            cancelled = !queryLock.isQueryRunning();
+                            cancelled |= !queryLock.isQueryRunning();
```
jwomeara: if you get to this statement, `cancelled` is already known to be false.  no need for `|=` in that case.

warehouse/query-core/src/main/java/datawave/query/iterator/builder/IvaratorBuilder.java
```diff
@@ -44,10 +45,20 @@ protected void validateIvaratorControlDir(IvaratorCacheDir ivaratorCacheDir) {
...
-            hdfsCacheURI = new URI(ivaratorCacheDirURI);
-            hdfsFileSystem.mkdirs(new Path(hdfsCacheURI));
+            final Path hdfsCachePath = new Path(new URI(ivaratorCacheDirURI));
+            // get the parent directory
+            final Path parentCachePath = hdfsCachePath.getParent();
+            final URI parentURI = parentCachePath.toUri();
+            if (!hdfsFileSystem.exists(parentCachePath)) {
+                // being able to make the parent directory is proof enough
+                hdfsFileSystem.mkdirs(parentCachePath);
+            } else if (parentURI.getScheme().equals("file")) {
```
jwomeara, line 60: So, if the parent exists in hdfs, we just assume that the directory is valid?  The previous logic would attempt to make a directory whether it's a local filesystem or hdfs - this doesn't.

warehouse/query-core/src/main/java/datawave/query/iterator/ivarator/IvaratorCacheDirConfig.java
```diff
@@ -34,6 +34,9 @@ public class IvaratorCacheDirConfig {
     // the minimum percent of available storage required to use this filesystem
     final protected double minAvailableStoragePercent;
     
+    // Do we require the ivarators to cleanup after completion (true by default)
+    final protected boolean requiresCleanup;
```
jwomeara, line 41: Are we ever realistically going to set this to false?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
```diff
@@ -1058,21 +1058,22 @@ public Object visit(ASTNumberLiteral node, Object o) {
...
+        // and lets increment the count for a unique subdirectory
+        subDirectoryBuilder.append('_').append(ivaratorCacheSubDirPrefix).append("term").append(Integer.toString(++ivaratorCount));
```
jwomeara: Integer.toString is unnecessary.

jwomeara: So the overall goal here is just to flatten out the directory structure?

jwomeara (changes requested):
Some minor comments included.

warehouse/query-core/src/main/java/datawave/query/iterator/builder/IvaratorBuilder.java
ivakegg: The directory could have already existed in which case it would return false.  Only and exception is a problem here.

warehouse/query-core/src/main/java/datawave/query/iterator/builder/IvaratorBuilder.java
ivakegg: done

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
ivakegg: The control dir is only the first one.  The other directories may have data in them and should be purged as well.

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
ivakegg: The QueryIterator is not creating all of the dirs as part of its verification of valid cache dirs.

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
ivakegg: woops

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
ivakegg: There is no need to precreate the control directory here, because when we create the ownership file, the parent directories will be created automatically.

warehouse/query-core/src/main/java/datawave/query/iterator/builder/IvaratorBuilder.java
ivakegg: This says hdfsFileSystem but that file systgem could also be a local file system.  I will rename the variable.
That bein said, the parent directory will be created on line 55 regardless.  The subsequent check for the canWrite and canRead can only really be done on the local file system.  The hdfs file system does not support such checks.

warehouse/query-core/src/main/java/datawave/query/iterator/ivarator/IvaratorCacheDirConfig.java
ivakegg: I wanted to leave the possibility of turning if off in case we could the code was deleting stuff if should not be deleting.  This might allow us to keep the system running until the bug is uncovered.  Just trying to be safe.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
ivakegg: true

ivakegg: Yes

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
jwomeara: Yup, I agree.  If it's time to cleanup, gotta hit all the ivarator cache dirs.

warehouse/query-core/src/main/java/datawave/query/iterator/builder/IvaratorBuilder.java
jwomeara: line 55 won't be hit at all if the parent directory already exists...  that wasn't really my concern.  I was curious why we would perform validation for a local filesystem, but not for hdfs.  Your comments about canWrite and canRead explain that.

However, comparing the new logic with the old, are we doing enough to validate hdfs?  If the parent directory already exists in the hdfs case does that guarantee that we will be able to create the ivarator cache directory?  In the past we tried to make the ivarator cache directory (local or hdfs).  Just wondering if this is equivalent to the previous code, and doing enough to validate for hdfs.

warehouse/query-core/src/main/java/datawave/query/iterator/builder/IvaratorBuilder.java
```diff
@@ -44,10 +45,22 @@ protected void validateIvaratorControlDir(IvaratorCacheDir ivaratorCacheDir) {
...
-            hdfsFileSystem.mkdirs(new Path(hdfsCacheURI));
+            final Path hdfsCachePath = new Path(new URI(ivaratorCacheDirURI));
+            // get the parent directory
+            final Path parentCachePath = hdfsCachePath.getParent();
+            final URI parentURI = parentCachePath.toUri();
+            if (!hdfsFileSystem.exists(parentCachePath)) {
+                // being able to make the parent directory is proof enough
+                hdfsFileSystem.mkdirs(parentCachePath);
+            } else if (hdfsFileSystem.getFileStatus(parentCachePath).isFile()) {
+                throw new IOException(parentCachePath.toString() + " exists but is a file.  Expecting directory");
```
jwomeara: toString is redundant.

jwomeara (changes requested):

warehouse/query-core/src/main/java/datawave/query/iterator/builder/IvaratorBuilder.java
ivakegg: OK, I guess I can go back to creating the directory as this one must be created.

ivakegg: Ah, now I figured out why I did this.  The problem is that I do not want the directory to be created if the ivarator never actually finds anything (the initial range seek comes up empty).  In that case the directory would be created and then immediately removed which seems wasteful.

ivakegg: However creating the directory is finding some holes where I need to cleanup the directory so working through those.

ivakegg: Cleanup those holes, but now I have the case where the ivarator never actually runs because an error earlier in the field index lookup.  So I am back to not creating the directory here.

jwomeara:
@ivakegg are you done making changes?  i can't tell looking at your latest comments.

ivakegg:
> @ivakegg are you done making changes? i can't tell looking at your latest comments.

I think I am.

jwomeara (dismissed):

PR #1850, opened 2023-02-13
Make jexlasthelper more efficient

### LINKED ISSUE #1839: Make JexlASTHelper more efficient
Within the `.deconstructIdentifier` method, we have inefficiencies when creating new objects. There exists extra string manipulation that can be condensed and optimized so that only a single new String is created.

mineralntl:


warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -594,29 +594,32 @@ public static String deconstructIdentifier(String fieldName) {
...
         }
         
-        return fieldName;
+        return (fieldName.charAt(0) == IDENTIFIER_PREFIX) ? fieldName.substring(1) : fieldName;
     }
     
     /**
      * Rebuild the identifier with the {@link #IDENTIFIER_PREFIX} if the identifier starts with an invalid character per the Jexl IDENTIFIER definition
-     * 
+     *
```
apmoriarty: The final line can still be `return fieldName`. You can simplify the body of the method to be along these lines (just pseudo code, I don't expect this to run)
```
int startIndex = fieldName.charAt(0) == IDENTIFIER_PREFIX ? 1 : 0;
int groupingOffset = fieldName.indexOf(GROUPING_CHAR_SEP)
int stopIndex = groupingOffset == -1 ? fieldName.length()-1 : groupingOffset

fieldName = new String(fieldName.getBytes, startIndex, stopIndex)
```

apmoriarty: Tests should be fairly straightforward as well
```
FIELD => FIELD
$FIELD => FIELD

# when includeGroupingContext is false
FIELD.1.2 => FIELD
$FIELD.1.2 => FIELD

# when includeGroupingContext is true
FIELD.1.2 => FIELD.1.2
$FIELD.1.2 => FIELD.1.2
```

apmoriarty (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
mineralntl: Updated method body as suggested and added unit tests.

apmoriarty (dismissed):
Beautiful

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -594,20 +594,28 @@ public static String deconstructIdentifier(String fieldName) {
...
      * @param fieldName
      * @param includeGroupingContext
      * @return
      */
     public static String deconstructIdentifier(String fieldName, boolean includeGroupingContext) {
         if (fieldName != null && fieldName.length() > 1) {
+            boolean idPrefix = fieldName.charAt(0) == IDENTIFIER_PREFIX;
+            int startIndex = idPrefix ? 1 : 0;
+            int stopIndex = idPrefix ? fieldName.length() - 1 : fieldName.length();
```
alerman: This variable is a misnomer. it is later being used in String(fieldName.getBytes(), startIndex, stopIndex); -- stopIndex here is not an index, it is a length. As a result, I was confused by this logic. The variable should be renamed to avoid confusion.

alerman (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
mineralntl: Updated as requested

d-hwang (dismissed):

apmoriarty (dismissed):

alerman (dismissed):

apmoriarty (approved):

cogross (approved):

PR #1853, opened 2023-02-16
Improve TLDEventDataFilter's query field parse

apmoriarty:
This addresses a defect where the `termOffsetMap` portion of a content function was considered a query field
- 5a7b4f4351f5ed3d9fb3c07124426215c4429869 correct defect 
- d7f2479255797cd1eb5a57cedc19414ceeb024b5 update test to address sonar/intellij warnings

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/predicate/TLDEventDataFilter.java
+++ b/warehouse/query-core/src/main/java/datawave/query/predicate/TLDEventDataFilter.java
@@ -2,11 +2,14 @@ package datawave.query.predicate;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 
+import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
 import datawave.query.tld.TLD;
 import datawave.query.util.TypeMetadata;
 import org.apache.accumulo.core.data.ByteSequence;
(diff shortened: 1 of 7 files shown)


warehouse/query-core/src/main/java/datawave/query/predicate/TLDEventDataFilter.java
```diff
@@ -618,6 +618,23 @@ private void extractQueryFieldsFromScript(ASTJexlScript script) {
...
     
+    /**
+     * Fields are always uppercase
+     * 
+     * @param field
+     *            the field
+     * @return true if all characters are uppercase
+     */
+    private boolean containsLowers(String field) {
+        for (char c : field.toCharArray()) {
```
ivakegg: This is hacky.  Lets use that QueryFieldVisitor that we created a year ago....

ivakegg (changes requested):

warehouse/query-core/src/test/java/datawave/query/predicate/TLDEventDataFilterTest.java
```diff
@@ -594,4 +590,18 @@ public void apply_acceptGroupingFields() throws ParseException {
         assertFalse(result3);
         assertTrue(result4);
     }
+    
+    @Test
+    public void testBadIdentifier() throws ParseException {
```
avgAGB, line 603: Would it make sense to add a similar test to this with an identifier that had mixed capitalization in it, to confirm that those fields would be rejected as well?

warehouse/query-core/src/main/java/datawave/query/predicate/TLDEventDataFilter.java
apmoriarty: Opting to use the configured fields in the QueryIterator per our discussion

warehouse/query-core/src/test/java/datawave/query/predicate/TLDEventDataFilterTest.java
apmoriarty: Can do

avgAGB (approved):

ivakegg (approved):

PR #1854, opened 2023-02-16
TLD TermFrequency enhancement

apmoriarty:
Key Points
- function intersects docs keys prior to fetching offsets (only for TLD queries)
- document key set only considers the ROW_COLFAM (ignores keys with different column visibilities)
- greater integration of TermFrequencyConfig

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/DocumentKeysFunction.java
+++ b/warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/DocumentKeysFunction.java
@@ -11,7 +11,6 @@ import datawave.query.jexl.functions.ContentFunctionsDescriptor;
 import datawave.query.jexl.functions.ContentFunctionsDescriptor.ContentJexlArgumentDescriptor;
 import org.apache.accumulo.core.data.Key;
 import org.apache.commons.jexl2.parser.ASTFunctionNode;
-import org.apache.commons.jexl2.parser.ASTNotNode;
 import org.apache.commons.jexl2.parser.JexlNode;
 import org.apache.log4j.Logger;
 
@@ -29,8 +28,6 @@ public class DocumentKeysFunction {
     
     private static final Logger log = Logger.getLogger(DocumentKeysFunction.class);
     
-    // cannot intersect on negated values
-    private final Set<String> negatedValues = new HashSet<>();
     private final Multimap<String,ContentFunction> contentFunctions = LinkedListMultimap.create();
     
     /**
@@ -69,26 +66,8 @@ public class DocumentKeysFunction {
                 
                 ContentFunction contentFunction = new ContentFunction(fieldsAndTerms[0].iterator().next(), fieldsAndTerms[1]);
                 contentFunctions.put(contentFunction.getField(), contentFunction);
-                
-                // need to record any values that are part of a negative function
-                if (isFunctionNegated(f)) {
-                    negatedValues.addAll(contentFunction.getValues());
-                }
             }
         }
-    }
-    
-    private boolean isFunctionNegated(Function f) {
-        JexlNode node = f.args().get(0);
-        if (node != null) {
-            while (node.jjtGetParent() != null) {
-                node = node.jjtGetParent();
-                if (node instanceof ASTNotNode) {
-                    return true;
-                }
-            }
-        }
-        return false;
     }
     
     /**
@@ -102,19 +81,13 @@ public class DocumentKeysFunction {
      */
     protected Set<Key> getDocKeys(Document d, Set<Key> docKeys) {
         Multimap<String,Key> valueToKeys = buildValueToKeys(d);
-        
-        // each content function's intersected search space is added
-        // along with the entire search space for any negated function
         Set<Key> filterKeys = buildFilterKeys(valueToKeys);
         
-        int origSize = docKeys.size();
-        docKeys.retainAll(filterKeys);
-        int nextSize = docKeys.size();
-        if (origSize != nextSize) {
-            log.info("reduced document keys from " + origSize + " to " + nextSize + " (-" + (origSize - nextSize) + ")");
+        if (log.isDebugEnabled() && docKeys.size() != filterKeys.size()) {
+            log.debug("reduced document keys from " + docKeys.size() + " to " + filterKeys.size() + " (-" + (docKeys.size() - filterKeys.size()) + ")");
         }
         
-        return docKeys;
+        return filterKeys;
     }
     
     /**
@@ -183,6 +156,8 @@ public class DocumentKeysFunction {
     
     /**
      * Builds the filter keys from the content functions and value-to-keys multimap
+     * <p>
+     * The filter keys are effectively the union of each content function's intersection
      *
      * @param valueToKeys
      *            a multimap of value to document keys
@@ -208,11 +183,6 @@ public class DocumentKeysFunction {
             if (searchSpace != null) {
                 filterKeys.addAll(searchSpace);
             }
-        }
-        
-        // add in all keys for all negated values
-        for (String value : negatedValues) {
-            filterKeys.addAll(valueToKeys.get(value));
         }
         
         return filterKeys;
--- a/warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/DocumentKeysFunction.java
+++ b/warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/DocumentKeysFunction.java
@@ -7,32 +7,18 @@ import datawave.query.attributes.Attribute;
 import datawave.query.attributes.Attributes;
 import datawave.query.attributes.Document;
 import datawave.query.attributes.PreNormalizedAttribute;
-import datawave.query.jexl.JexlASTHelper;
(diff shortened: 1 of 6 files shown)


warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/DocumentKeysFunction.java
```diff
@@ -0,0 +1,291 @@
...
+    
+    // functions may take different forms..
+    // within = {field, number, termOffsetMap, terms...}
+    // within = {number, termOffsetMap, terms...}
+    // adjacent = {field, termOffsetMap, terms...}
+    // adjacent = {termOffsetMap, terms...}
+    // phrase = {field, termOffsetMap, terms...}
+    // phrase = {termOffsetMap, terms...}
+    // scoredPhrase = {-1.5, termOffsetMap,'boy','car'}
+    private int findValueStartIndex(Function function) {
```
ivakegg: This seems like something the functions descriptor should be able to tell us.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/DocumentKeysFunction.java
```diff
@@ -0,0 +1,291 @@
...
+     * @param config
+     *            a {@link TermFrequencyConfig}
+     */
+    public DocumentKeysFunction(TermFrequencyConfig config) {
+        populateContentFunctions(config.getScript());
+    }
+    
+    protected void populateContentFunctions(JexlNode node) {
+        
+        Multimap<String,Function> functions = TermOffsetPopulator.getContentFunctions(node);
```
ivakegg, line 48: Why only content functions?  Also the Jexl utils and the function descriptors would be more suitable.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/DocumentKeysFunction.java
```diff
@@ -0,0 +1,291 @@
...
+        for (String key : functions.keySet()) {
+            Collection<Function> coll = functions.get(key);
+            for (Function f : coll) {
+                ContentFunction contentFunction = transformFunction(f);
+                contentFunctions.put(contentFunction.getField(), contentFunction);
+            }
+        }
+    }
+    
+    private ContentFunction transformFunction(Function function) {
```
ivakegg: This also seems like a weird place for this functionality.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/DocumentKeysFunction.java
apmoriarty: We're dealing with reducing the search space when aggregating term frequencies -- which have always been limited to content functions. See the TermOffsetPopulator for proof of that.

The TermOffsetPopulator call to `getContentFunctions` does make use of the JexlAstHelper which I think is the referenced jexl utility.

I'll take a look at using the content functions descriptor to extract fields and values.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/DocumentKeysFunction.java
apmoriarty: I'll take a cut at using the descriptors

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/DocumentKeysFunction.java
apmoriarty: Fair

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/DocumentKeysFunction.java
FineAndDandy: Will we ever use TF offsets for anything besides content functions?

apmoriarty: @FineAndDandy I don't believe so

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/DocumentKeysFunction.java
```diff
@@ -0,0 +1,220 @@
...
+        }
+    }
+    
+    private boolean isFunctionNegated(Function f) {
+        JexlNode node = f.args().get(0);
+        if (node != null) {
+            while (node.jjtGetParent() != null) {
+                node = node.jjtGetParent();
+                if (node instanceof ASTNotNode) {
+                    return true;
```
ivakegg: Might we handle a not within a not?  This should travel all the way to the top of the tree.

ivakegg: JexlASTHelper.isNegated(JexlNode node) might be in order.  I am sure there are many places we could be using that.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/DocumentKeysFunction.java
apmoriarty: Switched to using the ContentFunctionsDescriptor

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/DocumentKeysFunction.java
apmoriarty: Just documenting our earlier conversation: we shouldn't need special handling of negations in this case.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/DocumentKeysFunction.java
```diff
@@ -0,0 +1,190 @@
...
+                    throw new IllegalArgumentException("parent was not a function node");
+                }
+                
+                argsDescriptor = descriptor.getArgumentDescriptor((ASTFunctionNode) parent);
+                
+                // content, tf, and indexed fields are not actually needed to extract fields from the function node
+                fieldsAndTerms = argsDescriptor.fieldsAndTerms(Collections.emptySet(), Collections.emptySet(), Collections.emptySet(), null);
+                
+                if (fieldsAndTerms[0].size() != 1) {
+                    throw new IllegalStateException("content function had more than one field");
```
FineAndDandy, line 64: can an ANYFIELD expansion create this? or is it always expanded into multiple functions?

apmoriarty: I don't believe so. At this point all fields should have been expanded and all content functions rewritten to be one field per function (that is a multi-fielded content function becomes a union of two single fielded functions).

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TFFactory.java
```diff
@@ -61,9 +61,15 @@ public static com.google.common.base.Function<Tuple2<Key,Document>,Tuple3<Key,Do
...
         } else {
-            TermOffsetPopulator offsetPopulator = new TermOffsetPopulator(termFrequencyFieldValues, config.getContentExpansionFields(),
-                            config.getEvaluationFilter(), config.getSource());
-            return new TermOffsetFunction(offsetPopulator, config.getTfFields());
+            
+            DocumentKeysFunction docKeyFunction = null;
+            
+            if (config.isTld()) {
+                docKeyFunction = new DocumentKeysFunction(config);
+            }
```
FineAndDandy, line 69: why tld only?

apmoriarty: An event query by definition is for a single document, so the set of document keys will only be size one. 

TLD on the other hand could have many child document keys, and intersecting those before aggregating tfs will reduce the amount of work required.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermOffsetPopulator.java
```diff
@@ -37,20 +31,24 @@
...
+import java.util.Set;
+import java.util.SortedSet;
 
 import static datawave.query.Constants.TERM_FREQUENCY_COLUMN_FAMILY;
 
 public class TermOffsetPopulator {
     private static final Logger log = Logger.getLogger(TermOffsetPopulator.class);
     
-    private static final Set<String> phraseFunctions;
+    static final Set<String> phraseFunctions;
```
FineAndDandy: this can stay private

FineAndDandy (dismissed):

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/TermOffsetPopulator.java
apmoriarty: good catch. Corrected this in the next commit.

ivakegg (approved):

FineAndDandy (approved):

PR #1863, opened 2023-02-24
WIP: issue-1229: Save shard-specific query subplans to metrics

### LINKED ISSUE #1229: Save shard-specific query plans to metrics
Shard-specific query plans are created by the visitor function.  We should add these query plans to the metrics for the query, and make the metrics viewable via the metrics endpoint.  

The metrics endpoint should be updated to accept a shard query parameter which can be used to determine which query plan is displayed.  We could/should also display a column with the shards associated with the query, and potentially make each one a hyperlink to the metrics entry specific to that shard.

foster33:
Aims to add query subplans to the metrics for the query. Also will make the subplans viewable via the metrics endpoint.

Fixes #1229


warehouse/query-core/src/main/java/datawave/query/metrics/ShardTableQueryMetricHandler.java
```diff
@@ -560,6 +560,44 @@ public QueryMetricsDetailListResponse query(String user, String queryId, Datawav
         return response;
     }
     
+    // This method below needs to be changed to only show subplans
+    public QueryMetricsDetailListResponse subplan(String user, String queryId, DatawavePrincipal datawavePrincipal) {
```
ivakegg: Yes this method needs to be changed.  Firstly I would expect a different response object to be returned.

warehouse/query-core/src/main/java/datawave/query/metrics/ShardTableQueryMetricHandler.java
```diff
@@ -560,6 +560,44 @@ public QueryMetricsDetailListResponse query(String user, String queryId, Datawav
...
+            query.setQuery("QUERY_ID == '" + queryId + "'");
+            query.setQueryName(QUERY_METRICS_LOGIC_NAME);
+            query.setColumnVisibility(visibilityString);
+            query.setQueryAuthorizations(AuthorizationsUtil.buildAuthorizationString(authorizations));
+            query.setExpirationDate(DateUtils.addDays(new Date(), 1));
+            query.setPagesize(1000);
+            query.setUserDN(datawavePrincipal.getShortName());
+            query.setOwner(datawavePrincipal.getShortName());
+            query.setId(UUID.randomUUID());
+            query.setParameters(ImmutableMap.of(QueryOptions.INCLUDE_GROUPING_CONTEXT, "true"));
```
ivakegg, line 591: I am thinking that we may want some addition parameters here to restrict the fields that you want returned in the event.  QueryOptions.  Use QueryParameters.RETURN_FIELDS to specified the fields you want back.  So for example in this response you may not need the original query or the original query plan.

This also means that in the method about you may want to use QueryParameters.BLACKLISTED_FIELDS to avoid returning the subplans since they are not needed for the normal response.

ivakegg:
We need test cases!

PR #1865, opened 2023-02-24
KeyParser interface

apmoriarty:
Any and all comments are welcome. Feedback is especially welcome for the KeyParser interface, it's implementations, and for specific unit test edge cases. 

This draft merge request adds the concept of a KeyParser to address certain context-dependent deficiencies in the DatawaveKey.

The DatawaveKey is flexible in that it can parse and validate a number of different key types. There are certain drawbacks:
- a new instance of the DatawaveKey is required for each new accumulo key it parses
- keys are fully parsed up front, even if you only need to access a single logical subcomponent (for example, the uid)

The KeyParser focuses on the following features, in contrast to the DatawaveKey
- the parser is re-usable
- the parser is tailored to a specific key implementation
- the actual parse is done lazily

These features introduce a trade off which imposes the following constraints
- minimal verification is performed on the key
- the caller is responsible for ensuring the correct key types are given to the parser

The initial draft contains implementations for the following key types
- FieldIndex
- Event
- TermFrequency


warehouse/query-core/src/main/java/datawave/query/data/parsers/EventKey.java
```diff
@@ -0,0 +1,194 @@
...
+                throw new IllegalArgumentException("Failed to parse root uid from event key");
+            }
+        }
+        return rootUid;
+    }
+    
+    /**
+     * Performs an optimized scan of the uid.
+     * <p>
+     * A HashUID is minimally a 21 character string of three alphanumerics of length six, joined by two dots
```
FineAndDandy: 21 characters is too long for a minimal hash

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyIterator.java
```diff
@@ -192,11 +192,6 @@ protected void findTop() throws IOException {
             
             tfKey.parse(k);
```
FineAndDandy, line 193: does this now need to be reset after finding top?

warehouse/query-core/src/main/java/datawave/query/data/parsers/KeyParser.java
```diff
@@ -0,0 +1,71 @@
...
+     *
+     * @param k
+     *            a key
+     */
+    void parse(Key k);
+    
+    /**
+     * Clears existing state
+     */
+    void clearState();
```
FineAndDandy, line 36: making sure this is called everywhere that a key is parsed may not be simple/bug free.

warehouse/query-core/src/main/java/datawave/query/data/parsers/FieldIndexKey.java
```diff
@@ -4,16 +4,16 @@
...
+ * FieldIndex key structure is
  * <ul>
  * <li>row - the shard</li>
- * <li>column family - "fi\0FIELD_NAME"</li>
+ * <li>column family - "fi\0FIELD"</li>
  * <li>column qualifier - "value\0datatype\0uid"</li>
  * </ul>
  */
-public class FieldIndexKey {
+public class FieldIndexKey implements KeyParser {
```
FineAndDandy, line 17: In the EventKey you would parse on demand, here you are generally returning null. I think the different Key's should function the same, whichever way you prefer.

warehouse/query-core/src/main/java/datawave/query/data/parsers/TermFrequencyKey.java
```diff
@@ -0,0 +1,160 @@
...
+        }
+        return uid;
+    }
+    
+    @Override
+    public String getRootUid() {
+        if (uid == null) {
+            throw new IllegalArgumentException("Failed to parse root UID from tf key");
+        }
+        return null;
```
FineAndDandy: this doesn't look right

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyIndexIterator.java
```diff
@@ -79,7 +79,7 @@ public class TermFrequencyIndexIterator implements SortedKeyValueIterator<Key,Va
     private final Text cf = new Text();
     private final Text cq = new Text();
     
-    private final TFKey tfParser = new TFKey();
+    private final TermFrequencyKey tfParser = new TermFrequencyKey();
```
FineAndDandy, line 82: same concern here about state management of the keys

FineAndDandy (changes requested):

warehouse/query-core/src/main/java/datawave/query/data/parsers/EventKey.java
apmoriarty: I patterned this off of `TLD.estimateRootPointer()`...will check again for a reasonable minimum

warehouse/query-core/src/main/java/datawave/core/iterators/TermFrequencyIterator.java
apmoriarty: No. Internal state is cleared at the beginning of a `parse()` call.

warehouse/query-core/src/main/java/datawave/query/data/parsers/KeyParser.java
apmoriarty: By default, `clearState()` is called at the beginning of the parse method. An operator may optionally call it once the values of a key are no longer needed.

warehouse/query-core/src/main/java/datawave/query/data/parsers/TermFrequencyKey.java
apmoriarty: Hadn't implemented that yet for the TF key parser. Should be added now.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyIndexIterator.java
apmoriarty: Same comment from above, `clearState()` is called as part of the `parse()`, should be fine.

warehouse/query-core/src/main/java/datawave/query/data/parsers/FieldIndexKey.java
ivakegg: I am a little confused by this comment.  I don't see nulls being returned.

ivakegg (approved):

FineAndDandy (approved):

PR #1869, opened 2023-03-01
More ways to configure seeking in the QueryIterator

apmoriarty:
Add more fine-grained options for configuring seeking in the query iterator. Intended for future use in the aggregators and data filters.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
+++ b/warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
@@ -395,12 +395,13 @@ public class ShardQueryConfiguration extends GenericQueryConfiguration implement
     // The class for the excerpt iterator
     private Class<? extends SortedKeyValueIterator<Key,Value>> excerptIterator = TermFrequencyExcerptIterator.class;
     
-    private int fiFieldSeek;
-    private int fiNextSeek;
-    private int eventFieldSeek;
-    private int eventNextSeek;
-    private int tfFieldSeek;
-    private int tfNextSeek;
+    // controls when to issue a seek. disabled by default.
+    private int fiFieldSeek = -1;
+    private int fiNextSeek = -1;
+    private int eventFieldSeek = -1;
+    private int eventNextSeek = -1;
+    private int tfFieldSeek = -1;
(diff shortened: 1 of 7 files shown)


ivakegg:
I thought we decided that accumulo was handling the nexts before a seek.  Why are we creating all of this instumentation?

alerman (dismissed):

apmoriarty:
> I thought we decided that accumulo was handling the nexts before a seek. Why are we creating all of this instumentation?

Yes, we did determine that accumulo handles nexting before a seek! There are a few cases where we actually want to next a few times at the datawave level. One specific case is the tld event data filters, they build a seek range based on the last seen field so we want to roll into the next field before issuing the seek call to accumulo. 

In the worst case we are dealing with very high cardinality fields and values, and we don't gain anything from this approach. In the best case scenario we are able to trigger a seek that skips many, many fields and values. 

Ultimately we may decide to disable nexting and always seek in certain cases, but still want to retain at least one next call before issuing the seek. This merge request provides us with the most flexibility to make that happen.

FineAndDandy:
@apmoriarty are there specific places you see that this will be necessary? I understand as-is this doesn't change any behavior, but if we don't need the knob and don't expect to need the knob I'm not sure we need it?

apmoriarty:
> @apmoriarty are there specific places you see that this will be necessary? I understand as-is this doesn't change any behavior, but if we don't need the knob and don't expect to need the knob I'm not sure we need it?

Here's a rough draft of what I'm thinking
- field index field threshold - potentially required by ivarators if/when ANYFIELD terms are handled by the query iterator
- field index next threshold - will be used by tld event data filters
- event field threshold - will be used by tld event data filter
- event next threshold - will be used by tld event data filter (and will want that to be different from the fi next threshold)
- tf field threshold - should not be explicitly necessary, but will need this config since the data filters are for the most part key type agnostic, and are currently cloned between the various uses (fi, event, tf)
- tf next threshold - potentially configured off, but again, will want to configure this separately from fi and event next thresholds

If you think any of those use cases aren't worth considering, let me know.

FineAndDandy (approved):
After talking to drew this is more about workflow and teasing apart existing functionality than counting seeks prior to next

alerman (approved):

Issue #1873, opened 2023-03-06
Check for null when getting defaultDataTypeHandlers in IngestJob


FineAndDandy:
not all types have handlers, verify they exist to prevent NPE

PR #1875, opened 2023-03-06
Prevent NPE in IngestJob.setupHandlers (#1873)

FineAndDandy:


javease (dismissed):

apmoriarty (approved):
Added a commit from integration to fix the runner issue, otherwise LGTM

nonessentialprototype:
Do we want `log.info` or `log.warn` on scenarios where the handlers is null?

ivakegg (approved):

Issue #1877, opened 2023-03-07
RemoteQueryMetricService voidResponseReader not initialized


billoley:
RemoteQueryMetricService.voidResponseReader was previously initialized, but that line was reomved (possibly because voidResponseReader is declared and initialized in the superclass).  However, since it is declared and not initialized, we are getting an NPE whenever that response reader is used.

PR #1878, opened 2023-03-07
re #1877 RemoteQueryMetricService voidResponseReader not initialized

billoley:
RemoteQueryMetricService.voidResponseReader was previously initialized, but that line was reomved (possibly because voidResponseReader is declared and initialized in the superclass). However, since it is declared and not initialized, we are getting an NPE whenever that response reader is used.


ivakegg (approved):

keith-ratcliffe (approved):
LGTM

PR #1886, opened 2023-03-14
Removing datatype filter from termfrequencyindexiterator

avgAGB:


ivakegg (commented):
Why exactly are we removing the dataType filter?

avgAGB:
> Why exactly are we removing the dataType filter?

The description of the ticket states that the TermFrequencyIndexIterator is scanning the TF column family, and because the datatype is built into the scan range no datatype filter needs to be applied.

matthpeterson (dismissed):
It appears that datatypeFilter was never read.  Looks good to me.

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/iterator/builder/TermFrequencyIndexBuilder.java
jwomeara, line 1: If we're not using this.datatypeFilter anymore, let's remove it from the class completely.

jwomeara (changes requested):
I get that the field index range contains the type, so adding type filtering at this level doesn't make sense, but are we verifying at a higher level that we don't generate ranges against disallowed types?  i.e. if someone sets a type filter upstream, are we ensuring that we don't create term frequency ranges against those types?  i suspect not, but maybe we should.

warehouse/query-core/src/main/java/datawave/query/iterator/builder/TermFrequencyIndexBuilder.java
avgAGB: Done, removed from the latest commit.

avgAGB:
> I get that the field index range contains the type, so adding type filtering at this level doesn't make sense, but are we verifying at a higher level that we don't generate ranges against disallowed types? i.e. if someone sets a type filter upstream, are we ensuring that we don't create term frequency ranges against those types? i suspect not, but maybe we should.

To clarify a bit, did you want to verify that if someone sets a type filter on a query that the term frequency fields that are retrieved are the ones that are part of the type filter set?

jwomeara:
> > I get that the field index range contains the type, so adding type filtering at this level doesn't make sense, but are we verifying at a higher level that we don't generate ranges against disallowed types? i.e. if someone sets a type filter upstream, are we ensuring that we don't create term frequency ranges against those types? i suspect not, but maybe we should.
> 
> To clarify a bit, did you want to verify that if someone sets a type filter on a query that the term frequency fields that are retrieved are the ones that are part of the type filter set?

I believe that's what I'm saying, yes.  If a user specified only wanting to query against a specific set of types, we should make sure that we don't generate TF ranges against any types but those allowed by the user.

avgAGB:
> > > I get that the field index range contains the type, so adding type filtering at this level doesn't make sense, but are we verifying at a higher level that we don't generate ranges against disallowed types? i.e. if someone sets a type filter upstream, are we ensuring that we don't create term frequency ranges against those types? i suspect not, but maybe we should.
> > 
> > 
> > To clarify a bit, did you want to verify that if someone sets a type filter on a query that the term frequency fields that are retrieved are the ones that are part of the type filter set?
> 
> I believe that's what I'm saying, yes. If a user specified only wanting to query against a specific set of types, we should make sure that we don't generate TF ranges against any types but those allowed by the user.

This one slipped through the cracks for me, but I managed to verify with test queries that even with these removed, the metadata helper will only build a set of term frequency fields based on the filters that the user specifies. I'll clean up the merge conflicts and request for some reviews again when I get an opportunity.

apmoriarty (approved):

jwomeara (approved):

Issue #1887, opened 2023-03-15
RemoteHttpService parent class should not call init after @PostConstruct


billoley:
While some new Spring-created subclasses call both their @Override int() method and the base class init() method while protecting against executing the logic twice, calling init() from RemoteHttpService is causing other non-Spring-created subclasses to be constantly re-initialized whenever executePostMethodWithRuntimeException or executeGetMethodWithRuntimeException are called.

PR #1888, opened 2023-03-15
re #1887 RemoteHttpService parent class should not call init after @P…

billoley:
RemoteHttpService parent class should not call init after @PostConstruct #1887

While some new Spring-created subclasses call both their @OverRide int() method and the base class init() method while protecting against executing the logic twice, calling init() from RemoteHttpService is causing other non-Spring-created subclasses to be constantly re-initialized whenever executePostMethodWithRuntimeException or executeGetMethodWithRuntimeException are called.


jwomeara (approved):
Thanks!  I hate it!

keith-ratcliffe (approved):

Issue #1891, opened 2023-03-16
ShardTableConfigHelper should optionally override maxuids on index tables


FineAndDandy:
Depends on #1889. Allow configuration based table overrides of max uids

Issue #1889, opened 2023-03-16
PropogatingIterator should allow options to be passed to configured Aggregators


FineAndDandy:
Optionally allow options to be passed to a configured aggregator by passing <cf>.opts.<option> as an option on the PropogatingIterator. If set call agg.validateOptions(options) on the extracted options

PR #1892, opened 2023-03-17
ShardTableConfigHelper supports setting maxUids on index and reverse index tables

FineAndDandy:
See #1891 dependent on #1889. Should be a single commit after merge.

### DIFF
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/table/config/ShardTableConfigHelper.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/table/config/ShardTableConfigHelper.java
@@ -14,6 +14,7 @@ import datawave.ingest.table.balancer.ShardedTableTabletBalancer;
 import datawave.ingest.table.bloomfilter.ShardKeyFunctor;
 import datawave.ingest.table.bloomfilter.ShardIndexKeyFunctor;
 
+import datawave.iterators.PropogatingIterator;
 import org.apache.accumulo.core.client.AccumuloException;
 import org.apache.accumulo.core.client.AccumuloSecurityException;
 import org.apache.accumulo.core.client.IteratorSetting;
@@ -32,6 +33,7 @@ public class ShardTableConfigHelper extends AbstractTableConfigHelper {
     protected static final String SHARDED_TABLET_BALANCER_CLASS = ShardedTableTabletBalancer.class.getName();
     
     public static final String KEEP_COUNT_ONLY_INDEX_ENTRIES = "index.tables.keep.count.only.entries";
+    public static final String MAX_INDEX_UIDS = "index.tables.max.uids";
     
     public static final String SHARD_TABLE_BALANCER_CONFIG = "shard.table.balancer.class";
     protected String shardTableBalancerClass = SHARDED_TABLET_BALANCER_CLASS;
@@ -203,6 +205,16 @@ public class ShardTableConfigHelper extends AbstractTableConfigHelper {
                 aggClass = KeepCountOnlyUidAggregator.class.getName();
             }
             
+            if (conf.get(MAX_INDEX_UIDS) != null) {
+                String maxUids = conf.get(MAX_INDEX_UIDS);
+                String maxUidProperty = stem + PropogatingIterator.getOptString("*", GlobalIndexUidAggregator.MAX_UIDS);
+                setPropertyIfNecessary(tableName, maxUidProperty, maxUids, tops, log);
+                
+                // anytime a property is set also must set this property for Combiner properties
+                String combinerProperty = stem + PropogatingIterator.getOptString("*", "all");
+                setPropertyIfNecessary(tableName, combinerProperty, "true", tops, log);
+            }
+            
             setPropertyIfNecessary(tableName, stem + "*", aggClass, tops, log);
             
             if (markingsSetupIteratorEnabled) {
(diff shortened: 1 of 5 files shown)


FineAndDandy:
wait for merge of #1889 prior to reviewing

apmoriarty (approved):

nonessentialprototype (approved):

apmoriarty (changes requested):
Approach needs to be reworked

warehouse/ingest-core/src/main/java/datawave/iterators/PropogatingIterator.java
jwomeara, line 1: Propagating being spelled wrong makes me sad...

warehouse/ingest-core/src/main/java/datawave/iterators/PropogatingIterator.java
```diff
@@ -16,20 +16,25 @@
...
 
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
```
jwomeara, line 30: Collections, HashSet, and Set are unused.

warehouse/ingest-core/src/main/java/datawave/iterators/PropogatingIterator.java
```diff
@@ -319,11 +331,49 @@ public boolean validateOptions(Map<String,String> options) {
         
         PropogatingCombiner propAgg = null;
         
+        Map<String,Map<String,String>> allAggOptions = new HashedMap();
```
jwomeara, line 334: any reason to use HashedMap vs HashMap?

warehouse/ingest-core/src/main/java/datawave/iterators/PropogatingIterator.java
```diff
@@ -319,11 +331,49 @@ public boolean validateOptions(Map<String,String> options) {
...
+        // make a first pass of the options to find any aggregator options and build them up for each aggregator
         for (Entry<String,String> familyOption : options.entrySet()) {
+            String option = familyOption.getKey();
+            // if this property specifies aggregator options skip
```
jwomeara, line 338: It doesn't look like we're skipping if the property contains aggregator options.

warehouse/ingest-core/src/main/java/datawave/iterators/PropogatingIterator.java
```diff
@@ -319,11 +331,49 @@ public boolean validateOptions(Map<String,String> options) {
...
+            if (option.contains(AGGREGATOR_OPTS)) {
+                // get the aggregator name
+                String[] splits = option.split("\\.");
+                String aggName = splits[0];
+                if (splits.length < 3) {
+                    throw new IllegalArgumentException("When specifying options for an aggregator specify <cf>." + AGGREGATOR_OPTS + "<option>");
```
jwomeara, line 344: instead of \<option\>, use \<CombinerOptionName\> since that is what was used in the describeOptions method. Also, I guess this means we can't have any `.` characters in our combiner option names.

warehouse/ingest-core/src/main/java/datawave/iterators/PropogatingIterator.java
```diff
@@ -319,11 +331,49 @@ public boolean validateOptions(Map<String,String> options) {
...
+                }
+                
+                Map<String,String> aggOptions = allAggOptions.get(aggName);
+                if (aggOptions == null) {
+                    aggOptions = new HashMap<>();
+                    allAggOptions.put(aggName, aggOptions);
+                }
```
jwomeara, line 351: this can be replaced with computeIfAbsent

warehouse/ingest-core/src/main/java/datawave/iterators/PropogatingIterator.java
```diff
@@ -319,11 +331,49 @@ public boolean validateOptions(Map<String,String> options) {
...
+                
+                // strip off the first two sections of the option to build an option for the aggregator
+                String aggOption = option.replace(splits[0] + "." + splits[1] + ".", "");
```
jwomeara, line 354: if we've already split the option into it's three components (e.g. COLFAM.opts.MYOPTION -> 1) COLFAM, 2) opts, 3) MYOPTION) why can't jkust set aggOption as follows? `aggOption = splits[2]`

warehouse/ingest-core/src/main/java/datawave/iterators/PropogatingIterator.java
```diff
@@ -319,11 +331,49 @@ public boolean validateOptions(Map<String,String> options) {
...
+                aggOptions.put(aggOption, familyOption.getValue());
+            }
+        }
+        
+        for (Entry<String,String> familyOption : options.entrySet()) {
+            // if this property specifies aggregator options skip
+            if (familyOption.getKey().contains(AGGREGATOR_OPTS)) {
+                continue;
+            }
```
jwomeara, line 363: As an alternative, on your first pass over the options, you could create a new map of options containing only the non-AGGREGATOR_OPTS options, then iterate over that here, and skip this check.

warehouse/ingest-core/src/main/java/datawave/iterators/PropogatingIterator.java
```diff
@@ -319,11 +331,49 @@ public boolean validateOptions(Map<String,String> options) {
...
+            
+            // the property should an aggregator class since it isn't aggregator class options
```
jwomeara, line 365:
```suggestion
            // the property should be an aggregator class since it isn't an aggregator class option
```

warehouse/ingest-core/src/main/java/datawave/ingest/table/config/ShardTableConfigHelper.java
```diff
@@ -203,6 +205,16 @@ protected void configureGidxTable(TableOperations tops) throws AccumuloException
                 aggClass = KeepCountOnlyUidAggregator.class.getName();
             }
             
+            if (conf.get(MAX_INDEX_UIDS) != null) {
+                String maxUids = conf.get(MAX_INDEX_UIDS);
+                String maxUidProperty = stem + PropogatingIterator.getOptString("*", GlobalIndexUidAggregator.MAX_UIDS);
```
jwomeara, line 210: Just want to make sure I understand this correctly.  Is the idea that `*` means that this setting applies to all column families?  Does the logic in PropogatingIterator bear that out?  It doesn't look like it to me.  Looks like it looks for an exact column family match.

jwomeara (changes requested):
See comments.

Issue #1895, opened 2023-03-17
Improvements to QueryMetricsWriter


billoley:
Make number of updates that are pulled off of the queue at a time configurable.  This impacts the RemoteQueryMetricService path more than the ShardTableQueryMetricHandler path because we send the updates as one POST request to the remote service while the updates one at a time to the ShardTableQueryMetricHandler

Get rid of code duplication and use common methods for 1) fetching the updates from the queue 2) handling of failed writing of updates

Improve logging of failure handling

Use an EJB managed thread and a Runnable to continuously process metric updates instead of relying on a `@Scheduled` method where the executions overlap and produce log messages warning about this.

Improve ShardTableQueryMetricHandler throughput by relying on the BatchWriter's latency and maxmemory settings instead of calling flush() 

Delete `@RolesAllowed` and `@DeclareRoles` lines as users don't interact with QueryMetricsWriter

PR #1896, opened 2023-03-20
re #1895 Refactor QueryMetricsWriter to improve reliability, failure …

billoley:
Make number of updates that are pulled off of the queue at a time configurable. This impacts the RemoteQueryMetricService path more than the ShardTableQueryMetricHandler path because we send the updates as one POST request to the remote service while the updates one at a time to the ShardTableQueryMetricHandler

Get rid of code duplication and use common methods for 1) fetching the updates from the queue 2) handling of failed writing of updates

Improve logging of failure handling

Use an EJB managed thread and a Runnable to continuously process metric updates instead of relying on a @Scheduled method where the executions overlap and produce log messages warning about this.

Improve ShardTableQueryMetricHandler throughput by relying on the BatchWriter's latency and maxmemory settings instead of calling flush()

Delete @RolesAllowed and @DeclareRoles lines as users don't interact with QueryMetricsWriter

### DIFF
--- a/web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsWriter.java
+++ b/web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsWriter.java
@@ -3,29 +3,32 @@ package datawave.webservice.query.metric;
 import java.text.DecimalFormat;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.Comparator;
-import java.util.Date;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeSet;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 
 import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
 import javax.annotation.Resource;
-import javax.annotation.security.DeclareRoles;
-import javax.annotation.security.RolesAllowed;
 import javax.annotation.security.RunAs;
 import javax.ejb.LocalBean;
 import javax.ejb.Lock;
 import javax.ejb.LockType;
-import javax.ejb.Schedule;
 import javax.ejb.Singleton;
 import javax.ejb.Startup;
+import javax.enterprise.concurrent.ManagedThreadFactory;
 import javax.enterprise.event.Observes;
 import javax.enterprise.inject.spi.BeanManager;
 import javax.inject.Inject;
+import javax.jms.ConnectionFactory;
 import javax.jms.JMSConsumer;
 import javax.jms.JMSContext;
 import javax.jms.Message;
@@ -34,7 +37,6 @@ import javax.jms.Queue;
 
 import org.apache.commons.collections4.map.LRUMap;
 import org.apache.commons.lang.StringUtils;
-import org.apache.commons.lang.builder.CompareToBuilder;
 import org.apache.deltaspike.core.api.exclude.Exclude;
 import org.apache.log4j.Logger;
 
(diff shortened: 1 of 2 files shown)


keith-ratcliffe (dismissed):

apmoriarty (dismissed):

web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsWriter.java
```diff
@@ -189,158 +206,145 @@ public long getAge() {
...
-                itr.remove();
-            } else {
+            if (queryMetricsWriterConfiguration.getUseRemoteService()) {
                 try {
                     writeMetricsToRemoteService(Collections.singletonList(f.getMetric()));
                     itr.remove();
-                    anySuccess = true;
-                } catch (Exception e1) {
-                    log.error("metric failed write to remote service, will retry - " + e1.getMessage());
+                    successful++;
```
nonessentialprototype, line 298: The naming for this variable seems slightly confusing.  An exception is a success?

nonessentialprototype (dismissed):

web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsWriter.java
billoley: No - if writeMetricsToRemoteService(Collections.singletonList(f.getMetric())) returns with no exception, then we remove that item (itr.remove()) and increment successful.  The catch clause has no code since the metric that was beng written during the exception will remain in failedMetrics and we process the items remaining in that list later.

billoley: You seem to be using the "unified" view and it's mis-representing the diff.

web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsWriter.java
```diff
@@ -41,53 +39,73 @@
...
     private QueryMetricsWriterConfiguration queryMetricsWriterConfiguration;
     
     private UdpClient timelyClient = null;
     private Map<String,Long> lastPageMetricMap;
-    
-    private Map<String,Long> lastPageMap;
-    private List<QueryMetricHolder> metricQueue;
     private DecimalFormat df = new DecimalFormat("0.00");
+    private Future future = null;
+    private boolean shuttingDown = false;
```
jwomeara: this should probably be volatile

web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsWriter.java
```diff
@@ -189,158 +206,145 @@ public long getAge() {
...
+                    log.error(e.getMessage(), e);
+                }
+            }
         }
     }
     
-    private void useRemoteService() {
-        if (receivingMetrics.compareAndSet(false, true)) {
+    private void processQueryMetricsWithRemoteService(List<QueryMetricHolder> metricHolderQueue) {
+        List<BaseQueryMetric> metricQueue = metricHolderQueue.stream().map(h -> h.getQueryMetric()).collect(Collectors.toList());
```
jwomeara: `h -> h.getQueryMetric()` could be replaced with `QueryMetricHolder::getQueryMetric`

web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsWriter.java
```diff
@@ -189,158 +206,145 @@ public long getAge() {
...
+                log.trace("failures individual/total:" + iFailures + "/" + tFailures + " for metric update " + queryId);
             }
-        });
-        return failedMetrics;
+        }
+        if (discardForFailureCount > 0) {
+            log.error("Discarding " + discardForFailureCount + " updates from queries " + discardForFailureMetrics + " for repeated failures");
+        }
+        if (discardForTimeCount > 0) {
+            log.error("Discarding " + discardForTimeCount + " updates from queries " + discardForTimeMetrics + " for exceeding max time in failure queue");
```
jwomeara, line 347: I appreciate the logging here.

web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsWriterConfiguration.java
```diff
@@ -7,6 +7,7 @@ public class QueryMetricsWriterConfiguration {
     
     private String timelyHost = null;
     private int timelyPort = 0;
+    private int batchSize = 100;
```
jwomeara, line 10: 100 vs 1000.  was this intentional?  if so, disregard.

jwomeara (changes requested):
1 minor nit about making the `shuttingDown` boolean volatile.  other than that, looks good.

web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsWriterConfiguration.java
billoley: Yes - intentional

jwomeara (dismissed):

keith-ratcliffe (dismissed):

PR #1902, opened 2023-03-22
Fix nested field group names, edge efficiencies

hlgp:
Fixes nested field group names so we don't have to specify all possible combinations of offset lineage in configurations. e.g.:
  FIELD.Parent1_0.Parent2_0.FIELD_0 
  FIELD.Parent1_1.Parent2_0.FIELD_0
  FIELD.Parent1_1.Parent2_1.FIELD_0
  ...
  FIELD.Parent1_N.Parent2_M.FIELD_X
  used to have the groups explicitly defined for all real integers, N,M, and X but can now all be referenced with Parent1.Parent2. 
 The event field name remains unchanged.

Fixes the case where we were generating edges from a field to itself (not a different instance. its literal self)

Moves a short circuit in edge generation up so we don't waste time checking preconditions on events that don't even have the necessary fields.

Allows edge definitions for child datatypes.

Addresses #1905


warehouse/ingest-core/src/main/java/datawave/ingest/data/TypeRegistry.java
```diff
@@ -243,6 +244,15 @@ private TypeRegistry(Configuration config) {
                     Type t = new Type(typeName, outputName, helperClass, readerClass, handlerClassNames, filterPriority, filterClassNames);
                     log.debug("Registered type " + t);
                     this.put(typeName, t);
+                    
+                    if (null != config.get(typeName + DataTypeOverrideHelper.Properties.DATA_TYPE_VALUES)) {
```
ivakegg, line 248: excellent.

warehouse/ingest-core/src/main/java/datawave/ingest/data/normalizer/SimpleGroupFieldNameParser.java
```diff
@@ -35,20 +30,37 @@ public SimpleGroupFieldNameParser(boolean shouldIncludeDelimInGroupName) {
      */
```
ivakegg, line 30: Add comments please describing what "nested" means here.

warehouse/ingest-core/src/main/java/datawave/ingest/data/normalizer/SimpleGroupFieldNameParser.java
```diff
@@ -17,16 +18,10 @@
 public class SimpleGroupFieldNameParser {
     private static final long serialVersionUID = 1035918631638323565L;
     private static final Logger log = Logger.getLogger(SimpleGroupFieldNameParser.class);
-    private final boolean shouldIncludeDelimInGroupName;
+    private static final Pattern parentGroupingPattern = Pattern.compile("_\\d");
```
ivakegg: I don't have the answer here, but pattern matching can be somewhat expensive.  I expect this class is used quite a lot so I am wondering whether we should use a more optimized routine:
```
    static String trimGroup(String groupStr) {
        StringBuilder group = new StringBuilder();
        boolean inSubgroup = false;
        int lastDotIndex = -1;
        int groupStart = -1;
        for (int i = 0; i < groupStr.length(); i++) {
            char c = groupStr.charAt(i);
            if (inSubgroup) {
                if (c == '.') {
                    lastDotIndex = group.length();
                    group.append(c);
                    inSubgroup = false;
                } else if (Character.isDigit(c)) {
                    // continue
                } else {
                    group.append(groupStr.substring(groupStart, i+1));
                    inSubgroup = false;
                }
            } else if (c == '_') {
                inSubgroup = true;
                groupStart = i;
            } else if (c == '.') {
                lastDotIndex = group.length();
                group.append('.');
            } else {
                group.append(c);
            }
        }
        if (lastDotIndex >= 0) {
            return group.substring(0, lastDotIndex);
        } else {
            return group.toString();
        }
    }

```

warehouse/ingest-core/src/main/java/datawave/ingest/data/normalizer/SimpleGroupFieldNameParser.java
```diff
@@ -35,20 +30,37 @@ public SimpleGroupFieldNameParser(boolean shouldIncludeDelimInGroupName) {
...
     public NormalizedContentInterface extractFieldNameComponents(NormalizedContentInterface origField) {
         String originalIndexedFieldName = origField.getIndexedFieldName();
-        String normParts[] = StringUtils.split(originalIndexedFieldName, '.');
+        String[] splits = StringUtils.split(originalIndexedFieldName, '.');
         int index = originalIndexedFieldName.indexOf('.');
         
-        if (index > -1) {
-            String baseFieldName = originalIndexedFieldName.substring(0, index);
-            String group = originalIndexedFieldName.substring(index + (shouldIncludeDelimInGroupName ? 0 : 1));
+        if (index > 1) {
```
ivakegg: Can't we just test if splits.length > 1 instead of looking for a dot again?

warehouse/ingest-core/src/main/java/datawave/ingest/data/normalizer/SimpleGroupFieldNameParser.java
```diff
@@ -35,20 +30,37 @@ public SimpleGroupFieldNameParser(boolean shouldIncludeDelimInGroupName) {
...
             
-            // we are nested
+            String baseFieldName = null;
+            String group = null;
+            String subgroup = null;
+            
+            baseFieldName = splits[0];
+            group = originalIndexedFieldName.substring(index + 1);
+            
+            if (parentGroupingMatch.matcher(group).matches()) {
```
ivakegg: since we went through the trouble of splitting the string on dots, maybe simply pulling off every thing after the _ on each token is sufficient and then simply merge those tokens back together again for the "group".

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
```diff
@@ -756,6 +758,29 @@ private void setupPreconditionEvaluation(Multimap<String,NormalizedContentInterf
         }
     }
     
+    private boolean eventLacksEdgeFields(EdgeDefinition edgeDef) {
```
ivakegg: Calculating the negation can always be confusing for some and can lead to errors down the line.  Can we change this to containsEdgeFields(edgeDef) ?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
```diff
@@ -599,27 +613,12 @@ public long process(KEYIN key, RawRecordContainer event, Multimap<String,Normali
             
             String enrichmentFieldName = getEnrichmentFieldName(edgeDef);
             
-            Multimap<String,NormalizedContentInterface> mSource = null;
-            Multimap<String,NormalizedContentInterface> mSink = null;
+            Multimap<String,NormalizedContentInterface> mSource = depthFirstList.get(edgeDef.getSourceFieldName());
+            Multimap<String,NormalizedContentInterface> mSink = depthFirstList.get(edgeDef.getSinkFieldName());
             
             String sourceGroup = getGroup(edgeDef.getSourceFieldName());
```
ivakegg: Should we not be using all of that good work we did above to determine the group and subgroup?

ivakegg (changes requested):

warehouse/ingest-core/src/main/java/datawave/ingest/data/normalizer/SimpleGroupFieldNameParser.java
hlgp: the problem is that sometimes we have PARENT_FIELD_NAME, sometimes PARENT_FIELD_NAME_0, and PARENT_0, so we can't be that lax with everything after the last _

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
hlgp: But I don't not like double negatives.  Jk, no problem.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
hlgp: Please review my refactor MR.

warehouse/ingest-core/src/main/java/datawave/ingest/data/normalizer/SimpleGroupFieldNameParser.java
hlgp: Sure.  Not exactly but I  get the gist of what you want.

warehouse/ingest-core/src/main/java/datawave/ingest/data/normalizer/SimpleGroupFieldNameParser.java
ivakegg: I figured I was missing something.  I really don't get this comment.  Can you give me an example?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
ivakegg: The one that is a WIP ?

warehouse/ingest-core/src/main/java/datawave/ingest/data/normalizer/SimpleGroupFieldNameParser.java
ivakegg: oh, I see what you mean.  ok then check for digits as I did in the trim routine.   but we still can use the tokens to rebuild.  Just a suggestion.

warehouse/ingest-core/src/main/java/datawave/ingest/data/normalizer/SimpleGroupFieldNameParser.java
hlgp: Special case where the field name begins with ".", but sure, I could check if the first split is empty.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
hlgp: moved some of those changes into this one

warehouse/ingest-core/src/main/java/datawave/ingest/data/normalizer/SimpleGroupFieldNameParser.java
```diff
@@ -35,20 +31,41 @@ public SimpleGroupFieldNameParser(boolean shouldIncludeDelimInGroupName) {
...
+            String origGroup = null;
+            String group = null;
+            String subgroup = null;
+            
+            baseFieldName = splits[0];
+            origGroup = extractGroup(splits);
+            
+            group = trimGroup(origGroup);
+            
+            if (!trimmed) {
```
ivakegg: won't trimmed always be false here?

ivakegg: I am guessing that trimmed should be set to (!group.equals(origGroup)) ?

ivakegg: Oh, I see what you are doing.  I do not like the parser holding state however.  Especially static state!  That makes this no longer thread safe..  I think you can remove that variable and just test for group.equals(origGroup).

hlgp: No, I tried that first, thanks.  It's not true that the trimmed value will equal the original for this purpose.  I could probably edit the extract and trim methods such that it might... will try.

ivakegg: Then we need to return the trimmed boolean back with the method as well as the resulting group.
Perhaps pass in a StringBuilder for the trimmed result and the method should return a boolean as to whether it was "trimmed".

warehouse/ingest-core/src/main/java/datawave/ingest/data/normalizer/SimpleGroupFieldNameParser.java
```diff
@@ -57,9 +71,55 @@ public NormalizedContentInterface extractFieldNameComponents(NormalizedContentIn
         }
     }
     
+    private String extractGroup(String[] splits) {
```
ivakegg, line 146: perhaps a comment that this is returning all but the first and last split joined together.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
```diff
@@ -637,6 +635,9 @@ public long process(KEYIN key, RawRecordContainer event, Multimap<String,Normali
                     for (String subGroup : commonKeys) {
                         for (NormalizedContentInterface ifaceSource : mSource.get(subGroup)) {
                             for (NormalizedContentInterface ifaceSink : mSink.get(subGroup)) {
+                                if (ifaceSource.equals(ifaceSink)) {
```
ivakegg: How can this happen?  Would this only happen if we have an edge with the same source and sink?  Would this happen if the value of the field is the same on both sides (in which case this may be ok)?  This may be ok I am just trying to understand why.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
```diff
@@ -756,6 +757,26 @@ private void setupPreconditionEvaluation(Multimap<String,NormalizedContentInterf
         }
     }
     
+    private boolean eventContainsEdgeFields(EdgeDefinition edgeDef) {
+        
+        if (depthFirstList.containsKey(edgeDef.getSourceFieldName()) && depthFirstList.containsKey(edgeDef.getSinkFieldName())) {
+            mSource = depthFirstList.get(edgeDef.getSourceFieldName());
```
ivakegg: Again with the class member variables.  Can we not pass these back via the method parameters instead?  Doing it this way we are creating a method with side effects that are not obvious and also prevents thread safety if we ever need that with this class down the line.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
```diff
@@ -809,6 +830,9 @@ protected EdgeDataBundle createEdge(EdgeDefinition edgeDef, RawRecordContainer e
         }
         
         String typeName = event.getDataType().typeName();
+        String outputName = event.getDataType().outputName();
+        
+        typeName = edgeTypeLookup.containsKey(outputName) ? outputName : typeName;
```
ivakegg: This sequence looks similar to the one above.  Perhaps refactor into a common method.

ivakegg (changes requested):

warehouse/ingest-core/src/main/java/datawave/ingest/data/normalizer/SimpleGroupFieldNameParser.java
hlgp: Sure

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
hlgp: We simply take the product of the source and sink fields.  In the case where we are trying to make edges between multiple instances of the same field (see the updated test tvmaze-edge-definition for an example), we were previously including the main diagonal if the source field name and group matched in the edge definition.  

This would only happen if all elements of the NCI are identical, not just if the values are equal.  We'd still make edges if the group or subgroup are different.

This comment made me realize I was still testing some of that. Thanks!

Now that I think about it, though, we should probably fix some of the processors before implementing this since they are not always subgroup smart...  But until then we will still produce duplicates edges to the field itself.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
hlgp: Yeah, assigning them in here felt funky, but I guess I misinterpreted what you wanted.  I've got no problem changing it and it's not that I don't agree with you, but look at the top of this class at all the class member variables I did not create.  No one should use this class thinking it's thread safe.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
hlgp: Sure

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
ivakegg: ok, I see what you mean.  But then at least good comments denoting the side effect.

The method name does not suggest that anything else is happening

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
ivakegg: ah, I see. OK, I guess it is possible then.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
```diff
@@ -558,11 +559,31 @@ public long process(KEYIN key, RawRecordContainer event, Multimap<String,Normali
...
+            // don't bother evaluating preconditions if we know this event doesn't have the necessary fields for this edge
+            if (depthFirstList.containsKey(edgeDef.getSourceFieldName()) && depthFirstList.containsKey(edgeDef.getSinkFieldName())) {
+                mSource = depthFirstList.get(edgeDef.getSourceFieldName());
+                mSink = depthFirstList.get(edgeDef.getSinkFieldName());
+            } else {
+                continue;
+            }
+            
+            // bail if the event doesn't contain any values for the source or sink field
+            if (null == mSource || null == mSink) {
```
ivakegg, line 588: So.... why check for containsKey above if this check would catch the case where one of the fields does not exist?

ivakegg (dismissed):

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
ivakegg: Also I think CollectionUtils.isEmpty(collection) will take care of both == null and .isEmpty().  A little more succinct.

hlgp: This is the old code, just moved up so we're not needlessly checking the jexl evaluation on every record.  Good suggestions, I was just going to save this kind of cleanup for the refactor MR.

warehouse/ingest-core/src/main/java/datawave/ingest/data/normalizer/SimpleGroupFieldNameParser.java
```diff
@@ -17,16 +19,9 @@
 public class SimpleGroupFieldNameParser {
     private static final long serialVersionUID = 1035918631638323565L;
     private static final Logger log = Logger.getLogger(SimpleGroupFieldNameParser.class);
-    private final boolean shouldIncludeDelimInGroupName;
+    private static final String DOT = ".";
```
jwomeara: Minor nit - if you make this `private static final char DOT = '.';` instead, you can replace all of the `'.'` instances in this code as well as the `"."` instances.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
jwomeara: Do you need to check for null or empty at all?  Unless we are removing values from depthFirstList somewhere, in the process method we only ever add an entry if there is a valid key & value.  Might be able to drop both checks and just keep the one you added.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
```diff
@@ -637,6 +648,9 @@ public long process(KEYIN key, RawRecordContainer event, Multimap<String,Normali
                     for (String subGroup : commonKeys) {
                         for (NormalizedContentInterface ifaceSource : mSource.get(subGroup)) {
                             for (NormalizedContentInterface ifaceSink : mSink.get(subGroup)) {
+                                if (ifaceSource == ifaceSink) {
+                                    continue;
```
jwomeara, line 652: Just making sure I'm reading this correctly - this (and the continue added below) is where you have fixed the whole mapping to itself issue, right?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
```diff
@@ -956,13 +982,18 @@ private NormalizedContentInterface getNullKeyedNCI(String fieldValue, Multimap<S
     
     protected String getGroupedFieldName(NormalizedContentInterface value) {
         String fieldName = value.getIndexedFieldName();
+        String group;
```
jwomeara: Any reason not to move this up into the `if (!grouped.getGroup().isEmpty()) {` block?

warehouse/ingest-json/src/test/resources/config/ingest/tvmaze-edge-definitions.xml
```diff
@@ -34,180 +34,16 @@ http://www.springframework.org/schema/util/spring-util-4.0.xsd">
                                in the graph.
```
jwomeara: This comment above is no longer valid, right?

warehouse/ingest-json/src/test/resources/config/ingest/tvmaze-edge-definitions.xml
```diff
@@ -34,180 +34,16 @@ http://www.springframework.org/schema/util/spring-util-4.0.xsd">
...
                              -->
                             <bean class="datawave.ingest.mapreduce.handler.edge.define.EdgeNode">
```
jwomeara, line 32: You have this edge defined twice in this list.  Was that intentional?

jwomeara (changes requested):
tricky pr.  nice work.  only a few minor nits.

warehouse/ingest-core/src/main/java/datawave/ingest/data/normalizer/SimpleGroupFieldNameParser.java
hlgp: I think that was the goal and I just forgot about it, thanks.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
hlgp: yep

warehouse/ingest-json/src/test/resources/config/ingest/tvmaze-edge-definitions.xml
hlgp: ah, yes, I should update that

warehouse/ingest-json/src/test/resources/config/ingest/tvmaze-edge-definitions.xml
hlgp: Yep. You have to define a source and a sink, even if it's from the same field to itself.

jwomeara: do we need to do the same under the `group2` property?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
hlgp: nope

warehouse/ingest-json/src/test/resources/config/ingest/tvmaze-edge-definitions.xml
hlgp: No, since that's making edges from fields in group1 to fields in group2

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
hlgp: Yep.  Planning to clean this kind of thing up in the refactor MR.  I still want to short circuit since the precondition evaluation is expensive even when the event lacks the edge fields.

jwomeara (approved):

ivakegg (approved):

PR #1914, opened 2023-04-04
Add aggregation functionality for #GROUP_BY function

### LINKED ISSUE #1687: Provide the ability to aggregate fields when grouping
The basic ask is to be able to sum up a field and return that sum when using the grouping functionality.  So perhaps the query would be something like 
... AND #GROUPBY(FIELD1) AND #SUM(FIELD2)
In this case the results would be grouped by FIELD1, and FIELD2 would also be returned with its sum within each group.

This could be extended to do other aggregations such as MAX, MIN, COUNT, and AVERAGE.

lbschanno:
Provide the ability to aggregate fields when grouping. Specifically, add the new functions #SUM, #MIN, #MAX, #COUNT, #AVERAGE that will determine the aggregate value for any specified fields against the entries that match any particular grouping.

Resolves #1687


warehouse/query-core/src/main/java/datawave/query/common/grouping/GroupingAttribute.java
```diff
@@ -0,0 +1,58 @@
...
+        if (this.isMetadataSet() == other.isMetadataSet()) {
+            if (this.isMetadataSet()) {
+                return this.metadata.compareRow(other.getMetadata().getRow()) == 0;
+            }
+        }
+        return false;
+    }
+    
+    /**
+     * Returns a hashcode of the attribute's value.
```
avgAGB: Really small thing, could you add a @return notation for the hashCode javadoc?

lbschanno: Not a problem, I just added a bunch of documentation that included this change.

warehouse/query-core/src/main/java/datawave/query/common/grouping/AggregatedFields.java
```diff
@@ -0,0 +1,312 @@
...
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * This class provides functionality to aggregate values for specified target fields using specified aggregation operations.
+ */
+public class AggregatedFields {
+    
+    private final Multimap<String,Aggregator<?>> aggregatorMap;
```
jwomeara: Thoughts on updating this to be Map<String, Map<String, Aggregator>> keyed by both the field, and the aggregator enum name?  Might make the merge operation faster.

warehouse/query-core/src/main/java/datawave/query/common/grouping/AggregatedFields.java
```diff
@@ -0,0 +1,312 @@
...
+         * @return a new {@link Factory} instance with deconstructed fields
+         */
+        public Factory deconstructIdentifiers() {
+            Set<String> sumFields = this.sumFields.stream().map(JexlASTHelper::deconstructIdentifier).collect(Collectors.toSet());
+            Set<String> maxFields = this.maxFields.stream().map(JexlASTHelper::deconstructIdentifier).collect(Collectors.toSet());
+            Set<String> minFields = this.minFields.stream().map(JexlASTHelper::deconstructIdentifier).collect(Collectors.toSet());
+            Set<String> countFields = this.countFields.stream().map(JexlASTHelper::deconstructIdentifier).collect(Collectors.toSet());
+            Set<String> averageFields = this.averageFields.stream().map(JexlASTHelper::deconstructIdentifier).collect(Collectors.toSet());
+            return new Factory().withSumFields(sumFields).withMaxFields(maxFields).withMinFields(minFields).withCountFields(countFields)
+                            .withAverageFields(averageFields);
```
jwomeara: Do we not want to deconstruct the identifiers when fields are added in the first place?

warehouse/query-core/src/main/java/datawave/query/common/grouping/AverageAggregator.java
```diff
@@ -0,0 +1,151 @@
...
+        } else {
+            numerator = numerator.add(number);
+            divisor = divisor.add(BigDecimal.ONE);
+        }
+        average = numerator.divide(divisor, MATH_CONTEXT);
+        columnVisibilities.add(value.getColumnVisibility());
+    }
+    
+    @Override
+    public void merge(Aggregator<?> other) {
```
jwomeara, line 140: While reading this code, I had a brief concern about the potential for concurrent calls to merge in the pipeline iterator.  I know that we are currently not running with more than one evaluation pipeline, but if we did, would this code still work?  Actually, would grouping work at all?  Are we testing this somewhere already?

warehouse/query-core/src/main/java/datawave/query/common/grouping/CountAggregator.java
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Determines the total count of aggregated field values. This supports values of all {@link Attribute} types.
+ */
+public class CountAggregator extends AbstractAggregator<Integer> {
+    
+    /**
+     * The total number of times the field was seen.
+     */
+    private int count;
```
jwomeara: int, really?

warehouse/query-core/src/main/java/datawave/query/common/grouping/Groups.java
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Represents a set of groups found during a #GROUP_BY operation.
+ */
+public class Groups {
+    
+    /**
+     * A map of distinct grouping values to their groups.
+     */
+    private final Map<Set<GroupingAttribute<?>>,Group> groups = new HashMap<>();
```
jwomeara: A Set seems like an expensive object to use as a key in a hashmap.  Each time it calls hashCode on the set (at a minimum every time we call `get` on the hashmap), it iterates through every key value pair in the set and computes the hashcode.

warehouse/query-core/src/main/java/datawave/query/iterator/GroupingIterator.java
```diff
@@ -71,27 +75,17 @@ public class GroupingIterator implements Iterator<Map.Entry<Key,Document>> {
...
-    /**
-     * Length of time in milliseconds that a client will wait for a results to be returned. If a result is not collected before the timeout, a key with an
-     * "intermediate" document will be returned.
-     */
-    private final long resultTimeout;
-    
-    public GroupingIterator(Iterator<Map.Entry<Key,Document>> previousIterators, MarkingFunctions markingFunctions, Collection<String> groupFieldsSet,
-                    int groupFieldsBatchSize, YieldCallback<Key> yieldCallback, long resultTimeout) {
+    public GroupingIterator(Iterator<Map.Entry<Key,Document>> previousIterators, MarkingFunctions markingFunctions, Collection<String> groupFields,
+                    int groupFieldsBatchSize, YieldCallback<Key> yieldCallback, AggregatedFields.Factory aggregateFieldsFactory) {
```
jwomeara: Did we lose the functionality that utilizes the resultTimeout?

warehouse/query-core/src/main/java/datawave/query/language/builder/jexl/JexlTreeBuilder.java
```diff
@@ -63,7 +68,7 @@ public class JexlTreeBuilder extends QueryTreeBuilder {
             new DateFunction(), new OccurrenceFunction(), new MatchesInGroupFunction(), new MatchesInGroupLeft(), new GetAllMatches(),
             new MatchesAtLeastCountOf(), new Jexl(), new TimeFunction(), new AtomValuesMatchFunction(), new Options(), new Unique(), new GroupBy(),
             new EvaluationOnly(), new UniqueByDay(), new UniqueByHour(), new UniqueByTenthOfHour(), new UniqueByMonth(), new UniqueByYear(),
-            new UniqueByMinute(), new NoExpansion(), new Compare(), new ExcerptFields()};
+            new UniqueByMinute(), new NoExpansion(), new Compare(), new ExcerptFields(), new Sum(), new Min(), new Max(), new Average(), new Count()};
```
jwomeara: These should be added to allowedQueryFunctions in QueryLogicFactory.xml as well.

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -2049,7 +2143,11 @@ public Set<String> getOptionalQueryParameters() {
         optionalParams.add(datawave.webservice.query.QueryParameters.QUERY_PAGESIZE);
         optionalParams.add(datawave.webservice.query.QueryParameters.QUERY_PAGETIMEOUT);
         optionalParams.add(datawave.webservice.query.QueryParameters.QUERY_EXPIRATION);
-        optionalParams.add(datawave.webservice.query.QueryParameters.QUERY_MAX_RESULTS_OVERRIDE);
```
jwomeara, line 2102: Was this removal intentional?

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
jwomeara, line 1: This class seems to share some code with GroupingIterator.  Any chance we can identify the common code and pull it out to a common location?

warehouse/query-core/src/test/java/datawave/query/transformer/GroupingTest.java
jwomeara, line 1: Lots of changes in here - I admit that I only skimmed.  Did we remove the teardown aspects of these tests though?

jwomeara (changes requested):
Overall, looks really good.  Just a few comments throughout.  

There were a few places where it seemed like the logic Ceria added to return a blank page after a timeout might have been lost, but maybe it was just moved and I had trouble tracking.

One overall design thought - do we want to be able to impose limits on which fields can be aggregated?  A configurable allow list perhaps?

Because of how extensive these changes are, we should really put this through it's paces before merging IMO.

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
lbschanno: It was not, I think this got missed when resolving some conflicts with updates from integration. I'll put it back in.

warehouse/query-core/src/main/java/datawave/query/transformer/GroupingTransform.java
lbschanno: No problem, I'll do so.

warehouse/query-core/src/main/java/datawave/query/common/grouping/CountAggregator.java
lbschanno: Can you please elaborate on your specific concerns and what you would like to see here instead?

warehouse/query-core/src/main/java/datawave/query/common/grouping/AverageAggregator.java
lbschanno: As far as I know, we're not testing grouping with concurrent calls.

warehouse/query-core/src/main/java/datawave/query/common/grouping/AggregatedFields.java
lbschanno: Agreed, I'll see what I can do here to improve it.

warehouse/query-core/src/main/java/datawave/query/common/grouping/AggregatedFields.java
lbschanno: We should, I'll make the change.

warehouse/query-core/src/main/java/datawave/query/common/grouping/Groups.java
lbschanno: This is something I've also wanted to improve upon, it's how the original code handled finding groupings. I'll work on coming up with a more efficient key object to use instead of a set.

warehouse/query-core/src/main/java/datawave/query/language/builder/jexl/JexlTreeBuilder.java
lbschanno: Will do.

warehouse/query-core/src/main/java/datawave/query/iterator/GroupingIterator.java
lbschanno: It looks like resultTimeout was added in commit 3b7a68004289971232b7833206156db5b94fa2e4 by Ivan, but no functionality was added that actually used it. 

@ivakegg, was there supposed to be functionality added to GroupingIterator for resultTimeout?

warehouse/query-core/src/test/java/datawave/query/transformer/GroupingTest.java
lbschanno: No, the teardowns and interrupts are still used. The function `collectQueryResults()` will run the query under test against each permutation of the teardown/interrupts and collect the results for validations. I simplified the creation of the set of teardowns and interrupts via `EnumSet.allOf()`.

warehouse/query-core/src/main/java/datawave/query/language/builder/jexl/JexlTreeBuilder.java
lbschanno: After double-checking, these were already added to allowedQueryFunctions in QueryLogicFactory.xml.

lbschanno:
@jwomeara 
> One overall design thought - do we want to be able to impose limits on which fields can be aggregated? A configurable allow list perhaps?

I think @ivakegg will need to be the one to decide if that's a requirement for this work. If so, it should be relatively simple to implement.

warehouse/query-core/src/main/java/datawave/query/common/grouping/CountAggregator.java
jwomeara: just wondering if we ever expect to overrun an int while counting.  in that case, a long might be a better choice.  i'll leave that decision to you and Ivan.

warehouse/query-core/src/main/java/datawave/query/language/builder/jexl/JexlTreeBuilder.java
jwomeara: I noticed that as well and forgot to delete this comment.  My apologies.

lbschanno: No worries.

warehouse/query-core/src/main/java/datawave/query/common/grouping/CountAggregator.java
lbschanno: It's not an unreasonable concern to have, and I don't mind making it a long to help safeguard against overruns.

warehouse/query-core/src/main/java/datawave/query/iterator/GroupingIterator.java
ivakegg: There is actually a story behind that one.  The grouping mechanism essentially cannot return results to the user until all results on the back end have been gathered.  We were working the ability to allow the iterator stack to return all the way back to the RunningQuery which could then return an empty page to the user and subsequently continue the query with a new next call.  This would prevent the query from timing out.  So to that end this timeout was supposed to work in concert with the timeout in the grouping transform which is used webserver side.
  You will see that the timeout mechanism was actually implemented in the GroupingTransform which is used webserver side.  Then we realized that the RunningQuery could essentially just run the back end "next" call in a separate thread and manage the returning of an empty results page back to the user without having to modify the grouping logic.
  The timeout being passed to the GroupingIterator is a leftover of that.  I was going to go back and actually use that to allow the iterator stack to return and be torn down and rebuilt to avoid scan thread starvation.  However Bill is working that problem in a different way so that parameter can be removed.

ivakegg:
> 

If there are any restriction on fields to be grouped on or fields to be summed/aggregated, that would be done by the webserver.  Currently there are no restrictions as such and no requirements for any.

warehouse/query-core/src/main/java/datawave/query/iterator/GroupingIterator.java
lbschanno: Thanks for the background info, I'll leave the parameter out.

warehouse/query-core/src/main/java/datawave/query/common/grouping/CountAggregator.java
lbschanno: Done.

warehouse/query-core/src/main/java/datawave/query/common/grouping/AggregatedFields.java
lbschanno: Done.

warehouse/query-core/src/main/java/datawave/query/common/grouping/AggregatedFields.java
lbschanno: Done.

warehouse/query-core/src/main/java/datawave/query/common/grouping/AverageAggregator.java
ivakegg: The grouping is being done after the pipeline iterator so no need for thread safety here.

ivakegg (dismissed):

ivakegg (dismissed):

jwomeara (approved):

warehouse/query-core/src/main/java/datawave/query/common/grouping/Groups.java
jwomeara: I approved, but I just want to double check that we are going to proceed with a set as the key for this map.  Just wondering where your thought process lead you with this one.

jwomeara (dismissed):
One minor comment left unresolved, but otherwise looks great!

warehouse/query-core/src/main/java/datawave/query/common/grouping/Groups.java
lbschanno: So I found myself going back and forth on this, trying to figure out a workable solution that wasn't effectively just a wrapper around a Set. One approach we could try is an immutable HashSet implementation that computes the hashcode once at instantiation, retaining that as an attribute to be returned from hashCode(), and using that as the map key. Thoughts?

jwomeara: I am ok with whatever you decide - i just wanted to hear your thoughts.  If you're ok with it as-is, then we can move forward with that.

lbschanno: I think I'm okay with it as is, but I also don't really have any data on how large these grouping sets can get, nor how large the map keyset can get either, so I can't really say if this optimization is necessary or not. If we want to take a safer approach, we should try to optimize it. If we want to wait and see, it can stay as is.

lbschanno: @jwomeara I pushed a commit that optimizes the hashcode() usage, please take a look if you get a chance and let me know what you think.

ivakegg (dismissed):

ivakegg (dismissed):

ivakegg (approved):

hlgp (approved):

apmoriarty (approved):

PR #1916, opened 2023-04-04
WIP: 1913: add a check for unfielded queries

wgtong:


warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -1641,6 +1641,34 @@ public String getFormattedViolations() {
         }
     }
     
+    public static void addUnfieldedQueriesToList(List<String> unfieldedQueries, JexlNode node) {
+        if (node.jjtGetNumChildren() > 0) {
+            for (int i = 0; i < node.jjtGetNumChildren(); i++) {
+                addUnfieldedQueriesToList(unfieldedQueries, node.jjtGetChild(i));
+            }
+        } else if (node instanceof ASTStringLiteral) {
+            // Go to grandparent and check grandchildren for "Identifiers"
```
wgtong, line 1650: Inspect this logic closely as i'm not well versed in the possible JEXL queries

wgtong:
As this is, there are a couple of Edge dependencies which will break. (Edge classes which pass in lists without specifying the list syntax). This should be held off until the fixes for List syntax queries is in

hlgp:
Stale.  Can be reopened when active.

PR #1927, opened 2023-04-11
* Refactored Document and TimingMetadata to not extend Attribute

ivakegg:
* Simplified any logic looking for nested documents


warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
```diff
@@ -24,136 +20,63 @@
...
+        metadata.setMetadata(key);
     }
     
-    public ColumnVisibility getColumnVisibility() {
-        if (isMetadataSet()) {
-            return metadata.getColumnVisibilityParsed();
-        }
-        return Constants.EMPTY_VISIBILITY;
+    public boolean isMetadataSet() {
+        return (metadata.isMetadataSet());
```
apmoriarty: extra parens

apmoriarty (dismissed):
LGTM. Nice cleanup.

warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
ivakegg: removed

apmoriarty (dismissed):

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -1282,17 +1281,18 @@ public boolean validateOptions(Map<String,String> options) {
...
         if (options.containsKey(PROJECTION_FIELDS)) {
             this.projectResults = true;
-            this.useWhiteListedFields = true;
             
             String fieldList = options.get(PROJECTION_FIELDS);
-            if (fieldList != null && EVERYTHING.equals(fieldList)) {
-                this.whiteListedFields = UniversalSet.instance();
-            } else if (fieldList != null && !fieldList.trim().equals("")) {
+            if (fieldList != null && !EVERYTHING.equals(fieldList) && !fieldList.trim().equals("")) {
+                this.useWhiteListedFields = true;
```
jwomeara: I'm a little confused on this one.  We used to use UniversalSet if the field list was *, now we don't do that.  We only set the whitelisted fields if we have a non-`*`, non-empty list of fields.  So I guess we're saying that is the field list is set to `*`, we want to instead go down the blacklisted fields path with an empty set of fields.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -1282,17 +1281,18 @@ public boolean validateOptions(Map<String,String> options) {
...
                 Collections.addAll(this.whiteListedFields, StringUtils.split(fieldList, Constants.PARAM_VALUE_SEP));
-            }
-            if (options.containsKey(HIT_LIST) && Boolean.parseBoolean(options.get(HIT_LIST))) {
-                this.whiteListedFields.add(JexlEvaluation.HIT_TERM_FIELD);
+                if (options.containsKey(HIT_LIST) && Boolean.parseBoolean(options.get(HIT_LIST))) {
+                    this.whiteListedFields.add(JexlEvaluation.HIT_TERM_FIELD);
+                }
+            } else {
+                this.useBlackListedFields = true;
+                this.blackListedFields = new HashSet<>();
```
jwomeara: I don't think you need this.  blackListedFields initial value is `new HashSet<>()`  It's actually set on line 1310 below as well.  that wasn't your addition, but that can also be removed.

warehouse/query-core/src/main/java/datawave/query/predicate/EmptyDocumentFilter.java
```diff
@@ -21,7 +21,7 @@ public class EmptyDocumentFilter implements Predicate<Entry<Key,Document>> {
     
     @Override
     public boolean apply(Entry<Key,Document> input) {
-        boolean nonempty = (input.getValue().size() > 0);
+        boolean nonempty = (!input.getValue().isEmpty() || input.getValue().hasTimingMetadata());
```
jwomeara, line 24: I am not picking up the significance of having timing metadata, and how that changes this operation, but it's also late on a friday.  Can you explain a bit?

jwomeara (changes requested):
I kind of rushed through this at the end once i got to the tests, but I had a few comments.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
ivakegg: That is correct.  The UniversalSet mechanism was actually messing things up.  So instead a whitelist of * is the equivalent of an empty blacklist.  Although I suppose I could change this to avoid applying a projection at all in that case.  I will change it to do that instead.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
ivakegg: ah, ok.  Thanks.

warehouse/query-core/src/main/java/datawave/query/predicate/EmptyDocumentFilter.java
ivakegg: Previously the TimeingMetadata was a Document that was being added to the dict making the document non-empy.  Basically it is valid to have a document with only timing metadata and that should not be considered an empty document in this case.

apmoriarty (dismissed):

jwomeara (dismissed):

billoley:
In DocumentProjection, the following constructor arguments and class memebers are no longer used.  Is this intentional?
If so, they should probably be removed from the constructor and the class.

    private final boolean includeGroupingContext;
    private final boolean reducedResponse;    
    private boolean trackSizes = true;

    public DocumentProjection(boolean includeGroupingContext, boolean reducedResponse, boolean trackSizes, Set<String> projections, Projection.ProjectionType projectionType) {
        this.includeGroupingContext = includeGroupingContext;         // no longer used
        this.reducedResponse = reducedResponse;                            // no longer used
        this.projection = new Projection(projections, projectionType);
        this.trackSizes = trackSizes;                                                      // no longer used
    }

ivakegg:
@billoley Cleaned up those members and constructors.  Thanks!

ivakegg:
Some bedrock review:
 1. TimingMetadata Writable serialization still writes the wrong map                                                                                                                                                                
                                                                                                                                                                                                                                    
 File: warehouse/query-core/src/main/java/datawave/query/attributes/TimingMetadata.java:157-173                                                                                                                                     
                                                                                                                                                                                                                                    
 The method accepts a map parameter:                                                                                                                                                                                                
                                                                                                                                                                                                                                    
 ```java                                                                                                                                                                                                                            
   private void write(Map<String,Long> metadata, DataOutput out)                                                                                                                                                                    
 ```                                                                                                                                                                                                                                
                                                                                                                                                                                                                                    
 but ignores it:                                                                                                                                                                                                                    
                                                                                                                                                                                                                                    
 ```java                                                                                                                                                                                                                            
   WritableUtils.writeVInt(out, this.metadata.size());                                                                                                                                                                              
                                                                                                                                                                                                                                    
   for (Map.Entry<String,Long> entry : this.metadata.entrySet()) {                                                                                                                                                                  
 ```                                                                                                                                                                                                                                
                                                                                                                                                                                                                                    
 It is called as:                                                                                                                                                                                                                   
                                                                                                                                                                                                                                    
 ```java                                                                                                                                                                                                                            
   write(metadata, out);                                                                                                                                                                                                            
   write(stageTimers, out);                                                                                                                                                                                                         
 ```                                                                                                                                                                                                                                
                                                                                                                                                                                                                                    
 Therefore, the second map written is metadata again, not stageTimers.                                                                                                                                                              
                                                                                                                                                                                                                                    
 A Writable round trip will deserialize the ordinary metadata map as the stage-timer map. This affects the Hadoop serialization path, while the Kryo path appears to use the correct map.                                           
                                                                                                                                                                                                                                    
 Fix:                                                                                                                                                                                                                               
                                                                                                                                                                                                                                    
 ```java                                                                                                                                                                                                                            
   WritableUtils.writeVInt(out, metadata.size());                                                                                                                                                                                   
                                                                                                                                                                                                                                    
   for (Map.Entry<String,Long> entry : metadata.entrySet()) {                                                                                                                                                                       
 ```                                                                                                                                                                                                                                
                                                                                                                                                                                                                                    
 A round-trip test containing distinct values in both maps should be added.                                                                                                                                                         
                                                                                                                                                                                                                                    
 ────────────────────────────────────────────────────────────────────────────────                                                                                                                                                   
                                                                                                                                                                                                                                    
 2. TimingMetadata.hashCode() still throws for the normal default state                                                                                                                                                             
                                                                                                                                                                                                                                    
 File: TimingMetadata.java:104-107                                                                                                                                                                                                  
                                                                                                                                                                                                                                    
 ```java                                                                                                                                                                                                                            
   return metadata.hashCode() + stageTimers.hashCode() + host.hashCode();                                                                                                                                                           
 ```                                                                                                                                                                                                                                
                                                                                                                                                                                                                                    
 host is nullable and defaults to null. Thus:                                                                                                                                                                                       
                                                                                                                                                                                                                                    
 ```java                                                                                                                                                                                                                            
   new TimingMetadata().hashCode()                                                                                                                                                                                                  
 ```                                                                                                                                                                                                                                
                                                                                                                                                                                                                                    
 throws NullPointerException.                                                                                                                                                                                                       
                                                                                                                                                                                                                                    
 This also means any collection or containing object that calculates the hash code of a default TimingMetadata can fail.                                                                                                            
                                                                                                                                                                                                                                    
 Use a null-safe implementation, for example:                                                                                                                                                                                       
                                                                                                                                                                                                                                    
 ```java                                                                                                                                                                                                                            
   return Objects.hash(metadata, stageTimers, host);                                                                                                                                                                                
 ```                                                                                                                                                                                                                                
                                                                                                                                                                                                                                    
 ────────────────────────────────────────────────────────────────────────────────                                                                                                                                                   
                                                                                                                                                                                                                                    
 3. Nested document projections are still incorrect                                                                                                                                                                                 
                                                                                                                                                                                                                                    
 File: warehouse/query-core/src/main/java/datawave/query/function/DocumentProjection.java:65-79                                                                                                                                     
                                                                                                                                                                                                                                    
 The new implementation applies the projection only to the immediate children of the supplied document:                                                                                                                             
                                                                                                                                                                                                                                    
 ```java                                                                                                                                                                                                                            
   for (Entry<String,Attribute<? extends Comparable<?>>> entry : dict.entrySet()) {                                                                                                                                                 
       String fieldName = entry.getKey();                                                                                                                                                                                           
       Attribute<?> attr = entry.getValue();                                                                                                                                                                                        
       attr.setToKeep(attr.isToKeep() && projection.apply(fieldName));                                                                                                                                                              
   }                                                                                                                                                                                                                                
 ```                                                                                                                                                                                                                                
                                                                                                                                                                                                                                    
 It does not recursively process nested Document or Attributes values.                                                                                                                                                              
                                                                                                                                                                                                                                    
 For a document such as:                                                                                                                                                                                                            
                                                                                                                                                                                                                                    
 ```text                                                                                                                                                                                                                            
   PERSON = {                                                                                                                                                                                                                       
       NAME = "Alice",                                                                                                                                                                                                              
       AGE = 42                                                                                                                                                                                                                     
   }                                                                                                                                                                                                                                
 ```                                                                                                                                                                                                                                
                                                                                                                                                                                                                                    
 a projection intended to retain PERSON.NAME cannot selectively remove AGE. The projection is evaluated only against PERSON, and the nested children retain their previous toKeep values.                                           
                                                                                                                                                                                                                                    
 The previous implementation explicitly recursed through nested documents and attribute collections. This remains a regression.                                                                                                     
                                                                                                                                                                                                                                    
 There is also a semantic change: apply() now mutates the original document and returns it, whereas the prior implementation created a projected copy. Callers that reuse the original document may observe fields unexpectedly     
 removed after projection.                                                                                                                                                                                                          
                                                                                                                                                                                                                                    
 ────────────────────────────────────────────────────────────────────────────────                                                                                                                                                   
                                                                                                                                                                                                                                    
 4. Document.removeAll() no longer removes nested fields                                                                                                                                                                            
                                                                                                                                                                                                                                    
 File: warehouse/query-core/src/main/java/datawave/query/attributes/Document.java:507-527                                                                                                                                           
                                                                                                                                                                                                                                    
 The current implementation only removes matching direct entries. The previous implementation recursively called removeAll() on nested Document values.                                                                             
                                                                                                                                                                                                                                    
 For example:                                                                                                                                                                                                                       
                                                                                                                                                                                                                                    
 ```java                                                                                                                                                                                                                            
   document.removeAll("FIELD");                                                                                                                                                                                                     
 ```                                                                                                                                                                                                                                
                                                                                                                                                                                                                                    
 will remove a top-level FIELD, but leave a FIELD contained inside a nested document. This is a behavior change in a public document-manipulation operation.                                                                        
                                                                                                                                                                                                                                    
 If recursion is intentionally being removed, the method should be renamed or its contract/documentation changed. Otherwise, the prior recursive behavior should be restored.                                                       
                                                                                                                                                                                                                                    
 ────────────────────────────────────────────────────────────────────────────────                                                                                                                                                   
                                                                                                                                                                                                                                    
 5. Timing metadata serialization is not backward compatible                                                                                                                                                                        
                                                                                                                                                                                                                                    
 The PR replaces the old Metadata/Document representation of timing data with a standalone map-based TimingMetadata representation and changes both Writable and Kryo layouts.                                                      
                                                                                                                                                                                                                                    
 Old serialized timing metadata cannot be read by the new implementation, and vice versa. This is potentially an issue during rolling upgrades or when serialized documents are passed between components running different         
 versions.                                                                                                                                                                                                                          
                                                                                                                                                                                                                                    
 This may be acceptable if all query components are upgraded atomically, but it should be explicitly documented and covered by compatibility testing if mixed-version operation is supported.                                       
                                                                                                                                                                                                                                    
 Recommendation                                                                                                                                                                                                                     
                                                                                                                                                                                                                                    
 The compilation fixes address the original build blocker, but I would still request changes for items 1–4 before merging. The most immediately damaging runtime bugs are the Writable stage-timer serialization error and the      
 non-recursive projection behavior.

PR #1929, opened 2023-04-13
Only run setup handlers for active Type(s)

FineAndDandy:


apmoriarty (dismissed):
No test?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
```diff
@@ -538,7 +539,20 @@ public int run(String[] args) throws Exception {
     }
     
     private void setupHandlers(Configuration conf) {
-        for (Type t : TypeRegistry.getTypes()) {
+        // default to all types
+        Collection<Type> types = TypeRegistry.getTypes();
+        
+        // if a datatype is defined for this job, setup it only
+        String dataType = conf.get(DataTypeHelper.Properties.DATA_NAME);
```
ivakegg: I believe this property will always be set to something, albeit random when all of the configuration files are loaded.  I am pretty sure this will break the other system.

ivakegg (changes requested):
I don't this this will work for everybody.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/IngestJob.java
ivakegg: When using the event sequence reader, the data type is set based on the Event, not the data.type property.

FineAndDandy: I'm okay with setting a flag for initializing all or just the active datatype

ivakegg (approved):

apmoriarty (approved):

PR #1933, opened 2023-04-15
Feature/limit to matching fields

ivakegg:
This feature will allow the user to specify a set of matching fields that in combination with the limit fields will force additional field groups to be included in the returned event.  The idea is that if you have some fields that match across field groups, then the user will want those returned as well as the field group that hit in the query.


warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
```diff
@@ -9,25 +9,66 @@
...
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.stream.Collectors;
 
+//@formatter:off
```
apmoriarty: Java docs?

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1018,10 +1018,12 @@ public Entry<DocumentData,Document> apply(@Nullable Entry<Key,Document> input) {
...
             if (gatherTimingDetails()) {
-                documents = Iterators.transform(documents,
-                                new EvaluationTrackingFunction<>(QuerySpan.Stage.LimitFields, trackingSpan, new LimitFields(this.getLimitFieldsMap())));
+                documents = Iterators.transform(
+                                documents,
+                                new EvaluationTrackingFunction<>(QuerySpan.Stage.LimitFields, trackingSpan, new LimitFields(this.getLimitFieldsMap(), this
+                                                .getMatchingFieldSets())));
             } else {
-                documents = Iterators.transform(documents, new LimitFields(this.getLimitFieldsMap()));
+                documents = Iterators.transform(documents, new LimitFields(this.getLimitFieldsMap(), this.getMatchingFieldSets()));
```
apmoriarty: Should limit fields be encapsulated? Then you could have something like
```
documents = Iterators.transform(documents, getLimitFields());
```

getLimitFields could look like this
```
getLimitFields() {
    return new LimitFields(this.getLimitFieldsMap(), this.getMatchingFieldSets())
}

# or..
getLimitFields() {
if ( limitFields == null){
    limitFields = new LimitFields(this.getLimitFieldsMap(), this.getMatchingFieldSets());
}
return limitFields
```

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -533,6 +533,7 @@ private void configureIterator(ShardQueryConfiguration config, IteratorSetting c
         setCommonIteratorOptions(config, cfg);
         
         addOption(cfg, QueryOptions.LIMIT_FIELDS, config.getLimitFieldsAsString(), true);
+        addOption(cfg, QueryOptions.MATCHING_FIELD_SETS, config.getMatchingFieldSetsAsString(), true);
```
apmoriarty: The last argument should be false. Otherwise we will serialize the key with an empty value for every range generated. In fact...shouldn't most of these options have flag set to false?

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
ivakegg: right.  I will fix those comments.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
ivakegg: hmm.... I guess I don't have a problem with that but what does that help us with?

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
ivakegg: ah, ok.

ivakegg:
TODO:
1) ensure that the fields in the matching fields sets are added to the document (i.e. return.fields may have stripped them out)
2) expand the matching field sets to include model expansions
3) handling normalizations (optional)

apmoriarty (dismissed):
This will need to be updated against integration, otherwise looks good.

hlgp (dismissed):

warehouse/query-core/src/main/java/datawave/query/QueryParameters.java
```diff
@@ -137,6 +137,10 @@ public class QueryParameters {
      * Used to limit the number of values returned for specific fields
      */
     public static final String LIMIT_FIELDS = "limit.fields";
+    /**
+     * Used to tie field groups together such that if a field in one group is not being limited the those fields in matching groups will not be limited.
```
jwomeara:
```suggestion
     * Used to tie field groups together such that if a field in one group is not being limited the fields in matching groups will not be limited.
```

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
```diff
@@ -9,25 +9,70 @@
 import datawave.query.attributes.Content;
 import datawave.query.attributes.Document;
 import datawave.query.attributes.Numeric;
+import datawave.query.jexl.JexlASTHelper;
```
jwomeara: Unused.

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
```diff
@@ -9,25 +9,70 @@
...
 import datawave.util.StringUtils;
 import org.apache.accumulo.core.data.Key;
+import org.apache.commons.collections4.CollectionUtils;
```
jwomeara: Unused.

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
```diff
@@ -9,25 +9,70 @@
...
 import org.apache.log4j.Logger;
+import org.apache.lucene.util.CollectionUtil;
```
jwomeara: Unused.

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
```diff
@@ -9,25 +9,70 @@
...
 
+import java.util.ArrayList;
+import java.util.Collections;
```
jwomeara: Unused.

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
```diff
@@ -9,25 +9,70 @@
...
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
```
jwomeara: Unused.

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
```diff
@@ -9,25 +9,70 @@
...
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.stream.Collectors;
```
jwomeara: Unused.

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
```diff
@@ -9,25 +9,70 @@
...
 
+/**
+ * <p>
+ * LimitFields will reduce the attributes in a document given the limits specified for fields. Attributes that are in the set of hits for a document will never
+ * be dropped. Also matching field sets can be specified which will avoid dropping fields in the same group when the values match between the fields in a
+ * matching field set. for examples given the following field/values:
```
jwomeara:
```suggestion
 * matching field set. For example given the following field/values:
```

warehouse/query-core/src/main/java/datawave/query/function/MatchingFieldGroup.java
```diff
@@ -0,0 +1,26 @@
+package datawave.query.function;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class MatchingFieldGroup {
+    private final Set<String> matchingFieldSet;
+    private final Set<String> matchingValues;
```
jwomeara: Do we need to organize the values by field?  Is value matching across fields ok?  Can we add a test showcasing that if that was the intent?

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1293,9 +1297,17 @@ protected DocumentProjection getProjection() {
         DocumentProjection projection = new DocumentProjection(this.isIncludeGroupingContext(), this.isReducedResponse(), isTrackSizes());
         
         if (this.useWhiteListedFields) {
+            // make sure we including any fields being matched in the limit fields mechanism
```
jwomeara:
```suggestion
            // make sure we include any fields being matched in the limit fields mechanism
```

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1293,9 +1297,17 @@ protected DocumentProjection getProjection() {
...
+            if (!this.matchingFieldSets.isEmpty()) {
+                this.whiteListedFields.addAll(this.matchingFieldSets.stream().flatMap(s -> s.stream()).collect(Collectors.toList()));
+            }
             projection.setIncludes(this.whiteListedFields);
             return projection;
         } else if (this.useBlackListedFields) {
+            // make sure we including any fields being matched in the limit fields mechanism
```
jwomeara:
```suggestion
            // make sure we exclude any fields being matched in the limit fields mechanism
```

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1317,6 +1329,10 @@ protected DocumentProjection getCompositeProjection() {
                 }
             }
         }
+        // make sure we including any fields being matched in the limit fields mechanism
```
jwomeara:
```suggestion
        // make sure we include any fields being matched in the limit fields mechanism
```

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
jwomeara, line 1: were the changes to the allowBlankValue field intentional throughout?

warehouse/query-core/src/test/java/datawave/query/function/HitsAreAlwaysIncludedTest.java
```diff
@@ -260,8 +276,16 @@ public void testHitForIndexedQueryTerm() throws Exception {
...
-                        "FOO_1.FOO.1.3:good");
+        //@formatter:off
+        Set<String> goodResults = Sets.newHashSet(
+                "FOO_1_BAR.FOO.3:good<cat>", "FOO_3_BAR.FOO.3:defg<cat>", "FOO_3.FOO.3.3:defg", "FOO_4.FOO.4.3:yes", "FOO_1.FOO.1.3:good",
+                "FOO_1.FOO.1.0:yawn",
+                "FOO_1_BAR.FOO.0:yawn<cat>", "FOO_1_BAR.FOO.1:yawn<cat>",
+                "FOO_1_BAR_1.FOO.0:2021-03-24T16:00:00.000Z",
+                "FOO_3.FOO.3.0:abcd",
+                "FOO_3_BAR.FOO.0:abcd<cat>",
+                "FOO_4.FOO.4.0:purr", "FOO_4.FOO.4.1:purr");
```
jwomeara: Is there a rhyme or reason to some lines having multiple values, and others having single values?  If we're going to group by field (which I think was your intention) let's do that for all of the results, not just the new ones you added.

warehouse/query-core/src/test/java/datawave/query/function/HitsAreAlwaysIncludedTest.java
```diff
@@ -381,37 +458,107 @@ public void testHitWithExceededOrThreshold() throws Exception {
...
+        runTestQuery(queryString, format.parse("20091231"), format.parse("20150101"), extraParameters, expectedHits, goodResults);
+    }
+    
+    @Test
+    public void testGroupedHitsWithMatchingField() throws Exception {
+        Map<String,String> extraParameters = new HashMap<>();
+        extraParameters.put("include.grouping.context", "true");
+        extraParameters.put("hit.list", "true");
+        extraParameters.put("limit.fields", "FOO_1_BAR=-1,FOO_1=-1,FOO_3=-1,FOO_3_BAR=-1,FOO_4=-1,FOO_1_BAR_1=-1");
+        extraParameters.put("matching.field.sets", "FOO_4");
```
jwomeara, line 606: A test with multiple matching field sets would be good.

jwomeara (changes requested):
A few minor comments, and some additional tests requested.

warehouse/query-core/src/main/java/datawave/query/function/MatchingFieldGroup.java
ivakegg: The matching values contain values for one of the fields that is a hit in the query or of the same group as a hit in the query.  So this is simply keeping track of values that we would match to other fields in this fields set.  Perhaps I need to rename it?

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
ivakegg: Actuall the comment is correct, the code is incorrect here.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
ivakegg: yes it was per Moriarty's comment.

warehouse/query-core/src/test/java/datawave/query/function/HitsAreAlwaysIncludedTest.java
ivakegg: The first line is the set of hits, and the other lines are the addition fields included beyond the hits.  I will try to document as such.

warehouse/query-core/src/test/java/datawave/query/function/HitsAreAlwaysIncludedTest.java
ivakegg: Very true.  I will work that.  Thanks.

warehouse/query-core/src/main/java/datawave/query/function/MatchingFieldGroup.java
jwomeara: I see what you mean.  Looking at this again,  you aren't using the `matchingFieldSet` variable in this class.  The `containsField` method is unused as well.  Is there a good reason to keep the field set if we're not using it?

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -1448,6 +1460,16 @@ public boolean validateOptions(Map<String,String> options) {
             }
         }
         
+        if (options.containsKey(MATCHING_FIELD_SETS)) {
+            String matchingFieldSets = options.get(MATCHING_FIELD_SETS);
+            for (String fieldSet : Splitter.on(',').omitEmptyStrings().trimResults().split(matchingFieldSets)) {
+                String[] fields = Iterables.toArray(Splitter.on('=').omitEmptyStrings().trimResults().split(fieldSet), String.class);
+                if (fields != null) {
```
jwomeara: fields can't be null, but it could have length == 0.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
jwomeara: So you want to add the fields being matched to the blacklisted fields?

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
jwomeara: just checking

warehouse/query-core/src/test/java/datawave/query/function/HitsAreAlwaysIncludedTest.java
jwomeara: thank you!

jwomeara (changes requested):

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
jwomeara: What change needs to be made?  I was just commenting on the grammar.

jwomeara: You may have already updated the code, but I can't tell since you force pushed your changes.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
ivakegg: ah, right.... thanks

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1293,9 +1297,17 @@ protected DocumentProjection getProjection() {
         DocumentProjection projection = new DocumentProjection(this.isIncludeGroupingContext(), this.isReducedResponse(), isTrackSizes());
         
         if (this.useWhiteListedFields) {
+            // make sure we include any fields being matched in the limit fields mechanism
+            if (!this.matchingFieldSets.isEmpty()) {
+                this.whiteListedFields.addAll(this.matchingFieldSets.stream().flatMap(s -> s.stream()).collect(Collectors.toList()));
```
drewfarris: You do the following 3 times - consider encapsulating it into a protected method?
```
this.matchingFieldSets.stream().flatMap(s -> s.stream()).collect(Collectors.toList())
```

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -762,6 +762,17 @@ protected void loadQueryParameters(ShardQueryConfiguration config, Query setting
             }
         }
         
+        // Get the MATCHING_FIELD_SETS parameter if given
+        String matchingFieldSets = settings.findParameter(QueryParameters.MATCHING_FIELD_SETS).getParameterValue().trim();
+        if (org.apache.commons.lang.StringUtils.isNotBlank(matchingFieldSets)) {
```
drewfarris: Consider using `org.apache.commons.lang3` instead?

drewfarris (dismissed):
Minor suggestions - otherwise approved.

dtspence (dismissed):

warehouse/query-core/src/main/java/datawave/query/function/MatchingFieldGroup.java
jwomeara: This comment still stands I think.  Why do we need matchingFieldSet?

warehouse/query-core/src/test/java/datawave/query/function/HitsAreAlwaysIncludedTest.java
jwomeara: thanks

warehouse/query-core/src/main/java/datawave/query/function/MatchingFieldGroup.java
```diff
@@ -0,0 +1,26 @@
+package datawave.query.function;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class MatchingFieldGroup {
+    private final Set<String> matchingFieldSet;
```
jwomeara: This isn't used.

jwomeara (changes requested):
remove unused set in MatchingFieldGroup

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
ivakegg: done

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
ivakegg: requires change quite a few lines in this class.... I will try.

jwomeara (dismissed):

jwomeara (approved):

PR #1936, opened 2023-04-17
Catch JexlException separately in BaseQueryMetricHandler and Datawave…

billoley:
…SelectorExtractor to avoid JEXL exception when calling getMessage()


apmoriarty (dismissed):

jwomeara (dismissed):
why not just update jexl?  /s

you think it's worth it to add a test?

avgAGB (dismissed):

billoley:
> why not just update jexl? /s
> 
> you think it's worth it to add a test?

Updating JEXL is much more involved than avoiding an NPE here
From what I saw, the cases where a JexlException is thrown involve large queries with some JEXL error that doesn't cause a different type of exception.  I will see what I can find.

jwomeara:
> > why not just update jexl? /s
> > you think it's worth it to add a test?
> 
> Updating JEXL is much more involved than avoiding an NPE here From what I saw, the cases where a JexlException is thrown involve large queries with some JEXL error that doesn't cause a different type of exception. I will see what I can find.

The '/s' was for 'sarcasm'...

hlgp (dismissed):

PR #1938, opened 2023-04-19
Task/document query parsing

jack-gitdev:
Add tests for verifying expected order of operations in absence of parenthesis. Add convenience methods for interacting with JexlNode objects.


warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -1237,6 +1256,44 @@ public static boolean isWithinAnd(JexlNode node) {
         return false;
     }
     
+    public static <T extends JexlNode> boolean isWithinNodeType(JexlNode node, Class<T> typeKey) {
+        while (null != node && null != node.jjtGetParent()) {
```
jwomeara: This returns after a single iteration, so there's no need for a while loop.  You didn't create them, but this is also true for isWithinAnd, and isWithinNot.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
jwomeara, line 1: Javadoc to describe what your methods are doing would be appreciated.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -1237,6 +1256,44 @@ public static boolean isWithinAnd(JexlNode node) {
         return false;
     }
     
+    public static <T extends JexlNode> boolean isWithinNodeType(JexlNode node, Class<T> typeKey) {
```
jwomeara: I know there are already a few methods named 'isWithin...' but to me all of these methods would be better named as 'isDescendantOf...'

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -1237,6 +1256,44 @@ public static boolean isWithinAnd(JexlNode node) {
...
+                
+                if (typeKey.isInstance(child)) {
+                    return true;
+                } else {
+                    foundType = childrenContainNodeType(child, typeKey);
+                }
+                
+                if (foundType) {
+                    return true;
+                }
```
jwomeara: could be simplified

```suggestion
                if (typeKey.isInstance(child) || childrenContainNodeType(child, typeKey)) {
                    return true;
                }
```

jwomeara (changes requested):
A few comments...

So we're adding these new methods, but they aren't being used anywhere outside of tests.  Is this something we foresee us needing as part of a future task?  Why do we need these methods?

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
jack-gitdev: Sounds good, I will change this is.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
jack-gitdev: Agreed, will change them.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
jack-gitdev: Thank you for this, I could not figure out how to simplify it but I know it didn't look great.

jack-gitdev:
I will make the recommended changes. I added the convenience methods for my own use in the new tests. I didn't see any methods for checking against certain node types so I added these  generalized methods that can be used to check against any of the JexlNode based classes.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
jack-gitdev: Will add Javadocs.

jwomeara (approved):
I think you covered all of my concerns.  LGTM

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -796,6 +820,43 @@ private static void getERNodes(JexlNode node, List<ASTERNode> erNodes) {
         }
     }
     
+    /**
+     * Iterate through provided node and its children, then return a list of nodes matching the provided class. The provided class must extend JexlNode.
+     * 
+     * @see org.apache.commons.jexl2.parser.JexlNode;
```
jwomeara: here

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -796,6 +820,43 @@ private static void getERNodes(JexlNode node, List<ASTERNode> erNodes) {
...
+        
+        getNodesOfType(node, nodes, typeKey);
+        
+        return nodes;
+    }
+    
+    /**
+     * Check if the provided node is an instance of the provided class. If yes, then add the node to the provided list. The provided class must extend JexlNode.
+     * 
+     * @see org.apache.commons.jexl2.parser.JexlNode;
```
jwomeara: here

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -1195,43 +1256,116 @@ public static LiteralRange<BigDecimal> getBigDecimalBoundedRange(JexlNode[] chil
...
+            return isDescendantOfAnd(parent);
+        }
+        
+        return false;
+    }
+    
+    /**
+     * Iterate through provided node's ancestors and return true if any are an instance of the provided class. The provided class must extend JexlNode.
+     * 
+     * @see org.apache.commons.jexl2.parser.JexlNode;
```
jwomeara: here

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -1195,43 +1256,116 @@ public static LiteralRange<BigDecimal> getBigDecimalBoundedRange(JexlNode[] chil
...
+        }
+        
+        return false;
+    }
+    
+    /**
+     * Iterate through all descendants of the provided node and return true if any are an instance of the provided class. The provided class must extend
+     * JexlNode.
+     * 
+     * @see org.apache.commons.jexl2.parser.JexlNode;
```
jwomeara: here

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
jwomeara, line 1: your unit tests are failing due to some of your javadoc.  apparently javadoc doesn't like a `@see` tag that ends with a semicolon.  I pointed out the offenders below.

jwomeara (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
jack-gitdev: Thanks. I removed them and it should build successfully now.

jwomeara (approved):

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -1195,43 +1256,116 @@ public static LiteralRange<BigDecimal> getBigDecimalBoundedRange(JexlNode[] chil
...
+     * Iterate through provided node's ancestors and return true if any are an instance of the provided class. The provided class must extend JexlNode.
+     * 
+     * @see org.apache.commons.jexl2.parser.JexlNode
+     *
+     * @param node
+     * @param typeKey
+     * @return True if any ancestor is an instance of the provided class. If not, then False.
+     */
+    public static <T extends JexlNode> boolean isDescendantOfNodeType(JexlNode node, Class<T> typeKey) {
+        if (null != node && null != node.jjtGetParent()) {
```
apmoriarty, line 1347: Instead of the recursive calls here, you could have a more memory efficient setup like this. I don't think you should make this change in this merge request, but maybe that's work doing as a separate unit of work. 
```
while (node.getParent() != null){
  node = node.getParent()
  if (node.isInstance(type)) {
    return true
  }
}
```

apmoriarty (approved):

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
apmoriarty: Created https://github.com/NationalSecurityAgency/datawave/issues/1946 to address this

PR #1939, opened 2023-04-20
Aggregator that supports no uids on the global index

apmoriarty:


jwomeara (commented):
Why do we need this?  

I mean that in two senses - 1) do we have a plan to use this, or are we adding it just to add it? 2) this is identical to KeepCountOnlyUidAggregator, but with a hard-coded 0 in the constructor.  

Why wouldn't we just use that?

apmoriarty:
> Why do we need this?
> 
> I mean that in two senses - 1) do we have a plan to use this, or are we adding it just to add it? 2) this is identical to KeepCountOnlyUidAggregator, but with a hard-coded 0 in the constructor.
> 
> Why wouldn't we just use that?
1) this is going to be used
2) configuration in conjunction with the propogating iterator is a nightmare. Hard coding the zero makes integration extremely easy.

jwomeara:
> 

*Where* is this going to be used?  Why have this as a separate PR instead of having this exist within the context of that work, and that PR?

I don't agree that setting this up would be a 'nightmare'  This iterator has an init method just like any other accumulo iterator. 
 The init method has access to whatever options have been set on the table, such as the max number of uids to keep...  Doesn't seem like a nightmare to me at all.

Even still, this seems like it's just a specialized version of KeepCountOnlyUidAggregator, so why not extend that instead and then all you need is a constructor passing 0 to the super...

jwomeara (changes requested):
See latest comments.

jwomeara (approved):

ivakegg (approved):

PR #1940, opened 2023-04-21
Add document key function

apmoriarty:
This is a redo of https://github.com/NationalSecurityAgency/datawave/pull/1854 which had to be reverted.

The original handling of negations was correct, but got refactored out based on review comments. I have confirmed that negated terms are not added to the document and thus negated values must be generated for all document keys.

This does not preclude negated terms existing outside of the document keys generated by the positive terms, but at the very least this work with not introduce any regressions.


warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/DocumentKeysFunction.java
```diff
@@ -0,0 +1,223 @@
...
+                contentFunctions.put(contentFunction.getField(), contentFunction);
+
+                if(isFunctionNegated(f)){
+                    negatedValues.addAll(contentFunction.getValues());
+                }
+            }
+        }
+    }
+
+    public boolean isFunctionNegated(Function f){
```
FineAndDandy, line 83: At this point won't any nested negations already be flattened?

FineAndDandy: https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownNegationVisitor.java

I think as long as this is still being used we will always prevent having more than 1

FineAndDandy (approved):

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/DocumentKeysFunction.java
apmoriarty: You're correct that this function could stop at the first ASTNotNode it finds. This function was written to be thorough in case the assumption about pushing down negations is ever invalidated.

jack-gitdev (approved):
Looks good to me!

PR #1941, opened 2023-04-26
Allowing configurable shards per table for tomorrow shards script

avgAGB:


jwomeara (dismissed):
Would it ever be the case that we want to set a default value to apply to all three tables?  If so, should we have that as an option in addition to being able to set them all separately?  If not, feel free to merge.

avgAGB:
> Would it ever be the case that we want to set a default value to apply to all three tables? If so, should we have that as an option in addition to being able to set them all separately? If not, feel free to merge.

I think in that case it may be preferable to update the base properties in that case so that the value could be tracked a bit easier, and the script could draw from that. Dylan raised a good point yesterday about checking to see if this script would gel with our shard cache file in HDFS, so I'll check to see that there isn't a potential clash there.

avgAGB:
> > Would it ever be the case that we want to set a default value to apply to all three tables? If so, should we have that as an option in addition to being able to set them all separately? If not, feel free to merge.
> 
> I think in that case it may be preferable to update the base properties in that case so that the value could be tracked a bit easier, and the script could draw from that. Dylan raised a good point yesterday about checking to see if this script would gel with our shard cache file in HDFS, so I'll check to see that there isn't a potential clash there.

Followed up on this and it looks like in our other repos the shard cache's last value is accounted for, so I'll update this PR accordingly to set the shards for the error/query metric tables.

alerman (dismissed):

matthpeterson (dismissed):

warehouse/ingest-scripts/src/main/resources/bin/ingest/create-tomorrows-shards.sh
```diff
@@ -22,23 +23,37 @@ if [[ "$2" == "" ]]; then
 else
   declare -i divisor=$2
 fi
-numshards=$((numshards * numerator / divisor))
 
-shardsPerSplit=1
-if [[ -n "$3" ]]; then
-    shardsPerSplit=$3
+if [[ "$3" == "" ]]; then
```
ivakegg, line 38: Need some comments at the top of this script as to what the arguments are.

ivakegg (changes requested):
need script comments.

warehouse/ingest-scripts/src/main/resources/bin/ingest/create-tomorrows-shards.sh
avgAGB: Can do, I'll add them to the next commit.

apmoriarty (dismissed):

matthpeterson (dismissed):

warehouse/ingest-scripts/src/main/resources/bin/ingest/create-tomorrows-shards.sh
```diff
@@ -22,23 +34,37 @@ if [[ "$2" == "" ]]; then
...
+daysToGenerate=1
+
+CLASSPATH=$ADDJARS $WAREHOUSE_ACCUMULO_HOME/bin/accumulo datawave.ingest.util.GenerateShardSplits $DATE ${daysToGenerate} ${shardTableNumShards} ${SHARD_TABLE_SHARDNUM_SPLIT_STEP} -addShardMarkers -addDataTypeMarkers $TYPES $USERNAME $PASSWORD ${SHARD_TABLE_NAME} $WAREHOUSE_INSTANCE_NAME $WAREHOUSE_ZOOKEEPERS
 
-CLASSPATH=$ADDJARS $WAREHOUSE_ACCUMULO_HOME/bin/accumulo datawave.ingest.util.GenerateShardSplits $DATE 1 ${numshards} ${shardsPerSplit} -addShardMarkers -addDataTypeMarkers $TYPES $USERNAME $PASSWORD ${ERROR_SHARD_TABLE_NAME} $WAREHOUSE_INSTANCE_NAME $WAREHOUSE_ZOOKEEPERS
+CLASSPATH=$ADDJARS $WAREHOUSE_ACCUMULO_HOME/bin/accumulo datawave.ingest.util.GenerateShardSplits $DATE ${daysToGenerate} ${errorTableNumShards} ${ERROR_SHARD_TABLE_SHARDNUM_SPLIT_STEP} -addShardMarkers -addDataTypeMarkers $TYPES $USERNAME $PASSWORD ${ERROR_SHARD_TABLE_NAME} $WAREHOUSE_INSTANCE_NAME $WAREHOUSE_ZOOKEEPERS
```
ivakegg, line 68: Did you update the ingest_env.sh with the new properties?

avgAGB: In the latest commit, slipped through the cracks there. thanks

matthpeterson (approved):

alerman (approved):

apmoriarty (approved):

ivakegg (approved):

PR #1943, opened 2023-05-01
Log Long Aggregations

apmoriarty:
- cleaned up KeyToDocumentData
- added thresholds for event and term offset aggregation

### DIFF
--- a/warehouse/query-core/src/test/resources/datawave/query/QueryLogicFactory.xml
+++ b/warehouse/query-core/src/test/resources/datawave/query/QueryLogicFactory.xml
@@ -238,6 +238,9 @@
         <property name="eventNextSeek" value="-1" />
         <property name="tfFieldSeek" value="-1" />
         <property name="tfNextSeek" value="-1" />
+        <!--    control when to log warnings for long aggregations    -->
+        <property name="docAggregationThresholdMs" value="-1" />
+        <property name="tfAggregationThresholdMs" value="-1" />
     </bean>
 
     <bean id="TLDEventQuery" scope="prototype" parent="BaseEventQuery" class="datawave.query.tables.TLDQueryLogic">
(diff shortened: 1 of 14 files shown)


magrable (dismissed):

warehouse/query-core/src/test/resources/datawave/query/QueryLogicFactory.xml
```diff
@@ -238,6 +238,9 @@
         <property name="eventNextSeek" value="-1" />
         <property name="tfFieldSeek" value="-1" />
         <property name="tfNextSeek" value="-1" />
+        <!--    control when to log warnings for long aggregations    -->
+        <property name="docAggregationThresholdMs" value="-1" />
+        <property name="tfAggregationThresholdMs" value="-1" />
```
jwomeara, line 243: Do we need to set this since we're just using the default value here?

web-services/deploy/configuration/src/main/resources/datawave/query/QueryLogicFactory.xml
```diff
@@ -306,6 +306,9 @@
         <property name="selectorExtractor" ref="DatawaveSelectorExtractor" />
         <property name="evaluationOnlyFields" value="${evaluation.only.fields}" />
         <property name="lazySetMechanismEnabled" value="false" />
+        <!--    control when to log warnings for long aggregations    -->
+        <property name="docAggregationThresholdMs" value="-1" />
+        <property name="tfAggregationThresholdMs" value="-1" />
```
jwomeara, line 311: Do we need to set this since we're just using the default value here?

jwomeara (approved):
looks good.  just a minor comment.

warehouse/query-core/src/test/resources/datawave/query/QueryLogicFactory.xml
apmoriarty: It's not strictly necessary, the example here is for illustration

magrable (approved):

PR #1944, opened 2023-05-02
Remove batched query option

apmoriarty:


warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -373,7 +366,6 @@ public class QueryOptions implements OptionDescriber {
     protected boolean compressResults = false;
     
     protected Boolean compressedMappings = false;
-    protected boolean limitOverride = false;
```
jwomeara, line 382: It appears that this variable isn't used anywhere and that's why you removed it.  Just want to make sure I understand this correctly.

jwomeara (dismissed):
Effort to remove batched query looks good.  Can I ask why it's being removed though?  Are we just not using it?  Is it buggy, or non-performant?  Just curious, otherwise looks good to go.

avgAGB (dismissed):

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
apmoriarty: Correct, this is just removing an unused variable.

apmoriarty:
> Effort to remove batched query looks good. Can I ask why it's being removed though? Are we just not using it? Is it buggy, or non-performant? Just curious, otherwise looks good to go.

We evaluated it several years ago and found it to be buggy. It's been disabled via configs as far as I can remember. At one point we considered investing the time to make it work correctly, but decided not to.

avgAGB (approved):

jwomeara (approved):

PR #1951, opened 2023-05-05
Update query limits handling to allow rules based on systemFrom values

drewfarris:
Currently, Datawave provides a mechanism to override default query result limits based on the DN of the requester. 

This patch adds the ability to override default query result limits based on the value specified in the `systemFrom` request parameter. In cases where both a `systemFrom` and DN-based result limit override is in place, the DN-based limit takes precedence.

HTTP parameters are not an ideal mechanism for specifying result override behavior because any caller may put an arbitrary value in the `systemFrom`, which is used currently as a label for metrics. As such, using `systemFrom` limits is _weak_ in the sense that there is no validation that a particular caller is whom they claim. 

As such, DN-based limits remain the preferred mechanism for overriding query limits and `systemFrom` is provided as an alternative for cases where multiple systems' calls originate from a connection associated with a single DN.

### DIFF
--- a/web-services/query/src/main/java/datawave/webservice/query/logic/BaseQueryLogic.java
+++ b/web-services/query/src/main/java/datawave/webservice/query/logic/BaseQueryLogic.java
@@ -30,6 +30,7 @@ public abstract class BaseQueryLogic<T> implements QueryLogic<T> {
     private String logicDescription = "Not configured";
     private AuditType auditType = null;
     private Map<String,Long> dnResultLimits = null;
+    private Map<String,Long> systemFromResultLimits = null;
     protected long maxResults = -1L;
     protected ScannerBase scanner;
     @SuppressWarnings("unchecked")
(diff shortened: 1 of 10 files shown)


apmoriarty (approved):

drewfarris:
See also https://github.com/NationalSecurityAgency/datawave/compare/integration...drewfarris:datawave:systemfrom-query-limits-3.41-backport for a backport of this change to release/version3.41

magrable:
Is there a use case for wanting to overwrite a DN limit that may or may not exist with a lower systemFrom limit?  Going with the lower limit seems to mitigate the 'weakness' of using systemFrom because we are choosing the more restrictive option.  If it is not a use case, LGTM.

drewfarris:
> Is there a use case for wanting to overwrite a DN limit that may or may not exist with a lower systemFrom limit? Going with the lower limit seems to mitigate the 'weakness' of using systemFrom because we are choosing the more restrictive option.

In practice, we use this to make exceptions to existing query limits. For example, a query logic may be configured to return a maximum of 100 results, but certain DNs can ask for up to 1000 results. The specific case we're addressing here is when multiple 'systems' share the same DN and we don't want to raise the limit for all systems.

In this case, we will not add a DN rule at all and use the default limit for most systems running queries behind this DN. For specific systems where we need to allow more results, we will add systemFrom rules as exceptions.

We're not always looking for the 'smallest' limit so we don't want to get into the business of comparing the DN rules vs. the systemFrom rules and choosing the smallest of the two. There are use cases where it is more appropriate to return the largest number possible.

magrable (approved):

PR #1958, opened 2023-05-08
WIP: issue-1957: MaskedValueFilterFactory.get needs to be thread safe.

friedlou:


warehouse/query-core/src/main/java/datawave/query/function/MaskedValueFilterFactory.java
jwomeara, line 1: Why can't we just remove this call to `initializeCache()` ?  There is already a static initializer which should call this method the first time the class is accessed.

jwomeara (changes requested):

warehouse/query-core/src/main/java/datawave/query/function/MaskedValueFilterFactory.java
friedlou: Not sure what you mean. ` initializeCache()` is only called in `get()` and is the static initializer.  `get()` is the only public static method in the class.

friedlou: You are correct. Closing this PR.

PR #1969, opened 2023-05-19
optionally disable creating uids for the global index

apmoriarty:
Change Summary
- optionally disable creating uids for the shard index and shard reverse index by setting the `SHARD_INDEX_CREATE_UIDS` property to `false`
- consolidated Uid.List creation logic in `ShardedDataTypeHandler` and `ExtendedContentIndexingColumnBasedHandler`

To verify this changeset locally using the datawave quickstart, do the following
1. Checkout this branch, set `SHARD_INDEX_CREATE_UIDS` to false, and run `datawaveBuildDeploy`
2. Verify configs took effect by running 
`grep -A 1 "create.uids" /path/to/git/datawave/contrib/datawave-quickstart/datawave-ingest/config/*`
3. run `allStop`, `allUninstall` and `allInstall`
4. ashell and `scan -t datawave.shardIndex`, and `scan -t datawave.shardReverseIndex`


warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
```diff
@@ -1141,21 +1150,40 @@ protected void createShardFieldIndexColumn(RawRecordContainer event, Multimap<Bu
...
+     */
+    protected Uid.List createUidList(String uid, boolean deleteMode) {
+        Uid.List.Builder uidBuilder = Uid.List.newBuilder();
+
+        //  delete mode takes precedent over disabled uid creation
+        if (deleteMode) {
+            uidBuilder.setIGNORE(false);
             uidBuilder.setCOUNT(-1);
-        else
+            uidBuilder.addUID(uid);
```
jschmidt10, line 1171: This should be `addREMOVEDUID()`

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
```diff
@@ -1141,21 +1150,40 @@ protected void createShardFieldIndexColumn(RawRecordContainer event, Multimap<Bu
...
+        //  if uid creation is disabled
+        if (!getShardIndexCreateUids()) {
+            uidBuilder.setIGNORE(true);
             uidBuilder.setCOUNT(1);
-        
+            return uidBuilder.build();
+        }
+
+        //  else create a uid
+        uidBuilder.setIGNORE(false);
```
jschmidt10: This if/early return structure is basically mimicking an if/else if/else. I think the latter is a bit easier to follow:

```
if (isDelete)
  // set delete stuff
else if (isCreateUids)
  // set uid list stuff
else
  // set count only
return uidBuilder.build();
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
jschmidt10: Wow. Looks like it was that way before. Have we ever used this code path? I wonder if this was left over from a previous implementation, pre-removal list.

warehouse/ingest-configuration/src/main/resources/config/shard-ingest-config.xml
```diff
@@ -19,6 +19,11 @@
   <value>${table.name.shard}</value>
 </property>
 
+<property>
+    <name>shard.index.create.uids</name>
```
jschmidt10, line 23: It feels like this property just belongs in `shard-ingest-config.xml`. Since we load *config.xml in our ingest scripts, this has the capacity to clobber the property. Like if you set it to false here but forgot to in the other files, it's not clear which one would win.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
apmoriarty: Yea....I'm trying to do least harm here. I don't want really want to change this to `addREMOVEDUID` unless there's a really good reason to.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
apmoriarty: I can certainly do that. Thoughts on order of precedent? I figured delete should take precedent over 'no uids', but I'm open to suggestions.

warehouse/ingest-configuration/src/main/resources/config/shard-ingest-config.xml
apmoriarty: I'll run a few experiments and let you know the results. Ideally we could set this once at the top level config, but it might e more helpful to be verbose about this property.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
jschmidt10: I don't think the `addUID` with a count of -1 works anymore as a "delete". I believe that logic has been removed from the GlobalIndexUidAggregator. I'm assuming we don't use this code path b/c if we did, I don't believe this would work.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
jschmidt10: Thanks! The order you used makes sense to me.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
jschmidt10: To your point, you can leave what you have and we could open a separate issue to investigate that.

jschmidt10 (dismissed):

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
```diff
@@ -1141,21 +1150,38 @@ protected void createShardFieldIndexColumn(RawRecordContainer event, Multimap<Bu
...
+        } else if (!getShardIndexCreateUids()) {
+            //  uid not created
+            uidBuilder.setIGNORE(true);
             uidBuilder.setCOUNT(1);
-        
-        uidBuilder.addUID(uid);
-        
-        Uid.List uidList = uidBuilder.build();
-        return new Value(uidList.toByteArray());
+            return uidBuilder.build();
```
jschmidt10: Oh missed this before I hit approve. This return can be removed.

apmoriarty: Missed that, fixed in next commit

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
```diff
@@ -1141,21 +1150,37 @@ protected void createShardFieldIndexColumn(RawRecordContainer event, Multimap<Bu
...
+    protected Uid.List createUidList(String uid, boolean deleteMode) {
+        Uid.List.Builder uidBuilder = Uid.List.newBuilder();
+
+        //  delete mode takes precedent over disabled uid creation
+        if (deleteMode) {
+            uidBuilder.setIGNORE(false);
             uidBuilder.setCOUNT(-1);
-        else
+            uidBuilder.addUID(uid);
+        } else if (!getShardIndexCreateUids()) {
```
jwomeara: why use the getter vs using the variable directly?

apmoriarty: that's leftover from when this method was in an extending class...can update

jwomeara (approved):

jschmidt10 (approved):

PR #1971, opened 2023-05-22
Add Field Index Aggregators

apmoriarty:
Change summary
- add FiAggregator and TLDFiAggregator, alternate implementations of the FieldIndexAggregator
- aggregators are not integrated into production code

Motivation: document building is a binary decision based on one of two things, a query that is satisfied from the field index or a field that is index only. This change set allows index only terms to not perform aggregation that could be very expensive in terms of time and memory.


warehouse/query-core/src/main/java/datawave/query/jexl/functions/FiAggregator.java
```diff
@@ -0,0 +1,202 @@
...
+
+    protected final FieldIndexKey parser = new FieldIndexKey();
+    protected final FieldIndexKey tkParser = new FieldIndexKey();
+
+
+    public FiAggregator() {
+        //  empty constructor
+    }
+
+    //  builder style methods for legacy values
```
jwomeara: If these are 'builder style' shouldn't they return 'this' so calls to these methods can be chained?

warehouse/query-core/src/main/java/datawave/query/jexl/functions/TLDFiAggregator.java
```diff
@@ -0,0 +1,96 @@
...
+
+    public TLDFiAggregator(){
+        //  empty constructor
+    }
+
+    //  apply(SortedKeyValueIterator<Key,Value> itr)
+    //  this path is handled by overriding sameUid()
+
+    //  apply(SortedKeyValueIterator<Key,Value> iter, Range range, Collection<ByteSequence> cfs, boolean includeColumnFamilies)
+    //  this path is handled by overriding getSeekRange()
```
jwomeara: delete?

jwomeara (changes requested):
What is the motivation for adding this?  Is there a corresponding ticket explaining why we should add these or how they would be used?

warehouse/query-core/src/main/java/datawave/query/jexl/functions/FiAggregator.java
apmoriarty: Updated in https://github.com/NationalSecurityAgency/datawave/pull/1971/commits/b596c4a93cecbe08befbda16e24c4af06dd74000

warehouse/query-core/src/main/java/datawave/query/jexl/functions/TLDFiAggregator.java
apmoriarty: Deleted in next commit, comments added to the `sameUid()` and `getSeekRange()` methods

cogross (approved):

PR #1973, opened 2023-05-22
Added four tests

MiguelRicardos:


warehouse/query-core/src/test/java/datawave/query/util/sortedset/MultiSetBackedSortedSetTest.java
```diff
@@ -0,0 +1,585 @@
...
+    @Test
+    public void testHeadSet() {
+        int end = sortedOrder.length / 3;
+        int start = 0;
+        SortedSet<byte[]> subSet = set.headSet(data[sortedOrder[end]]);
+
+        // verify contents
+        assertEquals(end - start, subSet.size());
+        for (int i = 0; i < data.length; i++) {
+            if (i >= start && i < end) {
```
jwomeara: start is 0, and i starts at 0.  how could i ever be >= start?

warehouse/query-core/src/test/java/datawave/query/util/sortedset/MultiSetBackedSortedSetTest.java
```diff
@@ -0,0 +1,585 @@
...
+            } else {
+                assertFalse(subHeadSet.contains(data[sortedOrder[i]]));
+            }
+        }
+    }
+    /**
+     * Test method for {@link datwave.query.util.sortedset.MultiSetBackedSortedSet#headSet(byte[])}.
+     */
+    @Test
+    public void testHeadSet() {
```
jwomeara, line 345: This test is currently failing.  Not sure if the test is invalid, or the setup is invalid or what...

jwomeara (changes requested):

warehouse/query-core/src/test/java/datawave/query/util/sortedset/MultiSetBackedSortedSetTest.java
MiguelRicardos: I've been futzing with it. I ended up commenting out the two assert statements that have issues. Funny enough the test does not work on the other network unless these two are commented out.

warehouse/query-core/src/test/java/datawave/query/util/sortedset/MultiSetBackedSortedSetTest.java
MiguelRicardos: I copied it as is. I can change them if you want.

warehouse/query-core/src/test/java/datawave/query/util/sortedset/MultiSetBackedSortedSetTest.java
MiguelRicardos: Resolving now that it is passing.

avgAGB (dismissed):

apmoriarty (dismissed):

warehouse/query-core/src/test/java/datawave/query/util/sortedset/MultiSetBackedSortedSetTest.java
jwomeara: So am I to understand that there was a time when those tests ran successfully with those assertions, but they don't today?  Shouldn't we figure out why those assertions don't work now instead of commenting them out?

jwomeara: If those assertions truly are invalid, then they should be removed, not commented out.  But we should make sure that they're actually invalid before we remove them.

MiguelRicardos: From my digging through this, it looks like the original  ```assertTrue(Arrays.equals(data[sortedOrder[start]], subSet.first()));``` has an issue where the sebSet.first() always returns a no such element exception. I replaced this line with ```assertArrayEquals(data[sortedOrder[start]], subSet.iterator().next());``` to verify that the contents of the first element of the subSet match data[sortedOrder[start]]. Along with a ```assertFalse(subSet.isEmpty());``` on the preceding line. On the ticket, I added a link to the comparison and pipeline on the other network showing that when moving this test from the original repo to the intended repo, they did not pass in their original configuration. This points me to something wrong with the underlying MultiSetBackedSortedSet first() and last() methods. Creating a follow on ticket to investigate what is wrong with the underlying first() and last() methods.

MiguelRicardos: @jwomeara Are there any additional requested changes?

MiguelRicardos:
@avgAGB  Could you review this again?

magrable (dismissed):

jack-gitdev (dismissed):

jwomeara (changes requested):
Thanks for updating the first `assertTrue` statement in MultiSetBackedSortedSetTest.  Can you give a similar treatment to the other `assertTrue` statement that you removed?

```
        //assertTrue(Arrays.equals(data[sortedOrder[end - 1]], subSet.last()));
```

jwomeara (approved):

apmoriarty (approved):

Issue #1974, opened 2023-05-22
Create RAW JSON Document Serialization


phrocker:
Create a way for Docs to be serialized into JSON and not incur the cost of deser/ser at the webserver. Consumers will receive documents that are "raw" The result will be a faster response time, but they may lose some fidelity of the objects returned in JSON form. 

Based on prior work there will be several follow-on PRs.

phrocker:
I will submit 4 PRs:

*  everything together
*  Webserver only changesets
* shardquery logic breakdown into a base query logic
* query logic changes

PR #1975, opened 2023-05-22
Create RAW JSON Response; UberPR

phrocker:
Create a RAW JSON response per #1974 . This PR is an "uber PR" containing the JSON functionality; however, I would want to poll the community with my intent to create three additional PRs:

* One for the ShardQueryLogic Changeset that creates a base query logic
* One for the Webserver side changes
* One for the Document logic change set. 

The goal for these PRs is to reduce this to a reviewable and manageable set of PRs. The purpose of this PR is to give context to the others that will be submitted. There are a few files I may decide to trim from this PR, but in general it will be quite large since it encompasses a lot of tests. Three smaller PRs will be easier to review and lessen chance of artifacts sneaking by. I'm open to feedback on whether the three smaller PRs is just added work and the bigger one would be fine.

Naturally some of these changes are the result of copied code but the thought is that DocumentBatchScanner would replace BatchScanner if found valuable; however, I'd love the community's feedback about this as well. Since this encompasses a lot of changes I'm happy to hear preferences on how to submit. A tad over 35% of the change set are test files with some of these files being on the longer side; however, the bulk of the changes will be in the third category, identified above. 

Considerations:

**Why is this beneficial?** 

This is beneficial because each tablet server would generate a final document for the webserver to return. There is a transformer but ideally we would do as little work as possible at that point. This would be an experimental feature, as we need significant testing and some assumptions are made. For example, Row and CF are removed from metadata keys in returned documents to save space. 

Using perfjobs the RawDocumentQuery has been closer to that of other query testing frameworks and nearly double the speed of the traditional query logic when all fields are enabled. With additional changes that were excluded from this PR we should see 3-4x the performance of the ShardQueryLogic. It is expected that the return JSON is not identical to the JSON Document POJO converted into an Event XML or JSON return. 

**Who would this impact?**

No one who is NOT using this feature. This *should* have no impact to the ShardQeueryLogic customers in the best case scenario. 

**How does this benefit users?**

More lightweight responses means we could, in theory, distribute how queries are performed and view microservices work a bit differently. The many scans being performed at the tablet server would do the translation to JSON that would be returned to the user. 

**I see a custom batch scanner. what gives? Shouldn't that be in Accumulo?**

This probably should go into Accumulo. The idea here is that we reduce the number of copies of the binary data. In the case of a Raw JSON document we don't even need to convert the object into a Value and copy the bytes. We could use the TKeyValue directly up until the point it is returned to the end user. This is a BIG assumption and one we should validate. 

If the DW community does not believe this scanner should go into DW then the next logical place would be Accumulo, but I don't know if the broader Accumulo community has a need for more strongly typed objects coming from the Scanner since the existing interfaces are clearly Key/Value tuples. 

**If we minimize the fields to be returned DW and This change set are closer than 2-3x**

Yes, the fewer fields requested the less this change set benefits. With subsequent changes ( that I've broken out into separate PRs ) the Document Logic should still be "faster"; however, the change set comes with functional changes to calls ( with a new endpoint ) and a new type output. So I don't believe that speed should be the only deciding factor and the changes are positioned such that we should allow users to decide.


phrocker:
I hit the button to merge 5.1 and update my branch. I'm going to delete that commit then rebase from that branch

warehouse/query-core/src/main/java/datawave/query/index/lookup/ScannerStream.java
```diff
@@ -15,7 +15,6 @@
  * Note that certain delayed terms may create a ScannerStream without an underlying RangeStreamScanner.
  */
 public class ScannerStream extends BaseIndexStream {
```
phrocker: This was an artifact of doing dev on Windows and Linux. Will remove this change.

warehouse/query-core/src/main/java/datawave/query/tables/BatchScannerImpl.java
```diff
@@ -0,0 +1,106 @@
+package datawave.query.tables;
+
+import org.apache.accumulo.core.client.BatchScanner;
```
phrocker: This is an unused file. Removing

warehouse/query-core/src/main/java/datawave/query/tables/BatchScan.java
```diff
@@ -0,0 +1,710 @@
+package datawave.query.tables;
+
+
+import org.apache.accumulo.core.client.*;
+import org.apache.accumulo.core.clientImpl.*;
```
phrocker: This is an unused file. Removing

web-services/query/src/main/java/datawave/webservice/query/logic/AbstractQueryLogicTransformer.java
```diff
@@ -10,7 +11,7 @@
     public static final String PARTIAL_RESULTS = "Partial/incomplete page of results returned probably due to memory constraints";
     
     public abstract BaseQueryResponse createResponse(List<Object> resultList);
-
```
phrocker: This was more of moving between windows and linux where the EOL character changed. Will remove these as well.

web-services/query/src/main/java/datawave/webservice/query/logic/deser/ResultsPageAdapter.java
```diff
@@ -0,0 +1,41 @@
+package datawave.webservice.query.logic.deser;
+
+import com.google.common.base.Joiner;
+import com.google.gson.TypeAdapter;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonWriter;
+
+import java.io.IOException;
+import java.util.Objects;
+
```
phrocker, line 10: TODO: add a comment explaining structure

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -2064,6 +2066,229 @@ private BaseQueryResponse next(final String id, boolean checkForContentLookup) {
         
         return response;
     }
+
+
+    @GET
+    @Path("/{id}/raw/next")
```
phrocker, line 2070: I waffled between raw or direct, so I'm open to feedback on how to relay an endpoint that suggests we are not transforming or converting the JSON for the end user.

warehouse/query-core/src/test/java/datawave/query/iterator/IteratorBuildingVisitorIT.java
```diff
@@ -0,0 +1,1502 @@
...
+import java.util.UUID;
+
+import static datawave.query.iterator.QueryOptions.*;
+import static org.apache.commons.pool.impl.GenericObjectPool.WHEN_EXHAUSTED_BLOCK;
+
+/**
+ * Integration tests for the IteratorBuildingVisitor
+ *
+ */
+public class IteratorBuildingVisitorIT extends EasyMockSupport implements SourceFactory<Key,Value> {
```
phrocker: This will also go into a separate PR as stated in the description.

warehouse/query-core/src/test/java/datawave/query/jexl/functions/ContentFunctionQueryTest.java
```diff
@@ -77,24 +77,11 @@
...
 import static datawave.query.QueryParameters.DATE_RANGE_TYPE;
-import static datawave.webservice.query.QueryParameters.QUERY_AUTHORIZATIONS;
-import static datawave.webservice.query.QueryParameters.QUERY_BEGIN;
-import static datawave.webservice.query.QueryParameters.QUERY_END;
-import static datawave.webservice.query.QueryParameters.QUERY_EXPIRATION;
-import static datawave.webservice.query.QueryParameters.QUERY_LOGIC_NAME;
-import static datawave.webservice.query.QueryParameters.QUERY_NAME;
-import static datawave.webservice.query.QueryParameters.QUERY_PERSISTENCE;
-import static datawave.webservice.query.QueryParameters.QUERY_STRING;
+import static datawave.webservice.query.QueryParameters.*;
```
phrocker, line 90: ugh, I was intellij'd. Nonetheless this will go into an isolated PR

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IteratorBuildingVisitorTest.java
```diff
@@ -149,7 +149,7 @@ public void buildLiteralRange_middleWildcardTest() throws ParseException {
         Assert.assertTrue(range.getUpper().equals("bar" + Constants.MAX_UNICODE_STRING));
         Assert.assertTrue(range.isUpperInclusive());
     }
-
```
phrocker: Same windows EOL change that snuck in.

apmoriarty:
Several of these ancillary changes are good candidates for independent merge requests
- QueryLogic stack uses generics instead of Entry<Key,Value>
- option to cache TypeMetadata for QueryIterator
- option to use base64 serialization for field sets
- option to ignore RangeStream pruning

phrocker:
> Several of these ancillary changes are good candidates for independent merge requests
> 
>     * QueryLogic stack uses generics instead of Entry<Key,Value>
> 
>     * option to cache TypeMetadata for QueryIterator
> 
>     * option to use base64 serialization for field sets
> 
>     * option to ignore RangeStream pruning

Good call. I appreciate the suggestions. First and last are slam dunks, but the second and third ones I hadn't considered. I'll do those as well. Thanks.

I was on the fence about including two and three into this PR, but I always planned on making them as options solely for the DocumentLogic as they originally had different type metadata -- so in my head I unnecessarily tied them together -- but I'm re-validating assumptions as I submit additional "broken out PRs" from the various branches. Breaking them out is a smart idea since they aren't needed.

As an aside, my hope is that this is purely a reference from the branch where I put it all together to do testing. Would it be better as a WIP PR?

hlgp:
Closing, deferred.

PR #1977, opened 2023-06-01
Remove/detect star imports in repo

avgAGB:


cogross (dismissed):

hlgp (dismissed):

jwomeara (changes requested):
`junit.framework.*` are JUnit 3.x-style asserts.  Do we not want to use `org.junit.*` for JUnit 4.x-style asserts?

avgAGB:
> `junit.framework.*` are JUnit 3.x-style asserts. Do we not want to use `org.junit.*` for JUnit 4.x-style asserts?

I'll add it in for the next commit

mineralntl (dismissed):

cogross (dismissed):

warehouse/accumulo-extensions/src/test/java/datawave/ingest/table/volumeChoosers/ShardedTableDateBasedTieredVolumeChooserTest.java
```diff
@@ -1,8 +1,7 @@
 package datawave.ingest.table.volumeChoosers;
 
-import static org.junit.Assert.*;
+import static junit.framework.TestCase.assertTrue;
```
apmoriarty: `junit.framework` is deprecated. Looks like intellij defaulted to the wrong import.

apmoriarty (changes requested):

warehouse/accumulo-extensions/src/test/java/datawave/ingest/table/volumeChoosers/ShardedTableDateBasedTieredVolumeChooserTest.java
avgAGB: Thanks, will fix.

MiguelRicardos (dismissed):
LGTM

matthpeterson:
> `junit.framework.*` are JUnit 3.x-style asserts. Do we not want to use `org.junit.*` for JUnit 4.x-style asserts?

I interpreted this comment to mean that we should not have `org.junit.*` as a bannedImport because we use it.  I see it's still there so I must be misunderstanding.  Can you help clear this up for me?

I'm also confused how the automated checks are passing.  Maybe this plugin isn't getting executed by them yet.

pom.xml
```diff
@@ -1661,6 +1661,41 @@
...
+                                <phase>process-sources</phase>
+                                <configuration>
+                                    <rules>
+                                        <RestrictImports>
+                                            <!-- Define an explanatory reason why these imports are prohibited -->
+                                            <reason>Remove JUnit Framework</reason>
+                                            <!-- Specify a single pattern to be banned -->
+                                            <bannedImport>junit.framework.*</bannedImport>
+                                            <bannedImport>org.junit.*</bannedImport>
+                                            <bannedImport>**.'*'</bannedImport>
```
matthpeterson, line 1688: I think this bannedImport entry will prevent any imports containing a `*`.  Is that correct?  Is it worth having this entry in a separate group such that the `reason` is correct for it?

avgAGB: Yep that's correct. I'll space this out to another category

avgAGB:
> > `junit.framework.*` are JUnit 3.x-style asserts. Do we not want to use `org.junit.*` for JUnit 4.x-style asserts?
> 
> I interpreted this comment to mean that we should not have `org.junit.*` as a bannedImport because we use it. I see it's still there so I must be misunderstanding. Can you help clear this up for me?
> 
> I'm also confused how the automated checks are passing. Maybe this plugin isn't getting executed by them yet.

My thinking was across the board we don't want to use starred imports, so I added in patterns to cover both junit3/junit4 package patterns. I don't think we necessarily need to rely on org.junit.* in any unit tests, as I've went ahead and refactored any classes which were using it and switched to single imports as needed. The plugin may not be taking effect for the gitlab flow until the branch is merged into integration.

jwomeara:
> > > `junit.framework.*` are JUnit 3.x-style asserts. Do we not want to use `org.junit.*` for JUnit 4.x-style asserts?
> > 
> > 
> > I interpreted this comment to mean that we should not have `org.junit.*` as a bannedImport because we use it. I see it's still there so I must be misunderstanding. Can you help clear this up for me?
> > I'm also confused how the automated checks are passing. Maybe this plugin isn't getting executed by them yet.
> 
> My thinking was across the board we don't want to use starred imports, so I added in patterns to cover both junit3/junit4 package patterns. I don't think we necessarily need to rely on org.junit.* in any unit tests, as I've went ahead and refactored any classes which were using it and switched to single imports as needed. The plugin may not be taking effect for the gitlab flow until the branch is merged into integration.

My initial comment has been addressed by your updates.  The request was to stop using `junit.framework` in our tests and use the equivalent `org.junit` imports.

jwomeara (dismissed):
i'm happy with this

matthpeterson:
> My thinking was across the board we don't want to use starred imports, so I added in patterns to cover both junit3/junit4 package patterns. I don't think we necessarily need to rely on org.junit.* in any unit tests, as I've went ahead and refactored any classes which were using it and switched to single imports as needed. 

I thought that the `org.junit.*` entry specified that even single imports would be banned.  This is based on the first of the three special literals listed under https://github.com/skuzzle/restrict-imports-enforcer-rule#package-patterns, copied below.  It looks like the `'*'` entry is what will look for a literal `*` within the import.

> Package patterns are dot separated strings that can be compared case sensitively part by part. Every part must adhere to the java identifier rules with the exception of a some special literals:
> \* matches every package part but exactly one.
> ** matches multiple package parts but at least one.
> '*' matches a literal * in an import statement.

avgAGB:
> > My thinking was across the board we don't want to use starred imports, so I added in patterns to cover both junit3/junit4 package patterns. I don't think we necessarily need to rely on org.junit.* in any unit tests, as I've went ahead and refactored any classes which were using it and switched to single imports as needed.
> 
> I thought that the `org.junit.*` entry specified that even single imports would be banned. This is based on the first of the three special literals listed under https://github.com/skuzzle/restrict-imports-enforcer-rule#package-patterns, copied below. It looks like the `'*'` entry is what will look for a literal `*` within the import.
> 
> > Package patterns are dot separated strings that can be compared case sensitively part by part. Every part must adhere to the java identifier rules with the exception of a some special literals:
> > * matches every package part but exactly one.
> > ** matches multiple package parts but at least one.
> > '*' matches a literal * in an import statement.

Ah, makes sense then. Yeah the semantics on the banned import regexes is a bit confusing, but the '*' in this case is literal. If I wanted to cover everything underneath a package then the '**' would be used.  I updated the latest commit to reflect this for the junit imports.

matthpeterson (approved):

apmoriarty (approved):

Issue #1978, opened 2023-06-01
Limit days in RangeStream for canary testing and V&V


phrocker:
Initial reqs: 

1) The ability to ability to limit days and remove the original shards per day threshold.
2) Capability must not impact existing ops and must be enabled behind a specialized configuration.

PR #1979, opened 2023-06-01
Create RangeStreamLimitDays and enable bypass for #1978

phrocker:
* Provide a capability to re-enable previous intersection logic for comparison and canary testing with upcoming features. Also remove concept of days, leveraging a "ServiceConfiguration" which is a configuration entity meant to track features that are either only for testing or on the path to deprecation. Two Indexing Configuration options to be enabled via a Profile enables these features. This paves the way for certain query configurations to be built for testing new features and enable easy comparison of prior assumptions. 

A good example of this is that doc specific ranges are costly but may have some benefit, enabling the configurations defined in the QueryLogicFactory help with testing these various assumptions a bit more easily.  The config option isEnableIndexInfoUidToDayIntersectionBypass was a change in logic in January. This re-enables that capability behind a non-default configuration option. 

Add ServiceConfiguration and IndexConfiguration.

update QLF

Fix naming in QLF


warehouse/query-core/src/main/java/datawave/query/tables/RangeStreamScanner.java
```diff
@@ -592,11 +592,11 @@ public Value writeInfoToValue(IndexInfo info) {
         }
     }
     
-    private int dequeue() {
+    int dequeue() {
```
phrocker, line 595: I'll make this protected to be consistent in the next round of changes

hlgp:
OBE.  Some underlying assumptions have since changed.

Issue #1982, opened 2023-06-05
QueryStopwatch for bounded range lookup adds total twice


phrocker:
The line with "total" in it is added to the total summary increasing the total time for query setup. To avoid this I've created a "NonSummarized" TraceStopWatch

PR #1983, opened 2023-06-06
Add ability to not summarize a stop watch that is a total of other stop watches

phrocker:
Solves #1982  and allows the bounded total to not be summarized. It will still be printed in logs but not tabulated in the total.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/util/QueryStopwatch.java
+++ b/warehouse/query-core/src/main/java/datawave/query/util/QueryStopwatch.java
@@ -42,8 +42,6 @@ public class QueryStopwatch {
 
         TraceStopwatch sw = new TraceStopwatch(header);
 
-        System.out.println("ohhh");
-
         return sw;
     }
 
--- a/warehouse/query-core/src/main/java/datawave/query/util/QueryStopwatch.java
+++ b/warehouse/query-core/src/main/java/datawave/query/util/QueryStopwatch.java
@@ -61,7 +61,7 @@ public class QueryStopwatch {
 
     /**
      * Creates a new stopwatch that will be printed at summarization iff nonSummarized is false
-     * @param header header header string to print during summarization
+     * @param header stopwatch header to print during summarization
      * @param nonSummarized boolean to determine if this stopwatch's time is tabulated during summarization
      * @return newly created instance
      */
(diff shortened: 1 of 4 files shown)


warehouse/query-core/src/main/java/datawave/query/util/QueryStopwatch.java
```diff
@@ -24,104 +24,139 @@
...
+    TraceStopwatch newStopwatch(String header) {
         checkNotNull(header);
-        
+
         TraceStopwatch sw = new TraceStopwatch(header);
-        
-        watches.add(Maps.immutableEntry(header, sw));
-        
+
+        System.out.println("ohhh");
```
keith-ratcliffe: This System.out probably no longer needed here :-)

keith-ratcliffe (changes requested):

warehouse/query-core/src/main/java/datawave/query/util/QueryStopwatch.java
phrocker: hahaha thanks

phrocker: my desperate attempt to find out why the debugger wasn't catching a call only to realize I mockito'd myself in the wrong way.

keith-ratcliffe (dismissed):
Looks good

warehouse/query-core/src/main/java/datawave/query/util/QueryStopwatch.java
```diff
@@ -24,104 +24,137 @@
...
+
+        TraceStopwatch sw = new NonSummarizedStopWatch(header);
+
+        return sw;
+    }
+
+
+    /**
+     * Creates a new stopwatch that will be printed at summarization iff nonSummarized is false
+     * @param header header header string to print during summarization
```
hlgp: going for a triple header, impressive

hlgp (dismissed):

warehouse/query-core/src/main/java/datawave/query/util/QueryStopwatch.java
phrocker: haha I'll fix. thanks

hlgp (approved):

hlgp:
OBE

PR #1984, opened 2023-06-06
Breakout shard query logic

phrocker:
Breakout shard query logic  into sharded base query logic to support different return types for upcoming JSON PR.

This is in support of #1974

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/config/Profile.java
+++ b/warehouse/query-core/src/main/java/datawave/query/config/Profile.java
@@ -12,31 +12,31 @@ import org.apache.accumulo.core.data.Value;
 /**
  * Purpose: Provides a mechanism to provide a user configurable way to tune a user's query, such that he or she may take advantage of features that could have
  * negative implications on other user queries.
- * 
+ *
  * Design: Simple interface. User would select the profile through a bean
- * 
+ *
  *
  */
-public interface Profile {
+public interface Profile<T> {
     /**
      * Tune the query logic
-     * 
+     *
      * @param logic
      *            the query logic
      */
-    void configure(BaseQueryLogic<Entry<Key,Value>> logic);
-    
+    void configure(BaseQueryLogic<T> logic);
+
     /**
      * Tune the query planner.
-     * 
+     *
      * @param planner
      *            the query planner.
      */
     void configure(QueryPlanner planner);
-    
+
     /**
      * Tune the query configuration object
-     * 
+     *
      * @param configuration
      *            the query configuration object
      */
--- a/warehouse/query-core/src/main/java/datawave/query/scheduler/Scheduler.java
+++ b/warehouse/query-core/src/main/java/datawave/query/scheduler/Scheduler.java
@@ -1,37 +1,35 @@
 package datawave.query.scheduler;
 
-import java.util.Collection;
-import java.util.Map.Entry;
-
+import com.google.common.collect.Lists;
 import datawave.query.CloseableIterable;
 import datawave.query.config.ShardQueryConfiguration;
 import datawave.query.tables.ScannerFactory;
 import datawave.query.tables.stats.ScanSessionStats;
 import datawave.webservice.query.configuration.QueryData;
-
 import org.apache.accumulo.core.client.BatchScanner;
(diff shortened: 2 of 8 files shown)


tomnelson:
Glad to see this. 
I'm not sure it's clear in this PR, but the parameterization is a great solution for when you won't be returning Key/Value entries, but JSON instead.
Deep inside, I wish that spring configured the *QueryConfiguration (model?) instead of the *QueryLogic (controller?)
That would probably create big problems for other contributors.

warehouse/query-core/src/main/java/datawave/query/tables/BaseScannerSession.java
```diff
@@ -0,0 +1,14 @@
+package datawave.query.tables;
+
+import com.google.common.util.concurrent.AbstractExecutionThreadService;
+
+import java.util.Iterator;
+
+/**
+ * This will handles running a scan against a set of ranges. The actual scan is performed in a separate thread which places the results in a result queue. The
```
tomnelson, line 8: AI did not write this line.

phrocker: Indeed. AI probably would have made a cogent comment, ha. Thanks for pointing this out

tomnelson:
This PR is against a branch called release/version5.1?
I would have expected a branch (or tag?) with that name to be the (final) code that was in the actual 5.1 release and not a moving snapshot.
The desired result is for this to be included in a 5.1.x release? It seems like way more than a fix to 5.1.
I probably put my PRs against the 'integration' branch, which may be waaay out in the future.

warehouse/query-core/src/main/java/datawave/query/tables/ShardedBaseQueryLogic.java
tomnelson, line 1: I would have named this AbstractShardQueryLogic, but I've never been good with names.

phrocker: I wondered if people would be confused about the intent of a ShardQueryLogic and an AbstrctShardQueryLogic so I tried to incorporate "Base," but I've never been good with naming. I've become indifferent about naming. I don't know if I'm the best judge of whether intent is easily understood so I accept a rename is probably a good idea.

warehouse/query-core/src/main/java/datawave/query/config/Profile.java
```diff
@@ -12,31 +12,31 @@
 /**
  * Purpose: Provides a mechanism to provide a user configurable way to tune a user's query, such that he or she may take advantage of features that could have
  * negative implications on other user queries.
- * 
+ *
  * Design: Simple interface. User would select the profile through a bean
- * 
+ *
  *
```
avgAGB, line 18: Small thing, could you add the <T> generic in the annotation with a @param tag? I think maven picks those up with warnings if they aren't included

warehouse/query-core/src/main/java/datawave/query/scheduler/Scheduler.java
```diff
@@ -1,37 +1,35 @@
...
 import org.apache.accumulo.core.client.TableNotFoundException;
-import org.apache.accumulo.core.data.Key;
-import org.apache.accumulo.core.data.Value;
 
-import com.google.common.collect.Lists;
+import java.util.Collection;
+
+/**
+ *
+ */
```
avgAGB, line 17: Similar thing here, would like the generic parameter set in the annotation if there is a block for it. Otherwise removing the empty block could work too.

phrocker: I don't recall if I added the block with intent to fill it out or my IDE did it (I'd have to check settings), but I filled it out. Thanks

warehouse/query-core/src/main/java/datawave/query/config/Profile.java
phrocker: I made these changes locally while I await feedback. Shamefully/Honestly this isn't something I ever really thought about while writing javadoc, so I really appreciate you pointing it out and helping me clue into it.

hlgp:
Closing, deferred.

PR #1985, opened 2023-06-07
Move createBatchScanner method

tomnelson:
I have mixed feelings. On the one hand, the static createBatchScanner method in ShardQueryLogic feels out of place. Its 'config' parameter acts like a 'this' pointer pointing at a 'that'. Even though SQL has-a ShardQueryConfiguration, that is not the one used (and the method is static anyway).
The method seems to work fine as a non-static method in ShardQueryConfiguration, but that class might be a bad place to start dumping actual logic methods instead of just getter/setters.
Thoughts?

(edited)
Maybe the method would be happy in ScannerFactory instead. 
I moved it to ScannerFactory.


hlgp:
I dig it.  It does feel more natural to have it in ScannerFactory

hlgp:
Closing for now.  Will revive when someone can deconflict.

PR #1988, opened 2023-06-07
Updating formatter version and formatting all code

### LINKED ISSUE #1986: Update formatting plugin version
Recently it was discovered that the move to Java11 broke the mvn formatting plugin. We need to update the plugin to a compatible version and run the formatter across the entire codebase to bring everything up to baseline.

mineralntl:


ivakegg (dismissed):

keith-ratcliffe (dismissed):

alerman (dismissed):

tomnelson:
I see a lot of changes that are only the addition of trailing whitespace characters (tabs or spaces) to lines, including blank lines. This current formatting adds the trailing whitespace chars.
Do you really want that?

By updating the formatter version to 2.18.0 it will apply the a new (true by default) property to remove trailing whitespace characters. The downside is that there are many, many more lines in the codebase that already have this trailing whitespace, so the number of changes lines in this PR would be much higher (but you'd have no trailing whitespace chars anywhere).

The decision would be either to have or to not have trailing whitespace characters consistently across the codebase.
One potential issue if you add trailing whitespace during the build, is that some IDEs might remove trailing whitespace chars whenever someone saves a file, and so any PR they submit would include those noisy changes.

Also note that if you leave it like this, and someday someone updates the formatter plugin version without adding a property, they will suddenly see thousands of changes as the trailing whitespace is removed.

ivakegg:
Good point @tomnelson.  In general I might prefer the removal of the trailing whitespace.  I always thought it kinda odd to have blank lines that required spaces.

mineralntl:
We originally chose 2.16.0 to match the metadata-utils repo (since we knew it worked) but I'm not personally against using 2.18.0 - or whatever version that works - as long as we update ALL the repos so we can at least attempt to have some sort of consistency. If there are no particular objections, I can update this PR to use 2.18.0, re-run the formatter, and commit those changes instead.

tomnelson:
The property removeTrailingWhitespace was introduced in version 2.17.0 and defaulted to 'true' in version 2.18.0
I'd probably update our version to 2.18.0 and also include that property (set to true) just to make it clear to devs in the future, and to have it still work if someone ever changes to 2.17.0 

I 'get' how Eclipse likes to place the cursor in a handy spot, pre-indented for typing on each new line, but the effect of using that same Eclipse formatter to format existing files is seen in many project pull requests, especially when star is changed to star-space.

alerman (approved):

hlgp (approved):

keith-ratcliffe (approved):

PR #1990, opened 2023-06-09
No longer duplicate parses on identical key

avgAGB:


apmoriarty (changes requested):
Looks good, can you wrap the debug statements in `if (log.isDebugEnabled())`?

avgAGB:
> Looks good, can you wrap the debug statements in `if (log.isDebugEnabled())`?

Sure, will do on next commit.

apmoriarty (dismissed):

warehouse/query-core/src/test/java/datawave/query/data/parsers/EventKeyTest.java
```diff
@@ -59,6 +59,10 @@ public void testRepeatedParsing() {
 
         parser.parse(eventKeyWithChildUidNoDashes);
         assertKeyWithChildUidNoDashes(parser);
+
+        // this should skip actual processing, since it was already parsed up once.
+        parser.parse(eventKeyWithChildUidNoDashes);
```
ivakegg: this does not actually test that it is not parsed again.  Perhaps you can use a separate key that equals but is not == to the original key.  Then you can test that the cached pointer still points to the original one.

avgAGB: sure, will update this on the next commit

warehouse/query-core/src/main/java/datawave/query/data/parsers/EventKey.java
```diff
@@ -30,10 +31,18 @@ public class EventKey implements KeyParser {
...
 
+    private static final Logger log = Logger.getLogger(EventKey.class);
+
     @Override
     public void parse(Key k) {
-        clearState();
-        this.key = k;
+        if (this.key != null && this.key.equals(k)) {
+            if (log.isDebugEnabled()) {
+                log.debug("Same key given to parse call. Skipping re-load: " + this.key.toString());
```
jschmidt10: If we're using this key for debugging, you might want to use `key.toStringNoTruncate()`. By default, Accumulo's Key.toString() will truncate making it hard to find the problematic key. Applies to the log below as well.

warehouse/query-core/src/main/java/datawave/query/data/parsers/TermFrequencyKey.java
```diff
@@ -1,8 +1,10 @@
 package datawave.query.data.parsers;
 
+import datawave.ingest.util.GenerateMultipleNumShardsCacheFile;
```
jschmidt10: Unused import.

warehouse/query-core/src/test/java/datawave/query/data/parsers/EventKeyTest.java
```diff
@@ -59,6 +61,17 @@ public void testRepeatedParsing() {
 
         parser.parse(eventKeyWithChildUidNoDashes);
         assertKeyWithChildUidNoDashes(parser);
+
+        // this should skip actual processing, since it was already parsed up once.
+        Key dupeKey = new Key("row", "datatype\0d8zay2.3pnndm.anolok.12.34", "FIELD\0value");
+        parser.parse(dupeKey);
+        assertKeyWithChildUidNoDashes(parser);
+        assertNotNull(parser.getDatatype());
```
jschmidt10: Can we assert the values instead of just asserting "not null"? You can shove them in some variables and reuse. Applies to the fi and tf cases below too.

apmoriarty (dismissed):

warehouse/query-core/src/test/java/datawave/query/data/parsers/EventKeyTest.java
avgAGB: yep, will fix this.

warehouse/query-core/src/main/java/datawave/query/data/parsers/EventKey.java
avgAGB: alright, will update this.

jwomeara (changes requested):
I agree with the comments made so far.  Happy to approve once the changes are made.

apmoriarty (approved):

warehouse/query-core/src/main/java/datawave/query/data/parsers/EventKey.java
```diff
@@ -31,10 +32,18 @@ public class EventKey implements KeyParser {
...
     public void parse(Key k) {
-        clearState();
-        this.key = k;
+        if (this.key != null && this.key.equals(k)) {
+            if (log.isDebugEnabled()) {
+                log.debug("Same key given to parse call. Skipping re-load: " + this.key.toStringNoTruncate());
+            }
+        } else {
+            clearState();
+            this.key = k;
```
dtspence, line 45: I don't know if this would ever occur - however, there are a few mutable methods on `Key`. Should/could the caller's `k` always get assigned to `key`? This comment would carry to the other `*Key` types too.

PR #1991, opened 2023-06-09
Most recent unique feature

ivakegg:


warehouse/query-core/src/test/java/datawave/query/UniqueTest.java
```diff
@@ -351,4 +354,34 @@ public void testUniquenessWithModelAliases() throws Exception {
         String queryString = "UUID:/^[CS].*/ AND #UNIQUE(BOTH_NULL)";
         runTestQueryWithUniqueness(expected, queryString, startDate, endDate, extraParameters);
     }
+
+    @Test
+    public void testMostRecentUniqueness() throws Exception {
```
hlgp, line 377: nice

warehouse/query-core/src/test/java/datawave/query/UniqueTest.java
```diff
@@ -351,4 +354,34 @@ public void testUniquenessWithModelAliases() throws Exception {
...
+        extraParameters.put("unique.fields", "$DEATH_DATE,BIRTH_DATE");
+        runTestQueryWithUniqueness(expected, queryString, startDate, endDate, extraParameters);
+
+        expected.add(Sets.newHashSet(WiseGuysIngest.sopranoUID));
+        expected.add(Sets.newHashSet(WiseGuysIngest.corleoneUID));
+        expected.add(Sets.newHashSet(WiseGuysIngest.caponeUID));
+        extraParameters.put("unique.fields", "death_date,birth_date");
+        runTestQueryWithUniqueness(expected, queryString, startDate, endDate, extraParameters);
+    }
+
```
hlgp, line 411: I might have missed it, but do we have a test to make sure this also works with the #UNIQUE variant, just to make sure it doesn't break in future?

hlgp: Maybe some more negative cases, too

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -1287,7 +1287,7 @@ protected void setupRowBasedHdfsBackedSet(String row) throws IOException {
                 this.createdRowDir = false;
             }
 
-            this.set = new HdfsBackedSortedSet<>(null, hdfsBackedSetBufferSize, ivaratorCacheDirs, row, maxOpenFiles, numRetries, persistOptions,
+            this.set = new HdfsBackedSortedSet<>(null, null, hdfsBackedSetBufferSize, ivaratorCacheDirs, row, maxOpenFiles, numRetries, persistOptions,
```
jwomeara: 9 parameters for this constructor now.  At what point do we switch to a builder?

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
```diff
@@ -1648,9 +1659,10 @@ public UniqueFields getUniqueFields() {
     }
 
     public void setUniqueFields(UniqueFields uniqueFields) {
-        this.uniqueFields = uniqueFields;
-        // If unique fields are present, make sure they are deconstructed by this point.
-        if (uniqueFields != null) {
+        if (uniqueFields == null) {
+            this.uniqueFields = new UniqueFields();
```
jwomeara: It feels odd to set uniqueFields to a new instance when null is passed.  In what cases do we do this?  It wasn't obvious to me when I loaded the code in my ide.  Also, just an observation that the uniqueFields parameter is initialized as `new UniqueFields()` when ShardQueryConfiguration is created.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1557,11 +1557,11 @@ public Comparator<Object> getValueComparator(Tuple3<Key,Document,Map<String,Obje
         return new ValueComparator(from.second().getMetadata());
     }
 
-    protected UniqueTransform getUniqueTransform() {
+    protected UniqueTransform getUniqueTransform() throws IOException {
         if (uniqueTransform == null && getUniqueFields() != null && !getUniqueFields().isEmpty()) {
             synchronized (getUniqueFields()) {
                 if (uniqueTransform == null) {
-                    uniqueTransform = new UniqueTransform(getUniqueFields());
+                    uniqueTransform = new UniqueTransform(this, getUniqueFields());
```
jwomeara: I get why it was done (to gain access to a bunch of the query iterator's variables), but passing the query iterator to the unique transform seems odd to me.  I would be in favor of expanding the constructor params here to avoid passing the entire query iterator, adding a builder for UniqueTransform, or possibly adding a factory method to the query iterator that creates UniqueTransform objects...  Just my 2 cents.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -849,7 +854,7 @@ public void setHdfsSiteConfigURLs(String hadoopConfigURLs) {
     }
 
     public FileSystemCache getFileSystemCache() throws MalformedURLException {
-        if (this.fsCache == null && this.hdfsSiteConfigURLs != null) {
```
jwomeara: If getFileSystemCache always returns a non-null object, we have some downstream code that will be affected.  In QueryLock.java, if fsCache (which is set via getFileSystemCache) is never null, we will never create a lock using the hdfsSiteConfigs param.  If that's intentional, then we should remove that code path.

Also, if we allow a file system cache to be created without hdfs configs, we shouldn't expect to be able to access any hdfs paths, right?  If so, then there are a few downstream locations that should be updated because they assume that if this returns a non-null FileSystemCache that hdfs paths will be accessible (e.g. QueryLock.build())

We should make sure there are no other places where this assumption is made.

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -582,7 +584,11 @@ protected String getStopwatchHeader(ShardQueryConfiguration config) {
     @Override
     public QueryLogicTransformer getTransformer(Query settings) {
         if (this.transformerInstance != null) {
-            addConfigBasedTransformers();
+            try {
+                addConfigBasedTransformers();
+            } catch (QueryException qe) {
+                throw new RuntimeException("Unable to configure transformers", qe);
```
jwomeara: Why not throw the runtime exception directly from the addConfigBasedTransformers method?

Also, do we need to throw a runtime exception?  Is the QueryException being swallowed up somewhere?

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -625,7 +635,11 @@ private void addConfigBasedTransformers() {
                 if (alreadyExists != null) {
                     ((UniqueTransform) alreadyExists).updateConfig(getConfig().getUniqueFields(), getQueryModel());
                 } else {
-                    ((DocumentTransformer) this.transformerInstance).addTransform(new UniqueTransform(this, getConfig().getUniqueFields()));
+                    try {
+                        ((DocumentTransformer) this.transformerInstance).addTransform(new UniqueTransform(this, getConfig().getUniqueFields()));
```
jwomeara: Same comment here as I made above w.r.t. passing QueryIterator to initialize UniqueTransform, only this time we're using ShardQueryLogic.  

It seems like a bad practice to pass the whole object when we are only using parts of it.  What's to stop a developer from modifying QueryIterator or ShardQueryLogic from within the UniqueTransform constructor?

jwomeara: Are we more likely to see IOExceptions from the HDFSBackedSortedSet constructor now that FileSystemCache can return a FileSystem without valid hdfsSiteConfigs?

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
```diff
@@ -387,7 +387,7 @@ protected void pruneEmptyOptions(IteratorSetting settings) {
      *            an {@link IteratorSetting}
      */
     protected void pruneIvaratorConfigs(ASTJexlScript script, IteratorSetting settings) {
-        if (script != null && !IvaratorRequiredVisitor.isIvaratorRequired(script)) {
+        if (script != null && !IvaratorRequiredVisitor.isIvaratorRequired(script) && !settings.getOptions().containsKey(QueryOptions.MOST_RECENT_UNIQUE)) {
```
jwomeara: Probably more efficient to do the MOST_RECENT_UNIQUE check before calling the visitor.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -1,39 +1,65 @@
...
 /**
  * This iterator will filter documents based on uniqueness across a set of configured fields. Only the first instance of an event with a unique set of those
- * fields will be returned. This transform is thread safe.
+ * fields will be returned unless mostRecentUnique is specified in which case the most recent instance of an event will be returned. This transform is thread
+ * safe.
+ *
+ * 1) FileByteKeySortedSet: ability to keep the most recent key for the same byte array done: setRewriteStrategy(rewriteStrategy) 2) MultiSetBackedSortedSet:
+ * ability to keep the mote recent key for the same value when doing a merge sort done 3) Buffer all values in the file backed sorted set until we flush done 4)
+ * Update the ShardQueryLogic to use the MostRecentUniqueTransform done 5) Create a MostRecentUniqueIterator akin to the GroupingIterator done 6) Update the
+ * QueryIterator to apply the MostRecentUniqueIterator done 7) Create most recent test cases
```
jwomeara: This reads like notes that were used during feature development.  Consider rewriting?

warehouse/query-core/src/main/java/datawave/query/util/sortedset/BufferedFileBackedSortedSet.java
```diff
@@ -21,7 +21,7 @@
  * @param <E>
  *            type of the set
  */
-public class BufferedFileBackedSortedSet<E> implements SortedSet<E> {
+public class BufferedFileBackedSortedSet<E> implements RewritableSortedSet<E>, SortedSet<E> {
```
jwomeara: RewritableSortedSet extends SortedSet, so SortedSet is redundant here.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -1,39 +1,65 @@
...
 import java.util.Map.Entry;
 import java.util.TreeSet;
-import java.util.stream.Collectors;
+import java.util.UUID;
 
 /**
  * This iterator will filter documents based on uniqueness across a set of configured fields. Only the first instance of an event with a unique set of those
- * fields will be returned. This transform is thread safe.
+ * fields will be returned unless mostRecentUnique is specified in which case the most recent instance of an event will be returned. This transform is thread
+ * safe.
```
jwomeara, line 54: This object might be thread safe within a single QueryIterator instance, but are we going to be thread safe across multiple scans/QueryIterator instances when accessing the same file backed sorted set?  How often will the transformer reread the files from the file backed sorted set?  Is it not possible that two scans could update the same value at the same time?

warehouse/query-core/src/main/java/datawave/query/util/sortedset/FileByteDocumentSortedSet.java
```diff
@@ -0,0 +1,234 @@
...
+     *            the sorted set
+     * @param handler
+     *            the sorted set file handler
+     */
+    public FileByteDocumentSortedSet(RewritableSortedSet<Map.Entry<byte[],Document>> set, SortedSetFileHandler handler) {
+        super(set, new ByteDocumentFileHandler(handler), new FileByteDocumentSortedSet.Factory());
+    }
+
+    /**
+     * Create an sorted set out of another sorted set. If persist is true, then the set will be directly persisted using the set's iterator which avoid pulling
```
jwomeara:
```suggestion
     * Create a sorted set out of another sorted set. If persist is true, then the set will be directly persisted using the set's iterator which avoids pulling
```

warehouse/query-core/src/main/java/datawave/query/util/sortedset/FileKeyValueSortedSet.java
```diff
@@ -0,0 +1,233 @@
...
+     *            the sorted set
+     * @param handler
+     *            the sorted set file handler
+     */
+    public FileKeyValueSortedSet(RewritableSortedSet<Map.Entry<Key,Value>> set, SortedSetFileHandler handler) {
+        super(set, new KeyValueFileHandler(handler), new FileKeyValueSortedSet.Factory());
+    }
+
+    /**
+     * Create an sorted set out of another sorted set. If persist is true, then the set will be directly persisted using the set's iterator which avoid pulling
```
jwomeara, line 114:
```suggestion
     * Create a sorted set out of another sorted set. If persist is true, then the set will be directly persisted using the set's iterator which avoids pulling
```

warehouse/query-core/src/main/java/datawave/query/util/sortedset/FileSortedSet.java
```diff
@@ -573,79 +601,68 @@ public void clear() {
...
         }
     }
 
     @Override
     public E last() {
-        boolean gotLast = false;
-        E last = null;
         if (persisted) {
+            boolean gotLast = false;
+            E last = null;
```
jwomeara: I think these changes will go away once you merge integration in.

warehouse/query-core/src/main/java/datawave/query/util/sortedset/RewritableSortedSet.java
```diff
@@ -0,0 +1,35 @@
...
+         * Determine if the object should be rewritten
+         *
+         * @param original
+         * @param update
+         * @return true of the original should be replaced with the update
+         */
+        boolean rewrite(E original, E update);
+    }
+
+    RewritableSortedSetImpl.RewriteStrategy getRewriteStrategy();
```
jwomeara:
```suggestion
    RewritableSortedSetImpl.RewriteStrategy<E> getRewriteStrategy();
```

warehouse/query-core/src/main/java/datawave/query/util/sortedset/RewritableSortedSetImpl.java
```diff
@@ -0,0 +1,318 @@
...
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.NavigableMap;
+import java.util.NoSuchElementException;
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.function.Predicate;
+
+/**
+ * A sorted set that can use a replacement strategy to determine when adding an element that is already in the set should result in replacing that element. This
```
jwomeara:
```suggestion
 * A sorted set that can use a replacement strategy to determine when to replace an element that is already contained in the set. This
```

warehouse/query-core/src/main/java/datawave/query/util/sortedset/rfile/RFileKeyValueInputStreamBase.java
```diff
@@ -1,17 +1,18 @@
...
 import org.apache.accumulo.core.data.Range;
 import org.apache.accumulo.core.data.Value;
 
-public class RFileKeyInputStream implements FileSortedSet.SortedSetInputStream<Key> {
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Iterator;
+import java.util.Map;
+
+public class RFileKeyValueInputStreamBase {
```
jwomeara: Should this class be abstract, with the readObject method from FileSortedSet.SortedSetInputStream being abstract?

warehouse/query-core/src/test/java/datawave/query/util/sortedset/MultiSetBackedSortedSetTest.java
```diff
@@ -8,6 +8,8 @@
 import java.util.Arrays;
 import java.util.Comparator;
 import java.util.Iterator;
+import java.util.List;
+import java.util.NoSuchElementException;
```
jwomeara: Unused.

jwomeara (changes requested):

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
ivakegg: I was trying for "null" meaning to clear out the existing unique field list.  However I do not want a null member so I was creating a new empty class.  I could change this to be this.uniqueFields.clear() ot something like that.

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
ivakegg: Agreed.  Created a builder pattern.

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
ivakegg: Updated the copy to be final adn instead created appropriate set methods.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
ivakegg: Created a builder pattern.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
ivakegg: I will have to work through why I did this.  There is a null check in the FileSystemCache constructor.

ivakegg: I see now.  One can correclty create file systems even if you do not supply additional hdfs configuration files.  The local system may already have the hdfs configuration setup to be loaded.  I believe this is ok.

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
ivakegg: Throwing QueryException from this method creates a whole cascade of methods that need to throw QueryException: getTransformer(), getExtendedTransformer(), getTransformIterator(), and potentially others.  That is not a great excuse however so I will follow adding exceptions all the way through.

ivakegg: ok, updating to pass QueryException through

warehouse/query-core/src/test/java/datawave/query/UniqueTest.java
ivakegg: Yes, we do have #UNIQUE tests in this very class.

ivakegg: What do you mean by negative cases?

ivakegg: You cannot negate the #UNIQUE function, or rather the negation would not result in any difference.

warehouse/query-core/src/main/java/datawave/query/util/sortedset/FileSortedSet.java
ivakegg: Not sure what you meant here.  integration is merged in.

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
ivakegg: I created a builder for the unique transform.  I am hoping this is enough to mitigate your concerns.  The knowledge as to what pieces are required to create an appropriate hdfs backed sorted set is specific to the transform.  Also the set only needs to be created if the mostRecent flag is on.  Let me know if what I have done is not sufficient for you.

I also added comments in the builder method denoting that the logic and queryIterator cannot be cached or modified in any way.

ivakegg: Only if we do not configure the system correctly.  If we are using file based backed sorted sets, then an hdfs-site.xml will not need to be supplied.

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
ivakegg: good call.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
ivakegg: oops, left my notes in here.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
ivakegg: You are correct.  Added synchronization appropriately around the file backed sorted set.  Good catch!

ivakegg: indeed I missed synchronizing around the file backed sorted set.... good catch!

warehouse/query-core/src/main/java/datawave/query/util/sortedset/rfile/RFileKeyValueInputStreamBase.java
ivakegg: sure, why not.

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
jwomeara: I have some feedback for this change in a separate branch: https://github.com/NationalSecurityAgency/datawave/pull/2065

I cleaned up the builder usage so that you don't need to override all of the base builder's methods just to call the super implementation...  That kind of defeats the purpose of having a base and extended version of the builder, does it not?

I also updated the formatting around all of the builder calls to make it easier to read.

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
jwomeara: So now we have a set method on UniqueFields which copies the internals of `that` UniqueFields object into `this` UniqueFields object?  Why?  Is there a good reason not to just have `ShardQueryConfiguration.setUniqueFields` set the uniqueFields class member to the one that was passed in?  Does that variable need to be final for some reason that I'm not understanding?

Also, if we're going to call `uniqueFields.deconstructIdentifierFields` every time we set the unique fields, why not call that method from the uniqueFields `set` method?

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
jwomeara: The builder you've implemented isn't really what I had in mind with my comment.  It is strange to me that the query iterator (or query logic for that matter) should be necessary to build a unique transform object.  The unique transform doesn't use the query iterator or query logic in _any_ capacity.  The query iterator and query logic are simply a delivery mechanism for the multiple objects ACTUALLY needed by the unique transform.  To be clear, I think the builder should have setters for the objects that are actually needed by the unique transform.  

I have put an example of what I would like to see in a PR: https://github.com/NationalSecurityAgency/datawave/pull/2065

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
jwomeara: Your comment does not address my concern about the downstream code in QueryLock.java that is affected by this change.  With this change, there are inaccessible code paths in QueryLock that should be removed...

For my second concern, it is possible now to create a FileSystemCache without hdfs configs, while having an hdfs ivarator URI configured.  In QueryLock.build(), this combination would send us down a path where we try to use a FileSystemCache NOT configured for hdfs with an hdfs ivarator uri path.  How is that not an issue?

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
jwomeara: Comments won't prevent the query logic or iterator from being accessed in a way that could impact our query.  Not passing them in the first place does prevent this though.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
jwomeara: I'm not sure that your `synchronized` additions address my concern.  I'm not worried about two threads trying to access the same sorted set instance.  My concern is about multiple scans against the shard table reading/writing from/to the same set of files backing the sorted set.  

For instance let's imagine that our query results in `N` subplans.  Each of those `N` subplans is going to run in a separate QueryIterator stack on a bunch of different tablet servers.  Each of those QueryIterator stacks is going to setup it's own file backed sorted set object, but those sorted sets are all going to be backed by the same file (whether it be in hdfs or the local filesystem).  So, now we have `N` separate threads/processes all reading and potentially modifying the same file concurrently.  That sounds like a problem to me.  Am I missing something?

I don't think synchronized access to the sorted set object is the issue.  The issue is simultaneous access to the same file by multiple threads/nodes/tservers.

warehouse/query-core/src/main/java/datawave/query/util/sortedset/FileSortedSet.java
jwomeara: On June 29th when I wrote this comment, integration hadn't been merged in and there were some changes missing here.

warehouse/query-core/src/main/java/datawave/query/util/sortedset/rfile/RFileKeyValueInputStreamBase.java
jwomeara: I should have explained this in a bit more detail.  Every class that extends RFileKeyValueInputStreamBase also implements the SortedSetInputStream interface.  So, my question should have been "why not have RFileKeyValueInputStreamBase implement FileSortedSet.SortedSetInputStream, leaving the `readObject` method to be implemented by each extending class?"

I added a commit to my PR to show you what I meant: https://github.com/NationalSecurityAgency/datawave/pull/2065

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
ivakegg: merged

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
ivakegg: ok, I see your point.  I will revert this line and update the test cases appropriately.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
ivakegg: Each query iterator will be using its own unique transform and its own file backed sorted set.  The directory for the sorted set will be unique by query id and scan id (set on the webserver side).

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
ivakegg: ok, no problem.  Merged your updates in.

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
ivakegg: Updated to avoid the final thing and set the class appropriately.

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
ivakegg: got it.

warehouse/query-core/src/main/java/datawave/query/util/sortedset/rfile/RFileKeyValueInputStreamBase.java
ivakegg: I think it is now organized as you wished it.

warehouse/query-core/src/test/java/datawave/query/UniqueTest.java
hlgp: No, I meant that in addition to a test adding just unique.fields that we should have an additional test with most recent set to true using #UNIQUE or f:unique in the query.  I saw that there were already standalone tests of these... thanks....  to further elaborate, I would expect

Q1:

- extraParameters.put(QueryParameters.MOST_RECENT_UNIQUE, "true");
- extraParameters.put("unique.fields", "DEATH_DATE,$MAGIC");
- String queryString = "UUID =~ '^[CS].*'";

To return the same results as

Q2:
- extraParameters.put(QueryParameters.MOST_RECENT_UNIQUE, "true");
- String queryString = "UUID =~ '^[CS].*' && f:unique(death_date,magic)";

but, for some reason that does not appear to be happening in my local testing.

By negative testing, I mean supplying invalid or contradictory input to confirm desired behavior / error handling.

ivakegg: You are absolutely correct, the most recent flag is getting cleared in the second query.  Fixing

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -634,7 +641,26 @@ private void addConfigBasedTransformers() {
...
-                    ((DocumentTransformer) this.transformerInstance).addTransform(new UniqueTransform(this, getConfig().getUniqueFields()));
+                    try {
+                        // @formatter:off
+                        ((DocumentTransformer) this.transformerInstance).addTransform(new UniqueTransform.Builder()
+                                .withUniqueFields(getConfig().getUniqueFields())
+                                .withModel(getQueryModel())
+                                .withBufferPersistThreshold(getUniqueCacheBufferSize())
+                                .withIvaratorCacheDirConfigs(getIvaratorCacheDirConfigs())
+                                .withHdfsSiteConfigURLs(getHdfsSiteConfigURLs())
+                                .withSubDirectory(getConfig().getQuery().getId().toString())
```
matthpeterson, line 658: QueryIterator is using
```                            .withSubDirectory(getQueryId() + "-" + getScanId())```
Are these intentionally different?

File: warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java [hunk 21]
```diff
@@ -4,89 +4,97 @@
...
-            modelMapping = HashMultimap.create();
-            // reverse the reverse query mapping which will give us a mapping from the final field name to the original field name(s)
-            for (Map.Entry<String,String> entry : model.getReverseQueryMapping().entrySet()) {
-                modelMapping.put(entry.getValue(), entry.getKey());
+    public void updateConfig(UniqueFields uniqueFields, QueryModel model) {
+        // only reset the bloom filter if changing the field set
+        if (!this.uniqueFields.equals(uniqueFields)) {
+            this.uniqueFields = uniqueFields.clone();
+            log.info("Resetting unique fields on the unique transform");
+            this.bloom = BloomFilter.create(new ByteFunnel(), 500000, 1e-15);
```
matthpeterson, line 83: Perhaps these arguments would be worth making configurable, maybe in a future iteration.

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
matthpeterson: Maybe that's just how it works - forgive my ignorance on this one.

warehouse/query-core/src/test/java/datawave/query/util/sortedset/BufferedFileBackedRewritableSortedSetTest.java
```diff
@@ -0,0 +1,82 @@
...
+    public abstract V createValue(byte[] values);
+
+    public abstract void testFullEquality(Map.Entry<K,V> expected, Map.Entry<K,V> value);
+
+    @Test
+    public void testRewrite() throws Exception {
+        // create a new set of data, only half of which has greater Values
+        Map.Entry<K,V>[] data2 = new Map.Entry[template.length * 2];
+        for (int i = 0; i < template.length; i++) {
+            byte[] buffer = new byte[i + 11];
```
matthpeterson, line 60: These tests may benefit from additional comments.

warehouse/query-core/src/test/java/datawave/query/util/WiseGuysIngest.java
```diff
@@ -597,116 +613,130 @@ public static void writeItAll(AccumuloClient client, WhatKindaRange range) throw
...
-            mutation.put("fi\u0000" + "AGE", numberType.normalize("18") + "\u0000" + datatype + "\u0000" + sopranoUID, columnVisibility, timeStamp, emptyValue);
+            mutation.put("fi\u0000" + "AGE", numberType.normalize("16") + "\u0000" + datatype + "\u0000" + sopranoUID, columnVisibility,
+                            timeStamp + sopranoTimeStampDelta, emptyValue);
+            mutation.put("fi\u0000" + "AGE", numberType.normalize("18") + "\u0000" + datatype + "\u0000" + sopranoUID, columnVisibility,
+                            timeStamp + sopranoTimeStampDelta, emptyValue);
 
             // geo
             for (String normalized : ((OneToManyNormalizerType<Geometry>) geoType).normalizeToMany("POINT(20 20)")) {
-                mutation.put("fi\u0000" + "GEO", normalized + "\u0000" + datatype + "\u0000" + corleoneUID, columnVisibility, timeStamp, emptyValue);
+                mutation.put("fi\u0000" + "GEO", normalized + "\u0000" + datatype + "\u0000" + sopranoUID, columnVisibility, timeStamp + sopranoTimeStampDelta,
```
matthpeterson, line 687: Most of these Keys only changed timestamps and UIDs did not change.  This one's UID and timestamp went form corleone to soprano.  I'm wondering if that was a mistake.

matthpeterson: Disregard.  I think it's meant to align with the prior keys.  I see a similar change below.

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
ivakegg: Yes, that is intentional.

warehouse/query-core/src/test/java/datawave/query/util/sortedset/BufferedFileBackedRewritableSortedSetTest.java
ivakegg: ok

warehouse/query-core/src/test/java/datawave/query/UniqueTest.java
```diff
@@ -354,4 +371,62 @@ public void testUniquenessWithModelAliases() throws Exception {
...
+        extraParameters.put("unique.fields", "DEATH_DATE,$MAGIC");
+        String queryString = "UUID =~ '^[CS].*'";
+        runTestQueryWithUniqueness(expected, queryString, startDate, endDate, extraParameters);
+
+        extraParameters.clear();
+        extraParameters.put(QueryParameters.MOST_RECENT_UNIQUE, "true");
+        queryString = "UUID =~ '^[CS].*' && f:unique(death_date,magic)";
+        runTestQueryWithUniqueness(expected, queryString, startDate, endDate, extraParameters);
+
+        this.getClass().getMethod("testHannahHypothesis").getName().replace("Hypothesis", "Theory");
```
hlgp, line 430: NICE!

matthpeterson (dismissed):

jwomeara:
Since this is a big PR, please merge your changes up into the feature/queryMicroservices branch locally and ensure there are no major conflicts that need to be addressed.  Ideally, you would merge this up into feature/queryMicroservices after it is merged into integration.

ivakegg:
@jwomeara The merge into the query microservices branch is relatively clean.  The only conflicts were imports.

jwomeara (approved):

apmoriarty (approved):

PR #1992, opened 2023-06-09
Fix view name

MiguelRicardos:
Allows querying on the column qualifier column so that you can filter results on content.view.name optional parameter.


warehouse/query-core/src/main/java/datawave/query/tables/content/ContentQueryTable.java
```diff
@@ -54,7 +54,9 @@ public class ContentQueryTable extends BaseQueryLogic<Entry<Key,Value>> {
     private static final String ALL = "\u10FFFF";
 
     private int queryThreads = 100;
-    private ScannerFactory scannerFactory;
+
+    ScannerFactory scannerFactory;
```
apmoriarty, line 59: Does the unit test require public access? I would prefer this stay private and any tests use the getter

warehouse/query-core/src/test/java/datawave/query/tables/content/ContentQueryTableTest.java
```diff
@@ -0,0 +1,200 @@
...
+
+    @Before
+    public void setup() throws TableNotFoundException {
+        contentQueryTable = new ContentQueryTable();
+        mockScannerFactory = mock(ScannerFactory.class);
+        mockScanner = mock(BatchScanner.class);
+        mockGenericConfig = mock(GenericQueryConfiguration.class);
+        mockContentConfig = mock(ContentQueryConfiguration.class);
+
+        contentQueryTable.scannerFactory = mockScannerFactory;
```
apmoriarty, line 56: Could this be:
```
        contentQueryTable.setScannerFactory(mockScannerFactory);
```

MiguelRicardos: making the change right now

warehouse/query-core/src/main/java/datawave/query/tables/content/ContentQueryTable.java
MiguelRicardos: fair point

warehouse/query-core/pom.xml
```diff
@@ -338,6 +338,24 @@
...
+            <artifactId>junit-jupiter</artifactId>
+            <version>RELEASE</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <version>RELEASE</version>
+            <scope>test</scope>
+        </dependency>
```
jwomeara: duplicate entries.  i would recommend picking a specific version vs. using RELEASE, and add a dependency management entry in the root pom, using a property for the version.  just my 2 cents.

warehouse/query-core/pom.xml
```diff
@@ -338,6 +338,24 @@
...
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>3.3.3</version>
+            <scope>test</scope>
+        </dependency>
```
jwomeara: easymock is already a test dependency for query core - speaking as someone who doesn't understand the finer points of mockito vs easymock vs. any other mock api, why choose mockito?  Is it better than the other options in some way? 
 genuinely curious...  

powermock api mockito is defined in the root pom which includes mockito-core:3.3.3 transitively.  would it make sense to use that?

otherwise, please add this dependency to the root pom's dependency management section, and break the version out into a property.

warehouse/query-core/src/main/java/datawave/query/tables/content/ContentQueryTable.java
```diff
@@ -260,4 +277,8 @@ public Set<String> getRequiredQueryParameters() {
     public Set<String> getExampleQueries() {
         return Collections.emptySet();
     }
+
+    public Set<Authorizations> getAuthorizations() {
+        return Collections.singleton(this.scanner.getAuthorizations());
+    }
```
jwomeara: Doesn't look like this method is used.  Were you using it for testing at some point?

warehouse/query-core/src/main/java/datawave/query/tables/content/ContentQueryTable.java
```diff
@@ -55,7 +55,9 @@ public class ContentQueryTable extends BaseQueryLogic<Entry<Key,Value>> {
     private static final String ALL = "\u10FFFF";
 
     private int queryThreads = 100;
-    private ScannerFactory scannerFactory;
+
+    ScannerFactory scannerFactory;
```
jwomeara, line 59: If this is package-private, do we really need the getter/setter since they are only used in a test sharing the same package?

warehouse/query-core/src/main/java/datawave/query/tables/content/ContentQueryTable.java
```diff
@@ -55,7 +55,9 @@ public class ContentQueryTable extends BaseQueryLogic<Entry<Key,Value>> {
...
+
     private String viewName = null;
```
jwomeara: This could also be package private if it's only used for tests.

jwomeara (changes requested):

warehouse/query-core/pom.xml
MiguelRicardos: rgr, I am removing and moving.

warehouse/query-core/src/main/java/datawave/query/tables/content/ContentQueryTable.java
MiguelRicardos: rgr, fixing.

warehouse/query-core/src/main/java/datawave/query/tables/content/ContentQueryTable.java
MiguelRicardos: rgr, fixing.

warehouse/query-core/src/main/java/datawave/query/tables/content/ContentQueryTable.java
MiguelRicardos: I planned to use this for a test but then did not need it. Will remove.

warehouse/query-core/pom.xml
MiguelRicardos: I've mainly used Mockito and just went with what has worked. I moved it up and broke out the version number to a property. On this backport, I had to specify a different version of Mokito https://github.com/NationalSecurityAgency/datawave/pull/2014/files due to a transitive dependency issue. I think I prefer the specific dependency version, just so there's no confusion.

apmoriarty (dismissed):

pom.xml
```diff
@@ -73,7 +73,7 @@
         <version.jjwt>0.11.2</version.jjwt>
         <version.jts>1.19.0</version.jts>
         <version.junit>4.13.2</version.junit>
-        <version.junit.jupiter>5.5.2</version.junit.jupiter>
+        <version.junit.jupiter>5.9.3</version.junit.jupiter>
         <version.junit.vintage>5.5.2</version.junit.vintage>
```
jwomeara: Can we keep the junit jupiter and vintage versions in sync, please?

jwomeara (changes requested):
LGTM aside from the out of sync junit versions.

pom.xml
MiguelRicardos: Fixed this version. Rebase is making the diff look weird. Checking it out right now.

jwomeara (approved):

apmoriarty (approved):

PR #1993, opened 2023-06-11
add the ability to use more precise field datatype regular expressions

hlgp:


apmoriarty (dismissed):

avgAGB (dismissed):

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
```diff
@@ -91,6 +92,12 @@ public abstract class BaseIngestHelper extends AbstractIngestHelper implements C
      */
     public static final String FIELD_TYPE = ".data.field.type.class";
 
+    /**
+     * Configuration parameter to specify whether to specify whether to use type regex hierarchy. If true, this will use the datatype associated with the "most
```
jwomeara:
```suggestion
     * Configuration parameter to specify whether to use type regex hierarchy. If true, this will use the datatype associated with the "most
```

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
```diff
@@ -602,16 +613,43 @@ public List<datawave.data.type.Type<?>> getDataTypes(String fieldName) {
...
+                Collection<datawave.data.type.Type<?>> patternTypes = typeCompiledPatternMap.get(patternMatcher);
 
                 if (patternMatcher.reset(fieldName).matches()) {
-                    Collection<datawave.data.type.Type<?>> patternTypes = typeCompiledPatternMap.get(patternMatcher);
-                    types.addAll(patternTypes);
-                    typeFieldMap.putAll(fieldName, patternTypes);
+
+                    if (useMostPreciseFieldTypeRegex) {
+                        int patternLength = patternMatcher.pattern().toString().length();
+                        if (patternLengths.contains(Integer.valueOf(patternLength))) {
```
jwomeara: Intellij recommends removing the Integer.valueOf boxing.

jwomeara (changes requested):
looks good.  minor nit

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
hlgp: Nice!  Sometimes I can't help repeating myself repeating myself.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
hlgp: changed in next commit

jwomeara (dismissed):

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
```diff
@@ -602,16 +614,43 @@ public List<datawave.data.type.Type<?>> getDataTypes(String fieldName) {
...
+                Collection<datawave.data.type.Type<?>> patternTypes = typeCompiledPatternMap.get(patternMatcher);
 
                 if (patternMatcher.reset(fieldName).matches()) {
-                    Collection<datawave.data.type.Type<?>> patternTypes = typeCompiledPatternMap.get(patternMatcher);
-                    types.addAll(patternTypes);
-                    typeFieldMap.putAll(fieldName, patternTypes);
+
+                    if (useMostPreciseFieldTypeRegex) {
+                        int patternLength = patternMatcher.pattern().toString().length();
+                        if (patternLengths.contains(patternLength)) {
```
ivakegg, line 628: I think we only care if we have 2 or more patterns with the bestMatch length.

ivakegg (changes requested):
see comments

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
ivakegg: OK, perhaps a warning in any case is good.  Lets update the comment to include that this implies that the normalizer for only one of them will be used.  Also this may be acceptible to the user if we have the same normalizer for each of them.  Hence log.error should be log.warn instead.

ivakegg (approved):

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
```diff
@@ -602,16 +614,43 @@ public List<datawave.data.type.Type<?>> getDataTypes(String fieldName) {
                 compilePatterns();
             }
 
+            List<Integer> patternLengths = new ArrayList<>();
+            int bestMatch = 0;
+            Collection<datawave.data.type.Type<?>> bestMatchTypes = null;
+
             for (Matcher patternMatcher : typeCompiledPatternMap.keySet()) {
```
ivakegg, line 621: Additional suggestions that are for extra credit only:

1) only supply a warning message when we have two or more regexes with the same length that have different normalizers
2) If those two or more regexes are the final "bestMatch", and there are differing normalizers, then supply an error message instead
    2.a) Lets reevaluate whether we should fail in this circumstance

ivakegg (approved):

apmoriarty (approved):

PR #1996, opened 2023-06-13
Investigate and update MultiSetBackedSortedSet isEmpty(), first(), and last()

MiguelRicardos:
… updated first(), last(), and isEmpty()

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/util/sortedset/MultiSetBackedSortedSet.java
+++ b/warehouse/query-core/src/main/java/datawave/query/util/sortedset/MultiSetBackedSortedSet.java
@@ -5,6 +5,7 @@ import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
+import java.util.NoSuchElementException;
 import java.util.SortedSet;
 import java.util.TreeSet;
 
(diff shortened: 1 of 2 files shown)


apmoriarty (dismissed):

cogross (dismissed):

warehouse/query-core/src/main/java/datawave/query/util/sortedset/MultiSetBackedSortedSet.java
```diff
@@ -141,26 +145,38 @@ public SortedSet<E> tailSet(E fromElement) {
...
+        if (firstSet.isEmpty()) {
+            throw new NoSuchElementException("No elements in input sets");
+        }
         return firstSet.first();
     }
 
     @Override
-    public E last() {
+    public E last() throws NoSuchElementException {
+        if (sets == null || !sets.iterator().hasNext()) {
```
jwomeara: sets.iterator().hasNext() vs sets.isEmpty().  Why not go with the simpler call?

jwomeara (changes requested):

warehouse/query-core/src/main/java/datawave/query/util/sortedset/MultiSetBackedSortedSet.java
MiguelRicardos: Good catch! Updating

MiguelRicardos:
updated to the latest integration, rebased.

hlgp (approved):

apmoriarty (approved):

jack-gitdev (approved):

jwomeara (approved):

PR #1997, opened 2023-06-13
Feature/jexl3

jwomeara:
This updates our jexl dependency to 3.3.

Here's a summary of the major changes made with this update:
- Jexl no longer parents the following nodes with an ASTReference: ASTAndNode, ASTOrNode, ASTFunctionNode, ASTIdentifier, ASTStringLiteral, ASTNumberLiteral, ASTReferenceExpression.  So any logic that assumed this to be true has been updated.
- There are no more custom Datawave JexlNodes.  The following have been removed: ASTDelayedPredicate, ASTEvaluationOnly, ASTUnknownFieldERNode, ASTUnsatisfiableERNode, ASTUnsatisfiableGENode, ASTUnsatisfiableGTNode, ASTUnsatisfiableLENode, ASTUnsatisfiableLTNode, BoundedRange, ExceededtermThresholdJexlNode, ExceededValueThresholdJexlNode, IndexHoleMarkerJexlNode, 
- QueryPropertyMarker has been simplified due to the removal of the custom Datawave Jexl Nodes.  Nodes are marked similarly to how they were before, but the nodes themselves will be Jexl primitives, and not custom Datawave Jexl Nodes.
- ASTFunctionNodes now have 2 children: ASTNamespaceIdentifier and ASTArguments.  ASTNamespaceIdentifier contains the function namespace and name.  ASTArguments contains all of the arguments.
- JexlNodes no longer have an 'image' member.  Each individual node uses it's own parameter to store it's value.  A JexlNodes.getImage method has been added to simulate the prior functionality.
- JexlNodes no longer give direct access to the child array.  Therefore, the JexlNodes.children() method has been removed, since it is impossible to return the original child array, and I wanted to avoid confusion over whether what was returned is a copy of the original array or the array itself.

That is all that I can think of at the moment that has changed, but I'm sure I'm missing something.  Please take a close look at the changes and let me know what you think.


warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
```diff
@@ -208,12 +208,15 @@ public static boolean excludeRegex(Iterable<?> values, String regex) {
      * <li>{@code args[2,...]}: the regexes to use to find matches</li>
      * </ul>
      *
+     * Note: As of jexl3, jexl does not consider array parameters to be the same as varargs, so if you expect a variable number of arguments, you must use
```
apmoriarty, line 215: +1

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IsNotNullPruningVisitor.java
```diff
@@ -121,34 +116,36 @@ public static JexlNode prune(JexlNode node) {
...
-        if (notNullExists && !fields.isEmpty()) {
-            List<JexlNode> next = new ArrayList<>(node.jjtGetNumChildren());
-            JexlNode built;
-            for (JexlNode child : JexlNodes.children(node)) {
-                built = pruneNode(child, fields);
-                if (built != null) {
-                    next.add(built);
-                }
+        // do not process marker nodes
+        if (!QueryPropertyMarkerVisitor.getInstance(node).isAnyType()) {
```
apmoriarty, line 121: This is a good functional change

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
```diff
@@ -1174,17 +1157,16 @@ public void ivarateList(JexlNode rootNode, JexlNode sourceNode, Object data) thr
...
-                        exceededOrEvaluationCache.put(id, fst);
+                        exceededOrEvaluationCache.put(exceededOr.getId(), fst);
                 }
                 
                 // If this is actually negated, then this will be added to excludes. Do not negate in the ivarator
                 listIterBuilder.setNegated(false);
             }
             
-            builder.setField(field);
+            builder.setField(JexlASTHelper.deconstructIdentifier(exceededOr.getField()));
```
apmoriarty, line 1164: Shouldn't the field be deconstructed already? If we accept that dot fields get no support with the move to jexl3, perhaps there should be a visitor that deconstructs all identifiers early in query planning.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IvaratorRequiredVisitor.java
```diff
@@ -27,7 +28,7 @@ public static boolean isIvaratorRequired(JexlNode node) {
     @Override
     public Object visit(ASTAndNode and, Object data) {
         QueryPropertyMarker.Instance instance = QueryPropertyMarker.findInstance(and);
-        if (instance.isAnyTypeOf(ExceededOrThresholdMarkerJexlNode.class, ExceededValueThresholdMarkerJexlNode.class)) {
+        if (instance.isAnyTypeOf(EXCEEDED_OR, EXCEEDED_VALUE)) {
             ivaratorRequired = true;
```
apmoriarty, line 33: I think the class level docs and the method docs for visiting an AND node should be updated. The QueryPropertyMarker class indicates there are three ivarator nodes. If the query planner fails to expand a `_Term_` marker we explicitly fail the query. 
```
# for reference
        EXCEEDED_OR("_List_", true, true),
        EXCEEDED_TERM("_Term_", true, true),
        EXCEEDED_VALUE("_Value_", true, true),
```

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPruningVisitor.java
```diff
@@ -2,61 +2,55 @@
...
-import org.apache.commons.jexl2.parser.ASTTernaryNode;
-import org.apache.commons.jexl2.parser.ASTTrueNode;
-import org.apache.commons.jexl2.parser.ASTUnaryMinusNode;
-import org.apache.commons.jexl2.parser.ASTVar;
-import org.apache.commons.jexl2.parser.ASTWhileStatement;
-import org.apache.commons.jexl2.parser.JexlNode;
-import org.apache.commons.jexl2.parser.JexlNodes;
-import org.apache.commons.jexl2.parser.ParserTreeConstants;
-import org.apache.commons.jexl2.parser.SimpleNode;
+import org.apache.commons.jexl3.parser.ASTAddNode;
```
apmoriarty, line 9: QueryPruningVisitor looks good.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ValidateFilterFunctionVisitor.java
```diff
@@ -1,65 +1,41 @@
...
+import org.apache.commons.jexl3.parser.ASTStringLiteral;
+import org.apache.commons.jexl3.parser.ASTTernaryNode;
+import org.apache.commons.jexl3.parser.ASTTrueNode;
+import org.apache.commons.jexl3.parser.ASTUnaryMinusNode;
+import org.apache.commons.jexl3.parser.ASTVar;
+import org.apache.commons.jexl3.parser.ASTWhileStatement;
+import org.apache.commons.jexl3.parser.JexlNode;
 import org.apache.log4j.Logger;
 
 import java.util.Set;
```
apmoriarty: Good change

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/MinimalReferenceExpressionsVisitor.java
```diff
@@ -199,20 +168,19 @@ private void invalidate(JexlNode node, reason cause) {
...
-    @Override
-    public Object visit(SimpleNode node, Object data) {
-        if (isValid) {
-            node.childrenAccept(this, data);
-        }
-        return data;
-    }
-    
     @Override
     public Object visit(ASTReferenceExpression node, Object data) {
```
apmoriarty, line 173: I was really curious to see how the Jexl3 upgrade would affect this visitor in particular. Looks good.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IsNotNullPruningVisitor.java
jwomeara: You'll notice this in a lot of different visitors, and that's because these checks used to be done for ASTReference.  Now that that node is no longer created, the checks had to move down to the next node in the marker topology - the and node.  Just providing some additional context...

.github/workflows/tests.yml
```diff
@@ -130,6 +130,12 @@ jobs:
   compose-build-and-test:
     runs-on: ubuntu-latest
     steps:
+    - name: Free up some space
+      run: |
```
ivakegg: These are very curious directories to be removing.  Using sudo no less.  Is this not a little dangerous?

warehouse/query-core/src/test/java/datawave/test/ShallowJexlNodeComparator.java
```diff
@@ -32,12 +31,6 @@ public int compare(JexlNode first, JexlNode second) {
         if (first.jjtGetNumChildren() != second.jjtGetNumChildren()) {
             return 1;
         }
-        if (!Objects.equals(first.image, second.image)) {
```
ivakegg, line 35: Why did we drop this code?

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IteratorBuildingVisitorTest.java
```diff
@@ -51,7 +52,8 @@ private IteratorBuildingVisitor getDefault() {
      */
     @Test
     public void visitEqNode_nullValueTest() {
-        ASTEQNode node = (ASTEQNode) JexlNodeFactory.buildEQNode("FIELD", null);
+        ASTEQNode node = (ASTEQNode) JexlNodeFactory.buildEQNode("FIELD", "blah");
```
apmoriarty, line 56: does replacing the null value invalidate these two tests?

warehouse/query-core/src/main/java/datawave/query/index/lookup/Intersection.java
```diff
@@ -143,7 +144,7 @@ public Intersection(Collection<? extends IndexStream> streams, UidIntersector ui
                     case DELAYED_FIELD:
                     case UNKNOWN_FIELD:
                     case EXCEEDED_TERM_THRESHOLD:
-                        this.delayedNodes.add(stream.currentNode());
+                        this.delayedNodes.add(JexlNodes.wrap(stream.currentNode()));
```
ivakegg, line 148: So now we are "wrapping" the node here?  Can you tell me why?

warehouse/query-core/src/main/java/datawave/query/index/lookup/ScannerStream.java
```diff
@@ -66,7 +65,7 @@ public static ScannerStream variable(BaseIndexStream itr, JexlNode currNode) {
     
     // exceeded value threshold, so we can evaluate with data but may need special handling
     public static ScannerStream exceededValueThreshold(Iterator<Tuple2<String,IndexInfo>> itr, JexlNode currNode) {
-        JexlNode resultNode = JexlNodeFactory.wrap(currNode);
+        JexlNode resultNode = currNode;
```
ivakegg, line 70: And we are no longer wrapping the node here? Can you tell me why?

warehouse/query-core/src/main/java/datawave/query/index/lookup/Union.java
```diff
@@ -85,7 +86,7 @@ public Union(Iterable<? extends IndexStream> streams) {
                 case UNKNOWN_FIELD:
                 case EXCEEDED_TERM_THRESHOLD:
                     // these nodes need to be persisted via the set of delayedNodes
-                    delayedNodes.add(stream.currentNode());
+                    delayedNodes.add(JexlNodes.wrap(stream.currentNode()));
```
ivakegg, line 89: Same comment as for the Intersection class.

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
```diff
@@ -490,15 +484,6 @@ public Object visit(ASTAndNode node, Object data) {
         }
     }
     
-    /** {@inheritDoc} */
-    public Object visit(ASTSizeMethod node, Object data) {
```
ivakegg, line 531: Should this be changed to ASTSizeFunction instead of dropping it?

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/QueryFieldsVisitorTest.java
```diff
@@ -34,10 +34,6 @@ public void testEQ() throws ParseException {
         // identifiers
         query = "$12 == 'bar'";
         test(query, Collections.singleton("12"));
-        
-        // grouping context
-        query = "FOO.12 == 'bar'";
-        test(query, Collections.singleton("FOO"));
```
apmoriarty: Noting our previous conversation, grouping notation on query fields is explicitly NOT supported by datawave in Jexl3. For a query with grouping notation fields we may want to either fail the query or rewrite the fields.

apmoriarty (approved):
Overall this looks really good. I'm excited to work in a world with fewer reference nodes. Several questions popped up but none of them would block a merge.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
jwomeara: The field is not deconstructed in the exceeded or node actually.  I tried deconstructing the field earlier in the process and that actually broke the logic that adds the event values for that field to the jexl context.  For some reason that I didn't uncover, the logic that pulls fields out of the event and into the jexl context relies on the field not already being deconstructed.  Good question though - this was one of the places where i had to put a little more thought into what should be done.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
ivakegg: interesting.  Good safety tip.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IvaratorRequiredVisitor.java
jwomeara: Updated where?  In IvaratorRequiredVisitor, or in QueryPropertyMarker?  I'm not understanding based on the context of this class.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IteratorBuildingVisitorTest.java
jwomeara: The problem is that JexlNodeFactory.buildEQNode eventually calls JexlNodes.setLiteral.  If that is called with a null value, it will throw an exception for a precondition 'not null' error.  To get around that, I first create a valid EQ node with a throwaway value, then on the next line I replace the string literal child with one that has no value.  So, the tests should be completely equivalent - just the setup is different.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/QueryFieldsVisitorTest.java
jwomeara: Just clarifying that it's not supported for jexl2 or jexl3.  jexl2 would fail silently and munge the field name.  With the jexl3 update, this actually WILL fail if you attempt to use it because our code won't be able to find the identifier.  That's why it had to be removed from the test.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IteratorBuildingVisitorTest.java
apmoriarty: That was the context I needed, perfect.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IvaratorRequiredVisitor.java
jwomeara: I updated what I could find in QueryPropertyMarker and QueryPropertyMarkerVisitor.  Let me know if there's something else specific you think should be udpated.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IsNotNullPruningVisitor.java
```diff
@@ -586,11 +559,6 @@ public Object visit(ASTMethodNode node, Object data) {
         return data;
     }
     
-    @Override
-    public Object visit(ASTSizeMethod node, Object data) {
```
ivakegg, line 591: ASTSizeFunction

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
```diff
@@ -633,12 +644,6 @@ public Object visit(ASTMethodNode node, Object data) {
         return null;
     }
     
-    @Override
-    public Object visit(ASTSizeMethod node, Object data) {
```
ivakegg, line 637: ASTSizeFunction

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlStringBuildingVisitor.java
```diff
@@ -365,12 +366,26 @@ public Object visit(ASTNotNode node, Object data) {
...
         
-        node.childrenAccept(this, sb);
+        return sb;
+    }
+    
+    @Override
+    protected Object visit(ASTIdentifierAccess node, Object data) {
+        StringBuilder sb = (StringBuilder) data;
+        
+        sb.append(".");
```
ivakegg, line 391: should this only be if sb.length() > 0

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlStringBuildingVisitor.java
```diff
@@ -550,28 +507,28 @@ public Object visit(ASTJexlScript node, Object data) {
...
+                sb.append(" + ");
+            }
             node.jjtGetChild(i).jjtAccept(this, sb);
         }
         
         return sb;
     }
     
     @Override
-    public Object visit(ASTSizeMethod node, Object data) {
```
ivakegg, line 573: ASTSizeFunction?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnexecutableNodesVisitor.java
```diff
@@ -417,11 +392,6 @@ public Object visit(ASTMethodNode node, Object data) {
         return node;
     }
     
-    @Override
-    public Object visit(ASTSizeMethod node, Object data) {
```
ivakegg, line 422: ASTSizeFunction?

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/ValidComparisonVisitorTest.java
```diff
@@ -78,7 +78,7 @@ public void testInvalidGE() throws ParseException {
         ValidComparisonVisitor.check(script);
     }
     
-    @Test(expected = DatawaveFatalQueryException.class)
+    @Test(expected = ParseException.class)
```
ivakegg, line 82: Do we not want a DatawaveFatalQueryException ?

.github/workflows/tests.yml
jwomeara: Followed the recommendations here which seems to have helped some people.  
https://github.com/actions/runner-images/issues/2840#issuecomment-790492173

warehouse/query-core/src/main/java/datawave/query/index/lookup/Intersection.java
jwomeara: The wrapping just moved to a different line of code.  If you leave this unwrapped, when it is added back to the tree (unwrapped) it marks the entire node that it is added to which unintentionally will mark addition nodes as being delayed.

warehouse/query-core/src/test/java/datawave/test/ShallowJexlNodeComparator.java
jwomeara: The JexlNode value is an internal parameter used for evaluation that we should not be comparing from one tree to another.  In jexl3 there is state information saved there which specifies what parser was used to create the jexl tree.  Our rebuilding visitor doesn't copy this information so it was coming up as non-equal even though for our purposes they would have been functionally equivalent.

warehouse/query-core/src/main/java/datawave/query/index/lookup/ScannerStream.java
jwomeara: As I said in the comment above, the wrapping just moved to a more common location.

warehouse/query-core/src/main/java/datawave/query/index/lookup/Union.java
jwomeara: Same comment as above.  Also, keep in mind that the original code would have identified the reference node as the marked node, and the reference was the parent of the reference expression, so we didn't need to wrap anything with the existing code.  With the new code, since there are no reference nodes, and since AND nodes are now identified as the marked node, the AND node will need to be wrapped so that it doesn't mark the node it is added to if the tree is flattened.

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
jwomeara: I don't think so.  The parent implementation for ASTSizeFunction should be fine.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IsNotNullPruningVisitor.java
jwomeara: There could be a null within a size function, so I think we should leave this as-is since the parent implementation for ASTSizeFunction will cover that case.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
jwomeara: Would we not want to visit the children of a size function node?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlStringBuildingVisitor.java
jwomeara: No.  If someone creates a query like this ".something == 'blah'", I still want to print '.something' even though there is no identifier tied to it.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnexecutableNodesVisitor.java
jwomeara: yes!  I will add it.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlStringBuildingVisitor.java
jwomeara: yes!  i will add it.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnexecutableNodesVisitor.java
jwomeara: Do we not want to visit the children of a size function?

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/ValidComparisonVisitorTest.java
jwomeara: In the old code it failed because datawave didn't allow it.  In the new code it fails because JEXL doesn't allow it.

.github/workflows/tests.yml
jwomeara: Given that the compose test still isn't completing successfully I need to continue looking at this.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlStringBuildingVisitor.java
jwomeara: pushed.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExecutableDeterminationVisitor.java
```diff
@@ -714,51 +711,7 @@ public Object visit(ASTNENode node, Object data) {
...
-        } else if (instance.isType(IndexHoleMarkerJexlNode.class)) {
-            state = STATE.NON_EXECUTABLE;
-            if (output != null) {
-                output.writeLine(data, node, "( index hole )", state, true);
-            }
-        } else {
-            state = allOrNone(node, data);
-        }
-        return state;
+        return allOrNone(node, data);
```
ivakegg, line 717: I thought ASTReference went away around expressions and AND statement?  Should this visit function simply be dropped?

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -207,30 +204,28 @@ private static ASTJexlScript parseQueryWithBackslashes(String query, Parser pars
         while (!workingStack.isEmpty()) {
             JexlNode node = workingStack.pop();
             
-            if (node.image != null) {
-                int numToReplace = StringUtils.countMatches(node.image, placeholder);
+            if (node instanceof ASTStringLiteral) {
```
ivakegg, line 209: We used to parse the image of all nodes.  Are we we sure ASTStringLiteral is the only one we need to handle?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
ivakegg: I guess I have to respond with a question.  Can one of the children of the size function be an expression like maybe another function?  I am thinking of something like size(f:func(bla))

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnexecutableNodesVisitor.java
ivakegg: I guess I have to respond with a question. Can one of the children of the size function be an expression like maybe another function? I am thinking of something like size(f:func(bla))

warehouse/query-core/src/test/java/datawave/test/ShallowJexlNodeComparator.java
ivakegg: Well then this comparator really does not do the comparison we want anymore.  We need to compare the value in the nodes somehow.

ivakegg: Perhaps something like Objects.equals(JexlNodes.getImage(first), JexlNodes.getImage(second)) ?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
ivakegg: Well, actually in the context of an IteratorBuildingVisitor, it would have to be something like:
size(((_Value_ = true) && (x == y))).

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryFieldsVisitor.java
```diff
@@ -318,11 +301,6 @@ public Object visit(ASTMethodNode node, Object data) {
         return data;
     }
     
-    @Override
-    public Object visit(ASTSizeMethod node, Object data) {
-        return data;
```
ivakegg, line 323: We may need to process the size argument.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryModelVisitor.java
```diff
@@ -153,11 +154,6 @@ public Object visit(ASTMethodNode node, Object data) {
         return node.jjtAccept(this.simpleQueryModelVisitor, data);
     }
     
-    @Override
-    public Object visit(ASTSizeMethod node, Object data) {
```
ivakegg, line 157: I think we still need to process a size function

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryPruningVisitor.java
```diff
@@ -525,11 +486,6 @@ public Object visit(ASTMethodNode node, Object data) {
         return TruthState.UNKNOWN;
     }
     
-    @Override
-    public Object visit(ASTSizeMethod node, Object data) {
```
ivakegg, line 529: replace with ASTSizeFunction

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnexecutableNodesVisitor.java
ivakegg: In the context of this function I guess it does not make sense as we were not doing it before.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
ivakegg: But now that I look more closely we were not doing anything with the size function before, so nevermind.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RegexIndexExpansionVisitor.java
```diff
@@ -266,14 +263,7 @@ public Object visit(ASTERNode node, Object data) {
                     log.info(logLine);
                 }
                 
-                // If we've *never* seen this field, we want to denote the difference against it not being indexed
-                if (fieldName.equals(Constants.ANY_FIELD)) {
-                    return node;
-                } else if (!allFields.contains(fieldName)) {
-                    return RebuildingVisitor.copyInto(node, ASTUnknownFieldERNode.create());
```
ivakegg, line 276: I am a little scared we are dropping something.  Do we have a replacement for the ASTUnknownFieldERNode and ASTUnsatisfiableERNode ?

warehouse/query-core/src/test/java/datawave/test/ShallowJexlNodeComparator.java
jwomeara: Well, w.r.t. `value` the comparator really wasn't doing the comparison you expect anyway.  W.R.T. image, I can add an equivalent comparison.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExecutableDeterminationVisitor.java
jwomeara: i think we still need this.  if you call a method on something a reference is used.  For example, for this query `filter:includeRegex(NAM, 'MICHAEL').size() == 1`, the filter function will be parsed into an ASTFunctionNode and the size method will be parsed into an ASTMethodNode, and they will both be parented by the same ASTReference node.

If we remove this method, we won't perform an executability check on these kinds of nodes.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
jwomeara: The image parameter no longer exists.  

For context, there are 3 kinds of nodes that contain a literal now: ASTRegexLiteral (new, and unused by us), ASTStringLiteral, ASTNumberLiteral. 

This check was added by me a while ago to fix the right hand (String literal) side of EQ, ER, or NR nodes whose image contain backslashes so that we can escape them properly.  At the time, keying off of the `image` parameter was more convenient than checking to see whether the node was an ASTStringLiteral.  I can say pretty confidently that I am not interested in modifying anything that is not an ASTStringLiteral with this segment of code.  If you can think of something else that needs to be included here, let me know, but nothing comes to my mind.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RegexIndexExpansionVisitor.java
jwomeara: In the old code we create ASTUnsatisfiableERNode instances, but nowhere in the code will you find an example where we check to see if a node is an ASTUnsatisfiableERNode.  So, it was effectively unused.  

The existence of ASTUnknownFieldERNode triggered special logic in one place - RangeStream.  If you check RangeStream, you will see that I moved the check being performed in the RegexIndexExpansionVisitor to that visitor in order to trigger the creation of an unknown field scanner stream.  This was the only place where this type was checked for, so this should be equivalent.    

Also, keep in mind that these custom datawave JexlNode types do NOT survive if you print the jexl tree as a string, and parse that string back into a tree again - so this kind of logic is fairly brittle.  Jexl doesn't know what these nodes are, so to me it made sense to move away from custom types and stick with the primitives.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
ivakegg: I see now.  np

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlStringBuildingVisitor.java
apmoriarty: If  no test exists for that use case then it should get added

.github/workflows/tests.yml
jwomeara: working now.

ivakegg (approved):

PR #1999, opened 2023-06-15
Running sort import plugin on repo

avgAGB:


jwomeara (commented):
Do we want to consider using a plugin like this to keep the imports sorted?

https://github.com/revelc/impsort-maven-plugin

avgAGB:
> Do we want to consider using a plugin like this to keep the imports sorted?
> 
> https://github.com/revelc/impsort-maven-plugin

I recently merged in a ticket requesting that the plugin be kept separate from the default build - @apmoriarty do you think we should keep this out of the default build flow or should it be always running?

jwomeara:
> > Do we want to consider using a plugin like this to keep the imports sorted?
> > https://github.com/revelc/impsort-maven-plugin
> 
> I recently merged in a ticket requesting that the plugin be kept separate from the default build - @apmoriarty do you think we should keep this out of the default build flow or should it be always running?

If I'm allowed to have a say I would advocate for giving it the same precedence as the maven formatter plugin.

apmoriarty:
> > > Do we want to consider using a plugin like this to keep the imports sorted?
> > > https://github.com/revelc/impsort-maven-plugin
> > 
> > 
> > I recently merged in a ticket requesting that the plugin be kept separate from the default build - @apmoriarty do you think we should keep this out of the default build flow or should it be always running?
> 
> If I'm allowed to have a say I would advocate for giving it the same precedence as the maven formatter plugin.

I'm all for running this at the same time as the formatter plugin. 

For context the original plan was to add the impsort plugin behind a profile flag in one merge request and then follow up with the massive import format in a secondary merge request (this one).

avgAGB:
> > > > Do we want to consider using a plugin like this to keep the imports sorted?
> > > > https://github.com/revelc/impsort-maven-plugin
> > > 
> > > 
> > > I recently merged in a ticket requesting that the plugin be kept separate from the default build - @apmoriarty do you think we should keep this out of the default build flow or should it be always running?
> > 
> > 
> > If I'm allowed to have a say I would advocate for giving it the same precedence as the maven formatter plugin.
> 
> I'm all for running this at the same time as the formatter plugin.
> 
> For context the original plan was to add the impsort plugin behind a profile flag in one merge request and then follow up with the massive import format in a secondary merge request (this one).

cool, I'll tweak this in the next commit

apmoriarty (approved):
LGTM. Would like to see a slightly larger consensus on this merge request.

ivakegg:
I am good with this as long as we are all willing to take on the pain of updating their merge requests.  Just like the formatting changes which have already created a lot of churn, we will need to go through it again.  Again, I am ok with that but just want to make sure we have a quorum.

keith-ratcliffe:
> ... just want to make sure we have a quorum.

Assuming it doesn't add an inordinate amount of time to the build, I'm good with it as well

cogross (approved):
Looks fine to me.

mineralntl (approved):

PR #2003, opened 2023-06-21
Only configure excerpt query options when necessary

apmoriarty:
Change Summary
- QueryOptions for excerpts are only added to iterator settings if term frequencies are required
- optionally prune QueryOptions on a per-tablet basis via the VisitorFunction. Initial implementation only supports excerpt options.
- ShardQueryConfigTest for new additions now has a short summary for adding and removing configs


warehouse/query-core/src/test/java/datawave/query/config/ShardQueryConfigurationTest.java
```diff
@@ -415,12 +417,14 @@ public void testDeepCopyConstructor() {
...
         // assert seeks
-        Assert.assertEquals(12, other.getFiFieldSeek());
-        Assert.assertEquals(13, other.getFiNextSeek());
-        Assert.assertEquals(14, other.getEventFieldSeek());
-        Assert.assertEquals(15, other.getEventNextSeek());
+        Assert.assertEquals(12, config.getFiFieldSeek());
+        Assert.assertEquals(13, config.getFiNextSeek());
+        Assert.assertEquals(14, config.getEventFieldSeek());
+        Assert.assertEquals(15, config.getEventNextSeek());
         Assert.assertEquals(16, other.getTfFieldSeek());
```
avgAGB: should the tffieldseek switch to config as well, or should we keep that the same?

apmoriarty: yes it should, good catch

avgAGB (dismissed):

cogross (approved):

MiguelRicardos (approved):

jwomeara (approved):

PR #2004, opened 2023-06-21
Updated first()/last() methods to not rely on a boolean, clean up 

MiguelRicardos:


apmoriarty (approved):
refactor looks good to me, someone with more experience in the ivarators should take a look... @ivakegg or @jwomeara ?

mineralntl (approved):

jwomeara (approved):
Changes look good.  Not sure the conditionals in toString and compare were needed as that seems more like a style preference, but... ok.

MiguelRicardos:
updated to the latest integration, rebased.

PR #2005, opened 2023-06-22
Refactor field parsing in EventDataQueryExpressionFilter

apmoriarty:
extract field parsing from the EventDataQueryExpressionFilter's apply method

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/predicate/EventDataQueryExpressionFilter.java
+++ b/warehouse/query-core/src/main/java/datawave/query/predicate/EventDataQueryExpressionFilter.java
@@ -3,12 +3,10 @@ package datawave.query.predicate;
 import com.google.common.collect.Maps;
 import datawave.query.data.parsers.DatawaveKey;
 import datawave.query.attributes.AttributeFactory;
-import datawave.query.attributes.Document;
 import datawave.query.jexl.JexlASTHelper;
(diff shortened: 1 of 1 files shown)


cogross (approved):

MiguelRicardos (approved):

warehouse/query-core/src/main/java/datawave/query/predicate/EventDataQueryExpressionFilter.java
```diff
@@ -103,15 +101,26 @@ protected boolean apply(Key key, boolean update) {
             throw new RuntimeException("The EventDataQueryExpressionFilter was not initialized");
         }
 
-        final DatawaveKey datawaveKey = new DatawaveKey(key);
-        final String fieldName = JexlASTHelper.deconstructIdentifier(datawaveKey.getFieldName(), false);
+        final String fieldName = getFieldNameFromKey(key);
```
jwomeara, line 104: Is this used in more than one location?  Otherwise, why refactor?

apmoriarty: The idea was to use a key parser and avoid the datawave key, but exploratory refactors in that direction did not prove useful.

apmoriarty:
The goal here was to replace the datawave key with a dedicated key parser, but there was no immediate path forward.

PR #2012, opened 2023-06-28
Add identifiers used for content retrieval to lookupContentUUID response

drewfarris:
The `/lookupContentUUID` call retrieves the content for an object by scanning for the 'd' column in accumulo. When making this call, users can either provide an internal id (e.g., shard/datatype/uid) or a uuid to specify the object of interest. 

- If a uuid is specified, it will be translated into an internal id by performing a `/lookupUUID` call, and then the internal ids will be used to retrieve the content from the 'd' column in the shard table.
- If an internal id is specified datawave scan scan the shard table directly and will not need to perform the initial uuid lookup. 

It is possible to specify a batch of internal ids or uuids to look up the content for multiple objects at once using the `POST` variant of the `/lookupContentUUID` call.

The `lookupContentUUID` response contains the internal id for each content object retrieved. Unfortunately any linkage to the original uuid is not maintained. This makes it difficult for folks using the batch call to correlate the returned content with one of the ids specified in their original request.

This PR addresses that issue, using HIT_TERM to maintain the uuid in the results of the `/lookupUUID` call, and then injecting these identifiers into the content lookup process.

A side-effect of this PR is that end users that are querying by internal id can also add arbitrary labels to batch `/lookupContentUUID` responses if they add internal identifiers to the internal id's they request using the form `DOCUMENT:shard/datatype/uid!customlabel`

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/tables/content/ContentQueryTable.java
+++ b/warehouse/query-core/src/main/java/datawave/query/tables/content/ContentQueryTable.java
@@ -100,7 +100,7 @@ public class ContentQueryTable extends BaseQueryLogic<Entry<Key,Value>> {
             this.viewName = p.getParameterValue();
         }
         
-        // Decide whether or not to include the content of child events
+        // Decide whether to include the content of child events
         String end;
         p = settings.findParameter(QueryParameters.CONTENT_VIEW_ALL);
         if ((null != p) && (null != p.getParameterValue()) && StringUtils.isNotBlank(p.getParameterValue())) {
@@ -180,12 +180,17 @@ public class ContentQueryTable extends BaseQueryLogic<Entry<Key,Value>> {
             if (!term.isEmpty()) {
                 // Get the next value
                 int fieldSeparation = term.indexOf(':');
-                final String value;
+                final String valueIdentifier;
                 if (fieldSeparation > 0) {
-                    value = term.substring(fieldSeparation + 1);
+                    valueIdentifier = term.substring(fieldSeparation + 1);
                 } else {
-                    value = term;
+                    valueIdentifier = term;
                 }
+                
+                // Remove the identifier if present - we won't it here, but will extract them from the query
+                // later in the ContentQueryTransformer
+                int idSeparation = valueIdentifier.indexOf("!");
+                final String value = idSeparation > 0 ? valueIdentifier.substring(0, idSeparation) : valueIdentifier;
                 
                 // Validate the value
                 final String[] parts = value.split("/");
--- a/warehouse/query-core/src/main/java/datawave/query/tables/content/ContentQueryTable.java
+++ b/warehouse/query-core/src/main/java/datawave/query/tables/content/ContentQueryTable.java
@@ -187,7 +187,7 @@ public class ContentQueryTable extends BaseQueryLogic<Entry<Key,Value>> {
                     valueIdentifier = term;
                 }
                 
-                // Remove the identifier if present - we won't it here, but will extract them from the query
+                // Remove the identifier if present - we won't use it here, but will extract them from the query
                 // later in the ContentQueryTransformer
                 int idSeparation = valueIdentifier.indexOf("!");
                 final String value = idSeparation > 0 ? valueIdentifier.substring(0, idSeparation) : valueIdentifier;
--- a/warehouse/query-core/src/main/java/datawave/query/transformer/ContentQueryTransformer.java
+++ b/warehouse/query-core/src/main/java/datawave/query/transformer/ContentQueryTransformer.java
@@ -15,20 +15,92 @@ import datawave.webservice.result.EventQueryResponseBase;
 import org.apache.accumulo.core.data.Key;
 import org.apache.accumulo.core.data.Value;
 import org.apache.accumulo.core.security.Authorizations;
+import org.apache.log4j.Logger;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Map.Entry;
 
 public class ContentQueryTransformer extends BaseQueryLogicTransformer<Entry<Key,Value>,EventBase> {
     
+    private static final Logger log = Logger.getLogger(ContentQueryTransformer.class);
+    
     protected final Authorizations auths;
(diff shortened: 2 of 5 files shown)


drewfarris:
I can't link any of the changes made here back to the changes I had to make in order to get the unit tests to pass, so I'm a little worried I have an environment issue. Keeping a close eye on the CI/CD results.

warehouse/query-core/src/main/java/datawave/query/transformer/ContentQueryTransformer.java
```diff
@@ -15,20 +15,92 @@
...
+    /**
+     * Extract optional identifiers from the query. Expected query format is:
+     * <p>
+     * 'DOCUMENT:shard/datatype/uid!optionalIdentifier1 DOCUMENT:shard/datatype/uid!optionalIdentifier2 .. DOCUMENT:shard/datatype/uid!optionalIdentifier3'
+     * <p>
+     * The identifiers are not required, so this will parse 'DOCUMENT:shard/datatype/uid' as well.
+     *
+     * @param querySettings
+     *            the current query for which we are transforming results.
+     * @return a map of shard/datatye/uid mapped to their corresponding identifiers.
```
apmoriarty, line 50: This is great documentation, it should also go in the query_help doc.

apmoriarty (changes requested):
LGTM. A little concerning that the mocked unit test is failing, I'll spend some time tomorrow trying to sort through that.

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/transformer/ContentQueryTransformer.java
drewfarris: Good point. I've been meaning to chase down the location of that.

warehouse/query-core/src/main/java/datawave/query/tables/content/ContentQueryTable.java
```diff
@@ -180,13 +180,18 @@ private Collection<Range> createRanges(final Query settings, final String endKey
...
+                final String valueIdentifier;
                 if (fieldSeparation > 0) {
-                    value = term.substring(fieldSeparation + 1);
+                    valueIdentifier = term.substring(fieldSeparation + 1);
                 } else {
-                    value = term;
+                    valueIdentifier = term;
                 }
                 
+                // Remove the identifier if present - we won't it here, but will extract them from the query
```
jwomeara: we won't it here?

web-services/query/src/test/java/datawave/webservice/query/runner/ExtendedQueryExecutorBeanTest.java
```diff
@@ -394,7 +394,7 @@ public void testAdminCancel_LookupAccumuloQuery() throws Exception {
         expect(this.queryLogicFactory.getQueryLogic("ql1", principal)).andReturn((QueryLogic) this.queryLogic1);
         expect(this.queryLogic1.getConnectionPriority()).andReturn(Priority.NORMAL);
         expect(this.queryLogic1.getCollectQueryMetrics()).andReturn(false);
-        expect(this.queryLogic1.isLongRunningQuery()).andReturn(false);
+        // expect(this.queryLogic1.isLongRunningQuery()).andReturn(false);
```
jwomeara: what happened here?  same question for the subsequent commented out expect statements.

web-services/query/src/test/java/datawave/webservice/query/runner/QueryExecutorBeanTest.java
```diff
@@ -286,7 +286,7 @@ private void defineTestRunner(QueryImpl q, MultivaluedMap p) throws Exception {
         EasyMock.expect(logic.containsDNWithAccess(dnList)).andReturn(true);
         EasyMock.expect(logic.getMaxPageSize()).andReturn(0);
         EasyMock.expect(logic.getCollectQueryMetrics()).andReturn(Boolean.FALSE);
-        EasyMock.expect(logic.isLongRunningQuery()).andReturn(false);
+        // EasyMock.expect(logic.isLongRunningQuery()).andReturn(false);
```
jwomeara: what happened here?

jwomeara (changes requested):
looks good for the most part - just a few questions/comments.

warehouse/query-core/src/main/java/datawave/query/tables/content/ContentQueryTable.java
drewfarris: Fixed.

web-services/query/src/test/java/datawave/webservice/query/runner/ExtendedQueryExecutorBeanTest.java
drewfarris: It looks like this was something introduced in the `release/version3.42` branch. I rebased this PR to `release/version3.41` and these unit test modifications were no longer needed. Also, after backing out all changes from `release/version3.42` these unit test modifications were still required, so I'm going to look more closely at the `release/version3.42` branch to see what is happening there separately from this PR. 

Bottom line, I suspect these unit tests are failing in `release/version3.42` _in its current state_ without these changes.

web-services/query/src/test/java/datawave/webservice/query/runner/QueryExecutorBeanTest.java
drewfarris: Same comment as: https://github.com/NationalSecurityAgency/datawave/pull/2012#discussion_r1247902509

web-services/query/src/test/java/datawave/webservice/query/runner/ExtendedQueryExecutorBeanTest.java
drewfarris: Confirmed `release/version3.42` was indeed bad due to a bad merge. Addressed this in https://github.com/NationalSecurityAgency/datawave/commit/cc0b9ea18895792118e90df31f6da38af6f538fe

jwomeara (approved):

ivakegg (approved):

drewfarris:
Replaced this PR with #2020 because this originated from a personal fork.

PR #2014, opened 2023-06-29
Fix view name3.41

MiguelRicardos:


drewfarris (approved):
`content.view.name` never worked before because it did not take into account that the column qualifier contains more than just the view name (it contains the datatype and uid as well in `null` delimited form). 

This change ultimately works because the `content.view.name` is now handled as a regex find (e.g., substring match) instead of a regex match (e.g., a whole string match). This has the side effect that `lookupContentUUID` will return view names where the value of the content.view.name parameter partially matches a view name.

apmoriarty (approved):

Issue #2008, opened 2023-06-26
Update model lenient mechanism to be more obvious


ivakegg:
The current "lenient" mechanism for a model field is done by simply mapping the model field to an internal field name of "lenient".  This is a little hokey and should be done more explicitly.  Perhaps we should create a mechanism by which multiple attributes can be applied to a model field using a list in the columnQualifier or in the Value.

ivakegg:
After discussions with the customer, it appears that lenient needs to be the default, and strict will be the special case.

PR #2019, opened 2023-07-04
Feature/datawave 2008

ivakegg:
https://github.com/NationalSecurityAgency/datawave/issues/2008

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
@@ -46,7 +46,6 @@ import org.apache.commons.jexl2.parser.ASTStringLiteral;
 import org.apache.commons.jexl2.parser.ASTTrueNode;
 import org.apache.commons.jexl2.parser.JexlNode;
 import org.apache.commons.jexl2.parser.JexlNodes;
-import org.apache.commons.jexl2.parser.LenientExpression;
 import org.apache.commons.jexl2.parser.ParserTreeConstants;
 
 import java.math.BigDecimal;
@@ -263,73 +262,6 @@ public class JexlNodeFactory {
                     return JexlNodes.children(junction, children.toArray(new JexlNode[0]));
             }
         }
-    }
-
-    /**
-     * This method is currently only used from the QueryModelVisitor to expand a binary node into multiple pairs. If one side contains the lenient marker, then
-     * that will be extended around the binary node.
-     *
-     * @param containerType
-     * @param node
-     * @param pairs
-     * @return The container of binary nodes based on the specified node type
-     */
-    public static JexlNode createNodeTreeFromPairs(ContainerType containerType, JexlNode node, Set<List<JexlNode>> pairs) {
-        if (1 == pairs.size()) {
-            List<JexlNode> pair = pairs.iterator().next();
-
-            if (2 != pair.size()) {
-                throw new UnsupportedOperationException("Cannot construct a node from a non-binary pair: " + pair);
-            }
-
-            return buildUntypedBinaryNode(node, pair.get(0), pair.get(1));
-        }
-
-        JexlNode parentNode = (containerType.equals(ContainerType.OR_NODE) ? new ASTOrNode(ParserTreeConstants.JJTORNODE)
-                        : new ASTAndNode(ParserTreeConstants.JJTANDNODE));
-
-        int i = 0;
-        JexlNodes.ensureCapacity(parentNode, pairs.size());
-        for (List<JexlNode> pair : pairs) {
-            if (2 != pair.size()) {
-                throw new UnsupportedOperationException("Cannot construct a node from a non-binary pair: " + pair);
-            }
-
-            JexlNode leftNode = pair.get(0);
-            JexlNode rightNode = pair.get(1);
-            QueryPropertyMarker.Instance leftType = QueryPropertyMarker.findInstance(leftNode);
-            QueryPropertyMarker.Instance rightType = QueryPropertyMarker.findInstance(rightNode);
-            boolean lenient = false;
-            if (leftType.isType(LenientExpression.class)) {
-                lenient = true;
-                leftNode = leftType.getSource();
-            }
-            if (rightType.isType(LenientExpression.class)) {
-                lenient = true;
-                rightNode = rightType.getSource();
-            }
-            JexlNode child = buildUntypedBinaryNode(node, leftNode, rightNode);
-            if (lenient) {
-                child = LenientExpression.create(child);
-            }
-            parentNode.jjtAddChild(child, i);
-
-            // We want to override the default parent that would be set by
-            // buildUntypedNode because we're attaching this to an OR
-            child.jjtSetParent(parentNode);
-
-            i++;
-        }
-
-        // If we have more than one element in the new node, wrap it in
-        // parentheses
-        JexlNode wrapped = wrap(parentNode);
-
-        // Set the parent pointer
-        wrapped.jjtSetParent(node.jjtGetParent());
-
-        return wrapped;
-
     }
 
     public static JexlNode createNodeTreeFromFieldNames(ContainerType containerType, JexlNode node, Object literal, Collection<String> fieldNames) {
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
@@ -39,6 +39,7 @@ import org.apache.commons.jexl2.parser.JexlNode;
 import org.apache.commons.jexl2.parser.JexlNodes;
 import org.apache.commons.jexl2.parser.LenientExpression;
 import org.apache.commons.jexl2.parser.ParserTreeConstants;
+import org.apache.commons.jexl2.parser.StrictExpression;
 import org.apache.log4j.Logger;
 
 import java.text.MessageFormat;
@@ -152,10 +153,15 @@ public class ExpandMultiNormalizedTerms extends RebuildingVisitor {
             return node;
         }
 
-        // if we found a lenient expression, then pass it through
+        // if we found a strict or lenient expression, then pass it through
+        boolean strict = false;
         boolean lenient = false;
         JexlNode originalSource = null;
-        if (marker.isType(LenientExpression.class)) {
+        if (marker.isType(StrictExpression.class)) {
+            strict = true;
+            data = marker;
+            originalSource = copy(marker.getSource());
+        } else if (marker.isType(LenientExpression.class)) {
             lenient = true;
             data = marker;
             originalSource = copy(marker.getSource());
@@ -169,13 +175,13 @@ public class ExpandMultiNormalizedTerms extends RebuildingVisitor {
             returnNode = (JexlNode) (super.visit(node, data));
         }
 
-        // if we had a lenient marker, then unwrap it and handle all dropped
-        if (lenient) {
+        // if we have a strict or lenient marker, then unwrap it and handle all dropped
+        if (strict || lenient) {
             QueryPropertyMarker.Instance newMarker = QueryPropertyMarker.findInstance(returnNode);
             returnNode = newMarker.getSource();
 
-            // now we need to check if everything was dropped in which case we return the original node source eval only
-            if (isAllDropped(returnNode)) {
+            // if everything was dropped and we are strict, then wrap the original node with eval only
+            if (strict && isAllDropped(returnNode)) {
                 returnNode = ASTEvaluationOnly.create(originalSource);
             }
         }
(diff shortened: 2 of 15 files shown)


warehouse/query-core/src/test/java/datawave/query/AnyFieldQueryTest.java
```diff
@@ -54,6 +55,7 @@ public static void filterSetup() throws Exception {
 
         accumuloSetup.setData(FileType.CSV, dataType);
         client = accumuloSetup.loadTables(log);
+        Logger.getLogger(DefaultQueryPlanner.class).setLevel(Level.DEBUG);
```
apmoriarty: leftover from testing?

ivakegg: yikes, it is.  I will remove it.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTermsTest.java
```diff
@@ -614,7 +649,7 @@ public void testFailedRegexNormalizersAndNRNodes() throws ParseException {
 
         // in this case the numeric normalization fails, so we need to mark as evaluation only
```
hlgp: should the comment be updated?

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
```diff
@@ -268,7 +269,7 @@ public static JexlNode createNodeFromValuesForField(String field, IndexLookupMap
     }
 
     /**
-     * This method is currently only used from the QueryModelVisitor to expand a binary node into multiple pairs. If one side contains the lenient marker, then
+     * This method is currently only used from the QueryModelVisitor to expand a binary node into multiple pairs. If one side contains the strict marker, then
```
jwomeara: If this method is specific to QueryModelVisitor why is it here instead of in QueryModelVisitor?  (is there a better answer than 'because it's always been here'?)

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
```diff
@@ -154,9 +155,15 @@ public Object visit(ASTReference node, Object data) {
...
+        // if we found a strict or lenient expression, then pass it through
+        boolean strict = false;
         boolean lenient = false;
         JexlNode originalSource = null;
+        if (marker.isType(StrictExpression.class)) {
+            strict = true;
+            data = marker;
+            originalSource = copy(marker.getSource());
+        }
         if (marker.isType(LenientExpression.class)) {
```
jwomeara: should this be `else if`?  If not, can you add a test where lenient and strict are both true?

warehouse/query-core/src/test/java/datawave/query/AnyFieldQueryTest.java
```diff
@@ -13,6 +13,7 @@
 import java.util.Collections;
 
 import org.apache.accumulo.core.data.Key;
+import org.apache.log4j.Level;
```
jwomeara: Unused.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/QueryModelVisitorTest.java
jwomeara, line 1: in `testLenientLiteralQuery` why is the string literal being marked, and not the full equals expression?

web-services/client/src/main/java/datawave/webservice/model/Model.java
jwomeara, line 1: lots of unused imports in here.

web-services/client/src/main/java/datawave/webservice/model/Model.java
```diff
@@ -133,28 +138,30 @@ public String getMainContent() {
...
-                if (field.isLenientMarker()) {
-                    lenientModelFields.add(field.getModelFieldName());
-                }
+            if (!field.isFieldMapping()) {
+                modelFieldAttributes.putAll(field.getModelFieldName(), field.getAttributes());
             }
         }
 
         for (FieldMapping f : this.getFields()) {
             // don't include the lenient marker itself
```
jwomeara: should this be updated?
```suggestion
            // don't include model field attributes
```

jwomeara (changes requested):
looks good overall

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
ivakegg: There is not a better answer than 'because it's always been here'.  I have no problems moving it.

ivakegg: well, actually this method seems consistent with a bunch of the other methods in this class.  Unless you have a good reason, I am tending to lean towards leaving it in here.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
ivakegg: You are correct that this should be and else if

warehouse/query-core/src/test/java/datawave/query/AnyFieldQueryTest.java
ivakegg: removed

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
jwomeara: Assuming I'm looking at the same methods as you, the difference is that those methods are used in multiple places, not just the QueryModelVisitor.  I am still in favor of moving this method since it is specific to QueryModelVisitor.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTermsTest.java
ivakegg: yup.  this is a case where the LcNoDiacritics succeeds but the numeric fails.  updating comment

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/QueryModelVisitorTest.java
ivakegg: This was just a test to explore allowing the user to only mark part of the expression.  This is really a contrived example that I don't ever expect to happen.  I can drop the test.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
jwomeara: Created a PR with a potential refactor: https://github.com/NationalSecurityAgency/datawave/pull/2022

ivakegg: But it could be used elsewhere....

ivakegg: well.... since you went through all that trouble.  Thanks.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/QueryModelVisitorTest.java
jwomeara: I ask because you have logic in JexlNodeFactory that specifically checks for this case.  If it's contrived, and users shouldn't query that way, should we support it?

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
```diff
@@ -301,17 +302,26 @@ public static JexlNode createNodeTreeFromPairs(ContainerType containerType, Jexl
...
                 leftNode = leftType.getSource();
             }
-            if (rightType.isType(LenientExpression.class)) {
+            if (rightType.isType(StrictExpression.class)) {
+                strict = true;
+                rightNode = rightType.getSource();
+            } else if (rightType.isType(LenientExpression.class)) {
                 lenient = true;
                 rightNode = rightType.getSource();
             }
```
jwomeara: These lines are related to the test that you said was a contrived example.  Curious if we need this if we don't want to support that case.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
jwomeara: Where else would we use that method though?  If you genuinely think this method is useful outside of the QueryModelVisitor, then I will close my PR (which was no trouble at all btw).

ivakegg: I merged in your PR.  I verified it was not being used elsewhere

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
ivakegg: hmm.... let me reevaluate this.

ivakegg: removed.

ivakegg:
I marked this as linked, but the linked PR is yet to be created.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryModelVisitor.java
```diff
@@ -361,6 +365,40 @@ protected JexlNode expandBinaryNodeFromModel(JexlNode node, Object data) {
         return node;
     }
 
+    /**
+     * This method is currently only used from the QueryModelVisitor to expand a binary node into multiple pairs. If one side contains the strict marker, then
+     * that will be extended around the binary node.
```
apmoriarty: Suggested comment

```
Expands a binary node into multiple pairs. If one side contains a strict marker, then that will be extended around the binary node.
```

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryModelVisitor.java
```diff
@@ -450,8 +488,45 @@ public void setStrictFields(Set<String> strictFields) {
         this.strictFields.addAll(strictFields);
     }
 
+    /**
+     * We decided that we should be lenient by default
```
apmoriarty: "Model mappings are lenient by default"

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryModelVisitor.java
```diff
@@ -450,8 +488,45 @@ public void setStrictFields(Set<String> strictFields) {
...
+        }
+
+        // lenient by default
+        return true;
+    }
+
+    /**
+     * Strict if not lenient
+     *
+     * @param field
```
apmoriarty, line 526: One day we will fail merge requests with incomplete docs

apmoriarty (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryModelVisitor.java
ivakegg: good, except the "one side contains..." part is no longer valid.  Updating comment.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryModelVisitor.java
ivakegg: check

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryModelVisitor.java
ivakegg: check.

jwomeara (approved):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryModelVisitor.java
```diff
@@ -33,21 +32,21 @@
...
 import com.google.common.collect.Sets;
 
 import datawave.query.Constants;
 import datawave.query.exceptions.DatawaveFatalQueryException;
 import datawave.query.jexl.JexlASTHelper;
 import datawave.query.jexl.JexlNodeFactory;
-import datawave.query.jexl.JexlNodeFactory.ContainerType;
 import datawave.query.jexl.LiteralRange;
 import datawave.query.jexl.nodes.BoundedRange;
+import datawave.query.jexl.nodes.QueryPropertyMarker;
```
jwomeara: unused

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryModelVisitor.java
```diff
@@ -33,21 +32,21 @@
 import org.apache.commons.jexl2.parser.LenientExpression;
```
jwomeara: unused

jwomeara (changes requested):
noticed a few unused imports in QueryModelVisitor.  can quickly approve once fixed.

jwomeara (dismissed):

apmoriarty (dismissed):

hlgp (dismissed):

PR #2038, opened 2023-07-17
Standardize Handling of FieldSets

apmoriarty:
Change Set
- fieldset serialization is now standardized
- fieldset reduction is now standardized and options exist for one reduction at the end of query planning or reduction applied on a per-shard basis (for use cases where range stream reduction is applied)
- fieldset compression inspired by https://github.com/NationalSecurityAgency/datawave/pull/1975 was added and is compatible with either mode of fieldset reduction. Note: compression only applied to indexed fields, index-only fields, and composite fields.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
+++ b/warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
@@ -101,8 +101,16 @@ public class ShardQueryConfiguration extends GenericQueryConfiguration implement
     private boolean enforceUniqueTermsWithinExpressions = false;
     // should this query reduce the set of fields prior to serialization
     private boolean reduceQueryFields = false;
+    // should the query planner intersect field sets with query fields
+    private boolean reduceFieldSets = false;
+    // should the VisitorFunction reduce field sets
+    private boolean reduceFieldSetsPerShard = false;
+    // should the query planner compress the serialized field sets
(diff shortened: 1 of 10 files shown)


warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
```diff
@@ -2031,14 +2041,51 @@ public void setEnforceUniqueTermsWithinExpressions(boolean enforceUniqueTermsWit
...
     }
 
+    /**
+     * Deprecated in favor of {@link #getReduceFieldSets()}
+     *
+     * @return a boolean
+     */
+    @Deprecated
     public boolean getReduceQueryFields() {
         return reduceQueryFields;
```
ivakegg, line 2056: Should we be dropping this field and instead call one of the getters below?

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
apmoriarty: I don't believe so. If a user has this option configured and we suddenly delegate to a completely new option that is not yet configured this pr would become a breaking change.

apmoriarty:
We did extensive testing of this feature and found no cases where compressing the field sets led to an increase in performance.

Issue #704, opened 2020-01-13
Ivarators should use the yielding mechanism


ivakegg:
Currently once you seek on an ivarator (DatawaveFieldIndexCachingIteratorJexl), the ivarator will fill the complete set before returning.  We need to enable yielding on ivarators (using Accumulo's yield mechanism) as we do in the Query Iterator on results.

SethSmucker:
Done in #2042. Closing.

PR #2042, opened 2023-07-21
Add abilty to yield in Ivarators, AndIterator, OrIterator and return metrics before yield (#704)

billoley:
Add abilty to yield in Ivarators, AndIterator, OrIterator and return metrics before yield (#704)

WaitWindowObserver object tracks the remaining time before a yield and has a bunch of convenience methods for manipulating keys.  !YIELD_AT_BEGIN and \uffffYIELD_AT_END are used to yield either before a given key or after a given key.  They are used most often in the colFam but sometimes in the colQual if the query is using non-sorted UIDs as an optimization.  The ! sorts before alphanumeric characters and the \uffff sorts after alphanumerics.  The rest of the strings are for easy identification.

The remaining time is tracked in the WaitWindowObserver using a separate thread to minimize the calls to System.currentTimeMillis. 

When time is expired, a WaitWindowOverrunException is thrown and then caught/propagated (determining the correct yieldKey) at each AndIterator/OrIterator up to the WaitWindowOverseerIterator which is under either the SerialIterator of the PipelineIterator but above the rest of the boolean stack of iterators.

If collectTimingDetails=false, we simply yield when either the SerialIterator of the PipelineIterator detects a WaitWindowOverrun.  If collectTimingDetails=true, then we return a document with a WAIT_WINDOW_OVERRUN attribute and a TIMING_METADATA that hands off its souce, next, yield, etc metrics to the DocumentTransformerSupport class and then gets ignored.  On the next call to the QueryIterator, the hasTop method checks with the WaitWindowObserver and yields.

IveratorFutures are now tracked in IteratorThreadPoolManager so that when an Ivarator yields before fillSortedSet is completed for all ranges, a post-yield call can reclaim the HDFS-backed sorted set from the IvaratorFuture.

Ivarator yielding and timeouts:

Yielding is taken care of by checks in WaitWindowObserver and uses property query.iterator.yield.threshold.ms

IteratorThreadPoolManager has a timer-based check that an Ivarator has not been in fillSortedSet more than ivaratorCacheScanTimeout which is set by the logic property IvaratorCacheScanTimeoutMinutes which is set by the property query.max.call.time.minutes (default 60 minutes).  When an Ivarator suspends and resumes, the start time is maintained such that this check operates against the cumulative time in fillSortedSet for that Ivarator.

IteratorThreadPoolManager has a timer-based check that an IvaratorRunnable (used in fillSortedSet to fill the set from part of the Range) is not running for greater than tserver.datawave.ivarator.runnableTimeoutMinutes which is set in Accumulo properties and checked frequently for changes on a timer.  The default is also 60 minutes -- so it will likely never be involked, but can be used as a failsafe.  It can also be set lower temporarily to force shutdown of all IvaratorRunnables.

IteratorThreadPoolManager also uses the tserver.datawave.ivarator.runnableTimeoutMinutes property to ensure that IvaratorFuture objectd are evicted from the Caffeine cache after 1.1 * that setting.  These objects are removed when an Ivarator completes but could be abandoned if an Ivarator is suspended on yield and then not resumed.

Included fixes to the PipelineIterator discovered during testing:

1) When yielding, PipelineIterator should evaluate all possible keys to find the lowest, including the evaluationQueue, results, and yieldKey from WaitWindowOverrun.  This could have caused missed data when yielding in the PipelineIterator

2) Must fill the evaluationQueue after calling flushCompletedResults in getNext otherwise flushCompletedResults might empty the evaluationQueue with no valid results and on the next call to getNext, results will be empty and when cacheNextResult is called, evaluationQueue will be empty and PipelineIterator declares itself done.  This could have caused missed data when yielding in the PipelineIterator

3) Pipeline should save any caught Exception so that the PipelineIterator can propagate the Exception when calling getResult.  This is particularly important when running Pipline in an Executor via the PipelineIterator.  Otherwise, the evaluation of that Document just returns null which is the same as a Document not matching.  The SerialIterator runs the Pipline.run() method inline and was not subject to this bug.

QueryIteratorIT and extended classes WaitWindowQueryIteratorSerialIT and WaitWindowQueryIteratorPipelineIT use a test harness that more closely simulates Accumulo's LookupTask.

IvaratorYieldingTest now tests Pipeline/Serial, sortedUIDs/unsortedUIDs, and collectTimingDetails T/F for 8 variations instead of the previous 2 variations Serial + sortedUIDs/unsortedUIDs.


warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -865,16 +881,34 @@ public boolean add(long val) {
 
     private void fillSortedSets() throws IOException {
         String sourceRow = this.fiRow.toString();
-        setupRowBasedHdfsBackedSet(sourceRow);
-
-        // if keys is not null, then we already had a completed set which was loaded in setupRowBasedHdfsBackedSet
-        if (keys != null) {
-            moveToNextRow();
-            return;
+        // if we are running fillSortedSets as port of a re-seek after yield, then the fillSet threads would not have
```
apmoriarty:
```suggestion
        // if we are running fillSortedSets as part of a re-seek after yield, then the fillSet threads would not have
```

warehouse/query-core/src/main/java/datawave/query/iterator/waitwindow/WaitWindowObserver.java
```diff
@@ -0,0 +1,417 @@
...
+    };
+
+    protected YieldCallback yieldCallback = null;
+    // When the wait window is over. Set during the initial seek
+    protected long endOfWaitWindow;
+    // Remaining time in the wait window. Updated by the timerTask
+    protected AtomicLong remainingTimeMs = new AtomicLong();
+    protected TimerTask timerTask = new WaitWindowTimerTask();
+    // How often the timerTask gets run
+    protected long checkPeriod = 50;
```
apmoriarty:
```suggestion
    protected long checkPeriodMillis = 50;
```

warehouse/query-core/src/main/java/datawave/query/iterator/waitwindow/WaitWindowObserver.java
```diff
@@ -0,0 +1,417 @@
...
+    static public boolean hasMarker(Text text) {
+        return hasBeginMarker(text) || hasEndMarker(text);
+    }
+
+    static public boolean hasBeginMarker(Key key) {
+        return hasBeginMarker(key.getColumnFamily()) || hasBeginMarker(key.getColumnQualifier());
+    }
+
+    static public boolean hasBeginMarker(Text text) {
+        return text.toString().contains(YIELD_AT_BEGIN_STR);
```
ivakegg: beginsWith ?

warehouse/query-core/src/main/java/datawave/query/iterator/waitwindow/WaitWindowObserver.java
```diff
@@ -0,0 +1,417 @@
...
+    }
+
+    static public boolean hasEndMarker(Key key) {
+        return hasEndMarker(key.getColumnFamily()) || hasEndMarker(key.getColumnQualifier());
+    }
+
+    static public boolean hasEndMarker(Text text) {
+        return text.toString().contains(YIELD_AT_END_STR);
```
ivakegg: endsWith ?

warehouse/query-core/src/main/java/datawave/query/iterator/waitwindow/WaitWindowObserver.java
```diff
@@ -0,0 +1,417 @@
...
+            }
+            if (!isShardKey && key.compareTo(seekStartKey, PartialKey.ROW_COLFAM_COLQUAL) <= 0) {
+                // document range
+                return seekStartKey.followingKey(PartialKey.ROW_COLFAM_COLQUAL);
+            }
+        }
+        return key;
+    }
+
+    private Key yieldKeyAfterOverrun(Key key) {
```
ivakegg, line 340: Need some method comments describing what each non-obvious method is doing please.  Applies to all methods in this class.

warehouse/query-core/src/main/java/datawave/query/iterator/waitwindow/WaitWindowOverseerIterator.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.accumulo.core.data.Key;
+import org.apache.accumulo.core.data.Range;
+import org.apache.accumulo.core.iterators.IteratorEnvironment;
+
+import datawave.query.attributes.Document;
+import datawave.query.exceptions.WaitWindowOverrunException;
+import datawave.query.iterator.NestedIterator;
+import datawave.query.iterator.SeekableNestedIterator;
+
+public class WaitWindowOverseerIterator extends SeekableNestedIterator<Key> {
```
ivakegg, line 22: At least a class level description please.

warehouse/query-core/src/main/java/datawave/query/iterator/waitwindow/WaitWindowObserver.java
```diff
@@ -0,0 +1,417 @@
...
+import org.apache.log4j.Logger;
+
+import datawave.query.attributes.Document;
+import datawave.query.attributes.DocumentKey;
+import datawave.query.attributes.WaitWindowExceededMetadata;
+import datawave.query.exceptions.WaitWindowOverrunException;
+import datawave.query.iterator.ResultCountingIterator;
+import datawave.query.iterator.profile.QuerySpan;
+
+// This class maintains common state and logic to determine if the QueryIterator's stack of boolean logic
```
ivakegg: In general, it would be good to show examples of keys in the various states as appropriate for each of the functions below.  That might make it easier to follow.

warehouse/query-core/src/main/java/datawave/query/iterator/waitwindow/WaitWindowObserver.java
```diff
@@ -0,0 +1,417 @@
...
+        return readyToYield && yieldKey != null;
+    }
+
+    public Key getYieldKey() {
+        return yieldKey;
+    }
+
+    static public boolean isShardKey(Key key) {
+        Text colFam = key.getColumnFamily();
+        return colFam.equals(new Text()) || hasBeginMarker(colFam);
```
ivakegg: Why is this check sufficient?  It is not obvious that a begin markiner in the colFam is necessarily a shard key.

billoley: Comments added.  Document-specific keys have YIELD_AT_BEGIN or YIELD_AT_END in the colQual, whicle shard keys (non-document) have these markers in the colFam.

warehouse/query-core/src/main/java/datawave/query/iterator/waitwindow/WaitWindowObserver.java
billoley: Comments added.

warehouse/query-core/src/main/java/datawave/query/iterator/waitwindow/WaitWindowObserver.java
billoley: Comments added.

warehouse/query-core/src/main/java/datawave/query/iterator/waitwindow/WaitWindowOverseerIterator.java
billoley: Comments added.

warehouse/query-core/src/main/java/datawave/query/iterator/waitwindow/WaitWindowObserver.java
billoley: Won't work because we have to add one or more  \x00 to the end sometimes

warehouse/query-core/src/main/java/datawave/query/iterator/waitwindow/WaitWindowObserver.java
billoley: This one works.  OK

Rexblk81 (approved):

warehouse/query-core/src/main/java/datawave/query/iterator/waitwindow/WaitWindowObserver.java
```diff
@@ -0,0 +1,505 @@
...
+            } else {
+                yieldKey = highestYieldKey(keys);
+            }
+        }
+        throw new WaitWindowOverrunException(yieldKey);
+    }
+
+    /*
+     * When yieldKey is set and collectTimingDetails=true, then we yield on the second call from QueryIterator.hasTop so that the first call can return the
+     * document that contains the WAIT_WINDOW_OVERRUN and TIMING_METADATA attributes. readyToYield will be set to true after the first call to yieldOnOverrun
```
ivakegg, line 215: Remind me again, why are we bothering to yield on the second call if we are returning a document with the TIMING_METADATA.  Does not returning a document allow the tablet to execute the other scans which is the point of yielding?

warehouse/query-core/src/main/java/datawave/core/iterators/IteratorThreadPoolManager.java
```diff
@@ -70,15 +131,23 @@ private ThreadPoolExecutor createExecutorService(int maxThreads, String name) {
         return pool;
     }
 
-    private int getMaxThreads(final String prop, AccumuloConfiguration conf) {
-        if (conf != null) {
+    private int getIntPropertyValue(final String prop, int defaultValue, AccumuloConfiguration config) {
+        if (config != null) {
             Map<String,String> properties = new TreeMap<>();
-            conf.getProperties(properties, k -> Objects.equals(k, prop));
+            config.getProperties(properties, k -> Objects.equals(k, prop));
```
ivakegg: confused between this method and the getLongPropertyValue.  Why not use config.get(prop) ?

warehouse/query-core/src/main/java/datawave/query/exceptions/WaitWindowOverrunException.java
```diff
@@ -0,0 +1,36 @@
...
+        return yieldKey;
+    }
+
+    // disallow other constructors that don't include a yieldKey
+    private WaitWindowOverrunException() {
+        super();
+    }
+
+    // disallow constructors that don't include a yieldKey
+    private WaitWindowOverrunException(String message) {
```
ivakegg: This constructor would not be available anyway.  No need to override.  Same comment for all constructors.  You are supplying a constructor so even the default constructor would not be available.

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -554,14 +579,17 @@ public void seek(Range r, Collection<ByteSequence> columnFamilies, boolean inclu
...
-            int fieldnameIndex = cq.indexOf('\0');
-            if (fieldnameIndex >= 0) {
-                String cf = startKey.getColumnFamily().toString();
-                lastFiKey = new Key(startKey.getRow().toString(), "fi\0" + cq.substring(0, fieldnameIndex),
-                                cq.substring(fieldnameIndex + 1) + '\0' + cf + '\0');
+        Key startKey = r.getStartKey();
+        if (!sortedUIDs) {
+            String cq = WaitWindowObserver.removeMarkers(startKey.getColumnQualifier()).toString();
+            if (r.getStartKey().getColumnFamily().getLength() > 0 && cq.length() > 0) {
+                String cqSuffix = WaitWindowObserver.hasBeginMarker(startKey.getColumnQualifier()) ? "" : "\0";
```
ivakegg: comments on why a begin marker results in a different suffix would be good.

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -865,16 +881,34 @@ public boolean add(long val) {
...
-            moveToNextRow();
-            return;
+        // if we are running fillSortedSets as port of a re-seek after yield, then the fillSet threads would not have
+        // completed when a WaitWindowOverrunException was thrown and therefore the set would not have been marked as
+        // complete when setupRowBasedHdfsBackedSet was called at the top of this method. We will try to copy the
+        // RowBasedHdfsBackedSet from the previously used Ivarator since there may be unpersisted results or
+        // a long-running Ivarator may still be filling the RowBasedHdfsBackedSet
+        // If this re-seek is from an Ivarator yielding, the startKey will have a yield marker in either the
+        // colFam (for a shard range) or colQual (for a document range)
+        boolean usePreviousSortedSet = WaitWindowObserver.hasMarker(lastRangeSeeked.getStartKey());
```
ivakegg, line 891: I think we are missing the case where the ivarator filled the set, and the tablet migrates.  We need to think through those scenarios.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/KeyAdjudicator.java
```diff
@@ -46,17 +44,18 @@ public KeyAdjudicator() {
     @Override
     public Entry<Key,T> apply(Entry<Key,T> entry) {
         final Key entryKey = entry.getKey();
-        return Maps.immutableEntry(new Key(entryKey.getRow(), entryKey.getColumnFamily(), colQualRef, entryKey.getColumnVisibility(), entryKey.getTimestamp()),
-                        entry.getValue());
+        if (WaitWindowObserver.hasMarker(entryKey)) {
```
ivakegg, line 49: I need an explaination as to why we should not be updating the cq in this case.

warehouse/query-core/src/main/java/datawave/query/iterator/waitwindow/WaitWindowObserver.java
billoley: As long as itr.hasTop(), TabletBase continues to add a new KVEntry(key, value) to results and call next().
Documents with TIMING_METADATA and WAIT_WINDOW_OVERRUN are seen as resular documents at this point.
It's necessary to yield after that to force the results back to the caller and rebuild the iterator.

warehouse/query-core/src/main/java/datawave/core/iterators/IteratorThreadPoolManager.java
billoley: The note on AccumuloConfiguration.get() says that it is inefficient for keys that are not properties and recommends this approach.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/KeyAdjudicator.java
billoley: Added comment: if the key has a YIELD_AT_BEGIN or YIELD_AT_END marker, then don't modify the key
because doing so will adversely affect the subsequent yield and re-seek position

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
billoley: added comment

warehouse/query-core/src/main/java/datawave/query/exceptions/WaitWindowOverrunException.java
billoley: deleted unnecessary constructors

warehouse/query-core/src/main/java/datawave/query/iterator/logic/OrIterator.java
```diff
@@ -119,56 +141,68 @@ public T next() {
...
-                includeHeads = advanceIterators(lowest);
+                // regardless of where we hit make sure to advance includeHeads if it matches there
+                if (includeHeads != null && includeHeads.containsKey(lowest)) {
+                    includeHeads = advanceIterators(lowest);
+                }
+            }
+        } catch (WaitWindowOverrunException e) {
+            // if prev != null then it's a match that has not been returned
+            Key possibleYieldKey = null;
+            if (prev != null || !candidateSet.isEmpty()) {
```
apmoriarty: I think this needs to be `if (!isContextRequired() && (prev != null || !candidateSet.isEmpty()) {`

There is a case where the candidate set is populated but context is required to evaluate it `(A || !B)` or `(A || (!B && !C)`. In these cases the union can't actually map to a top key. An attempt to solve that problem is over in #2149

warehouse/query-core/src/main/java/datawave/query/iterator/logic/AndIterator.java
```diff
@@ -45,10 +49,15 @@ public class AndIterator<T extends Comparable<T>> implements NestedIterator<T>,
...
     public AndIterator(Iterable<NestedIterator<T>> sources) {
-        this(sources, null);
+        this(sources, null, null);
     }
 
     public AndIterator(Iterable<NestedIterator<T>> sources, Iterable<NestedIterator<T>> filters) {
+        this(sources, filters, null);
+    }
+
+    public AndIterator(Iterable<NestedIterator<T>> sources, Iterable<NestedIterator<T>> filters, WaitWindowObserver waitWindowObserver) {
```
apmoriarty, line 62: Yielding in the AndIterator looks good

warehouse/query-core/src/main/java/datawave/query/iterator/logic/OrIterator.java
billoley: But if prev != null, then that is a look-ahead key that was found to be a match and would have been returned next.  Assuming your logic is true....  shouldn't the change be:

`
if (prev != null || (!isContextRequired() && !candidateSet.isEmpty())
`

apmoriarty: Right you are!

billoley: Committed change

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/iterator/pipeline/SerialIterator.java
```diff
@@ -83,14 +86,52 @@ public void remove() {
...
+                result = handleWaitWindowOverrun(e);
             }
         } else {
             result = null;
         }
     }
+
+    // If collectTimingDetails == true, then we wil set the (future) yieldKey in waitWindowObserver and return
+    // an entry with the yield key and a WAIT_WINDOW_OVERRUN document to which the timing details can be added
+    // If collectTimingDetails == false, then we yield and return a null
```
apmoriarty: I would prefer to see this as a formal method doc, but not necessary for approval

warehouse/query-core/src/main/java/datawave/query/iterator/profile/QuerySpanCollector.java
```diff
@@ -42,18 +45,25 @@ public void addQuerySpan(QuerySpan querySpan) {
         }
     }
 
-    private QuerySpan combineQuerySpans() {
-
+    private QuerySpan snapshot(boolean reset) {
```
apmoriarty, line 55: Add method doc

warehouse/query-core/src/test/java/datawave/query/iterator/QueryIteratorIT.java
```diff
@@ -128,7 +134,10 @@ public void setup() throws IOException {
...
 
         // set a query id
-        options.put(QUERY_ID, "000001");
+        options.put(QUERY_ID, UUID.randomUUID().toString());
+
+        // set a scanId
+        options.put(SCAN_ID, UUID.randomUUID().toString());
 
         // set a scanId
         options.put(SCAN_ID, "000002");
```
apmoriarty: This overrides the random UUID. Scan ID should be either a uuid or "00002"

warehouse/query-core/src/test/java/datawave/query/iterator/QueryIteratorIT.java
```diff
@@ -258,47 +267,47 @@ public void indexOnly_documentSpecific_test() throws IOException {
         // build the seek range for a document specific pull
         Range seekRange = getDocumentRange("123.345.456");
         String query = "INDEX_ONLY_FIELD1 == 'apple'";
-        indexOnly_test(seekRange, query, false, Collections.emptyList(), Collections.emptyList());
+        indexOnly_test(seekRange, query, false, Collections.EMPTY_LIST, Collections.EMPTY_LIST);
```
apmoriarty: I would prefer to keep the emptyList() access of the static empty list

apmoriarty (dismissed):
Noted a few preferences, otherwise looks good.

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexListIteratorJexl.java
```diff
@@ -100,14 +100,27 @@ public SortedKeyValueIterator<Key,Value> deepCopy(IteratorEnvironment env) {
         return new DatawaveFieldIndexListIteratorJexl(this, env);
     }
 
+    @Override
+    public String toStringNoQueryId() {
```
ivakegg: Why override this method if it is identical to the parent?

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexListIteratorJexl.java
```diff
@@ -100,14 +100,27 @@ public SortedKeyValueIterator<Key,Value> deepCopy(IteratorEnvironment env) {
...
+        return toStringImpl(false);
+    }
+
     @Override
     public String toString() {
+        return toStringImpl(true);
```
ivakegg: Why override this method if it is identical to the parent?

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexListIteratorJexl.java
```diff
@@ -100,14 +100,27 @@ public SortedKeyValueIterator<Key,Value> deepCopy(IteratorEnvironment env) {
...
+    }
+
+    private String toStringImpl(boolean includeQueryId) {
```
ivakegg: Make protected and @ override

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexRangeIteratorJexl.java
```diff
@@ -107,12 +107,25 @@ public SortedKeyValueIterator<Key,Value> deepCopy(IteratorEnvironment env) {
         return new DatawaveFieldIndexRangeIteratorJexl(this, env);
     }
 
+    @Override
+    public String toStringNoQueryId() {
+        return toStringImpl(false);
```
ivakegg: Why override if identical to parent method

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexRangeIteratorJexl.java
```diff
@@ -107,12 +107,25 @@ public SortedKeyValueIterator<Key,Value> deepCopy(IteratorEnvironment env) {
...
+    }
+
     @Override
     public String toString() {
+        return toStringImpl(true);
+    }
+
+    private String toStringImpl(boolean includeQueryId) {
```
ivakegg: Make protected and @ override

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexRangeIteratorJexl.java
```diff
@@ -107,12 +107,25 @@ public SortedKeyValueIterator<Key,Value> deepCopy(IteratorEnvironment env) {
...
     }
 
+    @Override
+    public String toStringNoQueryId() {
+        return toStringImpl(false);
+    }
+
     @Override
     public String toString() {
+        return toStringImpl(true);
```
ivakegg: Why override if identical to parent method

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexRegexIteratorJexl.java
```diff
@@ -78,11 +78,24 @@ public SortedKeyValueIterator<Key,Value> deepCopy(IteratorEnvironment env) {
         return new DatawaveFieldIndexRegexIteratorJexl(this, env);
     }
 
+    @Override
+    public String toStringNoQueryId() {
+        return toStringImpl(false);
```
ivakegg: Why override if identical to parent

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexRegexIteratorJexl.java
```diff
@@ -78,11 +78,24 @@ public SortedKeyValueIterator<Key,Value> deepCopy(IteratorEnvironment env) {
...
+    }
+
     @Override
     public String toString() {
+        return toStringImpl(true);
```
ivakegg: Why override if identical to parent

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexFilterIteratorJexl.java
```diff
@@ -68,12 +68,25 @@ public SortedKeyValueIterator<Key,Value> deepCopy(IteratorEnvironment env) {
...
+    public String toStringNoQueryId() {
+        return toStringImpl(false);
+    }
+
     @Override
     public String toString() {
+        return toStringImpl(true);
+    }
+
+    private String toStringImpl(boolean includeQueryId) {
```
ivakegg: Make protected and @ override

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexRegexIteratorJexl.java
```diff
@@ -78,11 +78,24 @@ public SortedKeyValueIterator<Key,Value> deepCopy(IteratorEnvironment env) {
...
+    }
+
+    private String toStringImpl(boolean includeQueryId) {
```
ivakegg: Make protected and @ override

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexFilterIteratorJexl.java
```diff
@@ -68,12 +68,25 @@ public SortedKeyValueIterator<Key,Value> deepCopy(IteratorEnvironment env) {
         return new DatawaveFieldIndexFilterIteratorJexl(this, env);
     }
 
+    @Override
+    public String toStringNoQueryId() {
+        return toStringImpl(false);
```
ivakegg: Why override if identical to parent

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexFilterIteratorJexl.java
```diff
@@ -68,12 +68,25 @@ public SortedKeyValueIterator<Key,Value> deepCopy(IteratorEnvironment env) {
...
+    }
+
     @Override
     public String toString() {
+        return toStringImpl(true);
```
ivakegg: Why override if identical to parent

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -1655,16 +1730,68 @@ public void setCollectTimingDetails(boolean collectTimingDetails) {
...
+    public String toStringNoQueryId() {
+        return toStringImpl(false);
+    }
+
     @Override
     public String toString() {
+        return toStringImpl(true);
+    }
+
+    private String toStringImpl(boolean includeQueryId) {
```
ivakegg: make protected

ivakegg: comment as to why we have this includeQueryId switch

warehouse/query-core/src/main/java/datawave/query/attributes/WaitWindowExceededMetadata.java
```diff
@@ -0,0 +1,12 @@
+package datawave.query.attributes;
+
+/**
+ * Indicates that the scan session was ended after exceeding the wait window
```
ivakegg: Explain why this is an Attribute (i.e. added to document to pass to .....)

warehouse/query-core/src/main/java/datawave/query/iterator/waitwindow/WaitWindowObserver.java
```diff
@@ -0,0 +1,584 @@
...
+     * document that contains the WAIT_WINDOW_OVERRUN and TIMING_METADATA attributes. Since we can not yield on the same key that was just returned, we have to
+     * yield on a key that follows the yieldKey. For YIELD_AT_END, we add a null char after YIELD_AT_END which is at the end of the colFam or colQual. For
+     * YIELD_AT_BEGIN, we add a null character after that marker.
+     */
+    private Key yieldKeyAfterOverrun(Key key) {
+        Text row = key.getRow();
+        Text colFam = key.getColumnFamily();
+        Text colQual = key.getColumnQualifier();
+        if (hasBeginMarker(key)) {
+            if (hasBeginMarker(colFam)) {
```
ivakegg, line 345: I am trying to figure out why we can have a marker in wither the colFam or the colQual.  Why not always place in the colQual regardless?  I believe you have have a key with an empty (0 length) colFam and still have data in the colQual.   If that presumpting is incorrect then this makes perfect sense to me.  If that presumtion is correct, then my question stands.

ivakegg (commented):
I still have to complete this review, but here is a start.

warehouse/query-core/src/main/java/datawave/query/iterator/waitwindow/WaitWindowObserver.java
ivakegg: nevermind... explained in the createDocumentYieldKey method above.

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
billoley: includeQueryId parameter to toStringImpl is so we can have the same logic as in toString() when using this infomation in IvaratorRunnable logs, Ivarator logs, as well as in task names and thread names.  The queryId is part of a standard set of information that will occur in the same order in these places and we don't need to have it there twice.

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexListIteratorJexl.java
billoley: Fixed

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexListIteratorJexl.java
billoley: Fixed

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexRangeIteratorJexl.java
billoley: Fixed

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexRangeIteratorJexl.java
billoley: Fixed

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexFilterIteratorJexl.java
billoley: Fixed

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexFilterIteratorJexl.java
billoley: Fixed

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
billoley: Fixed

warehouse/query-core/src/main/java/datawave/query/attributes/WaitWindowExceededMetadata.java
billoley: Fixed

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexListIteratorJexl.java
billoley: Fixed

apmoriarty (approved):

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -544,34 +576,37 @@ public void seek(Range r, Collection<ByteSequence> columnFamilies, boolean inclu
...
-            Key startKey = r.getStartKey();
-            String cq = startKey.getColumnQualifier().toString();
-            int fieldnameIndex = cq.indexOf('\0');
-            if (fieldnameIndex >= 0) {
-                String cf = startKey.getColumnFamily().toString();
-                lastFiKey = new Key(startKey.getRow().toString(), "fi\0" + cq.substring(0, fieldnameIndex),
-                                cq.substring(fieldnameIndex + 1) + '\0' + cf + '\0');
+        Key startKey = r.getStartKey();
+        if (!sortedUIDs) {
+            String cq = WaitWindowObserver.removeMarkers(startKey.getColumnQualifier()).toString();
```
ivakegg, line 592: Can we get a little more inline documentation as to what we are doing when we update the WaitWindowObserver?

ivakegg (approved):

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
billoley: I can add more documentation as necessary.  However, this line is a static call to strip the YIELD_AT_BEGIN or YIELD_AT_END marker from the colQual (if it exists).  The WaitWindowObserver itself is not affected.

PR #2043, opened 2023-07-25
Pass query parameters to remoteQueryService.planQuery call

drewfarris:
* Previously, no parameters were passed as a part of the remote call to the `/plan` endpoint. This meant the receiver did not get any information about the query to plan, resulting in an odd `415 Unsupported Media Type` error due to the empty POST body.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/tables/RemoteEventQueryLogic.java
+++ b/warehouse/query-core/src/main/java/datawave/query/tables/RemoteEventQueryLogic.java
@@ -103,7 +103,7 @@ public class RemoteEventQueryLogic extends BaseQueryLogic<EventBase> implements 
     
     @Override
     public String getPlan(Connector connection, Query settings, Set<Authorizations> auths, boolean expandFields, boolean expandValues) throws Exception {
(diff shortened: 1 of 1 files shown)


apmoriarty (approved):

mineralntl (approved):

drewfarris:
Interestingly, this issue was present in the [`release/version3.41` branch](https://github.com/NationalSecurityAgency/datawave/blob/cf02185564115b67d103552e5b7547feedbd872d/warehouse/query-core/src/main/java/datawave/query/tables/RemoteEventQueryLogic.java#L107) and the [`integraton` branch](https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/query-core/src/main/java/datawave/query/tables/RemoteEventQueryLogic.java#L107), but not the [`release/version5.2`](https://github.com/NationalSecurityAgency/datawave/blob/release/version5.2/warehouse/query-core/src/main/java/datawave/query/tables/RemoteEventQueryLogic.java#L106) and [`release/version5.8`](https://github.com/NationalSecurityAgency/datawave/blob/release/version5.8/warehouse/query-core/src/main/java/datawave/query/tables/RemoteEventQueryLogic.java#L106) branches.

We'll need to be careful when we merge upwards.

PR #2049, opened 2023-07-28
fixes #1920

### LINKED ISSUE #1920: Ivarators don't always create 'complete' files
In DatawaveFieldIndexCachingIteratorJexl, after we have filled all of the sorted sets we make a decision as to whether or not we should create a 'complete' file in the ivarator directory.  

If the set representing the data has any entries buffered in memory that are not persisted to disk, we persist those entries to disk, and then create a 'complete' file.  If the set representing the data has a null buffer (meaning all data has already been persisted) we don't bother creating the 'complete' file.  

We should investigate the logic to determine whether it is safe to mark the ivarator directory as complete in both cases.

ivakegg:
* Added comment explaining why we do not persist the ivarator set in some situations.


matthpeterson (approved):

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -814,6 +814,9 @@ else if (key.compareTo(this.lastRangeSeeked.getStartKey()) < 0) {
                 }
 
                 // if we have any persisted data or we have scanned a significant number of keys, then persist it completely
+                // Note that we will not persist the data if we are not over the scan threshold and the data fits into
+                // one buffer (no persisted data). That means that if we need to rebuild the sorted set that we can do
+                // it relatively quickly so no need for a persisted set with a complete marker.
```
jwomeara, line 820: I'm not sure what this comment has to do with the original issue I reported.  

In the scenario I described, one of two things happen.  

Either, 

1) We have entries buffered in memory, and that triggers the ivarator to persist the entries to disk and create a `complete` file.

OR 

2) We have no entries buffered in memory (because everything was ALREADY persisted to disk), and that does NOT trigger the creation of a `complete` file.

I don't think this has anything to do with the scan threshold.  If it did, then in scenario `#1` the code wouldn't persist the buffer to disk, (but it absolutely does!!!).  Scenario `#1` is not the problem.  

Scenario `#2` where data has already been written to disk, but there are no results in the in-memory buffer is the problem.  We already have data written to disk, but we aren't creating the `complete` file.

jwomeara (changes requested):

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
ivakegg: I am pretty sure that "this.set.hasPersistedData()" is the part of the check that takes care of your scenario 2.

hlgp: @jwomeara are you saying that the code in scenario 2 is not making an accurate determination about the sate of persistence?  That it's not enough that the buffer is null to call it "persisted".  Should the check be buffer == null && isPersisted instead of an || ?  Or am I misunderstanding altogether?

PR #2052, opened 2023-08-01
Updated AsyncIndexLookup to wait for timed out threads to finish befo…

jwomeara:
…re accessing shared properties.


billoley:
Changes look fine.  A possible enhancement since you're adding another parameter (CoundDownLatch) that gets passed from three different places would be to encapsulate the common parameters in a class that extends Future and contains the CountDownLatches and timeouts.

timedScanFuture aka Future<Boolean> that is passed along with two CountdDownLatches, startTimeMillis, and timeout could be enclosed in a class TimedScanFuture that would get created in BoundedRangeIndexLookup.createTimedCallable() and FieldNameIndexLookup.createTimedCallable() by submitting to the executor and returning the future there and in RegexIndexLookup.createTimedCallable() by doing the same thing and then storing the TimedScanFuture in RegexLookupData.

billoley (dismissed):

jwomeara:
> Changes look fine. A possible enhancement since you're adding another parameter (CoundDownLatch) that gets passed from three different places would be to encapsulate the common parameters in a class that extends Future and contains the CountDownLatches and timeouts.
> 
> timedScanFuture aka Future that is passed along with two CountdDownLatches, startTimeMillis, and timeout could be enclosed in a class TimedScanFuture that would get created in BoundedRangeIndexLookup.createTimedCallable() and FieldNameIndexLookup.createTimedCallable() by submitting to the executor and returning the future there and in RegexIndexLookup.createTimedCallable() by doing the same thing and then storing the TimedScanFuture in RegexLookupData.

So you're suggesting I create a wrapper around future then?  Not necessarily an extension of future, right?  Because the future is created by the executor on submit, so I can't control what type of future is created - i can only wrap what I receive.  

The timedWaitScan call doesn't have that many arguments IMO, and to me, passing the latches as arguments in the method in 3 places is no different than passing them to the constructor of whatever wraps the future in 3 places.  I would prefer to leave this as-is.

billoley:
> So you're suggesting I create a wrapper around future then? Not necessarily an extension of future, right? Because the future is created by the executor on submit, so I can't control what type of future is created - i can only wrap what I receive.
> 
> The timedWaitScan call doesn't have that many arguments IMO, and to me, passing the latches as arguments in the method in 3 places is no different than passing them to the constructor of whatever wraps the future in 3 places. I would prefer to leave this as-is.

Yes, a wrapper (can still implement Future interface and encapsulate the returned Future) but I'm OK with leaving them as parameters.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BoundedRangeIndexLookup.java
```diff
@@ -206,111 +207,116 @@ public synchronized IndexLookupMap lookup() {
...
+                            throw new Exception("Timeout exceeded for bounded range lookup");
+                        }
 
-                    k.getRow(holder);
-                    String uniqueTerm = holder.toString();
+                        Key k = entry.getKey();
 
-                    SortedMap<Key,Value> keymap = WholeRowIterator.decodeRow(entry.getKey(), entry.getValue());
+                        if (log.isTraceEnabled()) {
+                            log.trace("Foward Index entry: " + entry.getKey());
```
avgAGB: small typo here: Forward Index entry:

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/RegexIndexLookup.java
```diff
@@ -251,107 +251,113 @@ public synchronized IndexLookupMap lookup() {
...
-                        log.trace("Foward Index entry: " + entry.getKey());
-                    }
+                        Key topKey = entry.getKey();
 
-                    // Get the column qualifier from the key. It contains the datatype and normalizer class
-                    if (null != topKey.getColumnQualifier()) {
-                        String colq = topKey.getColumnQualifier().toString();
-                        int idx = colq.indexOf(Constants.NULL);
+                        if (log.isTraceEnabled()) {
+                            log.trace("Foward Index entry: " + entry.getKey());
```
avgAGB: Similar typo here.

keith-ratcliffe (approved):
The changes here look good to me

hlgp (approved):

PR #2053, opened 2023-08-03
re #2008: Make the default normalization expansions as they were before

ivakegg:
* This will make any lenient/script code updates only applicable if a model field is explicitly marked as such
* The default normalization strategy remains as it was before.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryModelVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryModelVisitor.java
@@ -2,14 +2,12 @@ package datawave.query.jexl.visitors;
 
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Lists;
-import com.google.common.collect.Multimap;
 import com.google.common.collect.SetMultimap;
 import com.google.common.collect.Sets;
 import datawave.query.Constants;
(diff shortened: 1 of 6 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryModelVisitor.java
```diff
@@ -317,6 +316,14 @@ protected JexlNode expandBinaryNodeFromModel(JexlNode node, Object data) {
             List<JexlNode> expressions = new ArrayList<>();
             for (List<String> fieldPair : fieldPairs) {
                 boolean strict = isStrict(fieldPair.get(0)) || isStrict(fieldPair.get(1));
+                boolean lenient = isLenient(fieldPair.get(0)) || isLenient(fieldPair.get(1));
+
+                // if somehow this expression is marked as lenient and strict, then default to ole behaviour
```
apmoriarty:
```suggestion
                // if somehow this expression is marked as lenient and strict, then default to old behavior
```

ivakegg: fixing....

apmoriarty (dismissed):

hlgp (dismissed):

Issue #2055, opened 2023-08-03
Investigate why quickstart docker build is suddenly broken in GitHub Actions


keith-ratcliffe:
So far, it's seemingly due to a change in docker and BuildKit  behavior on the github runners, likely due to a version change.
E.g., the experimental `--squash` feature is no longer working, our `.dockerignore` file is suddenly being ignored, etc, all of which is causing the image size to grow very large, perhaps beyond github's limits.

Still investigating...

Update: 
The main culprit was that the symlink [created here](https://github.com/NationalSecurityAgency/datawave/blob/662a47e07c4a49bcc420b1e7905349d1fc620f40/contrib/datawave-quickstart/docker/docker-build.sh#L143) for [datawave-quickstart/docker/.dockerignore](https://github.com/NationalSecurityAgency/datawave/blob/662a47e07c4a49bcc420b1e7905349d1fc620f40/contrib/datawave-quickstart/docker/.dockerignore) stopped being loaded correctly by docker build, resulting in a massively bloated and error prone image. Not 100% sure what caused that behavior to suddenly change in Github's `ubuntu-latest` runner, but using a direct copy of `.dockerignore`, rather than a symlink, has resolved the issue

keith-ratcliffe:
Fix cherry-picked and pushed to all the 5.x branches

PR #2057, opened 2023-08-04
Update Dockerfile

MiguelRicardos:


keith-ratcliffe:
FWIW, I already have a WIP pr for this. See #2055/#2056

There are larger issues at play. The new git symptom is due to the fact that the .dockerignore file is being ignored / no longer working, and thus the .git dir is making it into the build context

That is causing other issues to surface down the line, even after working around the git issue

keith-ratcliffe:
Just a heads up that this can be closed, if you want. My fix in #2056 (#2055) has been merged and cherry-picked into integration and all the remaining 5.x branches

PR #2058, opened 2023-08-04
eliminate duplicate normalization when fetching term offsets

apmoriarty:
- term offset populator no longer normalizes query terms
- small code cleanups


jack-gitdev (approved):
Looks good. I was thinking the null docKeyFunction might cause an exception but the other TermOffsetFunction constructor uses a null value for it so that should be good.

MiguelRicardos (approved):

PR #2061, opened 2023-08-07
Fixing sudo commands in start-loader

avgAGB:


matthpeterson (dismissed):

warehouse/ingest-scripts/src/main/resources/bin/ingest/start-loader.sh
```diff
@@ -83,15 +83,15 @@ if [[${TOTAL} >0 ]]; then
...
     while [[ ! -z $portInUse ]]; do
       SHUTDOWN_PORT=$((SHUTDOWN_PORT + 1))
-      portInUse=$(sudo netstat -tupln | grep $SHUTDOWN_PORT)
+      portInUse=$(ps -eaf | grep [b]ulkIngestMap | grep $SHUTDOWN_PORT)
     done
     for ((x = 0; x < $COUNT; x = $((x + 1)))); do
       $MAPFILE_LOADER_CMD -srcHdfs ${MAP_LOADER_HDFS_NAME_NODE} -destHdfs ${MAP_LOADER_HDFS_NAME_NODE} -shutdownPort ${SHUTDOWN_PORT} >>$LOG_DIR/map-file-loader.$LOADER$COUNT.log 2>&1 &
       SHUTDOWN_PORT=$((SHUTDOWN_PORT + 1))
-      portInUse=`sudo netstat -tupln | grep $SHUTDOWN_PORT`
+      portInUse=$(ps -eaf | grep [b]ulkIngestMap | grep $SHUTDOWN_PORT)
```
keith-ratcliffe: Looking at this more closely, 'portInUse' isn't used in this loop. Seems like there's nothing stopping a used port from being chosen for subsequent loaders

warehouse/ingest-scripts/src/main/resources/bin/ingest/start-loader.sh
```diff
@@ -83,15 +83,15 @@ if [[${TOTAL} >0 ]]; then
     COUNT=${NUM_MAP_LOADERS_COPY[$LOADER]}
     echo "starting $COUNT map file loaders for ${MAP_LOADER_HDFS_NAME_NODES[$LOADER]} ..."
     SHUTDOWN_PORT=24100
-    portInUse=$(sudo netstat -tupln | grep $SHUTDOWN_PORT)
+    portInUse=$(ps -eaf | grep [b]ulkIngestMap | grep $SHUTDOWN_PORT)
```
keith-ratcliffe: Rather than `ps -eaf | grep [b]ulkIngestMap | grep $SHUTDOWN_PORT`, would it be better to use something like `lsof -i :$SHUTDOWN_PORT` instead, so at least all the current user's processes can be queried?

warehouse/ingest-scripts/src/main/resources/bin/ingest/start-loader.sh
```diff
@@ -121,10 +121,10 @@ if [[${TOTAL} >0 ]]; then
     for ((CUSTOM_LOADER = 0; CUSTOM_LOADER < ${#MAP_LOADER_CUSTOM[@]}; CUSTOM_LOADER = $((CUSTOM_LOADER + 1)))); do
       echo "starting additional map file loader: ${MAP_LOADER_CUSTOM[$CUSTOM_LOADER]}"
       SHUTDOWN_PORT=25100
-      portInUse=$(sudo netstat -tupln | grep $SHUTDOWN_PORT)
+      portInUse=$(ps -eaf | grep [b]ulkIngestMap | grep $SHUTDOWN_PORT)
       while [[ ! -z $portInUse ]]; do
         SHUTDOWN_PORT=$((SHUTDOWN_PORT + 1))
-        portInUse=$(sudo netstat -tupln | grep $SHUTDOWN_PORT)
+        portInUse=$(ps -eaf | grep [b]ulkIngestMap | grep $SHUTDOWN_PORT)
       done
```
keith-ratcliffe: In hindsight, maybe better to replace the three instances of this logic with a single function(?)

warehouse/ingest-scripts/src/main/resources/bin/ingest/start-loader.sh
avgAGB: Both could work - Since the process we're looking to grab is an ingest process we may not necessarily need to bring in anything more processes than that in order to verify the loaders aren't using the port in question. I'll test this out a little and adjust if needed. 

EDIT: One of my reservations on changing the command is the difference in permissions on what commands the user can run depending on the cluster environment (for example ran into an issue before where 'netstat' wasn't permissible to run in areas where the loaders were trying to start without sudo) To be on the safe side, the ps -eaf command has pulled the processes that we seek to verify in testing on DEV so I think I'll keep this as is.

warehouse/ingest-scripts/src/main/resources/bin/ingest/start-loader.sh
avgAGB: Good point, will fix this

warehouse/ingest-scripts/src/main/resources/bin/ingest/start-loader.sh
avgAGB: Makes sense. Will adjust this in the next.

ivakegg (approved):

jack-gitdev (approved):

PR #2073, opened 2023-08-26
Refactoring hadoop-common dependency - replacing with client-api/client-runtime

avgAGB:


pom.xml
```diff
@@ -653,6 +653,16 @@
                     </exclusion>
                 </exclusions>
             </dependency>
+            <dependency>
+                <groupId>org.apache.hadoop</groupId>
+                <artifactId>hadoop-client-api</artifactId>
+                <version>3.3.3</version>
```
alerman: You added version 3.3.5 everywhere else. Move the versions up to the top level pom and make a property to make our lives easier when we update.

alerman (changes requested):

pom.xml
avgAGB: Ah thanks. Will fix this.

alerman (dismissed):

keith-ratcliffe (dismissed):

mineralntl (dismissed):

awrigh11 (dismissed):

mineralntl (approved):

matthpeterson (approved):

PR #2081, opened 2023-09-05
Fail when AndIterator  drops IndexOnly term

apmoriarty:
Addressed two different bugs in the AndIterator

- dropping an index-only term from an intersection will now appropriately fail the query
- documents are now built using the `contextIncludes` iterators when context is necessary

As a note: several additional defects were discovered by testing with random inputs. These unit tests in the `AndOrIteratorIT` are disabled and a fix for those will be coming later.


warehouse/query-core/src/test/java/datawave/query/iterator/logic/AndIteratorIT.java
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    /**
+     * Triggers the "Lookup of event field failed, precision of query reduced" warning
+     *
+     * @throws IOException
+     *             on purpose
+     */
+    @Test
+    void testIterationInterruptedOnInitialSeekOfEventField() throws IOException {
```
FineAndDandy, line 82: why don't we want this to throw an exception? If B is index only, hits will be missed

warehouse/query-core/src/main/java/datawave/query/iterator/NestedIterator.java
```diff
@@ -77,4 +77,12 @@ default void setEnvironment(IteratorEnvironment env) {}
      *            non-null context for evaluation
      */
     void setContext(T context);
+
+    /**
+     * This information is required for safely dropping terms from an intersection in the AndIterator
+     *
+     * @return true if the field is index only
+     */
+    boolean isIndexOnly();
```
FineAndDandy: can we avoid tracking this explicitly by just checking if the context iterator has a document?

warehouse/query-core/src/main/java/datawave/query/iterator/Util.java
```diff
@@ -142,4 +142,33 @@ public static <T> Document buildNewDocument(Iterable<? extends NestedIterator<T>
...
+            for (NestedIterator<T> include : includes.get(lowest)) {
+                d.putAll(include.document().getDictionary().entrySet().iterator(), false);
+            }
+        }
+
+        if (contextIncludes != null) {
+            // context includes may not map to the lowest provided key
+            for (NestedIterator<T> contextInclude : contextIncludes.get(lowest)) {
+                Document doc = contextInclude.document();
+                if (doc != null) {
```
FineAndDandy, line 166: this check is what made me realize we might not need to track index only state, but rather just if the underlying document exists.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/AndIterator.java
```diff
@@ -256,17 +258,30 @@ public void seek(Range range, Collection<ByteSequence> columnFamilies, boolean i
             try {
                 for (NestedIterator<T> itr : child.leaves()) {
                     if (itr instanceof SeekableIterator) {
-                        ((SeekableIterator) itr).seek(range, columnFamilies, inclusive);
+                        try {
+                            ((SeekableIterator) itr).seek(range, columnFamilies, inclusive);
+                        } catch (Exception e2) {
+                            if (itr.isIndexOnly()) {
```
FineAndDandy: could be replaced by a check of if the itr had a document

warehouse/query-core/src/test/java/datawave/query/iterator/logic/AndIteratorTest.java
```diff
@@ -294,7 +296,31 @@ public void testContextWhenNotRequiredSkipLowestMoveLowest() {
...
+        includes.add(getItr(Lists.newArrayList("a", "b", "c"), false));
+        includes.add(getInterruptableItr(Lists.newArrayList("a", "b", "c"), false));
+
+        AndIterator<String> iterator = new AndIterator<>(includes);
+        iterator.initialize();
+
+        Assert.assertTrue(iterator.hasNext());
+        Document d = iterator.document();
+        Assert.assertNull(d);
+    }
```
FineAndDandy: this test is kinda weird. It would return a document if you called next on it.

warehouse/query-core/src/test/java/datawave/query/iterator/logic/AndOrIteratorIT.java
```diff
@@ -0,0 +1,847 @@
...
+
+            driveIntersectionWithSimpleNestedUnion(uidsA, uidsB, uidsC, expectedUids, expectedFieldCounts);
+        }
+    }
+
+    // this test is used to find edge cases
+    // A && (B || !C)
+    @Disabled
+    @Test
+    void testByVolumeIntersectionWithNestedUnionThatContainsANegation() throws IOException {
```
apmoriarty, line 437: This is going to be solved later

warehouse/query-core/src/main/java/datawave/query/iterator/logic/AndIterator.java
ivakegg: Should this be isNonEventField() instead?  non-event fields (which include index-only) require index lookups as well .

apmoriarty: NonEventField is much better than IndexOnly, thanks for pointing that out

warehouse/query-core/src/main/java/datawave/query/iterator/logic/OrIterator.java
```diff
@@ -354,4 +357,35 @@ public boolean isContextRequired() {
...
+
+    @Override
+    public boolean isNonEventField() {
+        return nonEventField;
+    }
+
+    /**
+     * Method to determine if any iterator is for a non-event field
+     */
+    private void determineNonEventState() {
```
FineAndDandy: should this state be updated anytime an iterator is exhausted?

warehouse/query-core/src/main/java/datawave/query/iterator/logic/AndIterator.java
```diff
@@ -527,4 +553,29 @@ public boolean isContextRequired() {
     public void setContext(T context) {
         this.evaluationContext = context;
     }
+
+    @Override
+    public boolean isNonEventField() {
```
FineAndDandy, line 566: should this state only be determined by iterators that aren't exhausted?

warehouse/query-core/src/test/java/datawave/query/iterator/logic/AndIteratorIT.java
FineAndDandy: nevermind, I see you are tracking if its a nonevent field by setting if the document will be built

warehouse/query-core/src/test/java/datawave/query/iterator/logic/IndexIteratorTest.java
```diff
@@ -0,0 +1,254 @@
...
+
+    /**
+     * Drive an iterator and assert document correctness for the field and uids
+     *
+     * @param itr
+     *            the IndexIterator
+     * @param field
+     *            the field
+     * @param uids
+     *            the list of uids
```
FineAndDandy: reverse order of javadoc

FineAndDandy: the driver iterator methods look pretty similar between classes. Can they be condensed?

FineAndDandy (changes requested):
really exceptional job with all the tests

warehouse/query-core/src/main/java/datawave/query/iterator/logic/AndIterator.java
```diff
@@ -256,17 +258,30 @@ public void seek(Range range, Collection<ByteSequence> columnFamilies, boolean i
             try {
                 for (NestedIterator<T> itr : child.leaves()) {
                     if (itr instanceof SeekableIterator) {
-                        ((SeekableIterator) itr).seek(range, columnFamilies, inclusive);
+                        try {
+                            ((SeekableIterator) itr).seek(range, columnFamilies, inclusive);
+                        } catch (Exception e2) {
```
dtspence, line 267: The exception here can potentially be an `IterationInterruptedException`. Is it possible to include specific handling for the exception and/or if appropriate lowering the log-levels?

warehouse/query-core/src/main/java/datawave/query/iterator/NestedIterator.java
apmoriarty: I don't think so. The initial state for every iterator prior to seeking has a null document, so until iterators are seeking they would all flag as index only.

warehouse/query-core/src/main/java/datawave/query/iterator/Util.java
apmoriarty: See previous comment wrt initial state

warehouse/query-core/src/test/java/datawave/query/iterator/logic/AndIteratorTest.java
apmoriarty: This test is more than kinda weird, I don't remember why it was written this way.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/OrIterator.java
apmoriarty: I see what you're saying. If a union contains a non event field the whole union is marked as non-event. The non-event source could be exhausted, and if an exception is thrown on the other tem then the whole query would incorrectly fail.

I'll work something up.

warehouse/query-core/src/test/java/datawave/query/iterator/logic/IndexIteratorTest.java
apmoriarty: Unfortunately not. Each iterator handles assertions slightly differently, so while the code looks similar the important bits are uniquely different.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/AndIterator.java
apmoriarty: What do you have in mind?

warehouse/query-core/src/main/java/datawave/query/iterator/logic/AndIterator.java
apmoriarty: Eventually, es. That change is part of https://github.com/NationalSecurityAgency/datawave/pull/2149. The ripple effect of removing iterators brings too much scope creep into this merge request.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/OrIterator.java
apmoriarty: Documented this case as a test in the next commit.

warehouse/query-core/src/test/java/datawave/query/iterator/logic/AndIteratorTest.java
apmoriarty: Corrected this test

warehouse/query-core/src/main/java/datawave/query/iterator/logic/AndIterator.java
```diff
@@ -379,14 +396,23 @@ protected TreeMultimap<T,NestedIterator<T>> advanceIterators(T key) {
                     highest = transform;
                 }
             } catch (Exception e) {
-                // only need to actually fail if we have nothing left in the AND clause
-                if (includeHeads.isEmpty()) {
-                    throw e;
+                seenException = true;
+                if (itr.isNonEventField()) {
+                    // dropping a non-event term from the query means that the accuracy of the query
```
ivakegg, line 409: Do you mean an event only field here?  We can certainly lookup non-event fields in the index so I am a little confused.  What am I missing?

apmoriarty: We *can* look up non-event or index-only fields in the index, but only under certain cases. 

The DelayedNonEventIndexContext only looks for non-event fields within delayed expressions, which by definition do not contribute iterators to the AndIterator.

The IndexOnlyContextCreator only created an index only context if the query contains the special marker "@LAZY_SET_FOR_INDEX_ONLY_FUNCTION_EVALUATION". 

The case we're dealing with here is `INDEXED_EVENT_FIELD == 'foo' && INDEX_ONLY_FIELD == 'bar'`. If INDEXED_EVENT_FIELD is dropped from the intersection we can still drive the query with INDEX_ONLY_FIELD because we will pick up the INDEXED_EVENT_FIELD while aggregating the event.

However, if we drop the INDEX_ONLY_FIELD there's no way to get that back into the document.

warehouse/query-core/src/test/java/datawave/query/iterator/logic/NegationFilterTest.java
```diff
@@ -176,5 +177,49 @@ public boolean isContextRequired() {
         public void setContext(K context) {
             // no-op
         }
+
+        // this iterator shouldn't need to use the index only flag
```
hlgp: index only != nonEvent

hlgp: there are places where we've conflated these terms and caused bugs.  Double check which you mean and make methods and comments consistent

warehouse/query-core/src/main/java/datawave/query/iterator/logic/AndIterator.java
ivakegg: I am so sorry.  I missed the context if the check here entirely.  You are correct.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/AndIterator.java
ivakegg: I need to update handling of the IteratorionInterruptedException which we explicitly catch in the QueryIterator.  We may want to more gracefully handle that exception here and pass it up prior to checking for the non event field.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/AndIterator.java
```diff
@@ -379,14 +396,23 @@ protected TreeMultimap<T,NestedIterator<T>> advanceIterators(T key) {
                     highest = transform;
                 }
             } catch (Exception e) {
-                // only need to actually fail if we have nothing left in the AND clause
-                if (includeHeads.isEmpty()) {
-                    throw e;
+                seenException = true;
```
ivakegg, line 407: I believe the point made above about the iteration interrupted exception should be handled here as well.

avgAGB (dismissed):

alerman (dismissed):

hlgp (dismissed):

ivakegg (dismissed):

warehouse/query-core/src/test/java/datawave/query/iterator/logic/AndOrIteratorIT.java
FineAndDandy: if this case doesn't have a solution is this code ready for merge?

FineAndDandy (approved):

ivakegg (approved):

PR #2086, opened 2023-09-11
Feature/clean up projection classes for deprecation plan

MiguelRicardos:


cogross (dismissed):

warehouse/query-core/src/main/java/datawave/query/function/DocumentProjection.java
```diff
@@ -38,40 +38,43 @@ public class DocumentProjection implements DocumentPermutation {
     private boolean trackSizes = true;
 
     public DocumentProjection() {
-        this(false, false);
+        this.includeGroupingContext = false;
```
ivakegg: Why not leave this as : this(false, false)

MiguelRicardos: I get a compilation failure when I do it that way.

apmoriarty (approved):

warehouse/query-core/src/main/java/datawave/query/function/DocumentProjection.java
apmoriarty: You can change this constructor to
```java
public DocumentProjection() {
    this(false, false);
}
```
without getting a compilation error

warehouse/query-core/src/main/java/datawave/query/function/DocumentProjection.java
```diff
@@ -38,40 +38,43 @@ public class DocumentProjection implements DocumentPermutation {
...
 
     public DocumentProjection(boolean includeGroupingContext, boolean reducedResponse) {
-        this(includeGroupingContext, reducedResponse, true);
+        this.includeGroupingContext = includeGroupingContext;
+        this.reducedResponse = reducedResponse;
+        this.projection = new Projection();
     }
 
-    public DocumentProjection(boolean includeGroupingContext, boolean reducedResponse, boolean trackSizes) {
+    public DocumentProjection(boolean includeGroupingContext, boolean reducedResponse, boolean trackSizes, Set<String> projections,
```
apmoriarty, line 52: Deprecate the existing constructors.

warehouse/query-core/src/main/java/datawave/query/function/DocumentProjection.java
```diff
@@ -38,40 +38,43 @@ public class DocumentProjection implements DocumentPermutation {
...
         this.projection.setIncludes(includes);
     }
 
-    /**
-     * Configure the delegate {@link Projection} with the fields to exclude
-     *
-     * @param excludes
-     *            the set of fields to exclude
-     */
+    public DocumentProjection(Set<String> projections, Projection.ProjectionType projectionType) {
```
apmoriarty, line 64: please group constructors together at the top of the class.

apmoriarty (changes requested):
None of the changes requested are functional changes, so there's no need to retest.

warehouse/query-core/src/main/java/datawave/query/function/DocumentProjection.java
MiguelRicardos: rgr

warehouse/query-core/src/main/java/datawave/query/function/DocumentProjection.java
MiguelRicardos: you got it.

warehouse/query-core/src/main/java/datawave/query/function/DocumentProjection.java
apmoriarty: This change wasn't made?

MiguelRicardos: Change has been made.

warehouse/query-core/src/main/java/datawave/query/function/DocumentProjection.java
```diff
@@ -37,37 +37,42 @@ public class DocumentProjection implements DocumentPermutation {
...
-     * @param includes
-     *            the set of fields to include
-     */
+    @Deprecated
+    public DocumentProjection(Set<String> projections, Projection.ProjectionType projectionType) {
+        this(false, false, projections, projectionType);
+    }
+
+    @Deprecated
+    public DocumentProjection(boolean includeGroupingContext, boolean reducedResponse, Set<String> projections, Projection.ProjectionType projectionType) {
```
apmoriarty, line 69: Does the new constructor need to be deprecated?

warehouse/query-core/src/main/java/datawave/query/function/DocumentProjection.java
```diff
@@ -37,37 +37,42 @@ public class DocumentProjection implements DocumentPermutation {
...
     }
 
-    /**
-     * Set the delegate {@link Projection} with the fields to include
-     *
-     * @param includes
-     *            the set of fields to include
-     */
+    @Deprecated
+    public DocumentProjection(Set<String> projections, Projection.ProjectionType projectionType) {
```
apmoriarty, line 64: Add documentation that this constructor will not include the grouping context and will not reduce the document.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1208,22 +1210,14 @@ private void debugBooleanLogicIterator(NestedIterator<Key> root, String prefix)
...
+                            Projection.ProjectionType.INCLUDES);
         } else if (this.useBlackListedFields) {
-            // make sure we are not excluding any fields being matched in the limit fields mechanism
-            if (!this.matchingFieldSets.isEmpty()) {
-                this.blackListedFields.removeAll(getMatchingFieldList());
-            }
-            projection.setExcludes(this.blackListedFields);
-            return projection;
+            return new DocumentProjection(this.isIncludeGroupingContext(), this.isReducedResponse(), isTrackSizes(), this.blackListedFields,
+                            Projection.ProjectionType.EXCLUDES);
```
apmoriarty: This change lost the removal of matching field sets. I think a better approach in this case is to build the fields first, then the projection, then the document projection.


And typing this out I realize we actually need to add a constructor for document projection that looks like this
```java
public DocumentProjection(boolean includeGroupingContext, boolean isReducedResponse, boolean isTrackSizes, Projection projection){
   ...
}
```

warehouse/query-core/src/main/java/datawave/query/function/DocumentProjection.java
MiguelRicardos: This constructor is from the parent class that will be removed in the future.

apmoriarty: Perfect, I missed that

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1208,22 +1210,14 @@ private void debugBooleanLogicIterator(NestedIterator<Key> root, String prefix)
...
-        DocumentProjection projection = new DocumentProjection(this.isIncludeGroupingContext(), this.isReducedResponse(), isTrackSizes());
 
         if (this.useWhiteListedFields) {
-            // make sure we include any fields being matched in the limit fields mechanism
-            if (!this.matchingFieldSets.isEmpty()) {
-                this.whiteListedFields.addAll(getMatchingFieldList());
-            }
-            projection.setIncludes(this.whiteListedFields);
-            return projection;
+
```
apmoriarty: Need to add this bit
```java
            // make sure we include any fields being matched in the limit fields mechanism
            if (!this.matchingFieldSets.isEmpty()) {
                this.whiteListedFields.addAll(getMatchingFieldList());
            }
```

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1208,22 +1210,14 @@ private void debugBooleanLogicIterator(NestedIterator<Key> root, String prefix)
...
+            return new DocumentProjection(this.isIncludeGroupingContext(), this.isReducedResponse(), isTrackSizes(),
+                            new Projection(this.whiteListedFields, Projection.ProjectionType.INCLUDES));
         } else if (this.useBlackListedFields) {
-            // make sure we are not excluding any fields being matched in the limit fields mechanism
-            if (!this.matchingFieldSets.isEmpty()) {
-                this.blackListedFields.removeAll(getMatchingFieldList());
-            }
-            projection.setExcludes(this.blackListedFields);
-            return projection;
+            return new DocumentProjection(this.isIncludeGroupingContext(), this.isReducedResponse(), isTrackSizes(),
```
apmoriarty, line 1226: Also need to handle the exclude fields
```java
            // make sure we are not excluding any fields being matched in the limit fields mechanism
            if (!this.matchingFieldSets.isEmpty()) {
                this.blackListedFields.removeAll(getMatchingFieldList());
            }
```

MiguelRicardos: added this back.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
MiguelRicardos: added this back.

apmoriarty (approved):

cogross (approved):

PR #2093, opened 2023-09-15
Updates to try and avoid hanging ivarators

ivakegg:
fixed configuration issues of the thread pools


dtspence (dismissed):

keith-ratcliffe (dismissed):

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -2089,6 +2089,7 @@ protected Future<IteratorSetting> loadQueryIterator(final MetadataHelper metadat
             addOption(cfg, QueryOptions.MAX_EVALUATION_PIPELINES, Integer.toString(config.getMaxEvaluationPipelines()), false);
             addOption(cfg, QueryOptions.MAX_PIPELINE_CACHED_RESULTS, Integer.toString(config.getMaxPipelineCachedResults()), false);
             addOption(cfg, QueryOptions.MAX_IVARATOR_SOURCES, Integer.toString(config.getMaxIvaratorSources()), false);
+            addOption(cfg, QueryOptions.MAX_IVARATOR_SOURCE_WAIT, Long.toString(config.getMaxIvaratorSourceWait()), false);
```
apmoriarty, line 2101: Since the default value is always 30 minutes in the ShardQueryConfig, the QueryOptions, and the Ivarator we only need to add the option if it is a non-default value.

warehouse/query-core/src/test/java/datawave/query/config/ShardQueryConfigurationTest.java
```diff
@@ -516,47 +517,6 @@ public void testIsTldQuery() {
         Assert.assertTrue(config.isTldQuery());
     }
 
-    /**
-     * This test will fail if a new variable is added improperly to the ShardQueryConfiguration
-     * <p>
-     * If a configuration option was added do the following as appropriate
-     * <ol>
-     * <li>Add getter/setters to the ShardQueryLogic that delegate to the ShardQueryConfiguration</li>
```
apmoriarty: This guard should be disabled instead of removed. While it doesn't do much in the way of actual testing it does alert a developer that adding a new config option requires certain steps. 

I'm thinking through what a better solution would look like.

apmoriarty (changes requested):
Looks good so far but a few things need to happen before it's ready to go

- ShardQueryLogic needs getters/setters for the new config option in order for this to be truly configurable via the QueryLogicFactory.xml
- Ideally an illustrative default value would be added to the test query logic factory
- Add a simple test to the ShardQueryConfig that verifies default value and verifies a new value survives a configuration clone operation

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
ivakegg: That would need to be done for all of the properties.  I think I will leave that to a separate PR.

warehouse/query-core/src/test/java/datawave/query/config/ShardQueryConfigurationTest.java
ivakegg: We definitly need a better solution.  We could actually verify that there are setter and getter methods for each of the properties, and could actually verify they are copied in the cloning constructor by using reflection.

ivakegg:
@apmoriarty I created #2112 to hopefully update the tests to enable better verification when properties are added.  Putting the method back in for this PR.  Please re-review.

apmoriarty (dismissed):

properties/default.properties
```diff
@@ -408,6 +408,8 @@ beq.maxTermExpansionThreshold=2000
 beq.fieldIndexRangeSplit=16
 # The max number of sources that can be created across ivarators for one scan
 beq.maxIvaratorSources=20
+# The max wait time in ms for an ivaratory source (default 30 minutes)
```
keith-ratcliffe: Typo nit: `ivaratory`

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -1937,8 +1937,12 @@ public void setMaxIvaratorSources(int maxIvaratorSources) {
         getConfig().setMaxIvaratorSources(maxIvaratorSources);
     }
 
-    public long getMaxIvaratorResults() {
-        return getConfig().getMaxIvaratorResults();
```
keith-ratcliffe: `getMaxIvaratorResults` removed inadvertently?

keith-ratcliffe (changes requested):

apmoriarty (approved):

keith-ratcliffe (approved):

PR #2094, opened 2023-09-19
Enrich 'i' and 'ri' rows in metadata table with event date

### LINKED ISSUE #825: Handle the cases were a field is both indexed and not indexed within a time range

lbschanno:
Modify the generation of 'i' (indexed rows) and 'ri' (reverse indexed rows) in the metadata table such that the column qualifier contains the event date, and the value contains the counts of events for that day. This is required as a first step to support efforts for issue #825 so that we can identify dates when an event was ingested and included in a frequency count for an associated 'f' row, but was not indexed.

**Original format:**
|Row|ColF|ColQ|Value|
|-----|------|------|-----|
|FIELD|i|TYPE|null|
|FIELD|ri|TYPE|null|

**New format:**
|Row|ColF|ColQ|Value|
|-----|-----|-----|-----|
|FIELD|i|TYPE\0yyyyMMdd|COUNT|
|FIELD|ri|TYPE\0yyyyMMdd|COUNT|


ivakegg (dismissed):

avgAGB (dismissed):

apmoriarty (dismissed):

warehouse/ingest-core/src/main/java/datawave/ingest/table/config/MetadataTableConfigHelper.java
```diff
@@ -58,6 +60,26 @@ private String setFrequencyCombiner(TableOperations tops, String scopeName) thro
...
+        setPropertyIfNecessary(tableName, stem + ".opt.lossy", "true", tops, log);
+        setPropertyIfNecessary(tableName, stem + ".opt.type", "VARLEN", tops, log);
+        return stem;
+    }
+
+    // Add the SummingCombiner to the reverse indexed column.
+    private String setReverseIndexCombiner(TableOperations tops, String scopeName) throws AccumuloException, TableNotFoundException, AccumuloSecurityException {
+        String stem = String.format("%s%s.%s", Property.TABLE_ITERATOR_PREFIX, scopeName, "ReverseIndexCombiner");
+        setPropertyIfNecessary(tableName, stem, "12," + SummingCombiner.class.getName(), tops, log);
+        setPropertyIfNecessary(tableName, stem + ".opt.columns", ColumnFamilyConstants.COLF_I.toString(), tops, log);
```
ivakegg: This should be the COLF_RI column family.

lbschanno: Good catch! Fixed.

ivakegg (dismissed):

keith-ratcliffe (dismissed):

keith-ratcliffe (approved):

ivakegg (approved):

PR #2119, opened 2023-10-06
change warning levels for two things

austin007008:
change warning level in iteratorbuildingvisitor from warn to trace and in document from warn to info(just commented out the "else if" for 10MB) 

also other small changes like fixing deprecated code.


MiguelRicardos (approved):
LGTM

ivakegg (approved):

Issue #2118, opened 2023-10-06
Need composite query that runs sequentially


ivakegg:
We need a composite query logics that runs sequentially where the first query to returns results will end the sequence.

PR #2120, opened 2023-10-07
re #2118: composite sequential execution mode

ivakegg:
* Added a sequential execution mode
* Executes the logics in sequential order (logicName order)
* The first logic that returns results will disable the following logics
* Added some debugging in the RemoteHttpService to verify we are closing


keith-ratcliffe (dismissed):

jwomeara:
So when you say "sequential", what you really mean is "sequential AND short circuiting".  If it were just the sequential version of what we already have, then it would run each of the logic to completion, but sequentially as opposed to in parallel as it exists today.  Might be worth making that a bit clearer with the naming/terminology.  I will try to spend more time giving this a proper review, but that's an initial thought.

ivakegg:
changed sequentialExecution to shortCircuitExecution

hgklohr (approved):

keith-ratcliffe (approved):

PR #2140, opened 2023-10-17
re #707: Updated to make use of intermediate results for the Unique

ivakegg:
function


warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -46,7 +47,14 @@ public class UniqueTransform extends DocumentTransform.DefaultDocumentTransform
...
     private Multimap<String,String> modelMapping;
 
-    public UniqueTransform(UniqueFields uniqueFields) {
+    /**
+     * Length of time in milliseconds that a client will wait while results are collected. If a full page is not collected before the timeout, a blank page will
+     * be returned to signal the request is still in progress.
+     */
+    private final long queryExecutionForPageTimeout;
+
+    public UniqueTransform(UniqueFields uniqueFields, long queryExecutionForPageTimeout) {
```
apmoriarty, line 66: Documentation would be nice, with a brief explanation for how the added `queryExecutionForPageTimeout` parameter changes the behavior of the transform

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
ivakegg: done

apmoriarty (approved):

alerman (approved):

PR #2141, opened 2023-10-18
Add visitor that prunes query terms based on ingest types

apmoriarty:


avgAGB (dismissed):

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IngestTypePruningVisitorTest.java
```diff
@@ -0,0 +1,419 @@
...
+        query = "((_Delayed_ = true) && (A == '1' && D == '2'))";
+        test(query, null);
+    }
+
+    @Test
+    public void testDelayedBoundedMarker() {
+        String query = "((_Delayed_ = true) && ((_Bounded_ = true) && (A > '2' && A < '4')))";
+        test(query, query);
+
+        // C term drives pruning of double nested marker
```
hlgp, line 366: I've been reading through this trying to figure out how to set the type we prune by, but this is suggesting we just pick the types for the first term we see?   If so, it would be good to indicate this in the description of the parameter, e.g., "should the query prune fields by the ingest type of the first term we happen to see in the query".

apmoriarty: It's not the first term that drives pruning, it's any term that maps to a mutually exclusive datatype.
```
(A && B && C)
B and C map to datatype 1
A maps to datatype 2
A term drives pruning
```
The order of the terms doesn't mater, datatypes are gathered first and then any potential pruning happens second.

avgAGB (approved):

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IngestTypePruningVisitorTest.java
hlgp: Got it, thanks

hlgp (approved):

PR #2146, opened 2023-10-24
fix enforcer for quickstart

cogross:
- takari-local-repository in .mvn/extensions
 - enforcer up to 3.4.1

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -5,7 +5,11 @@
         <artifactId>takari-smart-builder</artifactId>
         <version>0.6.1</version>
     </extension>
-
+    <extension>
+        <groupId>io.takari.aether</groupId>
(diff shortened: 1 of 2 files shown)


.mvn/extensions.xml
```diff
@@ -5,7 +5,11 @@
         <artifactId>takari-smart-builder</artifactId>
         <version>0.6.1</version>
     </extension>
-
+    <extension>
+        <groupId>io.takari.aether</groupId>
+        <artifactId>takari-local-repository</artifactId>
+        <version>0.11.3</version>
+    </extension>
```
apmoriarty, line 12: This project is archived and indicates that the maven resolver plugin is the successor, is there a specific reason to use the takari project over the maven resolver plugin?

PR #2149, opened 2023-10-26
NestedIterator Bugfix

apmoriarty:
Supercedes #2081 

### High Level Changeset
- dropping a non-event field from an intersection will now appropriately fail a query
- nested iterators now build documents from both the include and context include sources, fixing a bug where non-event fields may never be added to the document, thus causing a missed hit
- Ivarators running with unsorted uids are made slightly more efficient upon a yield, teardown and rebuild
- Added `seek()` and `isNonEventField()` methods to the NestedIterator interface
- Removed `setContext()` and `initialize()` methods from NestedIterator interface 
- NestedIterators are now evaluated in sorted order of Leaf -> Intersection -> Union
- AndIteratorBuilder now allows building an intersection with no includes (an intersection of all excludes)
- IteratorBuildingVisitor now allows building a union of negated terms so long as the parent is not the root
- IteratorBuildingVisitor now allows building an intersection of negated terms so long as the parent is not the root

### NestedIterator Changes
- Cache layer removed from IndexIteratorBridge, AndIterator, and OrIterator
- `move(context)` used to mean "move to next hit, or next highest hit". That implicit method contract has now changed to "does this iterator match the provided context?"
- Junction iterators now handle move operations for each component separately as interpreting the results is different depending on the junction type and source iterator type. For example, the move operation for an include source is the same for both an intersection or a union. The result of that move is handled differently for an intersection vs. a union
- Introduced the concept of expired iterators - these are iterators that do not map to a top key after a seek or move.
- Introduced the concept of exhausted iterators - these iterators have no more elements and this information may be useful for evaluation certain logic (for example, a union with an exhausted exclude source implies the union will always match whatever context is asked for, an intersection with an exhausted include source will never match).

Outstanding work

- [x] more complex unit tests
- [x] removing the initialize method from the NestedIterator interface
- [ ] examine using document status as a clue to non event status
- [ ] special handling of IterationInterruptedException (throw them)


billoley:
1) Is there any way to make AndIterator and OrIterator easier to review.  The diff is incredibly difficult to follow.
2) Is this PR really fixing a single bug?

warehouse/query-core/src/main/java/datawave/query/iterator/logic/AndIterator.java
```diff
@@ -73,220 +107,239 @@ public AndIterator(Iterable<NestedIterator<T>> sources, Iterable<NestedIterator<
...
-        if (!isInitialized()) {
-            throw new IllegalStateException("initialize() was never called");
+    /**
+     * Handles removing an iterator from the underlying collections if an exception occurs.
+     *
+     * @param iterator
+     *            the iterator to remove
+     */
+    private void removeIterator(NestedIterator<T> iterator) {
+        log.error("removing iterator");
```
cogross, line 318: Shouldn't this be a different log level? And maybe include some info to say which iterator is being removed to aid in debugging?

warehouse/query-core/src/test/java/datawave/query/iterator/logic/IndexIteratorBridgeTest.java
```diff
@@ -0,0 +1,216 @@
...
+
+        assertFalse(itr.hasNext());
+        assertEquals(uids, foundUids);
+    }
+
+    private void driveIteratorWithContext(SortedSet<String> uids, SortedSet<String> context) {
+        IndexIteratorBridge itr = createIndexIteratorBridge("FIELD_A", uids, true);
+        SortedSet<String> expected = new TreeSet<>(Sets.intersection(uids, context));
+
+        // log.info("uids : " + uids);
```
cogross, line 140: Should the commented out log.info() calls get removed?

cogross (dismissed):
A few minor comments to address, but seems good.

apmoriarty:
Closing due to inactivity

hlgp:
OBE

PR #2151, opened 2023-10-27
TLDEquality refactored to perform an in-line byte comparison

apmoriarty:
TLDEquality refactored to perform an in-line byte comparison instead of delegating to expensive TLD utility methods

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/function/TLDEquality.java
+++ b/warehouse/query-core/src/main/java/datawave/query/function/TLDEquality.java
@@ -35,6 +35,6 @@ public class TLDEquality implements Equality {
                 return true;
             }
         }
-        return true;
+        return len != 0;
     }
 }
--- a/warehouse/query-core/src/main/java/datawave/query/function/TLDEquality.java
+++ b/warehouse/query-core/src/main/java/datawave/query/function/TLDEquality.java
@@ -3,11 +3,9 @@ package datawave.query.function;
 import org.apache.accumulo.core.data.ByteSequence;
 import org.apache.accumulo.core.data.Key;
 
-import datawave.query.tld.TLD;
-
 /**
  * A key equality implementation that compares to the root pointers of two doc Ids together.
- *
+ * <p>
  * For example, two IDs `h1.h2.h3.a.b.c.d` and `h1.h2.h3.e.f` would be considered equal by this check.
  */
 public class TLDEquality implements Equality {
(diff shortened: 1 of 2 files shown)


warehouse/query-core/src/main/java/datawave/query/function/TLDEquality.java
```diff
@@ -23,8 +21,20 @@ public class TLDEquality implements Equality {
...
+        ByteSequence otherCf = other.getColumnFamilyData();
+
+        int dotCount = 0;
+        int len = Math.min(keyCf.length(), otherCf.length());
+        for (int i = 0; i < len; i++) {
+            byte a = keyCf.byteAt(i);
+            byte b = otherCf.byteAt(i);
+            if (a != b) {
+                return false;
+            } else if (a == '.' && ++dotCount == 3) {
```
MiguelRicardos, line 34: why 3?

MiguelRicardos (dismissed):

warehouse/query-core/src/main/java/datawave/query/function/TLDEquality.java
apmoriarty: A root uid only has two dots, a child uid has three or more. Thus, if a third dot is encountered and everything up to that point has matched, then the root uids are the same.

warehouse/query-core/src/main/java/datawave/query/function/TLDEquality.java
```diff
@@ -23,8 +21,20 @@ public class TLDEquality implements Equality {
...
+                return true;
+            }
+        }
+        return true;
```
d-hwang: maybe this doesn't happen, but this default return will make any comparison to an empty string true.

d-hwang: return len == 0 ? false : true;

??

d-hwang: but then this will make empty value comparisons to false..

apmoriarty: I like that solution. The return statement could be simplified to `return len != 0;`

d-hwang: existing code will also evaluate these two to be the same? a.b.c and a.b.c1
maybe each part of UID has fixed length, but comparing values up to shorter length sounds a bit dangerous to me.

apmoriarty: Given the structure of a uid I don't think that case is possible, but I added a test to document that behavior anyway.

d-hwang (approved):

MiguelRicardos (approved):

PR #2155, opened 2023-11-02
better TLD parsing and clean related classes

austin007008:
Replaced TLD parsing with dedicated code for each method to greatly reduce object creation.

Cleaned some related classes.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/functions/TLDEventFieldAggregator.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/functions/TLDEventFieldAggregator.java
@@ -1,13 +1,10 @@
 package datawave.query.jexl.functions;
 
-import java.util.ArrayList;
+import static datawave.query.tld.TLD.parseDatatypeAndRootUidFromEvent;
 
-import org.apache.accumulo.core.data.ArrayByteSequence;
 import org.apache.accumulo.core.data.ByteSequence;
 
-import datawave.data.hash.UID;
 import datawave.query.predicate.EventDataQueryFilter;
-import datawave.query.tld.TLD;
 import datawave.query.util.TypeMetadata;
 
 public class TLDEventFieldAggregator extends EventFieldAggregator {
@@ -15,20 +12,14 @@ public class TLDEventFieldAggregator extends EventFieldAggregator {
         super(field, filter, maxNextCount, typeMetadata, defaultTypeClass);
     }
 
+    /**
+     * Event Data Key Structure (row, cf='datatype\0uid', cq='field\0value')
+     *
+     * @param cf
+     *            - a ByteSequence representing the Key's ColumnFamily
+     */
     @Override
-    protected ByteSequence parsePointer(ByteSequence columnFamily) {
-        // find the null between the dataType and Uid
-        ArrayList<Integer> nulls = TLD.instancesOf(0, columnFamily, 1);
-        final int start = nulls.get(0) + 1;
-        // uid is from the null byte to the end of the cf
-        ByteSequence uidByteSequence = columnFamily.subSequence(start, columnFamily.length());
-
-        UID uid = UID.parseBase(uidByteSequence.toString());
-        if (uid.getBaseUid().length() != uidByteSequence.length()) {
-            return new ArrayByteSequence(columnFamily.subSequence(0, start) + uid.getBaseUid());
-        }
-
-        // no reduction necessary, already tld
-        return columnFamily;
+    protected ByteSequence parsePointer(ByteSequence cf) {
+        return parseDatatypeAndRootUidFromEvent(cf);
     }
 }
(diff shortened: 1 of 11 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/functions/TLDEventFieldAggregator.java
```diff
@@ -1,34 +1,41 @@
...
-        ArrayList<Integer> nulls = TLD.instancesOf(0, columnFamily, 1);
-        final int start = nulls.get(0) + 1;
-        // uid is from the null byte to the end of the cf
-        ByteSequence uidByteSequence = columnFamily.subSequence(start, columnFamily.length());
-
-        UID uid = UID.parseBase(uidByteSequence.toString());
-        if (uid.getBaseUid().length() != uidByteSequence.length()) {
-            return new ArrayByteSequence(columnFamily.subSequence(0, start) + uid.getBaseUid());
+    protected ByteSequence parsePointer(ByteSequence cf) {
+        int index = 0;
```
apmoriarty: The TLD utility already has a method for `parseRootPointerFromFI(ByteSequence bytes)`, I think we should add a new method `parseRootPointerFromEvent(ByteSequence bytes)` and call that here.

warehouse/query-core/src/test/java/datawave/query/predicate/TLDEventDataFilterTest.java
```diff
@@ -497,7 +495,7 @@ public void functional_ContentExpansionTest() throws ParseException {
         ASTJexlScript newScript = FunctionIndexQueryExpansionVisitor.expandFunctions(config, helper, helper2, script);
 
         Map<String,ExpressionFilter> expressionFilters = getExpressionFilters(newScript, new AttributeFactory(mockTypeMetadata));
-        filter = new TLDEventDataFilter(newScript, Sets.newHashSet("FOO", "BAR"), expressionFilters, null, null, -1, 1);
+        filter = new TLDEventDataFilter(newScript, new HashSet<>(Arrays.asList("FOO", "BAR")), expressionFilters, null, null, -1, 1);
```
apmoriarty: I would have left the `Sets.newHashSet()` calls

warehouse/query-core/src/main/java/datawave/query/tld/TLD.java
```diff
@@ -23,112 +19,202 @@
  * </ul>
  */
 public class TLD {
+    private final static char NULL_BYTE = '\u0000';
+    private final static char DOT = '.';
```
apmoriarty: Sonar lint is suggesting the following
```suggestion
    private static final char NULL_BYTE = '\u0000';
    private static final char DOT = '.';
```

warehouse/query-core/src/test/java/datawave/query/predicate/TLDEventDataFilterTest.java
austin007008: guava says to use the regular java constructor and that "newHashSet()" will likely be deprecated in the future.
Will be using "Set.of" as a hashset is not needed here.

warehouse/query-core/src/main/java/datawave/query/tld/TLD.java
austin007008: done

warehouse/query-core/src/main/java/datawave/query/jexl/functions/TLDEventFieldAggregator.java
austin007008: done with the name "parseDatatypeAndRootUidFromEvent" as we also need the datatype, not just the root uid

apmoriarty (dismissed):

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/TLDEventFieldAggregator.java
```diff
@@ -1,34 +1,25 @@
...
 public class TLDEventFieldAggregator extends EventFieldAggregator {
     public TLDEventFieldAggregator(String field, EventDataQueryFilter filter, int maxNextCount, TypeMetadata typeMetadata, String defaultTypeClass) {
         super(field, filter, maxNextCount, typeMetadata, defaultTypeClass);
     }
 
+    /**
+     * Event Data Key Structure (row, cf='datatype\0uid', cq='field\0value')
+     *
+     * @param cf
+     *            - a ByteSequence representing the Key's ColumnFamily
```
avgAGB, line 19: tickytack, but you could use a @return annotation here for the bytesequence line here

avgAGB (dismissed):

avgAGB (approved):

apmoriarty (approved):

PR #2160, opened 2023-11-06
Remove JMS queue between QueryMetricsBean and QueryMetricsWriter (#2159)

### LINKED ISSUE #2159: Remove JMS queue between QueryMetricsBean and QueryMetricsWriter
After investigating the ongoing periodic timeouts while adding query metrics to the JMS queue and the resulting webserver deadlocks, it appears that there are deadlock issues in the JMS server reattach connection code (not our original issue, but found in testing), and the original timeouts might be due to the large message path through the JMS server.

It seems less error prone at this point to remove the JMS queue from between QueryMetricsBean and QueryMetricWriter and instead inject QueryMetricsWriter into QueryMetricsBean transferring the metrics via a LinkedBlockingQueue.

billoley:
closes #2159


drewfarris:
Seems pretty straightforward. My only question is why did you remove the `remove-datawave-configuration.cli` and `restore-default-wildfly.sh` config/scripts?

billoley:
> Seems pretty straightforward. My only question is why did you remove the `remove-datawave-configuration.cli` and `restore-default-wildfly.sh` config/scripts?

The remove script isn't used anymore and hasn't been consistently updated and tested.  Instead of just removing the like where the QueryMetrics queue is added, I removed the files.

drewfarris (dismissed):

keith-ratcliffe (dismissed):

apmoriarty (dismissed):

hgklohr (approved):

keith-ratcliffe (approved):

Issue #2162, opened 2023-11-07
composite query with shortCircuit should not fail upfront if the remote user operations fails


ivakegg:
The composite query with shortCircuit enabled can have a failure scenario event if the second query logic is not required.  The scenario is that the second logic is a "remote" system that has a separate userOperations mechanism.  Say that remote system is down, then the userOperations will fail as well as the query.  However in "shortCircuit" mode the composite query should succeed if the first query logic is successful.  The problem is that the composite query uses the remote user operations up front when setting up the query and hence fails the entire query in this scenario.

PR #2163, opened 2023-11-07
re #2162: Updated the composite query to be lenient when gathering

ivakegg:
effective authorizations in shortCircuitExecution mode

### DIFF
--- a/web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeUserOperations.java
+++ b/web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeUserOperations.java
@@ -59,7 +59,7 @@ public class CompositeUserOperations implements UserOperations {
                     AuthorizationsListBase.SubjectIssuerDNPair dn = entry.getKey();
                     authMap.put(dn, Sets.union(authMap.containsKey(dn) ? authMap.get(dn) : Collections.emptySet(), entry.getValue()));
                 }
-            } catch (AuthorizationException e) {
+            } catch (Exception e) {
                 // ignore the exception if shortCircuitExecution is specified as we may never even call that remote logic
                 if (!shortCircuitExecution) {
                     throw new AuthorizationException(e);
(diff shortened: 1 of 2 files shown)


web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeUserOperations.java
```diff
@@ -45,14 +49,21 @@ public AuthorizationsListBase listEffectiveAuthorizations(Object callerObject) t
...
+                                remoteAuths.getIssuerDn());
+                authMap.put(userDn, Sets.union(authMap.containsKey(userDn) ? authMap.get(userDn) : Collections.emptySet(), remoteAuths.getAllAuths()));
+                Map<AuthorizationsListBase.SubjectIssuerDNPair,Set<String>> remoteAuthMap = remoteAuths.getAuths();
+                for (Map.Entry<AuthorizationsListBase.SubjectIssuerDNPair,Set<String>> entry : remoteAuthMap.entrySet()) {
+                    AuthorizationsListBase.SubjectIssuerDNPair dn = entry.getKey();
+                    authMap.put(dn, Sets.union(authMap.containsKey(dn) ? authMap.get(dn) : Collections.emptySet(), entry.getValue()));
+                }
+            } catch (AuthorizationException e) {
+                // ignore the exception if shortCircuitExecution is specified as we may never even call that remote logic
+                if (!shortCircuitExecution) {
```
apmoriarty, line 64: I'm curious if we could throw an AuthorizationException for a non-remote user operations object. Should this be
```java
if (!shortCircuitExecution && op instanceof RemoteUserOperations) {
```

ivakegg: I don't think it needs to be exclusively a RemoteUserOperations.  We could have other mechanisms or implementations that need the same handling from the composite user operations standpoint.

apmoriarty (approved):

hlgp (approved):

PR #2164, opened 2023-11-07
Add option to not add VersioningIterator 

avgAGB:


warehouse/ingest-core/src/main/java/datawave/ingest/table/config/AbstractTableConfigHelper.java
```diff
@@ -67,6 +67,14 @@ public static void setPropertyIfNecessary(String tableName, String propertyName,
         }
     }
 
+    public static void disableVersioningIfNecessary(TableOperations tops, Logger log) throws AccumuloException, AccumuloSecurityException {
+
+        for (IteratorScope iterScope : IteratorScope.values()) {
+            tops.removeProperty(Property.TABLE_ITERATOR_PREFIX + iterScope.name() + ".vers", "20," + VersioningIterator.class.getName());
```
jschmidt10: The API says that the first argument is the tableName and the 2nd is the property name. You don't need to specify the value of the property to remove.

warehouse/ingest-core/src/main/java/datawave/ingest/table/config/AbstractTableConfigHelper.java
```diff
@@ -67,6 +67,14 @@ public static void setPropertyIfNecessary(String tableName, String propertyName,
         }
     }
 
+    public static void disableVersioningIfNecessary(TableOperations tops, Logger log) throws AccumuloException, AccumuloSecurityException {
```
jschmidt10: The log argument appears to be unused. I'd remove it if we don't need it.

warehouse/ingest-core/src/main/java/datawave/ingest/table/config/AbstractTableConfigHelper.java
jschmidt10: It looks like Accumulo has a nicer API for this so you don't have to hard-code the options to remove. Check out `TableOperatoins.removeIterator()`.

warehouse/ingest-core/src/main/java/datawave/ingest/table/config/AbstractTableConfigHelper.java
avgAGB: Thanks, will do

warehouse/ingest-core/src/main/java/datawave/ingest/table/config/AbstractTableConfigHelper.java
avgAGB: Thanks for the heads up. I'll adjust this and test with the new call.

PR #2165, opened 2023-11-09
2013 - jexl uuid query parser research/changes

cogross:
```
In [20806624] LuceneToJexlUUIDQueryParser.java#L15, the LuceneToJexlUUIDQueryParser was marked deprecated, but no suggested replacement is indicated. It is used in the current QueryLogicFactory.xml#L295 . 
If this class isn't really needed anymore, we should update QueryLogicFactory.xml with an alternate configuration that replaces it.  If the class is still needed, we should remove the @Deprecated tag.

```
In testing the other classes that implement the QueryParser, I determined that no existing ones can replace the "Deprecared" LuceneQueryParser. Removed the @Deprecated tags.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/LuceneToJexlUUIDQueryParser.java
+++ b/warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/LuceneToJexlUUIDQueryParser.java
@@ -12,7 +12,6 @@ import datawave.query.search.FieldedTerm;
 import datawave.query.search.RangeFieldedTerm;
 import datawave.query.search.WildcardFieldedTerm;
 
-@Deprecated
 public class LuceneToJexlUUIDQueryParser extends LuceneToJexlQueryParser {
     private List<UUIDType> uuidTypes = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


avgAGB:
Curious question, is it possible to just pull out all of the beans that deal with the LuceneToJexlUUIDQueryParser in the QueryLogicFactory as well, or does that end up breaking stuff up?

cogross:
> Curious question, is it possible to just pull out all of the beans that deal with the LuceneToJexlUUIDQueryParser in the QueryLogicFactory as well, or does that end up breaking stuff up?

I wasn't able to find any Parsers to sub it with successfully, so I also assumed that removing references would mess things up.  But i gave it a try.  I cut out the LUCENE-UUID lines from the configs and whatnot, and it did blow up for the `datawaveWebTest`s.  So, I think that de-`@Deprecate`ing is the right approach for now.

avgAGB (approved):

matthpeterson (approved):

matthpeterson (approved):

apmoriarty (approved):

PR #2166, opened 2023-11-14
1285 - user check for start/stop scripts

cogross:
Added a check to ensure the datawaveStart/Stop scripts aren't run as root.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -136,6 +136,7 @@ jobs:
         DW_DATAWAVE_BUILD_COMMAND: "mvn -B -V -e -ntp -Dstyle.color=always -Pdev -Ddeploy -Dtar -DskipTests clean package"
         DOCKER_BUILD_OPTS: "--squash --force-rm"
       run: |
+        su - datawave
         TAG=$(mvn -q -N -Dexec.executable='echo' -Dexec.args='${project.version}' exec:exec)
         contrib/datawave-quickstart/docker/docker-build.sh ${TAG} --docker-opts "${DOCKER_BUILD_OPTS}"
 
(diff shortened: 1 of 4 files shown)


keith-ratcliffe (changes requested):
These changes break the 'quickstart-build-and-test' job in our github actions workflow, because the root user is used in the docker context. But I'm also confused about why this check is necessary for the quickstart in the first place.  The user that bootstraps the quickstart environment, root or other, is going to be the user that runs _all_ of the services, not just datawave web and datawave ingest. That is, if user 'foo' bootstraps the quickstart, then the root user won't even have the **datawaveIngestStart** and **datawaveWebStart** functions defined in it's environment in the first place

If the goal is just to limit datawave services, then I think a global check in [ingest-env.sh](https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/ingest-scripts/src/main/resources/bin/ingest/ingest-env.sh) is probably in order for ingest, and a similar check in [setup-wildfly.sh](https://github.com/NationalSecurityAgency/datawave/blob/integration/web-services/deploy/application/src/main/wildfly/setup-wildfly.sh) for the webservice (and/or something that dynamically updates wildfly's standalone.sh to perform the check)

But those would still constitute breaking changes for 'quickstart-build-and-test' as it currently stands, so some sort of override capability would be needed for the quickstart, or else a change in how we build/run the docker image.

keith-ratcliffe:
Is the main goal of this work to prevent the user from _accidentally_ running the DW start/stop scripts as root?

cogross:
> Is the main goal of this work to prevent the user from _accidentally_ running the DW start/stop scripts as root?

I believe that was the goal.  There weren't many details in the ticket.

PR #2168, opened 2023-11-16
updated bootstrap scripts

cbevard1:
The bootstrap scripts weren't correctly identifying the state of the hadoop, accumulo, zookeeper deployments and causing issues with hadoopStop, hadoopStart, accumuloStop, accumuloStart.

This changes fixes those issues.


cbevard1:
Also, can I get commit access to the repo? I had to fork in order to push much changes to github.

contrib/datawave-quickstart/bin/services/accumulo/bootstrap.sh
```diff
@@ -129,13 +126,20 @@ function accumuloStart() {
        hadoopStart
        echo
     fi
-    eval "${DW_ACCUMULO_CMD_START}"
+    if ! accumuloIsRunning ; then
+       eval "${DW_ACCUMULO_CMD_START}"
+       echo
```
MiguelRicardos, line 131: what is the extra echo for?

contrib/datawave-quickstart/bin/services/accumulo/bootstrap.sh
```diff
@@ -104,22 +104,19 @@ export PATH=${ACCUMULO_HOME}/bin:${ZOOKEEPER_HOME}/bin:$PATH
...
-
-    zookeeperIsRunning
-
-    [[ -z "${DW_ACCUMULO_PID_LIST}" && -z "${DW_ZOOKEEPER_PID_LIST}" ]] && return 1 || return 0
+    DW_ACCUMULO_PID_COUNT="$(eval "${DW_ACCUMULO_CMD_FIND_ALL_PIDS_COUNT}")"
+    [ "${DW_ACCUMULO_PID_COUNT}" -eq 4 ] && return 0 || return 1
 }
 
 function accumuloStart() {
-    accumuloIsRunning && echo "Accumulo is already running" && return 1
```
MiguelRicardos, line 122: Don't we want it to fail if accumulo is already running?

contrib/datawave-quickstart/bin/services/accumulo/bootstrap.sh
```diff
@@ -104,22 +104,20 @@ export PATH=${ACCUMULO_HOME}/bin:${ZOOKEEPER_HOME}/bin:$PATH
 
 DW_ZOOKEEPER_CMD_START="( cd ${ZOOKEEPER_HOME}/bin && ./zkServer.sh start )"
 DW_ZOOKEEPER_CMD_STOP="( cd ${ZOOKEEPER_HOME}/bin && ./zkServer.sh stop )"
-DW_ZOOKEEPER_CMD_FIND_ALL_PIDS="ps -ef | grep 'zookeeper.server.quorum.QuorumPeerMain' | grep -v grep | awk '{ print \$2 }'"
+DW_ZOOKEEPER_CMD_FIND_ALL_PIDS="ps -ef | grep 'zookeeper.server.quorum.QuorumPeerMain' | grep -v grep"
```
keith-ratcliffe: The command should output the pid of the ZK server, if present, and nothing else. This change causes the entire output of ```ps -ef``` for the ZK process to be stored in the variable, which is incorrect

contrib/datawave-quickstart/bin/services/accumulo/bootstrap.sh
```diff
@@ -274,15 +283,6 @@ function accumuloPrintenv() {
...
 }
 
-function accumuloPidList() {
-   # Refresh pid lists
-   accumuloIsRunning
-   zookeeperIsRunning
-   if [[ -n "${DW_ACCUMULO_PID_LIST}" || -n "${DW_ZOOKEEPER_PID_LIST}" ]] ; then
-      echo "${DW_ACCUMULO_PID_LIST} ${DW_ZOOKEEPER_PID_LIST}"
-   fi
-}
```
keith-ratcliffe: This removal, combined with the other changes here, breaks ```allStop --hard``` [here](https://github.com/NationalSecurityAgency/datawave/blob/integration/contrib/datawave-quickstart/bin/common.sh#L163)

contrib/datawave-quickstart/bin/services/hadoop/bootstrap.sh
```diff
@@ -95,11 +95,12 @@ export PATH=${HADOOP_HOME}/bin:$PATH
 
 DW_HADOOP_CMD_START="( cd ${HADOOP_HOME}/sbin && ./start-dfs.sh && ./start-yarn.sh && mapred --daemon start historyserver )"
 DW_HADOOP_CMD_STOP="( cd ${HADOOP_HOME}/sbin && mapred --daemon stop historyserver && ./stop-yarn.sh && ./stop-dfs.sh )"
-DW_HADOOP_CMD_FIND_ALL_PIDS="pgrep -d ' ' -f 'proc_datanode|proc_namenode|proc_secondarynamenode|proc_nodemanager|proc_resourcemanager|mapreduce.v2.hs.JobHistoryServer'"
+DW_HADOOP_CMD_FIND_ALL_PIDS_COUNT="pgrep -f 'proc_datanode|proc_namenode|proc_secondarynamenode|proc_nodemanager|proc_resourcemanager|mapreduce.v2.hs.JobHistoryServer' | wc -l"
```
keith-ratcliffe: Same as above (i.e., breaks ```allStop --hard```)

contrib/datawave-quickstart/bin/services/accumulo/bootstrap.sh
```diff
@@ -104,22 +104,20 @@ export PATH=${ACCUMULO_HOME}/bin:${ZOOKEEPER_HOME}/bin:$PATH
...
+DW_ACCUMULO_CMD_FIND_ALL_PIDS_COUNT="pgrep -f 'o.start.Main manager|o.start.Main tserver|o.start.Main monitor|o.start.Main gc' | wc -l"
 
 function accumuloIsRunning() {
-    DW_ACCUMULO_PID_LIST="$(eval "${DW_ACCUMULO_CMD_FIND_ALL_PIDS}")"
-
-    zookeeperIsRunning
-
-    [[ -z "${DW_ACCUMULO_PID_LIST}" && -z "${DW_ZOOKEEPER_PID_LIST}" ]] && return 1 || return 0
+    DW_ACCUMULO_PID_COUNT="$(eval "${DW_ACCUMULO_CMD_FIND_ALL_PIDS_COUNT}")"
+    [ "${DW_ACCUMULO_PID_COUNT}" -eq 4 ] && return 0 || return 1
```
keith-ratcliffe: I feel like it's probably not a good idea to have running/stopped status depend on the presence of a specific number of pids. The original goal was that the presence of any pids for a service should signal 'running' status, and for the pids themselves to be collected uniformly into a ```*_PID_LIST```  variable, so that you can easily ```kill -9``` them and/or report them to stdout on demand, if needed

That's not to say what's currently here is bulletproof by any means.

Could you provide more detail about the specific issue(s) you encountered when trying to start/stop things? That would help me better understand how we might fix it. 

One thing I've considered is breaking out Zookeeper as its own top-level service, rather than have it be embedded with Accumulo. That could go a long way toward simplifying things and probably squash some existing bugs here.

contrib/datawave-quickstart/bin/services/hadoop/bootstrap.sh
```diff
@@ -95,11 +95,12 @@ export PATH=${HADOOP_HOME}/bin:$PATH
...
+
 
 function hadoopIsRunning() {
-    DW_HADOOP_PID_LIST="$(eval "${DW_HADOOP_CMD_FIND_ALL_PIDS}")"
-    [ -z "${DW_HADOOP_PID_LIST}" ] && return 1 || return 0
+    DW_HADOOP_PID_COUNT="$(eval "${DW_HADOOP_CMD_FIND_ALL_PIDS_COUNT}")"
+    [ "${DW_HADOOP_PID_COUNT}" -eq 5 ] && return 0 || return 1
```
keith-ratcliffe: Same as above regarding reliance on specific pid count

keith-ratcliffe (changes requested):

MiguelRicardos:
can you make sure to run a ```mvn formatter:format``` to make the first stage happy? Also please rebase and squash your commits.

contrib/datawave-quickstart/bin/services/accumulo/bootstrap.sh
```diff
@@ -109,17 +109,15 @@ DW_ZOOKEEPER_CMD_FIND_ALL_PIDS="ps -ef | grep 'zookeeper.server.quorum.QuorumPee
 DW_ACCUMULO_CMD_START="( cd ${ACCUMULO_HOME}/bin && ./accumulo-cluster start )"
 DW_ACCUMULO_CMD_STOP="( cd ${ACCUMULO_HOME}/bin && ./accumulo-cluster stop )"
 DW_ACCUMULO_CMD_FIND_ALL_PIDS="pgrep -d ' ' -f 'o.start.Main manager|o.start.Main tserver|o.start.Main monitor|o.start.Main gc|o.start.Main tracer'"
+DW_ACCUMULO_CMD_FIND_ALL_PIDS_COUNT="pgrep -f 'o.start.Main manager|o.start.Main tserver|o.start.Main monitor|o.start.Main gc' | wc -l"
```
keith-ratcliffe: If the goal is to allow start commands to be retried on a partially-started Accumulo or Hadoop, then I think I see the benefit of the pid counting logic, since that could definitely save the user some time.

But rather than duplicating the ```pgrep -f '...'``` command just to determine the count via ```wc -l```, it would be better just to leverage the existing ```*PidList``` functions, so that we don't incur the extra maintenance cost of keeping both commands in sync over time.

For example, something like this could suffice instead...
```
...
# Convert pid list string into local array
local pidArray=( $(accumuloPidList) )

[[ ${#pidArray[@]} -eq 4 ]] && return 0 || return 1
...
```

keith-ratcliffe (changes requested):

contrib/datawave-quickstart/bin/services/hadoop/bootstrap.sh
```diff
@@ -129,6 +129,7 @@ function hadoopStatus() {
         local _pid
         local _opt=pid
 
+        DW_HADOOP_PID_LIST="$(eval "${DW_HADOOP_CMD_FIND_ALL_PIDS}")"
```
keith-ratcliffe, line 132: Sorry....last issue, I think:

I noticed that this function still requires ```hadoopIsRunning```  (line 128) to return success in order to begin evaluating the status of each process in the loop. So, even if there are some processes running, the status printed out will be... 
```
[DW_WARN] - X is not running
```
...for all of them b/c the pid count is less than 5

So, this line just needs to be moved up with ```test -n ...``` added, exactly like your fix for ```accumuloStatus``` above.
```
...
L128      DW_HADOOP_PID_LIST="$(eval "${DW_HADOOP_CMD_FIND_ALL_PIDS}")"
L129      test -n "${DW_HADOOP_PID_LIST}" && {
```

keith-ratcliffe (changes requested):

hlgp:
Closing, deferred.  Reopen when revived.

PR #2183, opened 2023-12-06
Fix Marker Pruning

apmoriarty:
The IngestTypePruningVisitor doesn't prune the whole marker node in certain cases. This merge request addresses that.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypePruningVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypePruningVisitor.java
@@ -241,7 +241,6 @@ public class IngestTypePruningVisitor extends BaseVisitor {
         Set<String> dts = (Set<String>) visit(source, data);
 
         if (source.jjtGetParent() == null) {
-            pruneNodeFromParent(source);
             pruneNodeFromParent(node);
         }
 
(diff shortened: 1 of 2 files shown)


FineAndDandy (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypePruningVisitor.java
```diff
@@ -240,7 +240,7 @@ private Set<String> visitMarker(QueryPropertyMarker.Instance instance, JexlNode
         JexlNode source = node.jjtGetChild(1);
         Set<String> dts = (Set<String>) visit(source, data);
 
-        if (source.jjtGetNumChildren() == 0) {
+        if (source.jjtGetParent() == null) {
             pruneNodeFromParent(source);
```
ivakegg: Do we need this and would if fail?

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypePruningVisitor.java
apmoriarty: We do not need this extra method call. It won't fail, there's a safeguard that does an early null check.

FineAndDandy (approved):

ivakegg (approved):

PR #2185, opened 2023-12-07
Add hit term call out in excerpt fields

mineralntl:


apmoriarty (approved):
LGTM

alerman (approved):
Id consider changing the property negation. I dont love that a value of true for excerpt.fields.no.hit.callout results in something getting excluded, but I can live with it.

PR #2186, opened 2023-12-07
Feature/filter user ops

ivakegg:


apmoriarty (dismissed):

web-services/query/src/main/java/datawave/webservice/query/util/LookupUUIDUtil.java
```diff
@@ -418,7 +419,7 @@ public <T> T createUUIDQueryAndNext(final AbstractUUIDLookupCriteria unvalidated
         return response;
     }
     
-    private String getAuths(String logicName, String queryAuths, Principal principal) {
+    private String getAuths(String logicName, MultivaluedMap<String,String> queryParameters, String queryAuths, Principal principal) {
```
d-hwang, line 435: Hmm, I don't see how queryParameters are being used. Were you going to add a logic to change the user operation behavior based on certain query parameter?

ivakegg: You are correct, I was unable to figure out how to create the Query object easily from the queryParameters to modify this section.  Let me take a look to see if I can do that.  Otherwise I will remove it.

ivakegg: ok, updated getAuths appropriately.

web-services/query/src/main/java/datawave/webservice/query/util/LookupUUIDUtil.java
```diff
@@ -418,10 +419,27 @@ public <T> T createUUIDQueryAndNext(final AbstractUUIDLookupCriteria unvalidated
         return response;
     }
     
-    private String getAuths(String logicName, String queryAuths, Principal principal) {
+    private Query createSettings(MultivaluedMap<String,String> queryParameters) {
+        Query query = responseObjectFactory.getQueryImpl();
+        query.setOptionalQueryParameters(queryParameters);
+        for (String key : queryParameters.keySet()) {
```
d-hwang: I see that the mock returns emptySet for testing, but should we add a null check just in case?

ivakegg: sure... ok hold on

ivakegg: done

apmoriarty (approved):

d-hwang (approved):

PR #2187, opened 2023-12-08
Feature/composite timestamp

hlgp:


warehouse/core/src/main/java/datawave/util/CompositeTimestamp.java
```diff
@@ -0,0 +1,42 @@
+package datawave.util;
+
+public class CompositeTimestamp {
```
ivakegg, line 13: A class description would be good that describes how the long value is split into two values.

warehouse/core/src/main/java/datawave/util/CompositeTimestamp.java
```diff
@@ -0,0 +1,42 @@
...
+        } else {
+            return ts;
+        }
+    }
+
+    public static long getCompositeTimeStamp(long eventDate, long ageOffDate) {
+        if (ageOffDate < eventDate) {
+            throw new IllegalArgumentException("age off date must be greater than the event date");
+        }
+        long diffInDays = (ageOffDate - eventDate) / MILLIS_PER_DAY;
```
ivakegg: Should we round up ?

warehouse/core/src/main/java/datawave/util/CompositeTimestamp.java
```diff
@@ -0,0 +1,42 @@
...
+        long compositeTS = (diffInDays << allocationForEventDate) | (eventDate & mask);
```
ivakegg: Can we check bounds on the diffInDays and the eventDate?

warehouse/core/src/main/java/datawave/util/CompositeTimestamp.java
```diff
@@ -0,0 +1,42 @@
...
+        if (isCompositeTimestamp(ts)) {
+            return (ts & mask);
+        } else {
+            return ts;
+        }
+    }
+
+    public static long getAgeOffDate(long ts) {
+        if (isCompositeTimestamp(ts)) {
+            return ((ts >> allocationForEventDate) * MILLIS_PER_DAY) + (ts & mask);
```
ivakegg: I think we want to do an unsigned right shift here (>>>)

warehouse/core/src/main/java/datawave/util/CompositeTimestamp.java
```diff
@@ -0,0 +1,42 @@
+package datawave.util;
+
+public class CompositeTimestamp {
+
+    private static final int allocationForEventDate = 47;
+    private static final long mask = 0x7fffffffffffL;
+    public static final long MILLIS_PER_DAY = 1000 * 60 * 60 * 24;
+
+    public static boolean isCompositeTimestamp(long ts) {
+        if (ts >> allocationForEventDate > 0) {
```
ivakegg: I think we want an unsigned right shift here (>>>)

warehouse/core/src/test/java/datawave/util/CompositeTimestampTest.java
```diff
@@ -0,0 +1,27 @@
+package datawave.util;
+
+import java.time.Instant;
+import java.time.format.DateTimeFormatter;
+
+import org.junit.Test;
+import org.locationtech.jts.util.Assert;
+
+public class CompositeTimestampTest {
```
ivakegg, line 17: Lets create some tests that test either side of the bounds for the event data and the ageoff date

warehouse/core/src/main/java/datawave/util/CompositeTimestamp.java
ivakegg: After seeing the test case that fails in https://github.com/NationalSecurityAgency/datawave/pull/2268, it made me realize that we cannot handle event dates prior to epoch.  Do we want treat the 47bit event date as a signed value ?

ivakegg (approved):

warehouse/core/src/main/java/datawave/util/CompositeTimestamp.java
```diff
@@ -0,0 +1,212 @@
...
+    private static final long mask = -1L >>> (8 * 8 - allocationForEventDate);
+    // The max age off delta in days
+    private static final long maxDiff = (-1L << (allocationForEventDate + 1)) >>> (allocationForEventDate + 1);
+    // A useful constant
+    public static final long MILLIS_PER_DAY = 1000 * 60 * 60 * 24;
+
+    // A constant for an invalid timestamp
+    public static final long INVALID_TIMESTAMP = Long.MIN_VALUE;
+    // The minimum event date
+    public static final long MIN_EVENT_DATE = (0 - mask); // also equivalent to Long.MIN_VALUE + 1
```
apmoriarty, line 27: Is this comment correct? If so, why not just initialize the min event date to Long.MIN_VALUE + 1?

warehouse/core/src/main/java/datawave/util/CompositeTimestamp.java
```diff
@@ -0,0 +1,212 @@
...
+ * bits. The last 17 bits (except for the sign bit) will be used to encode how many days after the event date that we should base the ageoff on. If the
+ * timestamp is negative, then to calculate the values the complement is taken and then the two values are extracted. The ageoff is encoded as an age off delta
+ * which is the number of days after the event date.
+ */
+public class CompositeTimestamp {
+
+    // The number of bits for the event date
+    private static final int allocationForEventDate = 46;
+    // A mask for the event date
+    private static final long mask = -1L >>> (8 * 8 - allocationForEventDate);
```
apmoriarty, line 18: Why 8*8 instead of 64?

warehouse/core/src/test/java/datawave/util/CompositeTimestampTest.java
```diff
@@ -0,0 +1,371 @@
...
+    public void testPostDoomsday() {
+        long eventDate = Instant.from(DateTimeFormatter.ISO_INSTANT.parse("4199-11-25T00:00:00Z")).toEpochMilli();
+        long ageOff = Instant.from(DateTimeFormatter.ISO_INSTANT.parse("4299-06-01T00:00:00Z")).toEpochMilli();
+
+        try {
+            CompositeTimestamp.getCompositeTimeStamp(eventDate, ageOff);
+            Assert.fail("Expected event date to fail");
+        } catch (IllegalArgumentException e) {
+            // expected
+        }
```
apmoriarty, line 205: Junit5 allows you to do `assertThrows(Exception, () -> CompositeTimeStamp.dostuff())`

apmoriarty (approved):
LGTM. 

The changes to the global index uid aggregator warrants a write up in the merge request description.

warehouse/core/src/main/java/datawave/util/CompositeTimestamp.java
ivakegg: in case we change the mask

warehouse/core/src/main/java/datawave/util/CompositeTimestamp.java
ivakegg: Letting the compiler work for it?  8 bits * 8 bytes.

warehouse/core/src/test/java/datawave/util/CompositeTimestampTest.java
ivakegg: Cool, I will use that next time.

PR #2189, opened 2023-12-08
Use multiple threads to write query metric updates from QueryMetricsWriter to the query metric microservice

billoley:


web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsWriter.java
```diff
@@ -128,7 +135,7 @@ public void onRefresh(@Observes RefreshEvent event, BeanManager bm) {
 
     public void addMetricToQueue(QueryMetricHolder queryMetricHolder) {
         try {
-            if (shuttingDown) {
+            if (shutDownQueue) {
                 log.error(String.format("shutting down - dropping metric for queryId %s", queryMetricHolder.getQueryMetric().getQueryId()));
             } else if (!blockingQueue.offer(queryMetricHolder)) {
                 if (queryMetricHolder.getQueryMetric().getLifecycle().compareTo(Lifecycle.INITIALIZED) <= 0) {
```
ivakegg, line 171: Just curious why we do not log dropping a metric even when we are not yet initialized?

ivakegg (commented):
The code looks good as far as I can tell.  I would like at least some function level documentation.

web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsWriter.java
billoley: 1) During shutdown of the W/S, shutDownQueue gets set to true first to prevent any more metrics from being added.  At that point, log metric updates that are dropped.  If all queries were completed, there shouldn't be any.

2) The else if clause is for if we fail adding a metric update to the queue (queue is full).  This PR and changes to the metric service are meant to prevent this situation, but if it happens (again) then we would print a log message for every query event (create, initialized, every page, closed, etc).  The if statement to only log a metric update when the Lifecycle is <= INITIALIZED (i.e. CREATED, INITIALIZED) is to limit the flood of logs while still alerting us that the queue is full and we are dropping metrics.

I'm open to suggestions, but the hope is that in adding threads on the W/S side to hit http receiver threads on the metric service side that we avoid this happening.

billoley:
> The code looks good as far as I can tell. I would like at least some function level documentation.

Just pushed a commit with some documentation

web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsWriter.java
```diff
@@ -200,21 +273,36 @@ public long getFailuresWhenOthersSucceeded() {
...
+        private LinkedBlockingQueue<QueryMetricHolder> blockingQueue;
+
+        public MetricProcessor(LinkedBlockingQueue<QueryMetricHolder> blockingQueue) {
+            this.blockingQueue = blockingQueue;
+        }
+
         @Override
         public void run() {
-            while (!shuttingDown) {
+            while (!shutDownMetricProcessors) {
```
jwomeara, line 299: So this loop is going to run forever, as fast as the processor will allow it to run, whether there are metrics or not?  This concerns me - it's going to eat up a lot of CPU, isn't it?  At the bare minimum, we should Thread.sleep for a bit if there are no metrics in the queue.  

Alternatively, you could `poll` the queue, and give a timeout for the poll call.  If there's a metric available to write, it will do so immediately, if not, it will wait the specified amount of time.  Calling `take()` would wait indefinitely until a metric is available, which might be even better in this case.

web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsWriter.java
```diff
@@ -225,236 +313,284 @@ public void run() {
...
+            List<BaseQueryMetric> metricList = metricHolderList.stream().map(QueryMetricHolder::getQueryMetric).collect(Collectors.toList());
+            if (!metricList.isEmpty()) {
+                try {
+                    writeMetricsToRemoteService(metricList);
+                    log.debug(String.format("wrote %d metric updates to RemoteQueryMetricService", metricList.size()));
+                } catch (Exception e) {
+                    log.error(String.format("%d metric updates failed write to RemoteQueryMetricService as a batch, will retry individually - %s",
+                                    metricList.size(), e.getMessage()), e);
+                    metricList.forEach(m -> {
+                        failedMetrics.add(new FailureRecord(m, false));
```
jwomeara, line 351: Why have failed metrics at all?  Why not just put the failed metric back into the blocking queue?

jwomeara (changes requested):
Only took a brief glance, but this ended up being more complicated than I expected it to be.  I'm not sure I understand the need for a separate queue for failed metrics, and I think the runaway while loop in the QueryMetricsWriter could be a problem.

ivakegg (approved):

web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsWriter.java
jwomeara: I see that I didn't read close enough...  The poll is in the getMetricsFromQueue method, which existed prior to this PR, so was not obvious to me.  Disregard this comment.

billoley: That loop calls getMetricsFromQueue which calls poll with a timeout and loops until batchSize or maxLatency  is reached.

web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsWriter.java
billoley: To track whether the failure is because of the metric itself or a system failure.  The data stored in FailureRecord helps weather and recover from metric-based, intermittent, and long-running failures

billoley: Also -- the primary reason for this PR is to have multiple MetricProcessor threads reading off of the queue for the remote service case.  The other functionality is unchanged.

jwomeara (approved):

PR #2192, opened 2023-12-13
Feature/query microservices

jwomeara:


contrib/datawave-quickstart/bin/modification.sh
```diff
@@ -0,0 +1,295 @@
...
+   else
+       error "Command set to ${DW_MODIFICATION_COMMAND}.  Command must be one of INSERT, UPDATE, DELETE, or REPLACE." && return 1
+   fi
+
+   DW_CURL_DATA="-d '$BODY'"
+}
+
+function modificationHelp() {
+    echo
+    echo " The $( printGreen "datawaveModification" ) shell function allows you submit modification requestson demand to DataWave's"
```
foster33:
```suggestion
    echo " The $( printGreen "datawaveModification" ) shell function allows you submit modification requests on demand to DataWave's"
```
Minor typo

.github/workflows/tests.yml
```diff
@@ -73,55 +76,57 @@ jobs:
...
-  #         USER_NAME: ${{ secrets.USER_NAME }}
-  #         ACCESS_TOKEN: ${{ secrets.ACCESS_TOKEN }}
-  #       run: |
-  #         RUN_TESTS="mvn -s $GITHUB_WORKSPACE/.github/workflows/settings.xml -V -B -e -Dmaven.build.cache.enabled=false verify"
-  #         cd microservices
-  #         $RUN_TESTS \
-  #           || { echo "***** TESTS FAILED. Attempting retry."; $RUN_TESTS; } \
-  #           || { echo "***** TESTS FAILED. Attempting final retry."; $RUN_TESTS; }
+  # Build the code and run the microservice unit/integration tests.
+  build-and-test-microservices:
```
alerman: Is the intention to leave this uncommented moving forward?

.github/workflows/tests.yml
```diff
@@ -142,14 +147,92 @@ jobs:
...
-        DW_DATAWAVE_BUILD_COMMAND: "mvn -s $GITHUB_WORKSPACE/.github/workflows/settings.xml -B -V -e -ntp -Dstyle.color=always -Dmaven.build.cache.enabled=false -Pdev -Ddeploy -Dtar -DskipTests clean package"
+        DW_DATAWAVE_BUILD_COMMAND: "mvn -s $GITHUB_WORKSPACE/.github/workflows/settings.xml -B -V -e -ntp -Dstyle.color=always -Dmaven.build.cache.enabled=false -Pdev -DskipServices -Ddeploy -Dtar -DskipTests -pl -:config-service clean package"
         DOCKER_BUILD_OPTS: "--squash --force-rm"
         USER_NAME: ${{ secrets.USER_NAME }}
         ACCESS_TOKEN: ${{ secrets.ACCESS_TOKEN }}
       run: |
         TAG=$(mvn -s $GITHUB_WORKSPACE/.github/workflows/settings.xml -q -N -Dmaven.build.cache.enabled=false -Dexec.executable='echo' -Dexec.args='${project.version}' exec:exec)
         contrib/datawave-quickstart/docker/docker-build.sh ${TAG} --docker-opts "${DOCKER_BUILD_OPTS}"
 
+  compose-build-and-test:
```
alerman: Is the plan to require this step to pass on all branch builds? Should this be a MR only check or otherwise restricted or do we believe it needs to be run on all commits?

.github/workflows/tests.yml
```diff
@@ -142,14 +147,92 @@ jobs:
...
+          ${{ runner.os }}-maven-
+    # Builds the quickstart and microservice docker images and runs a query test
+    - name: Docker Compose Query Tests
+      env:
+        USER_NAME: ${{ secrets.USER_NAME }}
+        ACCESS_TOKEN: ${{ secrets.ACCESS_TOKEN }}
+      run: |
+        mvn -s $GITHUB_WORKSPACE/.github/workflows/settings.xml -B -V -e -Pcompose -Dmicroservice-docker -Dquickstart-docker -Ddeploy -Dtar -DskipTests clean install
+        
+        # free up some space so that we don't run out
```
alerman: Is there some minimum amount of space we need that we could check for, then perform cleanup if necessary?

contrib/datawave-quickstart/bin/services/accumulo/bootstrap.sh
```diff
@@ -42,6 +42,7 @@ syncLimit=5
 clientPort=2181
 dataDir=${DW_CLOUD_DATA}/zookeeper
 maxClientCnxns=100
+4lw.commands.whitelist=ruok,wchs
```
alerman, line 45: Any reason not to use the wildcard (*) to enable all here?

.github/workflows/tests.yml
jwomeara: This is going to have to go away actually.  I had something similar for the spring-boot-2.x branch, and it's a nice idea, but unless everything is a snapshot version it doesn't really add much value.  I would prefer to comment it out AFTER this is merged just to make sure we are running this github action right up until the last possible second.

.mvn/maven-build-cache-config.xml
```diff
@@ -71,6 +71,22 @@
                         <goal>install</goal>
                     </goals>
                 </goalsList>
+                <goalsList artifactId="docker-maven-plugin">
```
jwomeara, line 74: @alerman I'm not sure any of these changes actually help.  I still have issues with the build cache when building everything.

contrib/datawave-quickstart/bin/services/datawave/install-ingest.sh
```diff
@@ -82,6 +82,8 @@ else
    if [ -d ${DW_DATAWAVE_INGEST_HOME}/accumulo-warehouse/lib/ext ]; then
       cp ${DW_DATAWAVE_INGEST_HOME}/accumulo-warehouse/lib/ext/*.jar ${ACCUMULO_HOME}/lib/ext > /dev/null 2>&1
    fi
+   info "Removing bundled guava jar from ${ACCUMULO_HOME}/lib"
```
jwomeara: I'm not sure this is necessary anymore.  I'll look into it.

core/connection-pool/src/main/java/datawave/core/query/runner/AccumuloConnectionRequestMap.java
```diff
@@ -0,0 +1,105 @@
...
+
+    private static Logger log = Logger.getLogger(AccumuloConnectionRequestMap.class);
+
+    /**
+     * This maps the query-id to a pair containing the tracking map (see the AccumuloConnectionFactory) and the thread handling the request
+     */
+    private Map<String,List<Pair<Map<String,String>,Thread>>> getConnectionThreadMap = new HashMap<>();
+
+    public boolean cancelConnectionRequest(String id, String userDn) {
+        // this call checks that the Principal used for the connection request and th connection cancel are the same
```
foster33:
```suggestion
        // this call checks that the Principal used for the connection request and the connection cancel are the same
```
Typo

core/map-reduce/src/main/java/datawave/core/mapreduce/bulkresults/map/ApplicationContextAwareMapper.java
```diff
@@ -0,0 +1,76 @@
...
+import org.springframework.core.env.EnumerablePropertySource;
+import org.springframework.core.env.Environment;
+import org.springframework.core.io.ClassPathResource;
+import org.springframework.core.io.support.ResourcePropertySource;
+
+public class ApplicationContextAwareMapper<KEYIN,VALUEIN,KEYOUT,VALUEOUT> extends Mapper<KEYIN,VALUEIN,KEYOUT,VALUEOUT> {
+
+    private static Logger log = Logger.getLogger(ApplicationContextAwareMapper.class);
+
+    //
```
foster33:
```suggestion
```
empty comment

core/query/src/main/java/datawave/core/query/logic/QueryLogic.java
```diff
@@ -285,23 +296,23 @@ default String getResponseClass(Query query) throws QueryException {
...
 
-    /** @return the connPoolName */
+    /**
+     * @return the connPoolName
+     */
     String getConnPoolName();
 
     /**
-     * Check that the user has one of the required roles principal my be null when there is no intent to control access to QueryLogic
+     * Check that the user has one of the required roles. userRoles my be null when there is no intent to control access to QueryLogic
```
foster33:
```suggestion
     * Check that the user has one of the required roles. userRoles may be null when there is no intent to control access to QueryLogic
```

core/query/src/main/java/datawave/core/query/logic/composite/CompositeQueryLogic.java
```diff
@@ -598,21 +599,83 @@ public Set<String> getExampleQueries() {
...
+        if (!isCheckpointable() || !(checkpoint instanceof CompositeQueryCheckpoint)) {
+            throw new UnsupportedOperationException("Cannot update a non-composite query checkpoint with the composite query logic.");
+        }
+
+        CompositeQueryCheckpoint compositeCheckpoint = (CompositeQueryCheckpoint) checkpoint;
+
+        CheckpointableQueryLogic logic = (CheckpointableQueryLogic) queryLogics.get(compositeCheckpoint.getDelegateQueryLogic());
+        if (logic == null) {
+            throw new UnsupportedOperationException(
+                            "Cannot update query checkpoint because delegate query logic [" + compositeCheckpoint.getDelegateQueryLogic() + "]does not exist");
```
foster33:
```suggestion
                            "Cannot update query checkpoint because delegate query logic [" + compositeCheckpoint.getDelegateQueryLogic() + "] does not exist");
```

core/query/src/main/java/datawave/core/query/logic/composite/CompositeQueryLogic.java
```diff
@@ -598,21 +599,83 @@ public Set<String> getExampleQueries() {
...
+        }
+
+        CompositeQueryConfiguration compositeConfig = (CompositeQueryConfiguration) config;
+
+        CompositeQueryCheckpoint compositeCheckpoint = (CompositeQueryCheckpoint) checkpoint;
+
+        CheckpointableQueryLogic logic = (CheckpointableQueryLogic) queryLogics.get(compositeCheckpoint.getDelegateQueryLogic());
+        if (logic == null) {
+            throw new UnsupportedOperationException(
+                            "Cannot update query checkpoint because delegate query logic [" + compositeCheckpoint.getDelegateQueryLogic() + "]does not exist");
```
foster33:
```suggestion
                            "Cannot update query checkpoint because delegate query logic [" + compositeCheckpoint.getDelegateQueryLogic() + "] does not exist");
```

warehouse/query-core/src/main/java/datawave/query/scheduler/Scheduler.java
```diff
@@ -1,24 +1,27 @@
...
 import datawave.query.config.ShardQueryConfiguration;
 import datawave.query.tables.ScannerFactory;
 import datawave.query.tables.stats.ScanSessionStats;
-import datawave.webservice.query.configuration.QueryData;
-
-public abstract class Scheduler implements CloseableIterable<Entry<Key,Value>> {
 
+/**
+ *
+ */
```
foster33: empty

warehouse/query-core/src/main/java/datawave/query/tables/BatchScannerSession.java
```diff
@@ -40,18 +40,32 @@
...
 import datawave.query.tables.async.Scan;
 import datawave.query.tables.async.ScannerChunk;
 import datawave.query.tables.async.SessionArbiter;
 import datawave.query.tables.async.SpeculativeScan;
-import datawave.webservice.query.Query;
 
-public class BatchScannerSession extends ScannerSession implements Iterator<Entry<Key,Value>>, FutureCallback<Scan>, SessionArbiter, UncaughtExceptionHandler {
+/**
+ *
+ */
```
foster33: empty

.github/workflows/tests.yml
```diff
@@ -73,55 +76,57 @@ jobs:
...
+            || { echo "***** TESTS FAILED. Attempting retry."; $RUN_TESTS; } \
+            || { echo "***** TESTS FAILED. Attempting final retry."; $RUN_TESTS; }
 
   quickstart-build-and-test:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout Code
       uses: actions/checkout@v3
+      with:
+        submodules: 'recursive'
```
jwomeara: This is useful while this exists as a PR, but I will remove this once the PR is merged.

.github/workflows/tests.yml
jwomeara: This is useful while this is a PR, but I will remove this once it's merged.

.gitignore
```diff
@@ -52,3 +52,6 @@ services/sample_configuration/*-dev.yml
 /.metadata/
 /.dockerignore
 /.maven-dockerignore
+
+/.dockerignore
```
jwomeara: i'll remove these dupes.

.github/workflows/tests.yml
```diff
@@ -56,6 +57,8 @@ jobs:
     steps:
     - name: Checkout Code
       uses: actions/checkout@v3
+      with:
```
jwomeara: This is useful while this exists as a PR, but I will remove this once the PR is merged.

core/connection-pool/src/main/java/datawave/core/common/result/ConnectionFactoryResponse.java
```diff
@@ -77,7 +77,7 @@ public String getMainContent() {
 
         builder.append("\n");
 
-        builder.append("table.connectionRequests td, table.connectionPools th {");
```
jwomeara: @ivakegg is this correct?

core/query/src/main/java/datawave/core/query/configuration/GenericQueryConfiguration.java
```diff
@@ -29,13 +32,19 @@
  * </p>
  *
  */
-public class GenericQueryConfiguration {
+public class GenericQueryConfiguration implements Serializable {
+    // is this execution expected to be checkpointable (changes how we allocate ranges to scanners)
+    private boolean checkpointable = false;
```
jwomeara, line 37: is a check for `this instanceof CheckpointableQueryConfiguration` not suitable enough for this?

core/query/src/main/java/datawave/core/query/configuration/GenericQueryConfiguration.java
```diff
@@ -219,10 +291,32 @@ public boolean canRunQuery() {
...
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o)
+            return true;
+        if (o == null || getClass() != o.getClass())
+            return false;
+        GenericQueryConfiguration that = (GenericQueryConfiguration) o;
+        return isCheckpointable() == that.isCheckpointable() && getBaseIteratorPriority() == that.getBaseIteratorPriority()
```
jwomeara, line 308: Probably need to make sure all of the class members are accounted for in equals and hashcode since this branch has been around as long as it has.

warehouse/core/src/main/java/datawave/ingest/util/cache/Loader.java
```diff
@@ -70,6 +70,7 @@ public ListenableFuture<V> reload(final K key, final V oldValue) {
                 log.trace("Reloading synchronously");
             try {
                 build(null);
+
```
alerman, line 73: Adding this new line intentional?

warehouse/ingest-configuration/src/main/resources/config/myjson-ingest-config.xml
```diff
@@ -208,11 +208,11 @@
         <description>ColumnVisibility expression to be applied to the "SUMMARY" field</description>
     </property>
 
-    <!--<property>-->
-        <!--<name>myjson.data.category.date</name>-->
-        <!--<value>PREMIERED</value>-->
-        <!--<description>Known date field to be used, if found, for the shard row id. Otherwise, current date will be used</description>-->
-    <!--</property>-->
+    <property>
```
alerman, line 211: Any Idea why this had been commented out? Seems like we would want this on in general?

warehouse/query-core/.attach_pid18134
alerman, line 1: inadvertant add?

core/connection-pool/src/main/java/datawave/core/common/cache/SharedCacheCoordinator.java
```diff
@@ -98,12 +91,7 @@ public interface EvictionCallback {
...
      *            number of locks
      */
-    @Inject
-    public SharedCacheCoordinator(@ConfigProperty(name = "dw.cache.coordinator.namespace") String namespace,
-                    @ConfigProperty(name = "dw.warehouse.zookeepers") String zookeeperConnectionString,
-                    @ConfigProperty(name = "dw.cacheCoordinator.evictionReaperIntervalSeconds", defaultValue = "30") int evictionReaperIntervalInSeconds,
-                    @ConfigProperty(name = "dw.cacheCoordinator.numLocks", defaultValue = "300") int numLocks,
-                    @ConfigProperty(name = "dw.cacheCoordinator.maxRetries", defaultValue = "10") int maxRetries) {
+    public SharedCacheCoordinator(String namespace, String zookeeperConnectionString, int evictionReaperIntervalInSeconds, int numLocks, int maxRetries) {
         ArgumentChecker.notNull(namespace, zookeeperConnectionString);
```
alerman, line 95: Do we have a dependency on Guava? If so, any reason not to use Preconditions.checkNotNull rather than implementing ourselves?

core/connection-pool/src/test/java/datawave/core/common/cache/SharedCacheCoordinatorTest.java
```diff
@@ -181,7 +182,7 @@ public void stateChanged(CuratorFramework client, ConnectionState newState) {}
 
             testingZooKeeperServer.kill();
 
-            for (int i = 0; i < 15; ++i) {
+            for (int i = 0; i < 20; ++i) {
```
alerman, line 185: Why 20? Are we just arbitrarily guessing how many iterations to do?

core/connection-pool/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,6 @@
+log4j.rootCategory=INFO, CONSOLE
+
```
alerman, line 2: Not sure if this is a log4j that would get passed through into production, but consider adding the monitorInterval setting to allow updates without app restart.

    monitorInterval = 30

core/query/src/main/java/datawave/core/query/configuration/GenericQueryConfiguration.java
alerman: For other projects I have used project lombok's @EqualsAndHashCode. Matbe we consider that? https://projectlombok.org/features/EqualsAndHashCode

core/query/src/main/java/datawave/core/query/logic/QueryKey.java
```diff
@@ -0,0 +1,104 @@
...
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class QueryKey implements Serializable {
+    private static final long serialVersionUID = -2589618312956104322L;
```
alerman, line 13: Curious how you decided on this serialVersionUID? Same question for other places. If you used some method to generate, we should be explicit about that wherever it is used so that future devs know how to update it.

core/query/src/main/java/datawave/core/query/logic/CheckpointableQueryLogic.java
```diff
@@ -0,0 +1,60 @@
...
+
+public interface CheckpointableQueryLogic {
+
+    /**
+     * This will allow us to check if a query logic is actually checkpointable. Even if the query logic supports it, the caller may have to tell the query logic
+     * that it is going to be checkpointed.
+     *
+     * @return true if checkpointable
+     */
+    boolean isCheckpointable();
```
jwomeara, line 17: @ivakegg this has always struck me as a little odd.  If the query logic implements CheckpointableQueryLogic, isn't that sign enough that the query logic is checkpointable?  Are you using this as more of an on/off switch?

docker/README.md
```diff
@@ -111,7 +163,7 @@ Docker Compose version v2.21.0
 
 #### Datawave Quickstart
 
-Prior to starting docker compose, you need to use the Datawave Quickstart to deploy working instances of Hadoop, Zookeeper, and Accumulo, along with some sample datasets for query.
+Prior to starting docker compose, you need to use the Qatawave Quickstart to deploy working instances of Hadoop, Zookeeper, and Accumulo, along with some sample datasets for query.
```
alerman: Typo QataWave. 

Obviously, it should be QuataWave

web-services/atom/src/main/java/datawave/webservice/atom/AtomServiceBean.java
```diff
@@ -120,8 +120,12 @@ public void setup() {}
     public Categories getCategories() {
         Principal p = ctx.getCallerPrincipal();
         Set<Authorizations> auths = new HashSet<>();
+        String userDN = null;
+        Collection<String> proxyServers = null;
         if (p instanceof DatawavePrincipal) {
```
alerman, line 125: This code is repeated in this class. Create a method to get the DN and server from the DP object perhaps?

warehouse/query-core/src/main/java/datawave/query/language/functions/lucene/LuceneQueryFunction.java
```diff
@@ -8,7 +8,6 @@
 import datawave.query.search.FieldedTerm;
 import datawave.query.search.WildcardFieldedFilter;
 
-@Deprecated
 public abstract class LuceneQueryFunction implements QueryFunction {
```
alerman: Are we undeprecating this?

microservices/README.md
```diff
@@ -0,0 +1,95 @@
+# DATAWAVE Query Microservices
```
jwomeara: Probably need to review this for accuracy.

web-services/common/src/main/java/datawave/webservice/common/audit/AuditBean.java
```diff
@@ -14,6 +14,7 @@
 
 import org.apache.log4j.Logger;
 import org.jboss.resteasy.annotations.GZIP;
+import org.springframework.util.MultiValueMap;
```
alerman, line 17: Whats the difference betweein this and MultivaluedMap? Can we use all of one or the other? Im worried about this causing confusion for developers down the line.

alerman: Same comment for other classes in this package.

alerman (commented):
Few small comments, but over all it's good.

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
```diff
@@ -713,6 +723,69 @@ public ShardQueryConfiguration(ShardQueryConfiguration other) {
...
     }
 
+    /**
+     * This constructor is used when we are creating a checkpoint for a set of ranges (i.e. QueryData objects). All configuration required for post planning
+     * needs to be copied over here.
+     *
+     * @param other
+     * @param queries
+     */
+    public ShardQueryConfiguration(ShardQueryConfiguration other, Collection<QueryData> queries) {
```
jwomeara, line 732: This probably needs to be updated.

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
```diff
@@ -2698,4 +2753,160 @@ public boolean isSortQueryByCounts() {
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o)
+            return true;
+        if (o == null || getClass() != o.getClass())
+            return false;
+        if (!super.equals(o))
+            return false;
+        ShardQueryConfiguration that = (ShardQueryConfiguration) o;
+        return isTldQuery() == that.isTldQuery() && isDisableIndexOnlyDocuments() == that.isDisableIndexOnlyDocuments()
```
jwomeara: this and the hash method probably need to be updated.

warehouse/query-core/src/main/java/datawave/query/iterator/profile/EvaluationTrackingNestedIterator.java
```diff
@@ -25,6 +25,7 @@ public EvaluationTrackingNestedIterator(QuerySpan.Stage stageName, QuerySpan myS
 
     @Override
     public T next() {
+        mySpan.next();
```
jwomeara, line 28: @ivakegg just checking if this is still needed.  not sure what the context surrounding this change was.

warehouse/query-core/src/main/java/datawave/query/language/functions/lucene/LuceneQueryFunction.java
jwomeara: i think i questioned whether these were actually unused or not.  i'll add the deprecation back in.

warehouse/query-core/src/main/java/datawave/query/tables/RemoteEventQueryLogic.java
```diff
@@ -69,16 +68,19 @@ public RemoteEventQueryLogic clone() {
 
     @Override
     public Set<String> getOptionalQueryParameters() {
+        // @TODO Get an instanceof the underlying query logic?
```
jwomeara: @ivakegg is this still a TODO?

warehouse/query-core/src/main/java/datawave/query/tables/shard/CountResultPostprocessor.java
```diff
@@ -0,0 +1,100 @@
...
+import org.apache.log4j.Logger;
+
+import com.google.common.collect.Sets;
+
+import datawave.core.query.logic.ResultPostprocessor;
+import datawave.marking.MarkingFunctions;
+import datawave.webservice.query.result.event.EventBase;
+import datawave.webservice.query.result.event.FieldBase;
+
+public class CountResultPostprocessor implements ResultPostprocessor {
```
jwomeara, line 19: @ivakegg I think we need result postprocessors for other things as well, like group by queries.

web-services/atom/src/main/java/datawave/webservice/atom/AtomServiceBean.java
jwomeara: i could go either way on this one...

web-services/common/src/main/java/datawave/webservice/common/audit/AuditBean.java
jwomeara: MultivaluedMap is a jboss/CDI/wildfly construct, so it should only really ever have been used in the webservice code.  MultiValueMap is the spring boot version of that class.

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -437,7 +466,7 @@ private QueryData validateQuery(String queryLogicName, MultivaluedMap<String,Str
 
         log.debug(queryParameters);
         qp.clear();
-        qp.setRequestHeaders(httpHeaders != null ? httpHeaders.getRequestHeaders() : null);
+        qp.setRequestHeaders(httpHeaders != null ? MapUtils.toMultiValueMap(httpHeaders.getRequestHeaders()) : null);
```
jwomeara, line 469: I'm not happy with this MapUtils stuff.  I will take another look at that.

core/query/src/main/java/datawave/core/query/configuration/GenericQueryConfiguration.java
apmoriarty: Another alternative is Apache's EqualsBuilder and HashCodeBuilder, we've used those elsewhere in the project.

core/query/src/main/java/datawave/core/query/configuration/GenericQueryConfiguration.java
apmoriarty: My preference is for instanceOf, it's more reliable than a boolean that could be accidentally set by a non-checkpointable query logic

core/query/src/main/java/datawave/core/query/logic/CheckpointableQueryLogic.java
apmoriarty: I agree. The only use I can see for this is if you wanted to do A/B testing with checkpointable logic disabled.

.gitmodules
foster33, line 1: This is missing a module for query-storage. For example, when `git submodule update --init --recursive` is ran, a fatal error occurs and outputs ```fatal: No url found for submodule path 'microservices/services/query-storage' in .gitmodules```

core/query/src/main/java/datawave/core/query/configuration/GenericQueryConfiguration.java
jwomeara: the only downside with removing this is that you wouldn't be able to force the query to run in an uncheckpointable way without the boolean.  not sure if we care about that at all.  @ivakegg?

core/query/src/main/java/datawave/core/query/configuration/GenericQueryConfiguration.java
jwomeara: @alerman unfortunately, for some of these classes i think there are class members which we *don't* want to consider when comparing for equality.

.github/workflows/tests.yml
jwomeara: this should go away AFTER the PR is merged.  I want to keep it in there until that point though just to make sure any changes to the microservices are tested appropriately.

.github/workflows/tests.yml
jwomeara: This is an issue I ran into when building all of the docker images.  I don't think this will be an issue once this github action is removed after merging.

contrib/datawave-quickstart/bin/services/accumulo/bootstrap.sh
jwomeara: I don't even remember why this was added or whether it's still needed.  @ivakegg?

warehouse/core/src/main/java/datawave/ingest/util/cache/Loader.java
jwomeara: Doing a blame on this line in intellij shows that this was part of the original code commit...

warehouse/ingest-configuration/src/main/resources/config/myjson-ingest-config.xml
jwomeara: I think I talked with @keith-ratcliffe at the time.  I think I had to do this in order to make the date ranges for my queries work reliably?  it's been a while, so im not sure.

core/connection-pool/src/main/java/datawave/core/common/cache/SharedCacheCoordinator.java
jwomeara: The notNull check wasn't added by me.  I don't really have an opinion here - I just left it as-is.

core/connection-pool/src/test/java/datawave/core/common/cache/SharedCacheCoordinatorTest.java
jwomeara: This is a timing thing.  Bumping the limit up increases the amount of time to perform the operation.

core/connection-pool/src/test/resources/log4j.properties
jwomeara: This is the test configuration.  As a rule I try not to deploy test code to production.  ;)

core/query/src/main/java/datawave/core/query/logic/QueryKey.java
jwomeara: I have never understood how these are generated, or even what their usefulness is.  @ivakegg?

.gitmodules
jwomeara: got it.  i intended to remove it.  i'll handle that properly with the next commit.

jwomeara: let me know if that's working for you now.

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
jwomeara: ehhhh, after looking at the usages of this class (MapUtils) I do think it's still necessary...

core/query/src/main/java/datawave/core/query/logic/CheckpointableQueryLogic.java
jwomeara: I think I could get on board for renaming this to something along the lines of 'enableCheckpoints' or something like that.  This is less about whether the query logic is checkpointable, and more about whether we want to use it in a checkpointable fashion.

microservices/README.md
jwomeara: Actually, I'm going to get rid of the entire thing since it was more of a place to take notes during early development.

core/query/src/main/java/datawave/core/query/configuration/GenericQueryConfiguration.java
jwomeara: After looking at this, I think all of the members are accounted for in both equals and hashCode.  I don't have a strong opinion about which method to use, other than to say that what's there is already implemented...  🤷

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
jwomeara: Looking at this a little closer, I guess it's ok.  The full copy constructor can manage copying all of the fields, while this one only handles the ones that are important for a checkpoint.  @ivakegg how did you decide on the list of members that need to be copied to support checkpointable queries, and could these members be moved to the CheckpointableQueryConfiguration class (or something equivalent) to better distinguish between properties that are important to checkpointable queries, and properties that aren't?

.gitmodules
foster33: It is good now

PaulsGitHubs: I'm actually so relieved it's working.

core/common-util/src/main/java/datawave/core/query/cache/ResultsPage.java
```diff
@@ -0,0 +1,45 @@
+package datawave.core.query.cache;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ *
+ */
+public class ResultsPage<T> {
```
FineAndDandy, line 9: is a ResultsPage intended to be a one to one to a query or one to many?

core/connection-pool/src/main/java/datawave/core/common/connection/AccumuloConnectionFactoryImpl.java
```diff
@@ -0,0 +1,381 @@
...
+        this.connectionPoolsConfiguration = config;
+        log.info("Initializing AccumuloConnectionFactoryImpl with " + config.getDefaultPool() + " and " + config.getPoolNames());
+        init();
+    }
+
+    public void init() {
+        this.pools = new HashMap<>();
+
+        if (this.connectionPoolsConfiguration == null) {
+            log.error("connectionPoolsConfiguration was null - aborting init()");
```
FineAndDandy, line 69: should this throw IllegalStateException?

core/connection-pool/src/main/java/datawave/core/query/runner/AccumuloConnectionRequestMap.java
```diff
@@ -0,0 +1,105 @@
...
+ * For storing a map of queryId to Thread that is requesting an AccumuloConnection
+ */
+public class AccumuloConnectionRequestMap {
+
+    private static Logger log = Logger.getLogger(AccumuloConnectionRequestMap.class);
+
+    /**
+     * This maps the query-id to a pair containing the tracking map (see the AccumuloConnectionFactory) and the thread handling the request
+     */
+    private Map<String,List<Pair<Map<String,String>,Thread>>> getConnectionThreadMap = new HashMap<>();
```
FineAndDandy:
```suggestion
    private Map<String,List<Pair<Map<String,String>,Thread>>> connectionThreadMap = new HashMap<>();
```

core/modification/src/main/java/datawave/modification/MutableMetadataUUIDHandler.java
```diff
@@ -322,13 +329,7 @@ else if (f.getName().equalsIgnoreCase(event.getIdType()) && fieldCount < 1 && co
                                         log.trace("Submitting request to MutableMetadataHandler from MutableMetadataUUIDHandler: " + modReq);
 
                                     // make sure user isn't null or empty
```
FineAndDandy: this is no longer being checked add the null/empty check or remove the comment

core/modification/src/main/java/datawave/modification/cache/ModificationCache.java
```diff
@@ -0,0 +1,120 @@
...
+                log.trace("datatype = " + e.getKey() + ", fieldcount = " + e.getValue().size());
+            }
+            // now atomically replace the cache
+            this.cache = cache;
+        } catch (Exception e) {
+            log.error("Error during initialization of ModificationCacheBean", e);
+            throw new RuntimeException("Error during initialization of ModificationCacheBean", e);
+        } finally {
+            if (null != s)
+                s.close();
```
FineAndDandy, line 83: if s.close() can throw an exception it would be possible to leak the connection

core/modification/src/main/java/datawave/modification/configuration/ModificationServiceConfiguration.java
```diff
@@ -47,14 +48,18 @@ public void setSecurityMarkingExemptFields(List<String> securityMarkingExemptFie
     /**
      * Handle to query service in case the modification service needs to run queries.
      *
-     * @return RemoteQueryExecutor
+     * @return ModificationQueryService
      */
-    public QueryExecutorBean getQueryService() {
-        return queryService;
+    public ModificationQueryService getQueryService(ProxiedUserDetails userDetails) {
+        return queryServiceFactory.createService(userDetails);
```
FineAndDandy, line 54: potential npe

core/query/src/main/java/datawave/core/query/configuration/GenericQueryConfiguration.java
FineAndDandy: there could definitely be a case for a best effort no checkpoint query

core/query/src/main/java/datawave/core/query/configuration/QueryData.java
```diff
@@ -0,0 +1,576 @@
...
+    public String getQuery() {
+        return query;
+    }
+
+    public String getTableName() {
+        return tableName;
+    }
+
+    public void setTableName(String tableName) {
+        this.tableName = tableName;
```
FineAndDandy, line 185: resetHashCode()

core/query/src/main/java/datawave/core/query/configuration/QueryData.java
```diff
@@ -0,0 +1,576 @@
...
+
+            for (int i = 0; i < node.size(); i++) {
+                list.add(getIteratorSetting(node.get(i)));
+            }
+
+            return list;
+        }
+
+        private IteratorSetting getIteratorSetting(JsonNode node) throws JsonProcessingException {
+            IteratorSetting setting = new IteratorSetting(1, "a", "a");
```
FineAndDandy: this initialization is a bit strange without a comment

core/common-util/src/main/java/datawave/core/query/cache/ResultsPage.java
jwomeara: I don't understand your question.  ResultsPage isn't a new concept - it's currently in use on integration.

core/connection-pool/src/main/java/datawave/core/common/connection/AccumuloConnectionFactoryImpl.java
jwomeara: Good question.  This is identical to what is currently in use on integration though.

core/connection-pool/src/main/java/datawave/core/query/runner/AccumuloConnectionRequestMap.java
jwomeara: yuck.  good catch.

core/modification/src/main/java/datawave/modification/MutableMetadataUUIDHandler.java
jwomeara: on it.

core/modification/src/main/java/datawave/modification/cache/ModificationCache.java
jwomeara: FWIW this code was lifted, unchanged from ModificationCacheBean.  This was just an attempt to strip the core functionality from the webservices module, and put it in a common place.

core/modification/src/main/java/datawave/modification/configuration/ModificationServiceConfiguration.java
jwomeara: FWIW a similar problem exists on integration.  On integration we just grab the query service and start using it.  What would you like to see done here?  Personally, if the factory is null I would like to see it blow up so that we know we aren't configured correctly.

core/query/src/main/java/datawave/core/query/configuration/QueryData.java
jwomeara: this `resetHashCode()` stuff feels really hokey, but ok.

core/query/src/main/java/datawave/core/query/configuration/QueryData.java
jwomeara: I agree.  @ivakegg what is the rhyme or reason for these constructor params?

core/modification/src/main/java/datawave/modification/configuration/ModificationServiceConfiguration.java
FineAndDandy: throw an IllegalStateException?

core/query/src/main/java/datawave/core/query/logic/DelegatingQueryLogic.java
```diff
@@ -353,6 +323,51 @@ public void validate(Map<String,List<String>> parameters) throws IllegalArgument
...
+    @Override
+    public ProxiedUserDetails getServerUser() {
+        return delegate.getServerUser();
+    }
+
+    @Override
+    public void setServerUser(ProxiedUserDetails serverUser) {
+        delegate.setServerUser(serverUser);
+    }
+
```
FineAndDandy, line 370: after seeing identical code in the BaseQueryLogic maybe this should be pulled out into a separate class and composed?

core/query/src/main/java/datawave/core/query/logic/QueryKey.java
FineAndDandy: the serialVersionUID has to be assigned for any Serializable class for compatibility. This value is likely the value automatically generated by the compiler, but it can also be versioned manually for compatibility

core/query/src/main/java/datawave/core/query/logic/composite/CompositeQueryLogic.java
```diff
@@ -241,9 +234,9 @@ public Set<Authorizations> updateRuntimeAuthorizationsAndQueryAuths(QueryLogic<?
         settings.setQueryAuthorizations(validQueryAuthorizations);
 
         // recalculate the runtime query authorizations (no need to pass in userService as we have already recalculated the principal)
-        Set<Authorizations> downgradedAuths = WSAuthorizationsUtil.getDowngradedAuthorizations(validQueryAuthorizations, principal, queryPrincipal);
+        Set<Authorizations> downgradedAuths = AuthorizationsUtil.getDowngradedAuthorizations(validQueryAuthorizations, currentUser, queryUser);
         if (log.isTraceEnabled()) {
-            log.trace("Principal auths for user " + principal.getPrimaryUser().getCommonName() + " are " + principal.getPrimaryUser().getAuths());
+            log.trace("Principal auths for user " + currentUser.getPrimaryUser().getCommonName() + " are " + currentUser.getPrimaryUser().getAuths());
```
FineAndDandy, line 239: can these names be changed or javadoced to reflect what they are intended to represent?

core/query/src/main/java/datawave/core/query/logic/DelegatingQueryLogic.java
jwomeara: I don't understand your comment.  BaseQueryLogic isn't doing any kind of delegation like this class is doing.

core/query/src/main/java/datawave/core/query/logic/composite/CompositeQueryLogic.java
jwomeara: Which names?  Changed to what?

core/modification/src/main/java/datawave/modification/configuration/ModificationServiceConfiguration.java
jwomeara: Both of those are RuntimeExceptions...  If it's all the same to you, I would rather just stick with the NPE.

contrib/datawave-quickstart/bin/services/accumulo/bootstrap.sh
alerman: In one of the versions of ZK they made them all disabled by default. I am sure that at least ruok had to be added for health checks, I just don't know why we would limit it and not just enable all.

warehouse/query-core/src/main/java/datawave/query/tables/shard/CountResultPostprocessor.java
ivakegg: Do we not use the DocumentTransformer mechanism?  The GroupingTransform and UniqueTransform should be incorporated into that.  What we may be missing is the persistence of the grouping results or the unique bloom filter since we no longer have a stateful RunningQuery mechanism anymore.  I will check into that.  This is on the list of things to check in the ticket.

warehouse/query-core/src/main/java/datawave/query/tables/RemoteEventQueryLogic.java
ivakegg: I don't think so.  I will remove in the main line.

ivakegg: This was already removed in the main line.  Perhaps a bad merge somewhere along the line.  I will remove it here.

warehouse/query-core/src/main/java/datawave/query/tables/shard/CountResultPostprocessor.java
jwomeara: By the time the results get to the query service they are already transformed results, instead of individual events/documents that can be aggregated by the document transformer.  So we need a way to aggregate the results in that format, which is what the result postprocessor does.  

I already added functionality to persist the results in the queue for queries that need some form of final aggregation (e.g. count queries).  To see how that works, check out NextCall.java and search for `returnIntermediateResults`.  Essentially, if this is a query that needs to be aggregated fully before it can be returned, and the next call times out or short circuits or whatever, the aggregated results seen so far will be published back into the queue.  

This is where results are published back into the queue.
https://github.com/NationalSecurityAgency/datawave-query-service/blob/feature/queryMicroservices/service/src/main/java/datawave/microservice/query/runner/NextCall.java#L158

There is also a new boolean in the GenericQueryConfiguration (reduceResults) which is used to indicate that results need to be reduced before anything is returned to the user.

warehouse/query-core/src/main/java/datawave/query/tables/RemoteEventQueryLogic.java
ivakegg: done

warehouse/query-core/src/main/java/datawave/query/iterator/profile/EvaluationTrackingNestedIterator.java
ivakegg: I am pretty sure this is still correct.  This is tracking the next/seek calls in the evaluation pipeline.

core/query/src/main/java/datawave/core/query/logic/QueryKey.java
ivakegg: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/Serializable.html

core/query/src/main/java/datawave/core/query/logic/CheckpointableQueryLogic.java
ivakegg: So the setCheckpointable method below would be the method to change to enableCheckpoints if you wish.  Its the CreateTask/NextTask classes used by the query executor service that needs to know whether the query can be checkpointed of not.  If not, then in the create task all of the results are pulled at that point.  If is it then other executor service instances can handle pulling results as needed.  This mechanism allows us to migrate query logics into the query microservices framework but still support those query logics that have not been migrated.

jwomeara: I would prefer enableCheckpoints or checkpointsEnabled, but that's not a hill I'm willing to die on.

warehouse/query-core/src/main/java/datawave/query/iterator/profile/EvaluationTrackingNestedIterator.java
jwomeara: ok, cool.  I just wasn't familiar with it so wanted to bring it to your attention.

warehouse/query-core/src/main/java/datawave/query/tables/RemoteEventQueryLogic.java
jwomeara: cool, thanks.  i did many, many, MANY merges so no surprise to me if i merged something in an undesirable way.

core/query/src/main/java/datawave/core/query/configuration/GenericQueryConfiguration.java
ivakegg: So the point of this is that an implementation may support checkpointing, however we may not want to run it in a checkpointable way for latency purposes (think UUID lookup which uses the shard query logic and the shard query configuration)

jwomeara: If you update the other class to be `enableCheckpoints` or `checkpointsEnabled` can you do the same here?

core/connection-pool/src/main/java/datawave/core/common/result/ConnectionFactoryResponse.java
ivakegg: reverted

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
ivakegg: There may be a few of the properties that are common to all checkpointable queries (i.e. need to consider other query logics like the edge queries for example).  I will try to assess the commonalities.

ivakegg: The common fields required are already covered by the GenericQueryConfiguration copy constructor.  I am cleaning out any redundencies however.

ivakegg: Found some redundency with the super copy constructor and some incorrectly overridden Query members.

core/query/src/main/java/datawave/core/query/configuration/QueryData.java
ivakegg: reworked to avoid hokey constructor params

web-services/query/src/test/java/datawave/webservice/query/runner/QueryExecutorBeanTest.java
```diff
@@ -145,7 +146,7 @@ public class QueryExecutorBeanTest {
     private AuditService auditService;
     private QueryMetricsBean metrics;
     private QueryLogicFactoryImpl queryLogicFactory;
-    private QueryExpirationConfiguration queryExpirationConf;
+    private QueryExpirationProperties queryExpirationConf;
```
billoley, line 149: Maybe: queryExpirationProperties

web-services/query/src/test/java/datawave/webservice/query/runner/ExtendedQueryExecutorBeanTest.java
```diff
@@ -242,10 +240,10 @@ public void setupBefore() throws Exception {
 
         queryLogic2 = PowerMock.createMock(QuerySyntaxParserQueryLogic.class);
 
-        queryExpirationConf = new QueryExpirationConfiguration();
-        queryExpirationConf.setPageSizeShortCircuitCheckTime(45);
-        queryExpirationConf.setPageShortCircuitTimeout(58);
-        queryExpirationConf.setCallTime(60);
+        queryExpirationConf = new QueryExpirationProperties();
```
billoley, line 243: queryExpirationProperties

web-services/query/src/test/java/datawave/webservice/query/cache/RunningQueryTimingImplTest.java
billoley, line 1: QueryExpirationProperties conf -->  QueryExpirationProperties queryExpirationProperties (multiple occurrences)

web-services/query/src/test/java/datawave/webservice/query/cache/QueryExpirationBeanTest.java
```diff
@@ -70,11 +71,11 @@ public void testRemoveIdleOrExpired() throws Exception {
     private QueryExpirationBean createBean(int expireTime) throws IllegalArgumentException, IllegalAccessException {
         QueryExpirationBean bean = new QueryExpirationBean();
 
-        QueryExpirationConfiguration expirationConfiguration = new QueryExpirationConfiguration();
-        setInternalState(expirationConfiguration, "idleTimeMinutes", expireTime);
-        setInternalState(expirationConfiguration, "callTimeMinutes", expireTime);
+        QueryExpirationProperties expirationConfiguration = new QueryExpirationProperties();
```
billoley, line 74: queryExpirationProperties

web-services/query/src/main/java/datawave/webservice/query/cache/QueryExpirationBean.java
```diff
@@ -44,7 +45,7 @@ public class QueryExpirationBean {
 
     @Inject
     @SpringBean(refreshable = true)
-    private QueryExpirationConfiguration conf;
+    private QueryExpirationProperties conf;
```
billoley, line 48: queryExpirationProperties

warehouse/query-core/src/main/java/datawave/query/tables/edge/EdgeQueryLogic.java
billoley, line 1: Not the only place that this changes, but I really like how you got rid of the dupication of settings by using the configuration objects

web-services/query/src/test/java/datawave/webservice/query/runner/QueryExecutorBeanTest.java
jwomeara: I was just trying to keep the name consistent with what it has always been.  Do you have a strong opinion about changing this?

PR #2197, opened 2023-12-14
Move geo response classes to query-metric-api, ensure json/xml/html responses and map link work from webservice

billoley:
Move geo response classes to query-metric-api, ensure json/xml/html responses and map link work from webservice

In the querymetric-serice, the isGeoQuery field is now serialized back to the webserver
Both the query metric and geo html responses will include the correct urls for css and js includes such that if the html is returned from the webserver, then those will be fetched from the webserver and if it is returned from the querymetric service then those will be fetched from the querymetric service

Combined with the changes in the querymetric service, this will allow the map link to be displayed on querymetric html pages


ivakegg (dismissed):

keith-ratcliffe (dismissed):

keith-ratcliffe (approved):

ivakegg (approved):

apmoriarty (approved):

billoley:
Superceded by https://github.com/NationalSecurityAgency/datawave/pull/2402

PR #2198, opened 2023-12-15
updated to use jexl3

jwomeara:
I recreated this branch for integration since I don't feel like waiting for query microservices for this...  The original description follows:

This updates our jexl dependency to 3.3.

Here's a summary of the major changes made with this update:

- Jexl no longer parents the following nodes with an ASTReference: ASTAndNode, ASTOrNode, ASTFunctionNode, ASTIdentifier, ASTStringLiteral, ASTNumberLiteral, ASTReferenceExpression. So any logic that assumed this to be true has been updated.
- There are no more custom Datawave JexlNodes. The following have been removed: ASTDelayedPredicate, ASTEvaluationOnly, ASTUnknownFieldERNode, ASTUnsatisfiableERNode, ASTUnsatisfiableGENode, ASTUnsatisfiableGTNode, ASTUnsatisfiableLENode, ASTUnsatisfiableLTNode, BoundedRange, ExceededtermThresholdJexlNode, ExceededValueThresholdJexlNode, IndexHoleMarkerJexlNode,
- QueryPropertyMarker has been simplified due to the removal of the custom Datawave Jexl Nodes. Nodes are marked similarly to how they were before, but the nodes themselves will be Jexl primitives, and not custom Datawave Jexl Nodes.
- ASTFunctionNodes now have 2 children: ASTNamespaceIdentifier and ASTArguments. ASTNamespaceIdentifier contains the function namespace and name. ASTArguments contains all of the arguments.
- JexlNodes no longer have an 'image' member. Each individual node uses it's own parameter to store it's value. A JexlNodes.getImage method has been added to simulate the prior functionality.
- JexlNodes no longer give direct access to the child array. Therefore, the JexlNodes.children() method has been removed, since it is impossible to return the original child array, and I wanted to avoid confusion over whether what was returned is a copy of the original array or the array itself.
- That is all that I can think of at the moment that has changed, but I'm sure I'm missing something. Please take a close look at the changes and let me know what you think.


warehouse/query-core/src/main/java/datawave/query/discovery/FindLiteralsAndPatternsVisitor.java
```diff
@@ -44,17 +42,17 @@ public static QueryValues find(JexlNode root) {
 
     @Override
     public Object visit(ASTStringLiteral node, Object data) {
-        // strings are always wrapped in a reference node, so the op is the gparent
-        JexlNode op = node.jjtGetParent().jjtGetParent();
+        // strings are NEVER wrapped in a reference node, so the op is the parent
```
apmoriarty, line 45: Good doc update

warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
```diff
@@ -401,48 +401,88 @@ public IndexStream visit(ASTOrNode node, Object data) {
 
     @Override
     public IndexStream visit(ASTAndNode node, Object data) {
-        Intersection.Builder builder = Intersection.builder();
-        builder.setUidIntersector(uidIntersector);
+        QueryPropertyMarker.Instance instance = QueryPropertyMarker.findInstance(node);
```
apmoriarty, line 404: RangeStream mods look good. Marker nodes were handled in part by the visit to ASTReference but are now part of the ASTAndNode visit.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/FunctionJexlNodeVisitor.java
```diff
@@ -44,27 +40,23 @@ public static FunctionJexlNodeVisitor eval(JexlNode node) {
...
     public static ASTFunctionNode makeFunctionFrom(String ns, String functionName, JexlNode... arguments) {
-        ASTFunctionNode fn = new ASTFunctionNode(ParserTreeConstants.JJTFUNCTIONNODE);
-        ASTIdentifier namespace = JexlNodes.makeIdentifierWithImage(ns);
-        ASTIdentifier function = JexlNodes.makeIdentifierWithImage(functionName);
-        ArrayList<JexlNode> nodes = Lists.newArrayList();
-        nodes.add(namespace);
-        nodes.add(function);
-        Collections.addAll(nodes, arguments);
-        return JexlNodes.children(fn, nodes.toArray(new JexlNode[nodes.size()]));
+        return JexlNodes.makeFunction(ns, functionName, arguments);
```
apmoriarty, line 61: Much cleaner, love it

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -364,32 +349,26 @@ public static String getIdentifier(JexlNode node) throws NoSuchElementException
...
-                            return (deconstruct ? deconstructIdentifier(grandChild.image) : grandChild.image);
-                        }
-                    }
-                } else if (child instanceof ASTIdentifier) {
-                    return (deconstruct ? deconstructIdentifier(child.image) : child.image);
-                }
+            JexlNode child = node.jjtGetChild(0);
 
+            if (child instanceof ASTIdentifier) {
+                // as of jexl3, identifiers are NOT parented by ASTReference nodes unless they are 'ant-style' and contain '.' characters
```
apmoriarty, line 355: good doc update

warehouse/query-core/src/main/java/datawave/query/discovery/FindLiteralsAndPatternsVisitor.java
```diff
@@ -44,17 +42,17 @@ public static QueryValues find(JexlNode root) {
...
+        JexlNode op = node.jjtGetParent();
         if (op instanceof ASTEQNode) {
-            JexlNode field = JexlNodes.otherChild(op, node.jjtGetParent()).jjtGetChild(0);
+            JexlNode field = JexlNodes.otherChild(op, node);
             if (log.isTraceEnabled()) {
-                log.trace("Found field " + JexlASTHelper.deconstructIdentifier(field.image) + "==" + node.getLiteral());
+                log.trace("Found field " + JexlASTHelper.deconstructIdentifier(((ASTIdentifier) field).getName()) + "==" + node.getLiteral());
```
ivakegg: Are we sure this will be an identifier ?

warehouse/query-core/src/main/java/datawave/query/jexl/functions/CompareFunctionValidator.java
```diff
@@ -31,14 +32,14 @@ public static void validate(String function, List<JexlNode> args) throws Illegal
                             MessageFormat.format("{0} requires 4 arguments", function));
             throw new IllegalArgumentException(qe);
         } else {
-            if (!operators.contains(args.get(1).image)) {
+            if (!operators.contains(String.valueOf(JexlNodes.getImage(args.get(1))))) {
```
ivakegg: Can the image be null in which case String.valueOf(null) will return "null"?  That might not be what we want here.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/CompareFunctionValidator.java
```diff
@@ -31,14 +32,14 @@ public static void validate(String function, List<JexlNode> args) throws Illegal
...
                 BadRequestQueryException qe = new BadRequestQueryException(DatawaveErrorCode.INVALID_FUNCTION_ARGUMENTS,
                                 MessageFormat.format("{0} requires valid comparison operator (<, <=, >, >=, ==/= or !=) as 2nd arguments", function));
                 throw new IllegalArgumentException(qe);
             }
 
             try {
-                Mode.valueOf(args.get(2).image);
+                Mode.valueOf(String.valueOf(JexlNodes.getImage(args.get(2))));
```
ivakegg: Same comment as above

warehouse/query-core/src/main/java/org/apache/commons/jexl3/parser/JexlNodes.java
```diff
@@ -195,25 +193,113 @@ public static <T extends JexlNode> T swap(T parent, JexlNode a, JexlNode b) {
...
+    }
+
+    public static void setIdentifierAccess(ASTIdentifierAccess identifierAccess, String value) {
+        Preconditions.checkNotNull(identifierAccess);
+        Preconditions.checkNotNull(value);
+
+        identifierAccess.setIdentifier(value);
+    }
+
+    public static Object getImage(JexlNode node) {
```
ivakegg: I think we may need to create a getStringImage(JexlNode node) that returns the string version instead of using String.valueOf everywhere.  This way we can also handle returning a null image as null instead of the string "null".

warehouse/query-core/src/main/java/datawave/query/jexl/functions/CompareFunctionValidator.java
ivakegg: I made a comment in JexlNodes to add a getStringImage method that can handle this case appropriately.  Then we can drop all of the String.valueOf calls everywhere.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FixNegativeNumbersVisitor.java
```diff
@@ -19,23 +20,34 @@ public static ASTJexlScript fix(JexlNode root) {
...
+            negated = -number.longValue();
+        } else if (number instanceof BigInteger) {
+            negated = ((BigInteger) number).negate();
+        } else if (number instanceof Float) {
+            negated = -number.floatValue();
+        } else if (number instanceof Double) {
+            negated = -number.doubleValue();
+        } else if (number instanceof BigDecimal) {
+            negated = ((BigDecimal) number).negate();
+        }
```
ivakegg, line 54: perhaps the default action should be negated = -number.doubleValue();

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/DroppedExpression.java
```diff
@@ -0,0 +1,82 @@
...
+        Object queryObj = assignments.get(REASON_LABEL);
+        if (queryObj != null) {
+            return String.valueOf(queryObj);
+        } else {
+            return null;
+        }
+    }
+
+    protected JexlNode createJexlNode() {
+        List<JexlNode> nodes = new ArrayList<>();
```
ivakegg, line 70: We used to have ((\_Drop_=true) && ((\_Query_='   ') && (\_Reason_='  ')))
What happened to the \_Drop_ part?

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java
```diff
@@ -1,216 +1,166 @@
...
 import datawave.query.jexl.visitors.QueryPropertyMarkerVisitor;
 
 /**
- * This is a node that can be put in place of an underlying reference node to place a property on an underlying query sub-tree (e.g. ExceededValueThreshold)
+ * This is a node that can be put in place of a given node to place a property on an underlying query sub-tree (e.g. ExceededValueThreshold)
  */
-public abstract class QueryPropertyMarker extends ASTReference {
+public class QueryPropertyMarker {
+
+    public enum MarkerType {
```
ivakegg, line 28: NIce, I like this

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IsNotNullPruningVisitor.java
```diff
@@ -122,34 +117,36 @@ public static JexlNode prune(JexlNode node) {
      */
     @Override
     public Object visit(ASTAndNode node, Object data) {
+        // do not process marker nodes
```
ivakegg, line 120: not traversing marker nodes may not be what we want here.  We were traversing them previously.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExecutableDeterminationVisitor.java
```diff
@@ -717,55 +714,7 @@ public Object visit(ASTNENode node, Object data) {
 
     @Override
     public Object visit(ASTReference node, Object data) {
```
ivakegg, line 716: Do we need to keep a visit method for the ASTReference ?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PullupUnexecutableNodesVisitor.java
```diff
@@ -152,61 +145,79 @@ private boolean containsChild(JexlNode node, ExecutableDeterminationVisitor.STAT
...
-                if (state == ExecutableDeterminationVisitor.STATE.EXECUTABLE) {
-                    executable = true;
+            swap(parent.jjtGetParent(), parent, source);
+            return source;
+        } else if (!ExecutableDeterminationVisitor.isExecutable(node, config, indexedFields, indexOnlyFields, nonEventFields, forFieldIndex, null, helper)) {
+            // we need to make sure one of these nodes executable, and we have no partials
+            // TODO we should use some cost/index stats info
+
+            // determine if we have any executable children
+            if (!ExecutableDeterminationVisitor.isExecutable(node, config, indexedFields, indexOnlyFields, nonEventFields, forFieldIndex, null, helper)) {
```
ivakegg: no need to call this method again here.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/SatisfactionVisitor.java
```diff
@@ -106,26 +106,20 @@ public Object visit(ASTEQNode node, Object data) {
     }
 
     @Override
-    public Object visit(ASTReference node, Object o) {
+    public Object visit(ASTAndNode and, Object data) {
+        QueryPropertyMarker.Instance instance = QueryPropertyMarker.findInstance(and);
         // Recurse only if not delayed
-        if (!QueryPropertyMarker.findInstance(node).isType(ASTDelayedPredicate.class)) {
-            super.visit(node, o);
+        if (!instance.isType(DELAYED)) {
```
ivakegg, line 112: if we are checking for delayed and exceeded, then does this not equate to looking for any marker

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IsNotNullPruningVisitor.java
apmoriarty: That's not exactly true. QueryPropertyMarkers are excluded via the visit to an ASTReference node in line 355
```
    @Override
        // do not descend into marker nodes
        if (!QueryPropertyMarkerVisitor.getInstance(node).isAnyType()) {
            node.childrenAccept(this, data);
        }
        return data;
    }
```

That said, this visitor should probably visit certain types of marker nodes when determining if an 'isNotNull' term can be pruned, for example:
```
# should prune. Value marker will put values for FOO into the context
!(FOO != null) && ((_Value_ = true) && (FOO =~ 'ba.*'))

# should prune. Delayed marker won't pull values for FOO into the context, but will exclude an event just as well as the isNotNull term.
!(FOO != null) && ((_Delayed_ = true) && (FOO =~ 'ba.*'))

# should NOT prune
!(FOO != null) && ((_Drop_ = true) && (FOO =~ 'ba.*'))
```

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownNegationVisitor.java
```diff
@@ -259,14 +261,13 @@ private JexlNode dropNot(ASTNotNode toDrop, NegationState state) {
...
-                if (child instanceof ASTReference && grandChild instanceof ASTReferenceExpression) {
-                    JexlNodes.swap(toDrop.jjtGetParent(), toDrop, greatGrandChild);
+            // check for and remove full node chain of 'NotNode - RefExpr'
+            if (child.jjtGetNumChildren() > 0) {
+                JexlNode grandChild = child.jjtGetChild(0);
+                if (child instanceof ASTReferenceExpression) {
+                    JexlNodes.swap(toDrop.jjtGetParent(), toDrop, grandChild);
                     state.setNegated(false);
-                    return greatGrandChild;
+                    return grandChild;
```
apmoriarty, line 270: Good change

warehouse/query-core/src/main/java/datawave/query/planner/rules/NodeTransformVisitor.java
```diff
@@ -106,12 +111,7 @@ public Object visit(ASTFunctionNode node, Object data) {
 
     @Override
     public Object visit(ASTReference node, Object data) {
```
ivakegg: should we drop this method ?  Are reference nodes still being used just not in the same places?  If so what are they used for.

warehouse/query-core/src/main/java/org/apache/commons/jexl3/parser/JexlNodes.java
```diff
@@ -65,24 +68,14 @@ public static int id(JexlNode n) {
     public static <T extends JexlNode> T newInstanceOfType(T node) {
         try {
             @SuppressWarnings("rawtypes")
-            Constructor constructor = node.getClass().getConstructor(Integer.TYPE);
+            Constructor constructor = node.getClass().getDeclaredConstructor(Integer.TYPE);
```
ivakegg, line 71: I believe this constructor appears to be public on the nodes.  Can we not still used getConstructor ?

ivakegg:
I made it through all 379 changed files.  This was a lot of work.  Thank you.

warehouse/query-core/src/main/java/datawave/query/discovery/FindLiteralsAndPatternsVisitor.java
jwomeara: In this method we have a string literal with an equals node as a parent.  Technically there has NEVER been a guarantee that the sibling node is an identifier, but that's the assumption this code has always made (as evidenced by the deconstructIdentifier call).  

This brings up an issue with our current understanding of how Jexl works.  In Jexl 2.x every node has an image property (string), but only a small subset of nodes actually use the image variable (String literal, Number literal, Identifier).  The image isn't always accurate either - for a negative number, the number literal will actually have the positive number as the image and will be wrapped in a unary minus node to signify that it's negative.  I mention this because I think we have made some inaccurate assumptions about how images work in our code.  

Getting back to this method, we have been relying on the sibling node to have a valid image value, but that will not be the case unless the sibling is a literal or an identifier.  If the other side of the equals node is a function or a method, then field.image would have returned null.  So, while it seems pretty clear to me that this code expected the sibling to be an identifier, I suppose I can update this to call JexlNodes.getImage instead, which will behave similarly to the previous logic.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/CompareFunctionValidator.java
jwomeara: The image CAN be null per my earlier comments.  I updated this to call JexlNodes.getIdentifierOrLiteralAsString which should match the previous behavior.

warehouse/query-core/src/main/java/datawave/query/discovery/FindLiteralsAndPatternsVisitor.java
jwomeara: Updated.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/CompareFunctionValidator.java
jwomeara: Updated.

warehouse/query-core/src/main/java/org/apache/commons/jexl3/parser/JexlNodes.java
jwomeara: check out JexlNodes.getIdentifierOrLiteralAsString.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FixNegativeNumbersVisitor.java
jwomeara: The previous code throws an illegal argument exception if the value isn't a known type (byte, short, integer, long, biginteger, float, double, bigdecimal), so I updated the code to match that behavior.

warehouse/query-core/src/main/java/datawave/query/jexl/nodes/DroppedExpression.java
jwomeara: https://github.com/NationalSecurityAgency/datawave/blob/jexl3/warehouse/query-core/src/main/java/datawave/query/jexl/nodes/QueryPropertyMarker.java#L32

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IsNotNullPruningVisitor.java
jwomeara: https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IsNotNullPruningVisitor.java#L357

I didn't make this logic up.  And nodes are no longer parented by reference nodes, so the reference node logic was added to the and node.  I am just trying to preserve existing behavior.  If we want to do something different we should write a new ticket.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExecutableDeterminationVisitor.java
jwomeara: In the previous code we checked for markers at the reference node - that logic has been added to the and node method.  There was no other special functionality for reference nodes, so unless there is some special functionality we want to run for these nodes, I think the base visitor method is fine.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PullupUnexecutableNodesVisitor.java
jwomeara: Sure, I can make that change.  This was a byproduct of moving the reference node logic into the and node, so we were running this check redundantly in the previous code for what it's worth.  Tests pass with this change.

warehouse/query-core/src/main/java/datawave/query/planner/rules/NodeTransformVisitor.java
jwomeara: Looking at the available NodeTransformRules, there aren't any that operate on reference nodes, so we can remove this.  

From what I have seen reference nodes are still used in certain situations, for instance with methods and identifier access nodes.  (e.g. something.someMethod(), and something.someProperty).  Identifier access nodes are a new concept with jexl3 and are used when using ant-style fields (i.e. fields with periods in them).

warehouse/query-core/src/main/java/org/apache/commons/jexl3/parser/JexlNodes.java
jwomeara: There are quite a few nodes which do not have a public constructor matching this signature.

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
```diff
@@ -190,29 +202,31 @@ public Object visit(ASTERNode node, Object data) {
...
+            result = new FunctionalSet(); // an empty set
         }
-        return false;
+        return result;
     }
 
-    public Object visit(ASTMethodNode node, Object data) {
-        if (data == null) {
-            data = new FunctionalSet(); // an empty set
+    public boolean hasMethodSibling(JexlNode node) {
```
lbschanno, line 214: This method will not return true if we see a method sibling before 'foundSelf' is set to true. If we need to support that case, we can simplify this to:
```java
public boolean hasMethodSibling(JexlNode node) {
    if (node.jjtGetParent() != null && node.jjtGetParent().jjtGetNumChildren() > 1) {
        JexlNode parent = node.jjtGetParent();
        for (int i = 0; i < parent.jjtGetNumChildren(); i++) {
            Node child = parent.jjtGetChild(i);
            if (child != node && child instanceof ASTMethodNode) {
                return true;
            }
        }
    }
    return false;
}
```

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
```diff
@@ -261,11 +262,78 @@ public static JexlNode createNodeFromValuesForField(String field, IndexLookupMap
...
+    /**
+     * This method is currently only used from the QueryModelVisitor to expand a binary node into multiple pairs. If one side contains the lenient marker, then
+     * that will be extended around the binary node.
+     *
+     * @param containerType
+     * @param node
+     * @param pairs
+     * @return The container of binary nodes based on the specified node type
+     */
+    public static JexlNode createNodeTreeFromPairs(ContainerType containerType, JexlNode node, Set<List<JexlNode>> pairs) {
```
lbschanno: This method is not used anywhere as far as I can see, do we need to keep this? If so, I'd like to see the `pairs` argument be a set of `Pair<JexlNode,JexlNode>` instances instead (`org.javatuples.Pair`). This way you don't need verify that each list has two elements.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionsDescriptor.java
```diff
@@ -557,17 +558,17 @@ public static JexlNode updateFunctionWithField(JexlNode function, String field)
...
-            if (zoneExists) {
-                iter.next(); // skip a zone argument if it exists
-            }
-            while (iter.hasNext()) {
-                updated.add(iter.next());
+
+            ASTArguments argsNode = (ASTArguments) functionCopy.jjtGetChild(1);
+            // skip a zone argument if it exists
+            for (int i = 0; i < argsNode.jjtGetNumChildren(); i++) {
+                if (i != 0 || !zoneExists) {
```
lbschanno: This can be simplified to avoid making checks each loop with the following:
```java
// Skip the first child if a zone exists.
for (int i = zoneExists ? 1 : 0; i < argsNode.jjtGetNumChildren(); i++) {
    updated.add(argsNode.jjtGetChild(i));
}
```

warehouse/query-core/src/main/java/datawave/query/jexl/functions/JexlFunctionArgumentDescriptorFactory.java
```diff
@@ -69,40 +67,19 @@ public static JexlArgumentDescriptor getArgumentDescriptor(ASTFunctionNode node)
          * Uses the FunctionReferenceVisitor to try and get the namespace from a function node
          */
         public static Class<?> extractFunctionClass(ASTFunctionNode node) {
-            String namespace = node.jjtGetChild(0).image;
-
-            Object possibleMatch = ArithmeticJexlEngines.functions().get(namespace);
-
-            if (possibleMatch != null && possibleMatch instanceof Class<?>) {
-                return (Class<?>) possibleMatch;
+            if (node.jjtGetNumChildren() == 2 && node.jjtGetChild(0) instanceof ASTNamespaceIdentifier && node.jjtGetChild(1) instanceof ASTArguments) {
```
lbschanno, line 70: Is it at all possible to have an ASTFunctionNode with more than 2 children? If so, consider checking that the function node has at least 2 children rather than exactly 2 children to safeguard against that scenario.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/FunctionReferenceVisitor.java
```diff
@@ -97,21 +99,20 @@ public Multimap<String,Function> functions() {
     @Override
     public Object visit(ASTFunctionNode node, Object data) {
         final int nChildren = node.jjtGetNumChildren();
-        if (nChildren < 3) {
-            log.error("Function node does have 3 children-- must supply a namespace, function name and at least one argument.");
+        if (nChildren < 2) {
```
lbschanno: This can be simplified to `if(node.jjtGetNumChildren() < 2)`. Additionally, the error message needs to be corrected to say that the function node does **not** have 2 children.

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
jwomeara: methods appear to the right of the identifier, so i only want to check the sibling immediately after the node in question.

lbschanno: Understood. In that case, I think a short comment noting that in the function itself would be helpful to anyone unfamiliar with it.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlNodeFactory.java
jwomeara: It looks like this got moved to QueryModelVisitor.  Great!  I'll remove it here.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionsDescriptor.java
jwomeara: cool.  updated in next commit.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/JexlFunctionArgumentDescriptorFactory.java
jwomeara: as far as i can tell, a function always has a namespace identifier node and an arguments node.  even the native jexl code assumes that there are two children per function node, so i think this is a safe assumption.

warehouse/query-core/src/main/java/datawave/query/postprocessing/tf/FunctionReferenceVisitor.java
jwomeara: good point.  i wonder why it was written like that in the first place.  anyway... updated.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/JexlFunctionArgumentDescriptorFactory.java
lbschanno: Cool, then I'm fine leaving it as is.

jwomeara:
Made all requested updates.

apmoriarty (approved):

jwomeara:
integration tests succeeded.  did i resolve everyone's comments?  if so, can you approve so i can merge?

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
jwomeara: added a comment

lbschanno (approved):

ivakegg (approved):

PR #2208, opened 2023-12-28
Feature/splittable rfiles

FineAndDandy:
Allow rfiles to be split by index block boundaries for parallel read processing.

### DIFF
--- a/warehouse/core/src/main/java/datawave/ingest/mapreduce/job/RFileSplit.java
+++ b/warehouse/core/src/main/java/datawave/ingest/mapreduce/job/RFileSplit.java
@@ -21,7 +21,7 @@ public class RFileSplit extends FileSplit {
         top = new Key();
     }
 
-    public RFileSplit(Path path, long fileStart, long fileLength, String[] hosts, int startBlock, int numBlocks, Key top) {
+    public RFileSplit(Path path, long fileStart, long fileLength, String[] hosts, long startBlock, long numBlocks, Key top) {
         super(path, fileStart, fileLength, hosts);
         this.startBlock = startBlock;
         this.numBlocks = numBlocks;
@@ -30,6 +30,10 @@ public class RFileSplit extends FileSplit {
 
     public long getStartBlock() {
         return startBlock;
+    }
+
+    public long getNumBlocks() {
+        return numBlocks;
     }
 
     public Key getTopKey() {
@@ -45,11 +49,6 @@ public class RFileSplit extends FileSplit {
      */
     public Range getSeekRange(Key start, Key end) {
         return new Range(start, true, end, false);
-    }
-
-    @Override
-    public long getLength() {
-        return numBlocks;
     }
 
     @Override
--- a/warehouse/core/src/main/java/datawave/ingest/mapreduce/job/RFileSplit.java
+++ b/warehouse/core/src/main/java/datawave/ingest/mapreduce/job/RFileSplit.java
@@ -1,0 +1,75 @@
+package datawave.ingest.mapreduce.job;
+
+import java.io.DataInput;
+import java.io.DataOutput;
+import java.io.IOException;
+
+import org.apache.accumulo.core.data.Key;
+import org.apache.accumulo.core.data.Range;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.mapreduce.lib.input.FileSplit;
+
+// For a file with N blocks
+// startBlock 0 - beginning of the file to the first block
+// startBlock N - last block to the end
+public class RFileSplit extends FileSplit {
+    private long startBlock;
+    private long numBlocks;
+    private Key top;
+
+    public RFileSplit() {
+        top = new Key();
+    }
+
+    public RFileSplit(Path path, long fileStart, long fileLength, String[] hosts, int startBlock, int numBlocks, Key top) {
+        super(path, fileStart, fileLength, hosts);
+        this.startBlock = startBlock;
+        this.numBlocks = numBlocks;
+        this.top = top;
+    }
+
+    public long getStartBlock() {
+        return startBlock;
+    }
+
+    public Key getTopKey() {
+        return top;
+    }
+
+    /**
+     * Create a range which is start key inclusive and end key exclusive
+     *
+     * @param start
+     * @param end
+     * @return
+     */
+    public Range getSeekRange(Key start, Key end) {
+        return new Range(start, true, end, false);
+    }
+
+    @Override
+    public long getLength() {
+        return numBlocks;
+    }
+
+    @Override
+    public String toString() {
+        return super.toString() + ":" + startBlock + "+" + numBlocks + "+" + top;
+    }
+
+    @Override
+    public void write(DataOutput out) throws IOException {
+        super.write(out);
+        top.write(out);
+        out.writeLong(startBlock);
+        out.writeLong(numBlocks);
+    }
+
+    @Override
+    public void readFields(DataInput in) throws IOException {
+        super.readFields(in);
+        top.readFields(in);
+        startBlock = in.readLong();
+        numBlocks = in.readLong();
+    }
+}
--- a/warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormat.java
+++ b/warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormat.java
@@ -30,7 +30,9 @@ public class SplittableRFileInputFormat extends RFileInputFormat {
 
     @Override
     protected boolean isSplitable(JobContext context, Path filename) {
-        return true;
+        // prevent FileInputFormat from splitting the raw files, potentially creating duplicates with InputSplit.getStart() != 0 when super.getSplits() is
+        // called
+        return false;
     }
 
     @Override
--- a/warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormat.java
+++ b/warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormat.java
@@ -1,0 +1,252 @@
+package datawave.ingest.mapreduce.job;
+
+import org.apache.accumulo.core.conf.DefaultConfiguration;
+import org.apache.accumulo.core.crypto.CryptoFactoryLoader;
+import org.apache.accumulo.core.data.Key;
+import org.apache.accumulo.core.data.Range;
+import org.apache.accumulo.core.data.Value;
+import org.apache.accumulo.core.file.FileOperations;
+import org.apache.accumulo.core.file.FileSKVIterator;
+import org.apache.accumulo.core.file.FileSKVWriter;
+import org.apache.accumulo.core.file.rfile.RFileOperations;
+import org.apache.accumulo.core.spi.crypto.CryptoEnvironment;
+import org.apache.accumulo.core.spi.crypto.CryptoService;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FSDataOutputStream;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.mapreduce.InputSplit;
+import org.apache.hadoop.mapreduce.JobContext;
+import org.apache.hadoop.mapreduce.RecordReader;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+import org.apache.hadoop.mapreduce.lib.input.FileSplit;
+import org.apache.log4j.Level;
+import org.apache.log4j.Logger;
+
+import java.io.EOFException;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import static org.apache.accumulo.core.conf.Property.TABLE_CRYPTO_PREFIX;
+
+public class SplittableRFileInputFormat extends RFileInputFormat {
+    @Override
+    protected boolean isSplitable(JobContext context, Path filename) {
+        return true;
+    }
+
+    @Override
+    public RecordReader<Key,Value> createRecordReader(InputSplit split, TaskAttemptContext context) throws IOException, InterruptedException {
+        return new SplittableRFileRecordReader();
+    }
+
+    @Override
+    public List<InputSplit> getSplits(JobContext job) throws IOException {
+        Configuration config = job.getConfiguration();
+
+        List<InputSplit> rfileSplits = new ArrayList<>();
+
+
+        // get the configured directories/files
+        for (InputSplit inputSplit : super.getSplits(job)) {
+            if (!(inputSplit instanceof FileSplit)) {
+                throw new IllegalArgumentException("Must have file splits");
+            }
+
+            // for each file get the index blocks
+            FileSplit fileSplit = (FileSplit) inputSplit;
+            rfileSplits.addAll(getSplits(config, fileSplit));
+        }
+
+        return rfileSplits;
+    }
+
+    public static List<InputSplit> getSplits(Configuration config, FileSplit fileSplit) throws IOException {
+        Path rfile = fileSplit.getPath();
+        FileSystem fs = rfile.getFileSystem(config);
+
+        if (!fs.exists(rfile)) {
+            throw new IllegalArgumentException(rfile + " does not exist");
+        }
+
+        FileOperations ops = RFileOperations.getInstance();
+        String file = rfile.toString();
+        CryptoService cs = CryptoFactoryLoader.getServiceForClient(CryptoEnvironment.Scope.TABLE, config.getPropsWithPrefix(TABLE_CRYPTO_PREFIX.name()));
+        FileSKVIterator iter = ops.newIndexReaderBuilder().forFile(file, fs, config, cs).withTableConfiguration(DefaultConfiguration.getInstance()).build();
+
+        List<InputSplit> splits = new ArrayList<>();
+        int blkCount = 0;
+        while (iter.hasTop()) {
+            splits.add(new RFileSplit(fileSplit, blkCount++, 1));
+            iter.next();
+        }
+
+        // add the last split
+        splits.add(new RFileSplit(fileSplit, blkCount, 1));
+
+        return splits;
+    }
+
+    // get the splits for the rfile
+    public static List<InputSplit> getSplits(Configuration config, Path rfile) throws IOException {
+        return getSplits(config, new FileSplit(rfile, 0, 0, new String[0]));
+    }
+
+    public static FileSKVIterator getIterator(Configuration config, RFileSplit split) throws IOException, InterruptedException {
+        return getKeys(config, split.getFileSplit().getPath(), split.getStartBlock(), split.getLength());
+    }
+
+    public static FileSKVIterator getKeys(Configuration config, Path rfile, int startIndexBlock, long numIndexBlocks) throws IOException {
+        FileSystem fs = rfile.getFileSystem(config);
+
+        if (!fs.exists(rfile)) {
+            throw new IllegalArgumentException(rfile.toString() + " does not exist");
+        }
+
+        FileOperations ops = RFileOperations.getInstance();
+        String file = rfile.toString();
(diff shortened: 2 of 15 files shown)


warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormat.java
```diff
@@ -0,0 +1,137 @@
...
+        log.info("getting splits for job");
+        // get the configured directories/files
+        for (InputSplit inputSplit : super.getSplits(job)) {
+            if (!(inputSplit instanceof FileSplit)) {
+                throw new IllegalArgumentException("Must have file splits");
+            }
+
+            // for each file get the index blocks
+            FileSplit fileSplit = (FileSplit) inputSplit;
+            log.info("creating splits for: " + fileSplit);
```
jschmidt10: This logline dupes the one in `getSplits(config, fileSplit)`

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormat.java
```diff
@@ -0,0 +1,137 @@
...
+     * @param fileSplit
+     * @return
+     * @throws IOException
+     */
+    public static List<InputSplit> getSplits(Configuration config, FileSplit fileSplit) throws IOException {
+        log.info("getting splits for: " + fileSplit);
+        Path rfile = fileSplit.getPath();
+        FileSystem fs = rfile.getFileSystem(config);
+
+        if (!fs.exists(rfile)) {
```
jschmidt10: Do you see us hitting this condition? It looks like we'll only get a `FileSplit` from `FileInputFormat` if it exists. This would be like an early catch if the RFile got compacted I guess? Just wondering if we should forego this for the sake of saving the roundtrips to the namenode when this would only trip in a narrow window between `FileInputFormat.getSplits()` and this call.

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormat.java
```diff
@@ -0,0 +1,137 @@
...
+        FileSKVIterator iter = rfileReader.getIndex();
+
+        // track the last split key to since multiple splits with the same split key MUST be in the same block
+        Key lastSplit = firstKey;
+        List<InputSplit> splits = new ArrayList<>();
+        int blkCount = 0;
+        int splitBlocks = 0;
+
+        int minBlocksPerSplit = config.getInt(MIN_BLOCKS_PER_SPLIT, 1);
+        log.info("Blocks per split: " + minBlocksPerSplit);
```
jschmidt10: This is going to log the same information for every underlying RFile we process. I'm picturing the log looking like:
```
[INFO] Blocks per split: 100
[INFO] Blocks per split: 100
...
```
Might be worth bumping this out to the top level `getSplits()`

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/RFileSplit.java
```diff
@@ -0,0 +1,75 @@
...
+    private long numBlocks;
+    private Key top;
+
+    public RFileSplit() {
+        top = new Key();
+    }
+
+    public RFileSplit(Path path, long fileStart, long fileLength, String[] hosts, int startBlock, int numBlocks, Key top) {
+        super(path, fileStart, fileLength, hosts);
+        this.startBlock = startBlock;
```
jschmidt10, line 26: We're converting `startBlock` and `numBlocks` from an int to a long. Do we need it to be a long?

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormat.java
```diff
@@ -0,0 +1,137 @@
...
+        Key top = null;
+        while (iter.hasTop()) {
+            splitBlocks++;
+            top = iter.getTopKey();
+
+            if (!top.equals(lastSplit) && splitBlocks >= minBlocksPerSplit) {
+                splits.add(new RFileSplit(fileSplit.getPath(), fileSplit.getStart(), fileSplit.getLength(), fileSplit.getLocations(), blkCount, splitBlocks,
+                                top));
+                blkCount += splitBlocks;
+                lastSplit = iter.getTopKey();
```
jschmidt10:
```suggestion
                lastSplit = top;
```

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormat.java
```diff
@@ -0,0 +1,137 @@
...
+                splitBlocks = 0;
+            }
+            iter.next();
+        }
+
+        // if there is a gap between the last index key top key and the lastKey in the file or there hasn't been a split created yet
+        if (top.equals(lastKey) || splits.isEmpty()) {
```
jschmidt10: The comment mentions "if there is a gap..." but the first condition is checking if they are equal. Should the check be for "not equal"?

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormat.java
```diff
@@ -0,0 +1,137 @@
...
+            splits.add(new RFileSplit(fileSplit.getPath(), fileSplit.getStart(), fileSplit.getLength(), fileSplit.getLocations(), blkCount, splitBlocks + 1,
+                            top));
+        }
+
+        return splits;
+    }
+
+    // get the splits for the rfile
+    public static List<InputSplit> getSplits(Configuration config, Path rfile) throws IOException {
+        return getSplits(config, new FileSplit(rfile, 0, 0, new String[0]));
```
jschmidt10: It looks like this method is only used by unit tests and it's a suboptimal split since it's stubbing out the length and the locations. Thoughts on moving it to the unit test as a utility method there, just to prevent anyone from using this in production code.

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/RFileSplit.java
```diff
@@ -0,0 +1,75 @@
...
+     * @param end
+     * @return
+     */
+    public Range getSeekRange(Key start, Key end) {
+        return new Range(start, true, end, false);
+    }
+
+    @Override
+    public long getLength() {
+        return numBlocks;
```
jschmidt10: The API mentions this is the number of bytes that will be read from the split. We're converting it to the a number of blocks - which is fine probably as the length relative to other splits will stay similar. Was there a reason to override this vs just leaving it as the default of `fileLength`?

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileRecordReader.java
```diff
@@ -0,0 +1,95 @@
...
+        FileSystem fs = rfile.getFileSystem(config);
+
+        if (!fs.exists(rfile)) {
+            throw new IllegalArgumentException(rfile + " does not exist");
+        }
+
+        CryptoService cs = CryptoFactoryLoader.getServiceForClient(CryptoEnvironment.Scope.TABLE, config.getPropsWithPrefix(TABLE_CRYPTO_PREFIX.name()));
+        CachableBlockFile.CachableBuilder cb = new CachableBlockFile.CachableBuilder().fsPath(fs, rfile).conf(config).cryptoService(cs);
+        RFile.Reader rfileReader = new RFile.Reader(cb);
+        FileSKVIterator iter = rfileReader.getIndex();
```
jschmidt10, line 53: The construction of this iterator takes about 10 lines of code that appears here and in the InputFormat. Might be worth wrapping that in a static utility method somewhere for reuse.

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/RFileSplit.java
jschmidt10: I see down below where we use the `getLength()` method to retrieve the number of index blocks. We could create a new method for `getNumBlocks()` like you did for the `startBlock`.

warehouse/core/src/test/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormatTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+    @Before
+    public void setup() {
+        Logger.getRootLogger().setLevel(Level.WARN);
+        config = new Configuration();
+    }
+
+    @After
+    public void cleanup() {
+
```
jschmidt10: If you made your `tmpFile` a class level variable, you could do something like:
```
if (tmpFile != null && tmpFile.exists())
  tmpFile.delete()
```
That'd get rid of the try/finally in all of your tests.

warehouse/core/src/test/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormatTest.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    @Test
+    public void testSingleMultiKeySplit() throws IOException, InterruptedException {
+        List<Map.Entry<Key,Value>> data = new ArrayList<>();
+        Key start = new Key(String.format("%08x", 0));
+        Key end = new Key(String.format("%08x", 1));
+        for (int i = 0; i < 2; i++) {
+            Key key = new Key(String.format("%08x", i));
+            data.add(new AbstractMap.SimpleEntry<>(key, new Value(String.format("%08x", i))));
```
jschmidt10: I couldn't tell where we were using the Value objects. If we aren't, we could just use `new Value()` for all the tests - it'd make the test code a little less dense/easier to read. Along those lines, introducing an `entry(key)` function could slim it up a bunch:

```
data.add(entry(key));
```
Then that `entry()` function could wrap your key in a SimpleEntry with an empty Value.

jschmidt10:
Awesome feature. Thanks for implementing this!

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormat.java
FineAndDandy: will remove

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormat.java
FineAndDandy: The check really isn't there for that case at all, but instead because the method is public, when being called from getSplits(JobContext) I don't believe this is a case to worry about.

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormat.java
FineAndDandy: agree

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/RFileSplit.java
FineAndDandy: these should all be longs, its very unlikely, but possible to extend this many blocks. I will update to be longs throughout

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormat.java
FineAndDandy: will fix

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormat.java
FineAndDandy: good catch, I'll add a test for this case as well

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormat.java
FineAndDandy: agree

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/RFileSplit.java
FineAndDandy: your right, there isn't a good reason to break the contract, will update

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileRecordReader.java
FineAndDandy: will condense

warehouse/core/src/test/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormatTest.java
FineAndDandy: will update

warehouse/core/src/test/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormatTest.java
FineAndDandy: I was originally using the Value objects to validate no keys were being lost or duplicated, but once I convinced myself this wasn't happening I ended up dropping those tests. I should bring them back, but also simplify the test creation by abstracting it.

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormat.java
FineAndDandy: After going back to this, the conditional is actually not needed at all. In all cases there is a final split to add.

jschmidt10 (dismissed):

avgAGB (dismissed):

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileRecordReader.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        Key end = null;
+        if (iter.hasTop()) {
+            end = iter.getTopKey();
+        }
+
+        Range seekRange = split.getSeekRange(start, end);
+
+        rfileReader.seek(seekRange, Collections.EMPTY_SET, false);
```
apmoriarty: sonar lint would prefer `Collections.emptySet()`

apmoriarty (dismissed):

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileRecordReader.java
```diff
@@ -0,0 +1,99 @@
...
+        } else {
+            super.initialize(split, context);
+        }
+    }
+
+    public static RFile.Reader getRFileReader(Configuration config, Path rfile) throws IOException {
+        FileSystem fs = rfile.getFileSystem(config);
+
+        if (!fs.exists(rfile)) {
+            throw new IllegalArgumentException(rfile + " does not exist");
```
drewfarris: Consider throwing a `FileNotFoundException` instead?

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileRecordReader.java
```diff
@@ -0,0 +1,99 @@
...
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.mapreduce.InputSplit;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+
+public class SplittableRFileRecordReader extends RFileRecordReader {
+    @Override
+    public void initialize(InputSplit split, TaskAttemptContext context) throws IOException, InterruptedException {
+        if (split instanceof RFileSplit) {
```
drewfarris, line 31: Consider commenting on the cases when an `RFileSplit` is expected versus some other type of `InputSplit`, it was a little difficult to trace through the code and figure this out.

Also, `RFileRecordReader` makes an assumption in its `initialize` method that `fileSplit.getStart()` will always return 0 and throws an `IOException` if it doesn't. I don't think that will cause problems in this case, but not 100% certain.

drewfarris (dismissed):

jschmidt10 (dismissed):

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileRecordReader.java
FineAndDandy: I did notice that because I'm calling super.getSplits(job) to get the file splits and I've overridden isSplittable() to return true there actually are pre-split rfiles returned. These will need to be de-duped/discarded in the SplittableRFileInputFormat. Good catch.

jschmidt10 (dismissed):

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormat.java
```diff
@@ -0,0 +1,121 @@
...
+import org.apache.accumulo.core.spi.crypto.CryptoService;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.mapreduce.InputSplit;
+import org.apache.hadoop.mapreduce.JobContext;
+import org.apache.hadoop.mapreduce.RecordReader;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+import org.apache.hadoop.mapreduce.lib.input.FileSplit;
+import org.apache.log4j.Logger;
```
drewfarris: Cleanup unused imports

warehouse/core/src/test/java/datawave/ingest/mapreduce/job/SplittableRFileInputFormatTest.java
```diff
@@ -0,0 +1,263 @@
...
+     * @param blkCount
+     *            the number of blocks in the split
+     * @param first
+     *            the first key of the split
+     * @param last
+     *            the last key of the split
+     * @param keyCount
+     *            the number of keys in the split
+     * @throws IOException
+     * @throws InterruptedException
```
drewfarris: Remove the @throws here and the throws statements for `InterruptedException` in the test method declarations above

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileRecordReader.java
```diff
@@ -0,0 +1,109 @@
...
+
+    public static RFile.Reader getRFileReader(Configuration config, Path rfile) throws IOException {
+        FileSystem fs = rfile.getFileSystem(config);
+
+        if (!fs.exists(rfile)) {
+            throw new FileNotFoundException(rfile + " does not exist");
+        }
+
+        CryptoService cs = CryptoFactoryLoader.getServiceForClient(CryptoEnvironment.Scope.TABLE, config.getPropsWithPrefix(TABLE_CRYPTO_PREFIX.name()));
+        CachableBlockFile.CachableBuilder cb = new CachableBlockFile.CachableBuilder().fsPath(fs, rfile).conf(config).cryptoService(cs);
```
drewfarris: This is some 'magic code' that's reused in the unit test, perhaps a good candidate for a static utility method?

drewfarris (dismissed):
LGTM. Minor nits only.

hlgp (approved):
Cool stuff!

hlgp (dismissed):

warehouse/core/src/main/java/datawave/ingest/mapreduce/job/SplittableRFileRecordReader.java
FineAndDandy: updated to point to utility class

FineAndDandy:
addressed final comments and rebased with integration

apmoriarty (approved):

avgAGB (approved):

PR #2214, opened 2024-01-11
Fix bypass accumulo option for accumulo2.1

apmoriarty:
- Enable remote debugging of shard ranges
- Update iterator environments to not use 'deprecated' methods.

### DIFF
--- a/warehouse/core/src/main/java/datawave/mr/bulk/BulkIteratorEnvironment.java
+++ b/warehouse/core/src/main/java/datawave/mr/bulk/BulkIteratorEnvironment.java
@@ -2,6 +2,7 @@ package datawave.mr.bulk;
 
 import java.io.IOException;
 
+import org.apache.accumulo.core.client.PluginEnvironment;
 import org.apache.accumulo.core.client.SampleNotPresentException;
 import org.apache.accumulo.core.client.sample.SamplerConfiguration;
 import org.apache.accumulo.core.conf.AccumuloConfiguration;
@@ -12,6 +13,7 @@ import org.apache.accumulo.core.iterators.IteratorEnvironment;
 import org.apache.accumulo.core.iterators.IteratorUtil.IteratorScope;
 import org.apache.accumulo.core.iterators.SortedKeyValueIterator;
 import org.apache.accumulo.core.security.Authorizations;
+import org.apache.accumulo.core.spi.common.ServiceEnvironment;
 
 public class BulkIteratorEnvironment implements IteratorEnvironment {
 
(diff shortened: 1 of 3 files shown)


alerman (approved):

warehouse/core/src/main/java/datawave/mr/bulk/BulkIteratorEnvironment.java
```diff
@@ -69,4 +71,14 @@ public SortedKeyValueIterator<Key,Value> reserveMapFileReader(String mapFileName
         return null;
     }
 
+    @Override
+    public ServiceEnvironment getServiceEnv() {
```
jschmidt10, line 75: I wouldn't bother overriding this as it's deprecated in Accumulo 2+. If we don't, does something break?

warehouse/core/src/main/java/datawave/mr/bulk/BulkIteratorEnvironment.java
```diff
@@ -69,4 +71,14 @@ public SortedKeyValueIterator<Key,Value> reserveMapFileReader(String mapFileName
...
+        return null;
+    }
+
+    @Override
+    public PluginEnvironment getPluginEnv() {
+        return null;
```
jschmidt10, line 81: In Accumulo 2, we're supposed to be fetching PluginEnv instead of caching Configuration objects (which we have done a few times in the past). I suspect having this return null will become a problem when iterators are migrated to caching the PluginEnv (like we did for ConfigurableAgeOffFilter). 

I think we might be able to construct a PluginEnvironment using the conf. 

If this works, we can also start here but I do think we might hit issues eventually.

warehouse/query-core/src/main/java/datawave/mr/bulk/RecordIterator.java
```diff
@@ -200,6 +200,15 @@ public SamplerConfiguration getSamplerConfiguration() {
             return null;
         }
 
+        @Override
+        public ServiceEnvironment getServiceEnv() {
```
jschmidt10, line 204: Same comment here - if we can get away without overriding this, I would drop it here as it'll eventually get removed from Accumulo's API.

apmoriarty: If you try to bypass accumulo today you will get an UnsupportedException from the default iterator environment. The overides are unfortunately required.

jschmidt10: Ok. Thanks!

jschmidt10 (approved):

warehouse/core/src/main/java/datawave/mr/bulk/BulkIteratorEnvironment.java
apmoriarty: I think that work will definitely need to get done at some point, preferably sooner rather than later. This merge request is just to get bypass accumulo working again.

PR #2216, opened 2024-01-12
Add FederatedQueryPlanner

### LINKED ISSUE #825: Handle the cases were a field is both indexed and not indexed within a time range

lbschanno:
Adds a FederatedQueryPlanner that will break up a query into multiple queries scanning over subsets of the original target date range if field index holes are identified to be present for the query in the target date range.

Note: the work in this PR is dependent on the work in:
- PR #2094
- PR [datawave-metadata-utils/30](https://github.com/NationalSecurityAgency/datawave-metadata-utils/pull/30)
- PR [datawave-metadata-utils/31](https://github.com/NationalSecurityAgency/datawave-metadata-utils/pull/31)

Closes #825


warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -444,12 +446,33 @@ public void initialize(ShardQueryConfiguration config, AccumuloClient client, Qu
                 config.setProjectFields(getCardinalityConfiguration().getRevisedProjectFields(queryModel, originalProjectFields));
             }
 
-            this.queries = getQueryPlanner().process(config, jexlQueryString, settings, this.getScannerFactory());
-
+            // If the planner is a DefaultQueryPlanner, delegate the execution to a FederatedQueryPlanner.
+            if (getQueryPlanner() instanceof DefaultQueryPlanner) {
```
ivakegg: Why not have the "getQueryPlanner()" simply return a FederatedQueryPlanner ?  This way we can do some A/B testing to ensure we are doing the right thing.  Currently we set the query planner to a configured DefaultQueryPlanner via the QueryLogicFactory.xml.  Perhaps we can set the delegated query planner via a setQueryPlanner on the FederatedQueryPlanner itself.

warehouse/query-core/src/main/java/datawave/query/planner/FederatedQueryPlanner.java
```diff
@@ -0,0 +1,315 @@
...
+        if (!(genericConfig instanceof ShardQueryConfiguration)) {
+            throw new ClassCastException("Config must be an instance of " + ShardQueryConfiguration.class.getSimpleName());
+        }
+
+        ShardQueryConfiguration config = (ShardQueryConfiguration) genericConfig;
+
+        log.debug("Query originally set to execute against date range " + dateFormat.format(originalBeginDate) + "-" + dateFormat.format(originalEndDate));
+
+        // Get the relevant date ranges.
+        // TODO - Determine if we should pass in fields and datatypes to filter on for this query. Can we do this before calling process() on the query?
```
ivakegg: The Query object should have the data types filter.  At this point is has already been passed to the configuration object via the ShardQueryLogic initialize method.

warehouse/query-core/src/main/java/datawave/query/planner/FederatedQueryPlanner.java
```diff
@@ -0,0 +1,315 @@
...
+        SortedSet<Pair<Date,Date>> dateRanges = getValidTargetDates(Collections.emptySet(), Collections.emptySet());
+
+        // TODO - Determine how restrictive we should be when evaluating whether or not to retain a date in the target date range, i.e. should we refrain from
+        // querying on a date if any index holes are seen on that day (current implementation) or only when we see index holes on a date for all fields and
+        // datatypes established for a query? What about when a query will include all fields and/or all datatypes?
```
ivakegg: I think we should only federate if the we see index holes for the fields in the query itself.  There is a visitor to return all the fields in the query, then apply the query model, and then you can determine the index holes to take into account.

warehouse/query-core/src/main/java/datawave/query/planner/FederatedQueryPlanner.java
```diff
@@ -0,0 +1,315 @@
...
+            ShardQueryConfiguration configCopy = new ShardQueryConfiguration(config);
+            configCopy.setBeginDate(dateRange.getLeft());
+            configCopy.setEndDate(dateRange.getRight());
+
+            // Create a copy of the original default query planner, and process the query with the new date range.
+            DefaultQueryPlanner planner = new DefaultQueryPlanner(originalPlanner);
+            results.addIterable(planner.process(config, query, settings, scannerFactory));
+
+            // Update the planned script.
+            // TODO - Verify if this is the way we want to consolidate the planned script for the multiple queries, and if ; works as a delimiter.
```
ivakegg: This is a good question.  Luke has worked on having the subplans added to the query metrics.  Perhaps that is enough to be able to see all of the variants.  I think just using the first query plan as the "planned script" for the federated query logic is probably sufficient.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/planner/FederatedQueryPlanner.java
lbschanno: What's the name of the visitor class? Would we need a different approach for any fields specified via a regex pattern?

warehouse/query-core/src/main/java/datawave/query/planner/FederatedQueryPlanner.java
lbschanno: Understood, I'll incorporate the datatype filters.

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
lbschanno: I took this approach initially because I wasn't sure what other logic in ShardQueryLogic would be dependent on having access directly to the original DefaultQueryPlanner instance. I'll switch to your suggested approach and see if I run into any issues.

warehouse/query-core/src/main/java/datawave/query/planner/FederatedQueryPlanner.java
lbschanno: Understood, I'll do that then.

warehouse/query-core/src/main/java/datawave/query/planner/FederatedQueryPlanner.java
apmoriarty: What about an unfielded term that expands into a field with an index hole? It seems like we should take that case into account.

ivakegg: oh right.... we would need to handle unfielded expansion.  hmm

lbschanno:
I have verified that the FederatedQueryPlanner is properly determining the date ranges that need to be used when creating the sub-queries. However, I am having difficulties getting the results from the sub-queries to return bundled together. At this point I think I'm running into a deficit of Datawave-specific knowledge on where exactly I need to make changes to support this.

I have added  `FederatedShardQueryConfiguration` and `ChainedScheduler` to ensure that a scheduler is created based off the configs for each sub-query, but so far, only results from the first sub-query are being returned (see `FederatedQueryTest`). I would appreciate any feedback on this.

warehouse/query-core/src/main/java/datawave/query/planner/FederatedQueryPlanner.java
```diff
@@ -0,0 +1,440 @@
...
+            try {
+                CloseableIterable<QueryData> queryData = subPlan.process(configCopy, query, settings, scannerFactory);
+                results.addIterable(queryData);
+                ShardQueryConfiguration configCopyCopy = new ShardQueryConfiguration(configCopy);
+                if (queryData != null) {
+                    configCopyCopy.setQueries(queryData.iterator());
+                }
+                configCopyCopy.setQueryString(subPlan.getPlannedScript());
+                federatedConfig.addConfig(configCopyCopy);
+            } catch (Exception e) {
```
ivakegg, line 343: Exception or DatawaveQueryException ?

warehouse/query-core/src/main/java/datawave/query/planner/FederatedQueryPlanner.java
```diff
@@ -0,0 +1,440 @@
...
+            ShardQueryConfiguration configCopy = new ShardQueryConfiguration(originalConfig);
+            configCopy.setBeginDate(dateRange.getLeft());
+            configCopy.setEndDate(dateRange.getRight());
+
+            // Create a copy of the original default query planner, and process the query with the new date range.
+            DefaultQueryPlanner subPlan = new DefaultQueryPlanner(queryPlanner);
+
+            // TODO - Check if it's okay to add this to the copy constructor.
+            // TODO - Check if anything else needs to be added to the copy constructor.
+            subPlan.setTransformRules(queryPlanner.getTransformRules());
```
ivakegg: I think this probably needs to be added to the copy constructor.

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -520,17 +533,79 @@ public void setupQuery(GenericQueryConfiguration genericConfig) throws Exception
...
+
+    private void setupFederatedQuery() {
+        List<ShardQueryConfiguration> configs = federatedConfig.getConfigs();
+        if (configs.size() == 1) {
+            setupQuery(configs.get(0));
+        } else {
+            ShardQueryConfiguration firstConfig = configs.get(0);
+            // If the first query could not be run, assume the remaining could not either. Return no results.
+            // TODO - Is this assumption correct? Should we check all sub-configs if any could be run?
+            if (!firstConfig.canRunQuery()) {
```
ivakegg: For what this entails today, only calling 1 is probably ok.  However it might be more robust to call em all.  If one of them cannot run, then we can't run.

ivakegg:
From @lbschanno 

I have been working on getting tests to pass when the FederatedQueryPlanner is the default query planner. Some cases have shown that it may not be enough to simply use the first config returned from a sub-query as the finalized query string.

For instance, in MaxExpansionIndexOnlyQueryTest.testMaxAnyField(), the sub-queries have the following results when the max value expansion threshold is set to 20:

Result 0 over 2015/04/04-2015/10/09
Query String: false
Query Data Iterable: Empty

Result 1 over 2015/10/10-2015/10/10
Query String: (CODE == 'b-code' || CITY == 'b-city' || CITY == 'b-2' || CITY == 'b-1' || STATE == 'b-state') && (CODE == 'a-code' || CITY == 'a-1' || STATE == 'a-state' || STATE == 'a-s2')
Query Data Iterable: contains 3 query datas

Sub-query 2 over 2015/10/11-2015/11/11
Query String: false
Query Data Iterable: empty

In MaxExpansionIndexOnlyQueryTest.testMaxValueRegexIndexOnly(), we receive the following sub-query results when the max expansion threshold is set to 20:

Sub-query 0 over 2015/04/04-2015/10/09
Query String: CITY == 'a-1' && STATE =~ 'b.*'
Query Data Iterable: Empty

Sub-query 1 over 2015/10/10-2015/10/10
Query String: CITY == 'a-1' && (STATE == 'b3-state' || STATE == 'b-state' || STATE == 'bi-s' || STATE == 'b2-state' || STATE == 'ba-s2')
Query Data Iterable: contains 2 query datas

Sub-query 2 over 2015/10/11-2015/11/11
Query String: CITY == 'a-1' && STATE =~ 'b.*'
Query Data Iterable: Empty

In MaxExpansionIndexOnlyQueryTest.testMaxValueNegAnyField(), we receive the following sub-query results when the max expansion threshold is set to 10.

Sub-query 0 over 2015/04/04-2015/10/09
Query String: false
Query Data Iterable: Empty

Sub-query 1 over 2015/10/10-2015/10/10
Query String: STATE == 'b-state' && !(((_Delayed_ = true) && (_ANYFIELD_ =~ 'a.*')) || CODE == 'a-code' || CITY == 'a-1' || STATE == 'a-state' || STATE == 'a-s2')
Query Data Iterable: contains 4 query datas

Sub-query 2 over 2015/10/11-2015/11/11
Query String: false
Query Data Iterable: Empty

I have seen some similar results for MaxExpansionQueryTest and AnyFieldQueryTest. Given that we can have differing query strings, how do you want to handle determining which query string to set in the original config that's passed into the FederatedQueryPlanner.process() method? Do we need the query string unique to the query data iterable returned from each sub-queries when setting up the schedulers in ShardQueryLogic.setUpQuery(GenericQueryConfiguration config)? Is there somewhere else where we need to know the specific query strings from each sub-query?

I have pushed an update that adds tests to MaxExpansionIndexOnlyQueryTest with versions of each test using either the DefaultQueryPlanner or FederatedQueryPlanner so that you can see the results for yourself.

I suppose another question would be do the results above even look correct to you?

ivakegg:
For documentation purposes, here is the response in the conversation we had:
1) The tests that test the query plan would need to be changed to handle the one returned by the federated query planner.
2) The federated query plan that gets returned should probably concatenate the plans (as a unique set) into something like this:
     ((plan = 1) && (<planA>)) || (plan = 2) && (<planB>)) ...
     simply use <planA> if only one plan is in the set
3) We can work later to allow the query metrics to handle muliple top level plans after the sub-plan work is pulled in and proven viable.

ivakegg (dismissed):

warehouse/query-core/src/test/java/datawave/query/QueryPlanTest.java
```diff
@@ -99,7 +102,7 @@ public void planInMetricsAfterInvalidQueryException() throws Exception {
     @Test
     public void planInMetricsAfterMissingIndexException() throws Exception {
         String query = "CITY == 'london' && CITY != 'london'";
-        String expectedPlan = "CITY == 'london' && !(CITY == 'london')";
+        String expectedPlan = "CITY == 'london' && CITY != 'london'";
```
hlgp: why should these have changed?

warehouse/query-core/src/test/java/datawave/query/planner/FederatedQueryTest.java
```diff
@@ -0,0 +1,518 @@
...
+        expectEvents("20130102", FieldIndexHoleDataIngest.corleoneUID, FieldIndexHoleDataIngest.caponeUID, FieldIndexHoleDataIngest.sopranoUID);
+        expectEvents("20130103", FieldIndexHoleDataIngest.corleoneUID, FieldIndexHoleDataIngest.caponeUID, FieldIndexHoleDataIngest.sopranoUID);
+        expectEvents("20130104", FieldIndexHoleDataIngest.corleoneUID, FieldIndexHoleDataIngest.caponeUID, FieldIndexHoleDataIngest.sopranoUID);
+        expectEvents("20130105", FieldIndexHoleDataIngest.corleoneUID, FieldIndexHoleDataIngest.caponeUID, FieldIndexHoleDataIngest.sopranoUID);
+
+        assertQueryResults();
+    }
+
+    /**
+     * Test a field index hole min threshold that some differing counts do and do not meet.
```
hlgp, line 470: Great tests

warehouse/query-core/src/test/java/datawave/query/testframework/GroupingAccumuloWriter.java
```diff
@@ -114,10 +114,10 @@ private void writeMetaData(BatchWriterConfig bwConfig, final List<Map.Entry<Mult
                         Value colVal = new Value(SummingCombiner.VAR_LEN_ENCODER.encode((long) rawData.get(column).size()));
                         mut.put(ColumnFamilyConstants.COLF_F, new Text(this.dataType + NULL_SEP + shardDate), colVal);
                         if (this.fieldConfig.getIndexFields().contains(column)) {
-                            mut.put(ColumnFamilyConstants.COLF_I, dtText, EMPTY_VALUE);
+                            mut.put(ColumnFamilyConstants.COLF_I, new Text(this.dataType + NULL_SEP + shardDate), colVal);
```
hlgp: Shouldn't these tests still work without the addition of the counts for i, ri for backward compatibility?

warehouse/query-core/src/test/java/datawave/query/QueryPlanTest.java
apmoriarty: Actually, could we keep this change in addition to the original? That way we test the `not-eq` and `ne` cases

warehouse/query-core/src/test/java/datawave/query/QueryPlanTest.java
```diff
@@ -112,7 +115,7 @@ public void planInMetricsAfterMissingIndexException() throws Exception {
     @Test
     public void planInMetricsAfterTableNotFoundException() throws Exception {
         String query = Constants.ANY_FIELD + " != " + "'" + TestCities.london + "'";
-        String expectedPlan = "!(_ANYFIELD_ == 'london')";
+        String expectedPlan = "_ANYFIELD_ != 'london'";
```
apmoriarty, line 161: same comment as above, can we run both versions of not equals

warehouse/query-core/src/test/java/datawave/query/QueryPlanTest.java
ivakegg: I will look at this one.

warehouse/query-core/src/test/java/datawave/query/testframework/GroupingAccumuloWriter.java
ivakegg: hmm.... maybe.  I will look into in

warehouse/query-core/src/test/java/datawave/query/QueryPlanTest.java
ivakegg: +1

warehouse/query-core/src/test/java/datawave/query/testframework/GroupingAccumuloWriter.java
lbschanno: I jumped the gun when doing this. After reverting these changes, all tests still pass with the original data format.

warehouse/query-core/src/test/java/datawave/query/QueryPlanTest.java
lbschanno: These changed initially after setting the logic to use the federated query planner by default because the ferderated query planner's planned script was not getting updated with the result from `DefaultQueryPlanner.process()` when a RuntimeException was thrown. After correcting this logic, the query planner now reflects the correct query plan. There are now four variants of this test case, two where the logic uses a federated query planner for not-eq and ne, and two where the default query planner is used for not-eq and ne.

warehouse/query-core/src/test/java/datawave/query/QueryPlanTest.java
lbschanno: There are now four variants of this test case, two where the logic uses a federated query planner to test `not-eq` and `ne`, and two where the logic uses a default query planner to test `not-eq` and `ne`. Unlike the asserts for the `QueryPlanTest.planInMetricsAfterMissingIndexException()` variations, we should not expect the query plan to be updated with the result from `DefaultQueryPlanner.process()` because the exception is thrown earlier when attempting to find field index holes from the missing metadata table.

warehouse/query-core/src/test/java/datawave/query/testframework/GroupingAccumuloWriter.java
hlgp: I would still like to see the tests without the addition of the new format to prevent us from accidentally breaking the old format handling in future.

hlgp (dismissed):

warehouse/query-core/src/test/java/datawave/query/testframework/GroupingAccumuloWriter.java
lbschanno: Reverted remaining changes to test data format. I believe I caught everything and that all tests should be passing with the original test data format.

ivakegg (approved):

hlgp (approved):

PR #2217, opened 2024-01-13
Updating excerpt callout based on added testing

hlgp:
Added TermOffsets to WiseGuysIngest for further testing


mineralntl (approved):
Updated hit term phrase call out logic after reviewing these test cases

apmoriarty (approved):

PR #2218, opened 2024-01-16
Avoid re-normalizing unfielded lookup results

### LINKED ISSUE #877: Avoid normalizing unfielded lookup results

foster33:
This PR closes #877 

- Adds a configurable cache that will prevent a field / term pair, that's actual value was already found in the index, from being re-normalized (if only one normalizer is being used).
- Adds a test for the cache


foster33:
I think additional tests should be added to ensure we are covering more cases... But I have removed the WIP and this is now ready for review.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/FieldNameIndexLookup.java
```diff
@@ -79,7 +79,6 @@ public void submit() {
             try {
                 if (!fields.isEmpty()) {
                     for (String term : terms) {
-
```
ivakegg: No actual changes here.  Please revert this class.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ShardIndexQueryTableStaticMethods.java
```diff
@@ -113,7 +113,6 @@ public static IndexLookup normalizeQueryTerm(JexlNode node, ShardQueryConfigurat
     public static IndexLookup normalizeQueryTerm(String literal, ShardQueryConfiguration config, ScannerFactory scannerFactory, Set<String> expansionFields,
                     Set<Type<?>> dataTypes, MetadataHelper helperRef, ExecutorService execService) throws TableNotFoundException {
         Set<String> terms = Sets.newHashSet(literal);
-
```
ivakegg: No real changes here.  Please revert this class.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ExpandedFieldCache.java
```diff
@@ -0,0 +1,46 @@
+package datawave.query.jexl.lookups;
+
+import java.util.Map;
+
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Multimap;
+
+public class ExpandedFieldCache {
```
ivakegg, line 12: A class level comment as to what this is used for please

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/ExpandMultiNormalizedTerms.java
```diff
@@ -216,37 +220,95 @@ private boolean isAllDropped(JexlNode node) {
...
+    }
+
     private JexlNode expandRangeForNormalizers(LiteralRange<?> range, JexlNode node) {
         Map<String,JexlNode> aliasedBounds = new HashMap<>();
         String field = range.getFieldName();
 
         // Get all of the indexed or normalized dataTypes for the field name
         Set<Type<?>> dataTypes = Sets.newHashSet(config.getQueryFieldsDatatypes().get(field));
         dataTypes.addAll(config.getNormalizedFieldsDatatypes().get(field));
+        boolean shouldCheckPreviouslyExpandedFieldCache = dataTypes.size() == 1;
```
ivakegg: We could include the && config.isCachePreviouslyExpandedFields() into this boolean avoiding that check down below.

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
```diff
@@ -106,6 +106,8 @@ public class ShardQueryConfiguration extends GenericQueryConfiguration implement
     private boolean pruneQueryByIngestTypes = false;
     // should this query reduce the set of fields prior to serialization
     private boolean reduceQueryFields = false;
+    // should previously expanded fields be cached to prevent the re-normalizing of actual values
+    private boolean cachePreviouslyExpandedFields = false;
```
ivakegg: Can we make the default true?

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ExpandedFieldCache.java
```diff
@@ -0,0 +1,54 @@
...
+                if (value.contains(literal)) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+    public void addExpansion(IndexLookupMap fieldstoterms) {
+        if (!containsExpansionsFor(fieldstoterms)) {
```
ivakegg, line 40: This does not look correct to me.  The test only tests of one of the terms actually exists in the cache.  So if only one of them exists then it will not add any of the fieldstoterms?  I dont think the test is desired here.

ivakegg (changes requested):

Issue #2104, opened 2023-09-28
A reasonable limit should exist for the number of ivarator terms in a query


ivakegg:
Queries should fail if they have more than a configurable number of "ivarators" in the query.  Alternatively queries with over this limit should be refactored to combine ivarators if possible.

PR #2220, opened 2024-01-16
re #2104: ivarator limit

ivakegg:
* Added max ivarator terms configuration property
* Enforce max ivarator term limit in validateQuerySize
* Added test


keith-ratcliffe (dismissed):
LGTM

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -672,6 +678,17 @@ public static void validateQuerySize(String lastOperation, final JexlNode queryT
             throw new DatawaveFatalQueryException(qe);
         }
 
+        // now check whether we are over the ivarator limit
+        if (maxIvaratorThreshold >= 0) {
+            NodeTypeCount nodeCount = NodeTypeCountVisitor.countNodes(queryTree);
+            int totalIvarators = nodeCount.getTotal(ExceededValueThresholdMarkerJexlNode.class) + nodeCount.getTotal(ExceededOrThresholdMarkerJexlNode.class);
```
apmoriarty: Seems like the NodeTypeCountVisitor should optionally accept a list of class names as a filter..

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
ivakegg: The visitor still needs to visit every node.  To add an additional check at every node to determine whether it is something we are interested in actually may be more expensive, but that is merely a hypothesis.  I will try to test that out.

ivakegg: Well, after much testing it appears that adding a set to test against is neglegible in terms of addition time.  Adding that in.

apmoriarty (approved):

keith-ratcliffe (approved):

MiguelRicardos (approved):

PR #2224, opened 2024-01-19
Feature/compose

jwomeara:


contrib/datawave-quickstart/bin/services/datawave/install-ingest.sh
```diff
@@ -75,9 +75,9 @@ else
    [ ! -d ${ACCUMULO_HOME}/lib/ext ] && fatal "Unable to update Accumulo classpath. ${ACCUMULO_HOME}/lib/ext does not exist!"
    info "Removing any existing jars from ${ACCUMULO_HOME}/lib/ext"
    rm -f ${ACCUMULO_HOME}/lib/ext/*.jar
-   info "Copying DataWave jars into ${ACCUMULO_HOME}/lib/ext"
+   info "Copying DataWave jars into ${ACCUMULO_HOME}/lib and ${ACCUMULO_HOME}/lib/ext"
    if [ -d ${DW_DATAWAVE_INGEST_HOME}/accumulo-warehouse/lib ]; then
-      cp ${DW_DATAWAVE_INGEST_HOME}/accumulo-warehouse/lib/*.jar ${ACCUMULO_HOME}/lib/ext > /dev/null 2>&1
+      cp ${DW_DATAWAVE_INGEST_HOME}/accumulo-warehouse/lib/*.jar ${ACCUMULO_HOME}/lib > /dev/null 2>&1
```
jwomeara, line 95: @keith-ratcliffe this seemed like a bug in the original code, but maybe im missing something?

keith-ratcliffe: Hmm. The intent was to set up `lib/ext` as a VFS classpath, in conjunction with the accumulo bootstrap where it sets `general.vfs.context.classpath.extlib` to that dir in accumulo.properties

jwomeara: The next few lines handle the lib/ext path.  Are we saying we want both the accumulo-warehouse/lib and accumulo-warehouse/lib/ext jars copied to ${ACCUMULO_HOME}/lib/ext ?

I just figured the intent was to copy the lib contents to lib, and the lib/ext contents to lib/ext.  

If it's wrong, tell me how it's supposed to look and I'll change it.

keith-ratcliffe: Sorry, I follow now....

I do recall that at some point long ago there were jar(s) in our `${DW_DATAWAVE_INGEST_HOME}/accumulo-warehouse/lib`  that were problematic for accumulo's parent classloader, and so putting everything in the lib/ext VFS dir instead was a viable workaround...but it's been so long that I don't remember which jar(s).

So, your update does restore the original intent. Assuming that the quickstart's ingest and query tests still function,  then the issue, whatever it was, is now OBE

keith-ratcliffe (dismissed):

contrib/datawave-quickstart/bin/services/accumulo/bootstrap.sh
```diff
@@ -57,7 +66,9 @@ DW_ACCUMULO_VFS_DATAWAVE_DIR="/datawave/accumulo-vfs-classpath"
 # accumulo.properties (Format: <property-name>=<property-value>{<newline>})
 
 DW_ACCUMULO_PROPERTIES="## Sets location in HDFS where Accumulo will store data
-instance.volumes=${DW_HADOOP_DFS_URI}/accumulo
+instance.volumes=${DW_HADOOP_DFS_URI_CLIENT}/accumulo
+
+tserver.memory.maps.native.enabled=false
```
jwomeara, line 71: @keith-ratcliffe is this ok?  This has been a hangup for me a few times now when loading the quickstart on my local box (i.e. not in docker).  I added this for testing purposes but didn't intend to commit it.  What are your thoughts?

ivakegg (approved):

PR #2228, opened 2024-01-23
CompositeQueryLogic persists the max results calculated by getResultLimit()

apmoriarty:
CompositeQueryLogic now respects the max results calculated by getResultLimit(). This allows overriding the max results by user dn and system from.

### DIFF
--- a/web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
+++ b/web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
@@ -98,6 +98,9 @@ public class CompositeQueryLogic extends BaseQueryLogic<Object> {
 
         public void setMaxResults(long maxResults) {
             this.maxResults = maxResults;
+            if (maxResults > logic.getMaxResults() && logic.getMaxResults() != -1) {
+                logic.setMaxResults(maxResults);
+            }
         }
 
         public long getMaxResults() {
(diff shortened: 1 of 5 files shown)


web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
```diff
@@ -685,4 +688,15 @@ public CountDownLatch getStartLatch() {
     public CountDownLatch getCompletionLatch() {
         return completionLatch;
     }
+
+    @Override
+    public void setMaxResults(long maxResults) {
+        this.maxResults = maxResults;
+        super.setMaxResults(maxResults);
+        if (queryLogics != null) {
```
ivakegg: use getQueryLogics().values() instead please.

ivakegg (changes requested):

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
apmoriarty: Can do

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
```diff
@@ -685,4 +688,15 @@ public CountDownLatch getStartLatch() {
     public CountDownLatch getCompletionLatch() {
         return completionLatch;
     }
+
+    @Override
+    public void setMaxResults(long maxResults) {
+        this.maxResults = maxResults;
+        super.setMaxResults(maxResults);
+        if (getQueryLogics() != null) {
```
ivakegg: cannot return null

ivakegg (approved):

d-hwang (approved):

hlgp (approved):

PR #2229, opened 2024-01-24
Moved SSDeep classes into new ssdeep-common module

drewfarris:
This will allow these classes to be re-used more effectively across ingest, query and trends.


warehouse/ssdeep-common/src/main/java/datawave/util/ssdeep/Tuple2.java
FineAndDandy, line 1: I'm pretty sure this class already exists elsewhere in the code. Are you sure this can't be reused from elsewhere?

apmoriarty: It's in query core

apmoriarty (dismissed):

warehouse/ssdeep-common/src/main/java/datawave/util/ssdeep/Tuple2.java
FineAndDandy: I was actually thinking something like this: https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/tuple/package-summary.html

drewfarris: Thank you both for picking up on the `Tuple2` blasphemy that I should have known better that trying to propagate :)

I'll use `ImmutablePair` from commons-lang most likely.

apmoriarty (approved):

FineAndDandy (approved):

PR #2230, opened 2024-01-24
Improved SSDeep ingest and query test framework

drewfarris:
* includes complete ingest testframework for SSDeep data that includes both shard data and ssdeepIndex data.
* cleans up some sins of the past related to the faceted query logic leaking into the test framework.
* Adds protoype datawave ingest SSDeepIndexHandler, good enough for testing at this point.


warehouse/ingest-ssdeep/src/main/java/datawave/ingest/mapreduce/handler/ssdeep/SSDeepIndexHandler.java
```diff
@@ -0,0 +1,238 @@
...
+    protected Text ssdeepIndexTableName;
+
+    protected MarkingFunctions markingFunctions;
+
+    protected TaskAttemptContext taskAttemptContext;
+
+    protected Set<String> ssdeepFieldNames;
+
+    NGramByteHashGenerator nGramGenerator;
+    BucketAccumuloKeyGenerator accumuloKeyGenerator;
```
apmoriarty: make these protected?

apmoriarty (dismissed):

apmoriarty (dismissed):

apmoriarty (dismissed):

ivakegg:
I am attempting to review this but I am afraid I need an overview of the table structure and how this works in general.  I did not find any class level comments that might help guide me.  Can we add something as such.  Maybe a README.md somewhere?  If there is already such documentation please point me to it.

drewfarris:
@ivakegg (or for anyone else) - see: https://github.com/NationalSecurityAgency/datawave/wiki/SSDeep-In-Datawave

warehouse/ingest-ssdeep/src/main/java/datawave/ingest/mapreduce/handler/ssdeep/SSDeepIndexHandler.java
```diff
@@ -0,0 +1,238 @@
...
+    protected int[] getNonNullTableLoaderPriorities(Configuration conf, int defaultPriority, String... properties) {
+        if (properties.length % 2 != 0) {
+            throw new IllegalArgumentException("Received an odd number of properties, expected an even number, "
+                            + "each table name property should be followed by the table priority property");
+        }
+
+        int[] priorities = new int[properties.length / 2];
+        int index = 0;
+
+        for (int i = 0; i < properties.length; i += 2) {
```
MiguelRicardos: Inside of a loop, do the strings need to be declared as final?

Couldn't it be done like:
```
        for (int i = 0; i < properties.length; i += 2) {
            String tableNameProp = properties[i];
            String tablePriorityProp = properties[i + 1];
            String tableName = conf.get(tableNameProp, null);
            if (null != tableName)
                priorities[index++] = conf.getInt(tablePriorityProp, defaultPriority);
        }
      
     ``` 
     
     or even
     
     ```
        for (int i = 0; i < properties.length; i += 2) {
            String tablePriorityProp = properties[i + 1];
            String tableName = conf.get(properties[i], null);
            if (null != tableName)
                priorities[index++] = conf.getInt(tablePriorityProp, defaultPriority);
        }
      
     ```

File: warehouse/ingest-ssdeep/src/main/java/datawave/ingest/mapreduce/handler/ssdeep/SSDeepIndexHandler.java [hunk 3]
```diff
@@ -0,0 +1,238 @@
...
+    protected BucketAccumuloKeyGenerator accumuloKeyGenerator;
+
+    @Override
+    public void setup(TaskAttemptContext context) {
+        markingFunctions = MarkingFunctions.Factory.createMarkingFunctions();
+        taskAttemptContext = context;
+
+        Configuration conf = context.getConfiguration();
+
+        final String t = ConfigurationHelper.isNull(conf, DataTypeHelper.Properties.DATA_NAME, String.class);
```
MiguelRicardos: Can you use a more descriptive variable than t?

warehouse/ingest-ssdeep/src/main/java/datawave/ingest/mapreduce/handler/ssdeep/SSDeepIndexHandler.java
```diff
@@ -0,0 +1,238 @@
...
+
+    public static final int DEFAULT_SSDEEP_BUCKET_ENCODING_BASE = BucketAccumuloKeyGenerator.DEFAULT_BUCKET_ENCODING_BASE;
+
+    public static final String SSDEEP_BUCKET_ENCODING_LENGTH = ".ssdeepIndex.bucket.encoding.length";
+
+    public static final int DEFAULT_SSDEEP_BUCKET_ENCODING_LENGTH = BucketAccumuloKeyGenerator.DEFAULT_BUCKET_ENCODING_LENGTH;
+
+    public static final String SSDEEP_INDEX_NGRAM_SIZE = ".ssdeepIndex.ngram.size.min";
+
+    public static final int DEFAULT_SSDEEP_INDEX_NGRAM_SIZE = 7;
```
MiguelRicardos: Can you add some comments as to why the default NGRAM size is 7?

warehouse/ingest-ssdeep/src/main/java/datawave/ingest/mapreduce/handler/ssdeep/SSDeepIndexHandler.java
```diff
@@ -0,0 +1,238 @@
...
+
+    public static final String SSDEEP_MIN_HASH_SIZE = ".ssdeepIndex.chunk.size.min";
+
+    public static final int DEFAULT_SSDEEP_MIN_HASH_SIZE = 3;
```
MiguelRicardos: Can you add some comments as to why the minimum hash size is 3?

warehouse/ingest-ssdeep/src/main/java/datawave/ingest/mapreduce/handler/ssdeep/SSDeepIndexHandler.java
```diff
@@ -0,0 +1,238 @@
...
+    @Override
+    public String[] getTableNames(Configuration conf) {
+        return getNonNullTableNames(conf, SSDEEP_INDEX_TABLE_NAME);
+    }
+
+    @Override
+    public int[] getTableLoaderPriorities(Configuration conf) {
+        // @formatter:off
+        return getNonNullTableLoaderPriorities(
+                conf, 40,
```
MiguelRicardos: Can you add some comments explaining why 40 is the default?

warehouse/query-core/src/test/java/datawave/query/tables/ssdeep/SSDeepSimilarityQueryTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+    protected SSDeepSimilarityQueryLogic logic;
+
+    protected DatawavePrincipal principal;
+
+    @BeforeClass
+    public static void loadData() throws Exception {
+        final String tableName = "ssdeepIndex";
+
+        InMemoryInstance i = new InMemoryInstance("ssdeepTestInstance");
```
MiguelRicardos: Can you use a more descriptive variable name than i in this context?

warehouse/query-core/src/test/java/datawave/query/tables/facets/FacetQueryTestTableHelper.java
```diff
@@ -0,0 +1,69 @@
...
+    @Override
+    public void dumpTables(Authorizations auths) {
+        super.dumpTables(auths);
+
+        try {
+            dumpTable(FACET_TABLE_NAME, auths);
+            dumpTable(FACET_HASH_TABLE_NAME, auths);
+            dumpTable(FACET_METADATA_TABLE_NAME, auths);
+        } catch (TableNotFoundException e) {
+            // should not happen
```
MiguelRicardos: Can you add a bit more explanation why this should not happen?

warehouse/ingest-ssdeep/src/main/java/datawave/ingest/mapreduce/handler/ssdeep/SSDeepIndexHandler.java
ivakegg: I believe this is somewhat arbitrary.  The loader priority scheme should probably be removed altogether.  It was an attempt to smooth out the bulk loading process and avoid pounding the accumulo cluster with all tables at the same time, but in practice we set the priority of all the tables to be the same value.  I should look into removing that mechanism.

warehouse/ingest-ssdeep/src/main/java/datawave/ingest/mapreduce/handler/ssdeep/SSDeepIndexHandler.java
```diff
@@ -0,0 +1,238 @@
...
+        nGramGenerator = new NGramByteHashGenerator(ngramSize, bucketCount, minHashSize);
+
+        String[] fieldNameArray = conf.getStrings(type.typeName() + SSDEEP_FIELD_SET);
+        ssdeepFieldNames = new HashSet<>(List.of(fieldNameArray));
+        ssdeepIndexTableName = new Text(ConfigurationHelper.isNull(conf, SSDEEP_INDEX_TABLE_NAME, String.class));
+    }
+
+    @Override
+    public String[] getTableNames(Configuration conf) {
+        return getNonNullTableNames(conf, SSDEEP_INDEX_TABLE_NAME);
```
ivakegg: This seems like a horribly complicated way to return new String[] {ssdeepIndexTableName};  If that table name is null then the setup call above would have thrown an exception first.

warehouse/ingest-ssdeep/src/main/java/datawave/ingest/mapreduce/handler/ssdeep/SSDeepIndexHandler.java
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    @Override
+    public int[] getTableLoaderPriorities(Configuration conf) {
+        // @formatter:off
+        return getNonNullTableLoaderPriorities(
```
ivakegg: This seems like a horribly complicated way to return new int[] {conf.get(SSDEEP_INDEX_TABLE_LOADER_PRIORITY, 40)}

warehouse/ingest-ssdeep/src/main/java/datawave/ingest/table/config/SSDeepIndexTableConfigHelper.java
```diff
@@ -0,0 +1,60 @@
...
+        if (ssdeepIndexTableName == null) {
+            throw new IllegalArgumentException("No Facet Table names are defined");
+        }
+
+        // TODO: generic balancer, markings, bloom filters, locality groups.
+
+        if (tableName.equals(ssdeepIndexTableName)) {
+            this.tableType = SSDeepTableType.INDEX;
+        } else {
+            throw new IllegalArgumentException("Invalid Facet Table Definition Fot: " + tableName);
```
ivakegg: facet table?

warehouse/ingest-ssdeep/src/main/java/datawave/ingest/mapreduce/handler/ssdeep/SSDeepIndexHandler.java
drewfarris: Yes, was really just following the pattern from elsewhere (probably the Facet handler or other index handlers).

warehouse/ingest-ssdeep/src/main/java/datawave/ingest/mapreduce/handler/ssdeep/SSDeepIndexHandler.java
drewfarris: Once again, more copypasta. I can simplify both of these.

warehouse/ingest-ssdeep/src/main/java/datawave/ingest/table/config/SSDeepIndexTableConfigHelper.java
drewfarris: More copypasta.

warehouse/ingest-ssdeep/src/main/java/datawave/ingest/mapreduce/handler/ssdeep/SSDeepIndexHandler.java
drewfarris: There's really no need for final when declaring the variables inside a loop - in fact, I suspect the code with or without the final would generate the same bytecode. I'll removed the finals and the extra property variable as you suggested.

warehouse/ingest-ssdeep/src/main/java/datawave/ingest/mapreduce/handler/ssdeep/SSDeepIndexHandler.java
ivakegg: We should clean those up.... sigh.

ivakegg: perhaps a test case as well for coverage

ivakegg (changes requested):
see comments

warehouse/ingest-ssdeep/src/main/java/datawave/ingest/mapreduce/handler/ssdeep/SSDeepIndexHandler.java
drewfarris: Resolved this by referencing a constant in the NGramGenerator and adding a comment there about the choice of an ngram length of 7.

warehouse/query-core/src/test/java/datawave/query/tables/facets/FacetQueryTestTableHelper.java
drewfarris: Removed this try-catch, it wasn't needed

warehouse/ingest-ssdeep/src/main/java/datawave/ingest/mapreduce/handler/ssdeep/SSDeepIndexHandler.java
drewfarris: Also replace this with a reference to a constant in NGramGenerator and included a comment there about what this means. In reality, we choose something larger.

drewfarris:
Addressed @ivakegg and @MiguelRicardos 's comments. Please review when you guys are able.

ivakegg (dismissed):

ivakegg (approved):

MiguelRicardos:
looking right now.

MiguelRicardos (approved):

PR #2234, opened 2024-01-26
Fixes #2233 - Bug in BulkIngestMapFileLoaderTest

### LINKED ISSUE #2233: Mitigate race condition bug in BulkIngestMapFileLoaderTest's previous commit (569f29aa)

keith-ratcliffe:
- Mitigates race condition bug in testLoadFailedWithMiniAccumuloCluster
- Minor cleanup/refactoring for clarity


warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/job/BulkIngestMapFileLoaderTest.java
```diff
@@ -793,18 +793,21 @@ public void testLoadFailedWithMiniAccumuloCluster() {
             // Start the loader
             new Thread(processor, "map-file-watcher").start();
 
-            // Wait up to 30 secs for the bulk loader to log the failure
+            // Wait up to 30 secs for the bulk loader to log & mark the failure
             for (int i = 1; i <= 10; i++) {
                 Thread.sleep(3000);
-                if (log.contains(expectedMsg)) {
+                if (log.contains(expectedMsg) && Files.exists(expectedMarkerFile)) {
```
keith-ratcliffe, line 799: Marker file and expected msg are created in different threads in BIMFL, so need to account for both here prior to assertions below to mitigate failing `expectedMarkerFile` assertion

PR #2237, opened 2024-01-29
Enforce term threshold prior to range stream lookups

apmoriarty:
In addition to checking term thresholds at the beginning and very end of query planning, also check just prior to the range stream


drewfarris:
The implementation lgtm. How hard would it be be to add a unit test that validates that the threshold exception is thrown when we expect it to be thrown?

drewfarris (dismissed):

apmoriarty:
> The implementation lgtm. How hard would it be be to add a unit test that validates that the threshold exception is thrown when we expect it to be thrown?

That's easy enough to do. I think this may turn into another threshold that's independent of the initial and final max term threshold.

hlgp (approved):

hlgp:
lgtm.  One thought came to mind though about ensuring compatibility between the different term limits, e.g. a final term limit that is smaller than the initial or intermediate limits might merit a warning of "oy mate, that doesn't make a lot of sense.  might want to double check those configs"

MiguelRicardos (approved):

PR #2238, opened 2024-01-30
Limit Query Logging in RangeStream

apmoriarty:
Optionally limit the number of lines printed when streaming queries from the global index

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
+++ b/warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
@@ -549,6 +549,12 @@ public class ShardQueryConfiguration extends GenericQueryConfiguration implement
     private DocumentScannerConfig documentScannerConfig;
 
     /**
+     * The maximum number of lines to print when streaming the query from the global index. Useful for limiting the logging footprint of large queries when
+     * debug logging is enabled.
+     */
+    private int maxLinesToPrint = -1;
+
+    /**
      * Default constructor
      */
     public ShardQueryConfiguration() {
@@ -800,6 +806,7 @@ public class ShardQueryConfiguration extends GenericQueryConfiguration implement
                         other.getNoExpansionIfCurrentDateTypes() == null ? null : Sets.newHashSet(other.getNoExpansionIfCurrentDateTypes()));
         this.setUseDocumentScheduler(other.isUseDocumentScheduler());
         this.setDocumentScannerConfig(other.getDocumentScannerConfig());
+        this.setMaxLinesToPrint(other.getMaxLinesToPrint());
     }
 
     /**
@@ -3384,5 +3391,13 @@ public class ShardQueryConfiguration extends GenericQueryConfiguration implement
 
     public void setUseDocumentScheduler(boolean useDocumentScheduler) {
         this.useDocumentScheduler = useDocumentScheduler;
+    }
+
+    public int getMaxLinesToPrint() {
+        return maxLinesToPrint;
+    }
+
+    public void setMaxLinesToPrint(int maxLinesToPrint) {
+        this.maxLinesToPrint = maxLinesToPrint;
     }
 }
--- a/warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
+++ b/warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
@@ -155,6 +155,9 @@ public class RangeStream extends BaseVisitor implements CloseableIterable<QueryP
 
     protected NumShardFinder numShardFinder;
 
+    private int maxLinesToPrint = -1;
+    private int linesPrinted = 0;
+
     public RangeStream(ShardQueryConfiguration config, ScannerFactory scanners, MetadataHelper metadataHelper) {
         this.config = config;
         this.scanners = scanners;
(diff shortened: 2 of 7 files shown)


warehouse/query-core/src/test/resources/datawave/query/QueryLogicFactory.xml
```diff
@@ -236,6 +236,7 @@
         <property name="numRangesToBuffer" value="15" />
         <property name="rangeBufferTimeoutMillis" value="15000" />
         <property name="rangeBufferPollMillis" value="100" />
+        <property name="maxLinesToPrint" value="100"/>
```
FineAndDandy: default to unlimited?

apmoriarty: Sure, I don't mind doing that for this QLF.xml. I think I'll leave the configuration in the EventQueryLogicFactory so there's at least some unit tests using this code path.

warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
```diff
@@ -250,8 +257,12 @@ public Iterator<QueryPlan> iterator() {
 
                 if (log.isDebugEnabled()) {
                     log.debug("Query returned a stream with a context of " + this.context);
+                    int count = 0;
                     for (String line : StringUtils.split(queryStream.getContextDebug(), '\n')) {
                         log.debug(line);
+                        if (maxLinesToPrint > 0 && ++count > maxLinesToPrint) {
```
drewfarris, line 298: Consider adding a final log line that indicates that the output was truncated because it hit this max. This way we can differentiate between cases where the log didn't have what we expected and the log was truncated.

warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
```diff
@@ -205,8 +208,12 @@ public CloseableIterable<QueryPlan> streamPlans(JexlNode script) {
         if (log.isDebugEnabled()) {
             log.debug("Query returned a stream with a context of " + this.context);
             if (queryStream != null) {
+                int count = 0;
                 for (String line : StringUtils.split(queryStream.getContextDebug(), '\n')) {
                     log.debug(line);
+                    if (maxLinesToPrint > 0 && ++count > maxLinesToPrint) {
```
drewfarris, line 226: See comment from line 263 re: adding a log message to indicate truncation

warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
```diff
@@ -475,22 +506,30 @@ public ScannerStream visit(ASTEQNode node, Object data) {
...
 
         // Check if field is not indexed
+        linesPrinted++;
         if (!isIndexed(fieldName, config.getIndexedFields())) {
             try {
                 if (this.getAllFieldsFromHelper().contains(fieldName)) {
-                    log.debug("{\"" + fieldName + "\": \"" + literal + "\"} is not indexed.");
+                    if (maxLinesToPrint > 0 && linesPrinted > maxLinesToPrint) {
+                        log.debug("{\"" + fieldName + "\": \"" + literal + "\"} is not indexed.");
+                    }
```
drewfarris, line 554: Likewise it could be useful to add a single log message that states we exceeded the threshold and won't print any more, but that's less straightforward to implement here. Alternately, logging at trace level after we hit the threshold could be an alternate approach because if we _do_ need the full log lines for some reason on an operational system, we can dynamically change the log level. Otherwise the only alternative would be to patch the configuration, restart and re-run the query.

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
```diff
@@ -451,6 +451,12 @@ public class ShardQueryConfiguration extends GenericQueryConfiguration implement
      */
     private boolean pruneQueryOptions = false;
 
+    /**
+     * The maximum number of lines to print when streaming the query from the global index. Useful for limiting the logging footprint of large queries when
+     * debug logging is enabled.
+     */
+    private int maxLinesToPrint = -1;
```
drewfarris, line 555: Consider a slightly more descriptive variable name? (e.g., maxDebugLogLines)

warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
apmoriarty: good call, I'll add that in the next commit

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
apmoriarty: I'm not certain that being more descriptive is better in this case. Right now this mostly affects log messages at the debug level, but this could (and should) be used to drive the number of lines printed elsewhere in query planning. Those other instances may be at the info or warn levels.

warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
apmoriarty: I like flipping to trace logging if the threshold is exceeded

drewfarris (dismissed):

jalphonso (approved):

hlgp (approved):

alerman (approved):

PR #2243, opened 2024-01-31
Refactor QueryPlan

apmoriarty:
Refactor QueryPlan to be more decoupled, deprecate multiple constructors in favor of builder style methods.
- QueryPlan hashcode is built on demand and invalidated when any setter method is called
- updated existing code to use non-deprecated constructors

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/index/lookup/ShardRangeStream.java
+++ b/warehouse/query-core/src/main/java/datawave/query/index/lookup/ShardRangeStream.java
@@ -1,5 +1,6 @@
 package datawave.query.index.lookup;
 
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
 import java.util.Iterator;
@@ -122,8 +123,14 @@ public class ShardRangeStream extends RangeStream {
         }
 
         public QueryPlan apply(Entry<Key,Value> entry) {
-            return new QueryPlan(node, new Range(new Key(entry.getKey().getRow(), entry.getKey().getColumnFamily()), true,
-                            entry.getKey().followingKey(PartialKey.ROW_COLFAM), false));
+            Key start = new Key(entry.getKey().getRow(), entry.getKey().getColumnFamily());
+            Key end = entry.getKey().followingKey(PartialKey.ROW_COLFAM);
+            Range range = new Range(start, true, end, false);
+            //  @formatter:off
+            return new QueryPlan()
+                            .withQueryTree(node)
+                            .withRanges(Collections.singleton(range));
+            //  @formatter:on
         }
     }
 }
--- a/warehouse/query-core/src/main/java/datawave/query/index/lookup/ShardRangeStream.java
+++ b/warehouse/query-core/src/main/java/datawave/query/index/lookup/ShardRangeStream.java
@@ -1,6 +1,5 @@
 package datawave.query.index.lookup;
 
-import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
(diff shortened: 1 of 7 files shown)


warehouse/query-core/src/main/java/datawave/query/scheduler/PushdownFunction.java
```diff
@@ -152,11 +158,33 @@ protected void redistributeQueries(Multimap<String,QueryPlan> serverPlan, Tablet
...
                 }
 
-                if (log.isTraceEnabled())
+                if (log.isTraceEnabled()) {
                     log.trace("Adding query tree " + JexlStringBuildingVisitor.buildQuery(currentPlan.getQueryTree()) + " " + currentPlan.getSettings().size()
                                     + " for " + server);
+                }
+
+                // update settings before building new query plan
+                List<IteratorSetting> updatedSettings = new ArrayList<>();
```
ivakegg, line 167: Before we were simply passing the original settings on.  Why are we reconstituting this here?

apmoriarty: This is a lift and shift of code that happens inside the constructor `public QueryPlan(JexlNode queryTree, Iterable<Range> rangeIter, List<IteratorSetting> settings, Collection<String> columnFamilies)`.

I didn't particularly care that iterator setting updates were hidden inside a single constructor, so I moved it out to be more explicit about what's happening.

warehouse/query-core/src/main/java/datawave/query/index/lookup/ShardRangeStream.java
```diff
@@ -1,5 +1,6 @@
 package datawave.query.index.lookup;
 
+import java.util.Collection;
```
avgAGB: This import doesn't appear to be used anywhere with this class (Collections is though) so it could be removed

warehouse/query-core/src/main/java/datawave/query/index/lookup/ShardRangeStream.java
```diff
@@ -122,8 +123,14 @@ public FieldIndexParser(JexlNode node) {
...
-            return new QueryPlan(node, new Range(new Key(entry.getKey().getRow(), entry.getKey().getColumnFamily()), true,
-                            entry.getKey().followingKey(PartialKey.ROW_COLFAM), false));
+            Key start = new Key(entry.getKey().getRow(), entry.getKey().getColumnFamily());
+            Key end = entry.getKey().followingKey(PartialKey.ROW_COLFAM);
+            Range range = new Range(start, true, end, false);
+            //  @formatter:off
+            return new QueryPlan()
+                            .withQueryTree(node)
+                            .withRanges(Collections.singleton(range));
+            //  @formatter:on
```
avgAGB, line 132: This seems cleaner, pretty nice.

warehouse/query-core/src/main/java/datawave/query/index/lookup/TupleToRange.java
```diff
@@ -1,5 +1,6 @@
 package datawave.query.index.lookup;
 
+import java.util.Collection;
```
avgAGB: Similar deal, this could be removed if you like.

apmoriarty: Good catch, thanks!

ivakegg (approved):

avgAGB (approved):

PR #2244, opened 2024-01-31
Refactor QueryData

apmoriarty:
Refactor QueryData to support optional builder like methods
- QueryData test no longer relies on mock objects
- QueryData now implements equals and hashcode methods
- QueryData hashcode is built on demand and invalidated when any setter method is called
- copy constructor now copies all fields

### DIFF
--- a/web-services/query/src/main/java/datawave/webservice/query/configuration/QueryData.java
+++ b/web-services/query/src/main/java/datawave/webservice/query/configuration/QueryData.java
@@ -144,7 +144,7 @@ public class QueryData {
 
     public void addIterator(IteratorSetting cfg) {
         this.settings.add(cfg);
-        hashCode = -1;
+        resetHashCode();
     }
 
     @Override
--- a/web-services/query/src/main/java/datawave/webservice/query/configuration/QueryData.java
+++ b/web-services/query/src/main/java/datawave/webservice/query/configuration/QueryData.java
@@ -2,43 +2,106 @@ package datawave.webservice.query.configuration;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
 
 import org.apache.accumulo.core.client.IteratorSetting;
 import org.apache.accumulo.core.data.Range;
-
-import com.google.common.collect.Lists;
-import com.google.common.collect.Sets;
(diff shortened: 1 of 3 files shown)


ivakegg (dismissed):

mineralntl (dismissed):

web-services/query/src/main/java/datawave/webservice/query/configuration/QueryData.java
```diff
@@ -2,43 +2,106 @@
...
 public class QueryData {
-    List<IteratorSetting> settings = Lists.newArrayList();
-    String query;
-    Collection<Range> ranges = Sets.newHashSet();
-    Collection<String> columnFamilies = Sets.newHashSet();
+    private String query;
+    private Collection<Range> ranges = new HashSet<>();
+    private Collection<String> columnFamilies = new HashSet<>();
+    private List<IteratorSetting> settings = new ArrayList<>();
+    private int hashCode = -1;
```
jschmidt10, line 21: `-1` is technically a valid hashCode. Not the end of the world but if something legitimately hashed to -1, then we would keep recomputing. Typically you'd use `-1` when a negative number is not feasible - like for a count. In this case, the behavior works b/c we will just recompute it again but a more robust check would be to have a boolean for the need to recompute the hash.

web-services/query/src/main/java/datawave/webservice/query/configuration/QueryData.java
```diff
@@ -67,19 +132,66 @@ public Collection<String> getColumnFamilies() {
...
     public void setRanges(Collection<Range> ranges) {
-        if (null != ranges)
-            this.ranges.addAll(ranges);
+        this.ranges = ranges;
+        resetHashCode();
     }
 
     public void addIterator(IteratorSetting cfg) {
         this.settings.add(cfg);
+        hashCode = -1;
```
jschmidt10:
```suggestion
       resetHashCode();
```

web-services/query/src/main/java/datawave/webservice/query/configuration/QueryData.java
ivakegg: +1

ivakegg (approved):

jschmidt10 (approved):

hlgp (approved):

PR #2250, opened 2024-01-31
Added remote edge query logic and tests

ivakegg:


warehouse/query-core/src/main/java/datawave/query/config/RemoteQueryConfiguration.java
```diff
@@ -89,6 +90,17 @@ public static RemoteQueryConfiguration create(RemoteEventQueryLogic remoteQueryL
...
+    /**
+     * Factory method that creates a RemoteQueryConfiguration deep copy from a RemoteQueryLogic
+     *
+     * @param remoteQueryLogic
+     *            - a configured RemoteQueryLogic
+     * @return - a RemoteQueryConfiguration
+     */
+    public static RemoteQueryConfiguration create(RemoteEdgeQueryLogic remoteQueryLogic) {
+        return create(remoteQueryLogic.getConfig());
+    }
```
drewfarris: Could you use the `RemoteQueryLogic` interface as the parameter type for this and thus have a single `create` method that applies to both the `RemoteEdgeQueryLogic` and the `RemoveEventQueryLogic`? You would have to add `getConfig()` to the RemoteQueryLogic interface though...

warehouse/query-core/src/main/java/datawave/query/tables/RemoteEdgeQueryLogic.java
drewfarris, line 1: It seems that there's a fair amount of duplication between `RemoteEventQueryLogic` and `RemoteEdgeQueryLogic` - would it make sense to consolidate somehow? Perhaps `RemoteQueryLogic` becomes an abstract class instead of an interface?

drewfarris (dismissed):
Added a couple comments for consideration, but lgtm. Could be merged as-is.

warehouse/query-core/src/main/java/datawave/query/config/RemoteQueryConfiguration.java
ivakegg: Thank you for reviewing.  That seemed like a simple enough add that it was not worth it.  I may remove those convience methods altogether.

warehouse/query-core/src/main/java/datawave/query/tables/RemoteEdgeQueryLogic.java
ivakegg: I will see if I can do that.

warehouse/query-core/src/main/java/datawave/query/tables/RemoteEdgeQueryLogic.java
```diff
@@ -0,0 +1,269 @@
...
+
+    @Override
+    public Set<String> getRequiredQueryParameters() {
+        return new ShardQueryLogic().getRequiredQueryParameters();
+    }
+
+    @Override
+    public Set<String> getExampleQueries() {
+        return new ShardQueryLogic().getExampleQueries();
+    }
```
apmoriarty: The standard EdgeQueryLogic implements both `getOptionalQueryParameters` and `getRequiredQueryParameters`, so I think these overrides could cause inconsistent behavior.

While the standard EdgeQueryLogic does not implement `getExampleQueries` in any meaningful way there may well be other extended query logics that do...personally I would like to see that method removed from the QueryLogic interface altogether, but that's a discussion for another time.

apmoriarty (changes requested):

warehouse/query-core/src/main/java/datawave/query/tables/RemoteEdgeQueryLogic.java
```diff
@@ -0,0 +1,269 @@
...
+            log.trace("Creating RemoteQueryLogic: " + System.identityHashCode(this));
+    }
+
+    /**
+     * Copy constructor
+     *
+     * @param other
+     *            - another ShardQueryLogic object
+     */
+    public RemoteEdgeQueryLogic(RemoteEdgeQueryLogic other) {
```
apmoriarty, line 44: IntelliJ flags this copy constructor as not copying `userOperations` and `transformerInstance`. Either copy those variables or add a comment explaining why these are excluded from the copy.

apmoriarty (changes requested):

warehouse/query-core/src/main/java/datawave/query/tables/RemoteEdgeQueryLogic.java
ivakegg: good catch.  This was a mistake on my part.  Agreed on the second part

warehouse/query-core/src/main/java/datawave/query/tables/RemoteEdgeQueryLogic.java
ivakegg: Done.... pulled out a BaseRemoteQueryLogic.

apmoriarty (approved):

drewfarris (approved):

PR #2253, opened 2024-02-01
Attributes build correct metadata key

apmoriarty:
Corrected several issues with how Attributes built their metadata key
- term frequency keys now build the correct column family of datatype/uid, the setMetadata method was looking at a TF key's column family. It now examines the column qualifier.
- field index keys could build incorrect metadata keys if the field value contained a null. changed the algorithm to iterate backwards through a field index key's column qualifier to avoid this problem

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
+++ b/warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
@@ -98,44 +98,46 @@ public abstract class Attribute<T extends Comparable<T>> implements WritableComp
     protected void setMetadata(Key key) {
         if (key == null) {
             this.metadata = null;
-        } else {
-            // convert the key to the form shard type\0uid cv, ts. Possible inputs are an event key, a fi key, or a tf key
-            final ByteSequence row = key.getRowData(), cf = key.getColumnFamilyData(), cv = key.getColumnVisibilityData();
-            if (isFieldIndex(cf)) {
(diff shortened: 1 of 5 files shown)


MiguelRicardos (dismissed):

warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
```diff
@@ -98,44 +98,46 @@ protected void setMetadata(ColumnVisibility vis, long ts) {
...
-                    }
+            return;
+        }
+
+        // convert the key to the form shard type\0uid cv, ts. Possible inputs are an event key, a fi key, or a tf key
+        final ByteSequence row = key.getRowData();
+        final ByteSequence cf = key.getColumnFamilyData();
+        final ByteSequence cv = key.getColumnVisibilityData();
+        if (isFieldIndex(cf)) {
+            // CQ is 'value\0datatype\0value
```
ivakegg: The value is at both ends?

ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
apmoriarty: good catch, I'll update that

ivakegg (approved):

avgAGB (approved):

PR #2256, opened 2024-02-01
Misc cleaning up of ingest scripts

mineralntl:
A couple of ingest scripts were incorrectly quoting regexs so in addition to fixing those, the legacy backticks were also removed in favor of the current $() syntax.


hlgp:
Still marked as Draft.   Anything it's waiting on?

mineralntl:
Nothing else to add - fell thru the cracks and that's on me; marking as ready to review

alerman (approved):

cogross (approved):

hlgp (approved):

PR #2257, opened 2024-02-02
Fix AddDateFilter to produce a flattened query tree

apmoriarty:
To see the original error, checkout integration and enable AST validation and run `ContentFunctionQueryTest`


warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -1945,7 +1946,11 @@ public ASTJexlScript addDateFilters(final ASTJexlScript queryTree, ScannerFactor
             // now for each field, add an expression to filter that date
             List<JexlNode> andChildren = new ArrayList<>();
             for (int i = 0; i < queryTree.jjtGetNumChildren(); i++) {
-                andChildren.add(JexlNodeFactory.createExpression(queryTree.jjtGetChild(i)));
+                if (queryTree.jjtGetChild(i) instanceof ASTAndNode) {
+                    andChildren.add(queryTree.jjtGetChild(i));
+                } else {
+                    andChildren.add(JexlNodeFactory.createExpression(queryTree.jjtGetChild(i)));
```
avgAGB, line 1952: Might need a little formatting tweak

avgAGB (approved):

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -1946,7 +1947,11 @@ public ASTJexlScript addDateFilters(final ASTJexlScript queryTree, ScannerFactor
             // now for each field, add an expression to filter that date
             List<JexlNode> andChildren = new ArrayList<>();
             for (int i = 0; i < queryTree.jjtGetNumChildren(); i++) {
-                andChildren.add(JexlNodeFactory.createExpression(queryTree.jjtGetChild(i)));
+                if (queryTree.jjtGetChild(i) instanceof ASTAndNode) {
```
ivakegg, line 1950: Should this be checking for AndNode or ReferenceExpressionNode?

apmoriarty: That's an excellent question. I want to say it should, but I'm not sure how that would play with marker nodes.

apmoriarty: I added two tests and found no issues with the code as-is. One test had a marker node and the other had a nested union.

The reason is the call to `createReference()` will return the original node if it's a reference expression (preventing a double wrapped node)

ivakegg: Well as this is, it makes no sense to me to wrap it with a reference expression (or ensure it is a reference expression) IFF is is not an 'and' node.

ivakegg: ok then I need a little more context here.  What exactly is this fixing?

ivakegg: If we want a flattened tree, then why not simply include all of the children of the and node as children of this node instead?

ivakegg: likewidse if this is a reference expression wrapping only and nodes.

ivakegg: I can approve this, but I am not sure this is the full solution you want.

ivakegg (approved):

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
apmoriarty: Once the query tree is flattened after the initial parse, every future operation should produce a flattened query tree -- and the 'AddDateFilters' step will produce an unflattened query tree
```
# input
A && B

# output -- not flattened
(A && B) && filter

# should produce this -- which is flattened
A && B && filter
```
If the query was a top level union then it's certainly appropriate to add a reference expression.

PR #2259, opened 2024-02-02
Dependabot/maven/commons io commons io 2.7

ivakegg:


apmoriarty (dismissed):

drewfarris (dismissed):

keith-ratcliffe (approved):

ivakegg:
I think we should be upgrading to the same version as in accumulo: 2.13

PR #2269, opened 2024-02-07
Removed unused code

matthpeterson:


warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/define/EdgeDataBundle.java
```diff
@@ -148,12 +114,8 @@ private void initVertexMasking(final VertexValue vertex, final MaskedFieldHelper
         }
     }
 
-    public void setMaskedVisibility(ColumnVisibility maskedVisibility) {
-        this.maskedVisibility = maskedVisibility;
-    }
-
```
matthpeterson, line 154: This needs to be reintroduced

matthpeterson: reintroduced

hlgp:
OBE

PR #2271, opened 2024-02-07
Feature/ingest filter

ivakegg:
Added an record predicate mechanism to the EventMapper to allow filtering out events


drewfarris (dismissed):

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/EventMapperTest.java
```diff
@@ -199,6 +202,115 @@ public void shouldNotWriteRawFile() throws IOException, InterruptedException {
         assertEquals(4, written.size());
     }
 
+    public static class BaseEventPredicate implements RawRecordPredicate {
```
apmoriarty: Consider breaking this into it's own class so there's an "AlwaysTrue" default implementation available, similar to how we construct a TimeFilter.

If these objects are going to be stored in a set then any implementation should override equals and hashcode.

ivakegg: hmm, not a bad idea.  will do

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/EventMapper.java
```diff
@@ -463,12 +506,26 @@ public void map(K1 key, V1 value, Context context) throws IOException, Interrupt
...
-            getCounter(context, IngestInput.OLD_EVENT).increment(1);
-            return;
+        if (!value.fatalError()) {
+            // Determine whether the event date is greater than the interval. Excluding fatal error events.
+            Long myInterval = dataTypeDiscardIntervalCache.get(value.getDataType().typeName());
+            if (null != myInterval && 0L != myInterval && (value.getDate() < (now.get() - myInterval))) {
+                if (log.isInfoEnabled())
+                    log.info("Event with time " + value.getDate() + " older than specified interval of " + (now.get() - myInterval) + ", skipping...");
+                getCounter(context, IngestInput.OLD_EVENT).increment(1);
+                return;
```
matthpeterson: Looks good.  Perhaps the date type discard interval could be made into a predicate in the future.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/RawRecordPredicate.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * The implementation of the java util Predicate method. This method should not be overridden.
+     *
+     * @param record
+     *            The raw record container under review
+     * @return true if the record should be ingested
+     */
+    @Override
+    default boolean test(RawRecordContainer record) {
```
matthpeterson, line 51: I don't think shouldProcess is necessary.  There's no default implementation of test in java.util.function.Predicate, so it should be possible to remove shouldProcess and replace lines 51-53 with:
```
@Override
    default boolean test(RawRecordContainer record);
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/EventMapper.java
ivakegg: That is indeed my expectation here.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/RawRecordPredicate.java
ivakegg: I was trying to make it obvious whether the test was to filter the event or to process the event.

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/EventMapperTest.java
```diff
@@ -199,6 +202,115 @@ public void shouldNotWriteRawFile() throws IOException, InterruptedException {
...
+    public void testFailedPredicates() throws IOException, InterruptedException {
+        try {
+            conf.set("file." + EventMapper.RECORD_PREDICATES, CantConfigureEventPredicate.class.getName());
+            eventMapper.setup(mapContext);
+            eventMapper.map(new LongWritable(1), record, mapContext);
+        } finally {
+            BaseEventPredicate.reset();
+        }
+    }
+
```
matthpeterson, line 326: Nice tests!

ivakegg: Thank you

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/EventMapper.java
ivakegg: I might do that in this PR... still debating

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/EventMapperTest.java
ivakegg: Since we pass in the classes by class name, the same mechanism as done in TimeFilter will not work exactly.  Making the base implementation abstract however and creating an explicit "AlwaysProcessPredicate" class to make that more obvious.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/EventMapper.java
ivakegg: doing it...

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/EventMapper.java
```diff
@@ -100,11 +102,8 @@ public class EventMapper<K1,V1 extends RawRecordContainer,K2,V2> extends StatsDE
 
     private static final Logger log = Logger.getLogger(EventMapper.class);
 
-    /**
-     * number which will be used to evaluate whether or not an Event should be processed. If the Event.getEventDate() is greater than (now - interval) then it
-     * will be processed.
-     */
-    public static final String DISCARD_INTERVAL = "event.discard.interval";
+    public static final String DISCARD_INTERVAL = DataTypeDiscardIntervalPredicate.DISCARD_INTERVAL;
```
matthpeterson, line 106: Thanks for moving out the discard interval predicate.  It looks like this line can be removed.

matthpeterson, line 106:
```suggestion
```

matthpeterson: See suggested edit below.

ivakegg: Actually there are other classes that are looking for the constant in the EventMapper, so I was trying to minimize those changes.

ivakegg: Well maybe its only in one place, so I will do it.

apmoriarty (dismissed):

matthpeterson (dismissed):

matthpeterson (approved):

avgAGB (approved):

PR #2274, opened 2024-02-08
Enforce term limits during EdgeQueries

jschmidt10:
We currently do not limit terms correctly for EdgeQueries due to the condition being checked in the parent AND or OR node before we have visited the terms and incremented the termCount.

Updated to increment the term count and check as we go.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EdgeTableRangeBuildingVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EdgeTableRangeBuildingVisitor.java
@@ -168,11 +168,6 @@ public class EdgeTableRangeBuildingVisitor extends BaseVisitor implements EdgeMo
      */
     @Override
     public Object visit(ASTAndNode node, Object data) {
-        if (termCount > maxTerms) {
-            log.error("Query has too many terms");
-            throw new IllegalArgumentException("Too many search terms " + termCount);
-        }
-
         // run the visitor against all of the children
         List<List<? extends EdgeContext>> childContexts = new ArrayList<>(node.jjtGetNumChildren());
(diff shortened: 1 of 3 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EdgeTableRangeBuildingVisitor.java
```diff
@@ -410,10 +400,17 @@ private void checkNotExclusion(IdentityContext context, String msg) {
...
     public Object visit(ASTEQNode node, Object data) {
-        termCount++;
+        incrementTermCountAndCheck();
         return visitExpresionNode(node, EQUALS);
     }
 
+    private void incrementTermCountAndCheck() {
+        if (++termCount > maxTerms) {
+            log.error("Query has too many terms");
+            throw new IllegalArgumentException("Too many search terms " + termCount);
```
apmoriarty: Even though this will error out at one over the threshold, I would make this a little more descriptive with something like 
```
"Term count X exceeded term threshold of Y"
```

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/EdgeTableRangeBuildingVisitor.java
jschmidt10: Since we terminate as we go, it'll always just say like `Term count 11 exceeded term threshold of 10` even if the query has 100 terms. I updated the verbiage to hopefully be more understandable. Let me know what you think.

apmoriarty (dismissed):

ivakegg (dismissed):

ivakegg (approved):

keith-ratcliffe (approved):

PR #2276, opened 2024-02-09
Upgrade Surefire

apmoriarty:
Change Summary
- replaced `surefire-junit47` provider with `junit-jupiter-engine` and `junit-vintage-engine`
- `junit-jupiter-api`, `junit-jupiter-engine`, and `junit-vintage-engine` are all controlled by `version.junit-bom`
- surefire upgraded from `3.0.0-M6` to `3.2.5` to avoid a bug that couldn't parse junit versions with patch versions

### DIFF
--- a/common-test/pom.xml
+++ b/common-test/pom.xml
@@ -63,5 +63,10 @@
             <scope>provided</scope>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/common-test/pom.xml
+++ b/common-test/pom.xml
@@ -65,7 +65,7 @@
         </dependency>
         <dependency>
             <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-api</artifactId>
+            <artifactId>junit-jupiter-engine</artifactId>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -26,21 +26,6 @@
                 <version>${version.junit}</version>
                 <scope>test</scope>
             </dependency>
-            <dependency>
-                <groupId>org.junit.jupiter</groupId>
-                <artifactId>junit-jupiter-api</artifactId>
-                <scope>test</scope>
-            </dependency>
-            <dependency>
-                <groupId>org.junit.jupiter</groupId>
-                <artifactId>junit-jupiter-engine</artifactId>
-                <scope>test</scope>
-            </dependency>
-            <dependency>
-                <groupId>org.junit.vintage</groupId>
-                <artifactId>junit-vintage-engine</artifactId>
-                <scope>test</scope>
-            </dependency>
         </dependencies>
     </dependencyManagement>
     <dependencies>
@@ -52,18 +37,6 @@
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <version>${version.junit}</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-engine</artifactId>
-            <version>${version.junit.bom}</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.junit.vintage</groupId>
-            <artifactId>junit-vintage-engine</artifactId>
-            <version>${version.junit.bom}</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -18,11 +18,6 @@
         <developerConnection>scm:git:ssh://git@fixme/core.git</developerConnection>
         <tag>HEAD</tag>
     </scm>
-    <properties>
-        <version.junit>4.13.2</version.junit>
-        <version.junit.jupiter>5.8.2</version.junit.jupiter>
-        <version.junit.vintage>5.8.2</version.junit.vintage>
-    </properties>
     <dependencyManagement>
         <dependencies>
             <dependency>
@@ -34,19 +29,16 @@
             <dependency>
                 <groupId>org.junit.jupiter</groupId>
                 <artifactId>junit-jupiter-api</artifactId>
-                <version>${version.junit.jupiter}</version>
                 <scope>test</scope>
             </dependency>
             <dependency>
                 <groupId>org.junit.jupiter</groupId>
                 <artifactId>junit-jupiter-engine</artifactId>
-                <version>${version.junit.jupiter}</version>
                 <scope>test</scope>
             </dependency>
             <dependency>
                 <groupId>org.junit.vintage</groupId>
                 <artifactId>junit-vintage-engine</artifactId>
-                <version>${version.junit.vintage}</version>
                 <scope>test</scope>
             </dependency>
         </dependencies>
@@ -59,16 +51,19 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
+            <version>${version.junit}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-engine</artifactId>
+            <version>${version.junit.jupiter}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.junit.vintage</groupId>
             <artifactId>junit-vintage-engine</artifactId>
+            <version>${version.junit.vintage}</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
@@ -90,6 +85,18 @@
                         <redirectTestOutputToFile>true</redirectTestOutputToFile>
                         <runOrder>random</runOrder>
                     </configuration>
+                    <dependencies>
+                        <dependency>
+                            <groupId>org.junit.jupiter</groupId>
+                            <artifactId>junit-jupiter-engine</artifactId>
+                            <version>${version.junit.jupiter}</version>
+                        </dependency>
+                        <dependency>
+                            <groupId>org.junit.vintage</groupId>
+                            <artifactId>junit-vintage-engine</artifactId>
+                            <version>${version.junit.vintage}</version>
+                        </dependency>
+                    </dependencies>
                 </plugin>
             </plugins>
         </pluginManagement>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -57,13 +57,13 @@
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-engine</artifactId>
-            <version>${version.junit.jupiter}</version>
+            <version>${version.junit.bom}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.junit.vintage</groupId>
             <artifactId>junit-vintage-engine</artifactId>
-            <version>${version.junit.vintage}</version>
+            <version>${version.junit.bom}</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
@@ -89,12 +89,12 @@
                         <dependency>
                             <groupId>org.junit.jupiter</groupId>
                             <artifactId>junit-jupiter-engine</artifactId>
-                            <version>${version.junit.jupiter}</version>
+                            <version>${version.junit.bom}</version>
                         </dependency>
                         <dependency>
                             <groupId>org.junit.vintage</groupId>
                             <artifactId>junit-vintage-engine</artifactId>
-                            <version>${version.junit.vintage}</version>
+                            <version>${version.junit.bom}</version>
                         </dependency>
                     </dependencies>
                 </plugin>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -36,7 +36,6 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>${version.junit}</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -85,8 +85,6 @@
         <version.json>20231013</version.json>
         <version.jts>1.19.0</version.jts>
         <version.junit>4.13.2</version.junit>
-        <version.junit.jupiter>5.10.2</version.junit.jupiter>
-        <version.junit.vintage>5.10.2</version.junit.vintage>
         <version.jython>2.7.2b2</version.jython>
         <version.kryo>2.20</version.kryo>
         <version.kryonet>2.20</version.kryonet>
@@ -1060,6 +1058,13 @@
                 <scope>import</scope>
             </dependency>
             <dependency>
+                <groupId>org.junit</groupId>
+                <artifactId>junit-bom</artifactId>
+                <version>${version.junit.bom}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+            <dependency>
                 <groupId>org.springframework</groupId>
                 <artifactId>spring-framework-bom</artifactId>
                 <version>${version.springframework}</version>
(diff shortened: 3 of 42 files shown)


warehouse/accumulo-extensions/pom.xml
```diff
@@ -42,6 +42,7 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
+            <version>${version.junit}</version>
```
ivakegg: Do you need the version here?  Is it not inherited from the parent pom?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypePruningVisitor.java
```diff
@@ -192,7 +192,7 @@ public Object visit(ASTAndNode node, Object data) {
             pruningTypes = ingestTypes;
         }
 
-        for (int i = 0; i < node.jjtGetNumChildren(); i++) {
+        for (int i = node.jjtGetNumChildren() - 1; i >= 0; i--) {
```
ivakegg, line 199: why go through the children backwards?

ivakegg: Oh, because we are pruning children as we go I guess.  Perhaps a comment to that effect would be useful.

warehouse/accumulo-extensions/pom.xml
apmoriarty: IntelliJ flagged this instance of junit as not matching the parent pom

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypePruningVisitor.java
apmoriarty: Definitely worth a comment

pom.xml
```diff
@@ -1146,6 +1146,18 @@
                 <version>${version.junit.jupiter}</version>
                 <scope>test</scope>
             </dependency>
+            <dependency>
```
jwomeara: Can we just update the pom to use the junit bom like we do in datawave-microservice-parent?  https://github.com/NationalSecurityAgency/datawave-parent/blob/main/pom.xml#L64

warehouse/accumulo-extensions/pom.xml
jwomeara: is junit:junit in the dependency management section of the parent pom?  if not, it should be, and you shouldn't need the version here.

warehouse/core/pom.xml
```diff
@@ -198,6 +198,16 @@
...
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
```
jwomeara: junit-jupiter-engine has a dependency on junit-jupiter-api.  so you probably only need junit-jupiter-engine here.

jwomeara (changes requested):
Looks good for the most part.  In addition to these changes, can you update the following tests to stop mixing junit4 and junit5 classes? (e.g. Assert vs Assertions, org.junit.Test vs org.junit.jupiter.api.Test, etc.)

- DatawaveJexlContextTest
- AncestorChildExpansionIteratorTest
- TFKeyUtilTest
- FiKeyUtilTest
- TermFrequencyIteratorTest
- SplitSelectorExtractorTest

warehouse/core/pom.xml
apmoriarty: All the example docs I read had the explicit dependencies called out. I think that's better in the long run for readability.

pom.xml
apmoriarty: I'm leaving this as an explicit dependency.

warehouse/accumulo-extensions/pom.xml
ivakegg: I wonder if the test scope somehow mucks with that.  You may need to add the test scope for that in the management section as well.  Just a guess.

warehouse/core/pom.xml
ivakegg: Not sure I agree with that.  That requires us to ensure we are using compatible versons.  It is simply more maintainance on our part.

pom.xml
apmoriarty: I'll see about pulling in the `junit.bom`

warehouse/core/pom.xml
apmoriarty: I'll look at pulling that in

warehouse/pom.xml
```diff
@@ -475,6 +475,18 @@
                             <log4j.configuration>file:${project.build.testOutputDirectory}/log4j.properties</log4j.configuration>
                         </systemPropertyVariables>
                     </configuration>
+                    <dependencies>
+                        <dependency>
+                            <groupId>org.junit.jupiter</groupId>
+                            <artifactId>junit-jupiter-engine</artifactId>
+                            <version>${version.junit.bom}</version>
```
ivakegg, line 488: isn't this already defined in the parent pom?

pom.xml
```diff
@@ -1468,9 +1486,14 @@
                     </configuration>
                     <dependencies>
                         <dependency>
-                            <groupId>org.apache.maven.surefire</groupId>
-                            <artifactId>surefire-junit47</artifactId>
-                            <version>${version.surefire.plugin}</version>
+                            <groupId>org.junit.jupiter</groupId>
```
ivakegg, line 1474: I am wondering if we add these dependencies in the parent pom to the dependency section as well as the dependency management sections, then do we need to add these dependencies to every pom?

apmoriarty: I had tried that initially but the build fails locally saying that a version is required. Once that is resolved by adding the version here in the dependencies section the build fails later on with a failure to pull junit-bom from the packages repo. Something is amiss here because you should be able to declare `junit-bom` in the dependencies section with just the `groupId` and `artifactId`

pom.xml
```diff
@@ -1143,7 +1149,19 @@
...
+            <dependency>
+                <groupId>org.junit.jupiter</groupId>
+                <artifactId>junit-jupiter-engine</artifactId>
+                <version>${version.junit.bom}</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.junit.vintage</groupId>
+                <artifactId>junit-vintage-engine</artifactId>
+                <version>${version.junit.bom}</version>
```
jwomeara: These should be removed.  The junit-bom is importing these.  Keeping them defeats the purpose of using the bom

jwomeara (changes requested):

core/pom.xml
```diff
@@ -34,19 +29,16 @@
...
                 <artifactId>junit-jupiter-engine</artifactId>
-                <version>${version.junit.jupiter}</version>
                 <scope>test</scope>
             </dependency>
             <dependency>
                 <groupId>org.junit.vintage</groupId>
                 <artifactId>junit-vintage-engine</artifactId>
-                <version>${version.junit.vintage}</version>
                 <scope>test</scope>
             </dependency>
```
jwomeara: You should remove these completely.  These dependencies are managed by the parent pom.  The child inherits the parent's dependency management, so if you had kept the versions in this would have been redundant.  Without the versions, I'm not sure what this accomplishes at all.  I think this is all that's in the dependency management section of this pom, so you can just remove dependency management completely.

core/pom.xml
```diff
@@ -59,16 +51,19 @@
...
             <artifactId>junit-jupiter-engine</artifactId>
+            <version>${version.junit.bom}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.junit.vintage</groupId>
             <artifactId>junit-vintage-engine</artifactId>
+            <version>${version.junit.bom}</version>
             <scope>test</scope>
         </dependency>
```
jwomeara: No versions needed here since these dependencies are managed by the parent.

jwomeara (changes requested):

pom.xml
jwomeara: Plugin dependencies can't be managed, unfortunately.  https://issues.apache.org/jira/browse/MNG-2496

warehouse/pom.xml
jwomeara: Plugin dependency management isn't a thing, unfortunately.  If you want to avoid having to redefine this plugin in multiple poms you could just move the maven-surefire-plugin configuration from the warehouse and web-services poms to the pluginManagement entry for the datawave parent pom.  That way, all of the child modules will inherit the same surefire configuration, and you will only have to define it in one place.

Honestly, I'm not sure why we aren't already doing that.  The surefire plugin configuration for warehouse and web-services is identical.

warehouse/accumulo-extensions/pom.xml
jwomeara: he already updated it to remove the version.  it wasn't needed.

i would actually advise against explicitly listing the test scope when managing dependencies.  if you specify a scope, it will only manage it for that scope, and OTHER imported dependency management sections could override your compile time version.

ivakegg (dismissed):

warehouse/pom.xml
```diff
@@ -444,6 +444,19 @@
...
             <artifactId>datawave-common-test</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
+        </dependency>
```
jwomeara: Are these needed here?  and if so, shouldn't they be test scoped?

warehouse/query-core/src/test/java/datawave/audit/SplitSelectorExtractorTest.java
```diff
@@ -1,23 +1,23 @@
...
 
 import org.apache.commons.lang.math.IntRange;
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
 
 import com.google.common.collect.Lists;
 
 import datawave.webservice.query.QueryImpl;
 
-public class SplitSelectorExtractorTest {
```
jwomeara, line 17: curious why public was removed throughout.

web-services/client/pom.xml
```diff
@@ -171,11 +171,6 @@
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-api</artifactId>
-            <scope>test</scope>
-        </dependency>
```
jwomeara: why are we removing this?  if we're not going to add junit-jupiter-api to each module, why remove it from the ones that already include it?

jwomeara (changes requested):

warehouse/pom.xml
apmoriarty: Yes these are needed and yes they should be test scoped.

web-services/client/pom.xml
apmoriarty: I removed it because it was unused, whereas with the other modules there are tests that rely on the jupiter api. I can add back the unused dependency if you like.

warehouse/query-core/src/test/java/datawave/audit/SplitSelectorExtractorTest.java
apmoriarty: SonarLint will flag public modifiers in JUnit5 classes and test methods. Using the default package visible is recommended to improve readability.

warehouse/pom.xml
jwomeara: the warehouse module itself has no source code, which means it has no tests to compile, let alone run, so why do you think these are needed?  for what it's worth I removed all of the dependencies in this pom and everything built just fine...

common-test/pom.xml
```diff
@@ -63,5 +63,10 @@
             <scope>provided</scope>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
```
jwomeara: junit-jupiter-engine

warehouse/common/pom.xml
```diff
@@ -55,5 +55,10 @@
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
```
jwomeara: junit-jupiter-engine

warehouse/pom.xml
```diff
@@ -444,6 +444,19 @@
...
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
```
jwomeara: junit-jupiter-engine

web-services/client/pom.xml
jwomeara: i guess what im saying is that this change wasn't required for the work you were doing.  if you're worried about scope creep, well i think this counts as that, right?  only change what you need

jwomeara (changes requested):

warehouse/pom.xml
jwomeara: happy to approve once this is removed.

jwomeara (changes requested):
just need some stuff removed from the warehouse pom.  also, not sure why `public` was removed from all of the test methods

warehouse/pom.xml
```diff
@@ -445,6 +445,11 @@
             <groupId>gov.nsa.datawave</groupId>
             <artifactId>datawave-common-test</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
```
jwomeara, line 452: You don't need this.

warehouse/ssdeep-common/pom.xml
```diff
@@ -24,5 +24,10 @@
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
```
jwomeara, line 30: does ssdeep have junit jupiter tests?

web-services/pom.xml
```diff
@@ -506,6 +506,11 @@
             <groupId>gov.nsa.datawave</groupId>
             <artifactId>datawave-common-test</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
```
jwomeara, line 513: does web-services have junit jupiter tests?

jwomeara (changes requested):

warehouse/pom.xml
apmoriarty: If you look at the recent commit history you can see how this was removed, and then added back in when other modules failed to build.

web-services/pom.xml
jwomeara: i was wrong.  let's keep this

warehouse/ssdeep-common/pom.xml
jwomeara: i think this can go away

warehouse/pom.xml
jwomeara: i was wrong.  if we keep this, all child modules will inherig junit-jupiter-engine, which is what we want.  sorry for the confusion

apmoriarty: all good!

warehouse/ssdeep-common/pom.xml
jwomeara: wrong again

jwomeara (approved):
ship it

ivakegg (approved):

drewfarris:
Would it be possible to squash commits like this in the future? The 17 commits made rebasing some feature branches really challenging when there were conflicts.

jwomeara:
I agree.  PR commits should be squashed when merging.

apmoriarty:
100% agree, when merging this one I forgot to change the default merge strategy. Perhaps that should be updated?

jwomeara:
You can't set a default merge strategy for a project. GitHub defaults to the last merge strategy you selected on a per user basis. So you need to manually select squash and merge the next time you merge a PR to reset your default.

PR #2287, opened 2024-02-28
Adding before/after parameters to hit callout for excerpt fields function

mineralntl:


warehouse/query-core/src/main/java/datawave/query/attributes/ExcerptFields.java
```diff
@@ -59,7 +59,8 @@ public static ExcerptFields from(String string) {
         String[] fieldParts = string.split(Constants.COMMA);
         for (String fieldPart : fieldParts) {
             String[] parts = StringUtils.split(fieldPart, Constants.FORWARD_SLASH);
-            excerptFields.put(parts[0], Integer.valueOf(parts[1]));
+            String direction = parts.length == 3 ? parts[2] : "both";
```
avgAGB: Minor quip, but it might be worth considering putting the "both" string as a constant for the two spots in this class where its used. Here and on line 140

warehouse/query-core/src/main/java/datawave/query/attributes/ExcerptFields.java
```diff
@@ -112,19 +113,47 @@ public boolean containsField(String field) {
...
+     * Put the offset to use when retrieving excerpts for the specified field. When no direction is specified, we default to returning both directions.
      *
      * @param field
      *            the field
      * @param offset
      *            the offset
      */
     public void put(String field, Integer offset) {
-        fieldMap.put(field, offset);
+        this.put(field, offset, "both");
```
avgAGB: Other spot for a potential constant here.

cogross (dismissed):

warehouse/query-core/src/main/java/datawave/query/attributes/ExcerptFields.java
ivakegg: I agree, a constant for "both" would be goodness

warehouse/query-core/src/main/java/datawave/query/attributes/ExcerptFields.java
mineralntl: Good catch, will update

warehouse/query-core/src/main/java/datawave/query/attributes/ExcerptFields.java
mineralntl: Aye aye

avgAGB (dismissed):

hlgp (dismissed):
Nice tests!

mineralntl:
Re-submitting after fixing rebase compilation errors

warehouse/query-core/src/test/java/datawave/query/ExcerptTest.java
```diff
@@ -225,6 +225,68 @@ public void simpleTest() throws Exception {
...
+    public void simpleTestBefore() throws Exception {
+        Map<String,String> extraParameters = new HashMap<>();
+        extraParameters.put("include.grouping.context", "true");
+        extraParameters.put("hit.list", "true");
+        extraParameters.put("return.fields", "HIT_EXCERPT");
+        extraParameters.put("query.syntax", "LUCENE");
+
+        String queryString = "QUOTE:(farther) #EXCERPT_FIELDS(QUOTE/2/before)";
+
+        // not sure why the timestamp and delete flag are present
```
hlgp, line 238: any idea why the ts and delete flag are showing up in the EXCERPT?

mineralntl: Haven't looked into that any further - I can try tracing it down if we think it's necessary before merging, or we can break that into a different ticket to get the functionality added in faster if not.

alerman (approved):

hlgp (approved):

warehouse/query-core/src/test/java/datawave/query/ExcerptTest.java
mineralntl: Upon further review - HIT_EXCERPT is added to the document as an Attribute which includes ts and delete flags whereas most (but not all) of the other fields are added as TypeAttributes which do not include ts or delete flags. This has been the case with the ExcerptTransform class since April 2022.

hlgp: Nice hunting

PR #2295, opened 2024-03-06
Upper-end limit for maximum terms printed for PrintingVisitor

avgAGB:


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PrintingVisitor.java
```diff
@@ -337,552 +346,978 @@ private void childrenAccept(SimpleNode node, ParserVisitor visitor, Object data)
...
     public Object visit(ASTJexlScript node, Object data) {
-        output.writeLine(data + node.toString());
-        childrenAccept(node, this, data + PREFIX);
+        if (maxTermsToPrint > 0) {
+            output.writeLine(data + node.toString());
+            maxTermsToPrint -= 1;
+            childrenAccept(node, this, data + PREFIX);
+        } else {
+            LOGGER.trace("maxTermsToPrint is exhausted. Ignoring: " + node);
+        }
```
apmoriarty: It looks like this code will still emit a log message for each node in the query that exceeds the threshold. Given that, I think I would prefer to see something like this (explanations for changes in-line)
```
int termsPrinted = 0;

...

public Object visit(node, data) {
    if (termsPrinted > maxTermsToPrint) {
        return null;  // return early without printing for each additional node visited after the threshold
    }

    termsPrinted++;

    if (termsPrinted > maxTermsToPrint) {
        LOGGER.trace("exceeded max terms to print threshold of " + maxTermsToPrint);  // just log the threshold, logging the whole node for a very large query could be less than constructive
    }

    output.writeLine(data + node.toString());
    childrenAccept(node, this, data + PREFIX);
    return null;
}
```

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/EventDataQueryExpressionVisitorTest.java
```diff
@@ -905,6 +909,54 @@ public static void printJexlScript(ASTJexlScript script) {
...
+    @Test
+    public void testPrintVisitorDefault() throws Exception {
+        // Between 5 and 12.
+        String originalQuery = "FOO == 'abc' AND ((_Bounded_ = true) && (BAZ >= '+aE5' AND BAZ <= '+bE1.2'))";
+        // @TODO, use ExpandMultiNormalizedTerms to normalize this query?
+
+        ASTJexlScript script = JexlASTHelper.parseJexlQuery(originalQuery);
+        final Map<String,ExpressionFilter> filter = EventDataQueryExpressionVisitor.getExpressionFilters(script, attrFactory);
+
+        final PrintStream stdOut = System.out;
```
matthpeterson: It looks like this line can be removed.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/EventDataQueryExpressionVisitorTest.java
```diff
@@ -905,6 +909,54 @@ public static void printJexlScript(ASTJexlScript script) {
...
+    @Test
+    public void testPrintVisitorDefault2() throws Exception {
+        // Between 5 and 12.
+        String originalQuery = "FOO == 'abc' AND ((_Bounded_ = true) && (BAZ >= '+aE5' AND BAZ <= '+bE1.2'))";
+        // @TODO, use ExpandMultiNormalizedTerms to normalize this query?
+
+        ASTJexlScript script = JexlASTHelper.parseJexlQuery(originalQuery);
+        final Map<String,ExpressionFilter> filter = EventDataQueryExpressionVisitor.getExpressionFilters(script, attrFactory);
+
+        final PrintStream stdOut = System.out;
```
matthpeterson: also removable, I think

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PrintingVisitor.java
```diff
@@ -101,15 +101,20 @@
 import org.apache.commons.jexl3.parser.SimpleNode;
 import org.apache.commons.jexl3.parser.StringProvider;
 import org.apache.commons.jexl3.parser.TokenMgrException;
+import org.apache.log4j.Logger;
 
 import com.google.common.collect.Lists;
 
+import datawave.data.hash.AbstractUIDBuilder;
```
matthpeterson: This import can be removed.  So can the JexlFeatures import which was already there.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/EventDataQueryExpressionVisitorTest.java
matthpeterson: In my efforts to better understand how this works, I wrote a few unit tests.  I've created an MR to add them to this branch, if you'd like them: https://github.com/NationalSecurityAgency/datawave/pull/2357

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PrintingVisitor.java
avgAGB: Thanks, will clean this up.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/EventDataQueryExpressionVisitorTest.java
avgAGB: Will do

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PrintingVisitor.java
```diff
@@ -337,550 +345,1316 @@ private void childrenAccept(SimpleNode node, ParserVisitor visitor, Object data)
...
 
     @Override
     protected Object visit(ASTQualifiedIdentifier node, Object data) {
+        if (termsPrinted > maxTermsToPrint) {
+            return null;
+        }
+        termsPrinted++;
+
+        if (termsPrinted > maxTermsToPrint) {
+            LOGGER.trace("exceeded max terms to print threshold of " + maxTermsToPrint);
```
matthpeterson: I think this trace will log just before the nth term, so maybe the word "exceeded" should be replaced with "reached"

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/EventDataQueryExpressionVisitorTest.java
```diff
@@ -37,6 +40,7 @@
 import datawave.query.util.MockMetadataHelper;
 import datawave.query.util.TypeMetadata;
 import datawave.test.JexlNodeAssert;
+import javassist.bytecode.ByteArray;
```
matthpeterson: Oops I think these new imports should be removed too.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PrintingVisitor.java
avgAGB: Makes sense, will fix.

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PrintingVisitor.java
```diff
@@ -337,550 +345,1316 @@ private void childrenAccept(SimpleNode node, ParserVisitor visitor, Object data)
...
     public Object visit(ASTJexlScript node, Object data) {
+        if (termsPrinted >= maxTermsToPrint) {
+            return null; // return early without printing for each additional node visited after the threshold
+        }
+        termsPrinted++;
+
+        if (termsPrinted >= maxTermsToPrint) {
+            LOGGER.trace("reached max terms for print threshold of " + maxTermsToPrint); // just log the threshold, logging the whole node for a very large
+                                                                                         // query could be less than constructive
+        }
```
keith-ratcliffe: Wouldn't it be better to consolidate all the duplicate instances of this logic with a single method call?

keith-ratcliffe (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PrintingVisitor.java
avgAGB: Ok, I updated the code to pull out the duplicates into a helper.

apmoriarty (approved):

matthpeterson (approved):

keith-ratcliffe (approved):

PR #2301, opened 2024-03-09
Remove parse from JexlStringBuildingVisitor, deprecate old methods

FineAndDandy:
Deprecated buildQueryWithoutParse() methods and removed parse from buildQuery(). We can remove buildQueryWithoutParse() in a future release.


FineAndDandy:
similar to #2300 but for integration rather than a patch

MiguelRicardos:
It builds fine locally. You need to rebase to work off the newer Maven image to get the GitHub action to pass.

PR #2302, opened 2024-03-11
Fix GlobalIndexUidAggregator complaining

mineralntl:
This seeks to fix an error seen from the GlobalIndexUidAggregator complaining about an inccorect Value type being passed from the ShardTableQueryMetricHandler. Instead of passing an incompatible type and clogging up the logs, we now move UIDs to the removalUidList instead of flipping the delete flag on the entire key being processed.


hlgp:
Deferred

PR #2304, opened 2024-03-12
Feature improve excerpt iterator

austin007008:


warehouse/query-core/src/main/java/datawave/query/iterator/logic/WordsAndScores.java
```diff
@@ -0,0 +1,67 @@
+package datawave.query.iterator.logic;
+
+import java.util.ArrayList;
+import java.util.Set;
+
+public class WordsAndScores {
```
ivakegg, line 27: A little documentation in this class might be nice.

austin007008: I working on adding documentation in now 👍

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -314,29 +314,62 @@ private Set<Excerpt> getExcerpts(PhraseIndexes phraseIndexes) {
...
-        try {
-            excerptIterator.init(source, excerptIteratorOptions, env);
-            excerptIterator.seek(range, Collections.emptyList(), false);
-            if (excerptIterator.hasTop()) {
-                Key topKey = excerptIterator.getTopKey();
-                String[] parts = topKey.getColumnQualifier().toString().split(Constants.NULL);
-                // The column qualifier is expected to be field\0phrase.
-                if (parts.length == 2) {
-                    return getHitPhrase(hitTermValues, parts);
+        for (int i = 0; i <= timesToTry; i++) {
```
ivakegg: I really do not like this multiple pass thing.  The number of terms in a document can be HUGE.  So making this pass multiple times will cause us significant performance issues.  How about filling the initial array of terms to be significantly outside the bounds of the initial request, and then trim down the excerpt to what we need.  This would be significantly less expensive.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/iterator/logic/WordsAndScores.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Set;
+
+/**
+ * An object used to save terms and their respective scores (if a score exists and is valid).
+ */
+public class WordsAndScores {
+    /** the arraylist of words */
+    private final ArrayList<String> words;
+    /** the arraylist of scores */
+    private final ArrayList<Integer> scores;
```
jack-gitdev, line 39: One possible approach in this class is to make a WordScorePair supporting class. This would allow you to have one array list here so you wouldn't have to coordinate between two arrays. Would also allow you to provide default values upon construction.

austin007008: good idea. might implement

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
austin007008: I've done a hybrid of the ideas.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
```diff
@@ -315,28 +386,225 @@ public void next() throws IOException {
...
+     * Trims off the front to get the correct size
+     *
+     * @param termsToOutput
+     *            the terms to create a phrase from
+     * @param beforeIndex
+     *            the index in the array to start at
+     * @param start
+     *            the index in the array where we want to start setting values to null at
+     */
+    private void trimBeginning(String[] termsToOutput, int beforeIndex, float start) {
```
keith-ratcliffe: Should `start` be `int` rather than `float`?

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
```diff
@@ -315,28 +386,225 @@ public void next() throws IOException {
...
+     * Trims off the end to get the correct size
+     *
+     * @param termsToOutput
+     *            the terms to create a phrase from
+     * @param afterIndex
+     *            the index in the array to start at
+     * @param start
+     *            the index in the array where we want to start setting values to null at
+     */
+    private void trimEnd(String[] termsToOutput, int afterIndex, float start) {
```
keith-ratcliffe: Should `start` be `int` rather than `float`?

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
austin007008: possibly. I don't remember why I made it a float in the first place. Just have to cast the computation for "start" to an int to get everything to compile still.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
austin007008: possibly. I don't remember why I made it a float in the first place. Just have to cast the computation for "start" to an int to get everything to compile still.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
```diff
@@ -66,6 +65,19 @@ public class TermFrequencyExcerptIterator implements SortedKeyValueIterator<Key,
     // the top value
     protected Value tv;
 
+    private float origHalfSize;
```
drewfarris: Comment?

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
```diff
@@ -66,6 +65,19 @@ public class TermFrequencyExcerptIterator implements SortedKeyValueIterator<Key,
...
+    // the list of hit terms
+    protected ArrayList<String> hitTermsList;
+    // the direction for the excerpt
+    private String direction;
+
+    private boolean trim;
```
drewfarris: Comment?

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
```diff
@@ -278,6 +300,16 @@ public void next() throws IOException {
...
+                    // check if the number of scores is equal to the number of offsets
+                    boolean useScores = info.getScoreCount() == info.getTermOffsetCount();
+                    boolean hasOnlyNegativeScores = true;
+                    List<Integer> scoreList = info.getScoreList();
+                    for (Integer integer : scoreList) {
+                        if (integer >= 0) {
+                            hasOnlyNegativeScores = false;
+                            break;
+                        }
+                    }
```
apmoriarty: In future work I would extract this to a helper method

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
```diff
@@ -315,28 +386,244 @@ public void next() throws IOException {
...
+    /**
+     * Looks for hit phrases (not separate hit terms) and puts the whole phrase in brackets
+     *
+     * @param terms
+     *            the terms to create a phrase from
+     */
+    private void checkForHitPhrase(WordsAndScores[] terms) {
+        ArrayList<String> hitPhrases = new ArrayList<>();
+        // checks for phrases (anything in the hit list with a space in it) and adds them to a new arrayList
+        for (String s : hitTermsList) {
```
apmoriarty, line 737: There's nothing preventing a value from having a space in it. This code will work but might discover more phrases than actually exist. Longer term we should have a separate attribute type for hit term phrases.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
```diff
@@ -315,28 +386,244 @@ public void next() throws IOException {
...
 
-        return joiner.join(largestTerms);
+    private void overrideOutputLongest(WordsAndScores[] terms) {
+        for (WordsAndScores ws : terms) {
+            if (ws == null) {
+                continue;
+            }
+            if (ws.getUseScores()) {
+                return;
+            }
```
apmoriarty: I would collapse these into a single check, but that's just personal preference.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
```diff
@@ -66,6 +65,19 @@ public class TermFrequencyExcerptIterator implements SortedKeyValueIterator<Key,
...
+    // the list of hit terms
+    protected ArrayList<String> hitTermsList;
+    // the direction for the excerpt
+    private String direction;
+
+    private boolean trim;
+
+    private static final String BEFORE = "BEFORE";
+    private static final String AFTER = "AFTER";
+    private static final String BOTH = "BOTH";
```
apmoriarty: This could be an enum

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -319,51 +336,64 @@ private Set<Excerpt> getExcerpts(PhraseIndexes phraseIndexes) {
...
-            excerptIterator.init(source, excerptIteratorOptions, env);
-            excerptIterator.seek(range, Collections.emptyList(), false);
-            if (excerptIterator.hasTop()) {
-                Key topKey = excerptIterator.getTopKey();
-                String[] parts = topKey.getColumnQualifier().toString().split(Constants.NULL);
-                // The column qualifier is expected to be field\0phrase.
-                if (parts.length == 2) {
-                    return getHitPhrase(hitTermValues, parts);
+        int i = 0;
+        while (i <= 1) {
```
apmoriarty: If I'm reading this correctly it's possible to run the offset iterator twice in a row. Is that what's intended here?

Typically a while loop is only used when configuring scan column families or ranges in a batch scanner. It's a bit odd to see the whole iterator execution inside a while loop.

warehouse/query-core/src/test/java/datawave/query/ExcerptTest.java
```diff
@@ -220,7 +220,7 @@ public void simpleTest() throws Exception {
         String queryString = "QUOTE:(farther) #EXCERPT_FIELDS(QUOTE/2)";
 
         // not sure why the timestamp and delete flag are present
-        Set<String> goodResults = Sets.newHashSet("HIT_EXCERPT:get much [farther] with a: : [] 9223372036854775807 false");
+        Set<String> goodResults = new HashSet<>(Collections.singleton("HIT_EXCERPT:get much [farther] with a: : [] 9223372036854775807 false"));
```
apmoriarty: Consider using `Set.of("string")` here and elsewhere.

warehouse/query-core/src/test/java/datawave/query/iterator/logic/TermFrequencyExcerptIteratorTest.java
```diff
@@ -208,7 +218,259 @@ private void givenOptions(String field, int start, int end) {
...
+
+        assertTrue(iterator.hasTop());
+
+        Key topKey = iterator.getTopKey();
+        assertEquals(row, topKey.getRow());
+        assertEquals(new Text("email" + Constants.NULL + "333.222.111"), topKey.getColumnFamily());
+        assertEquals(new Text("BODY" + Constants.NULL + "XXXNOTSCOREDXXX" + Constants.NULL
+                        + "we [like to repeat] [stuff] do not [ask questions] we [like to repeat] [stuff]" + Constants.NULL + "XXXNOTSCOREDXXX"),
+                        topKey.getColumnQualifier());
+    }
```
apmoriarty, line 518: great additions

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
austin007008: I agree. Right now this code will never actually be triggered when running because I learned that datawave does not pass down the whole phrases into the hit excerpt code currently. plans for part 2 of this include finding a way to pass that down and differentiate the original phrases from the broken up pieces.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
austin007008: yep. If we hit a certain condition while the iterator is running the first time, we exit out and run the iterator with a greater range the second time. Would it be more proper to put it in a for loop instead of a while loop?

warehouse/query-core/src/test/java/datawave/query/ExcerptTest.java
austin007008: done

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
```diff
@@ -278,6 +300,16 @@ public void next() throws IOException {
                 try {
                     // parse the offsets from the value
                     TermWeight.Info info = TermWeight.Info.parseFrom(source.getTopValue().get());
+                    // check if the number of scores is equal to the number of offsets
+                    boolean useScores = info.getScoreCount() == info.getTermOffsetCount();
+                    boolean hasOnlyNegativeScores = true;
+                    List<Integer> scoreList = info.getScoreList();
+                    for (Integer integer : scoreList) {
```
drewfarris: it appears that you only need to check the scores if `useScores` is true. Consider setting `useScores` to false here if you find out that all scores are negative. This simplifies the check below. 

Also, don't we only want to look at the scores for the terms between our start and end offset? I suppose there could be positive scores only outside of the range of offsets we're interested in, and if we only consider the scores for the terms between our start and end offset there's a possibility that they will all be negative.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
drewfarris:
```suggestion
    private enum Direction {
        BEFORE,
        AFTER,
        BOTH
    };
```
There will be other things to fix if you take this suggestion.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
```diff
@@ -264,7 +280,13 @@ public void next() throws IOException {
         Text cv = top.getColumnVisibility();
         long ts = top.getTimestamp();
         Text row = top.getRow();
-        List<String>[] terms = new List[endOffset - startOffset];
+        // set the size of the array to the amount of terms we need to choose
+        WordsAndScores[] terms = new WordsAndScores[endOffset - startOffset];
```
drewfarris: Consider renaming this `wordsAndScores` instead of `terms`. It will be easier to read when it's used later.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
```diff
@@ -302,9 +344,38 @@ public void next() throws IOException {
...
+        // check to see if we have something scored and if so, turn off outputting the scores
+        for (WordsAndScores term : terms) {
+            if (term == null) {
+                continue;
+            }
+            if (term.getUseScores()) {
+                usedScores = true;
+                term.setOutputScores(false);
+            }
+        }
```
drewfarris: Could we avoid calling `generatePhase` twice if we check this first or even check the boolean value of `useScores` here?

drewfarris (commented):
Didn't complete a review yet, but left some things to think about.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
austin007008: done

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
```diff
@@ -135,6 +147,10 @@ public void init(SortedKeyValueIterator<Key,Value> source, Map<String,String> op
         this.startOffset = Integer.parseInt(options.get(START_OFFSET));
         this.endOffset = Integer.parseInt(options.get(END_OFFSET));
         this.fieldName = options.get(FIELD_NAME);
+        hitTermsList = new ArrayList<>();
+        direction = BOTH;
+        origHalfSize = 0;
+        trim = false;
```
drewfarris: There's a `deepCopy` method above the `init` method, I believe you'll need to add these fields to that method as well or the deep copies created by that method will be incomplete.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/WordsAndScores.java
```diff
@@ -0,0 +1,384 @@
...
+
+    /**
+     * A method to add a word and score into the object and saves its index if it is the longest word or has the smallest (non-negative) score.
+     *
+     * @param word
+     *            the word to add
+     * @param score
+     *            the score to add
+     */
+    public boolean addTerm(String word, int score, List<String> hitTermsList) {
```
drewfarris: The `addTerm` method also returns a `boolean` value, but it's not really documented anywhere, nor is the fact that a specific WordAndScore instance is a stop word is recorded.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
austin007008: done

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
drewfarris: Looking at this further, this is a little confusing and could use some more explanation - I see we have an array of `WordsAndScores` here, but each instance of `WordsAndScores` has a list of words and scores. Could you somehow explain what each instance of this array captures?

austin007008: renamed to "wordsAndScoresArr"

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -319,51 +337,70 @@ private Set<Excerpt> getExcerpts(PhraseIndexes phraseIndexes) {
      * @return the excerpt
      */
     private String getExcerpt(String field, int start, int end, Range range, ArrayList<String> hitTermValues) {
+        if (start < 0) {
+            start = 0;
+        }
+        float origHalfSize = (float) (end - start) / 2;
+        int expandSize = 20;
         excerptIteratorOptions.put(TermFrequencyExcerptIterator.FIELD_NAME, field);
```
drewfarris: Consider making `excerptIteratorOptions` a local variable, I don't think there's a good reason to make it a class field.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -319,51 +337,70 @@ private Set<Excerpt> getExcerpts(PhraseIndexes phraseIndexes) {
...
-        try {
-            excerptIterator.init(source, excerptIteratorOptions, env);
-            excerptIterator.seek(range, Collections.emptyList(), false);
-            if (excerptIterator.hasTop()) {
-                Key topKey = excerptIterator.getTopKey();
-                String[] parts = topKey.getColumnQualifier().toString().split(Constants.NULL);
-                // The column qualifier is expected to be field\0phrase.
-                if (parts.length == 2) {
-                    return getHitPhrase(hitTermValues, parts);
+        int i = 0;
```
drewfarris: Instead of `i` perhaps you could call this variable `attempt` and add a comment like "we'll try to run the section twice. If the first attempt fails, we'll run it a second time with a greater allowed range of offsets..

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -319,51 +337,70 @@ private Set<Excerpt> getExcerpts(PhraseIndexes phraseIndexes) {
...
+        while (i <= 1) {
+            // set the start of the range
+            if (i == 0) {
```
drewfarris: Instead of having two separate conditional statements that check which attempt we're in, consider comining it into a single statement. Something like:

```
            if (attempt == 0) {
                // on the first attempt just use the start and end offsets provided.
                excerptIteratorOptions.put(TermFrequencyExcerptIterator.START_OFFSET, String.valueOf(start));
                excerptIteratorOptions.put(TermFrequencyExcerptIterator.END_OFFSET, String.valueOf(end));
            } else {
                // on the second attempt, set the start and end the range to include up to expandSize terms prior to
                // the start offset and after the end offset.
                if (start - expandSize > 0) {
                    excerptIteratorOptions.put(TermFrequencyExcerptIterator.START_OFFSET, String.valueOf(start - expandSize));
                } else {
                    excerptIteratorOptions.put(TermFrequencyExcerptIterator.START_OFFSET, String.valueOf(0));
                }
                excerptIteratorOptions.put(TermFrequencyExcerptIterator.END_OFFSET, String.valueOf(end + expandSize));
            }
```

It's a little more readable and isolates most of the the attempt specific behavior to a single section of code.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -319,51 +337,70 @@ private Set<Excerpt> getExcerpts(PhraseIndexes phraseIndexes) {
...
-                hitPhrase.add("[" + phrasePart + "]");
+            if (i == 0) {
+                excerptIteratorOptions.put(TermFrequencyExcerptIterator.END_OFFSET, String.valueOf(end));
             } else {
-                hitPhrase.add(phrasePart);
+                // set the end of the range
+                excerptIteratorOptions.put(TermFrequencyExcerptIterator.END_OFFSET, String.valueOf(end + expandSize));
             }
-        }
+            if (log.isDebugEnabled() && i != 0) {
```
drewfarris: Whenever we check which attempt we're on, try to be consistent - instead of using `i != 0` in some cases and `i == 1` in others, let's just use `i == 1` for example. 

Specifically talking about the checks in this line (364), and lines 375 and 391

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -319,51 +337,70 @@ private Set<Excerpt> getExcerpts(PhraseIndexes phraseIndexes) {
...
+                                excerptIteratorOptions.get(TermFrequencyExcerptIterator.START_OFFSET),
+                                excerptIteratorOptions.get(TermFrequencyExcerptIterator.END_OFFSET));
+            }
+            try {
+                excerptIterator.init(source, excerptIteratorOptions, env);
+                excerptIterator.setHitTermsList(hitTermValues);
+                excerptIterator.setDirection(excerptFields.getDirection(field).toUpperCase().trim());
+                excerptIterator.setOrigHalfSize(origHalfSize);
+                if (i != 0) {
+                    excerptIterator.setTrim(true);
```
drewfarris: Could you add a comment as to what `trim` means?

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -319,51 +337,70 @@ private Set<Excerpt> getExcerpts(PhraseIndexes phraseIndexes) {
...
 
-        // return phrase based on direction
-        String result = String.join(" ", hitPhrase); // if no direction given, return everything
-        String direction = excerptFields.getDirection(phraseParts[0]).toUpperCase().trim();
-        if (direction.equals(BEFORE)) { // remove tokens prior to hit term
-            result = StringUtils.substringBeforeLast(result, "]") + "]";
-        } else if (direction.equals(AFTER)) { // remove tokens after hit term
-            result = "[" + StringUtils.substringAfter(result, "[");
+                    // if we have reached the limit of times to try, or we have no stop words removed
+                    if (i == 1 || !parts[1].equals("XXXWESKIPPEDAWORDXXX")) {
```
drewfarris: Will we ever have a stop word removed on the first attempt (attempt 0)? If not, the '||' (or) is a little confusing here.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -319,51 +337,70 @@ private Set<Excerpt> getExcerpts(PhraseIndexes phraseIndexes) {
...
+                        return parts[1] + Constants.NULL + parts[2] + Constants.NULL + parts[3];
+                    }
+                } else {
+                    return null;
```
drewfarris: Would we possibly want to still make the second attempt in this case if we're only on the first? Consider calling `continue` or `break` instead?

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -319,51 +337,70 @@ private Set<Excerpt> getExcerpts(PhraseIndexes phraseIndexes) {
...
+                excerptIterator.seek(range, Collections.emptyList(), false);
+                // if an excerpt is returned...
+                if (excerptIterator.hasTop()) {
+                    Key topKey = excerptIterator.getTopKey();
+                    String[] parts = topKey.getColumnQualifier().toString().split(Constants.NULL);
+                    // The column qualifier is expected to be field\0phraseWithScores\0phraseWithoutScores\0oneBestExcerpt.
+                    if (parts.length != 4) {
+                        log.warn("{} returned top key with incorrectly-formatted column qualifier in key: {} when scanning for excerpt [{},{}] for field {} within range {}",
+                                        TermFrequencyExcerptIterator.class.getSimpleName(), topKey, start, end, field, range);
+                        return null;
```
drewfarris: Would we potentially want to make the second attempt in this case if we're only on the first? Consider calling break or continue instead?

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -319,51 +337,70 @@ private Set<Excerpt> getExcerpts(PhraseIndexes phraseIndexes) {
...
+                }
+            } catch (IOException e) {
+                throw new RuntimeException("Failed to scan for excerpt [" + start + "," + end + "] for field " + field + " within range " + range, e);
+            }
+            i++;
         }
-
-        return result;
+        // it should always return from inside the loop so if this is reached something went very wrong
+        throw new RuntimeException("This should never be reached. Something went wrong!");
```
drewfarris: Perhaps we return null here after we've exhausted all of our attempts instead of returning null from within the loop.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
drewfarris: A for loop might make it a little clearer that you expect only two attempts _and_ handle the incrementing of the variable.

```
for (int i=0; i < 2; i++) { //make two attempts to do X
```

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
austin007008: both done

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -319,51 +337,70 @@ private Set<Excerpt> getExcerpts(PhraseIndexes phraseIndexes) {
...
+                excerptIterator.init(source, excerptIteratorOptions, env);
+                excerptIterator.setHitTermsList(hitTermValues);
+                excerptIterator.setDirection(excerptFields.getDirection(field).toUpperCase().trim());
+                excerptIterator.setOrigHalfSize(origHalfSize);
+                if (i != 0) {
+                    excerptIterator.setTrim(true);
+                }
+                excerptIterator.seek(range, Collections.emptyList(), false);
+                // if an excerpt is returned...
+                if (excerptIterator.hasTop()) {
```
drewfarris, line 353: If I'm reading this section correctly, it looks like we'll never return an excerpt if the attempt is == 0 - we alwasy seem to return null in that case. Is this correct behavior? Aren't there some cases where we'll want to return what is produced by the excerptIterator during the first attempt?

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
austin007008: done

drewfarris (commented):
Another set of comments/questions/feedback

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
drewfarris: I noticed that there's another class `ExcerptFields` that uses this notion of direction as well - and that uses a String too. It might make sense for these two classes to use the same enum, but that would be a larger refactoring.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
austin007008: done

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
austin007008: done

warehouse/query-core/src/main/java/datawave/query/iterator/logic/WordsAndScores.java
austin007008: added @returns for the methods. 

checking for the stop words here is mainly for the "fast fail" on the first time the iterator is called. I hesitate to save whether or not each term is a stop word as we would need to save a list of booleans for that. The one extra check we do before returning a selected term is pretty low cost so I'm not sure if saving that information for all terms in a WordsAndScores is worth it.

thoughts/opinions?

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
austin007008: done

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
austin007008: good idea. done

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
austin007008: done

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
austin007008: done

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
austin007008: yep. done

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
austin007008: my bad, let me try to explain a bit: 
If we find a stop word the first time, we will always find a stop word the second time also. The "i == 1" is so that on the second attempt, we always return what we have no matter what. The "!parts[1].equals("XXXWESKIPPEDAWORDXXX"" is so that we can return on the first attempt if there were no skip words found (basically if the iterator did not fast-fail and it returned an excerpt to us).

This part of the code is related to TermFrequencyExcerptIterator around line 340ish where if a stop word is found and this is the first iteration (know by "trim" being false), we immediately return with "XXXWESKIPPEDAWORDXXX + Constants.NULL + XXXWESKIPPEDAWORDXXX + Constants.NULL + XXXWESKIPPEDAWORDXXX" in the key. 
The "XXXWEKIPPEDAWORDXXX" is put in place of all of the excerpts to be detected here in this if statement.

austin007008: actually now that I think about it the "i ==1" might not be needed because XXXWESKIPPEDAWORDXXX can't be seen after the first attempt. I think this was another piece that was just still in here from a different implementation I tried before this one.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
austin007008: I guess it wouldn't hurt to try. But I think if hasTop is still false after the second attempt we might want to throw/log an error as something probably went terribly wrong if this happens.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
austin007008: we can attempt and then if it still isn't working, then log/output an error

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
austin007008: as with the other two comments, no more returning null as having a blank excerpt with no message tells us nothing. 

Will output the message "something went wrong generating your excerpt"

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
austin007008: it will return an excerpt during attempt==0 if there are no stop words found when creating wordsAndScoresArr (the list of all of the terms we can choose from at all of the offsets in range)

austin007008:
still need to add comments on some new pieces

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
```diff
@@ -34,38 +34,59 @@
...
+     * Whether we might need to trim down the excerpt to the requested size. <br>
+     * <br>
+     * Is false if this is the first time running the iterator and true otherwise.
+     */
+    private boolean trim;
+
+    private static final String BEFORE = "BEFORE";
+    private static final String AFTER = "AFTER";
+    private static final String BOTH = "BOTH";
+    private static final String XXXWESKIPPEDAWORDXXX = "XXXWESKIPPEDAWORDXXX";
```
drewfarris: Suggesting a slight re-ordering of these fields so that related fields are close to each other in the code (this violates conventions with the goal of making the code more readable). Also an attempt to further define what these things mean. Please check that my descriptions are accurate. Also favoring a more compact comment style. Sprinkled in some `<br>`'s to make the auto-formatter happy.
```suggestion
    /** the list of hit terms: terms from the query that resulted in the current document being returned as a result */
    protected ArrayList<String> hitTermsList;
    /** the direction for the excerpt - controls which directions we build the excerpt from an originating hit term */
    private String direction;
    /** specifies that an excerpt is built using the terms prior to the hit term(s) */
    private static final String BEFORE = "BEFORE";
    /** specifies that an excerpt is built using the terms after the hit term(s) */
    private static final String AFTER = "AFTER";
    /** specifies that an excerpt is built using the terms that appear on both sides of the hit terms(s) */
    private static final String BOTH = "BOTH";

    /**
     * Whether we might need to trim down the excerpt to the requested size. <br>
     * Is false if this is the first time running the iterator and true otherwise.
     */
    private boolean trim;
    /** The size of half of the original desired excerpt length. Used during trimming. */
    private float origHalfSize;

    /**
     * A special term that is used to indicate we removed a candidate term because <br>
     * it was a member of the list of terms that should not be included in an excerpt.
     */
    private static final String XXXWESKIPPEDAWORDXXX = "XXXWESKIPPEDAWORDXXX";
    private static final String XXXWESKIPPEDAWORDXXX = "XXXWESKIPPEDAWORDXXX";
```

warehouse/query-core/src/main/java/datawave/query/iterator/logic/WordsAndScores.java
```diff
@@ -0,0 +1,387 @@
...
+/**
+ * An object used to save terms and their respective scores (if a score exists and is valid).
+ */
+public class WordsAndScores {
+    private static final Logger log = LoggerFactory.getLogger(WordsAndScores.class);
+
+    /** the arraylist of words */
+    private final ArrayList<String> words;
+    /** the arraylist of scores */
+    private final ArrayList<Integer> scores;
```
drewfarris:
```suggestion
/**
 * An object used to save terms and their respective scores (if a score exists and is valid). Each instance represents a word in a single position in a
 * document. There are conditions where multiple words may exist in the same position in a document, such as synonym tokens for standard text or speech-to-text
 * where multiple possible readings for single utterance exist.
 */
public class WordsAndScores {
    private static final Logger log = LoggerFactory.getLogger(WordsAndScores.class);

    /** the list of words, there may be multiple words in the same position */
    private final ArrayList<String> words;
    /**
     * the list of scores, each word in a single position may have a different score. <br>
     * the indexes in this list correspond to the indexes in the <code>words</code> array
     */
    private final ArrayList<Integer> scores;
```

warehouse/query-core/src/main/java/datawave/query/iterator/logic/WordsAndScores.java
```diff
@@ -0,0 +1,387 @@
...
+    private int longestWordIndex;
+    /** the index of the word with the smallest score (smallest non-negative is best) */
+    private int smallestScoreIndex;
+    /** the index of the hit term to output */
+    private int hitTermIndex;
+    /** the size of the words and scores arrays */
+    private int arrSize;
+
+    /** the index of the override to output */
+    private int overrideIndex;
```
drewfarris: Could your comment explain what an override is and how it's used? (Think about writing these like you'd write a dictionary definition - e.g., if the comment doesn't add much more than the variable name, it's probably not that useful)

warehouse/query-core/src/main/java/datawave/query/iterator/logic/WordsAndScores.java
```diff
@@ -0,0 +1,387 @@
...
+     * @param word
+     *            the word to add
+     * @param score
+     *            the score to add
+     * @return True if the added term is a stop word and false otherwise.
+     */
+    public boolean addTerm(String word, int score, List<String> hitTermsList) {
+        if (score < 0) {
+            return addTerm(word, hitTermsList);
+        } else {
```
drewfarris: We can remove the `else` clause since we're returning if the `score < 0` - this makes the code marginally more readable.

drewfarris (dismissed):

warehouse/query-core/src/main/java/datawave/query/iterator/logic/WordsAndScores.java
```diff
@@ -0,0 +1,501 @@
...
+
+        final int hitTermScore = scores.get(hitTermIndex);
+
+        // whether this word is longer than the existing best hit term word.
+        boolean currentTermIsLonger = word.length() > words.get(hitTermIndex).length();
+
+        // there are a couple cases here
+        // - if current has a score, and it's better than the best, we take current.
+        // - if current and best have equal scores, but current is longer, we take current instead of best
+        if (score >= 0 && hitTermScore > score) {
```
ivakegg, line 223: I am a little confused here.  Don't we want to choose the term that is the exact match with the actual hit ?

austin007008: yes but currently this deep in the code we do not have the information about which word was actually hit on in this position, we only have a list of all of the individual hit terms. 
This is something I'm planning to improve next as currently a lot of helpful data either isn't passed down this far or doesn't exist in the code yet to make those better decisions.

ivakegg (approved):

drewfarris (approved):

PR #2319, opened 2024-03-20
fixes #2303: composite and filtered query logics

### LINKED ISSUE #2303: CompositeQueryLogic needs to be aware of FilteredQueryLogic
If the composite query logic is configured with logics that are warpped with the FilteredQueryLogic, then the allMustInitialize flag cannot be set to false.  The reason is that if the filtered query logic is filtering, then it will always appear as being successful.  Hence one can get into the situation where one query logic may legitimately fail, but the second one is filtered and hence "succeeds" initialization.  In that situation the query will return as succesful went it should not have.

The solution is to make the composite query logic aware of the filtered query logic and should remove the logic from the intialized logics if it is being filtered out.

ivakegg:
* Updated to not drop filtered query logics allowing appropriate failure and success scenarios


web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
```diff
@@ -273,25 +274,42 @@ public GenericQueryConfiguration initialize(AccumuloClient client, Query setting
...
+                        if (logicQueryStringBuilder.length() > 0) {
+                            logicQueryStringBuilder.append(" || ");
+                        }
+                        logicQueryStringBuilder.append("( ( logic = '").append(logicName).append("' )");
+                        logicQueryStringBuilder.append(" && ").append("( filtered = true )").append(" )");
+                    } else {
+                        if (logicQueryStringBuilder.length() > 0) {
+                            logicQueryStringBuilder.append(" || ");
+                        }
+                        logicQueryStringBuilder.append("( ( logic = '").append(logicName).append("' )");
```
FineAndDandy: can be removed if moved up

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
```diff
@@ -273,25 +274,42 @@ public GenericQueryConfiguration initialize(AccumuloClient client, Query setting
...
+                        }
                     }
 
                 } catch (Exception e) {
                     exceptions.put(logicName, e);
                     log.error("Failed to initialize " + logic.getClass().getName(), e);
+                    if (logicQueryStringBuilder.length() > 0) {
+                        logicQueryStringBuilder.append(" || ");
+                    }
+                    logicQueryStringBuilder.append("( ( logic = '").append(logicName).append("' )");
```
FineAndDandy: can be removed if moved up

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
```diff
@@ -273,25 +274,42 @@ public GenericQueryConfiguration initialize(AccumuloClient client, Query setting
...
-                    if (isShortCircuitExecution()) {
-                        break;
+
+                    // only add this query logic to the initialized logic states if it was not simply filtered out
+                    if (logic instanceof FilteredQueryLogic && ((FilteredQueryLogic) logic).isFiltered()) {
+                        log.info("Dropping " + logic.getLogicName() + " as it was filtered out");
+                        if (logicQueryStringBuilder.length() > 0) {
+                            logicQueryStringBuilder.append(" || ");
+                        }
+                        logicQueryStringBuilder.append("( ( logic = '").append(logicName).append("' )");
```
FineAndDandy: move this above the if/else block to reduce redundancy

FineAndDandy (changes requested):

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
ivakegg: done

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
ivakegg: done

web-services/query/src/main/java/datawave/webservice/query/logic/composite/CompositeQueryLogic.java
ivakegg: done

FineAndDandy (approved):

apmoriarty (approved):

PR #2322, opened 2024-03-21
Added QueryLogIterator class

### LINKED ISSUE #2305: QueryIterators/ScanIterators should log query id first

SethSmucker:
Added a QueryLogIterator class. This iterator logs the QueryID and method name called before and after each method is executed by the following iterator.


warehouse/query-core/src/main/java/datawave/query/iterator/QueryLogIterator.java
```diff
@@ -0,0 +1,164 @@
...
+import java.util.Map;
+
+import static datawave.query.iterator.QueryOptions.QUERY_ID;
+
+public class QueryLogIterator implements SortedKeyValueIterator<Key, Value>{
+
+    protected static final Logger log = Logger.getLogger(QueryLogIterator.class);
+
+    protected SortedKeyValueIterator<Key,Value> source;
+    protected SortedKeyValueIterator<Key,Value> sourceForDeepCopies;
```
ivakegg: sourceForDeepCopies not needed, not used.  Same goes for initKeySource, seekKeySource, myEnvironment, and queryOptions.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryLogIterator.java
```diff
@@ -0,0 +1,164 @@
...
+    protected Map<String,String> queryOptions;
+    protected NestedIterator<Key> initKeySource, seekKeySource;
+    protected IteratorEnvironment myEnvironment;
+
+    protected Range range;
```
ivakegg: No need for Range, Key, Value, and yield

warehouse/query-core/src/main/java/datawave/query/iterator/QueryLogIterator.java
ivakegg: What we should be doing is attempting to pull the query id from the queryOptions for use in log messages.

ivakegg: That would be better than stoiring the entire map of options.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryLogIterator.java
```diff
@@ -0,0 +1,164 @@
...
+        this.seekKeySource = other.seekKeySource;
+        this.myEnvironment = other.myEnvironment;
+        this.queryOptions = other.queryOptions;
+
+    }
+    @Override
+    public void init(SortedKeyValueIterator<Key,Value> source, Map<String,String> options, IteratorEnvironment env) throws IOException {
+
+        try{
+            if (log.isInfoEnabled()) {
```
ivakegg: Perhaps a "start" method that takes in the name of the method .  That would make the code a lot simpler.  Also this method could be the one that sets the thread name.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryLogIterator.java
```diff
@@ -0,0 +1,164 @@
...
+                log.info("QueryLogIterator: init() Started QueryID: " + options.get(QUERY_ID));
+            }
+            this.queryOptions = options;
+            this.myEnvironment = env;
+            this.source = source;
+        }
+        finally{
+            if (log.isInfoEnabled()) {
```
ivakegg: Perhaps an "end" method akin to the start method.  Also this method could be the one that resets the thread name to its original form.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryLogIterator.java
```diff
@@ -0,0 +1,164 @@
...
+
+    @Override
+    public SortedKeyValueIterator<Key, Value> deepCopy(IteratorEnvironment iteratorEnvironment) {
+
+        QueryLogIterator newQLI = null;
+
+        try {
+            if(log.isInfoEnabled()){
+                log.info("QueryLogIterator: deepCopy() Started QueryID: " + this.queryOptions.get(QUERY_ID));
+                newQLI = new QueryLogIterator(this, this.myEnvironment);
```
ivakegg: move outside the if clause here.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryLogIterator.java
```diff
@@ -0,0 +1,164 @@
...
+        return newQLI;
+    }
+
+    @Override
+    public void seek(Range range, Collection<ByteSequence> collection, boolean b) throws IOException {
+
+        try {
+            if(log.isInfoEnabled()){
+                log.info("QueryLogIterator: seek() Started QueryID: " + this.queryOptions.get(QUERY_ID));
+                this.source.seek(range, collection, b);
```
ivakegg: move outside of the if clause

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -467,9 +472,14 @@ protected CloseableIterable<QueryData> process(ScannerFactory scannerFactory, Me
                 cfg = getQueryIterator(metadataHelper, config, settings, "", false, false);
             }
             configureIterator(config, cfg, newQueryString, isFullTable);
+
+            while (null == logCfg) {
+                logCfg = getQueryIterator(metadataHelper, config, settings, "", false, false);
```
ivakegg: I think this should be getQueryLogIterator

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -437,6 +439,7 @@ public class QueryOptions implements OptionDescriber {
 
     private Map<String,Long> fieldCounts;
     private Map<String,Long> termCounts;
+    private boolean activeTserverLogging = false;
```
ivakegg: lets make this "true" by default

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
```diff
@@ -469,6 +469,8 @@ public class ShardQueryConfiguration extends GenericQueryConfiguration implement
      */
     private boolean useTermCounts = false;
 
+    private boolean isTserverLoggingActive = false;
```
ivakegg: lets make this true by default

PR #2324, opened 2024-03-21
Improvements to SSDeepSimilarityDiscoveryQuery

drewfarris:
* When scoring a match based on ngram overlaps, collect the actual matching ngrams - not just the score.
* Use the matching ssdeep from the SSDeepSimilarityQuery in the SSDeepSimilarityDiscoveryQuery results, instead of the ssdeep returned from the Discovery query, because the field in the index may have been downcased, whereas the similarity query results retain the original case.
* Rename MATCH_SCORE to OVERLAP_SCORE to be clearer about what the score represents.
* Added return the list of overlapping ssdeep NGrams from the SSDeepSimilarityQueryLogic in the OVERLAP_SSDEEP_NGRAMS field.
* Added unit tests for overlapping ngrams

### DIFF
--- a/warehouse/ssdeep-common/src/main/java/datawave/util/ssdeep/SSDeepHashEditDistanceScorer.java
+++ b/warehouse/ssdeep-common/src/main/java/datawave/util/ssdeep/SSDeepHashEditDistanceScorer.java
@@ -5,7 +5,7 @@ import org.apache.log4j.Logger;
 
 /**
  * Implements functions to calculate a similarity score for a pair of SSDeepHashes. This implementation will calculate an edit distance between two hashes and
- * the create a normalized score between 0-100, with 100 being a great match and 0 being a horrible match. This is designed to give relatively equal footing to
+ * then create a normalized score between 0-100, with 100 being a great match and 0 being a horrible match. This is designed to give relatively equal footing to
  * hash comparisons that have different lenghts.
  */
 public class SSDeepHashEditDistanceScorer implements SSDeepHashScorer<Integer> {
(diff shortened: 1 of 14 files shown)


warehouse/ssdeep-common/src/main/java/datawave/util/ssdeep/SSDeepHashEditDistanceScorer.java
```diff
@@ -3,8 +3,12 @@
 import org.apache.commons.text.similarity.LevenshteinDistance;
 import org.apache.log4j.Logger;
 
-/** Implements functions to calculate a similarity score for a pair of SSDeepHashes */
-public class SSDeepHashEditDistanceScorer implements SSDeepHashScorer {
+/**
+ * Implements functions to calculate a similarity score for a pair of SSDeepHashes. This implementation will calculate an edit distance between two hashes and
+ * the create a normalized score between 0-100, with 100 being a great match and 0 being a horrible match. This is designed to give relatively equal footing to
```
KhakiField: Minor typo
```suggestion
 * then create a normalized score between 0-100, with 100 being a great match and 0 being a horrible match. This is designed to give relatively equal footing to
```

warehouse/ssdeep-common/src/main/java/datawave/util/ssdeep/SSDeepHashScorer.java
```diff
@@ -1,5 +1,10 @@
 package datawave.util.ssdeep;
 
-public interface SSDeepHashScorer {
-    public int apply(SSDeepHash signature1, SSDeepHash signature2);
+/**
+ * An interface for things that take two hashes and compare the to produce some sort of result.
```
KhakiField: Minor typo
```suggestion
 * An interface for things that take two hashes and compare them to produce some sort of result.
```

apmoriarty (approved):

MiguelRicardos (approved):

PR #2326, opened 2024-03-22
Age off modulation and new formatting utility  

matthpeterson:
- Extracts age-off code and tests to its own module
- Creates a new module containing age-off related utilities, such as generation of xml configuration files from a csv
- Ensures no transitive dependencies in new modules
- Eliminates transitive dependencies from common-test


warehouse/age-off-utils/src/main/java/datawave/age/off/util/AgeOffFileGenerator.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /**
+     * Writes the file to the writer.
+     *
+     * @throws IOException
+     *             i/o exception with writer
+     */
+    public void format() throws IOException {
+        openConfigurationElement();
```
jschmidt10, line 32: I like this style of declarative generation of the XML. If we wanted to one day take this further, we could make an object hierarchy that reflects the structure of the XML file and then just let something like jaxb or jackson serialize it out as XML at the end. Then adding a rule or a parent would just be manipulating Java objects rather than needing to properly format each line of output. 

I do think this is a good first step though if we don't want to do the whole object style initially - it's probably a significant rework.

warehouse/age-off-utils/src/main/java/datawave/age/off/util/AgeOffRuleConfiguration.java
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        @Override
+        public void write(String str) throws IOException {
+            writer.write(this.indentation + str);
+        }
+
+        @Override
+        public void write(char[] chars, int i, int i1) throws IOException {
+            writer.write(chars, i, i1);
```
jschmidt10: For the sake of preserving the API, can we either add the indent here or throw an `UnsupportedOperationException`? That way no one calls this thinking the indent will be respected. I think the arguments are `write(char[] chars, int begin, int length)`

warehouse/age-off-utils/src/main/java/datawave/age/off/util/AgeOffFileGenerator.java
```diff
@@ -0,0 +1,69 @@
...
+        for (AgeOffRuleConfiguration ruleConfiguration : this.configuration.getRuleConfigurations()) {
+            writeRule(ruleConfiguration);
+        }
+
+        this.writer.write(this.indent + "</rules>\n");
+    }
+
+    private void writeRule(AgeOffRuleConfiguration ruleConfiguration) throws IOException {
+        log.debug("formatting ruleConfiguration {}", ruleConfiguration.getRuleLabel());
+        new AgeOffRuleFormatter(ruleConfiguration).format();
```
jschmidt10: The writer object is sorta creeping across all of the Configuration and Formatter objects. I like the division between Conf vs Formatter where the Conf holds all the state we need to write into the XML and the Formatter knows how to convert a Conf to XML. But the writer creeping into the Conf objects makes it a bit hard to keep track of. 

What do you think about eliminating the Writer from any Conf objecs and requiring the Writer to be passed into any Formatter calls? If we need to keep track of the indent level, we could make like an XMLIndentingWriter like you have already but maybe open up an "indent()" and "outdent()" API.


```
writer = new IndentingXMLWriter(rawWriter);

// in AgeOffFileGenerator
for ruleConf in rulesConfs
  writer.indent();
  new AgeOffRuleFormatter(ruleConf).format(writer);
  writer.outdent();

// in AgeOffRuleFormatter
new AgeOffCsvToMatchPatternFormatter(matchConf).format(writer);

```

warehouse/age-off-utils/src/main/java/datawave/age/off/util/AgeOffCsvToMatchPatternFormatter.java
```diff
@@ -0,0 +1,212 @@
...
+    private void reformatLine(String[] tokens) throws IOException {
+        StringBuilder sb = new StringBuilder();
+
+        appendLabel(tokens, sb);
+
+        appendLiteral(tokens, sb);
+
+        appendEquivalenceSymbol(sb);
+
+        appendValue(sb, tokens);
```
jschmidt10: Nit: all the other functions use the parameter order: `tokens, sb`. I don't think I feel strongly about the ordering but aligning them all would make it quicker to follow.

warehouse/age-off-utils/src/main/java/datawave/age/off/util/AgeOffCsvToMatchPatternFormatterConfiguration.java
```diff
@@ -0,0 +1,159 @@
...
+    private String staticLabel;
+    private char quoteCharacter;
+    private boolean shouldQuoteLiteral;
+    private boolean useColons; // when false, use equals sign
+    private boolean shouldPadEquivalence; // when false, no spaces around equals sign
+    private boolean shouldLowerCaseLiterals;
+    private boolean shouldUpperCaseLiterals;
+    private Map<String,String> valueMapping;
+    private Writer writer;
+    private BufferedReader reader;
```
jschmidt10: If you are down with the proposal about Confs just holding the raw data and Formatters printing, then I'd suggest we get rid of the reader object in the Conf. Can we just read the CSV data during the building of the conf object and put the data in the conf - the most straightforward would be as Strings for each line.

If you wanted to stream these readers for memory reasons, then we could do something slicker where you just give the Conf an Iterator<String> so it's ignorant about the reader under the covers. Then the iterator could invoke the BufferedReader.readLine() and handle closing the stream at the end.

warehouse/age-off-utils/src/main/java/datawave/age/off/util/AgeOffCsvToMatchPatternFormatter.java
```diff
@@ -0,0 +1,212 @@
...
+        if (label.length() == 0) {
+            log.error("Unable to apply non-empty label {}", Arrays.toString(tokens));
+            throw new IllegalStateException("Empty token");
+        }
+        sb.append(label).append(SPACE);
+    }
+
+    private void appendLiteral(String[] tokens, StringBuilder sb) {
+        if (tokens.length <= columnInformation.patternColumnNumber) {
+            log.error("Unable to process literal {}", Arrays.toString(tokens));
```
jschmidt10, line 175: The logging of an error followed by throwing an Exception can usually lead to double logging of the Exception as the calling code with either catch it or let it kill the JVM (and will be logged at that level).

I think it's usually better to either 1) print it if you do not throw or 2) throw it.

warehouse/age-off-utils/src/main/java/datawave/age/off/util/AgeOffCsvToMatchPatternFormatter.java
```diff
@@ -0,0 +1,212 @@
...
+            throw new IllegalStateException("Not enough tokens");
```
jschmidt10, line 168: Instead of logging and throwing, what do you think about adding this context to the IllegalStateException. I think logging part of the error and throwing the Exception could potentially lead to the error message and the stacktrace being separated. Like if the caller catches the IllegalStateException and chooses not to log it for instance.

It could look like:
```
throw new IllegalStateException("Not enough tokens when parsing literal from " + Arrays.toString(tokens));
```

or something like that.

warehouse/age-off-utils/src/test/java/datawave/age/off/util/AgeOffCsvToMatchPatternFormatterTest.java
```diff
@@ -0,0 +1,394 @@
...
+    private static final String HEADER_WITHOUT_LABEL = "pattern,duration";
+
+    // @formatter:off
+    private static final String INPUT_TEXT =
+            "bakingPowder, 365d\n" +
+            "driedBeans,548d\n" +
+            "bakingSoda,\t720d\n" +
+            "      coffeeGround       ,        90d\n          " +
+            "coffeeWholeBean        ,183d\n" +
+            "         coffeeInstant,730d\n";
```
jschmidt10: Don't forget:

```
twinkies, infinite
```

warehouse/age-off-utils/src/test/java/datawave/age/off/util/AgeOffCsvToMatchPatternFormatterTest.java
```diff
@@ -0,0 +1,394 @@
...
+        StringWriter out = new StringWriter();
+        builder.setWriter(out);
+        AgeOffCsvToMatchPatternFormatter generator = new AgeOffCsvToMatchPatternFormatter(builder.build());
+        generator.write();
+
+        return out.toString();
+    }
+
+    private static String adjustEachLine(String input, Function<String,String> function) {
+        return Arrays.stream(input.split("\\n")).map(item -> {
```
jschmidt10: The `map()` method actually accepts a function, so you could do

```
Arrays.stream(input.split("\\n")).map(function).collect(Collectors.joining(...)
```
Maybe for readability, you could rename `function` to `lineFormatter` or something like that.

warehouse/age-off-utils/src/test/java/datawave/age/off/util/AgeOffCsvToMatchPatternFormatterTest.java
matthpeterson: Addressed in latest push.

warehouse/age-off-utils/src/test/java/datawave/age/off/util/AgeOffCsvToMatchPatternFormatterTest.java
matthpeterson: Good catch!  Addressed in latest push.

warehouse/age-off-utils/src/main/java/datawave/age/off/util/AgeOffCsvToMatchPatternFormatter.java
matthpeterson: Addressed in latest push.

warehouse/age-off-utils/src/main/java/datawave/age/off/util/AgeOffCsvToMatchPatternFormatterConfiguration.java
matthpeterson: The Iterator idea worked better than I expected.  Thanks for the feedback!

warehouse/age-off-utils/src/main/java/datawave/age/off/util/AgeOffCsvToMatchPatternFormatter.java
matthpeterson: Addressed in latest push.

warehouse/age-off-utils/src/main/java/datawave/age/off/util/AgeOffFileGenerator.java
matthpeterson: I removed the reader and writer from the Configuration objects.  I think that was good feedback.  The writer is now being passed to the format method which seems cleaner.  I stuck with the delegating indenting writer approach instead of changing it to do the indent/outdent.  Hopefully it looks ok this way now that the writer is passed to format.

warehouse/age-off-utils/src/main/java/datawave/age/off/util/AgeOffRuleConfiguration.java
matthpeterson: Good call.  Addressed in latest push.

warehouse/age-off-utils/src/main/java/datawave/age/off/util/AgeOffFileGenerator.java
matthpeterson: The object hierarchy is tough because there are so many different implementations of filters that make the rule so different.

I tried changing to XMLStreamWriter but it didn't support new lines nor indentation very well.  I reverted those changes for now.  

This is being left unresolved for now while I try a few alternatives.

matthpeterson: The implementation in the latest commit, using xml Document objects and a Transformer, worked out better than XMLStreamWriter.

jschmidt10 (dismissed):

dtspence (dismissed):

jschmidt10 (approved):

dtspence (approved):

PR #2328, opened 2024-03-28
Rebuild Attributes when reducing type metadata

apmoriarty:
Add AttributeRebuilder for enriching return types when reducing type metadata


warehouse/query-core/src/main/java/datawave/query/util/transformer/AttributeRebuilder.java
```diff
@@ -0,0 +1,123 @@
...
+            log.trace("Field " + field + " not found in TypeMetadata, falling back to QueryModel");
+        }
+
+        if (queryModel == null) {
+            log.warn("QueryModel is null, cannot populate normalizers for " + field + " from model");
+            return;
+        }
+
+        // check forward mappings
+        Multimap<String,String> forwardMappings = queryModel.getForwardQueryMapping();
```
ivakegg: I think you may be going the wrong way on the mappings here.  At the point where you are rebuilding attributes, if you have the fields as they exist in the DB then there is no mapping required.  If you have the attributes after applying the reverse mappings, then you would nee the reverse mappings here.  In any case I cannot see how the forward mappings would be applicable here.

warehouse/query-core/src/main/java/datawave/query/util/transformer/AttributeRebuilder.java
```diff
@@ -0,0 +1,124 @@
...
+        if (queryModel == null) {
+            this.fieldMap = new HashMap<>();
+        } else {
+            this.fieldMap = invertMap(queryModel.getReverseQueryMapping());
+        }
+        this.classCache = new HashMap<>();
+    }
+
+    private Map<String,String> invertMap(Map<String,String> map) {
+        Map<String,String> mappings = new HashMap<>();
```
ivakegg: This may need to be a Multimap.  There is nothing preventing us from having multiple fields with the same reverse mapping, although I don't believe we do this in practive.

apmoriarty: That sounds correct. How should the multiple normalizer case be handled? I could see argument for retaining a single normalizer, or duplicating values in the document one per unique normalizer.

avgAGB (dismissed):

apmoriarty:
Closing in favor of: https://github.com/NationalSecurityAgency/datawave/pull/2652

PR #2331, opened 2024-04-01
Add file integrity checks before service installation in dw-quickstart

foster33:
**Idea behind creating this PR:**
I noticed that some of the services' distribution URIs were using http and subsequently causing a 301 when being downloaded. The initial request before the redirect could pose a security issue. An even greater issue can arise for a user who puts http distribution links for their services and those URIs do not have redirects. There was then no file integrity check on the tarballs before they were installed.

**This PR adds:**
- Tarball checksum verification before services are installed
- A prompt to the user when an http link is detected, warning them of the potential security risk
- Updating of the distribution URIs that were using http and causing the 301s to use https


avgAGB (dismissed):

contrib/datawave-quickstart/bin/services/accumulo/bootstrap.sh
```diff
@@ -15,7 +15,10 @@ DW_ACCUMULO_SERVICE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 # Zookeeper config
 
 # You may override DW_ZOOKEEPER_DIST_URI in your env ahead of time, and set as file:///path/to/file.tar.gz for local tarball, if needed
+# DW_ZOOKEEPER_DIST_URI should, if possible, be using https. There are potential security risks by using http.
 DW_ZOOKEEPER_DIST_URI="${DW_ZOOKEEPER_DIST_URI:-https://archive.apache.org/dist/zookeeper/zookeeper-3.7.1/apache-zookeeper-3.7.1-bin.tar.gz}"
+# The sha512 checksum for the tarball. Value should be the hash value only and does not include the file name. Cannot be left blank.
+DW_ZOOKEEPER_DIST_SHA512_CHECKSUM="9103628a50745fa1a289bca666fda4a9c08ec17c55cf13e66887e7ba76e93dbae60a1f1ffd6c10798be3a16069344ecbc00cebb29bf03d9cd7096ccd098ed011"
```
ivakegg: Is there not a separate link to pull down the checksum?

foster33: Yes. There are separate links for most, if not all, the service dist URIs in the quickstart. My thought behind not using them was that if users decide to use other distribution URIs, those might not have specific links to pull the checksums. I can change the logic here and make it use either a link or a hash. Maybe my reasoning is wrong... but I am open to more feedback.

keith-ratcliffe: Whether separate link or not, there needs to be a way to override the value in the user's environment, as with the `*DIST_URI` values, so that the user doesn't have to modify the source code here to experiment with different versions. E.g....
```
DW_ZOOKEEPER_DIST_SHA512_CHECKSUM=${DW_ZOOKEEPER_DIST_SHA512_CHECKSUM:-9103628a50745fa1a289bca666fda4a9c08ec17c55cf13e66887e7ba76e93dbae60a1f1ffd6c10798be3a16069344ecbc00cebb29bf03d9cd7096ccd098ed011}
```
With that said, it's probably more straightforward just to leave as checksums, rather than links to checksums, so that the user can easily build/install their own local version of a tarball, if desired. That's a fairly common use case, as when testing Accumulo SNAPSHOT versions, for example

keith-ratcliffe (changes requested):

contrib/datawave-quickstart/bin/services/accumulo/bootstrap.sh
foster33: Good point. I went ahead and made changes based on your feedback.

keith-ratcliffe (approved):

avgAGB (approved):

PR #2333, opened 2024-04-02
Optionally build a precise datatype filter when no filter is specified

apmoriarty:
- added option to build a precise datatype filter
- improved IngestTypeVisitor
- the 'precise datatype filter' option *should* play nicely with the 'reduce ingest types' option

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypeVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypeVisitor.java
@@ -29,8 +29,6 @@ import org.apache.commons.jexl3.parser.ASTReferenceExpression;
 import org.apache.commons.jexl3.parser.JexlNode;
 import org.apache.log4j.Logger;
 
-import com.google.common.collect.Sets;
-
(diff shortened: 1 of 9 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypeVisitor.java
```diff
@@ -262,7 +260,7 @@ public Set<String> getIngestTypesForJunction(JexlNode node) {
     public Set<String> getIngestTypesForLeaf(JexlNode node) {
```
ivakegg, line 266: Why did we create a separate visitor for this?  Why not use the visitor that pulls fields from the query (which we duplicate for the most part in here), and then simply get all of the datatypes for that list of fields?  I feel this may be fragile as there may be changes to the query fields visitor that would not get incorporated in here and it would be a long time before that would be discovered.

apmoriarty: Updated the docs on the visitor per our conversation

ivakegg (approved):

mineralntl (approved):

PR #2334, opened 2024-04-02
RemoteHttpService Invalid Exception Fix

jack-gitdev:
Fixes the readVoidResponse method logic to avoid an invalid runtime exception being thrown.


apmoriarty (approved):
LGTM - would love to see a unit test that replicates the problem and verifies the solution

ivakegg (approved):

hlgp (approved):

PR #2347, opened 2024-04-17
Finish refactoring IngestTypeVisitor, add datatype focused test

apmoriarty:
- completed refactoring datatype-gathering logic from the IngestTypePruningVisitor to the IngestTypeVisitor.
- add ShapesTest, a WiseGuys-style test with a focus on querying across datatypes
- updated IngestTypePruningVisitor and IngestTypeVisitor tests based on integration testing

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypePruningVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypePruningVisitor.java
@@ -1,9 +1,8 @@
 package datawave.query.jexl.visitors;
 
-import java.util.Collections;
-import java.util.HashMap;
+import static datawave.query.jexl.visitors.IngestTypeVisitor.UNKNOWN_TYPE;
+
 import java.util.HashSet;
-import java.util.Map;
 import java.util.Set;
 
 import org.apache.commons.jexl3.parser.ASTAndNode;
@@ -49,20 +48,12 @@ import datawave.query.util.TypeMetadata;
 public class IngestTypePruningVisitor extends BaseVisitor {
     private static final Logger log = Logger.getLogger(IngestTypePruningVisitor.class);
 
-    private static final String UNKNOWN_TYPE = "UNKNOWN_TYPE";
-
-    // cache expensive calls to get ingest types per field
-    private final TypeMetadata typeMetadata;
-    private final Map<String,Set<String>> ingestTypeCache;
-
     private int termsPruned = 0;
     private int nodesPruned = 0;
 
     private final IngestTypeVisitor ingestTypeVisitor;
 
     public IngestTypePruningVisitor(TypeMetadata typeMetadata) {
-        this.typeMetadata = typeMetadata;
-        this.ingestTypeCache = new HashMap<>();
         this.ingestTypeVisitor = new IngestTypeVisitor(typeMetadata);
     }
 
@@ -165,12 +156,33 @@ public class IngestTypePruningVisitor extends BaseVisitor {
 
     @Override
     public Object visit(ASTReferenceExpression node, Object data) {
-        return visitOrPrune(node, data);
+        Object o = node.jjtGetChild(0).jjtAccept(this, data);
+        if (node.jjtGetNumChildren() == 0) {
+            pruneNodeFromParent(node);
+        }
+        return o;
     }
 
     @Override
     public Object visit(ASTOrNode node, Object data) {
-        return visitOrPrune(node, data);
+
+        if (data == null) {
+            // just a visit
+            return ingestTypeVisitor.getIngestTypesForJunction(node);
+        }
+
+        Set<String> pruningTypes = (Set<String>) data;
+
+        // must traverse the children in reverse order because of pruning
+        for (int i = node.jjtGetNumChildren() - 1; i >= 0; i--) {
+            node.jjtGetChild(i).jjtAccept(this, pruningTypes);
+        }
+
+        if (node.jjtGetNumChildren() == 0) {
+            pruneNodeFromParent(node);
+        }
+
+        return pruningTypes;
     }
 
     /**
@@ -188,16 +200,20 @@ public class IngestTypePruningVisitor extends BaseVisitor {
 
         QueryPropertyMarker.Instance instance = QueryPropertyMarker.findInstance(node);
         if (instance.isAnyType()) {
-            return visitMarker(instance, node, data);
+            Object o = visitMarker(instance, node, data);
+            if (node.jjtGetNumChildren() == 0) {
+                pruneNodeFromParent(node);
+            }
+            return o;
         }
 
         // getting ingest types for an intersection is different
-        Set<String> ingestTypes = getIngestTypesForIntersection(node);
+        Set<String> ingestTypes = ingestTypeVisitor.getIngestTypesForIntersection(node);
 
         // automatically prune if there is no common ingest type
         if (ingestTypes.isEmpty()) {
             pruneNodeFromParent(node);
-            return Collections.emptySet();
+            return new HashSet<>();
         }
 
         // the AndNode is where we can generate a set of ingest types used to prune child nodes
@@ -244,7 +260,7 @@ public class IngestTypePruningVisitor extends BaseVisitor {
         // ExceededOr marker can be handled on its own
         if (instance.isType(QueryPropertyMarker.MarkerType.EXCEEDED_OR)) {
             String field = new ExceededOr(instance.getSource()).getField();
-            Set<String> ingestTypes = getIngestTypesForField(field);
+            Set<String> ingestTypes = ingestTypeVisitor.getIngestTypesForField(field);
             if (data instanceof Set<?>) {
                 return pruneLeaf(ingestTypes, node, data);
             }
@@ -254,7 +270,7 @@ public class IngestTypePruningVisitor extends BaseVisitor {
         JexlNode source = node.jjtGetChild(1);
         Set<String> dts = (Set<String>) source.jjtAccept(this, data);
 
-        if (source.jjtGetParent() == null) {
+        if (source.jjtGetParent() == null || source.jjtGetNumChildren() == 0) {
             pruneNodeFromParent(node);
         }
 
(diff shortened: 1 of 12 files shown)


warehouse/query-core/src/test/java/datawave/query/ShapesTest.java
```diff
@@ -0,0 +1,864 @@
...
+import datawave.query.tables.ShardQueryLogic;
+import datawave.query.tables.edge.DefaultEdgeEventQueryLogic;
+import datawave.query.util.ShapesIngest;
+import datawave.util.TableName;
+import datawave.webservice.edgedictionary.RemoteEdgeDictionary;
+import datawave.webservice.query.QueryImpl;
+import datawave.webservice.query.configuration.GenericQueryConfiguration;
+
+/**
+ * A set of tests that emphasis the influence of datatypes on query planning and execution
```
foster33:
```suggestion
 * A set of tests that emphasize the influence of datatypes on query planning and execution
```
Minor typo

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypeVisitor.java
```diff
@@ -82,11 +83,13 @@ public static Set<String> getIngestTypes(JexlNode node, TypeMetadata typeMetadat
         Object o = node.jjtAccept(visitor, null);
         if (o instanceof Set) {
             Set<String> ingestTypes = (Set<String>) o;
-            if (!ingestTypes.contains(UNKNOWN_TYPE)) {
+            if (ingestTypes.contains(UNKNOWN_TYPE)) {
+                ingestTypes.retainAll(Collections.singleton(UNKNOWN_TYPE));
                 return ingestTypes;
```
alerman, line 90: Duplicated return. Remove this one and the flow doesnt change.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypePruningVisitor.java
```diff
@@ -324,7 +340,7 @@ private Set<String> pruneJunction(JexlNode node, Object data) {
         for (int i = node.jjtGetNumChildren() - 1; i >= 0; i--) {
             node.jjtGetChild(i).jjtAccept(this, data);
         }
-        return Collections.emptySet();
+        return new HashSet<>();
```
MiguelRicardos, line 360: Is this actually more efficient?

Collections.emptySet() is an immutable object that isn't going to allocate more memory every time that this is called, whereas a new HashSet<>() is going to. Even if it's small, it seems like this could be bad for memory management.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IngestTypePruningVisitorTest.java
```diff
@@ -438,7 +441,7 @@ public void testEvaluationOnlyField() {
     @Test
     public void testPruneNegation() {
         String query = "A == '1' || !((_Delayed_ = true) && (A == '1' && C == '2'))";
-        test(query, "A == '1'");
+        test(query, query);
```
alerman: Does this indicate we are not pruning the negation, if the expected now has the negation included?

alerman (commented):
Just a few small things. Nothing too major. Please check the unit test with changed expected output.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypePruningVisitor.java
```diff
@@ -165,12 +156,33 @@ public Object visit(ASTReference node, Object data) {
...
+        }
+
+        Set<String> pruningTypes = (Set<String>) data;
+
+        // must traverse the children in reverse order because of pruning
+        for (int i = node.jjtGetNumChildren() - 1; i >= 0; i--) {
+            node.jjtGetChild(i).jjtAccept(this, pruningTypes);
+        }
+
+        if (node.jjtGetNumChildren() == 0) {
```
MiguelRicardos, line 198: Should this check be placed before the for loop?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypeVisitor.java
apmoriarty: Not quite, in this case I only want to return a collection that contains the `UNKNOWN_TYPE`. I'll add a comment to that effect.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypePruningVisitor.java
apmoriarty: The collection returned by `Collections.emptySet()` will throw an UnsupportedOperationException in some very unfortunate circumstances, so I ripped out all Collections and Guava utility methods in favor of "native" collections. Union and intersection operations are now done via `.addAll` and `.retain`

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypeVisitor.java
```diff
@@ -354,25 +353,28 @@ public Set<String> getIngestTypesForField(String field) {
...
             if (ingestTypes.isEmpty()) {
                 ingestTypes = childIngestTypes;
             } else {
-                if (child instanceof ASTNotNode) {
-                    // special handling of negations. negated ingest types get OR'd together
-                    ingestTypes.addAll(childIngestTypes);
-                } else {
-                    ingestTypes = intersectTypes(ingestTypes, childIngestTypes);
-                }
+                ingestTypes = intersectTypes(ingestTypes, childIngestTypes);
```
MiguelRicardos, line 413: Could you add a comment or explain why the special handling of negations for instances of ASTNNotNode's are no longer needed?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypePruningVisitor.java
apmoriarty: No, a child of a union can self-prune in certain cases. Consider the query `(A && B) || (C && D)` where each term maps to a separate, unique datatype (no overlaps). The `(A && B)` branch will self prune to zero, and the `(C && D)` branch will prune to zero. Thus the union node now has zero children and must prune itself from it's parent.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypeVisitor.java
```diff
@@ -385,7 +387,9 @@ private Set<String> getIngestTypesForIntersection(ASTAndNode node) {
 
     private Set<String> intersectTypes(Set<String> typesA, Set<String> typesB) {
         if (typesA.contains(UNKNOWN_TYPE) || typesB.contains(UNKNOWN_TYPE)) {
-            return Collections.singleton(UNKNOWN_TYPE);
+            Set<String> unknown = new HashSet<>();
```
MiguelRicardos, line 435: Could you explain the benefit of going from:
```
            return Collections.singleton(UNKNOWN_TYPE);
``` 
to 
```
Set<String> unknown = new HashSet<>();
unknown.add(UNKNOWN_TYPE);
return unknown;
```
Is there a reason to go from an immutable Set<String> to a mutable Set<String> if all it's doing is returning one with the UNKNOWN_TYPE?

warehouse/query-core/src/test/java/datawave/query/CompositeFunctionsTest.java
```diff
@@ -491,7 +493,7 @@ public void testNotNulls() throws Exception {
                 "filter:isNotNull(UUID) || filter:isNotNull(NULL1)",
                 "filter:isNotNull(ONE_NULL)",
                 "UUID =~ '^[CS].*' AND filter:isNotNull(UUID)",
-                "UUID =~ '^[CS].*' AND filter:isNotNull(NULL1)",
+                "UUID =~ '^[CS].*' AND filter:isNotNull(NULL1)",    //  dis guy
```
MiguelRicardos: Could you add a more descriptive comment?

MiguelRicardos (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypePruningVisitor.java
MiguelRicardos: Can you add a comment to explain that this will only happen in some instances? Do we have a unit test that shows this?

apmoriarty: I'll add a comment. You can look at the `testDoubleNestedPruning()` unit test for some examples.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypeVisitor.java
apmoriarty: The result will likely have set operations performed (add, addAll, retain) so it can't be immutable.

warehouse/query-core/src/test/java/datawave/query/CompositeFunctionsTest.java
apmoriarty: That's a leftover comment, I'll remove it

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IngestTypePruningVisitorTest.java
apmoriarty: Previously the negation would self-prune, let me take a closer look at why this stopped visiting the negated branch..

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypeVisitor.java
apmoriarty: Does the method level comment answer the question about negations, or would you like to see something more explicit?

avgAGB (approved):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypePruningVisitor.java
MiguelRicardos: SGTM

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypeVisitor.java
MiguelRicardos: LGTM

MiguelRicardos (approved):
Nice work!

PR #2352, opened 2024-04-18
Feature/fi reindex integration

FineAndDandy:
Added support for batch processing, fixed bugs with tf processing offsets for multivalued fields, and extended unit test coverage


warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexMapper.java
foster33, line 1: There are a handful of places within `ShardReindexMapper` that mention TODOs for updating / adding documentation. We should probably make sure these are taken care of and then removed.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexJob.java
```diff
@@ -114,7 +114,39 @@ private Job setupJob() throws IOException, ParseException, AccumuloException, Ta
         }
 
         // set the propagate deletes flag
```
jschmidt10: This comment doesn't match the code anymore

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexJob.java
```diff
@@ -139,31 +171,72 @@ private Job setupJob() throws IOException, ParseException, AccumuloException, Ta
             throw new IllegalStateException("split.work.dir and job.output.table.names must be configured");
         }
 
+        // test that each of the output table names exist
+        Properties accumuloProperties = builder.build();
+        AccumuloClient client = Accumulo.newClient().from(accumuloProperties).build();
```
jschmidt10: This client should be closed. I also think Accumulo recommends creating a single client object and reusing for the duration of the app now. Its supposed to be a heavier weight utility than the old Connector objects. You use another client in your `getSplitsFromMetadata()` method. Might be worth promoting the client to a class member and closing it at the end.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexJob.java
```diff
@@ -139,31 +171,72 @@ private Job setupJob() throws IOException, ParseException, AccumuloException, Ta
...
+        for (String table : configuration.get("job.output.table.names").split(",")) {
```
jschmidt10: Blocks of code with a describing comment are great candidates for methods:

```
validateTablesExist(client, tableNames);
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexJob.java
```diff
@@ -139,31 +171,72 @@ private Job setupJob() throws IOException, ParseException, AccumuloException, Ta
...
+                    ranges = buildFiRanges(jobConfig.startDate, jobConfig.endDate, jobConfig.splitsPerDay);
+                }
 
-            Properties accumuloProperties = builder.build();
+                if (ranges.size() == 0) {
+                    throw new IllegalArgumentException("no ranges created from start: " + jobConfig.startDate + " end: " + jobConfig.endDate);
+                }
+
+                for (Range r : ranges) {
+                    log.info("Range: " + r);
```
jschmidt10: Can we add a little more context to this message? Maybe plug in the table name or highlight that these are going to be used as input splits. Sometimes our jobs scan so many tables its tough to know what this log line is telling us.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexJob.java
```diff
@@ -200,7 +273,47 @@ private Job setupJob() throws IOException, ParseException, AccumuloException, Ta
...
+
+        AccumuloClient client = Accumulo.newClient().from(accumuloProperties).build();
+        String tableId = client.tableOperations().tableIdMap().get(tableName);
+
+        if (tableId == null) {
+            throw new RuntimeException("Could not locate table: '" + tableName + "'");
+        }
+
+        try {
+            Scanner s = client.createScanner("accumulo.metadata");
```
jschmidt10: Ideally, we'd close the client and the scanner if an exception occurs to, with a finally block or better yet with a try-with-resources:

```
try (AccumuloClient client = .... ;
      Scanner metadataScanner = .... ) {
        // rest of your stuff
}
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexJob.java
```diff
@@ -200,7 +273,47 @@ private Job setupJob() throws IOException, ParseException, AccumuloException, Ta
...
+
+            ScannerBase.ConsistencyLevel consistencyLevel = ScannerBase.ConsistencyLevel.IMMEDIATE;
+            if (jobConfig.useScanServers) {
```
jschmidt10: I have never considered using scan servers for the accumulo.metadata table. Does that mean we might miss RFiles or attempt to read RFiles that have been GC'd?

jschmidt10: I guess the same is true even if we don't use the scan servers - so nvm!

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexMapper.java
```diff
@@ -0,0 +1,764 @@
...
+import datawave.ingest.data.config.ingest.IngestHelperInterface;
+import datawave.ingest.mapreduce.ContextWrappedStatusReporter;
+import datawave.ingest.mapreduce.handler.DataTypeHandler;
+import datawave.ingest.mapreduce.handler.shard.ShardedDataTypeHandler;
+import datawave.ingest.mapreduce.job.BulkIngestKey;
+import datawave.ingest.mapreduce.partition.MultiTableRangePartitioner;
+import datawave.ingest.protobuf.Uid;
+
+public class ShardReindexMapper extends Mapper<Key,Value,BulkIngestKey,Value> {
+    public static final String CLEANUP_SHARD = "cleanupShard";
```
jschmidt10: Thoughts on prefixing these? It can be nice to search the GUI for `ShardReindexMapper` and see all the applicable options.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexMapper.java
```diff
@@ -0,0 +1,764 @@
...
+                throw new IllegalArgumentException("can not create handler for data type handler: " + dataTypeHandler, e);
+            }
+
+            this.generateMetadata = config.getBoolean(GENERATE_METADATA, this.generateMetadata);
+
+            try {
+                this.event = new RawRecordContainerImpl();
+                if (eventOverride != null) {
+                    log.info("creating event override: " + this.eventOverride);
+                    this.event = (RawRecordContainer) Class.forName(this.eventOverride).getDeclaredConstructor().newInstance();
```
jschmidt10: Hadoop has a util to instantiate a class and inject the conf:

```
this.event = ReflectionUtils.newInstance(eventClass, config);
```
You could also use it for the data type handler above.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexMapper.java
```diff
@@ -0,0 +1,764 @@
...
+                    }
+                }
+            } catch (ClassNotFoundException | InvocationTargetException | InstantiationException | IllegalAccessException | NoSuchMethodException e) {
+                throw new RuntimeException("Could not create event of type: " + this.eventOverride, e);
+            }
+
+            this.enableReindexCounters = config.getBoolean(ENABLE_REINDEX_COUNTERS, this.enableReindexCounters);
+            this.dumpCounters = config.getBoolean(DUMP_COUNTERS, this.dumpCounters);
+            if (this.enableReindexCounters) {
+                this.counters = new HashMap<>();
```
jschmidt10, line 212: `counters` was already initialized in the member declaration

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexMapper.java
```diff
@@ -0,0 +1,764 @@
...
+        if (this.enableReindexCounters) {
+            for (String counterGroup : this.counters.keySet()) {
+                Map<String,Long> groupCounters = this.counters.get(counterGroup);
+                for (String counter : groupCounters.keySet()) {
+                    Long value = groupCounters.get(counter);
+                    // optionally dump the counters to logs
+                    if (this.dumpCounters) {
+                        log.info("COUNTER " + counterGroup + " " + counter + " " + value);
+                    } else {
+                        context.getCounter(counterGroup, counter).increment(value);
```
jschmidt10, line 264: The comment suggests that the dump counter logs in addition to incrementing but here we log OR increment. I like always incrementing.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexMapper.java
```diff
@@ -0,0 +1,764 @@
...
+        // 3. Get the data type. The data type is used to get the correct IngestHelperInterface which is used to make decisions regarding how a field is
+        // indexed. This may vary from data type to data type
+
+        // this is required for the partitioner, see EventMapper
+        MultiTableRangePartitioner.setContext(context);
+
+        // ensure the key is an fi
+        final byte[] cf = key.getColumnFamilyData().getBackingArray();
+        final int fiBaseLength = FI_START_BYTES.length;
+        if (cf.length <= fiBaseLength || WritableComparator.compareBytes(cf, 0, fiBaseLength, FI_START_BYTES, 0, fiBaseLength) != 0) {
```
jschmidt10: This conditional tree is sorta tricky. It reads "if this isn't an FI, then check for tf, then check for d, if its not a tf or a d, it must be an event, else its an fi"

Wondering if we could bubble all the key checks to a single level and then maybe follow your `processFI` pattern for each of the other key types:

```
if (isFiCol(key))
  processFiCol(key, context)
else if (isTfCol(key))
  processTfCol(key, context)
else if (isDCol(key))
  processDCol(key, context)
else
  processEventKey(key, context)
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexMapper.java
```diff
@@ -0,0 +1,764 @@
...
+
+    // TODO add javadoc
+    private void processFI(Context context, Key key) throws IOException, InterruptedException {
+        final byte[] cf = key.getColumnFamilyData().getBackingArray();
+
+        // check if it's the same target field as the last one
+        final int fiBaseLength = FI_START_BYTES.length;
+        final int fiBaseOffset = fiBaseLength + 1;
+        // TODO add comment here
+        if (this.lastFiBytes == null || WritableComparator.compareBytes(cf, fiBaseOffset, cf.length - fiBaseOffset, this.lastFiBytes, fiBaseOffset,
```
jschmidt10, line 417: Rather than commenting, you could turn this into a method call:

`if (isNewFiField(cf))`

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexJob.java
FineAndDandy: this comment was still helpful. I realized I missed setting the consistency level on the AccumuloInputFormat when it was in use.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexMapper.java
FineAndDandy: Sure. The beauty of constant is it doesn't matter

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexMapper.java
FineAndDandy: Updated the comment. This feature was required to support cases where the counters would overrun the MR limits and cause jobs to fail.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexMapper.java
FineAndDandy: I will do this, but will need to add some new state to prevent reparsing

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexJob.java
```diff
@@ -197,10 +268,61 @@ private Job setupJob() throws IOException, ParseException, AccumuloException, Ta
...
 
-    public static Collection<Range> buildRanges(String start, String end, int splitsPerDay) throws ParseException {
+    private void validateTablesExist(String[] tableNames) throws AccumuloException {
+        for (String table : tableNames) {
+            try {
+                Map<String,String> tableProperties = accumuloClient.tableOperations().getTableProperties(table);
+                if (tableProperties == null) {
+                    throw new IllegalArgumentException("configured output table: " + table + " does not exist");
+                }
+            } catch (TableNotFoundException tnfe) {
```
MiguelRicardos, line 382: Is there a better way than catching one exception just to throw another?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexMapper.java
```diff
@@ -0,0 +1,837 @@
...
+     * @throws InterruptedException
+     */
+    @Override
+    protected void map(Key key, Value value, Context context) throws IOException, InterruptedException {
+        // this is required for the partitioner, see EventMapper
+        MultiTableRangePartitioner.setContext(context);
+
+        ByteSequence cf = key.getColumnFamilyData();
+        String keyType;
+        if (isKeyD(cf)) {
```
MiguelRicardos: I wonder if it would be valuable to have a getKeyType(cf) that returns a string to assign to keyType and then do switch statements on this like:
```
switch (keyType) {
    case "d":
        processDKey(key, value, context);
        break;
    case "tf":
        processTFKey(key, value, context);
        break;
    case "fi":
        processFIKey(key, value, context);
        break;
    default:
        processEventKey(key, value, context);
        break;
}
```
or even better if the case of the d, tf, and fi keys are just an integer we can do the case statement on the length of the bytes in the cf:

```
int keyTypeCode = cf.length()

switch (keyTypeCode) {
    case 1:
        processDKey(key, value, context);
        break;
    case 2:
        processTFKey(key, value, context);
        break;
    case 4:
        //possible extra check on starting bytes
        processFIKey(key, value, context);
        break;
    default:
        processEventKey(key, value, context);
        break;
```

It could eliminate the need for the isKey* methods and might be more efficient.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexMapper.java
```diff
@@ -0,0 +1,837 @@
...
+
+        incrementCounter("fi.dataTypes", dataType);
+        incrementCounter("fi.fields", this.normalizedFieldName);
+
+        Text fieldValueText;
+        Text fieldText = null;
+        Text indexCq = null;
+        boolean indexed = false;
+
+        if (key.isDeleted() && !this.propagateDeletes) {
```
MiguelRicardos, line 461: Consider changing to:
```
if (key.isDeleted()) {
    if (!this.propagateDeletes) {
        incrementCounter("deletes", "skipped");
        return;
    }
    incrementCounter("deletes", "propagated");
}
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexMapper.java
```diff
@@ -0,0 +1,837 @@
...
+            incrementCounter("deletes", "skipped");
+            return;
+        } else if (key.isDeleted()) {
+            incrementCounter("deletes", "propagated");
+        }
+
+        // if the field is indexed and index only or events aren't being reprocessed
+        if (helper.isIndexedField(this.normalizedFieldName) && (!this.reprocessEvents || helper.isIndexOnlyField(this.normalizedFieldName))) {
```
MiguelRicardos, line 469: You might want to move:
```
(!this.reprocessEvents || helper.isIndexOnlyField(this.normalizedFieldName))
```

to a single if statment tha encompasses the ```(helper.isIndexedField(this.normalizedFieldName) ``` and ```(helper.isReverseIndexedField(this.normalizedFieldName)``` checks.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexMapper.java
```diff
@@ -0,0 +1,837 @@
...
+        }
+    }
+
+    /**
+     * Optionally floor a timestamp to the beginning of the day
+     *
+     * @param timestamp
+     * @return the original timestamp if this.floorTimestamps is false, otherwise the timestamp set to the same day at 0:0:0.000
+     */
+    private long floorTimestamp(long timestamp) {
```
MiguelRicardos, line 528: Might want to use java.time instead of Calendar.

```
private long floorTimestamp(long timestamp) {
    if (!this.floorTimestamps) {
        return timestamp;
    }

    return java.time.Instant.ofEpochMilli(timestamp)
            .atZone(java.time.ZoneId.systemDefault())  // Adjust the zone as necessary
            .toLocalDate()
            .atStartOfDay(java.time.ZoneId.systemDefault())
            .toInstant()
            .toEpochMilli();
}
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexMapper.java
jschmidt10: Got it. Thanks!

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexJob.java
jschmidt10: I am okay with this pattern since we're converting a checked exception into an unchecked one. This is us saying that if the table fails, there is no reasonable recovery. The conversion to an unchecked means you don't have to string `throws TableNotFoundException` up your entire call stack.

One thing I would do is add the originating exception to the causedBy of the one we're throwing for stacktrace traceability.

jschmidt10 (dismissed):

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexJob.java
MiguelRicardos: sounds reasonable to me.

apmoriarty (approved):

drewfarris:
Would it make sense to squash and rebase at this point?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/RepartitionerJob.java
drewfarris, line 1: Consider setting `mapreduce.job.name` (unless you're setting it somewhere else already...

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reindex/ShardReindexJob.java
```diff
@@ -0,0 +1,846 @@
...
+        private String resources;
+
+        @Parameter(names = "--reducers", description = "number of reducers to use", required = true)
+        private int reducers;
+
+        @Parameter(names = "--outputDir", description = "output directory that must not already exist", required = true)
+        private String outputDir;
+
+        @Parameter(names = "--destHdfs", description = "HDFS for --outputDir", required = true)
+        private String destHdfs;
```
drewfarris, line 738: In theory a separate `destHdfs` shouldn't be needed if `outputDir` is a fully qualified path like `hdfs://ingest/foo/bar` and you use something like `path.getFilesystem(conf)` when you need to get the `FileSystem` object.

drewfarris (approved):
Monster PR. Looks good to me - a couple minor nits, possibly not worth addressing.

PR #2354, opened 2024-04-19
Fixed bug where unique results of null values failed to return

ivakegg:
* The intermedite results mechanism which got fixed with the most recent unique mechanism resulted in this bug


warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -132,23 +103,6 @@ public void updateConfig(UniqueFields uniqueFields, QueryModel model) {
...
-    }
-
-    /**
-     * Set the query model from which the reverse query mappings are pulled.
-     *
-     * @param model
-     *            The query model
-     */
-    private void setModelMappings(QueryModel model) {
-        if (model != null) {
```
ivakegg, line 145: This whole mapping is already done in the DefaultQueryPlanner when setting up the query fields.  This was redundant.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -312,18 +265,22 @@ private void outputSortedFieldValues(Document document, DataOutputStream output)
      * @throws IOException
      *             for issues with read/write
      */
-    private int dumpValues(int count, String field, List<String> values, DataOutputStream output) throws IOException {
+    private void dumpValues(String field, Collection<String> values, DataOutputStream output) throws IOException {
+        String separator = "f-" + field + ":";
         if (!values.isEmpty()) {
-            Collections.sort(values);
-            String separator = "f-" + field + '/' + (count++) + ":";
```
ivakegg, line 318: Adding the count in here is not really useful.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -177,6 +131,10 @@ public Entry<Key,Document> apply(@Nullable Entry<Key,Document> keyDocumentEntry)
                 return keyDocumentEntry;
             }
 
+            if (keyDocumentEntry.getValue().isIntermediateResult()) {
+                return keyDocumentEntry;
+            }
+
```
ivakegg, line 137: This was the critical piece to fix the null document issue.

warehouse/query-core/src/test/java/datawave/query/transformer/UniqueTransformTest.java
hlgp, line 1: How tough would it be to add a test that would fail without these fixes?

ivakegg: Pretty tough, but I will think on it.

ivakegg: ok, tests added

apmoriarty (approved):

avgAGB (approved):

PR #2362, opened 2024-04-30
Allow table cache to be bypassed by ModelBean (#2359)

### LINKED ISSUE #2359: ModelBean should optionally allow table cache to be bypassed via WrappedAccumuloClient

keith-ratcliffe:
Depends on https://github.com/NationalSecurityAgency/datawave-accumulo-utils/pull/14

The goal here is to allow certain ModelBean operations to be performed in a more timely manner. E.g., in order to replace a given query model, the current implementation requires that a user first perform a model DELETE operation and then wait for an asynchronous cache refresh, so that the subsequent model import can succeed. If the cache isn't fully refreshed, the import fails with an error, because the deleted model is still resident in the stale cache. That process can be drastically streamlined by simply allowing the table cache to be bypassed by an "admin" user

This PR also fixes a bug in the `listModelNames` method that misidentifies row: `edge_key` colf: `version` as a query model

### DIFF
--- a/web-services/model/src/main/java/datawave/webservice/query/model/ModelBean.java
+++ b/web-services/model/src/main/java/datawave/webservice/query/model/ModelBean.java
@@ -23,9 +23,11 @@ import javax.inject.Inject;
 import javax.interceptor.Interceptors;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.DELETE;
+import javax.ws.rs.DefaultValue;
 import javax.ws.rs.FormParam;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
+import javax.ws.rs.PUT;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
@@ -37,6 +39,7 @@ import org.apache.accumulo.core.client.BatchWriterConfig;
 import org.apache.accumulo.core.client.IteratorSetting;
 import org.apache.accumulo.core.client.MutationsRejectedException;
 import org.apache.accumulo.core.client.Scanner;
+import org.apache.accumulo.core.client.TableNotFoundException;
 import org.apache.accumulo.core.data.Key;
 import org.apache.accumulo.core.data.Mutation;
 import org.apache.accumulo.core.data.Value;
@@ -58,6 +61,8 @@ import datawave.security.authorization.DatawavePrincipal;
 import datawave.security.util.ScannerHelper;
 import datawave.webservice.common.cache.AccumuloTableCache;
 import datawave.webservice.common.connection.AccumuloConnectionFactory;
+import datawave.webservice.common.connection.WrappedAccumuloClient;
+import datawave.webservice.common.connection.WrappedScannerHelper;
 import datawave.webservice.common.exception.DatawaveWebApplicationException;
 import datawave.webservice.common.exception.NotFoundException;
 import datawave.webservice.common.exception.PreConditionFailedException;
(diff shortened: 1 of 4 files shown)


web-services/model/src/main/java/datawave/webservice/query/model/ModelBean.java
```diff
@@ -86,7 +90,7 @@ public class ModelBean {
     private static final long BATCH_WRITER_MAX_MEMORY = 10845760;
     private static final int BATCH_WRITER_MAX_THREADS = 2;
 
-    private static final HashSet<String> RESERVED_COLF_VALUES = Sets.newHashSet("e", "i", "ri", "f", "tf", "m", "desc", "edge", "t", "n", "h");
+    private static final HashSet<String> RESERVED_COLF_VALUES = Sets.newHashSet("e", "i", "ri", "f", "tf", "m", "desc", "edge", "t", "n", "h", "version");
```
keith-ratcliffe, line 94: This fixes a preexisting bug: ```listModelNames``` method erroneously returns "version" (i.e., colf of the "edge_key" row) as an existing model

web-services/model/src/main/java/datawave/webservice/query/model/ModelBean.java
```diff
@@ -127,6 +131,32 @@ public class ModelBean {
     @GZIP
     @Interceptors(ResponseInterceptor.class)
     public ModelList listModelNames(@QueryParam("modelTableName") String modelTableName) {
+        return listModelNames(modelTableName, false);
```
ivakegg, line 135: If the default for skipCache is true down below, should this not be passing in true?  Moreover do we even need this method since the query parameter in the method below is basically optional?

web-services/model/src/main/java/datawave/webservice/query/model/ModelBean.java
```diff
@@ -205,7 +237,8 @@ else if (parts.length == 2)
     @GZIP
     @RolesAllowed({"Administrator", "JBossAdministrator"})
     @Interceptors(ResponseInterceptor.class)
-    public VoidResponse importModel(datawave.webservice.model.Model model, @QueryParam("modelTableName") String modelTableName) {
+    public VoidResponse importModel(datawave.webservice.model.Model model, @QueryParam("modelTableName") String modelTableName,
```
ivakegg: As an additional optimization, can the importModel do the delete as well instead of returning a 412 response?  I always found that annoying personally.

web-services/model/src/main/java/datawave/webservice/query/model/ModelBean.java
keith-ratcliffe: Since this endpoint is open to everyone, my assumption was that we wouldn't want a typical user to be able to circumvent the webservice's cache layer. So, the default of `true` below is only in effect for the admin-only override, just as a convenience. Same reasoning was applied to the `getModel` endpoint as well, since it's also open to everyone.

But if you think it's ok to relax that restriction in this case, then that would definitely simplify things. I could go either way here, honestly

web-services/model/src/main/java/datawave/webservice/query/model/ModelBean.java
keith-ratcliffe: Lol! I had an earlier iteration with that exact change. But then I chickened out, because it seemingly violates traditional `POST` method semantics, and it felt like the original author(s) had those strict semantics in mind here.

But I agree with the annoyance factor here, and I could really go either way. Just let me know.

FWIW, from my own testing, I can say that a lot of the annoyance goes away when you don't have to wait minute(s) for an asynchronous cache refresh just to make it through to the finish line :-)

keith-ratcliffe: In latest commit, I updated the `importModel` method to support an optional path parameter...
```
@Path("/import{pathDelimiter: (/)?}{modelName: ((?<=/)[a-zA-Z0-9_]*)?}")
```
...in order to enable update/replace semantics. So, user can now accomplish the delete and import with a single call. E.g.,..
```
curl -d @/path/to/MyExistingModel.xml -X PUT http://localhost:8443/DataWave/Model/import/MyExistingModelName
```
If you try to post an existing model without specifying the `/{modelName}` path param, the assumption is that you're intending to create a new model, and you'll get the `412` error response, same as always

apmoriarty (approved):

hlgp:
Closing, deferred for now

PR #2382, opened 2024-05-03
Adding term catch to load job cache script

avgAGB:


warehouse/ingest-scripts/src/main/resources/bin/ingest/load-job-cache.sh
```diff
@@ -42,9 +42,22 @@ sed s%${BEFORE}%${AFTER}% job-cache-env.sh > job-cache-env.tmp
...
 
+function no_ctrlc() {
+  let count++
+  echo
+  if [[ $count == 1 ]]; then
+    echo "Are you sure you want to stop the load-job-cache.sh process? Giving you a warning."
+  else
+    echo "Second SIGTERM received, dipping out"
+    rm -r -f "$tmpdir"
+    exit $?
```
jschmidt10: This is following the pattern that was established before you but this doesn't look right to me. I think if we Ctrl+C this script, we wouldn't want to return a successful exit code (0). But this is returning the exit code of the `rm` command above it, so if the `rm` works, then we return success.

Thoughts?

warehouse/ingest-scripts/src/main/resources/bin/ingest/load-job-cache.sh
```diff
@@ -42,9 +42,22 @@ sed s%${BEFORE}%${AFTER}% job-cache-env.sh > job-cache-env.tmp
 
 date
 
+function no_ctrlc() {
+  let count++
+  echo
+  if [[ $count == 1 ]]; then
+    echo "Are you sure you want to stop the load-job-cache.sh process? Giving you a warning."
```
jschmidt10: We should add some verbiage that describes what will occur because otherwise, you might not necessarily know the scope of the warning.

Maybe something like "Killing this script will leave the job cache in a most-likely unusable state. You will need to rerun this script."

warehouse/ingest-scripts/src/main/resources/bin/ingest/load-job-cache.sh
avgAGB: Hm, maybe instead we exit with a 1 here instead? I think that might be prudent.

jschmidt10: Works for me. Can you also add it to the KILL INT trap too?

warehouse/ingest-scripts/src/main/resources/bin/ingest/load-job-cache.sh
jschmidt10: Oh, and then maybe clarify that you have to run `Ctrl+C` again to terminate.

warehouse/ingest-scripts/src/main/resources/bin/ingest/load-job-cache.sh
avgAGB: Sure, will do.

jschmidt10 (dismissed):

warehouse/ingest-scripts/src/main/resources/bin/ingest/load-job-cache.sh
```diff
@@ -42,9 +42,23 @@ sed s%${BEFORE}%${AFTER}% job-cache-env.sh > job-cache-env.tmp
...
+    echo "Second SIGTERM received, exiting script."
+    rm -r -f "$tmpdir"
+    exit 1
+  fi
+}
+
 # prepare a directory with links to all of the files/directories to put into the jobcache
 tmpdir=`mktemp -d $MKTEMP_OPTS`
-trap 'rm -r -f "$tmpdir"; exit $?' INT TERM EXIT
+trap 'rm -r -f "$tmpdir"; exit 1' INT EXIT
```
keith-ratcliffe: Doesn't 
```
trap 'blah ; exit 1' EXIT
```
guarantee that the script will _always_ exit w/ non-zero status?

avgAGB: going off a discussion with Jeff it looks like if TERM/KILL signals are being sent while this script is running it probably makes sense to return with a non-zero status to imply the script did not exit clean. If the trap isn't encountered though, we would want to exit with a 0 in that case.

keith-ratcliffe: Trap arg for `EXIT` will fire even on normal exit of the script. So, I don't see how we'll ever get a zero exit status

E.g., if you run the script below, you'll get non-zero return status:
```
#!/bin/bash
trap 'echo trapping; exit 1' INT EXIT
exit 0
```

avgAGB: Ah ok, now I follow. I'll see if I can extract out the EXIT signal from that logic

keith-ratcliffe (dismissed):

keith-ratcliffe (approved):

alerman (approved):
LGTM

PR #2388, opened 2024-05-14
Fix #GROUPBY duplicate counts with query model

### LINKED ISSUE #2360: #GROUPBY and #COUNT is creating duplicate counts
It appears that when we have a groupby on a model field that maps to multiple fields that actually exist in the same record with the same value, then the counts being returned will be double the actual count of records.

lbschanno:
When query model mapping is applied to a document before grouping and aggregating, it is possible for multiple fields to be mapped to the same root model mapping with the same value, resulting in duplicate field-value pairings that were incorrectly included in grouping and aggregation.

Ensure that duplicate field-value pairings are disregarded. Add tests to verify this and update documentation to include details about how model mapping affects the #GROUPBY function.

Fixes #2360


warehouse/query-core/src/test/java/datawave/query/transformer/GroupingTest.java
```diff
@@ -1004,4 +1015,46 @@ public void testAveragingNonNumericalValue() {
...
+
+        // Run the test queries and collect their results.
+        collectQueryResults();
+
+        // Verify the results.
+        assertGroups();
+    }
+
+    /**
+     * Verify that when grouping and aggregating with model mapping, if a multiple fields with the same value are mapped to the same root model mapping in a
```
foster33:
```suggestion
     * Verify that when grouping and aggregating with model mapping, if multiple fields with the same value are mapped to the same root model mapping in a
```

apmoriarty (approved):

ivakegg (approved):

PR #2390, opened 2024-05-15
fix critical sonarqube issues

cogross:
fix up the "critical" issues found by sonarqube.
bugs and some code smells

### DIFF
--- a/warehouse/age-off-utils/src/main/java/datawave/age/off/util/AgeOffRuleFormatter.java
+++ b/warehouse/age-off-utils/src/main/java/datawave/age/off/util/AgeOffRuleFormatter.java
@@ -3,6 +3,7 @@ package datawave.age.off.util;
 import java.io.IOException;
 import java.io.StringWriter;
 import java.io.Writer;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import javax.xml.transform.OutputKeys;
 import javax.xml.transform.Transformer;
@@ -50,14 +51,14 @@ public class AgeOffRuleFormatter {
         writer.write(transformToXmlString(ruleConfig));
     }
 
-    private AgeOffRuleLoader.RuleConfig createRuleConfig(AgeOffRuleConfiguration configuration) throws IOException {
-        AgeOffRuleLoader.RuleConfig ruleConfig = new AgeOffRuleLoader.RuleConfig(this.configuration.getFilterClass().getName(), index++);
+    private static AgeOffRuleLoader.RuleConfig createRuleConfig(AgeOffRuleConfiguration configuration) throws IOException {
+        AgeOffRuleLoader.RuleConfig ruleConfig = new AgeOffRuleLoader.RuleConfig(configuration.getFilterClass().getName(), index++);
         ruleConfig.label(configuration.getRuleLabel());
-        ruleConfig.setIsMerge(this.configuration.shouldMerge());
-        ruleConfig.ttlValue(this.configuration.getTtlDuration());
-        ruleConfig.ttlUnits(this.configuration.getTtlUnits());
-        ruleConfig.matchPattern(buildMatchPattern());
-        ruleConfig.customElements(this.configuration.getCustomElements());
+        ruleConfig.setIsMerge(configuration.shouldMerge());
+        ruleConfig.ttlValue(configuration.getTtlDuration());
+        ruleConfig.ttlUnits(configuration.getTtlUnits());
+        ruleConfig.matchPattern(buildMatchPattern(configuration, configuration.getIndentation()));
+        ruleConfig.customElements(configuration.getCustomElements());
         return ruleConfig;
     }
 
@@ -93,7 +94,7 @@ public class AgeOffRuleFormatter {
         return Integer.toString(length);
     }
 
-    private String buildMatchPattern() throws IOException {
+    private static String buildMatchPattern(AgeOffRuleConfiguration configuration, String indent) throws IOException {
         if (configuration.getPatternConfiguration() == null) {
             return "";
         }
@@ -104,14 +105,14 @@ public class AgeOffRuleFormatter {
         AgeOffCsvToMatchPatternFormatter patternFormatter = new AgeOffCsvToMatchPatternFormatter(configuration.getPatternConfiguration());
 
         // add two indentations: one for items under the rule element and another for items under the matchPattern element
-        String extraIndentation = this.indent + this.indent;
+        String extraIndentation = indent + indent;
         patternFormatter.write(new IndentingDelegatingWriter(extraIndentation, writer));
 
         String result = writer.toString();
 
         // final indentation to precede the closing of matchPattern
         if (result.endsWith("\n")) {
-            return result + this.indent;
+            return result + indent;
         }
         return result;
     }
--- a/warehouse/age-off-utils/src/main/java/datawave/age/off/util/AgeOffRuleFormatter.java
+++ b/warehouse/age-off-utils/src/main/java/datawave/age/off/util/AgeOffRuleFormatter.java
@@ -3,7 +3,6 @@ package datawave.age.off.util;
 import java.io.IOException;
 import java.io.StringWriter;
 import java.io.Writer;
-import java.util.concurrent.atomic.AtomicInteger;
 
 import javax.xml.transform.OutputKeys;
 import javax.xml.transform.Transformer;
--- a/warehouse/core/src/main/java/datawave/edge/protobuf/EdgeData.java
+++ b/warehouse/core/src/main/java/datawave/edge/protobuf/EdgeData.java
@@ -328,12 +328,6 @@ public final class EdgeData {
                         case 0:
                             done = true;
                             break;
-                        default: {
-                            if (!parseUnknownField(input, unknownFields, extensionRegistry, tag)) {
-                                done = true;
-                            }
-                            break;
-                        }
                         case 8: {
                             bitField0_ |= 0x00000001;
                             count_ = input.readInt64();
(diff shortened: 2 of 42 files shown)


warehouse/age-off-utils/src/main/java/datawave/age/off/util/AgeOffRuleFormatter.java
```diff
@@ -51,8 +51,13 @@ void format(Writer writer) throws IOException {
         writer.close();
     }
 
+    private static void increment() {
+        index++;
+    }
+
     private AgeOffRuleLoader.RuleConfig createRuleConfig(AgeOffRuleConfiguration configuration) throws IOException {
-        AgeOffRuleLoader.RuleConfig ruleConfig = new AgeOffRuleLoader.RuleConfig(this.configuration.getFilterClass().getName(), index++);
+        AgeOffRuleLoader.RuleConfig ruleConfig = new AgeOffRuleLoader.RuleConfig(this.configuration.getFilterClass().getName(), index);
```
jschmidt10: I know this was sorta the quick fix to get thru the sonarqube complaint but I almost feel like we're hiding the problem.

Can we make `createRuleConfig()` static instead? It has some instance references to `configuration` but we actually pass `configuration` in already - so I think we can make it just use the passed in one.

@matthpeterson Does that sound reasonable?

jschmidt10: Looks like we'd need to pass indent in to `createRuleConfig()` and then to `buildMatchPattern()` as well.

warehouse/core/src/main/java/datawave/edge/protobuf/EdgeData.java
```diff
@@ -308,7 +308,7 @@ public EdgeValue getDefaultInstanceForType() {
             return defaultInstance;
         }
 
-        private final com.google.protobuf.UnknownFieldSet unknownFields;
+        private transient final com.google.protobuf.UnknownFieldSet unknownFields;
```
jschmidt10: The comment at the top says this was generated by protoc. It says to not edit, I do not know the impact if we do.

jschmidt10: This might be something where we're supposed to be generating this at build time instead of generating it once and adding it to source control by hand.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
```diff
@@ -796,7 +800,7 @@ protected Set<NormalizedContentInterface> normalize(NormalizedContentInterface n
                     values.add(normalize(normalizedContent, dataType));
                 }
                 if (log.isDebugEnabled()) {
-                    log.debug("added normalized field " + normalizedContent + " to values " + values);
+                    log.debug(ADDED_NORMALIZED_FIELD + normalizedContent + TO_VALUES + values);
```
jschmidt10: This `TO_VALUES` constant got me thinking that there's probably a cleaner way to address these. For one, this message is saying the same thing as the one above without the `TO_VALUES`. The only difference is we're logging the whole values set over and over here so I might go with the former.

Another thing we could do is add a method for `logNormalizedField(String value)` and that method could check the log level and log the message. Then we wouldn't have this block repeated over and over.

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/job/BulkIngestMapFileLoaderTest.java
```diff
@@ -1625,8 +1625,11 @@ public void testCleanUpJobDirectoryMakesDirectory() throws Exception {
             BulkIngestMapFileLoaderTest.WrappedLocalFileSystem fs = new BulkIngestMapFileLoaderTest.WrappedLocalFileSystem(createMockInputStream(), null, true,
                             false, false, false, new HashMap<>(), false, false);
 
-            Whitebox.invokeMethod(FileSystem.class, "addFileSystemForTesting", BulkIngestMapFileLoaderTest.FILE_SYSTEM_URI, conf, fs);
-
+            try {
+                Whitebox.invokeMethod(FileSystem.class, "addFileSystemForTesting", BulkIngestMapFileLoaderTest.FILE_SYSTEM_URI, conf, fs);
+            } catch (Exception e) {
```
MiguelRicardos, line 1632: Can you have it catch the specific type of exception it fails on for each of these catch statements?

warehouse/core/src/main/java/datawave/mr/bulk/BulkInputFormat.java
```diff
@@ -110,6 +110,8 @@ public class BulkInputFormat extends InputFormat<Key,Value> {
     protected static final String RANGESPLITSTRATEGY = PREFIX + ".split.strategy.class";
     protected static final String MOCK = ".useInMemoryInstance";
 
+    protected static final String UTF8 = "UTF-8";
```
MiguelRicardos, line 119: I'm not sure this is a critical vulnerability, but I'll give it to you.

warehouse/core/src/main/java/datawave/ingest/protobuf/TermWeight.java
```diff
@@ -125,7 +125,7 @@ public Info getDefaultInstanceForType() {
             return defaultInstance;
         }
 
-        private final com.google.protobuf.UnknownFieldSet unknownFields;
+        private final transient com.google.protobuf.UnknownFieldSet unknownFields;
```
MiguelRicardos: Won't this change get reverted when it is regenerated?

warehouse/core/src/main/java/datawave/ingest/protobuf/Uid.java
```diff
@@ -135,7 +135,7 @@ public List getDefaultInstanceForType() {
             return defaultInstance;
         }
 
-        private final com.google.protobuf.UnknownFieldSet unknownFields;
+        private final transient com.google.protobuf.UnknownFieldSet unknownFields;
```
MiguelRicardos: Same for this, won't this change get reverted when re-generated?

cogross: reverted.

warehouse/core/src/main/java/datawave/ingest/protobuf/TermWeight.java
cogross: reverted

warehouse/core/src/main/java/datawave/mr/bulk/BulkInputFormat.java
cogross: Agree, but its the reused string that it thinks could be done wrong.  Just trying to reduce the grand number of this its finding.

warehouse/core/src/main/java/datawave/edge/protobuf/EdgeData.java
cogross: missed that -- intellij collapsed the comment at the top.  reverted.

warehouse/age-off-utils/src/main/java/datawave/age/off/util/AgeOffRuleFormatter.java
cogross: reworked it as you described.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
cogross: Went for option 2. i think that seems cleaner and we can remove the 2 static strings.

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/job/BulkIngestMapFileLoaderTest.java
cogross: all the other spots just catch an Exception (invokeMethod throws Exception) - but addFileSystemForTesting throws IOException so I added that.

avgAGB (dismissed):

warehouse/core/src/main/java/datawave/edge/util/EdgeKey.java
```diff
@@ -44,6 +44,8 @@ public class EdgeKey {
 
     private static final Logger log = Logger.getLogger(EdgeKey.class);
 
+    private static final String CANT_ENCODE_UNKNOWN_ERROR_MSG = "Can't encode unknown edge key format.";
```
apmoriarty: `UNKNOWN_EDGE_KEY_FORMAT_MSG` seems more precise

warehouse/core/src/main/java/datawave/mr/bulk/BulkInputFormat.java
```diff
@@ -1289,8 +1291,8 @@ public AccumuloIteratorOption(String iteratorOption) {
             StringTokenizer tokenizer = new StringTokenizer(iteratorOption, FIELD_SEP);
             this.iteratorName = tokenizer.nextToken();
             try {
-                this.key = URLDecoder.decode(tokenizer.nextToken(), "UTF-8");
-                this.value = URLDecoder.decode(tokenizer.nextToken(), "UTF-8");
+                this.key = URLDecoder.decode(tokenizer.nextToken(), UTF8);
```
apmoriarty, line 1300: Why not `Charset.UTF_8`?

apmoriarty (dismissed):

warehouse/core/src/main/java/datawave/mr/bulk/BulkInputFormat.java
cogross: This is just replacing a String of "UTF-8", not a character set or anything fancy.

warehouse/core/src/main/java/datawave/edge/util/EdgeKey.java
cogross: Sure, changed the name.

alerman (approved):

jalphonso (approved):

PR #2398, opened 2024-05-21
Age Off File creation improvements (#2395)

matthpeterson:
* Stop closing writer after each rule

* Remove ismerge child element in generic case.  Just use mode='merge' instead

* Add header to xml files

---------


awrigh11 (approved):

matthpeterson:
Howard will be cherry-picking this change into the 6.14 release branch and integration

PR #2399, opened 2024-06-04
Ensure CompositeQueryLogic error codes are sent to metrics

### LINKED ISSUE #2340: Where O where did my Error Code go
We are no longer placing error codes in our metrics when using the federated query logic (CompositeQueryLogic).  E.g. if a query "times out" then the error code should be 500-27.

lbschanno:
The query metrics API expects QueryExceptions to be found in the cause of the original exception passed to BaseQueryMetric.setError(Throwable). In the case of CompositeQueryLogic, due to the additional wrapping of exceptions passed from its defined internal logics in a CompositeLogicException, query exceptions were not in the expected place in the stack hierarchy.

Update the CompositeLogicException to identify any QueryExceptions present in the stacks of exception arguments, and ensure that they are either already a QueryException, or that if they contain a QueryException in the stack, that they are subsequently wrapped in a CompositeRaisedQueryException with a copy of the desired error code.

Fixes #2340


lbschanno:
An alternative approach here would be to modify the query metrics service to search beyond the first cause in an exception's stack trace when searching for a QueryException to extract an error code from. If that approach is preferred, I will close this PR and create one for the query metrics service with that change.

ivakegg (approved):

apmoriarty (approved):

PR #2409, opened 2024-06-07

### LINKED ISSUE #2305: QueryIterators/ScanIterators should log query id first

SethSmucker:
- Create QueryLogIterator class that logs a query's ID to the TServer logs

Resolves #2305

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/QueryParameters.java
+++ b/warehouse/query-core/src/main/java/datawave/query/QueryParameters.java
@@ -1,5 +1,7 @@
 package datawave.query;
 
+import datawave.query.iterator.QueryLogIterator;
+
 public class QueryParameters {
 
     /**
@@ -221,4 +223,9 @@ public class QueryParameters {
      * Used to specify model or DB fields that must be treated as strict (cannot be skipped if normalization fails)
      */
     public static final String STRICT_FIELDS = "strict.fields";
+
+    /**
+     * Controls whether a query's ID is logged on the tserver using {@link QueryLogIterator}
+     */
+    public static final String TSERVER_LOGGING_ACTIVE = "tserver.logging.active";
 }
--- a/warehouse/query-core/src/main/java/datawave/query/iterator/QueryLogIterator.java
+++ b/warehouse/query-core/src/main/java/datawave/query/iterator/QueryLogIterator.java
@@ -16,8 +16,11 @@ import org.apache.accumulo.core.iterators.OptionDescriber;
 import org.apache.accumulo.core.iterators.SortedKeyValueIterator;
 import org.apache.log4j.Logger;
 
-
-
+/**
+ * An iterator used to log the start and end of each method run by the {@link SortedKeyValueIterator} above it in the iterator stack.
+ * Logs the QueryID associated with each method.
+ * Logs are written on the TServer where the iterator is running
+ */
 public class QueryLogIterator implements SortedKeyValueIterator<Key,Value>, OptionDescriber {
 
     private static final Logger log = Logger.getLogger(QueryLogIterator.class);
--- a/warehouse/query-core/src/main/java/datawave/query/iterator/QueryLogIterator.java
+++ b/warehouse/query-core/src/main/java/datawave/query/iterator/QueryLogIterator.java
@@ -1,0 +1,145 @@
+package datawave.query.iterator;
+
+import static datawave.query.iterator.QueryOptions.QUERY_ID;
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.apache.accumulo.core.data.ByteSequence;
+import org.apache.accumulo.core.data.Key;
+import org.apache.accumulo.core.data.Range;
+import org.apache.accumulo.core.data.Value;
+import org.apache.accumulo.core.iterators.IteratorEnvironment;
+import org.apache.accumulo.core.iterators.OptionDescriber;
+import org.apache.accumulo.core.iterators.SortedKeyValueIterator;
+import org.apache.log4j.Logger;
+
+public class QueryLogIterator implements SortedKeyValueIterator<Key,Value>, OptionDescriber {
+
+    private static final Logger log = Logger.getLogger(QueryLogIterator.class);
+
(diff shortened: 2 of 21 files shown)


warehouse/query-core/src/main/java/datawave/query/QueryParameters.java
```diff
@@ -1,5 +1,7 @@
 package datawave.query;
 
+import datawave.query.iterator.QueryLogIterator;
```
jschmidt10, line 3: This import is unused.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryLogIterator.java
```diff
@@ -0,0 +1,189 @@
...
+     * Wraps the init() method of the iterator above it, logging the start and end of the method along with its query id.
+     */
+    @Override
+    public void init(SortedKeyValueIterator<Key,Value> source, Map<String,String> options, IteratorEnvironment env) throws IOException {
+
+        try {
+            this.queryID = options.get(QUERY_ID);
+            this.source = source;
+            this.env = env;
+            logStartOf("init()");
```
jschmidt10, line 56: Seems like this should be the first thing called in the function here. Also there is no delegation to the `source.init()`.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
foster33, line 1: There are no actual changes here, so this class can probably be reverted.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryLogIterator.java
```diff
@@ -0,0 +1,189 @@
...
+    private static final String CLASS_NAME = QueryLogIterator.class.getSimpleName();
+
+    private String queryID;
+    private SortedKeyValueIterator<Key,Value> source;
+    private IteratorEnvironment env;
+
+    /**
+     * Class constructor
+     */
+    public QueryLogIterator() {}
```
avgAGB, line 35: Could this be pruned out if it isn't being used?

warehouse/query-core/src/main/java/datawave/query/QueryParameters.java
```diff
@@ -221,4 +223,9 @@ public class QueryParameters {
      * Used to specify model or DB fields that must be treated as strict (cannot be skipped if normalization fails)
      */
     public static final String STRICT_FIELDS = "strict.fields";
+
+    /**
+     * Controls whether a query's ID is logged on the tserver using {@link QueryLogIterator}
+     */
+    public static final String TSERVER_LOGGING_ACTIVE = "tserver.logging.active";
```
ivakegg, line 230: Please move this constant to the QueryOptions class instead.  QueryParameters are for parameters specified by users when submitting queries.  This is purly used as an option on the query log iterator.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryLogIterator.java
ivakegg: The tserver iterator setup will be calling the source.init with the options specific to that iterator.  No need to call source.init().

warehouse/query-core/src/main/java/datawave/query/QueryParameters.java
ivakegg, line 230: You are setting this as a iterator option as well as pulling it out of the query parameters.  I don't believe the user submitting the query should be able to turn logging off and on.  That should be done when the query logic is being created (see QueryLogicFactory.xml).  You may want to add a line in there.  I think this constant should go away, and the default query planner should optionally setup the QueryLogIterator if that boolean is true in the configuration which it appears you are already doing.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -469,9 +477,21 @@ protected CloseableIterable<QueryData> process(ScannerFactory scannerFactory, Me
...
+                    if (logCfg == null) {
+                        logCfg = getQueryLogIterator(config, settings);
+                    }
+                    break;
+                }
+            }
         }
 
-        final QueryData queryData = new QueryData().withQuery(newQueryString).withSettings(Lists.newArrayList(cfg));
+        List<IteratorSetting> iteratorSettings = logCfg == null ? Collections.singletonList(cfg) : Lists.newArrayList(cfg, logCfg);
```
ivakegg, line 493: We should only be adding the logCfg if config.isTserverLoggingActive().

ivakegg: Optionally avoid creating the logCfg altogether if config.IsTserverLoggingActive() == false;.,

warehouse/query-core/src/test/java/datawave/query/config/ShardQueryConfigurationTest.java
```diff
@@ -580,6 +580,9 @@ public void setUp() throws Exception {
         updatedValues.put("sortQueryBeforeGlobalIndex", true);
         defaultValues.put("sortQueryByCounts", false);
         updatedValues.put("sortQueryByCounts", true);
+
+        defaultValues.put("tserverLoggingActive", false);
```
ivakegg, line 584: I think the default should be true

warehouse/query-core/src/test/java/datawave/query/tables/ShardQueryLogicTest.java
```diff
@@ -360,4 +360,9 @@ public void testNegativeFilterRegexV2() throws Exception {
 
         runTestQuery(expected, queryString, format.parse("20091231"), format.parse("20150101"), extraParameters);
     }
+
+    @Test
+    public void testTserverLoggingActive() {
+
```
ivakegg, line 366: This is not much of a test....

ivakegg (changes requested):
See comments

SethSmucker:
Superseded by #2428, all changes suggested have been added there. That PR pulls directly from my fork instead of using a middle-man branch between my fork's branch and upstream/integration.

PR #2419, opened 2024-06-14
Add defaulting of non-"false" QueryOption fields to "true"

### LINKED ISSUE #2410: Tighter query options
In the QueryOptions.validateOptions we need to be able to set an option to an appropriate default value if the option exists but has no value.

For example looking at the REDUCED_RESPONSE option, it uses setReducedResponse(Boolean.parseBoolean(options.get(REDUCED_RESPONSE)));
The problem is we want the mere existance of the option (i.e. empty value) to also results in setting it to true.
This should be need this done for all boolean properties setup in that function.

Then we can update the DefaultQueryPlanner to set the options but with empty values.

The net results is that the amount of data sent from the webserver to the query iterator will be tighter and more readable when listing scans

SethSmucker:
- Create function evaluateBooleanOption() method in QueryOptions
- Use evaluateBooleanOption() in the validateOptions() method
- Add a test in QueryOptionsTest to test evaluateBooleanOption()

resolves #2410

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
+++ b/warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
@@ -1289,6 +1289,20 @@ public class QueryOptions implements OptionDescriber {
         return new IteratorOptions(getClass().getSimpleName(), "Runs a query against the DATAWAVE tables", options, null);
     }
 
+    /**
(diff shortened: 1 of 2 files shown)


foster33:
This PR seems to be identical to another one you opened up right after this. This one can be closed.

PR #2420, opened 2024-06-14
Optimize QueryOptions to only include entries for non-default value options

### LINKED ISSUE #2410: Tighter query options

SethSmucker:
- Create function evaluateBooleanOption() method in QueryOptions
- Use evaluateBooleanOption() in the validateOptions() method
- Add a test in QueryOptionsTest to test evaluateBooleanOption()

resolves #2410


warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -1289,6 +1289,20 @@ public IteratorOptions describeOptions() {
         return new IteratorOptions(getClass().getSimpleName(), "Runs a query against the DATAWAVE tables", options, null);
     }
 
+    /**
+     * Helper function for validateOptions(). Takes the String value of an option and returns the appropriate a boolean value.
```
foster33:
```suggestion
     * Helper function for validateOptions(). Takes the String value of an option and returns the appropriate boolean value.
```

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -1289,6 +1289,20 @@ public IteratorOptions describeOptions() {
...
+     *
+     * @param value:
+     *            The string value of an option that should be represented by a boolean.
+     * @return: True if the option is null or not "false". False if the option has the value "false" (not case-sensitive).
```
lbschanno:
```suggestion
     * @return True if the option is null or not "false". False if the option has the value "false" (not case-sensitive).
```

lbschanno (changes requested):

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -1289,6 +1289,20 @@ public IteratorOptions describeOptions() {
...
 
+    /**
+     * Helper function for validateOptions(). Takes the String value of an option and returns the appropriate boolean value.
+     *
+     * @param value:
+     *            The string value of an option that should be represented by a boolean.
+     * @return True if the option is null or not "false". False if the option has the value "false" (not case-sensitive).
+     */
+    private static boolean evaluateBooleanOption(String value) {
+        if (value == null) {
```
ivakegg: OK, this is great.  Now we need to add the second part of the ticket which is to update the DefaultQueryPlanner which is setting up these options to use null values instead of "true" when setting up these options on the iterator.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -20,6 +20,7 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
```
foster33: Unused import

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -2278,4 +2384,219 @@ public Equality getEquality() {
...
+            addOption(setting, option, value, (v) -> Short.toString((Short) v), allowBlankValues);
+        } else if (value instanceof Byte) {
+            addOption(setting, option, value, (v) -> Byte.toString((Byte) v), allowBlankValues);
+        } else if (value instanceof Character) {
+            addOption(setting, option, value, (v) -> Character.toString((Character) v), allowBlankValues);
+        } else if (value instanceof Collection<?>) {
+            addOption(setting, option, value, (v) -> COMMA_JOINER.join((Iterable<?>) v), allowBlankValues);
+        } else if (value instanceof Enum<?>) {
+            addOption(setting, option, value, (v) -> ((Enum<?>) v).name(), allowBlankValues);
+        } else if (value instanceof Object){
```
lbschanno: This can be simplified to the following. We do not need to check if Value is an Object if we've reached the last if statement. Additionally, we should default to using v.toString() if we did not match any other types above, and not throw an exception here. 

```suggestion
        } else {
            addOption(setting, option, value, (v) -> v.toString(), allowBlankValues);
        }

File: warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java [hunk 6]
```diff
@@ -58,6 +58,7 @@
 import datawave.ingest.data.config.ingest.CompositeIngest;
 import datawave.marking.MarkingFunctionsFactory;
 import datawave.query.Constants;
+import datawave.query.ancestor.AncestorQueryIterator;
```
foster33: unused import

warehouse/query-core/src/main/java/datawave/query/iterator/facets/DynamicFacetIterator.java
```diff
@@ -27,6 +27,7 @@
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 
+import datawave.query.ancestor.AncestorQueryIterator;
```
foster33: unused import

warehouse/query-core/src/main/java/datawave/query/iterator/facets/DynamicFacetIterator.java
```diff
@@ -37,6 +38,7 @@
 import datawave.query.function.MinimumEstimation;
 import datawave.query.iterator.AccumuloTreeIterable;
 import datawave.query.iterator.FieldIndexOnlyQueryIterator;
+import datawave.query.iterator.QueryOptions;
```
foster33: unused import

warehouse/query-core/src/main/java/datawave/query/iterator/facets/DynamicFacetIterator.java
```diff
@@ -282,5 +284,4 @@ public FieldIndexAggregator getFiAggregator() {
         }
         return fiAggregator;
     }
-
```
foster33, line 279: there are no changes to this class, so it can be reverted

warehouse/query-core/src/main/java/datawave/query/ancestor/AncestorQueryIterator.java
```diff
@@ -311,4 +312,11 @@ public Equality getEquality() {
         }
         return equality;
     }
+
+    @Override
+    protected DefaultOptions createDefaultOptions() {
```
lbschanno: We do not need to override createDefaultOptions() here, because a different default value for the HIT_LIST options is never actually established. All changes to this class can be reverted.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -153,6 +151,9 @@ public class QueryOptions implements OptionDescriber {
     public static final String GROUP_FIELDS = "group.fields";
     public static final String GROUP_FIELDS_BATCH_SIZE = "group.fields.batch.size";
     public static final String UNIQUE_FIELDS = "unique.fields";
+    public static final String MOST_RECENT_UNIQUE = "most.recent.unique";
```
lbschanno, line 158: The fields MOST_RECENT_UNIQUE and UNIQUE_CACHE_BUFFER_SIZE were removed in commit  2c2de3c0d0. Please delete them here.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -506,6 +508,7 @@ public void deepCopy(QueryOptions other) {
         this.ivaratorCacheDirConfigs = (other.ivaratorCacheDirConfigs == null) ? null : new ArrayList<>(other.ivaratorCacheDirConfigs);
         this.hdfsSiteConfigURLs = other.hdfsSiteConfigURLs;
         this.ivaratorCacheBufferSize = other.ivaratorCacheBufferSize;
+        this.uniqueCacheBufferSize = other.uniqueCacheBufferSize;
```
lbschanno, line 548: This line should be deleted.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -313,6 +314,7 @@ public class QueryOptions implements OptionDescriber {
     protected GroupFields groupFields = new GroupFields();
     protected int groupFieldsBatchSize = Integer.MAX_VALUE;
     protected UniqueFields uniqueFields = new UniqueFields();
+    protected int uniqueCacheBufferSize = 100;
```
lbschanno, line 333: This line should be deleted.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -958,6 +961,14 @@ public void setIvaratorCacheBufferSize(int ivaratorCacheBufferSize) {
         this.ivaratorCacheBufferSize = ivaratorCacheBufferSize;
     }
 
+    public int getUniqueCacheBufferSize() {
```
lbschanno, line 1108: `getUniqueCacheBufferSize()` and `setUniqueCacheBufferSize()` should be deleted.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -2259,4 +2270,214 @@ public Equality getEquality() {
...
+        defaultOptions.putDefaultValue(QueryOptions.EXCERPT_FIELDS, queryOptions.excerptFields);
+        defaultOptions.putDefaultValue(QueryOptions.EXCERPT_FIELDS_NO_HIT_CALLOUT, queryOptions.excerptFieldsNoHitCallout);
+        defaultOptions.putDefaultValue(QueryOptions.EXCERPT_ITERATOR, queryOptions.excerptIterator);
+
+        // This method can be overridden in sub-classes, like following example:
+        return defaultOptions;
+    }
+
+    /*
+     * protected DefaultOptions createDefaultOptions() { SubClass subClass = new SubClass(); DefaultOptions defaultOptions = super.createDefaultOptions(); //
```
lbschanno: This section of commented code should be deleted.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -2259,4 +2270,214 @@ public Equality getEquality() {
...
+                        queryOptions.includeGroupingContext || (queryOptions.groupFields != null && queryOptions.groupFields.hasGroupByFields()));
+        defaultOptions.putDefaultValue(QueryOptions.DOCUMENT_PERMUTATION_CLASSES, queryOptions.documentPermutationClasses);
+        defaultOptions.putDefaultValue(QueryOptions.LIMIT_FIELDS, queryOptions.limitFieldsMap);
+        defaultOptions.putDefaultValue(QueryOptions.MATCHING_FIELD_SETS, queryOptions.matchingFieldSets);
+        defaultOptions.putDefaultValue(QueryOptions.LIMIT_FIELDS_PRE_QUERY_EVALUATION, queryOptions.limitFieldsPreQueryEvaluation);
+        defaultOptions.putDefaultValue(QueryOptions.LIMIT_FIELDS_FIELD, queryOptions.limitFieldsField);
+        defaultOptions.putDefaultValue(QueryOptions.GROUP_FIELDS, queryOptions.groupFields);
+        defaultOptions.putDefaultValue(QueryOptions.GROUP_FIELDS_BATCH_SIZE, queryOptions.groupFieldsBatchSize);
+        defaultOptions.putDefaultValue(QueryOptions.UNIQUE_FIELDS, queryOptions.uniqueFields);
+        defaultOptions.putDefaultValue(QueryOptions.MOST_RECENT_UNIQUE, queryOptions.uniqueFields);
```
lbschanno: This line and the one with UNIQUE_CACHE_BUFFER_SIZE below should be deleted.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -2259,4 +2270,214 @@ public Equality getEquality() {
         }
         return equality;
     }
+
+    private static final Joiner COMMA_JOINER = Joiner.on(",").skipNulls();
+
+    // Map of class strings to DefaultOptions implementations.
```
lbschanno, line 2499: Overall, I would like to see more complete documentation here and for the new methods below that leverage javadoc and go into more complete detail about the contract defined here for specifying default options, as well as caveats for overriding them.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -2497,19 +2502,19 @@ protected void setCommonIteratorOptions(ShardQueryConfiguration config, Iterator
 
         // Include the EVENT_DATATYPE as a field
         if (config.getIncludeDataTypeAsField()) {
-            addOption(cfg, QueryOptions.INCLUDE_DATATYPE, Boolean.toString(true), false);
+            addOption(cfg, QueryOptions.INCLUDE_DATATYPE, Boolean.toString((true)), false);
```
lbschanno: There are several incidences of double wrapping with parentheses here that can be cleaned up to single wrapping.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -2401,6 +2402,10 @@ public static void configureTypeMappings(ShardQueryConfiguration config, Iterato
         }
     }
 
+    public static void addOption(IteratorSetting cfg, String option, boolean value) {
+        cfg.addOption(option, Boolean.toString(value));
+    }
+
     public static void addOption(IteratorSetting cfg, String option, String value, boolean allowBlankValue) {
```
lbschanno: I don't see a change here to actually leverage `QueryOptions.addOption()` so that we only populate the option in the iterator setting if the value is different. Please add it in.

lbschanno (changes requested):

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -2401,18 +2402,24 @@ public static void configureTypeMappings(ShardQueryConfiguration config, Iterato
...
-            }
-            cfg.addOption(option, value);
-        }
+    // public static <T> void addOption(IteratorSetting setting, String option, T value, com.google.common.base.Function<T,String> valueTransformer, boolean
+    // allowBlankValues) {
+    public static <T> void addOption(IteratorSetting cfg, String option, T value, boolean allowBlankValues) {
+        QueryOptions.addOption(cfg, option, value, allowBlankValues);
     }
 
+    // public static void addOption(IteratorSetting cfg, String option, String value, boolean allowBlankValue) {
```
lbschanno: We can remove these commented out lines.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -2259,4 +2260,290 @@ public Equality getEquality() {
...
+        }
+        setting.addOption(option, valueString);
+    }
+
+    /**
+     * Default options for QueryOptions. TODO: Make immutable with a builder that is used in createDefaultOptions().
+     */
+    protected static final class DefaultOptions {
+
+        public final Map<String,Object> defaultValues;
```
lbschanno: This should be private instead of public.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -2259,4 +2260,290 @@ public Equality getEquality() {
...
+        } else if (value instanceof Collection<?>) {
+            addOption(setting, option, value, (v) -> COMMA_JOINER.join((Iterable<?>) v), allowBlankValues);
+        } else if (value instanceof Enum<?>) {
+            addOption(setting, option, value, (v) -> ((Enum<?>) v).name(), allowBlankValues);
+        } else {
+            addOption(setting, option, value, (v) -> v.toString(), allowBlankValues);
+        }
+    }
+
+    // Add given option, using the specified to-string transformer.
```
lbschanno: This non-javadoc comment can be removed.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -2259,4 +2260,290 @@ public Equality getEquality() {
...
+        protected static class Builder {
+            private final Map<String,Object> values;
+
+            public Builder() {
+                values = new HashMap<>();
+            }
+
+            public Builder putDefaultValue(String option, Object value) {
+                if (value.getClass().isAssignableFrom(Collection.class)) {
+                    value = Collections.unmodifiableCollection((Collection<?>) value);
```
lbschanno, line 2776: Upon further thinking, this may cause issues with equalsDefaultValue() when trying to compare the stored immutable collections and any passed-in lists/sets. Can you add a test case to see equalsDefaultValues returns true when a set or list is passed in with default values?

lbschanno (changes requested):

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -2259,4 +2260,303 @@ public Equality getEquality() {
...
+                Class<?> clazz = Class.forName(className);
+                // Check if the given class inherits QueryOptions.
+                if (!QueryOptions.class.isAssignableFrom(clazz)) {
+                    DefaultOptions options = DefaultOptions.builder().build();
+                    defaultOptionsMap.put(className, options);
+                    return options;
+                }
+                // Attempt to fetch the default options from createDefaultOptions on an instance of the class.
+                try {
+                    QueryOptions instance = (QueryOptions) clazz.getConstructor().newInstance();
```
ivakegg, line 2399: Does anything actually extend QueryOptions?

ivakegg:
I had added several comments but then removed them once I figured out what the thought process is here.  I think this looks good, but I need to do some integration testing before bringing it in.

ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -763,6 +763,7 @@ private void configureIterator(ShardQueryConfiguration config, IteratorSetting c
             }
         }
         addOption(cfg, QueryOptions.HIT_LIST, Boolean.toString(config.isHitList()), false);
+        addOption(cfg, QueryOptions.HIT_LIST, Boolean.toString(config.isHitList()), false);
```
ivakegg: Adding this option twice ?

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -2684,18 +2685,22 @@ public void configureTypeMappings(ShardQueryConfiguration config, IteratorSettin
...
-                value = " ";
-            }
-            cfg.addOption(option, value);
-        }
+    public static <T> void addOption(IteratorSetting cfg, String option, T value, boolean allowBlankValues) {
+        QueryOptions.addOption(cfg, option, value, allowBlankValues);
     }
 
+    // public static void addOption(IteratorSetting cfg, String option, String value, boolean allowBlankValue) {
+    // if (StringUtils.isNotBlank(option) && (allowBlankValue || StringUtils.isNotBlank(value))) {
```
ivakegg: Delete this unused block please

warehouse/query-core/src/test/java/datawave/query/CompositeFunctionsTest.java
```diff
@@ -539,6 +539,7 @@ public void testNotNulls() throws Exception {
             // filter must be reset between each run when pruning ingest types
             eventQueryLogic.getConfig().setDatatypeFilter(Collections.emptySet());
             runTestQuery(expectedLists[i], queryStrings[i], format.parse("20091231"), format.parse("20150101"), extraParameters);
+            runTestQuery(expectedLists[i], queryStrings[i], format.parse("20091231"), format.parse("20150101"), extraParameters);
```
ivakegg: Running the tests twice?

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -2477,8 +2491,304 @@ public Equality getEquality() {
...
+    }
+
+    /**
+     * Used to populate the defaultOptionsMap with the default options for each class. When overriding this method, call super.createDefaultOptions() to ensure
+     * the default options are populated. Changes to these defaults will be reflected in the DefaultOptions object returned by this method. Add them with
+     * defaultOptions.putDefaultValue(key, value).
+     *
+     * @return the default options
+     */
+    protected DefaultOptions createDefaultOptions() {
```
ivakegg, line 2514: Can we create a test to verify that this set of options is complete?  See the ShardQueryConfiguratTest on how that might be done using the ObjectMapper.

ivakegg (approved):

lbschanno (dismissed):

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -296,6 +302,7 @@ public class QueryOptions implements OptionDescriber {
 
     protected String scanId;
     protected String query;
+    @JsonIgnore
```
ivakegg: I think we need to have comments as to why we are ignoring certain elements in this file.  It is not obvious.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
lbschanno: Alternatively, we could use a mixIn to modify the serialization of the QueryOptions class without needing to apply any jackson annotations directly to QueryOptions. May not be a huge concern in this case.

ivakegg: I think I would be much better with that.  Using those annotations just for test purposes does not sit well with me.

warehouse/query-core/src/test/java/datawave/query/iterator/QueryOptionsMixin.java
```diff
@@ -0,0 +1,151 @@
...
+import datawave.query.function.DocumentPermutation;
+import datawave.query.function.Equality;
+import datawave.query.function.serializer.DocumentSerializer;
+import datawave.query.iterator.filter.KeyIdentity;
+import datawave.query.jexl.functions.FieldIndexAggregator;
+import datawave.query.predicate.EventDataQueryFilter;
+import datawave.query.predicate.TimeFilter;
+import datawave.query.statsd.QueryStatsDClient;
+import datawave.query.util.sortedset.FileSortedSet;
+
```
lbschanno, line 26: Please add a class-level comment here describing why we have this mixin.

warehouse/query-core/src/test/java/datawave/query/config/ShardQueryConfigurationTest.java
```diff
@@ -644,6 +644,7 @@ private ArrayListMultimap<String,?> createArrayListMultimap(Multimap<String,?> m
     private void testValues(ShardQueryConfiguration config, Map<String,Object> values, Map<String,Predicate> predicates) throws Exception {
         ObjectMapper mapper = JsonMapper.builder().enable(MapperFeature.PROPAGATE_TRANSIENT_MARKER).build();
         JsonNode root = mapper.readTree(mapper.writeValueAsString(config));
+        System.out.println(root);
```
lbschanno: This should be deleted.

warehouse/query-core/src/test/java/datawave/query/iterator/QueryOptionsTest.java
```diff
@@ -259,4 +268,182 @@ protected static String decompressOption(final String buffer, Charset characterS
...
+        limitFields.put("field1", 1);
+        limitFields.put("field2", 2);
+
+        QueryOptions.addOption(iteratorSetting, QueryOptions.LIMIT_FIELDS, limitFields, false);
+
+        Assert.assertTrue(QueryOptions.getDefaultOptions(QueryOptionsWithNonEmptyCollectionsAsDefaultValues.class.getName())
+                        .equalsDefaultValue(QueryOptions.LIMIT_FIELDS, limitFields));
+
+    }
+
```
lbschanno, line 321: Please wrap this map creation with `// @formatter:off` and  `// @formatter:on`, and ensure each put statement is on a separate line for better readability.

lbschanno (changes requested):
Requesting some minor changes.

lbschanno (dismissed):

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -2474,8 +2494,315 @@ public Equality getEquality() {
...
+        DefaultOptions defaultOptions = getDefaultOptions(setting.getIteratorClass());
+        // If the value matches the default value, do not add it to the setting and instead remove it from the IteratorSetting if it exists.
+        if (defaultOptions.hasDefaultValue(option) && defaultOptions.equalsDefaultValue(option, value)) {
+            setting.removeOption(option);
+            return;
+        }
+        // Otherwise convert it, and if blank and blank values allowed, change it to something else, or it will fail in InputFormatBase when run through the
+        // MapReduce api.
+        String valueString = valueTransformer.apply(value);
+        if (allowBlankValues && org.apache.commons.lang.StringUtils.isBlank(valueString)) {
```
lbschanno, line 2731: We can do a static import for the `isBlank()` method here.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -2474,8 +2494,315 @@ public Equality getEquality() {
...
+            valueString = " ";
+        }
+        setting.addOption(option, valueString);
+    }
+
+    protected static final class DefaultOptions {
+
+        private final Map<String,Object> defaultValues;
+
+        private DefaultOptions() {
```
lbschanno, line 2749: This constructor is never used. Do we need it?

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -2474,8 +2494,315 @@ public Equality getEquality() {
...
+            public Builder putDefaultValues(Map<String,Object> values) {
+                values.forEach(this::putDefaultValue);
+                return this;
+            }
+
+            public boolean hasDefaultValue(String option) {
+                return values.containsKey(option);
+            }
+
+            public boolean equalsDefaultValue(String option, Object value) {
```
lbschanno, line 2845: This method is never used. Was it supposed to be used somewhere, or can it be removed?

lbschanno (changes requested):

PR #2422, opened 2024-06-18
2416 - maven build cache fix

cogross:
Changes to allow for the build to work with the maven build cache enabled.
 - mapreduce query microservice pom changed to copy resources earlier
 - quickstart docker changes (disable build cache, for now)

### DIFF
--- a/.mvn/maven-build-cache-config.xml
+++ b/.mvn/maven-build-cache-config.xml
@@ -35,7 +35,7 @@
     <input>
         <global>
             <glob>
-                {*.java,*.sh,*.py,*.properties}
+                {*.java,*.sh,*.py,*.properties,*.yml,*.xml}
             </glob>
             <includes>
                 <include>src/</include>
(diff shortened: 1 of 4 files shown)


jwomeara (changes requested):
I think at a minimum we're going to want to disable the build cache for the assemble-datawave and assemble-webservice modules as well.  I have run into issues with those failing to rebuild on subsequent build attempts.

jwomeara:
Also, as part of your testing can you try building and then rebuilding the full docker compose build, making the following changes independently between builds:
- modify something in ShardQueryLogic.java
- modify something in QueryLogicFactory.xml

I just want to make sure that as we make changes to things and rebuild, that the caching doesn't prevent us from building the full suite of artifacts since that's the kind of behavior that we've been seeing.

cogross:
> I think at a minimum we're going to want to disable the build cache for the assemble-datawave and assemble-webservice modules as well. I have run into issues with those failing to rebuild on subsequent build attempts.

Sure -- makes sense.  Easy enough to add.

cogross:
> Also, as part of your testing can you try building and then rebuilding the full docker compose build, making the following changes independently between builds:
> 
> * modify something in ShardQueryLogic.java
> * modify something in QueryLogicFactory.xml
> 
> I just want to make sure that as we make changes to things and rebuild, that the caching doesn't prevent us from building the full suite of artifacts since that's the kind of behavior that we've been seeing.

It worked for ShardQueryLogic.java.
I made a change to the maven config to let xml (not poms) and yml file changes be taken into account for the maven build cache.  Once I did that, a change to the xml file you gave would cause its module (spring-boot-starter-datawave-query) to be rebuilt.

Changes are pushed up, if you wanted to give it another look over.

cogross:
something got wonky on rebase -- recreating

PR #2424, opened 2024-06-20
Adding Badge and Adding Microservice build and test to gitlab action

doublejai:


.github/workflows/tests.yml
```diff
@@ -73,49 +73,50 @@ jobs:
           || { echo "***** TESTS FAILED. Attempting retry."; $RUN_TESTS; } \
           || { echo "***** TESTS FAILED. Attempting final retry."; $RUN_TESTS; }
 
-  # This step can be uncommented to support feature development.  When developing a 
-  # feature which spans multiple repos, you can update the poms to use the latest 
-  # SNAPSHOT across the board, and run this action as a sanity check to ensure 
+  # This step can be uncommented to support feature development.  When developing a
+  # feature which spans multiple repos, you can update the poms to use the latest
+  # SNAPSHOT across the board, and run this action as a sanity check to ensure
```
jwomeara: You can probably remove this.

README.md
```diff
@@ -2,8 +2,10 @@
    <img src="datawave-readme.png" />
 </p>
 
-[![Apache License][li]][ll] ![Build Status](https://github.com/NationalSecurityAgency/datawave/actions/workflows/tests.yml/badge.svg)
+# My Project
```
jwomeara: This can be removed.

README.md
```diff
@@ -2,8 +2,10 @@
...
 
+[![Apache License][li]][ll] ![Build Status](https://github.com/NationalSecurityAgency/datawave/actions/workflows/tests.yml/badge.svg)
+![2414](https://github.com/NationalSecurityAgency/datawave/actions/workflows/tests.yml/badge.svg?job=build-and-test-microservices)
```
jwomeara: I know it doesn't display, but instead of 2414, a more descriptive name would be good.

jwomeara: We should probably put the build status icons in a table and label them to indicate which is the overall status and which is the docker compose status.

properties/dev.properties
```diff
@@ -1,7 +1,7 @@
 CONFIGURATION=test
 RCPT_TO=hadoop@localhost
 
-docker.image.prefix=
+docker.image.prefix=ghcr.io/nationalsecurityagency/
```
jwomeara: Is this needed?

jwomeara (changes requested):

jwomeara:
I unintentionally lead you astray...  Instead of running the microservice build and test, i wanted to run the docker compose tests.  here's the entry for that.  you should be able to just swap this for the one you added.  sorry about that.
https://github.com/NationalSecurityAgency/datawave/blob/eb62d64971db18ff7f8bac5646b38b9f935bd9c0/.github/workflows/tests.yml#L158

jwomeara:
One thing that might not work if one of the jobs is failing (even though the overall status is successful) is the ability to merge a PR.  I *think* all of the jobs have to be passing to merge a PR.  I created a PR against your PR which should fail the microservice build and let us know one way or the other.

https://github.com/NationalSecurityAgency/datawave/pull/2426

jwomeara:
It looks like your changes might have gotten lost from tests.yml.  I don't see any new test configurations added.  Just some indentation changes, and comment removal.

.github/workflows/tests.yml
```diff
@@ -219,6 +219,65 @@ jobs:
...
+              echo "Docker compose started successfully"
+              break
+            elif [ $attempt -eq $max_attempts ] ; then
+              echo "Failed to start docker compose"
+              exit 1
+            fi
+          done
+          cd scripts
+          ./testAll.sh
+
```
jwomeara, line 290: Can you add the following so that it dumps docker compose logs if the tests fail?

```suggestion
      - name: Dump Logs
        if: failure()
        run: |
          cd docker
          docker compose logs
```

jwomeara (changes requested):
One small request, and then I think this is good to go.

jwomeara (approved):

Issue #37, opened 2018-02-21
Could my D4M documentation be adapted here?


medined:
About 4 years ago I described the D4M schema at https://github.com/medined/D4M_Schema. Could any of that material be adapted for DataWave? If so, I give permission to use any or all of it.

PR #2425, opened 2024-06-20
Add option to sort query by field cardinality

apmoriarty:
Update the sortQueryBeforeGlobalIndex option to get field cardinality from the DatawaveMetadata table

- integrates bugfix from in-memory-accumulo 4.0.1
- integrates forthcoming feature in metadata-utils 4.0.2
- bugfix for how QueryFieldVisitor handles method terms like `FOO.min() == 1`
- corrected test data format for datawave metadata F column

Note: until [metadata-utils #37](https://github.com/NationalSecurityAgency/datawave-metadata-utils/pull/37) is merged and the 4.0.2 tag created, this build will fail.


warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -2796,6 +2797,20 @@ public Tuple2<CloseableIterable<QueryPlan>,Boolean> getQueryRanges(ScannerFactor
         return new Tuple2<>(ranges, needsFullTable);
     }
 
+    protected ASTJexlScript timedSortQueryBeforeGlobalIndex(ShardQueryConfiguration config, MetadataHelper metadataHelper) throws DatawaveQueryException {
+        return visitorManager.timedVisit(config.getTimers(), "SortQueryBeforeGlobalIndex", () -> {
```
ivakegg, line 2796: I am missing if there is a test that covers this change.  Is there a way to test that it did what we want?  Maybe testing the final plan to see that is is different when isSortQueryBeforeGlobalIndex is set to true?

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
apmoriarty: Added a test to verify the query terms are re-ordered by field cardinality and discovered another problem with the RebuildingScannerHelper which led to it's removal from a test.

avgAGB (approved):

ivakegg (approved):

PR #2428, opened 2024-06-20

### LINKED ISSUE #2305: QueryIterators/ScanIterators should log query id first

SethSmucker:
Create QueryLogIterator class that logs a query's ID to the TServer logs

Resolves #2305


warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -65,6 +59,7 @@
 import datawave.ingest.data.config.ingest.CompositeIngest;
 import datawave.query.Constants;
 import datawave.query.DocumentSerialization;
+import datawave.query.QueryParameters;
```
foster33: Unused import.

warehouse/query-core/src/test/java/datawave/query/tables/ShardQueryLogicTest.java
```diff
@@ -360,4 +360,5 @@ public void testNegativeFilterRegexV2() throws Exception {
 
         runTestQuery(expected, queryString, format.parse("20091231"), format.parse("20150101"), extraParameters);
     }
+
```
foster33: This can be removed
```suggestion
```

warehouse/query-core/src/main/java/datawave/query/QueryParameters.java
```diff
@@ -221,4 +221,5 @@ public class QueryParameters {
      * Used to specify model or DB fields that must be treated as strict (cannot be skipped if normalization fails)
      */
     public static final String STRICT_FIELDS = "strict.fields";
+
```
foster33: This can be removed
```suggestion
```

SethSmucker:
Switched the PR over to NSA/log-iterator instead of SethSmucker/log-iterator. 

Replaced by #2451

PR #2433, opened 2024-06-21
support event and index-only fields within filter:includeRegex

hlgp:


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RegexFunctionVisitor.java
```diff
@@ -61,43 +65,75 @@ public Object visit(ASTFunctionNode node, Object data) {
...
+            if (extractChildren.size() == 0) {
+                // nothing to rewrite
+                return returnNode;
+            } else if (keepChildren.size() == 0) {
+                // rewrite all nodes
+                if (identifiers.size() == 1) {
+                    // we've already rewritten our one node
+                    returnNode = extractChildren.get(0);
+                } else {
+                    if (functionMetadata.name().equals(INCLUDE_REGEX)) {
```
hlgp: I originally merged this check with the one below, but found it reads a bit more clearly this way as to which case we're handling

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RegexFunctionVisitor.java
```diff
@@ -61,43 +65,75 @@ public Object visit(ASTFunctionNode node, Object data) {
         if (functionMetadata.name().equals(INCLUDE_REGEX) || functionMetadata.name().equals(EXCLUDE_REGEX)) {
             List<JexlNode> arguments = functionMetadata.args();
 
+            if (arguments.get(0) instanceof ASTAndNode) {
+                // without this check we would have changed the AND to an OR
+                log.warn("filter:includeRegex(FIELD1 && FIELD2...) is currently not supported for rewrite");
```
ivakegg: I am now wondering if it is more likely that we would hit this case with an EXCLUDE_REGEX?  Just check to make sure that the model expansion is not putting an AND node in here for an EXCLUDE_REGEX.  If it is then we need to handle it.

hlgp: I thought the model expansion was always ORs.  Will verify.  Thanks!

hlgp: https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/QueryModelVisitor.java#L602

Looks like it'll always be ORed in the excludeRegex.... should we fix that in a separate ticket?

hlgp: thinking to just handle it...

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RegexFunctionVisitor.java
```diff
@@ -62,42 +66,68 @@ public Object visit(ASTFunctionNode node, Object data) {
...
+            } else if (keepChildren.size() == 0) {
+                // rewrite all nodes
+                if (identifiers.size() == 1) {
+                    // we've already rewritten our one node
+                    returnNode = extractChildren.get(0);
+                } else {
+                    if (functionMetadata.name().equals(INCLUDE_REGEX) && arguments.get(0) instanceof ASTOrNode) {
+                        returnNode = JexlNodeFactory.createOrNode(extractChildren);
+                    } else {
+                        // build an AND node because of how DeMorgan's law works with expanding negations
```
ivakegg, line 94: The code looks good.  However I would like a comment here making it clear that we are ignoring the and/or for the exclude regex case because of ....

ivakegg (approved):

apmoriarty (approved):

PR #2434, opened 2024-06-25
Make #MATCHES_IN_GROUP work with a number

### LINKED ISSUE #2400: MATCHES_IN_GROUP should work with a number
If you do something like $MATCHES__IN_GROUP(field1,value1,field2,10) where field2 is typed as a number, and the value is a number (10 in this case), then the query will fail as follows:

Caused by : ...JexlException: ...JexlEvaluation... (grouping:matchesInGroup(field1,value1,field2,+bE1) method invocation error
Caused by: java.util.regex.PatternSyntaxException: Dangling meta charater '+' nead index 0 +bE1

Basically, we need to normalize the values into a regex.

This may bring up a larger issue of when a user supplies a raw number where a regex is required, we need to think about how to handle that case.

lbschanno:
Enable regex matching for functions in GroupingRequiredFilteredFunctions. Increase test coverage. Verify numeric regex matching works. Update existing unit tests in anticipating of merging in datawave-type-utils/pull/11 which makes changes to how numeric regexes are normalized.

Fixes #2400

NOTE: This PR is dependent on the work in https://github.com/NationalSecurityAgency/datawave-type-utils/pull/11 and should not be merged in before that. (completed, type-utils updated to version 3.0.1)


warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
```diff
@@ -269,10 +272,14 @@ public static FunctionalSet<ValueTuple> includeRegex(Object fieldValue, String r
                     return FunctionalSet.singleton(getHitTerm(fieldValue));
                 }
             } catch (PatternSyntaxException e) {
+                System.out.println("exception occurred " + e);
```
ivakegg: We need to remove these system.out statements please.  Use logging instead.  Or simply revert the changes to this class as nothing was actually changed functionally.

warehouse/query-core/src/test/java/datawave/query/jexl/functions/GroupingFiltersIngest.java
```diff
@@ -0,0 +1,437 @@
...
+        BatchWriter bw = null;
+        BatchWriterConfig bwConfig = new BatchWriterConfig().setMaxMemory(1000L).setMaxLatency(1, TimeUnit.SECONDS).setMaxWriteThreads(1);
+        Mutation mutation;
+
+        try {
+            // write the shard table :
+            bw = client.createBatchWriter(TableName.SHARD, bwConfig);
+            mutation = new Mutation(shard);
+
+            mutation.put(datatype + "\u0000" + corleoneUID, "NAME.BAT.ZIP.FOO.0" + "\u0000" + "SANTINO", columnVisibilityItalian, timeStamp, emptyValue);
```
ivakegg: Most all of our data fields name have only a group, subgroup, and instance.  So that would be something like NAME.BAT.FOO.0 (i.e. only 4-tuples).  I would like to make sure that things are working with data structured like that.  You can have a mix of data if you want to test it with additional grouping.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GroupingRequiredFilterFunctions.java
```diff
@@ -5,6 +5,7 @@
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.Iterator;
```
ivakegg: As far as I can tell there are no functional changes in this class, just reorganization of the code.  Am I correct in this observation?  This is ok, just makes it harder to review.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
lbschanno: Ack, forgot to remove these. Fixed.

warehouse/query-core/src/test/java/datawave/query/jexl/functions/GroupingFiltersIngest.java
lbschanno: Data has been updated to be 4 tuples instead. Verified that the tests passed.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GroupingRequiredFilterFunctions.java
lbschanno: Yes, that was an unused import. Removed.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
```diff
@@ -20,6 +21,8 @@
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
```
ivakegg: I think we can revert the changes to this class entirely.

ivakegg (approved):

apmoriarty (approved):

PR #2440, opened 2024-06-26
Fix ScoredPhrase Bug

### LINKED ISSUE #2441: ContentFunctionsDescriptor fails to parse field from scored functions
The ContentFunctionsDescriptor will fail to parse the field from a scored phrase when calling `fieldsAndTerms`

For example: 

Given a function like `content:scoredPhrase(FIELD, -1.5, termOffsetMap, 'foo', 'bar')` the call to fieldsAndTerms should return `{FIELD}, {foo, bar}`, but it will return an empty field set.

Given a function like `content:scoredPhrase((FIELD_A || FIELD_B), -1.5, termOffsetMap, 'foo', 'bar')` the call to fieldsAndTerms should return `{FIELD_A, FIELD_B}, {foo, bar}`, but it will return an empty field set.

apmoriarty:
Fixes https://github.com/NationalSecurityAgency/datawave/issues/2441

- added a test suite for ContentFunctionsDescriptor
- corrected how fields are parsed from a scored function when calling `fieldsAndTerms`

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionsDescriptor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionsDescriptor.java
@@ -300,36 +300,31 @@ public class ContentFunctionsDescriptor implements JexlFunctionArgumentDescripto
                         }
                     }
                 } else if (CONTENT_SCORED_PHRASE_FUNCTION_NAME.equals(funcName)) {
-                    JexlNode firstArg = args.next();
+                    JexlNode arg = args.next();
 
-                    if (firstArg instanceof ASTNumberLiteral || firstArg instanceof ASTUnaryMinusNode) {
-                        // firstArg is max score value, skip
-                        firstArg = args.next();
-                    }
-
-                    // we override the zones if the first argument is a string
-                    if (firstArg instanceof ASTStringLiteral) {
-                        fields = Collections.singleton(((ASTStringLiteral) firstArg).getLiteral());
-
-                        if (args.peek() instanceof ASTNumberLiteral || args.peek() instanceof ASTUnaryMinusNode) {
-                            args.next(); // max score not needed for fields and terms
-                        }
-
+                    if (arg instanceof ASTNumberLiteral || arg instanceof ASTUnaryMinusNode) {
+                        // if the first argument is a number, then no field exists
                         termOffsetMap = args.next();
                     } else {
-                        JexlNode nextArg = args.peek();
-
-                        // The zones may (more likely) be specified as an identifier
-                        if (!JexlASTHelper.getIdentifiers(firstArg).isEmpty() && !JexlASTHelper.getIdentifiers(nextArg).isEmpty()) {
-                            if (oredFields != null && firstArg instanceof ASTAndNode) {
-                                oredFields.setValue(false);
-                            }
-
-                            fields = JexlASTHelper.getIdentifierNames(firstArg);
-                            termOffsetMap = args.next();
+                        if (arg instanceof ASTIdentifier) {
+                            // single field case
+                            fields = Collections.singleton(String.valueOf(JexlASTHelper.getIdentifier(arg)));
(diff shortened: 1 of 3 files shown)


ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionsDescriptor.java
```diff
@@ -300,36 +300,31 @@ public Set<String>[] fieldsAndTerms(Set<String> termFrequencyFields, Set<String>
...
-                            }
-
-                            fields = JexlASTHelper.getIdentifierNames(firstArg);
-                            termOffsetMap = args.next();
+                        if (arg instanceof ASTIdentifier) {
+                            // single field case
+                            fields = Collections.singleton(String.valueOf(JexlASTHelper.getIdentifier(arg)));
                         } else {
-                            termOffsetMap = firstArg;
+                            // multi field case
```
drewfarris, line 315: it might be helpful to add an "e.g.," here and for the single field case above, such as:

```
// multi field case, e.g., content:scoredPhrase((FIELD_A || FIELD_B), -1.5, termOffsetMap, 'foo', 'bar')
```

Not critical b/c there are examples in the unit test, but it makes the code a bit more readable.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionsDescriptor.java
```diff
@@ -300,36 +300,31 @@ public Set<String>[] fieldsAndTerms(Set<String> termFrequencyFields, Set<String>
...
+                            if (!identifiers.isEmpty()) {
+                                fields = identifiers;
+
+                                if (oredFields != null && arg instanceof ASTAndNode) {
+                                    oredFields.setValue(false);
+                                }
+                            }
                         }
+
+                        // skip score
```
drewfarris: It would be nice to explain why - the code did this previously...

```
// max score not needed for fields and terms
```

drewfarris (dismissed):
It would be nice to add some additional commentary in the comments, but fundamentally this code looks ok.

For future work, it would be nice to consider restructuring the code to avoid a bunch of nested if/else statements. I'm curious how sonar would rate the cyclomatic complexity of this class.

warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionsDescriptor.java
apmoriarty: I can absolutely do that

apmoriarty:
> It would be nice to add some additional commentary in the comments, but fundamentally this code looks ok.
> 
> For future work, it would be nice to consider restructuring the code to avoid a bunch of nested if/else statements. I'm curious how sonar would rate the cyclomatic complexity of this class.

I originally wanted to break apart that method but decided not to for the sake of keeping the change set small. Since you've asked, I'm more than happy to split out the scored phrase logic into it's own method. Sonar Lint shows the method complexity at 100 -- far too much.

apmoriarty:
> > It would be nice to add some additional commentary in the comments, but fundamentally this code looks ok.
> > For future work, it would be nice to consider restructuring the code to avoid a bunch of nested if/else statements. I'm curious how sonar would rate the cyclomatic complexity of this class.
> 
> I originally wanted to break apart that method but decided not to for the sake of keeping the change set small. Since you've asked, I'm more than happy to split out the scored phrase logic into it's own method. Sonar Lint shows the method complexity at 100 -- far too much.

I took a look at this and to do it right would require a massive expansion of scope. Wrote https://github.com/NationalSecurityAgency/datawave/issues/2456 to track that.

alerman (approved):

keith-ratcliffe:
Synced w/integration just to get the Docker file fixes for the `quickstart-build-and-test` failure.

keith-ratcliffe (approved):

PR #2442, opened 2024-06-26
Fixing shard splits with date only, no suffixes

avgAGB:


warehouse/accumulo-extensions/src/main/java/datawave/ingest/table/volumeChoosers/ShardedTableDateBasedTieredVolumeChooser.java
```diff
@@ -52,6 +52,7 @@ public class ShardedTableDateBasedTieredVolumeChooser extends RandomVolumeChoose
     private static final DateTimeFormatter FORMATTER = DateTimeFormatter.ofPattern(DATE_PATTERN);
 
     private static Pattern SHARD_PATTERN = Pattern.compile("\\d{8}_\\d+");
+    private static Pattern SHARD_PATTERN_SUFFIXLESS = Pattern.compile("\\d{8}");
```
apmoriarty: I stumbled over the word suffixless several times in when reading it silently. Might be easier to understand if it was `SHARD_PATTERN_NO_SUFFIX`?

avgAGB: fair point, I can fix that.

alerman (approved):

keith-ratcliffe (approved):

apmoriarty (approved):

Issue #2445, opened 2024-06-27
QueryFunctionsDescriptor does not return correct fields


apmoriarty:
The QueryFunctionsDescriptor `fields()` method does not return the correct fields for certain functions.

While most query functions are removed from the query tree almost immediately by the QueryOptionsFromQueryVisitor, several functions will remain in the query tree. Notably the `includeText` function explicitly supports multiple fields.

foster33:
Seems like this was covered in #2446, can this be closed?

PR #2446, opened 2024-06-27
QueryFunctionsDescriptor returns all fields for multi-fielded functions

apmoriarty:
Addresses https://github.com/NationalSecurityAgency/datawave/issues/2445

- added unit test for QueryFunctionsDescriptor
- corrected bug that prevented all fields from being returned in a multi-fielded function
- added test cases to QueryOptionsFromQueryVisitorTest


warehouse/query-core/src/main/java/datawave/query/jexl/functions/QueryFunctionsDescriptor.java
```diff
@@ -126,24 +129,74 @@ public void addFilters(AttributeFactory attributeFactory, Map<String,EventDataQu
 
         @Override
         public Set<String> fieldsForNormalization(MetadataHelper helper, Set<String> datatypeFilter, int arg) {
-            // Do not normalize fields for the includeText function.
-            if (!name.equalsIgnoreCase(INCLUDE_TEXT)) {
-                // All other functions use the fields in the first argument for normalization.
-                if (arg > 0) {
-                    return fields(helper, datatypeFilter);
-                }
+            if (name.equalsIgnoreCase(QueryFunctions.INCLUDE_TEXT)) {
```
ivakegg: Read the function description.  The arg argument is to be used to denote argument we are to be returning the fields for normalization.  So for example if the first argument denotes the name of the fields, then for arg=0, this should return the empty set because we are not normalizing the field names themselves.

apmoriarty: Got it, I'll revert that change. Do any additional functions need to delegate to the fields method besides `includeText` (such as `matchRegex`) ?

warehouse/query-core/src/main/java/datawave/query/jexl/functions/QueryFunctionsDescriptor.java
```diff
@@ -138,12 +142,63 @@ public Set<String> fieldsForNormalization(MetadataHelper helper, Set<String> dat
...
+                case QueryFunctions.MAX:
+                case QueryFunctions.AVERAGE:
+                case QueryFunctions.GROUPBY_FUNCTION:
+                case QueryFunctions.NO_EXPANSION:
+                case QueryFunctions.LENIENT_FIELDS_FUNCTION:
+                case QueryFunctions.STRICT_FIELDS_FUNCTION:
+                    // In practice each of these functions should be parsed from the query
+                    // almost immediately. This implementation is added for consistency
+                    for (JexlNode arg : args) {
+                        fields.addAll(JexlASTHelper.getIdentifierNames(arg));
```
mineralntl, line 158: Just for clarification - does it matter that we're doing a simple addAll(identifier) here but down below we're doing a UniqueFields.from(identifier)?

apmoriarty: No, it does not matter. These functions will all have arguments that are ASTIdentifiers. The #UNIQUE function might have arguments that are either an ASTIdentifier OR an ASTStringLiteral.

mineralntl: 👍

mineralntl (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/QueryFunctionsDescriptor.java
ivakegg: So the idea was that the "fields" being returned are the the fields for which thier normalizers should be applied to the argument in question.  Originally all of the methods being handled by the descriptor essentially would need to normalize all arguments past the first one using all of the fields, except for include_text which was explicitly made to bypass any normalization.
So to answer your question, look at the arguments for all methods being handled by the QueryFunctions and see if that holds.  If not then we may need to adjust something.

apmoriarty: After some poking around I don't believe matchRegex needs to be added to the exclusion list.

alerman (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/QueryFunctionsDescriptor.java
```diff
@@ -208,7 +263,7 @@ private static void verify(String name, int numArgs) {
             case QueryFunctions.GROUPBY_FUNCTION:
             case QueryFunctions.EXCERPT_FIELDS_FUNCTION:
             case QueryFunctions.MATCH_REGEX:
-            case QueryFunctions.INCLUDE_TEXT:
+            case INCLUDE_TEXT:
```
jwomeara: Inconsistent with the other cases.  either do a static import for all terms, or none.

jwomeara (commented):
looks safe from a query microservices perspective.  happy to approve from a codeowner perspective once this has 2 approvals.

mineralntl (approved):

hlgp (approved):

jwomeara (approved):

PR #2450, opened 2024-06-27
Extract all fields from filter:includeRegex and filter:excludeRegex …

hlgp:
…when an index-only field is present

Correct RegexFunctionVisitor incorrectly calling index-only fields non-event fields 
Correctly apply De Morgan's law with ANDed exclusions

### DIFF
--- a/warehouse/query-core/src/test/java/datawave/query/CompositeFunctionsTest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/CompositeFunctionsTest.java
@@ -783,6 +783,22 @@ public abstract class CompositeFunctionsTest {
     }
 
     @Test
+    public void test() throws Exception {
+        eventQueryLogic.setParser(new LuceneToJexlQueryParser());
+        Map<String,String> extraParameters = new HashMap<>();
+        // @formatter:off
+        String[] queryStrings = {
+                "UUID:SOPRANO AND #INCLUDE(LOCATION || POSIZIONE || NAME, 'newjersey')"
+        };
+        @SuppressWarnings("unchecked")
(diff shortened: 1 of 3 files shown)


warehouse/query-core/src/test/java/datawave/query/CompositeFunctionsTest.java
```diff
@@ -782,6 +782,22 @@ public void testFilterFunctionsInvalidatedByIndexOnlyFields() throws ParseExcept
         }
     }
 
+    @Test
+    public void test() throws Exception {
```
apmoriarty: rename to something like test index only rewrite. using the 'plan' end point would allow you to verify the rewritten filter function.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/RegexFunctionVisitorTest.java
```diff
@@ -186,6 +186,13 @@ public void testDoubleEndedWildCard2() throws ParseException {
         assertVisitorResult(query, query, indexOnlyFields);
     }
 
+    @Test
+    public void testMultiFieldDoubleEndedWildCard() throws ParseException {
```
apmoriarty, line 190: add a comment explaining why multi fielded filter functions with an index only field are okay given the presence of a double ended wildcards?

hlgp: This functionality has not changed, it simply wasn't covered before.

apmoriarty (approved):

ivakegg (approved):

PR #2451, opened 2024-06-28

### LINKED ISSUE #2305: QueryIterators/ScanIterators should log query id first

SethSmucker:
Create QueryLogIterator class that logs a query's ID to the TServer logs

Resolves #2305


warehouse/query-core/src/main/java/datawave/query/iterator/QueryLogIterator.java
```diff
@@ -0,0 +1,191 @@
...
+        } finally {
+            logEndOf("init()");
+        }
+    }
+
+    /**
+     * Logs the query id and {@link SortedKeyValueIterator} method name before the method is run.
+     */
+    private void logStartOf(String methodName) {
+        if (log.isInfoEnabled()) {
```
ivakegg: I am thinking the info level may not be appropriate.  Maybe debug instead.

ivakegg (changes requested):

ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/iterator/QueryLogIterator.java
```diff
@@ -0,0 +1,191 @@
...
+    @Override
+    public void init(SortedKeyValueIterator<Key,Value> source, Map<String,String> options, IteratorEnvironment env) throws IOException {
+
+        try {
+            logStartOf("init()");
+            this.queryID = options.get(QUERY_ID);
+            this.source = source;
+            this.env = env;
+        } finally {
+            logEndOf("init()");
```
billoley, line 64: Not sure there is value in calling logStartOf and logEndOf here since we are not making any call on source between these two calls

warehouse/query-core/src/main/java/datawave/query/iterator/QueryLogIterator.java
```diff
@@ -0,0 +1,191 @@
...
+    public boolean hasTop() {
+        boolean result;
+
+        try {
+            logStartOf("hasTop()");
+            result = source.hasTop();
+        } finally {
+            logEndOf("hasTop()");
+        }
+        return result;
```
billoley: I think the start/end logs on this method are overkill, but would like to see changing the thread name and adding queryId to a MappedDiagnosticContext

warehouse/query-core/src/main/java/datawave/query/iterator/QueryLogIterator.java
```diff
@@ -0,0 +1,191 @@
...
+        Value v;
+        try {
+            logStartOf("getTopValue()");
+            v = source.getTopValue();
+        } finally {
+            logEndOf("getTopValue()");
+        }
+        return v;
+    }
+
```
billoley, line 185: I think the start/end logs on these methods are overkill, but would like to see changing the thread name and adding queryId to a MappedDiagnosticContext

warehouse/query-core/src/main/java/datawave/query/iterator/QueryLogIterator.java
```diff
@@ -0,0 +1,191 @@
...
+        QueryLogIterator copy;
+
+        try {
+            logStartOf("deepCopy()");
+            copy = new QueryLogIterator(this, this.env);
+        } finally {
+            logEndOf("deepCopy()");
+        }
+        return copy;
+    }
```
billoley, line 216: I think the start/end logs on this method are overkill, but would like to see changing the thread name and adding queryId to a MappedDiagnosticContext

warehouse/query-core/src/main/java/datawave/query/iterator/QueryLogIterator.java
```diff
@@ -0,0 +1,191 @@
...
+    @Override
+    public void seek(Range range, Collection<ByteSequence> collection, boolean b) throws IOException {
+
+        try {
+            logStartOf("seek()");
+            this.source.seek(range, collection, b);
+        } finally {
+            logEndOf("seek()");
+        }
+    }
```
billoley, line 239: Would like to see changing the thread name and adding queryId to a MappedDiagnosticContext

warehouse/query-core/src/main/java/datawave/query/iterator/QueryLogIterator.java
```diff
@@ -0,0 +1,191 @@
...
+     */
+    @Override
+    public void next() throws IOException {
+        try {
+            logStartOf("next()");
+            source.next();
+        } finally {
+            logEndOf("next()");
+        }
+    }
```
billoley, line 124: Would like to see changing the thread name and adding queryId to a MappedDiagnosticContext

billoley (changes requested):
How to update thread name and restore old name
Also how to add and remove a value to the MapedDiagnosticContext (MDC)

```
String oldName = Thread.currentThread().getName();
Thread.currentThread().setName(oldName + " -> " + queryId);
MDC.put("queryId", queryId);
try {
    // call corresponding method of the source
} finally {
    Thread.currentThread().setName(oldName);
    MDC.remove("queryId");
}
```


Keep the start/stop logging in seek and next, but it's overkill in the others
Add the code to change/restore thread name and add/remove MDC in all methods that are overridden from the base class and which call that method on source

hlgp:
Reopen when ready for eyes

PR #2463, opened 2024-07-08
Removing error for large document sizes

mineralntl:
This is the quick and dirty removing of error logging for large Document sizes. Adding the actual document size to the metrics coming in a follow-on PR.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/attributes/Document.java
+++ b/warehouse/query-core/src/main/java/datawave/query/attributes/Document.java
@@ -1,5 +1,27 @@
 package datawave.query.attributes;
 
+import com.esotericsoftware.kryo.Kryo;
+import com.esotericsoftware.kryo.io.Input;
+import com.esotericsoftware.kryo.io.Output;
+import com.google.common.collect.Iterators;
+import com.google.common.collect.Sets;
+import datawave.marking.MarkingFunctions;
+import datawave.query.Constants;
+import datawave.query.collections.FunctionalSet;
+import datawave.query.composite.CompositeMetadata;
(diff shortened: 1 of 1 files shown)


drewfarris:
Perhaps we could complete the other PR that adds the document size metrics elsewhere first?

I'm a little worried that this will get merged and the new feature won't - and we'll lose a tool for tracking causes of tserver memory issues as a result.

mineralntl:
> Perhaps we could complete the other PR that adds the document size metrics elsewhere first?
> 
> I'm a little worried that this will get merged and the new feature won't - and we'll lose a tool for tracking causes of tserver memory issues as a result.

That's fair - I can make those changes in this PR instead of splitting them out.

warehouse/query-core/src/main/java/datawave/query/attributes/Document.java
```diff
@@ -211,17 +211,8 @@ public Attribute<?> toDocKeyAttributes(Set<Key> docKeys, boolean keepRecordId) {
...
-            log.error("Document " + docKey + "; size = " + size() + "; bytes = " + bytes);
-        }
-        // if more than 10M, then warn
-        // else if (bytes > (1024l * 1000 * 10)) {
-        // log.warn("Document " + docKey + "; size = " + size() + "; bytes = " + bytes);
-        // }
-
         // if more than 1M, then info
-        else if (bytes > (ONE_M)) {
+        if (bytes > (ONE_M)) {
```
ivakegg, line 215: Perhaps too much info now.  Make this ONE_HUNDRED_M and the debug below for ONE_M.

mineralntl:
Closing as OBE

PR #2465, opened 2024-07-09
2417 dedicated rabbitmq for query results

cogross:
Create a dedicated rabbitmq for the query results to use.  It leaves the other one in place.


cogross:
stopped working after a rebase - new one pushed.

PR #2472, opened 2024-07-16
Adds github action that runs docker compose tests against the latest snapshots

### LINKED ISSUE #2439: Create a github action which runs docker compose tests with everything set to the latest snapshot
After we standardize our version property names via #2436 we should create a github action which changes every one of our datawave dependency versions to be the latest snapshot.  Then we should run the docker compose build, and the testAll script to see if the latest version of each of the services works together.

foster33:
This PR creates a github action which runs docker compose tests with all dependencies that have `<groupId>'s` that fall into `gov.nsa.*`. All of those dependencies are also updated to the latest SNAPSHOT

Closes #2439


.github/workflows/tests.yml
```diff
@@ -150,6 +150,90 @@ jobs:
...
+          cd scripts
+          ./testAll.sh
+
+      - name: Dump Logs
+        if: failure()
+        run: |
+          cd docker
+          docker compose logs
+
+      - name: Docker Compose Cleanup
```
jwomeara: We can probably get rid of this section.  Cleanup doesn't matter since it is a docker container.  This was a goof on my part.

jwomeara (dismissed):

.github/workflows/tests.yml
foster33: removed this section in latest commit

jwomeara (dismissed):

.github/workflows/tests.yml
```diff
@@ -150,6 +150,84 @@ jobs:
...
+          sudo rm -rf $AGENT_TOOLSDIRECTORY
+      - name: Checkout Code
+        uses: actions/checkout@v3
+        with:
+          submodules: 'recursive'
+      - name: Set up JDK ${{env.JAVA_VERSION}}
+        uses: actions/setup-java@v3
+        with:
+          distribution: ${{env.JAVA_DISTRIBUTION}}
+          java-version: ${{env.JAVA_VERSION}}
```
jsduncan2, line 170: I think we want to add the following lines here:
```suggestion
          java-version: ${{env.JAVA_VERSION}}
          maven-version: 3.9.5
          cache: 'maven'
```
And then we can remove the `actions/cache@v3` step (following this commit ea54c50592dd66723f1209791816e57fe1ed2e3c).

jsduncan2 (changes requested):

.github/workflows/tests.yml
foster33: Made these changes in my latest commit.

jsduncan2 (approved):

jwomeara (approved):

PR #2474, opened 2024-07-17
feature/mapService

jwomeara:
i pulled some logic out of the query module and into their own modules, with minimal dependencies

the new modules are
- datawave-core-exceptions
- datawave-core-geo
- datawave-core-jexl
- datawave-core-language

i wanted to be able to parse lucene and jexl queries, do some minimal jexl visitation, and obviously use some geo logic without having to include EVERYTHING in query-core

unfortunately pulling logic out of query-core means a lot of downstream changes

so many things needed to be updated, but it's mostly just package name changes


core/geo/src/main/java/datawave/core/geo/function/geo/ZIndexGeoFunctionDetails.java
```diff
@@ -0,0 +1,83 @@
...
+
+    public ZIndexGeoFunctionDetails(String name, Set<String> fields, double minLon, double maxLon, double minLat, double maxLat) {
+        this.name = name;
+        this.fields = fields;
+
+        // is the lower left longitude greater than the upper right longitude?
+        // if so, we have crossed the anti-meridian and should split
+        queryGeometry = new ArrayList<>();
+        if (minLon > maxLon) {
+            queryGeometry.add(new BoundingBox(minLon, MAX_LON, minLat, maxLat));
```
alerman, line 35: This logic looks to be repetitive. Consider moving to helper function

core/geo/src/main/java/datawave/core/geo/function/geo/NumericIndexGeoFunctionDetails.java
```diff
@@ -0,0 +1,147 @@
...
+
+    public NumericIndexGeoFunctionDetails(String name, String lonField, String latField, double minLon, double maxLon, double minLat, double maxLat) {
+        this.name = name;
+        this.lonField = lonField;
+        this.latField = latField;
+
+        // is the lower left longitude greater than the upper right longitude?
+        // if so, we have crossed the anti-meridian and should split
+        boundingBoxes = new ArrayList<>();
+        if (minLon > maxLon) {
```
alerman, line 48: This logic looks to be repetitive. Consider moving to helper function

core/geo/src/main/java/datawave/core/geo/utils/GeoQueryConfig.java
```diff
@@ -0,0 +1,159 @@
...
+        return geoMaxExpansion;
+    }
+
+    public boolean isOptimizeGeoRanges() {
+        return optimizeGeoRanges;
+    }
+
+    public static class Builder {
+        // geowave configuration
+        private int geowaveMaxEnvelopes = 4;
```
alerman, line 69: Are these configurable, and just the defaults are hardcoded here?

core/geo/src/main/java/datawave/core/geo/utils/GeoUtils.java
```diff
@@ -0,0 +1,617 @@
...
+        return shapeFactory.createCircle();
+    }
+
+    public static Polygon createRectangle(double minLon, double maxLon, double minLat, double maxLat) {
+        List<Coordinate> coordinates = new ArrayList<>();
+        coordinates.add(new CoordinateXY(minLon, minLat));
+        coordinates.add(new CoordinateXY(maxLon, minLat));
+        coordinates.add(new CoordinateXY(maxLon, maxLat));
+        coordinates.add(new CoordinateXY(minLon, maxLat));
+        coordinates.add(new CoordinateXY(minLon, minLat));
```
alerman, line 545: duplocate of 546? intentional or copy/paste error?

core/jexl/src/main/java/datawave/core/query/jexl/JexlASTHelper.java
```diff
@@ -0,0 +1,1794 @@
...
+                return range;
+            }
+
+            if (marked)
+                throw new DatawaveFatalQueryException("A bounded range must contain bounds with comparable types");
+            return null;
+        }
+    }
+
+    public static LiteralRange<String> getStringBoundedRange(JexlNode[] children, LiteralRange<String> range) {
```
alerman, line 1100: Would it be possible to genericize these similar methods?

core/jexl/src/main/java/datawave/core/query/jexl/JexlASTHelper.java
```diff
@@ -0,0 +1,1794 @@
...
+    /**
+     * Iterate through provided node's ancestors and return true if any are an instance of ASTNotNode.
+     *
+     * @see org.apache.commons.jexl3.parser.ASTNotNode
+     *
+     * @param node
+     *            a JexlNode
+     * @return True if any ancestor is an instance of ASTNotNode. If not, then False.
+     */
+    public static boolean isDescendantOfNot(JexlNode node) {
```
alerman, line 1335: Should all of thiese just call isDescendantOfNodeType and pass the appropriate type?

core/geo/src/main/java/datawave/core/geo/utils/GeoQueryConfig.java
jwomeara: they are configurable

core/geo/src/main/java/datawave/core/geo/utils/GeoUtils.java
jwomeara: intentional.  that's how polygon definitions work

core/jexl/src/main/java/datawave/core/query/jexl/JexlASTHelper.java
jwomeara: these aren't methods created by me.  they are just moved from their previous location

core/jexl/src/main/java/datawave/core/query/jexl/JexlASTHelper.java
jwomeara: same comment as above.  i didn't recreate any of this, i just moved it to a different file

core/geo/src/main/java/datawave/core/geo/function/AbstractGeoFunctionDetails.java
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        JexlNode finalNode;
+        if (!orNodes.isEmpty()) {
+            finalNode = orNodes.size() > 1 ? JexlNodeFactory.createOrNode(orNodes) : orNodes.get(0);
+        } else {
+            finalNode = new ASTTrueNode(ParserTreeConstants.JJTTRUENODE);
+        }
+
+        return finalNode;
```
apmoriarty, line 59: When a method ends on conditionally building a junction, I'm partial to the following syntax
```java
switch(orNodes.getNumChildren()){
  case 0:
    return new ASTTrueNode();
  case 1:
    return orNodes.get(0);
  default:
    return JexlNodeFactory.createOrNode(orNodes);
}
```

jwomeara: i don't think what you have is any more/less readable than what i currently have.  the current logic is less lines of code though.

i don't like multiple return statements in a method either.

core/geo/src/main/java/datawave/core/geo/utils/GeoQueryConfig.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o)
+            return true;
+        if (o == null || getClass() != o.getClass())
+            return false;
+        GeoQueryConfig that = (GeoQueryConfig) o;
+        return geowaveMaxEnvelopes == that.geowaveMaxEnvelopes && geometryMaxExpansion == that.geometryMaxExpansion
```
apmoriarty: Opportunity to place each variable on it's own line, similar to what you did with the shard query config/logic?

core/geo/src/test/java/datawave/core/geo/utils/GeoWaveUtilsTest.java
```diff
@@ -0,0 +1,232 @@
...
+import org.locationtech.jts.geom.Polygon;
+import org.locationtech.jts.io.WKTReader;
+
+import datawave.data.normalizer.GeometryNormalizer;
+import datawave.data.normalizer.PointNormalizer;
+
+public class GeoWaveUtilsTest {
+
+    // add tests that showcase geo fields used in geowave functions
+    // add tests that showcase geowave fields used in geo functions
```
apmoriarty: Are these outstanding TODO items? If so, do they need to be completed as part of this work or can it be future work?

core/language/src/main/java/datawave/core/query/language/builder/jexl/FieldQueryNodeBuilder.java
```diff
@@ -16,26 +29,13 @@
...
-import org.apache.lucene.queryparser.flexible.core.builders.QueryBuilder;
-import org.apache.lucene.queryparser.flexible.core.nodes.EscapedNodes;
-import org.apache.lucene.queryparser.flexible.core.nodes.FieldQueryNode;
-import org.apache.lucene.queryparser.flexible.core.nodes.FuzzyQueryNode;
-import org.apache.lucene.queryparser.flexible.core.nodes.QueryNode;
-import org.apache.lucene.queryparser.flexible.core.nodes.QuotedFieldQueryNode;
-import org.apache.lucene.queryparser.flexible.core.util.UnescapedCharSequence;
-import org.apache.lucene.search.TermQuery;
-
 import com.google.common.collect.Sets;
```
alerman, line 32: Should we put all the imports above the license comment?

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -1657,7 +57,7 @@ public static boolean isIndexed(JexlNode node, ShardQueryConfiguration config) {
         Preconditions.checkNotNull(indexedFieldsDatatypes);
```
apmoriarty, line 57: Can you briefly walk me through the thought process behind having two JexlASTHelpers? If I'm going to add a method to a JexlASTHelper, what's the criteria for placing a method in the core-query instance vs. the query-core instance?

core/language/src/main/java/datawave/core/query/language/builder/jexl/FieldQueryNodeBuilder.java
jwomeara: i have no idea.  that's just the import sorting plugin at work.

core/geo/src/test/java/datawave/core/geo/utils/GeoWaveUtilsTest.java
jwomeara: these comments can be removed.

core/geo/src/main/java/datawave/core/geo/utils/GeoQueryConfig.java
jwomeara: doesn't matter to me.  we are consistently inconsistent throughout the code.

core/language/src/main/java/datawave/core/query/language/builder/jexl/FieldQueryNodeBuilder.java
jwomeara: also, these are all deprecated.  that being said... as you wish.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
jwomeara: If you look at datawave.query.jexl.JexlASTHelper you will notice that all of the methods in this class (aside from the two at the end added by Charlie after I made this PR) depend on at least one class in datawave-query-core.  Some examples are RangeStream, JexlArgumentDescriptorFactory, and IndexStatsClient).  Those classes are more query-specific than they are jexl-generic, and thus don't really belong in the datawave-core-jexl module.  I also don't want datawave-core-jexl to have a dependency on datawave-query-core, so the only option would be to move those classes into yet another separate module (and deal with all of the implications of that) or extend JexlASTHelper in datawave-query-core to 1) maintain backwards compatibility, and 2) provide a JexlASTHelper class which has access to those constructs.  

I think an argument could be made that JexlASTHelper has become somewhat of a dumping ground, and that those methods in datawave.query.jexl.JexlASTHelper could and probably should be moved into a class other than JexlASTHelper.  If we ever achieve that, then we will be back to having a single JexlASTHelper.

ivakegg:
Put into the backlog and the ticket is being reprioritized.

PR #2475, opened 2024-07-18
Update Deprecated Accumulo API Calls

SethSmucker:
Replace deprecated Accumulo API calls that have a direct replacement.


billoley:
A better title might be:

Replace deprecated org.apache.commons.cli.Option calls

SethSmucker:
There are a few other updates that may be added along side the `org.apache.commons.cli.Option` call updates. Do you think it would be better to keep them within the same PR or break them into individual PRs?

PR #2476, opened 2024-07-18
Corrected logic in how negated, null, and not null terms are handled …

apmoriarty:
…by the IngestTypeVisitor and IngestTypePruningVisitor.

Added stochastic integration test for the IngestTypeVisitor

Updated QueryGenerator to support fine-grained query generation

Added concept of internal vs. external visits for the IngestTypeVisitor. Internal visits will return the effective types for a node or subtree (the types for executable terms), while an external visit will return all types regardless of a term's executability state.


alerman (approved):

avgAGB (approved):

jwomeara (approved):
i don't see any obvious query microservice breakages...

PR #2478, opened 2024-07-19
Fixed off-by-one error that would prevent final ssdeep ngram generation

drewfarris:


apmoriarty (approved):

austin007008 (approved):

jwomeara (approved):
looks safe from a query microservices perspective

PR #2479, opened 2024-07-22
CompositeQueryLogicTransformer must implement WritesQueryMetrics interface, gather stats from delegates

billoley:
Query logics using the CompositeQueryLogicTransformer are not populating sourceCount, seekCount, nextCount, yieldCount, fiRanges, docRanges in the query metric because CompositeQueryLogicTransformer doesn't implement the WritesQueryMetrics interface.

I considered delegating the totaling of the various fields to the metric-api module, but after adding "implements WritesQueryMetrics," to CompositeQueryLogicTransformer the class must implement the getters for sources, seeks, nexts, etc which iterate over the delegates to total these values.

With those methods in place, it is just simpler to use those totals in writeQueryMetrics.  Any future metric field counts held by a transformer would have to be added to the WritesQueryMetrics interface.


apmoriarty (dismissed):

keith-ratcliffe (dismissed):

keith-ratcliffe (approved):

apmoriarty (approved):

jwomeara (approved):
looks safe enough from a microservices perspective.

ivakegg (approved):

PR #2480, opened 2024-07-26
replace apache base64 with java base64

austin007008:
continued from cbevard1

removed apache commons base64 encoding/decoding

removed unused commons-codec dependency


warehouse/query-core/src/main/java/datawave/core/iterators/ColumnRangeIterator.java
```diff
@@ -102,22 +102,18 @@ public void setScanLimit(long scanLimit) {
 
     public static String encodeRange(Range range) throws IOException {
         ByteArrayOutputStream b = new ByteArrayOutputStream();
-        DataOutputStream d = new DataOutputStream(b);
 
-        try {
+        try (b; DataOutputStream d = new DataOutputStream(b)) {
```
apmoriarty: You could do something like this
```java
        try (ByteArrayOutputStream b = new ByteArrayOutputStream(); DataOutputStream d = new DataOutputStream(b)) {
            range.write(d);
            byte[] payload = Base64.encodeBase64(b.toByteArray());
            return new String(payload);
        } catch (Exception e) {
            throw new IOException(e);
        }
```

apmoriarty (dismissed):

apmoriarty (dismissed):

keith-ratcliffe (changes requested):
I'm pretty sure that  [org.apache.commons.codec.binary.Base64](https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/binary/Base64.html) as used here is based on the RFC2045 spec, which is not the same as [java.util.Base64.getDecoder](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/util/Base64.java#L154) / [java.util.Base64.getEncoder](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/util/Base64.java#L89), which could cause issues for any existing data that we have encoded

I think you want [java.util.Base64.getMimeDecoder](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/util/Base64.java#L175) / [java.util.Base64.getMimeEncoder](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/util/Base64.java#L110) to remain backward-compatible with any existing data

austin007008:
> I'm pretty sure that [org.apache.commons.codec.binary.Base64](https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/binary/Base64.html) as used here is based on the RFC2045 spec, which is not the same as [java.util.Base64.getDecoder](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/util/Base64.java#L154) / [java.util.Base64.getEncoder](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/util/Base64.java#L89), which could cause issues for any existing data that we have encoded
> 
> I think you want [java.util.Base64.getMimeDecoder](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/util/Base64.java#L175) / [java.util.Base64.getMimeEncoder](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/util/Base64.java#L110) to remain backward-compatible with any existing data

I will double check but from what I see, the encoder we were using from apache does not cut the data up into 76 character lines:
 "encodeBase64(byte[] binaryData)  Encodes binary data using the base64 algorithm but does not chunk the output."
so the basic java encoder should work fine
"Uses "The Base64 Alphabet" as specified in Table 1 of RFC 4648 and RFC 2045 for encoding and decoding operation. The encoder does not add any line feed (line separator) character."

austin007008:
> I'm pretty sure that [org.apache.commons.codec.binary.Base64](https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/binary/Base64.html) as used here is based on the RFC2045 spec, which is not the same as [java.util.Base64.getDecoder](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/util/Base64.java#L154) / [java.util.Base64.getEncoder](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/util/Base64.java#L89), which could cause issues for any existing data that we have encoded
> 
> I think you want [java.util.Base64.getMimeDecoder](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/util/Base64.java#L175) / [java.util.Base64.getMimeEncoder](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/util/Base64.java#L110) to remain backward-compatible with any existing data

I ran a test and this is the output:

original:cGxjc2RyZmdkdmhlYmFleXR2cXRwdGh3amtld2txbXF6emhsY3ZkdWhobmJvbmxxaWNlY3l2Z25pcmJpZWFhdmF3bmV4d3d6anlydWhycG91d2Rydm1kdW9qZ3Ria2Jk
java:cGxjc2RyZmdkdmhlYmFleXR2cXRwdGh3amtld2txbXF6emhsY3ZkdWhobmJvbmxxaWNlY3l2Z25pcmJpZWFhdmF3bmV4d3d6anlydWhycG91d2Rydm1kdW9qZ3Ria2Jk
mime:cGxjc2RyZmdkdmhlYmFleXR2cXRwdGh3amtld2txbXF6emhsY3ZkdWhobmJvbmxxaWNlY3l2Z25p
cmJpZWFhdmF3bmV4d3d6anlydWhycG91d2Rydm1kdW9qZ3Ria2Jk

the mime one split the text but the original apache and the java I replaced it with did not and have the same output so it should be ok.

warehouse/ingest-core/src/test/resources/datawave/ingest/mapreduce/job/all-splits.txt
```diff
@@ -1,11 +1,11 @@
-shard	[B@57c8b24d
-shard	[B@39320a41
-shard	[B@43f4ebd
-shard	[B@584a5c78
-shard	[B@79fcf790
-shard1	[B@79fcf790
+shard	B0E=
```
keith-ratcliffe: I'm probably just being overly paranoid. Sorry about that....

The thing that originally caught my attention was the fact that the data in these test files was changed, which seemed like a red flag, because, if we have to change existing data in order for it to be decoded successfully, then we could definitely have a problem, particularly if the data in question happens to already be on disk in Accumulo.

We had a similar situation a while back, where DW code was using Base64 encode/decode logic provided by a deprecated Infinispan dependency, and we found that we _did_ need to replace it with Java's mime encoder/decoder to ensure compatibility with existing data in Accumulo

That said, I just tried running some junits for these test files as they exist currently, and I apparently discovered the same thing you did....

I changed the decode logic in `NonShardedSplitsFile.Reader` to attempt both `java.util.Base64.getDecoder().decode` and `java.util.Base64.getMimeDecoder().decode` only to find that BOTH of those decoders fail with `java.lang.IllegalArgumentException` on these files.

So, taking a closer look at the encoded values in these files, my new question is: Why the heck doesn't commons-codec's Base64 choke on  them as well? AFAIK, they don't appear to be valid Base64 according to _either_ RFC spec. So, still kinda worrisome that there's that much difference in the behavior between commons-codec and Java

However, I do concede that it would be problematic, in this case specifically, for us to use to the mime encoder to generate the split files, since we rely on the table name and the encoded  value being on a single line. Large-ish values getting split across lines would definitely break our file parsing.

And it does seem like all of data in question here (in terms of the entire PR) is probably all ephemeral config-related stuff. Not 100% sure about that though

Still deserves some thorough integration testing just to be sure :-)

keith-ratcliffe (dismissed):

warehouse/ingest-core/src/test/resources/datawave/ingest/mapreduce/job/all-splits.txt
austin007008: yeah I'm definitely doing a bunch of testing lol. the reason the test data got changed (the issue was originally found be a dev who isn't with us anymore so hopefully I'm saying the issue right) is because the data in the original test files were actually memory locations from something and not actual base64 so it was giving bad input to the encoders/decoders. 

From what I can find/see, the big difference between apache and java base64 is that if you give apache bad data it will just gobble it up and return bad data to you were as java wont take bad input and will properly throw an error. I'm running into a similar problem downstream where a group of unit tests had been feeding the decoder non-base64 and now that it wont accept bad input, it's failing the test.

this is some high quality pair-programming :)

warehouse/core/src/main/java/datawave/ingest/data/config/ingest/AccumuloHelper.java
```diff
@@ -30,10 +32,17 @@ public class AccumuloHelper {
...
+
     public void setup(Configuration config) throws IllegalArgumentException {
         username = ConfigurationHelper.isNull(config, USERNAME, String.class);
-        byte[] pw = Base64.decodeBase64(ConfigurationHelper.isNull(config, PASSWORD, String.class).getBytes());
-        password = new PasswordToken(pw);
+        String encodedPW = ConfigurationHelper.isNull(config, PASSWORD, String.class);
+        if (encodedPW.length() >= 2) {
+            password = new PasswordToken(Base64.getDecoder().decode(encodedPW));
+        } else {
+            log.warn("no Accumulo password?");
```
ivakegg: only a warning?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/NonShardedSplitsFile.java
```diff
@@ -0,0 +1,206 @@
+package datawave.ingest.mapreduce.job;
```
ivakegg: What does adding this file have to do with replacing apache base64 ?

warehouse/core/src/main/java/datawave/ingest/data/config/ingest/AccumuloHelper.java
drewfarris: I don't think the check on `encodedPW.length()` is needed here. Just let the exception be thrown if the ConfigurationHelper returns garbage.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/NonShardedSplitsFile.java
austin007008: idk. it just says "Extracted from IngestJob" 

I know that as we were going through this code we realized that something was up with the splits so maybe it's related to that?

do you want me to remove this from the pr?

warehouse/core/src/main/java/datawave/ingest/data/config/ingest/AccumuloHelper.java
austin007008: changed

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/NonShardedSplitsFile.java
austin007008: removed

hlgp:
Reopen and address conflicts as priorities allow

PR #2482, opened 2024-07-29
Fixed a bug where the remote user operations cache always used the sa…

jwomeara:
…me key.

### DIFF
--- a/web-services/security/src/test/java/datawave/security/authorization/remote/RemoteUserOperationsImplHttpTest.java
+++ b/web-services/security/src/test/java/datawave/security/authorization/remote/RemoteUserOperationsImplHttpTest.java
@@ -1,6 +1,7 @@
 package datawave.security.authorization.remote;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
 
 import java.io.IOException;
 import java.math.BigInteger;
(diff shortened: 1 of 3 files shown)


ivakegg (approved):

ivakegg:
Excellent!

web-services/security/src/test/java/datawave/security/authorization/remote/RemoteUserOperationsImplHttpTest.java
```diff
@@ -142,15 +216,33 @@ public void after() {
...
+        // ensure that we get the cached user details
+        ProxiedUserDetails dupeReturnedUser = remote.getRemoteUser(principal);
+        assertEquals(returnedUser, dupeReturnedUser);
+
+        // setup the list effective auth response for the other user
+        setListEffectiveAuthResponse(otherUserDN, auths);
+
+        // ensure that we get the other user details, not the cached user details
+        ProxiedUserDetails newReturnedUser = remote.getRemoteUser(otherPrincipal);
+        assertNotEquals(returnedUser, newReturnedUser);
```
FineAndDandy, line 245: add an assert for the expected auths as well as that they are different from the cached auths?

FineAndDandy (approved):

web-services/security/src/test/java/datawave/security/authorization/remote/RemoteUserOperationsImplHttpTest.java
jwomeara: When you get a cache hit you literally get the exact same object (down to the reference) so i think this is sufficient..

PR #2487, opened 2024-07-31
Avoid useless object creation where possible by delegating to TypeFactory

apmoriarty:
This should not be merged until https://github.com/NationalSecurityAgency/datawave-type-utils/pull/35 is integrated into a tag of type-utils.

- stop creating a new Type instance for every field during event aggregation (including TLD aggregation)
- stop creating a new Type instance for every field during LimitFields operation (soon to be defunct given coming refactor, but including here as an example)
- added unit test to AttributeFactory enforce a new Type instance per attribute


alerman (dismissed):

warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java
```diff
@@ -64,6 +65,8 @@ public class LimitFields implements Function<Entry<Key,Document>,Entry<Key,Docum
     // should not be dropped
     private final Set<Set<String>> matchingFieldSets;
 
+    private final TypeFactory typeFactory = new TypeFactory();
```
ivakegg: configure size and ttl ?

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EventFieldAggregator.java
```diff
@@ -28,23 +25,16 @@
...
+
+    private final TypeMetadata typeMetadata;
+    private final TypeFactory typeFactory;
+    private final String defaultTypeClass;
 
     public EventFieldAggregator(String field, EventDataQueryFilter filter, int maxNextCount, TypeMetadata typeMetadata, String defaultTypeClass) {
         super(Collections.singleton(field), filter, maxNextCount);
 
         this.typeMetadata = typeMetadata;
+        this.typeFactory = new TypeFactory();
```
ivakegg: configure size and ttl?

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EventFieldAggregator.java
```diff
@@ -122,16 +112,48 @@ private Set<String> getNormalizedValues(String dataType, String fieldName, Strin
...
     }
+
+    /**
+     * Get the TypeFactory. If no TypeFactory exists one will be created. Configs for cache size and timeout may be configured.
+     *
+     * @return the TypeFactory
+     */
+    private TypeFactory getTypeFactory() {
+        if (typeFactory == null) {
+            if (typeCacheSize != -1 && typeCacheTimeoutMinutes != -1) {
```
alerman, line 131: What if one of the 2 is set?

apmoriarty: Then a type factory with default expiration values will be created.

avgAGB (approved):

ivakegg (approved):

Issue #2443, opened 2024-06-26
Remove uses of deprecated and non-public accumulo apis


ivakegg:
Se need to do a scrub of datawave and remove any uses of non-public and deprecated accumulo apis.  Note that for non-public apis, there may be situations where that cannot be done and we can subsequently request that accumulo make something public for us.

## Sub-Issues Progress

- [x] #3163 - Replace org.apache.accumulo.core.classloader.ClassLoaderUtil
- [x] #3176 - Replace Deprecated AbstractEventRecordReader
- [x] #3177 - Replace deprecated setDate(long)
- [x] #3178 - Replace deprecated AdvancedCacheLoader
- [x] #3179 - Replace deprecated KeyFilter
- [x] #3180 - Replace deprecated CacheLoaderTask
- [x] #3181 - Replace deprecated TaskContext
- [x] #3182 - Replace deprecated SingletonStoreConfiguration
- [x] #3184 - Replace deprecated InstanceOperations.getInstanceID()
- [x] #3185 - Replace deprecated TableId.getTableId()
- [x] #3187 - Replace deprecated Integer(int)
- [x] #3292 - Replace deprecated Float, Long, Double, Boolean constructors
- [x] #3293 - Replace deprecated Date(year, month, day) constructor
- [x] #3294 - Replace deprecated Job() constructor in MultiTableRRRangePartitionerTest
- [x] #3298 - Replace deprecated Byte, Character, Short constructors
- [x] #3300 - Replace deprecated Date(String) constructor
- [x] #3313 - Replace Charset.forName("UTF-8") with StandardCharsets.UTF_8
- [x] #3319 - Remove deprecated TabletLocator methods
- [x] #3335 - Fix incorrect WrappedConnector check in ScannerFactory
- [x] #3336 - Replace non-public IterationInterruptedException with class name check
- [x] #3337 - Remove unused SortedMapIterator import from ArrayIterator
- [x] #3339 - Replace ClientContext.getTableId with public API
- [x] #3341 - Replace non-public ScanTimedOutException import
- [x] #3343 - Remove deprecated TabletLocator usage
- [x] #3348 - Replace non-public BadArgumentException with IllegalArgumentException
- [x] #3358 - Replace non-public DateFormatSupplier
- [x] #3374 - Replace non-public ThreadPools
- [x] #3376 - Replace non-public CloneConfigurationImpl in test

[...]

PR #2490, opened 2024-08-01
Replace deprecated JexlArithmetic.compare()

SethSmucker:
The previous version used a String instead of a JexlOperator for the operator parameter. The following classes have been updated to use JexlArithmetic.compare(Object left, Object right, JexlOperator operator):

- datawave.query.jexl.DatawaveArithmetic
- datawave.query.jexl.DefaultArithmetic
- datawave.query.jexl.HitListArithmetic


avgAGB (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/HitListArithmetic.java
```diff
@@ -221,7 +222,7 @@ public boolean lessThan(final Object left, final Object right) {
             for (final Object o : set) {
                 Object normalizedO = ValueTuple.getNormalizedValue(o);
                 Object normalizedRight = ValueTuple.getNormalizedValue(right);
-                if (compare(normalizedO, normalizedRight, LESS_THAN) < 0) {
+                if (compare(normalizedO, normalizedRight, JexlOperator.LT) < 0) {
```
apmoriarty: I don't think the string constants on line 26 are needed anymore

warehouse/query-core/src/main/java/datawave/query/jexl/DefaultArithmetic.java
```diff
@@ -178,7 +179,7 @@ public boolean lessThan(Object left, Object right) {
             Set<Object> set = (Set<Object>) left;
 
             for (Object o : set) {
-                if (super.compare(o, right, LESS_THAN) < 0) {
+                if (super.compare(o, right, JexlOperator.LT) < 0) {
```
apmoriarty: I don't think the string constants on line 16 are needed anymore

apmoriarty (dismissed):

foster33 (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/HitListArithmetic.java
ivakegg: +1

warehouse/query-core/src/main/java/datawave/query/jexl/DefaultArithmetic.java
ivakegg: +1

ivakegg (changes requested):
Remove unused constants

warehouse/query-core/src/main/java/datawave/query/jexl/DefaultArithmetic.java
ivakegg: Still not resolved

warehouse/query-core/src/main/java/datawave/query/jexl/HitListArithmetic.java
ivakegg: still not resolved

ivakegg:
Why am I seeing changes to all of the docker files in the submodules in this PR.  Also the formatting is failing.

SethSmucker:
> Why am I seeing changes to all of the docker files in the submodules in this PR. Also the formatting is failing.

I think that they're from changes merged from integration, but I could be wrong. I've been working on the sub-modules in their own directories, so that could be it too.

PR #2491, opened 2024-08-01
Replace deprecated INSTANCE_DFS_URI/DIR with INSTANCE_VOLUMES

SethSmucker:
INSTANCE_DFS_DIR and INSTANCE_DFS_URI were deprecated, both being replaced with INSTANCE_VOLUMES


avgAGB (dismissed):

warehouse/query-core/src/main/java/datawave/mr/bulk/MultiRfileInputformat.java
```diff
@@ -152,8 +152,7 @@ public static List<InputSplit> computeSplitPoints(AccumuloClient client, Configu
 
             synchronized (MultiRfileInputformat.class) {
                 final InstanceOperations instOps = client.instanceOperations();
-                dfsUriMap.put(tableId, instOps.getSystemConfiguration().get(Property.INSTANCE_DFS_URI.getKey()));
-                dfsDirMap.put(tableId, instOps.getSystemConfiguration().get(Property.INSTANCE_DFS_DIR.getKey()));
+                dfsUriMap.put(tableId, instOps.getSystemConfiguration().get(Property.INSTANCE_VOLUMES.getKey()));
```
ivakegg, line 154: So if the dfsDirMap is no longer used, then we should update those other places where we check its contents.

ivakegg (changes requested):
see comment

ivakegg:
ok, now I need to actually test this out to make sure we are good.

ivakegg (approved):

apmoriarty (approved):

PR #2493, opened 2024-08-01
2416 fix maven build cache

cogross:
fixes build issue w/ cache
also needs https://github.com/NationalSecurityAgency/datawave-mapreduce-query-service/pull/6

### DIFF
--- a/.mvn/maven-build-cache-config.xml
+++ b/.mvn/maven-build-cache-config.xml
@@ -35,7 +35,7 @@
     <input>
         <global>
             <glob>
-                {*.java,*.sh,*.py,*.properties}
+                {*.java,*.sh,*.py,*.properties,*.yml,*.xml}
             </glob>
             <includes>
                 <include>src/</include>
--- a/contrib/datawave-quickstart/docker/pom.xml
+++ b/contrib/datawave-quickstart/docker/pom.xml
@@ -17,6 +17,8 @@
         <dist.maven>apache-maven-3.8.8-bin.tar.gz</dist.maven>
         <dist.wildfly>wildfly-17.0.1.Final.tar.gz</dist.wildfly>
         <dist.zookeeper>apache-zookeeper-3.7.2-bin.tar.gz</dist.zookeeper>
+        <docker.image.prefix />
+        <maven.build.cache.enabled>false</maven.build.cache.enabled>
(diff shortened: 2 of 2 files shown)


avgAGB (dismissed):

jwomeara:
OK, so I tested this out a bit today.  I ran into some issues...  Overall I think these are improvements to what we currently have on integration, but I think there's still more work to do to get this working consistently.

here's what I did.

```
# From the base datawave project
mvn -Pdev clean install -DskipTests

# Now that we have cached artifacts, run the full docker build (with the compose profile instead of dev)
mvn -Pcompose -Dmicroservice-docker -Pquickstart-docker -Ddeploy -Dtar -Ddist clean install -DskipTests

# once that finished, i started docker compose, and ran the test scripts.  everything worked fine
cd docker
docker compose up -d
cd scripts
./testAll.sh

# after that I made a change to QueryController.java in the query service.  i added a new rest endpoint.  you can see that below.

# i also made a change to QueryLogicFactory.xml in the web-services/deploy module.  you can find that change below as well.

# after that, I built the full project again with all of the microservice docker images (from the base datawave project directory)
mvn -Pcompose -Dmicroservice-docker -Pquickstart-docker -Ddeploy -Dtar -Ddist clean install -DskipTests

# doing that I saw a lot of errors when it got to the docker quickstart portion of the build.  those errors didn't fail the build, but they did result in a broken quickstart image.  you can verify this by attempting to start docker compose again.  the services will fail to start, and if you look at the quickstart logs you will see that it is attempting to build datawave (which is indicative of an error)

```

Here's what I added to QueryController.java
```
@RequestMapping(path = "blah", method = {RequestMethod.POST}, produces = {"application/xml", "text/xml", "application/json", "text/yaml", "text/x-yaml", "application/x-yaml", "application/x-protobuf", "application/x-protostuff"})
public String blah(@AuthenticationPrincipal DatawaveUserDetails currentUser) throws QueryException {
   return "blah";
}
```

Here's what I added for QueryLogicFactory.xml
```
<bean id="totallyNewResponseObjectFactory" class="datawave.core.query.result.event.DefaultResponseObjectFactory" />
```

jwomeara (changes requested):
see comments

PR #2499, opened 2024-08-06
Update deprecated usages of SequenceFile.createWriter()

lbschanno:
Update several deprecated calls to SequenceFile.createWriter() to use SequenceFile.createWriter(Configuration conf,
SequenceFile.Writer.Option... opts) instead.

Part of #2443


apmoriarty (dismissed):

avgAGB (dismissed):

ivakegg (dismissed):

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/MultiRFileOutputFormatter.java
```diff
@@ -594,7 +595,9 @@ public void close(TaskAttemptContext context) throws IOException, InterruptedExc
                 if (generateMapFileRowKeys && !shardMapFileRowKeys.isEmpty()) {
                     log.info("Writing mapFileRowKeys");
                     Path shardMapFilePath = new Path(workDir, getUniqueFile(context, "mapFileRowKeys", ".lst"));
-                    try (SequenceFile.Writer output = SequenceFile.createWriter(fs, conf, shardMapFilePath, Text.class, Text.class)) {
+                    Path qualifiedShardMapFilePath = fs.makeQualified(shardMapFilePath);
```
ivakegg: Why call fs.makeQualified?  I was looking through the deprecated code of SequenceFile.createWriter and I do not see that being called?  Perhaps I missed something however.

lbschanno: If I recall correctly, I believe I did it following the example of other calls to `SequenceFile.createWriter(Configuration conf, Writer.Option... opts)`, but you're correct that it is not called in the deprecated code of `SequenceFile.createWriter()`. I have removed the unnecessary calls to `fs.makeQualified()`.

ivakegg (approved):

apmoriarty (approved):

drewfarris (approved):

alerman (approved):

PR #2505, opened 2024-08-07
Update workflow to checkout branch

alerman:


foster33 (approved):
LGTM. I suspect this might not work with PRs based off of a fork, but that probably isn't that big of a deal.

keith-ratcliffe (approved):

PR #2508, opened 2024-08-09
Hazelcast config for multiple clusters

doublejai:
Configuration for creating an additional hazelcast cluster: Follow the example in the docker-compose.yml in this branch, where a new hazelcast cluster is created after the default one.
Make sure "- --spring.application.name=" is added under the "command" header in the additional cluster's config.

Configuration for the service: Point a service to a specifec hazelcast cluster by adding
hazelcast:
clusterName: ''
client:
enabled: true
clusterName: ''
(example of this can be seen in the query.yml and executor.yml files in this branch)

### DIFF
--- a/docker/config/executor.yml
+++ b/docker/config/executor.yml
@@ -65,3 +65,8 @@ datawave:
       logStatusPeriodMs: 600000
       logStatusWhenChangedMs: 300000
       queryMetricsUrlPrefix: https://localhost:8543/querymetric/v1/id/
+
+hazelcast:
+  client:
+    clusterName: cache2
+  clusterName: cache2
--- a/docker/config/executor.yml
+++ b/docker/config/executor.yml
@@ -65,8 +65,3 @@ datawave:
       logStatusPeriodMs: 600000
       logStatusWhenChangedMs: 300000
       queryMetricsUrlPrefix: https://localhost:8543/querymetric/v1/id/
-
-hazelcast:
-  client:
-    clusterName: cache2
-  clusterName: cache2
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -612,9 +612,9 @@ services:
   #    - ns datawaveUsers
   #    - m.entries
   management-center:
-#    profiles:
-#      - management
-#      - full
+    profiles:
+      - management
+      - full
     image: docker.io/hazelcast/management-center:5.1.2
     environment:
       - |-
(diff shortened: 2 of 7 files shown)


docker/config/executor.yml
```diff
@@ -65,3 +65,8 @@ datawave:
       logStatusPeriodMs: 600000
       logStatusWhenChangedMs: 300000
       queryMetricsUrlPrefix: https://localhost:8543/querymetric/v1/id/
+
+hazelcast:
```
jwomeara: Can we move this configuration to application-query.yml.  Then you can remove it from both executor.yml and query.yml because both the query and executor service depend on application-query.yml.

Can you also make the cluster name an environment variable (defaulting to the shared cache) so that we can decide which cache to use at docker compose runtime?

For instance, instead of what you have, can you change it to this?
```
hazelcast:
  client:
    clusterName: ${QUERY_CACHE:cache}
  clusterName: ${QUERY_CACHE:cache}
```

That way we can use the original cache with this command:
`docker compose up -d`

And the alternate cache with this command:
`QUERY_CACHE=cache2 docker compose up -d`

docker/docker-compose.yml
```diff
@@ -173,6 +173,35 @@ services:
       configuration:
         condition: service_started
 
+  cache2:
```
jwomeara: can you rename this to query-cache?

Can you also add a profile for this, so that it won't be run by default?  We don't want to run this way by default because the github actions which run these services have limited resources.

something like this would work:
```
query-cache:
    profiles:
        - querycache
```

jwomeara (changes requested):

jwomeara (changes requested):
I was really hoping for some simple docs to describe what changes are necessary in order to use a separate hazelcast cache for query.  Could you find a place to add that information as well?

docker/config/application-querycache.yml
jwomeara, line 1: You can delete this.  It's not needed.

docker/docker-compose.yml
```diff
@@ -576,9 +612,9 @@ services:
   #    - ns datawaveUsers
   #    - m.entries
   management-center:
-    profiles:
-      - management
-      - full
+#    profiles:
+#      - management
+#      - full
```
jwomeara: You probably commented these while testing, but they should stay uncommented for your PR.

jwomeara (changes requested):
Just a few minor requests, but otherwise I think this is good to go.  Let's try and get it merged today.

jwomeara (approved):

PR #2509, opened 2024-08-09
fixes #2507: Changed to use a more specific exception for query itera…

### LINKED ISSUE #2507: IterationInterruptedException not being passed up
Dave
- I've been looking into the "zombie" scan threads that were causing tablets to not unload. 
Somebody grabbed some stack traces and found a lot of zombie scans stuck in the same place. He provided me with the following links that correspond to lines from the stack traces that he has:
https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/query-core/src/main/java/datawave/query/function/LimitFields.java#L299
https://github.com/NationalSecurityAgency/datawave-type-utils/blob/main/src/main/java/datawave/data/type/GeometryType.java
https://github.com/NationalSecurityAgency/datawave-type-utils/blob/main/src/main/java/datawave/data/normalizer/GeometryNormalizer.java
The stack traces that he has, and the links above, just point to where the zombie scan threads are. I found the following, which seems concerning.
[https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/quer[…]y-core/src/main/java/datawave/query/iterator/QueryIterator.java](https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java#L533-L542)
Specifically, the fact that IterationInterruptedException is ignored. It appears that it's ignored if the scan has yielded. He said you might have more information as to why this code is ignoring that exception in this case. Looking at the Accumulo code, IterationInterruptedException is only raised when the interrupt flag is checked. I can't find a case where it's raised when the iterator has yielded. Based on the current logic in QueryIterator.handleException, any scan that has yielded and where the client has gone away, an exception will not be thrown, but the ScanSession will be cleaned up leaving the thread orphaned.

Dave
Related to this issue, I created https://github.com/apache/accumulo/pull/4761 this morning to possibly address an edge case, but I'm concerned that this could be ignored as well.

Ivarator

I will have to dig into the code to remember the whole yielding thing.  WRT the limit fields mechanism, Charlie is working on a fix for that (we had a meeting of the minds on that one).

Ivarator
Are you saying that the exception is ignored in the datawave code, or the accumulo code?

Daniel Roberts
I believe he was referring to the datawave code. Specifically this section:
https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/quer[…]y-core/src/main/java/datawave/query/iterator/QueryIterator.java

Ivarator
ah, I see..... ok I will need to dig up some neurons from 10 years ago to figure this out.

Dave
 - Thanks, and the implementation too: [https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/quer[…]y-core/src/main/java/datawave/query/iterator/QueryIterator.java](https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java#L533-L542)
I looked in the 1.10 code and I'm not seeing IterationInterruptedException being used for yielding either. Maybe it was used in that manner in an earlier version.

Dave
 - I think I found the problem. The DataWave code is raising the Accumulo exception for yielding at:
https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/quer[…]ain/java/datawave/query/iterator/pipeline/PipelineIterator.java
[https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/quer[…]ain/java/datawave/query/iterator/pipeline/PipelineIterator.java](https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/query-core/src/main/java/datawave/query/iterator/pipeline/PipelineIterator.java#L187)
The DataWave code is also raising the exception in SortedMultiMapIterator, but it's doing it in the same manner that the Accumulo code is doing it.
Suggest raising a different exception in PipelineIterator and handling that in QueryIterator.handleException.

Dave
If changing this doesn't resolve the "zombie" scan problem, we have at least eliminated it as a possibility.

Ivarator
ah, good find.  Easy enough to change the exception

ivakegg:
This PR will ensure that the IteratorInterruptionException is always thrown all the way up, even if we have yielded.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/core/iterators/CompositeSeekingIterator.java
+++ b/warehouse/query-core/src/main/java/datawave/core/iterators/CompositeSeekingIterator.java
@@ -76,7 +76,7 @@ public class CompositeSeekingIterator extends WrappingIterator {
                 try {
                     type = Class.forName(typeClass).asSubclass(DiscreteIndexType.class).getDeclaredConstructor().newInstance();
                 } catch (ClassNotFoundException | NoSuchMethodException | InstantiationException | IllegalAccessException | InvocationTargetException e) {
-                    log.warn("Unable to create DiscreteIndexType for class name: " + typeClass);
+                    log.warn("Unable to create DiscreteIndexType for class name: " + typeClass, e);
                 }
             }
 
--- a/warehouse/query-core/src/main/java/datawave/core/iterators/CompositeSeekingIterator.java
+++ b/warehouse/query-core/src/main/java/datawave/core/iterators/CompositeSeekingIterator.java
@@ -1,6 +1,7 @@
 package datawave.core.iterators;
 
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -74,7 +75,7 @@ public class CompositeSeekingIterator extends WrappingIterator {
             if (typeClass != null) {
                 try {
                     type = Class.forName(typeClass).asSubclass(DiscreteIndexType.class).getDeclaredConstructor().newInstance();
-                } catch (Exception e) {
+                } catch (ClassNotFoundException | NoSuchMethodException | InstantiationException | IllegalAccessException | InvocationTargetException e) {
                     log.warn("Unable to create DiscreteIndexType for class name: " + typeClass);
                 }
             }
--- a/warehouse/query-core/src/main/java/datawave/core/iterators/filesystem/FileSystemCache.java
+++ b/warehouse/query-core/src/main/java/datawave/core/iterators/filesystem/FileSystemCache.java
@@ -43,7 +43,7 @@ public class FileSystemCache {
         try {
             return new URI(scheme, authority, "/", null, null);
         } catch (URISyntaxException e) {
-            throw new IllegalArgumentException("Failed to create URI with only " + scheme + " and " + authority);
+            throw new IllegalArgumentException("Failed to create URI with only " + scheme + " and " + authority, e);
         }
     }
 
--- a/warehouse/query-core/src/main/java/datawave/core/iterators/filesystem/FileSystemCache.java
+++ b/warehouse/query-core/src/main/java/datawave/core/iterators/filesystem/FileSystemCache.java
@@ -3,6 +3,7 @@ package datawave.core.iterators.filesystem;
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.concurrent.TimeUnit;
 
@@ -41,7 +42,7 @@ public class FileSystemCache {
         }
         try {
             return new URI(scheme, authority, "/", null, null);
-        } catch (Exception e) {
+        } catch (URISyntaxException e) {
             throw new IllegalArgumentException("Failed to create URI with only " + scheme + " and " + authority);
         }
     }
--- a/warehouse/query-core/src/main/java/datawave/query/common/grouping/AverageAggregator.java
+++ b/warehouse/query-core/src/main/java/datawave/query/common/grouping/AverageAggregator.java
@@ -120,11 +120,7 @@ public class AverageAggregator extends AbstractAggregator<BigDecimal> {
     @Override
     public void aggregate(Attribute<?> value) {
         BigDecimal number;
-        try {
-            number = new BigDecimal(value.getData().toString());
-        } catch (Exception e) {
-            throw new IllegalArgumentException("Unable to calculate an average with non-numerical value '" + value.getData() + "'", e);
-        }
+        number = new BigDecimal(value.getData().toString());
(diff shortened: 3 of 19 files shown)


warehouse/query-core/src/main/java/datawave/query/tables/shard/CountAggregatingIterator.java
```diff
@@ -62,12 +62,7 @@ public boolean hasNext() {
                 ResultCountingIterator.ResultCountTuple tuple = unpackValue(entry.getValue());
 
                 // Merge the columnVisibilities
-                try {
-                    this.columnVisibilities.add(tuple.getVisibility());
```
ivakegg, line 66: no parsing is actually done here.... in fact no exception could really be thrown here.

dlmarion (commented):
LGTM

warehouse/query-core/src/main/java/datawave/core/iterators/CompositeSeekingIterator.java
```diff
@@ -74,7 +75,7 @@ public void init(SortedKeyValueIterator<Key,Value> source, Map<String,String> op
             if (typeClass != null) {
                 try {
                     type = Class.forName(typeClass).asSubclass(DiscreteIndexType.class).getDeclaredConstructor().newInstance();
-                } catch (Exception e) {
+                } catch (ClassNotFoundException | NoSuchMethodException | InstantiationException | IllegalAccessException | InvocationTargetException e) {
                     log.warn("Unable to create DiscreteIndexType for class name: " + typeClass);
```
alerman: Is it worth logging out e or e's message here?

warehouse/query-core/src/main/java/datawave/core/iterators/filesystem/FileSystemCache.java
```diff
@@ -41,7 +42,7 @@ private URI getSchemeAndAuthority(URI hdfsBaseURI) {
         }
         try {
             return new URI(scheme, authority, "/", null, null);
-        } catch (Exception e) {
+        } catch (URISyntaxException e) {
             throw new IllegalArgumentException("Failed to create URI with only " + scheme + " and " + authority);
```
alerman:
```suggestion
            throw new IllegalArgumentException("Failed to create URI with only " + scheme + " and " + authority, e);
```

warehouse/query-core/src/main/java/datawave/query/common/grouping/AverageAggregator.java
```diff
@@ -120,11 +120,7 @@ public BigDecimal getDivisor() {
     @Override
     public void aggregate(Attribute<?> value) {
         BigDecimal number;
-        try {
-            number = new BigDecimal(value.getData().toString());
-        } catch (Exception e) {
-            throw new IllegalArgumentException("Unable to calculate an average with non-numerical value '" + value.getData() + "'", e);
-        }
+        number = new BigDecimal(value.getData().toString());
```
alerman, line 123: Do we want/need to catch the NumberFormatException that can be thrown here?

warehouse/query-core/src/main/java/datawave/query/tables/stats/ScanSessionStats.java
```diff
@@ -86,8 +86,8 @@ public void stopOnFailure() {
         for (TIMERS timer : mergedTimers.keySet()) {
             try {
                 timers.get(timer).stop();
-            } catch (Exception e) {
-
+            } catch (IllegalStateException e) {
+                // thrown when timer was not yet running.... we can ignore that
```
alerman, line 90: is it worth logging anything at debug?

alerman (changes requested):

warehouse/query-core/src/main/java/datawave/core/iterators/CompositeSeekingIterator.java
ivakegg: +1

warehouse/query-core/src/main/java/datawave/query/common/grouping/AverageAggregator.java
ivakegg: Why not simply let it get passed up?  We were only turning it into an IllegalArgumentException anyway...

warehouse/query-core/src/main/java/datawave/query/tables/stats/ScanSessionStats.java
ivakegg: Not in this case.

warehouse/query-core/src/main/java/datawave/query/common/grouping/AverageAggregator.java
alerman: As long as something above is catching it that's fine, just wanted to make sure it had been considered

alerman (approved):

apmoriarty (approved):

FineAndDandy (approved):

PR #2511, opened 2024-08-13
Use generic function descriptors in all cases except content functions

apmoriarty:


alerman (approved):

avgAGB (approved):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypeVisitor.java
```diff
@@ -394,32 +388,17 @@ public Set<String> getIngestTypesForFunction(ASTFunctionNode node) {
...
-                return queryDescriptor.fields(null, Collections.emptySet());
-            default:
-                // do nothing
-                log.warn("Unhandled function namespace: " + visitor.namespace());
+        if (visitor.namespace().equals(CONTENT_FUNCTION_NAMESPACE)) {
+            // all content function fields are added
+            ContentFunctionsDescriptor.ContentJexlArgumentDescriptor contentDescriptor = new ContentFunctionsDescriptor().getArgumentDescriptor(node);
+            return contentDescriptor.fieldsAndTerms(Collections.emptySet(), Collections.emptySet(), Collections.emptySet(), null)[0];
+        } else {
+            JexlArgumentDescriptor descriptor = JexlFunctionArgumentDescriptorFactory.F.getArgumentDescriptor(node);
```
ivakegg, line 396: Huh?   What happened to all of the other functions descriptors?  Why would you get rid of those?

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IngestTypeVisitor.java
apmoriarty: That method call is smart enough to select the correct functions descriptor based on the function namespace and name. 

If the query planner were ever configured with an function descriptor that extended any of the previous descriptors, then this visitor would fail to function correctly.

ivakegg: Oh, I see, this is using the argument descriptor factory.... totally missed that. We are good.

ivakegg (approved):

PR #2512, opened 2024-08-13
Suggestions for term frequency excerpt iterator review

drewfarris:


warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
```diff
@@ -276,66 +323,483 @@ public void next() throws IOException {
...
+
+        // there are cases where we'll have no scores and we will want to choose a longer term for a particular position
+        // instead of the hit term we matched (e.g., when we're dealing with synonyms that are a fragment of a larger
+        // term.) This method will set an override for that position so that we'll choose the longer term instead of
+        // a shorter synonym, which is typically a substring of the longer term.
+        overrideOutputLongest(wordsAndScoresArr);
+
+        // create an array with the same length as the one we just passed in
+        String[] termsToOutput = new String[wordsAndScoresArr.length];
+        // TODO: what about both?
```
austin007008: "both" is the default so if a user does not specifically specify before/after, it will assume "both"

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
```diff
@@ -38,24 +38,111 @@ public class TermFrequencyExcerptIterator implements SortedKeyValueIterator<Key,
     private static final Collection<ByteSequence> TERM_FREQUENCY_COLUMN_FAMILY_BYTE_SEQUENCE = Collections
                     .singleton(new ArrayByteSequence(Constants.TERM_FREQUENCY_COLUMN_FAMILY.getBytes()));
 
-    /** The field name option */
-    public static final String FIELD_NAME = "field.name";
-    /** The start offset option */
-    public static final String START_OFFSET = "start.offset";
-    /** The end offset option */
-    public static final String END_OFFSET = "end.offset";
```
drewfarris, line 45: Moved these to the configuration class.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
```diff
@@ -38,24 +38,111 @@ public class TermFrequencyExcerptIterator implements SortedKeyValueIterator<Key,
...
+    }
 
     /** the underlying source */
     protected SortedKeyValueIterator<Key,Value> source;
-    /** the field name */
-    protected String fieldName;
-    /** the start offset (inclusive) */
-    protected int startOffset;
-    /** the end offset (exclusive) */
-    protected int endOffset;
```
drewfarris, line 54: moved these to Configuration class

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
```diff
@@ -302,11 +346,13 @@ public void next() throws IOException {
...
+        final int startOffset = config.startOffset;
+        final int endOffset = config.endOffset;
+        final List<String> hitTermsList = config.hitTermsList;
+        final String fieldName = config.fieldName;
+
         // get the pieces from the top key that will be returned
         Key top = source.getTopKey();
-        Text cv = top.getColumnVisibility();
-        long ts = top.getTimestamp();
-        Text row = top.getRow();
```
drewfarris, line 266: This as a little dangerous because `top` can change between here and where these values were used. We can just call the methods on `Key` to get these values when we need them.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
```diff
@@ -371,8 +417,8 @@ public void next() throws IOException {
         }
         // Now that the words and scores array is populated with all the tf data, it's time to generate an excerpt and
         // create a key that contains all of our excerpts to be read by the ExcerptTransform
-        final List<Object> carePackage = List.of(hitTermsList, direction, trimExcerpt, startOffset, endOffset, origHalfSize);
```
drewfarris: Replaced with the Configuration class.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
```diff
@@ -432,14 +478,14 @@ private static String generateExcerpt(WordsAndScores[] wordsAndScoresArr, final
         }
 
         // Generate the "phrase without scores" excerpt now that the scores flags are false.
-        String phraseWithoutScoresExcerpt = generatePhrase(wordsAndScoresArr, carePackage);
+        String phraseWithoutScoresExcerpt = generatePhrase(wordsAndScoresArr, config);
 
         // if the regular excerpt is blank, we will return a message saying that the excerpt was blank
         if (phraseWithoutScoresExcerpt.isBlank()) {
-            phraseWithoutScoresExcerpt = BLANKEXCERPT;
```
drewfarris: Slightly better variable name

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
```diff
@@ -902,24 +948,24 @@ private static String getDtUid(String str) {
     }
 
     public void setHitTermsList(List<String> hitTermsList) {
-        this.hitTermsList = (ArrayList<String>) hitTermsList;
+        this.config.hitTermsList = (ArrayList<String>) hitTermsList;
```
drewfarris, line 951: it's safe to to call `this.config` without a null check here b/c the `config` instance is _always_ created and _final_ so it can't be replaced inadvertently with a null.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -337,84 +347,88 @@ private Set<Excerpt> getExcerpts(PhraseIndexes phraseIndexes) {
      *            the term values to match
      * @return the excerpt
      */
-    private String getExcerpt(String field, int start, int end, Range range, ArrayList<String> hitTermValues) {
+    private Excerpt getExcerpt(String field, int start, int end, Range range, ArrayList<String> hitTermValues) {
```
drewfarris, line 350: Use the `Excerpt` class which was created for this purpose instead of splitting a returned string elsewhere to get three values.

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -337,84 +347,88 @@ private Set<Excerpt> getExcerpts(PhraseIndexes phraseIndexes) {
...
-            if (log.isDebugEnabled() && attempt == 1) {
-                log.debug("size of excerpt requested: {}", excerptFields.getOffset(field) * 2);
-                log.debug("original range is ({},{}) and the expanded range is ({},{})", start, end,
-                                excerptIteratorOptions.get(TermFrequencyExcerptIterator.START_OFFSET),
-                                excerptIteratorOptions.get(TermFrequencyExcerptIterator.END_OFFSET));
+
+                // if this is the second attempt, we want the iterator to trim the excerpt down to the size we want.
+                // (remember we run the iterator with an expanded range the second time, so we can potentially have a
+                // bigger excerpt than needed even after removing stop words)
+                excerptIterator.setTrimExcerpt(true);
```
austin007008, line 386: I think this has to be set to true after we initialize the iterator. If we set it before the initialization, it will be overridden with "false" during initialization

PR #2515, opened 2024-08-14
Improve query sort options and implementations

apmoriarty:
- consolidate and rename variables that control sorting query pre and post global index lookup
- expanded test cases for both pre and post index sorting
- reworked sort implementations


avgAGB (dismissed):

alerman (dismissed):

jwomeara (changes requested):
From a serialization perspective, your changes to ShardQueryConfiguration are not backwards compatible.  We serialize the ShardQueryConfiguration to hazelcast for the query microservices, so I would avoid making changes like this (ones that have no impact on functionality) in the future as it is likely to break query microservices.  However, since you made a change, please update ShardQueryConfiguration's serialVersionUID.  If you leave that unchanged we will run into serialization errors in the query microservices.

jwomeara (dismissed):
with the addition of the serialVersionUID update, this should be ok for query microservices.  not sure what renaming all of these buys us though...

ivakegg (dismissed):

avgAGB (dismissed):

ivakegg:
I had previously approved this, but since then this PR is showing a heck of a lot more changes than just the variable renaming.  What happened here?

apmoriarty:
> I had previously approved this, but since then this PR is showing a heck of a lot more changes than just the variable renaming. What happened here?

Someone requested additional tests which exposed several cases where sorting was not behavior according to expectation.

avgAGB (dismissed):

ivakegg:
> > I had previously approved this, but since then this PR is showing a heck of a lot more changes than just the variable renaming. What happened here?
> 
> Someone requested additional tests which exposed several cases where sorting was not behavior according to expectation.

Well you also added additional sorting mechanisms as well, not just test cases.

apmoriarty:
> Well you also added additional sorting mechanisms as well, not just test cases.

I have updated the title and description to reflect that this was not just an update to some test cases

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -2763,8 +2763,10 @@ public Tuple2<CloseableIterable<QueryPlan>,Boolean> getQueryRanges(ScannerFactor
             }
         }
 
-        if (config.isSortQueryBeforeGlobalIndex()) {
+        if (config.isSortQueryPreIndexWithFieldCounts()) {
             config.setQueryTree(timedSortQueryBeforeGlobalIndex(config, getMetadataHelper()));
+        } else if (config.isSortQueryPreIndexWithImpliedCounts()) {
+            config.setQueryTree(OrderByCostVisitor.order(config.getQueryTree()));
```
ivakegg: Should this second ordering option be timed as well?

apmoriarty: Yes, yes it should. Thanks!

avgAGB (dismissed):

avgAGB (approved):

alerman (approved):

PR #2520, opened 2024-08-15
Replace nonpublic Accumulo Threads with new Datawave equivalent

SethSmucker:
`org.apache.accumulo.core.util.threads.Threads` is not part of Accumulo's public API. The variable `UEH` is the only part of the class used in Datawave.

Created `datawave.util.threads.DatawaveUncaughtExceptionHandler` and `datawave.util.threads.Threads` which are similar to the Accumulo classes.

Replace all references of Accumulo's `Threads` and `AccumuloUncaughtExceptionHandler` with the Datawave classes.


SethSmucker:
This is the bare minimum to move to Datawave. If you'd like me to include the same functionality as the Accumulo version I can do that.

warehouse/core/src/main/java/datawave/util/threads/DatawaveUncaughtExceptionHandler.java
```diff
@@ -0,0 +1,55 @@
...
+    static boolean isError(Throwable t) {
+        return isError(t, 0);
+    }
+
+    @Override
+    public void uncaughtException(Thread t, Throwable e) {
+        if (isError(e)) {
+            try {
+                e.printStackTrace();
+                System.err.println("Error thrown in thread: " + t + ", halting VM.");
```
avgAGB: Just curious, would it be worthwhile to print this to LOG.error as well to ensure it gets picked up in logging?

warehouse/core/src/main/java/datawave/util/threads/Threads.java
```diff
@@ -0,0 +1,9 @@
+package datawave.util.threads;
+
+import java.lang.Thread.UncaughtExceptionHandler;
+
+public class Threads {
```
jschmidt10: I would ditch this Threads class and just use `new DatawaveUncaughtExceptionHandler()` in BulkInputFormat. We don't need this wrapper.

warehouse/core/src/main/java/datawave/util/threads/DatawaveUncaughtExceptionHandler.java
jschmidt10: Agree! Just to be explicit - We should drop the printStackTrace() and the System.err and use a `LOG.error("some message", e)`. That'll give us the message and the full stacktrace and let us configure where/if the logging occurs.

warehouse/core/src/main/java/datawave/util/threads/Threads.java
SethSmucker: Sounds good! Wasn't sure if we should include it in case it's used down the line, but that's probably unnecessary future proofing.

alerman (dismissed):

avgAGB (dismissed):

jschmidt10 (dismissed):

warehouse/core/src/main/java/datawave/util/threads/DatawaveUncaughtExceptionHandler.java
```diff
@@ -0,0 +1,83 @@
...
+    @Override
+    public void uncaughtException(Thread t, Throwable e) {
+        if (isError(e)) {
+            try {
+                LOG.error("Error thrown in thread: {}, halting VM.", t, e);
+            } catch (Throwable e1) {
+                // If e == OutOfMemoryError, then it's probably that another Error might be
+                // thrown when trying to print to System.err.
+            } finally {
+                Runtime.getRuntime().halt(-1);
```
ivakegg: Actually, I really do not like halting.  I realize the accumulo one did this, but in the context of a BulkInputFormat, we certainly do not want to halt the vm which would most likely be a yarn container.
Instead, lets save off the exception, and then add finaly blocks above to check if exceptions were thrown so that we can pass them back up.

ivakegg (changes requested):
see comment

warehouse/core/src/main/java/datawave/util/threads/DatawaveUncaughtExceptionHandler.java
```diff
@@ -0,0 +1,86 @@
...
+     * Virtual Machine.
+     *
+     * @param t
+     *            the thread
+     * @param e
+     *            the exception
+     */
+    @Override
+    public void uncaughtException(Thread t, Throwable e) {
+        Throwable savedException = null;
```
ivakegg: SOrry, I was not clear.  This savedException should be a member of this class.  In the BulkInputFormat we should check whether that member has an exception and subsequently pass the exception up.  I don't think throwing an exception in this method would actually be passed up as it would be in a separate thread.

ivakegg:
Also why all of the changes to the docker files in this PR ?

Issue #2530, opened 2024-08-23
LimitFields.isHit() should not normalize


FineAndDandy:
The attribute can be passed along with the Content in the JexlEvaluation so that the isHit() check does not need to normalize, but instead can just compare against the original attribute.

PR #2531, opened 2024-08-23
Use Attribute comparisons in LimitFields.isHit() (#2530)

FineAndDandy:


warehouse/query-core/src/main/java/datawave/query/attributes/Content.java
```diff
@@ -117,13 +145,31 @@ public void write(Kryo kryo, Output output, Boolean reducedResponse) {
...
+            Class sourceClass;
+            try {
+                sourceClass = Class.forName(clazz);
+                source = (Attribute<?>) sourceClass.newInstance();
+            } catch (ClassNotFoundException | InstantiationException | IllegalAccessException e) {
+                throw new RuntimeException("could not parse source", e);
+            }
+
+            source.read(kryo, input);
+        }
```
apmoriarty, line 172: This looks good. Can you update the ContentTest to include variants with a source attribute?

PR #2423, opened 2024-06-18
Replace usage of Accummulo Pair class

alerman:

PR #2534, opened 2024-08-26
Replace nonpublic Accumulo ColumnSet

SethSmucker:
ColumnSet is not part of Accumulo's public API. encode/decodeColumns are the only part of the class used in Datawave.

Replace the references of to Accumulo's ColumnSet with code that functions the same without the reference.

### DIFF
--- a/warehouse/core/src/main/java/datawave/util/ColumnSetUtil.java
+++ b/warehouse/core/src/main/java/datawave/util/ColumnSetUtil.java
@@ -1,6 +1,7 @@
 package datawave.util;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
+
 import org.apache.accumulo.core.util.Pair;
 import org.apache.hadoop.io.Text;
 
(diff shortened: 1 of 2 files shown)


avgAGB (dismissed):

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/reduce/AggregatingReducer.java
```diff
@@ -39,6 +38,7 @@
 import com.google.common.collect.Sets;
 
 import datawave.ingest.mapreduce.job.TableConfigurationUtil;
+import datawave.util.ColumnSetUtil;
 
 @SuppressWarnings("deprecation")
```
ivakegg: Can we remove the suppress warnings?

ivakegg (dismissed):

avgAGB (dismissed):

ivakegg (approved):

apmoriarty (approved):

warehouse/core/src/main/java/datawave/util/ColumnSetUtil.java
```diff
@@ -0,0 +1,93 @@
...
+        encode(sb, columnFamily);
+        if (columnQualifier != null) {
+            sb.append(':');
+            encode(sb, columnQualifier);
+        }
+
+        return sb.toString();
+    }
+
+    public static Pair<Text,Text> decodeColumns(String columns) {
```
ivakegg, line 27: Lets also avoid using Pair here (part of private api).  Use AbstractMap.SimpleEntry instead please

ivakegg (changes requested):
One more update required.

ivakegg:
Actualy, this is already handled in #2423 so I am going to close this one.  My apologies.

PR #2536, opened 2024-08-26
Updated the composite query plan to be parseable as Jexl

ivakegg:


avgAGB (approved):

apmoriarty (approved):
It would be cool to see a small test that calls the /plan endpoint

PR #2537, opened 2024-08-26
2535 disable maven build cache

cogross:
Disables build cache by default. Maven build cache causes problems for quickstart docker images.

### DIFF
--- a/contrib/datawave-quickstart/docker/pom.xml
+++ b/contrib/datawave-quickstart/docker/pom.xml
@@ -17,6 +17,7 @@
         <dist.maven>apache-maven-3.8.8-bin.tar.gz</dist.maven>
         <dist.wildfly>wildfly-17.0.1.Final.tar.gz</dist.wildfly>
         <dist.zookeeper>apache-zookeeper-3.7.2-bin.tar.gz</dist.zookeeper>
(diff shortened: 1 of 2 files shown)


contrib/datawave-quickstart/docker/pom.xml
```diff
@@ -17,6 +17,7 @@
         <dist.maven>apache-maven-3.8.8-bin.tar.gz</dist.maven>
         <dist.wildfly>wildfly-17.0.1.Final.tar.gz</dist.wildfly>
         <dist.zookeeper>apache-zookeeper-3.7.2-bin.tar.gz</dist.zookeeper>
+        <docker.image.prefix />
```
jwomeara: I don't think this is needed, and im not sure why it keeps being added.

jwomeara (changes requested):

contrib/datawave-quickstart/docker/pom.xml
cogross: I needed it to get the docker image to build before - was able to build without it in place just now so removing it in the next push.

avgAGB (approved):

jwomeara (approved):

PR #2538, opened 2024-08-27
Enable Seeking Event Aggregation

apmoriarty:
- EventDataQueryFieldFilter now supports seeking aggregation when possible
- feature flagged this behavior, default is to not seek
- option is configurable via the query logic factory, including lookup uuid tune profiles
- refactored a bunch of stuff to support

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/function/IndexOnlyKeyToDocumentData.java
+++ b/warehouse/query-core/src/main/java/datawave/query/function/IndexOnlyKeyToDocumentData.java
@@ -38,9 +38,9 @@ import datawave.webservice.query.exception.QueryException;
  * Fetches index-only tf key/values and outputs them as "standard" field key/value pairs
  */
 public class IndexOnlyKeyToDocumentData extends KeyToDocumentData implements Iterator<Entry<DocumentData,Document>> {
-    private static final Collection<ByteSequence> COLUMN_FAMILIES = Lists.<ByteSequence> newArrayList(new ArrayByteSequence("d"));
+    private static final Collection<ByteSequence> COLUMN_FAMILIES = Lists.newArrayList(new ArrayByteSequence("d"));
 
-    private static Logger LOG = Logger.getLogger(IndexOnlyKeyToDocumentData.class);
+    private static final Logger LOG = Logger.getLogger(IndexOnlyKeyToDocumentData.class);
 
     private static final Entry<Key,Value> INVALID_COLUMNQUALIFIER_FORMAT_KEY = Maps.immutableEntry(new Key("INVALID_COLUMNQUALIFIER_FORMAT_KEY"), EMPTY_VALUE);
 
@@ -159,7 +159,7 @@ public class IndexOnlyKeyToDocumentData extends KeyToDocumentData implements Ite
             // get the document key
             Key docKey = getDocKey(from.getKey());
 
-            // Ensure that we have a non-empty colqual
+            // Ensure that we have a non-empty column qualifier
             final Key stopKey = new Key(from.getKey().getRow().toString(), from.getKey().getColumnFamily().toString(),
                             from.getKey().getColumnQualifier().toString() + '\u0000' + '\uffff');
 
--- a/warehouse/query-core/src/main/java/datawave/query/function/KeyToDocumentData.java
+++ b/warehouse/query-core/src/main/java/datawave/query/function/KeyToDocumentData.java
@@ -54,8 +54,7 @@ public class KeyToDocumentData implements Function<Entry<Key,Document>,Entry<Doc
 
     private static final ByteSequence EMPTY_BYTE_SEQUENCE = new ArrayByteSequence(new byte[] {});
 
-    protected static final Collection<ByteSequence> columnFamilies = Lists.newArrayList(new ArrayByteSequence("tf"), new ArrayByteSequence("d"));
-    protected static final boolean inclusive = false;
+    protected final Collection<ByteSequence> columnFamilies = Lists.newArrayList(new ArrayByteSequence("tf"), new ArrayByteSequence("d"));
 
     private final DescendantCountFunction countFunction;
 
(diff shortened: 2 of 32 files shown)


warehouse/query-core/src/main/java/datawave/query/function/IndexOnlyKeyToDocumentData.java
```diff
@@ -38,9 +38,9 @@
  * Fetches index-only tf key/values and outputs them as "standard" field key/value pairs
  */
 public class IndexOnlyKeyToDocumentData extends KeyToDocumentData implements Iterator<Entry<DocumentData,Document>> {
-    private static final Collection<ByteSequence> COLUMN_FAMILIES = Lists.<ByteSequence> newArrayList(new ArrayByteSequence("d"));
+    private static final Collection<ByteSequence> COLUMN_FAMILIES = Lists.newArrayList(new ArrayByteSequence("d"));
```
FineAndDandy, line 41: is the d constant defined somewhere else it can be referenced here?

warehouse/query-core/src/main/java/datawave/query/function/KeyToDocumentData.java
```diff
@@ -54,8 +54,7 @@ public class KeyToDocumentData implements Function<Entry<Key,Document>,Entry<Doc
 
     private static final ByteSequence EMPTY_BYTE_SEQUENCE = new ArrayByteSequence(new byte[] {});
 
-    protected static final Collection<ByteSequence> columnFamilies = Lists.newArrayList(new ArrayByteSequence("tf"), new ArrayByteSequence("d"));
-    protected static final boolean inclusive = false;
```
FineAndDandy, line 58: was this always false before?

warehouse/query-core/src/main/java/datawave/query/predicate/EventDataQueryFieldFilter.java
```diff
@@ -93,18 +95,39 @@ public boolean keep(Key k) {
...
+        }
+        return apply(entry.getKey());
     }
 
     @Override
-    public boolean apply(@Nullable Map.Entry<Key,String> input) {
-        return keyProjection.apply(input);
+    public boolean peek(@Nullable Map.Entry<Key,String> entry) {
+        // equivalent to apply in the event column case, simple redirect
+        return apply(entry);
```
FineAndDandy: peeking the same field repeatedly max falsely drive up the nextCount

warehouse/query-core/src/main/java/datawave/query/predicate/EventDataQueryFieldFilter.java
```diff
@@ -120,49 +143,43 @@ public boolean peek(@Nullable Map.Entry<Key,String> input) {
...
+        }
     }
 
     @Override
     public int getMaxNextCount() {
-        // not yet implemented
+        // while technically implemented, do not return the max next count here. This method is only used
+        // by the ChainableEventDataQueryFilter which does NOT guarantee that the filter will exclusively
+        // be applied to event keys.
         return -1;
```
FineAndDandy: should this maybe throw an UnsupportedMethodException then?

warehouse/query-core/src/main/java/datawave/query/function/IndexOnlyKeyToDocumentData.java
apmoriarty: I checked but didn't find anything. I would re-use the one from KeyToDocumentData, but those column families include the TF column and the IndexOnly variant explicitly excludes that.

warehouse/query-core/src/main/java/datawave/query/function/KeyToDocumentData.java
apmoriarty: Yes

warehouse/query-core/src/main/java/datawave/query/predicate/EventDataQueryFieldFilter.java
apmoriarty: oooooh good point. I'll rework this.

warehouse/query-core/src/main/java/datawave/query/predicate/EventDataQueryFieldFilter.java
apmoriarty: That would be one way to ensure the filter is being used appropriately...I like  it.

FineAndDandy (approved):

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1094,9 +1097,12 @@ protected Iterator<Entry<Key,Document>> mapDocument(SortedKeyValueIterator<Key,V
             log.trace("mapDocument " + fieldIndexSatisfiesQuery);
         }
         if (fieldIndexSatisfiesQuery) {
+            //  @formatter:off
             final KeyToDocumentData docMapper = new KeyToDocumentData(deepSourceCopy, this.myEnvironment, this.documentOptions, getEquality(),
-                            getEvaluationFilter(), this.includeHierarchyFields, this.includeHierarchyFields).withRangeProvider(getRangeProvider())
-                                            .withAggregationThreshold(getDocAggregationThresholdMs());
+                            getEventFilter(), this.includeHierarchyFields, this.includeHierarchyFields)
```
ivakegg, line 1102: What happened to using the evaluation filter?

apmoriarty: The evaluation filter was never used in the QueryIterator, it was only used by the TLDQueryIterator. This change allows the standard QueryIterator to do field filtering (and field based seeking).

You can see later in the change set [where the TLDQueryIterator](https://github.com/NationalSecurityAgency/datawave/pull/2538/files#diff-dcb340cebaade3ad6aa598bd3737a93f7050cde082910d3294d1d8220a9be591R141-R150) intercepts `getEventFilter()` and swaps in the evaluation filter:

ivakegg: ah, I see now.

ivakegg (approved):

PR #2539, opened 2024-08-27
Replace nonpublic Accumulo IterationInterruptedException

SethSmucker:
IterationInterruptedException is not part of Accumulo's public API.

Create IterationInterruptedException class in DataWave, replacing Accumulo's version.


avgAGB (dismissed):

foster33 (dismissed):

warehouse/core/src/main/java/datawave/iterators/IterationInterruptedException.java
```diff
@@ -0,0 +1,15 @@
+package datawave.iterators;
+
+/**
+ * Exception thrown if an interrupt flag is detected.
+ */
+public class IterationInterruptedException extends RuntimeException {
```
alerman: I might suggest customizing the name slightly to avoid IDE's pulling in accumulo's import of the class of the same name to avoid confusion later.

alerman (changes requested):

warehouse/core/src/main/java/datawave/iterators/IterationInterruptedException.java
ivakegg: +1

ivakegg (changes requested):
See Adam's comment

SethSmucker:
> See Adam's comment

Is it still too close? I changed it from `IterationInterruptedException` to `IterationInterruptException`. Might be a bit confusing if they're this close, what do you think?

ivakegg:
> > See Adam's comment
> 
> Is it still too close? I changed it from `IterationInterruptedException` to `IterationInterruptException`. Might be a bit confusing if they're this close, what do you think?

Now that I am looking one layer deeper, I realize that we were throwing that exception because we are implementing InterruptibleIterator which is an accumulo class.  Is that class part of the public API?  If so then we need to push for the original IteratorInterruptedException to be part of the public api.  Can you push that question over to the accumulo group please?

ddanielr:
> > > See Adam's comment
> > 
> > 
> > Is it still too close? I changed it from `IterationInterruptedException` to `IterationInterruptException`. Might be a bit confusing if they're this close, what do you think?
> 
> Now that I am looking one layer deeper, I realize that we were throwing that exception because we are implementing InterruptibleIterator which is an accumulo class. Is that class part of the public API? If so then we need to push for the original IteratorInterruptedException to be part of the public api. Can you push that question over to the accumulo group please?

@ivakegg the `InterruptibleIterator` is not in the list of iterators in the [accumulo public API](https://www.javadoc.io/static/org.apache.accumulo/accumulo-core/2.1.3/org/apache/accumulo/core/iterators/package-summary.html) 
After looking through the code, there are two datawave iterators which implement `InterruptibleIterator`, [SortedMultiMapIterator](https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/ingest-core/src/test/java/datawave/iterators/SortedMultiMapIterator.java) and [ColumnRangeIterator](https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/query-core/src/main/java/datawave/core/iterators/ColumnRangeIterator.java). 

`SortedMultiMapIterator` is used for a test case so it could probably be modified to implement a public iterator class without too much impact to other code behavior. However `ColumnRangeIterator` might need further review to determine the next steps.

ivakegg:
Ok, it appears that whole mechanism is basically OBE.  So essentially, everywhere we were passing InterationInterruptedException should remain as is.  We simply should not be implementing the InterruptibleIterator and hence should not have any flag to check on.  @SethSmucker Please move that whole mechanism, and keep the places where we are passing the InterationInterruptedException back up the stack.

ivakegg:
If you make the changes in your other PR, then I believe this one becomes OBE and can be closed.  We still need to make sure that exception is thrown up the stack.

alerman (dismissed):

ivakegg:
We need to remove the InterruptibleIterator as you have done.  What I think is that the additional exception is not what we need here.  Instead we should simply make sure that accumulo's IterationInterruptedException can pass through.  So that means we probably should not be catching that exception anywhere.  I will put this in the PR.

Issue #2544, opened 2024-09-05
RemoteDatawaveUserService should retry requests on UnknownHostException 


billoley:
In a kubernetes deployment when all instances of the authorization service are being rolled, requests to the authorization service cause an UnknownHostException.  

DatawaveUnavailableRetryStrategy in RemoteHttpService accounts for response codes of HttpStatus.SC_SERVICE_UNAVAILABLE (default in parent class) and HttpStatus.SC_NOT_FOUND (which can happen during service startup).

DatawaveRetryHandler extends DefaultHttpRequestRetryHandler to modify the nonRetriable IOExceptions and to add a delay after ConnectException.

nonRetriableClasses in DefaultHttpRequestRetryHandler
    InterruptedIOException
    UnknownHostException
    ConnectException
    SSLException

nonRetriableClasses in DatawaveRetryHandler
    UnknownHostException
    SSLException

so we want to retry on exceptions that implement InterruptedIOException and ConnectException

additionally, DatawaveRetryHandler added a delay after ConnectException using the same unavailableRetryCount and unavailableRetryDelay that DatawaveUnavailableRetryStrategy uses.

We should make is so that subclasses of RemoteHttpService can override nonRetriableClasses and unavailableRetryClasses.

PR #2545, opened 2024-09-05
RemoteDatawaveUserService should retry requests on UnknownHostException

billoley:
RemoteDatawaveUserService should retry requests on UnknownHostException

See https://github.com/NationalSecurityAgency/datawave/issues/2544

### DIFF
--- a/web-services/common/src/main/java/datawave/webservice/common/remote/RemoteHttpService.java
+++ b/web-services/common/src/main/java/datawave/webservice/common/remote/RemoteHttpService.java
@@ -576,7 +576,7 @@ public abstract class RemoteHttpService {
     }
 
     /**
-     * classes that are instances of IOException that will cause DatawaveRetryHandler to retry with delay subclasses of RemoteHttpService should override this
+     * Classes that are instances of IOException that will cause DatawaveRetryHandler to retry with delay. Subclasses of RemoteHttpService should override this
      * method if necessary
      */
     protected List<Class<? extends IOException>> getUnavailableRetryClasses() {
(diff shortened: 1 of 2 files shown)


web-services/common/src/main/java/datawave/webservice/common/remote/RemoteHttpService.java
```diff
@@ -570,30 +575,61 @@ public void setConfig(RemoteHttpServiceConfiguration config) {
         this.config = config;
     }
 
+    /**
+     * classes that are instances of IOException that will cause DatawaveRetryHandler to retry with delay subclasses of RemoteHttpService should override this
+     * method if necessary
+     */
+    protected List<Class<? extends IOException>> getUnavailableRetryClasses() {
```
keith-ratcliffe, line 582: Would something like `getRetriableClasses()` be a more succinct and/or future-proof name?

web-services/common/src/main/java/datawave/webservice/common/remote/RemoteHttpService.java
```diff
@@ -570,30 +575,61 @@ public void setConfig(RemoteHttpServiceConfiguration config) {
         this.config = config;
     }
 
+    /**
+     * classes that are instances of IOException that will cause DatawaveRetryHandler to retry with delay subclasses of RemoteHttpService should override this
```
keith-ratcliffe: Thanks to formatter, some punctuation would be helpful :-)

web-services/common/src/main/java/datawave/webservice/common/remote/RemoteHttpService.java
```diff
@@ -570,30 +575,61 @@ public void setConfig(RemoteHttpServiceConfiguration config) {
...
+     * method if necessary
+     */
+    protected List<Class<? extends IOException>> getUnavailableRetryClasses() {
+        return Arrays.asList(ConnectException.class);
+    }
+
+    /**
+     * classes that are instances of IOException that should not cause a retry subclasses of RemoteHttpService should override this method if necessary the
```
keith-ratcliffe: Thanks to formatter, some punctuation would be helpful :-)

web-services/common/src/main/java/datawave/webservice/common/remote/RemoteHttpService.java
billoley: All IOExceptions are retriable except for those in nonRetriableClasses

Apache HttpClient uses a delay for response codes that indicate that the service is unavailable:SC_SERVICE_UNAVAILABLE.  Datawave adds SC_NOT_FOUND

Datawave uses the same paradigm in the IOException handler to introduce a delay for IOExceptions that indicate that the service is unavailable.  Previously ConnectException.  This PR adds HostUnknownException for the call to the authorization service.

So it's not just that these are retriable.  They are classes that indicate that the service is unavailable and should be retried with a delay.

keith-ratcliffe (approved):

ivakegg (approved):

ivakegg (approved):

PR #2548, opened 2024-09-05
Add feature to exclude datatypes

jack-gitdev:


warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -746,7 +746,32 @@ protected void loadQueryParameters(ShardQueryConfiguration config, Query setting
...
+                if (dataType.charAt(0) == '!') {
+                    excludeSet.add(StringUtils.substring(dataType, 1));
+                }
+                else {
+                    typeFilter.add(dataType);
+                }
+            }
+
+            if (!excludeSet.isEmpty()) {
+                if (typeFilter.isEmpty()) {
```
FineAndDandy, line 761: throw an error if your ignoring due to the typeFilter being non-empty

FineAndDandy (approved):

FineAndDandy (changes requested):
can you add a unit test that covers the new functionality?

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -746,7 +746,32 @@ protected void loadQueryParameters(ShardQueryConfiguration config, Query setting
...
+                    MetadataHelper metadataHelper = prepareMetadataHelper(config.getClient(), this.getMetadataTableName(), config.getAuthorizations(), config.isRawTypes());
+
+                    try {
+                        for (Type<?> type : metadataHelper.getAllDatatypes()) {
```
ivakegg: These are the wrong data types.  I know this is highly confusing, but these data types are the types assigned to fields (e.g. LcNoDiacritics).  Unfortunately we chose the initial "datatype" designation unwisely.  So "datatype" really needs a qualifier of some sort.  Field datatype vs event or ingest datatype ?  IngestDataTypes is used for various method parameters in the MetadataHelper.   Anyway, the method metadataHelper.getDatatypes((Set<String>)null) will get the data types you are looking for.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -746,7 +746,30 @@ protected void loadQueryParameters(ShardQueryConfiguration config, Query setting
...
+                                    config.isRawTypes());
+
+                    try {
+                        typeFilter.addAll(metadataHelper.getDatatypes(null));
+                    } catch (TableNotFoundException e) {
+                        throw new RuntimeException(e);
+                    }
+                }
+
+                typeFilter.removeAll(excludeSet);
```
FineAndDandy, line 810: consider adding a test that there is at least one dataType left?

FineAndDandy (changes requested):

jack-gitdev:
Need to rework unit tests and add data to testing environment. Current data set does not support testing of necessary edge cases.

jack-gitdev:
Added data to the test data set, adjusted ShardQueryLogicTest to support the expectation of no results, and adjusted unit tests affected by the addition of new data. Removed draft status.

FineAndDandy (approved):

ivakegg (approved):

PR #2557, opened 2024-09-10
Ensure Discovery Queries only return indexed fields

foster33:
Fixes the problem where Discovery Queries could return fields, that were not marked as indexed in the metadata table, as results.

Also adds a unit test to validate the fix.


warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryLogic.java
```diff
@@ -379,7 +384,12 @@ public Iterator<DiscoveredThing> apply(Entry<Key,Value> from) {
                 }
                 ArrayList<DiscoveredThing> thangs = Lists.newArrayListWithCapacity(aw.get().length);
                 for (Writable w : aw.get()) {
-                    thangs.add((DiscoveredThing) w);
+                    // Check to see if the field is currently indexed, if it's not, we should NOT be adding it to 'thangs'
+                    if (indexedFields.contains(((DiscoveredThing) w).getField())) {
```
apmoriarty, line 648: opportunity to rename existing variable to something more adult sounding like "items"?

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryLogic.java
```diff
@@ -75,6 +75,7 @@ public class DiscoveryLogic extends ShardIndexQueryTable {
     public static final String SEPARATE_COUNTS_BY_COLVIS = "separate.counts.by.colvis";
     public static final String SHOW_REFERENCE_COUNT = "show.reference.count";
     public static final String REVERSE_INDEX = "reverse.index";
+    private static Set<String> indexedFields;
```
apmoriarty: I wonder if this needs to be static.

ivakegg: +1

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryLogic.java
ivakegg: I protest, we are not adults here

ivakegg (changes requested):
Except for the static member, I am good with this

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryLogic.java
foster33: Fixed in latest commit

apmoriarty (dismissed):

ivakegg (dismissed):

apmoriarty (approved):

ivakegg (approved):

PR #2558, opened 2024-09-12
Task/GitHub 1522

doublejai:


core/common-util/src/main/java/datawave/core/common/logging/ThreadConfigurableLogger.java
```diff
@@ -3,216 +3,132 @@
...
-import org.apache.log4j.Appender;
-import org.apache.log4j.Level;
-import org.apache.log4j.Logger;
-import org.apache.log4j.Priority;
+import org.apache.logging.log4j.Level;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
-public class ThreadConfigurableLogger extends Logger {
+public class ThreadConfigurableLogger {
```
jsduncan2, line 10: Why did we stop extending a class here? Could we extend `org.apache.logging.log4j.Logger`?

doublejai: Logger is an interface in Log4j 2.x

jsduncan2: There's a `SimpleLogger` that might be analogous to the log4j 1 `Logger`. If that's not quite right, I'd keep looking around for some concrete class to extend that isn't as heavy as the log42 `Logger` interface. If we can't find something suitable, we could consider removing some usages of `ThreadConfigurableLogger` that aren't necessary.

jsduncan2: Just to capture what we talked about in-person. We're going to leave this as-is as we can't find a suitable replacement for Logger. There is an interface but it's too heavy for what we need.

hlgp:
Deferred until someone can resolve conflicts

PR #2564, opened 2024-09-17
Updated to ensure we are using the same query id for federated query

ivakegg:
plans

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/planner/FederatedQueryPlanner.java
+++ b/warehouse/query-core/src/main/java/datawave/query/planner/FederatedQueryPlanner.java
@@ -13,6 +13,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
+import java.util.UUID;
 import java.util.stream.Collectors;
 
 import org.apache.accumulo.core.client.TableNotFoundException;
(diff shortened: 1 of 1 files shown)


warehouse/query-core/src/main/java/datawave/query/planner/FederatedQueryPlanner.java
```diff
@@ -334,10 +334,14 @@ public CloseableIterable<QueryData> process(GenericQueryConfiguration genericCon
             configCopy.setBeginDate(dateRange.getLeft());
             configCopy.setEndDate(dateRange.getRight());
 
+            // we want to make sure the same query and same query id are used for tracking purposes and execution
+            configCopy.setQuery(originalConfig.getQuery());
```
apmoriarty: Is it okay to share the same object across two instances of the planner? I would think we should copy the Query object but maybe that's an incorrect assumption on my part.

ivakegg: The settings object is static at this point, so no need to have a copy

apmoriarty: Perfect

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/planner/FederatedQueryPlanner.java
ivakegg: But that being said, perhaps safety.... looking at it.

ivakegg: looks like the query object is actually already duplicated, so all I need to do is reset the id..... testing

apmoriarty (approved):

PR #2567, opened 2024-09-17
build new protobuf and upgrade version to 3.25.4

austin007008:
add new fields to the TermWeight protobuf and upgrade protobuf version


hlgp:
Closing, deferred for now

PR #2568, opened 2024-09-17
WIP: Bulk V2

keith-ratcliffe:


properties/dev.properties
```diff
@@ -41,7 +41,7 @@ LIVE_CHILD_MAP_MAX_MEMORY_MB=1024
 BULK_CHILD_REDUCE_MAX_MEMORY_MB=2048
 LIVE_CHILD_REDUCE_MAX_MEMORY_MB=1024
 
-BULK_INGEST_DATA_TYPES=shardStats
+BULK_INGEST_DATA_TYPES=shardStats,wikipedia,mycsv,myjson
 LIVE_INGEST_DATA_TYPES=wikipedia,mycsv,myjson
```
ivakegg, line 45: If you are moving those types to bulk, then remove from live

keith-ratcliffe: IIRC, my goal here was just to signal to the user that they can now use either live or bulk for the 3 test data types...although, I guess it really doesn't matter in the end, since both live and bulk types here get dumped into the `ingest.data.types` config list, and that list gets deduped in `TypeRegistry`

Makes me wonder, do we  really need to maintain separate variables for these? So far, I haven't found a case where the distinction matters to our code

Anyway, the live flag maker still polls the `datatypeName` dirs in hdfs, same as always. And the bulk flag maker is now configured to run, polling the new `datatypeName-bulk` dirs (created in quickstart's `install-ingest.sh` above)

PR #2570, opened 2024-09-18
Feature/most recent unique

ivakegg:


avgAGB (dismissed):

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -303,4 +355,207 @@ public void funnel(byte[] from, PrimitiveSink into) {
...
+         * Build a list of potential hdfs directories based on each ivarator cache dir configs.
+         *
+         * @param ivaratorCacheDirConfigs
+         * @param hdfsSiteConfigURLs
+         * @param subdirectory
+         * @return A path
+         * @throws IOException
+         *             for issues with read/write
+         */
+        private static List<IvaratorCacheDir> getIvaratorCacheDirs(List<IvaratorCacheDirConfig> ivaratorCacheDirConfigs, String hdfsSiteConfigURLs,
```
FineAndDandy, line 458: this looks like a copy/paste from IteratorBuildingVisitor. Can you shift to a common function?

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/BufferedFileBackedSortedMap.java
```diff
@@ -0,0 +1,560 @@
...
+
+    /**
+     * If the number of maps is over maxFiles, then start compacting those files down. The goal is to get the number of files down around 50% of maxFiles.
+     *
+     * @param maxFiles
+     *            the max number of files
+     * @throws IOException
+     *             for IO Exceptions
+     */
+    public void compact(int maxFiles) throws IOException {
```
FineAndDandy, line 280: I think this does what it say it does. What I'm not clear on is why its necessary. These files should be relatively short lived. What problem are we avoiding by making multiple passes over the data to compact them?

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/BufferedFileBackedSortedMap.java
```diff
@@ -0,0 +1,560 @@
...
+    public Set<K> keySet() {
+        persistIfMultipleMaps();
+        return map.keySet();
+    }
+
+    @Override
+    public Collection<V> values() {
+        persistIfMultipleMaps();
+        return map.values();
+    }
```
FineAndDandy: some getter functions are persisting before returning while others are not. Can you define what operations should require persisting and which shouldn't and why?

warehouse/query-core/src/main/java/datawave/query/transformer/DocumentTransform.java
```diff
@@ -31,7 +31,7 @@ public interface DocumentTransform extends Function<Map.Entry<Key,Document>,Map.
     class DefaultDocumentTransform implements DocumentTransform {
         protected Query settings;
         protected MarkingFunctions markingFunctions;
-        protected long queryExecutionForPageStartTime;
+        protected long queryExecutionForPageStartTime = System.currentTimeMillis();
```
apmoriarty: It's not obvious to me what this change does. It looks like the variable is reset on every next call, so I'm not sure what setting this buys us. If this change is required it would be nice to know why

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/FileSerializableSortedMap.java
```diff
@@ -0,0 +1,287 @@
...
+
+            while ((total < bytesToSkip) && ((cur = stream.skip(bytesToSkip - total)) > 0)) {
+                total += cur;
+            }
+
+            byte[] buffer = new byte[4];
+            stream.read(buffer);
+
+            return ((buffer[3] & 0xFF)) + ((buffer[2] & 0xFF) << 8) + ((buffer[1] & 0xFF) << 16) + ((buffer[0]) << 24);
+        }
```
FineAndDandy: can you explain what is going on here and why?

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/FileSerializableSortedMap.java
```diff
@@ -0,0 +1,287 @@
...
+            delegate.writeObject(value);
+        }
+
+        @Override
+        public void writeSize(int size) throws IOException {
+            delegate.write((size >>> 24) & 0xFF);
+            delegate.write((size >>> 16) & 0xFF);
+            delegate.write((size >>> 8) & 0xFF);
+            delegate.write((size >>> 0) & 0xFF);
+        }
```
FineAndDandy: this could be simplified with the above to just write the integer?

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/HdfsBackedSortedMap.java
```diff
@@ -0,0 +1,302 @@
...
+    private static final String FILENAME_PREFIX = "SortedMapFile.";
+
+    public static class Builder<B extends Builder<B,K,V>,K,V> extends BufferedFileBackedSortedMap.Builder<B,K,V> {
+        private List<IvaratorCacheDir> ivaratorCacheDirs;
+        private String uniqueSubPath;
+        private FileSortedSet.PersistOptions persistOptions;
+
+        public Builder() {
+            // change the default buffer persist threshold
+            withBufferPersistThreshold(10000);
```
apmoriarty: Do we really want to buffer 10,000 elements before persisting to disk? Is this configurable? Should this persist threshold be different from a standard ivarator? I.e., what if I want to buffer 5k ivarator keys but only 250 unique documents?

Small nit, would prefer the code read like 
```java
withBufferPersistThreshold(10_000);
```

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/FileKeyValueSortedMap.java
FineAndDandy, line 1: Is this class used?

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/rfile/KeyValueByteDocumentTransforms.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    public static Document valueToDocument(Value value) {
+        if (value == null) {
+            return null;
+        }
+        DocumentDeserializer deserializer = new KryoDocumentDeserializer();
+        Document document = deserializer.deserialize(new ByteArrayInputStream(value.get()));
+        return document;
+    }
```
apmoriarty: I'm almost 100% certain this is going to be cost prohibitive from a performance perspective. Document serializers/deserializers should be initialized once and reused. Any class using these static methods should instead have a private instance. 

A did a series of simple test, all times are normalized to nano seconds per document. After accounting for JVM warm up and other java goodness, it's distinctly faster to reuse the serializer/deserializer
```
        //  using static methods
        //  count, to value      to doc
        //  100k   66,396        117,381
        //  10k    67,760        129,000
        //  1k    171,247        220,385
        //  100   407,812        493,630
        //  10  2,291,080      2,213,140

        //  reusing ser/de
        //  100k   38,581         65,382
        //  10k    37,031         85,645
        //  1k     74,649        126,809
        //  100   159,763        349,851
        //  10    654,658      1,542,268
```

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/HdfsBackedSortedMap.java
```diff
@@ -0,0 +1,302 @@
...
+        super(other);
+    }
+
+    protected HdfsBackedSortedMap(Builder builder) throws IOException {
+        super(builder);
+        this.handlerFactories = createFileHandlerFactories(builder.ivaratorCacheDirs, builder.uniqueSubPath, builder.persistOptions);
+        // for each of the handler factories, check to see if there are any existing files we should load
+        for (SortedMapFileHandlerFactory handlerFactory : handlerFactories) {
+            // Note: All of the file handler factories created by 'createFileHandlerFactories' are SortedMapHdfsFileHandlerFactories
+            if (handlerFactory instanceof SortedMapHdfsFileHandlerFactory) {
```
FineAndDandy: if this is so can this just be typed to SortedMapHdfsFileHandlerFactory in the field to simplify the code?

warehouse/query-core/src/test/java/datawave/query/UniqueTest.java
```diff
@@ -182,7 +187,10 @@ protected void runTestQueryWithUniqueness(Set<Set<String>> expected, String quer
         TransformIterator iter = new DatawaveTransformIterator(logic.iterator(), transformer);
         List<Object> eventList = new ArrayList<>();
         while (iter.hasNext()) {
-            eventList.add(iter.next());
+            Object o = iter.next();
+            if (o != null) {
+                eventList.add(o);
+            }
```
apmoriarty, line 193: Does this imply that before the transform iterator could return a null element after 'hasNext' was true? Is this a departure from the way things worked before or is the test being brought into alignment with the way the transform iterator actually works?

warehouse/query-core/src/test/java/datawave/query/UniqueTest.java
```diff
@@ -298,12 +313,14 @@ public void testUniquenessUsingLuceneFunction() throws Exception {
...
 
         queryString = "UUID:/^[CS].*/ AND #UNIQUE(DEATH_DATE,$BIRTH_DATE)";
+        expected.clear();
         expected.add(Sets.newHashSet(WiseGuysIngest.sopranoUID));
         expected.add(Sets.newHashSet(WiseGuysIngest.corleoneUID));
         expected.add(Sets.newHashSet(WiseGuysIngest.caponeUID));
         runTestQueryWithUniqueness(expected, queryString, startDate, endDate, extraParameters);
 
         queryString = "UUID:/^[CS].*/ AND #UNIQUE(death_date,birth_date)";
+        expected.clear();
```
apmoriarty, line 323: Instead of clearing the collection here, could clear the collection at the end of the driver method

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/MultiMapBackedSortedMap.java
```diff
@@ -0,0 +1,397 @@
...
+                if (it != null && it.hasNext()) {
+                    return true;
+                }
+            }
+            return false;
+        }
+
+        @Override
+        public Entry<K,V> next() {
+            populate();
```
FineAndDandy, line 314: can the map be pulled from directly and only if empty populate?

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/MultiMapBackedSortedMap.java
FineAndDandy, line 1: no unit tests for this class

warehouse/query-core/src/test/java/datawave/query/iterator/FieldIndexIntegrationTest.java
```diff
@@ -79,9 +79,9 @@ public class FieldIndexIntegrationTest {
             put("FIELD_A", 2);
             put("FIELD_B", 2);
             put("FIELD_C", 2);
-            put("TLD_FIELD_A", 6);
-            put("TLD_FIELD_B", 26);
-            put("TLD_FIELD_C", 8);
+            put("TLD_FIELD_A", 4); // 3 record ids, 1 value
+            put("TLD_FIELD_B", 14); // 13 record ids, 1 value
+            put("TLD_FIELD_C", 5); // 4 record ids, 1 value
```
apmoriarty: What caused this assertion to change? Was the TLD aggregator not working before?

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/MultiMapBackedSortedMap.java
```diff
@@ -0,0 +1,397 @@
...
+                        // if the iterator is finished, or did not contribute to the value being returned
+                        // then null it out since the value returned is already in the map to compare
+                        // on the next round
+                        nextIterators.set(i, null);
+                    }
+                }
+                populated = true;
+            }
+        }
+    }
```
FineAndDandy: I'm very unclear on how this method is supposed to work. This could easily be a me problem, but I'm struggling to understand what is going on here and why.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -212,71 +237,90 @@ byte[] getBytes(Document document) throws IOException {
...
+     * Return the query-specified field that the provided document matches, if one exists, or otherwise return null.
+     *
+     * @param documentField
+     *            The document field
+     * @return The query specified field
+     */
     private String getUniqueField(String documentField) {
         String baseDocumentField = getFieldWithoutGrouping(documentField);
         return uniqueFields.getFields().stream().filter((field) -> isMatchingField(baseDocumentField, field)).findFirst().orElse(null);
     }
```
FineAndDandy, line 315: what happens in the case where there are multiple unique fields on the query?

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -66,66 +80,39 @@ public class UniqueTransform extends DocumentTransform.DefaultDocumentTransform
...
-            modelMapping = HashMultimap.create();
-            // reverse the reverse query mapping which will give us a mapping from the final field name to the original field name(s)
-            for (Map.Entry<String,String> entry : model.getReverseQueryMapping().entrySet()) {
-                modelMapping.put(entry.getValue(), entry.getKey());
+    public void updateConfig(UniqueFields uniqueFields) {
+        // only reset the bloom filter if changing the field set
+        if (!this.uniqueFields.equals(uniqueFields)) {
+            this.uniqueFields = uniqueFields.clone();
+            log.info("Resetting unique fields on the unique transform");
+            this.bloom = BloomFilter.create(new ByteFunnel(), 500000, 1e-15);
```
FineAndDandy: since the bloom is only used when not using isMostRecent() can we avoid creating the bloom at all when that is set?

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
ivakegg: They are very similar, but different enough that I think they warrent separate implementations.

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/BufferedFileBackedSortedMap.java
ivakegg: These files are NOT short lived.  We need a compaction mechanism to avoid the problem of too many open file handles when we do the final merge sort.

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/BufferedFileBackedSortedMap.java
ivakegg: This is a good question.  The idea was that when the maps have been fully populated, then they should be persisted such that they would survive iterator tear-down/rebuilds (at least within the context of the ivarators).  That idea was copied over into the map implementations here.   I guess we could also force the persisting in a bunch of the other methods.  Let me go over why this was being done and adjust if needed.  Thank you.

warehouse/query-core/src/main/java/datawave/query/transformer/DocumentTransform.java
ivakegg: this may not be needed, but I vaguely remember some test case that needed the initial value set.... I will investigate.

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/FileSerializableSortedMap.java
ivakegg: Its reading the size (in elements) from the last 4 bytes in the file.

ivakegg: I will add comments.

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/FileSerializableSortedMap.java
ivakegg: At one point I was having issues with little/big endian so this guarentees that I am reading and writing the size correctly.

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/HdfsBackedSortedMap.java
ivakegg: hmm, I thought this was configurable. I will fix this if that is not the case.  I expect that 10K is probably not the best default for the most recent unique functionality.  Also I like 10_000.... I will do that.

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/FileKeyValueSortedMap.java
ivakegg: It was the initial one I used, and then I created the ByteDocument and KeyDocument versions.  I did not see a reason to delete this one as it could be used in another context.

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/rfile/KeyValueByteDocumentTransforms.java
ivakegg: Nice observation!  I will do as you suggest.

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/HdfsBackedSortedMap.java
ivakegg: yeah, that kinda looks ugly.  Trying to figure out why I did that.

warehouse/query-core/src/test/java/datawave/query/UniqueTest.java
ivakegg: Bringing it into line.  It is valid for a null to be returned when a timeout happens underneath.  This was to allow the RunningQuery to return a null page.  However the RunningQuery was changed to pull the results in a separate thread allow the timeout to be controlled there instead.  I still want to allow the iterators to return a null document for the purposes of avoiding hanging scans in the tservers.

warehouse/query-core/src/test/java/datawave/query/UniqueTest.java
ivakegg: probably.  I will take this into consideration.

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/MultiMapBackedSortedMap.java
ivakegg: hmm... perhaps not directly.  I will see what I can do.

warehouse/query-core/src/test/java/datawave/query/iterator/FieldIndexIntegrationTest.java
ivakegg: I really don't know why this changed.. I will figure it out however.

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/MultiMapBackedSortedMap.java
ivakegg: We had the conversation off line for this one.  I am presumming this is now understood.  However I will try to add comments to make things more clear.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
ivakegg: probably, I will see if I can do that.

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/BufferedFileBackedSortedMap.java
```diff
@@ -0,0 +1,560 @@
...
+            this.rewriteStrategy = (FileSortedMap.RewriteStrategy<K,V>) rewriteStrategy;
+            return self();
+        }
+
+        public B withNumRetries(int numRetries) {
+            this.numRetries = numRetries;
+            return self();
+        }
+
+        public B withHandlerFactories(List<SortedMapFileHandlerFactory> handlerFactories) {
```
FineAndDandy, line 102: this method is never called. Not sure this class will work right without the handlerFactories being set

ivakegg: yikes... that does not sound good.  I will figure that out.

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/BufferedFileBackedSortedMap.java
ivakegg: Now I am thinking that the persistence should never be done automatically here.  The transform should drive that.

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/BufferedFileBackedSortedMap.java
ivakegg: Ah, I see, the handler factories are setup in the HdfsBackedSortedMap which extends the BufferedFileBackedSortedMap.  This is the builder that the UniqueTransform uses.

FineAndDandy: my mistake, sorry

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/rfile/KeyValueByteDocumentTransforms.java
FineAndDandy: also worth keeping in mind that Kryo, backed by KryoDocumentSerializer is NOT thread safe

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/BufferedFileBackedSortedMap.java
ivakegg: Updated to not automatically persisty and explicitly persist in the transform.

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/HdfsBackedSortedMap.java
ivakegg: Changed the constant to add the '_'.  This is being overridden in the UniqueTransform where the maps are built.

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/rfile/KeyValueByteDocumentTransforms.java
ivakegg: done

warehouse/query-core/src/main/java/datawave/query/transformer/DocumentTransform.java
ivakegg: determined this is more of a safety thing.  It certainly does not hurt.  Moved into constructor.

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/HdfsBackedSortedMap.java
ivakegg: Updated to use the correct type.

warehouse/query-core/src/test/java/datawave/query/UniqueTest.java
ivakegg: That would be a side effect and hence not obvious to the user.  I would rather leave the explicit clear calls where they are.

warehouse/query-core/src/test/java/datawave/query/iterator/FieldIndexIntegrationTest.java
ivakegg: no longer doing this.... I was attempting to fix the whole document/attribute equality mechanisms and it resulted in this change.  Undid all of that.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
ivakegg: done

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/MultiMapBackedSortedMap.java
ivakegg: done

FineAndDandy (dismissed):

apmoriarty (dismissed):

PR #2575, opened 2024-09-25
Offload potentially expensive query planning steps to async tasks.

apmoriarty:
Although the metadata helper is supposed to cache expensive method calls, this does not always work out in a production environment. Many of these expensive calls can be executed in parallel, thus reducing the setup time for a query. 

- Offloaded potentially expensive method calls to async tasks. 
- updated some planner method interfaces
- optionally log timed saved via async execution


warehouse/query-core/src/main/java/datawave/query/index/lookup/ShardRangeStream.java
```diff
@@ -134,4 +137,33 @@ public QueryPlan apply(Entry<Key,Value> entry) {
...
+        String nonIndexedTypes = QueryOptions.buildFieldNormalizerString(nonIndexedQueryFieldsDatatypes);
+        DefaultQueryPlanner.addOption(cfg, QueryOptions.NON_INDEXED_DATATYPES, nonIndexedTypes, false);
+
+        try {
+            String serializedTypeMetadata = metadataHelper.getTypeMetadata(config.getDatatypeFilter()).toString();
+            DefaultQueryPlanner.addOption(cfg, QueryOptions.TYPE_METADATA, serializedTypeMetadata, false);
+        } catch (TableNotFoundException e) {
+            throw new RuntimeException(e);
+        }
+
```
mineralntl, line 155: Could the above and below try statements be simplified into a single try since we end up throwing the same RTE regardless of no table or IO problems?

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -3229,8 +3259,254 @@ public static Date getEndDateForIndexLookup(Date endDate) {
...
+            }
+        }
+
+        if (nonEventFields == null) {
+            return Collections.emptySet();
+        }
+
+        return nonEventFields;
+    }
+
```
mineralntl, line 3372: Would it be preferable to condense the preceeding eight methods to call a single getCallable() method? Seems like a bit of redundant try/catch and emptySet returning.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -3229,8 +3259,254 @@ public static Date getEndDateForIndexLookup(Date endDate) {
...
+                stopwatch.stop();
+            }
+        }
+
+        if (termFrequencyFields == null) {
+            return Collections.emptySet();
+        }
+
+        return termFrequencyFields;
+    }
```
mineralntl, line 3404: This one too

mineralntl (dismissed):
LGTM other than possibly condensing a few methods but that's just personal preference

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
apmoriarty: I could certainly write a callable per return type, so a FieldSetCallable could handle several of these use cases. The others unfortunately need to remain because of the different return types.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -3229,8 +3259,254 @@ public static Date getEndDateForIndexLookup(Date endDate) {
...
+    protected CompositeMetadata getCompositeMetadata() {
+        if (compositeMetadata == null && compositeMetadataCallable != null) {
+            TraceStopwatch stopwatch = stageStopWatch.newStartedStopwatch(compositeMetadataCallable.stageName());
+            try {
+                while (compositeMetadata == null) {
+                    compositeMetadata = compositeMetadataFuture.get(concurrentTimeoutMillis, TimeUnit.MILLISECONDS);
+                }
+            } catch (InterruptedException | ExecutionException | TimeoutException e) {
+                log.error("Failed to fetch CompositeMetadata");
+                log.error(e.getStackTrace());
```
FineAndDandy: these lines can be combined so that log parsing can process this as a single event (all other cases as well)

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
FineAndDandy: it might also be possible to return the callable rather than the value, this way you could wrap the waiting and error handling on a single method that is rused, then all of these would either have a clean get or throw a RuntimeException

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -3229,8 +3259,254 @@ public static Date getEndDateForIndexLookup(Date endDate) {
...
+                throw new RuntimeException(e);
```
FineAndDandy: Maybe throw a specialty exception so we can track how often this leads to a query failing? This could be caught internally or returned with a new datawave error code

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
apmoriarty: combined in next commit

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
apmoriarty: done in next commit

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
apmoriarty: Common code for getting field sets is extracted to a common method in the next commit

avgAGB (approved):

mineralntl (approved):

PR #2578, opened 2024-09-27
Close ScannerSessiosn on timeout, correctly apply timeout fixes #2574

### LINKED ISSUE #2574: Interrupt AnyFieldScanners after timeouts

FineAndDandy:


warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BoundedRangeIndexLookup.java
```diff
@@ -50,7 +51,7 @@ public class BoundedRangeIndexLookup extends AsyncIndexLookup {
     private final LiteralRange<?> literalRange;
 
     protected Future<Boolean> timedScanFuture;
-    protected long lookupStartTimeMillis = Long.MAX_VALUE;
+    protected AtomicLong lookupStartTimeMillis = new AtomicLong(Long.MAX_VALUE);
```
ivakegg, line 54: I am guessing you did this because of thead access?  Could have used the "volatile" keyword as well.

ivakegg, line 54: Was this because of thread access issues?  Could have used the "volatile" qualifier.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/FieldNameIndexLookup.java
```diff
@@ -117,7 +124,10 @@ public synchronized IndexLookupMap lookup() {
         if (!sessions.isEmpty()) {
             try {
                 // for field name lookups, we wait indefinitely
-                timedScanWait(timedScanFuture, lookupStartedLatch, lookupStoppedLatch, lookupStartTimeMillis, Long.MAX_VALUE);
+                // TODO consider if this really should be Long.MAX_VALUE or some time less. Other index scanners are set to config.getMaxIndexScanTimeMillis().
```
ivakegg, line 127: Do you need this TODO if you are now passing config.getMaxAnyFieldScanTimeMillis?

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/FieldNameIndexLookup.java
```diff
@@ -135,13 +145,18 @@ protected Callable<Boolean> createTimedCallable(final Iterator<Entry<Key,Value>>
...
-                lookupStartTimeMillis = System.currentTimeMillis();
+                lookupStartTimeMillis.set(System.currentTimeMillis());
                 lookupStartedLatch.countDown();
 
                 final Text holder = new Text();
 
                 try {
                     while (iter.hasNext()) {
+                        // check for interrupt which may be triggered by closing the batch scanner
+                        if (Thread.currentThread().isInterrupted()) {
```
ivakegg: Hmm.... we really need this check?  How can we be running if we were interrupted?

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/FieldNameIndexLookup.java
ivakegg: If you are referring to the default value of the timeout, then we could change the default to be that of the query timeout.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/FieldNameIndexLookup.java
FineAndDandy: without this check the thread does not reliably interrupt

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/FieldNameIndexLookup.java
FineAndDandy: I wanted the TODO so we talk about what the value should be. Yes it's now configurable. But if no one looks at it it won't ever change.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BoundedRangeIndexLookup.java
FineAndDandy: This was because passing a long by reference doesn't work :) Also we are accessing this long in different threads.

ivakegg: ah, it was more because of the reference part.... got it.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/FieldNameIndexLookup.java
ivakegg: interesting, ok

ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/FieldNameIndexLookup.java
drewfarris: I always struggle to remember the difference between `Thread.interrupted()` vs `Thread.currentThread().interrupted()` in cases like this. I have a suspicion we should be clearing the interrupted flag on the thread by using `Thread.interrupted()` here instead, but it is worth considering further. The difference is that `Thread.interrupted()` will clear the interrupted flag on the current thread which might be what we want because this is effectively handing (and then resetting the interrupted flag?)

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/RegexIndexLookup.java
```diff
@@ -266,6 +267,10 @@ protected Callable<Boolean> createTimedCallable(final Iterator<Entry<Key,Value>>
                     }
 
                     while (iter.hasNext()) {
+                        // check if interrupted which may be triggered by closing a batch scanner
+                        if (Thread.currentThread().isInterrupted()) {
```
drewfarris: Consider `Thread.interrupted()` vs `Thread.isInterrupted()` here as well.

warehouse/query-core/src/test/java/datawave/query/AnyFieldQueryTest.java
```diff
@@ -1401,4 +1462,413 @@ protected void testInit() {
         this.auths = CitiesDataType.getTestAuths();
         this.documentKey = CityField.EVENT_ID.name();
     }
+
+    private static class DelayedClient implements AccumuloClient {
```
drewfarris, line 1466: Could be useful as a documented, top-level class?

drewfarris (commented):
Looks good to me - per discussion, I'll take a look once you've updated the thread interruption related code.

warehouse/query-core/src/test/java/datawave/query/AnyFieldQueryTest.java
FineAndDandy: definitely could be used for that. I'd like that to be a separate effort.

ivakegg (approved):

drewfarris (approved):

PR #2580, opened 2024-09-27
Apply Scan Hints to SessionOptions

apmoriarty:
Fixed several cases where scan hints were not persisted
- pushdown scheduler
- range stream
- shard index static methods (expansion)

Added default execution hint to shard index static methods

Configured test query logic factory to use ivarator hints


warehouse/query-core/src/test/resources/datawave/query/QueryLogicFactory.xml
```diff
@@ -231,6 +231,27 @@
                 <!-- JEXL is the default Jexl query syntax -->
             </map>
         </property>
+
+        <property name="tableHints">
```
ivakegg: Need to update query starter to enable this configuration.

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
```diff
@@ -753,6 +758,7 @@ public void copyFrom(ShardQueryConfiguration other) {
         this.setUseQueryTreeScanHintRules(other.isUseQueryTreeScanHintRules());
         this.setQueryTreeScanHintRules(other.getQueryTreeScanHintRules());
         this.setFieldIndexHoleMinThreshold(other.getFieldIndexHoleMinThreshold());
+        this.setIndexExpansionHintKey(other.getIndexExpansionHintKey());
```
ivakegg: I thought we were going to configure the ScannerFactory, probably just adding to the hintsByTable to have entries matching some of our functions.  So we would add a configuration to that map for "expansion" or "indexExpansion".  Then the getScanner(table, function) would look in that table first by the function, and if empty then lookup by table.  I would rather keep that configuration all together instead of add another config item here.

ivakegg: Ah, it appears that the table hints in the GenericQueryConfiguration does that.  I don't think we need to configure an index expansion hint key, that can be a reserved word.  We should add a comment (probably where the table hints are configured) that lists any reserved keys such as QUERY_EXPANSION.

ivakegg (approved):

drewfarris (approved):

PR #2600, opened 2024-10-09
Regex Rewrite Visitor

apmoriarty:
Added visitor to that rewrites regex terms without affecting the overall executability of the query.

Added visitor that detects if a term or subtree is executable, or an 'anchor'.


avgAGB (dismissed):

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -809,6 +822,11 @@ protected ASTJexlScript updateQueryTree(ScannerFactory scannerFactory, MetadataH
             config.setQueryTree(timedEnforceUniqueDisjunctionsWithinExpressions(timers, config.getQueryTree()));
         }
 
+        // rewrite regex nodes, optimistically
+        if (rewriteRegexTerms) {
```
hlgp: Would probably rather see this after the RegexIndexExpansion visitor call.  If we can expand to explicit terms, we should.  Otherwise, rewrite the parts that can't if the values are not selective enough.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -387,6 +393,7 @@ protected DefaultQueryPlanner(DefaultQueryPlanner other) {
         rangeStreamClass = other.rangeStreamClass;
         setSourceLimit(other.sourceLimit);
         setPushdownThreshold(other.getPushdownThreshold());
+        setRegexRewriteOptions(other.getRegexRewriteOptions());
```
ivakegg, line 396: Why not simply create more transform rules (see 2 lines below) instead of creating a whole new injected set of rewrite options?

apmoriarty: These are not simple transform rules. These rules are more tailored and intended to be run before and after index expansion.

To elaborate, the regex rewrite rules applied in a way that preserves query executability whereas the transform rules will always apply, even if means making a query non-executable.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/pushdown/AnchorDetectionVisitor.java
```diff
@@ -0,0 +1,208 @@
...
+import org.apache.commons.jexl3.parser.JexlNode;
+
+import datawave.query.jexl.JexlASTHelper;
+import datawave.query.jexl.nodes.QueryPropertyMarker;
+import datawave.query.jexl.visitors.ShortCircuitBaseVisitor;
+
+/**
+ * Determines if a subtree is an anchor for a given query
+ * <p>
+ * An anchor is defined as an executable leaf or subtree.
```
ivakegg, line 30: does not the executability visitor already do this?  How is this different?

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
ivakegg: but.... as far as I can tell you are simply transforming nodes within the tree.  I think I need to sit down with you to understand why a separate visitor.  The node transform rules are supplied the node and hence its subtree, so they can do whatever they need to keep the query executable.

PR #2603, opened 2024-10-09
Improvement for IsNotNullPruningVisitor

avgAGB:


warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IsNotNullPruningVisitorTest.java
```diff
@@ -575,6 +577,22 @@ public void testNoOpQueryPropertyMarkers() {
         test(query, query);
     }
 
+    @Test
+    public void testFutureCase_Custom() {
```
apmoriarty: rename to something like 'testPruningNestedUnionOfIsNotNullFunctions' ?

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IsNotNullPruningVisitorTest.java
```diff
@@ -575,6 +577,22 @@ public void testNoOpQueryPropertyMarkers() {
...
+        String query = "FOO == 'bar' && (!(FOO == null) || !(FOO2 == null) || !(FOO3 == null) || !(FOO4 == null))";
+        String expected = "FOO == 'bar'";
+
+        test(query, expected);
+    }
+
+    @Test
+    public void testFutureCase_CustomTwo() {
+        // logically, these unions are equivalent and the 'is not null' side can be pruned
+        String query = "FOO == 'bar' && (!(FOO2 == null) || !(FOO4 == null))";
```
apmoriarty, line 592: I told you the wrong thing with this case. We can't prune is not null terms for mutually exclusive fields. A document like `{FOO=bar, FOO2=baz, FOO4=fizz}` shouldn't be returned by this query, but if the 'not null' functions are pruned then the document will be returned.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IsNotNullPruningVisitorTest.java
```diff
@@ -493,13 +493,13 @@ public void testFutureCase_PartialPruneOfUnionViaUnion() {
...
-        test(query, query);
+        String expected = "(FOO == 'bar' || FOO == 'baz')";
+        test(query, expected);
 
         // should also work for filter:includeRegex
         query = "(!(FOO == null) || !(FOO2 == null)) && (filter:includeRegex(FOO, 'bar.*') || filter:includeRegex(FOO, 'baz.*'))";
-        // expected = "!(FOO2 == null) && (filter:includeRegex(FOO, 'bar.*') || filter:includeRegex(FOO, 'baz.*'))";
-        test(query, query);
+        expected = "(filter:includeRegex(FOO, 'bar.*') || filter:includeRegex(FOO, 'baz.*'))";
+        test(query, expected);
```
apmoriarty, line 502: great work

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IsNotNullPruningVisitorTest.java
```diff
@@ -537,10 +537,12 @@ public void testNoOpCases() {
...
         // cannot prune half of a union
         query = "(!(FOO == null) || !(FOO2 == null)) && FOO == 'bar'";
-        test(query, query);
+        String expected = "FOO == 'bar'";
+        test(query, expected);
 
         query = "(!(FOO == null) || !(FOO2 == null)) && FOO =~ 'ba.*'";
-        test(query, query);
+        expected = "FOO =~ 'ba.*'";
+        test(query, expected);
```
apmoriarty, line 545: fantastic

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IsNotNullPruningVisitorTest.java
avgAGB: You're correct, it turns out the comment for this test case was not updated. This particular string will not do any pruning since FOO2 and FOO4 are not the fields in the EQNode clause. This query will end up doing no action when passed through the pruning visitor.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IsNotNullPruningVisitorTest.java
avgAGB: Done in current commit

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/IsNotNullPruningVisitorTest.java
avgAGB: Updated the comment here to clarify the function, as we chatted about a little in office.

alerman (approved):

apmoriarty (approved):

PR #2608, opened 2024-10-11
Initial File Provider Service update

awrigh11:
Adding the file provider service to the microservices framework


docker/docker-compose.yml
```diff
@@ -413,6 +413,28 @@ services:
...
+    ports:
+      - "8280:8080"
+      - "8643:8443"
+    volumes:
+      - ${PKI_DIR:-./pki}:/etc/pki:ro
+      - ./logs:/logs
+    networks:
+      - demo
+    depends_on:
+      authorization:
```
awrigh11: This doesn't actually depend on the authorization service. Remove.

docker/README.md
```diff
@@ -328,6 +339,10 @@ Start the default services, the kafka services, and the dictionary service:
 
 ```docker compose --profile quickstart --profile dictionary --profile kafka up -d```
 
+Start the default services, and the file provider service:
+
+```docker compose --profile file-provider up -d```
```
avgAGB, line 344: would the quickstart profile need to be started beforehand in the case? Or since the file-provider might be independent of it maybe we don't need it?

avgAGB: would the quickstart profile need to be started beforehand in the case? Or since the file-provider might be independent of it maybe we don't need it?

awrigh11: Good catch. Added it.

apmoriarty (approved):

jjonesy16 (approved):

PR #2613, opened 2024-10-16
Improve Bounded Range Expansion

apmoriarty:
- replace most of the bounded index lookup iterator stack with a single BoundedRangeExpansionIterator
- add integration test for bounded index lookup
- add unit test for bounded range expansion iterator
- Integrate tests from commit 7092ee4f877d8931d7368238ad3ca472159ce07a

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/core/iterators/BoundedRangeExpansionIterator.java
+++ b/warehouse/query-core/src/main/java/datawave/core/iterators/BoundedRangeExpansionIterator.java
@@ -5,6 +5,7 @@ import java.util.Collection;
 import java.util.Map;
 import java.util.TreeSet;
 
+import datawave.query.jexl.LiteralRange;
 import org.apache.accumulo.core.data.ByteSequence;
 import org.apache.accumulo.core.data.Key;
 import org.apache.accumulo.core.data.Range;
@@ -25,7 +26,9 @@ import datawave.query.Constants;
 /**
  * A {@link SeekingFilter} that attempts to expand bounded ranges using the global index
  * <p>
- * The caller is responsible for fetching the appropriate column families
+ * The caller is responsible for fetching the appropriate column families. The range is constructed from a {@link LiteralRange}.
+ * <p>
+ * The only thing this iterator does is advance through datatypes if a filter is supplied, advance to the start date, and advance to the next row within the range.
  */
 public class BoundedRangeExpansionIterator extends SeekingFilter implements OptionDescriber {
 
@@ -76,7 +79,7 @@ public class BoundedRangeExpansionIterator extends SeekingFilter implements Opti
 
     @Override
     public FilterResult filter(Key k, Value v) {
-        log.debug("filter key: {}", k.toStringNoTime());
+        log.trace("filter key: {}", k.toStringNoTime());
 
         // shard + null + datatype
         String cq = k.getColumnQualifier().toString();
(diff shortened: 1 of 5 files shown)


warehouse/query-core/src/main/java/datawave/core/iterators/BoundedRangeExpansionIterator.java
```diff
@@ -0,0 +1,171 @@
...
+
+        super.init(source, options, env);
+    }
+
+    @Override
+    public IteratorOptions describeOptions() {
+        IteratorOptions opts = new IteratorOptions(getClass().getName(), "Expands bounded ranges using the global index", null, null);
+        opts.addNamedOption(START_DATE, "The start date");
+        opts.addNamedOption(END_DATE, "The end date");
+        opts.addNamedOption(DATATYPES_OPT, "The set of datatypes used to filter keys");
```
ivakegg: Denote that the datatype option is optional

warehouse/query-core/src/main/java/datawave/core/iterators/BoundedRangeExpansionIterator.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        String datatype = cq.substring(index + 1);
+        if (!datatypes.isEmpty() && !datatypes.contains(datatype)) {
+            log.trace("datatype {} was filtered out, advancing to next key", datatype);
+            return new FilterResult(false, AdvanceResult.NEXT);
+        }
+
+        if (prevRow != null && prevRow.equals(k.getRow())) {
+            // should never get here
```
ivakegg: I expect we would get here.  Is this not how we advance to the next value?

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BoundedRangeIndexLookup.java
```diff
@@ -137,20 +134,17 @@ public synchronized void submit() {
...
+                setting.addOption(BoundedRangeExpansionIterator.START_DATE, startDay);
+                setting.addOption(BoundedRangeExpansionIterator.END_DATE, endDay);
+                if (!config.getDatatypeFilter().isEmpty()) {
+                    setting.addOption(BoundedRangeExpansionIterator.DATATYPES_OPT, Joiner.on(',').join(config.getDatatypeFilter()));
+                }
+                bs.addScanIterator(setting);
 
                 // If this is a composite field, with multiple terms, we need to setup our query to filter based on each component of the composite range
-                if (config.getCompositeToFieldMap().get(literalRange.getFieldName()) != null) {
+                if (!config.getCompositeToFieldMap().get(literalRange.getFieldName()).isEmpty()) {
```
ivakegg, line 147: Do we need to check for null as well?

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BoundedRangeIndexLookup.java
```diff
@@ -239,68 +226,30 @@ protected Callable<Boolean> createTimedCallable(final Iterator<Entry<Key,Value>>
...
-                        String field = null;
+                        // safety check...
+                        Preconditions.checkState(field.equals(literalRange.getFieldName()),
+                                        "Got an unexpected field name when expanding range" + field + " " + literalRange.getFieldName());
 
-                        boolean foundDataType = false;
+                        // obtaining the size of a map can be expensive,
+                        // instead
+                        // track the count of each unique item added.
+                        indexLookupMap.put(field, uniqueTerm);
```
ivakegg, line 247: So much better.... nice

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BoundedRangeIndexLookup.java
apmoriarty: We do not. The `Multimap.get()` call returns an empty collection when nothing maps to the provided key. Which means that yes, we were slapping this iterator onto the stack every single time for no reason.

warehouse/query-core/src/main/java/datawave/core/iterators/BoundedRangeExpansionIterator.java
apmoriarty: Comment updated in the next commit. The iterator should only return a single key per unique row in the index, this is more of a safety check than anything else. At one point I had this code block throw an exception and the build passed without issue.

warehouse/query-core/src/main/java/datawave/core/iterators/BoundedRangeExpansionIterator.java
apmoriarty: done

ivakegg (approved):

PR #2614, opened 2024-10-16
Ingest field configuration helper cache

### LINKED ISSUE #2612: FieldConfigHelper results cache
JFR analysis showed there may be a cost with pattern evaluation in the `FieldConfigHelper`. In testing/benchmarks, it appears that caching the field lookup results would help with ingest performance. 

Evaluate optionally including a cache that can be configured to keep results of field evaluations. The implementation should also be aware of some configurable maximum limit.

dtspence:
Implementation of a field helper cache using an LRU map. 
* The cache can be optionally enabled on the `BaseIngestHelper` instance and will wrap an existing `FieldConfigHelper` implementation.
* By default the cache is disabled and must be enabled via configuration.

As part of the work, there was also a factory implementation in separate branch which would allow an implementation to be specified by configuration, which may be preferrable.

Closes #2612

### DIFF
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CachedFieldConfigHelper.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CachedFieldConfigHelper.java
@@ -1,8 +1,7 @@
 package datawave.ingest.data.config;
 
-import java.util.EnumMap;
 import java.util.Map;
-import java.util.function.Function;
+import java.util.function.Predicate;
 
 import org.apache.commons.collections4.map.LRUMap;
 
@@ -10,10 +9,10 @@ import com.google.common.annotations.VisibleForTesting;
 
 public class CachedFieldConfigHelper implements FieldConfigHelper {
     private final FieldConfigHelper underlyingHelper;
-    private final Map<String,ResultEntry> resultCache;
+    private final Map<String,CachedEntry> resultCache;
 
     enum AttributeType {
-        INDEXED_FIELD, REVERSE_INDEXED_FIELD, TOKENIZED_FIELD, REVERSE_TOKENIZED_FIELD, STORED_FIELD, INDEXED_ONLY
+        INDEXED_FIELD, REVERSE_INDEXED_FIELD, TOKENIZED_FIELD, REVERSE_TOKENIZED_FIELD, STORED_FIELD, INDEX_ONLY_FIELD
     }
 
     public CachedFieldConfigHelper(FieldConfigHelper helper, int limit) {
@@ -26,50 +25,96 @@ public class CachedFieldConfigHelper implements FieldConfigHelper {
 
     @Override
     public boolean isStoredField(String fieldName) {
-        return getOrEvaluate(AttributeType.STORED_FIELD, fieldName, underlyingHelper::isStoredField);
+        return getFieldResult(AttributeType.STORED_FIELD, fieldName, underlyingHelper::isStoredField);
     }
 
     @Override
     public boolean isIndexedField(String fieldName) {
-        return getOrEvaluate(AttributeType.INDEXED_FIELD, fieldName, underlyingHelper::isIndexedField);
+        return getFieldResult(AttributeType.INDEXED_FIELD, fieldName, underlyingHelper::isIndexedField);
     }
 
     @Override
     public boolean isIndexOnlyField(String fieldName) {
-        return getOrEvaluate(AttributeType.INDEXED_ONLY, fieldName, underlyingHelper::isIndexOnlyField);
+        return getFieldResult(AttributeType.INDEX_ONLY_FIELD, fieldName, underlyingHelper::isIndexOnlyField);
     }
 
     @Override
     public boolean isReverseIndexedField(String fieldName) {
-        return getOrEvaluate(AttributeType.REVERSE_INDEXED_FIELD, fieldName, underlyingHelper::isReverseIndexedField);
+        return getFieldResult(AttributeType.REVERSE_INDEXED_FIELD, fieldName, underlyingHelper::isReverseIndexedField);
     }
 
     @Override
     public boolean isTokenizedField(String fieldName) {
-        return getOrEvaluate(AttributeType.TOKENIZED_FIELD, fieldName, underlyingHelper::isTokenizedField);
+        return getFieldResult(AttributeType.TOKENIZED_FIELD, fieldName, underlyingHelper::isTokenizedField);
     }
 
     @Override
     public boolean isReverseTokenizedField(String fieldName) {
-        return getOrEvaluate(AttributeType.REVERSE_TOKENIZED_FIELD, fieldName, underlyingHelper::isReverseTokenizedField);
+        return getFieldResult(AttributeType.REVERSE_TOKENIZED_FIELD, fieldName, underlyingHelper::isReverseTokenizedField);
     }
 
     @VisibleForTesting
-    boolean getOrEvaluate(AttributeType attributeType, String fieldName, Function<String,Boolean> evaluateFn) {
-        return resultCache.computeIfAbsent(fieldName, ResultEntry::new).resolveResult(attributeType, evaluateFn);
+    boolean getFieldResult(AttributeType attributeType, String fieldName, Predicate<String> fn) {
+        return resultCache.computeIfAbsent(fieldName, CachedEntry::new).get(attributeType).getResultOrEvaluate(fn);
     }
 
-    private static class ResultEntry {
+    private static class CachedEntry {
         private final String fieldName;
-        private final EnumMap<AttributeType,Boolean> resultMap;
+        private final MemoizedResult indexed;
+        private final MemoizedResult reverseIndexed;
+        private final MemoizedResult stored;
+        private final MemoizedResult indexedOnly;
+        private final MemoizedResult tokenized;
+        private final MemoizedResult reverseTokenized;
 
-        ResultEntry(String fieldName) {
+        private CachedEntry(String fieldName) {
             this.fieldName = fieldName;
-            this.resultMap = new EnumMap<>(AttributeType.class);
+            this.indexed = new MemoizedResult();
+            this.reverseIndexed = new MemoizedResult();
+            this.stored = new MemoizedResult();
+            this.indexedOnly = new MemoizedResult();
+            this.tokenized = new MemoizedResult();
+            this.reverseTokenized = new MemoizedResult();
         }
 
-        boolean resolveResult(AttributeType attributeType, Function<String,Boolean> evaluateFn) {
-            return resultMap.computeIfAbsent(attributeType, (t) -> evaluateFn.apply(fieldName));
+        private MemoizedResult get(AttributeType attributeType) {
+            MemoizedResult result;
+            switch (attributeType) {
+                case INDEX_ONLY_FIELD:
+                    result = indexedOnly;
+                    break;
+                case INDEXED_FIELD:
+                    result = indexed;
+                    break;
+                case REVERSE_INDEXED_FIELD:
+                    result = reverseIndexed;
+                    break;
+                case TOKENIZED_FIELD:
+                    result = tokenized;
+                    break;
+                case REVERSE_TOKENIZED_FIELD:
+                    result = reverseTokenized;
+                    break;
+                case STORED_FIELD:
+                    result = stored;
+                    break;
+                default:
+                    throw new IllegalArgumentException("Undefined attribute type: " + attributeType);
+            }
+            return result;
+        }
+
+        private class MemoizedResult {
+            private boolean resultEvaluated;
+            private boolean result;
+
+            private boolean getResultOrEvaluate(Predicate<String> evaluateFn) {
+                if (!resultEvaluated) {
+                    result = evaluateFn.test(fieldName);
+                    resultEvaluated = true;
+                }
+                return result;
+            }
         }
     }
 }
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CachedFieldConfigHelper.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CachedFieldConfigHelper.java
@@ -1,0 +1,75 @@
+package datawave.ingest.data.config;
+
+import java.util.EnumMap;
+import java.util.Map;
+import java.util.function.Function;
+
+import org.apache.commons.collections4.map.LRUMap;
+
+import com.google.common.annotations.VisibleForTesting;
+
+public class CachedFieldConfigHelper implements FieldConfigHelper {
+    private final FieldConfigHelper underlyingHelper;
+    private final Map<String,ResultEntry> resultCache;
+
+    enum AttributeType {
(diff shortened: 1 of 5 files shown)


warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CachedFieldConfigHelper.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @VisibleForTesting
+    boolean getOrEvaluate(AttributeType attributeType, String fieldName, Function<String,Boolean> evaluateFn) {
+        return resultCache.computeIfAbsent(fieldName, ResultEntry::new).resolveResult(attributeType, evaluateFn);
+    }
+
+    private static class ResultEntry {
+        private final String fieldName;
+        private final EnumMap<AttributeType,Boolean> resultMap;
```
ivakegg: Do we not need to limit the size of this cache?

dtspence: I think the map here will be limited to the size of the AttributeType (which was in the current class - maybe we should change the name). I tried an alternate version which used a switch statement on the enum and individual boolean variables, but it did not seem to impact the timing w/JMH. I will confirm the timing again - and we can also update to have a switch statement and explicit boolean results.

warehouse/ingest-core/src/test/java/datawave/ingest/data/config/CachingFieldConfigHelperTest.java
```diff
@@ -0,0 +1,99 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+public class CachingFieldConfigHelperTest {
+    @SuppressWarnings("unchecked")
+    @Test
+    public void testCachingBehaviorWillCallBaseMethods() {
+        // @formatter:off
+        Stream.of(new Object[][] {
+            new Object[] {
```
jschmidt10: Are we actually calling all these pairs? 

You are iterating over an array of arrays but there's only 1 inner array with 12 elements in it. 
It seems like we want 6 inner arrays, each with two elements {action, verify}
I think we're actually only testing the isIndexOnlyField pair here.

jschmidt10: It also has me wondering if maybe it's just easier to do the manual calls in the test instead of dealing with all the method references and such.

```
cachedHelper.isIndexOnlyField(fieldName);
verify(mockHelper).isIndexOnlyField(fieldName);

cachedHelper.isIndexedField(fieldName);
verify(mockHelper).isIndexedField(fieldName);

// ...
```

dtspence: That is a bug, thank you for catching that. In regards to the manual tests - I updated, thank you.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CachedFieldConfigHelper.java
dtspence: An approach with a switch statement and primitive boolean variables was slightly quicker on second pass. I updated the logic, please let me know if we should evaluate the previous approach.

dtspence: Discussed additional performance ideas, going to run additional tests.

dtspence: The LinkedHashMap implementation tested slightly better in benchmarks. Please indicate if we should switch the implementation or restructure the representation.

jschmidt10 (dismissed):

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CachedFieldConfigHelper.java
```diff
@@ -0,0 +1,208 @@
...
+
+    private CachedEntry debugCachedEntryCreation(String fieldName) {
+        debugFieldComputes.incrementAndGet();
+        debugFieldUnique.add(fieldName);
+        return new CachedEntry(fieldName);
+    }
+
+    private void debugLogState() {
+        if (resultCache.hasLimitExceeded()) {
+            log.info("Field cache LRU limit exceeded [limit={}, debug={}, size={}, uniq={}]",
```
jschmidt10: Since this feature is "debugLogState" I was expecting this to be debug level logging. Not a deal breaker but I'd vote to make it debug.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CachedFieldConfigHelper.java
```diff
@@ -0,0 +1,208 @@
...
+            this.maxSize = maxSize;
+        }
+
+        boolean hasLimitExceeded() {
+            // thread-safe
+            return limitExceeded;
+        }
+
+        protected boolean removeEldestEntry(Map.Entry<K,V> eldest) {
+            boolean localLimitExceeded = size() > maxSize;
```
jschmidt10, line 175: Is `size()` threadsafe? Wondering if this could return a stale size in a multi-threaded scenario. Not sure how much we care.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
jalphonso, line 1: The `setup()` method is already too long. There are blocks of logic preceded by a comment explaining what's to come. I know this is a bit out of scope but it would help with readability if it were refactored so that each of those commented blocks turn into meaningfully named methods.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
```diff
@@ -255,10 +259,19 @@ public void setup(Configuration config) {
         // Load the field helper, which takes precedence over the individual field configurations
         final String fieldConfigFile = config.get(this.getType().typeName() + FIELD_CONFIG_FILE);
         if (fieldConfigFile != null) {
-            if (log.isDebugEnabled()) {
-                log.debug("Field config file " + fieldConfigFile + " specified for: " + this.getType().typeName() + FIELD_CONFIG_FILE);
+            final boolean fieldConfigCacheEnabled = config.getBoolean(this.getType().typeName() + FIELD_CONFIG_CACHE_ENABLED, false);
```
jalphonso, line 262: If the cache truly is the better option do we really want it to be feature based?

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
```diff
@@ -255,10 +259,19 @@ public void setup(Configuration config) {
...
+            final boolean fieldConfigCacheLimitDebug = config.getBoolean(this.getType().typeName() + FIELD_CONFIG_CACHE_KEY_LIMIT_DEBUG, false);
+            final int fieldConfigCacheLimit = config.getInt(this.getType().typeName() + FIELD_CONFIG_CACHE_KEY_LIMIT, 100);
```
jalphonso: how was the default of 100 chosen?

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
```diff
@@ -255,10 +259,19 @@ public void setup(Configuration config) {
         // Load the field helper, which takes precedence over the individual field configurations
         final String fieldConfigFile = config.get(this.getType().typeName() + FIELD_CONFIG_FILE);
         if (fieldConfigFile != null) {
-            if (log.isDebugEnabled()) {
-                log.debug("Field config file " + fieldConfigFile + " specified for: " + this.getType().typeName() + FIELD_CONFIG_FILE);
```
jalphonso: Should probably restore this

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CachedFieldConfigHelper.java
```diff
@@ -0,0 +1,208 @@
...
+                    .setPriority(NORM_PRIORITY)
+                    .setDaemon(true)
+                    .setNameFormat("CachedFieldConfigHelper.DebugState")
+                    .build()
+                // formatter:off
+            );
+            this.debugStateExecutor.scheduleAtFixedRate(this::debugLogState, DEFAULT_DEBUG_STATE_SECS, DEFAULT_DEBUG_STATE_SECS, SECONDS);
+        } else {
+            this.debugStateExecutor = null;
+        }
```
jalphonso: Should this debug related code be left in?

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CachedFieldConfigHelper.java
```diff
@@ -0,0 +1,208 @@
...
+        debugFieldUnique.add(fieldName);
+        return new CachedEntry(fieldName);
+    }
+
+    private void debugLogState() {
+        if (resultCache.hasLimitExceeded()) {
+            log.info("Field cache LRU limit exceeded [limit={}, debug={}, size={}, uniq={}]",
+                limit, debugFieldComputes.get(), debugFieldUnique.size(), debugLimitsEnabled);
+        }
+    }
```
jalphonso, line 124: more debug related code

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CachedFieldConfigHelper.java
```diff
@@ -0,0 +1,208 @@
...
+    @Override
+    public boolean isReverseTokenizedField(String fieldName) {
+        return getFieldResult(AttributeType.REVERSE_TOKENIZED_FIELD, fieldName, underlyingHelper::isReverseTokenizedField);
+    }
+
+    @VisibleForTesting
+    boolean getFieldResult(AttributeType attributeType, String fieldName, Predicate<String> fn) {
+        CachedEntry ce = !debugLimitsEnabled ?
+            resultCache.computeIfAbsent(fieldName, CachedEntry::new) :
+            resultCache.computeIfAbsent(fieldName, this::debugCachedEntryCreation);
```
jalphonso, line 104: debug code?

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CachedFieldConfigHelper.java
```diff
@@ -0,0 +1,208 @@
...
+        }
+
+        protected boolean removeEldestEntry(Map.Entry<K,V> eldest) {
+            boolean localLimitExceeded = size() > maxSize;
+            if (localLimitExceeded) {
+                limitExceeded = true;
+            }
+            return localLimitExceeded;
+        }
+    }
```
jalphonso, line 147: eldest entry isn't being removed

jalphonso (changes requested):

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
dtspence: I used a value that felt was a reasonable/minimal default and also fit with the sample datasets. I can change it to something different, would a higher value at `1_000` or `10_000` be better?

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
dtspence: It was elevated to an info - however, it is logged twice - there is another location within `XMLFieldHelper`  which outputs the file as info. I will update this back to debug.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
dtspence: My preference was to make it feature/opted-into based on the need to select an appropriate cache value and also so the capability could be turned off in the event of unexpected behavior.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CachedFieldConfigHelper.java
dtspence: I can make a comment and/or refactor - the intent was to have the limitExceeded variable be thread-safe for the debug log output.

EDIT: The refactored the implementation no longer uses a thread for the debug/diagnostic log messages.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CachedFieldConfigHelper.java
dtspence: I can update the name to be something different and/or refactor. The intent was to trace internals to view and help properly size the cache. It was meant to be conditionally enabled, so not to impact normal execution. I evaluated a way to link to Hadoop metrics, but I didn't immediately see a straight forward path to connect. In respect to the name, would generally changing this to an optional `view-state` semantic be better or is there another convention that would apply better?

EDIT: we could also update tracing/state logic to all be inline (i.e. remove threading), which when enabled would optionally log out the same information. That would simplify the maintenance/logic.

EDIT#2: refactored the logic to no longer have a background thread to log the message.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CachedFieldConfigHelper.java
dtspence: I can update - I set the level this way so it would show up in logs without having to also adjust logger/levels.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
jalphonso: I don't have a recommendation on size. I was just curious.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
jalphonso: With the new diagnostic code if it makes better sense to have it as info then I'm ok with it.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
jalphonso: That's reasonable. I suggest that once the confidence level is high, we remove the conditional logic.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CachedFieldConfigHelper.java
jalphonso: The diagnostic nomenclature works for me.

dtspence:
There is an alternate branch that simplified the approach - closing the PR and will re-open on the new branch.

PR #2620, opened 2024-10-28
Bad seek range

jack-gitdev:
Updated filtering so index-only fields are omitted from seek ranges related to event queries.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
+++ b/warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
@@ -319,6 +319,8 @@ public class QueryOptions implements OptionDescriber {
 
     // filter for any key type (fi, event, tf)
     protected EventDataQueryFilter evaluationFilter;
+    protected EventDataQueryFilter fiEvaluationFilter;
+    protected EventDataQueryFilter eventEvaluationFilter;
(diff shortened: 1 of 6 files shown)


apmoriarty (approved):

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -319,6 +319,8 @@ public class QueryOptions implements OptionDescriber {
 
     // filter for any key type (fi, event, tf)
     protected EventDataQueryFilter evaluationFilter;
+    protected EventDataQueryFilter fiEvaluationFilter;
+    protected EventDataQueryFilter eventEvaluationFilter;
```
drewfarris, line 323: I see we now have `evaluationFilter`, `fiEvaluationFilter` and `eventEvaluationFilter`, and the comment mentions that we're filtering for `fi`, `event` and `tf`. Does this mean that the `evaluationFilter` is only used for tf's or is it used in other cases as well? 

Questions to consider:
* Is the `evaluationFilter` still used for anything?
* Is this an opportunity to rename the `evaluationFilter` to better explain how it is now used now that the two other filters have been added?
*

drewfarris (approved):
One minor question/comment - otherwise lgtm.

PR #2621, opened 2024-11-01
Edge precondition support for ER and NR nodes

hlgp:


alerman (dismissed):

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/evaluation/EdgePreconditionArithmetic.java
```diff
@@ -417,4 +424,40 @@ public Object notEquals(Object left, Object right) {
 
         return matches;
     }
+
+    public Object notContains(Object left, Object right) {
```
ivakegg, line 448: Comparing this with the notEquals method, I expected that they would essentially follow each other.  However notEquals handles the left not being a collection and the right being a collection as a separate case.  Is it that the notEquals method can be simplified, this method needs to be expanded, or I am missing something (which is always the more likely case)?

hlgp: I never hit that case in testing with the way we define preconditions, but I can add it

apmoriarty (approved):

ivakegg (approved):

PR #2623, opened 2024-11-01
Task/consistent java version and distribution

ivakegg:
Use environment variables for consistent java version and distribution across github actions


drewfarris (approved):

alerman:
ENV doesnt play nicely with workflow_call jobs. I had it this way originally but the downstream repos will be unable to access the env. It can be set per job, but if we do that, we dont really improve anything. Closing.

PR #2411, opened 2024-06-10
Language Expansion Visitor

apmoriarty:
- Added a visitor to support language-based expansion of query terms
- LanguageAwareAnalyzer extends Lucene's Analyzer class.
- Default implementations for most Lucene analyzers provided.
- QueryParameters and QueryOptions can enable language analysis and disable stemming, lemmatization, bigramming or unigramming on a per-language basis

PR #2629, opened 2024-11-15
Improves Lucene query tokenization to properly handle Analyzers that produce alternate tokens at the same position

drewfarris:


drewfarris:
In the future, the functions in `CustomAnalyzerQueryNodeProcessor` should be integrated into the JEXL processing pipeline in a way similar to what's sketched out in #2411 - that's a larger scope change that I'm able to dig into at the moment because we need to take all of the phrase handling code and make that work with the proposed `LanguageExpansionVisitor`.

warehouse/query-core/src/main/java/datawave/query/language/processor/lucene/CustomAnalyzerQueryNodeProcessor.java
```diff
@@ -408,4 +438,68 @@ private String toStringEscaped(UnescapedCharSequence unescaped) {
...
+        public VariantBuilder appendVariant(String input) {
+            if (variants.isEmpty()) {
+                append(input);
+            } else {
+
+                List<List<String>> newVariants = new ArrayList<>();
+
+                for (List<String> b : variants) {
+                    // create a new variant of all the existing strings, replacing the
+                    List<String> newVariant = new ArrayList<>(b);
```
FineAndDandy, line 474: Since there is no random access, might it make sense to use a a LinkedList rather than an array list to prevent reallocating the list for each new variant? It seems like this could get expensive.

FineAndDandy (approved):
suggestion on List impl, otherwise looks good to me

austin007008 (approved):
lgtm

PR #2640, opened 2024-11-19
Improve Docker testAll.sh Script

foster33:
This PR aims to improve `docker/scripts/testAll.sh`

Before this change, the script had a few issues:
- If a test failed a configurable number of times, it would not let the other tests run.
- There were timeouts that caused the script to run longer than necessary.
- Test result outputs could be hard to understand when multiple attempts occurred.

This PR adds:
- Improved formatting of individual test results.
- Information about the amount of time each test takes.
- An overall test summary that shows the passing and failed tests.
- Allows all tests to run no matter the status of other tests.
- Removes unnecessary timeout after all 3 attempts already failed.
- An option to skip the cleanup stage, keeping all logs from tests ( `-noCleanup` )
- Adds more descriptive fail messages that give expected number of events / files.

I tried to improve this script to look like the output from `datawaveWebTest`, which I found to be quite nice.


foster33:
**Here is an example of the testAll.sh output before the change:**
_The script ends after this failure and does not let the other tests run._
image

**Here are some images of the testAll.sh output after the change:**
_Passing & failing tests:_
image
_Summary (not related to test results above):_
image

alerman (approved):

alerman (dismissed):

docker/scripts/testAll.sh
```diff
@@ -129,7 +190,11 @@ runTest query.sh 12 2
...
+# Gives option to skip the cleanup stage
+if [ "${1}" == "-noCleanup" ] ; then
+    printTestSummary
+    exit 0
+fi
 
-echo
-echo "All tests SUCCEEDED!"
+printTestSummary
+$SCRIPT_DIR/cleanup.sh
```
ivakegg: We should only cleanup all of the evidence if everything passed.  Otherwise we need a way to go and see what happened.

foster33: I agree. Good idea. The latest commit adds a rework of the cleanup script to only delete the logs of tests that have passed. Logs related to tests that fail are kept.

foster33: Could even remove `cleanup.sh` and put the cleanup in `testAll.sh` since that is the only script that uses the cleanup.

alerman (approved):

apmoriarty (approved):

PR #2645, opened 2024-11-21
Implement query validation endpoint

### LINKED ISSUE #2585: query validation endpoint
Currently we have a plan endpoint that syntactically validates the query, however there are many ways a user can create a valid query that is most obviously not what they had intended.  We need an end point where we can configure rules that find potential query issues and return text describing what those may be.

Here is a list if issues with examples that we should be able to detect.  Most all of these based on LUCENE queries:
1) Parens before field or completely missing parens when using ORs
     FIELD:1234 OR 5678 -> should be FIELD:(1234 OR 5678)
     (FIELD:1234 OR 5678) -> should be FIELD:(1243 OR 5687)
2) Phrases not quoted (where we would automatically AND)
     anywhere that the boolean operator is missing
     FIELD:term1 term2 -> should be FIELD:"term1 term2"
3) Index only fields used in a #INCLUDE or #EXCLUDE
4) Fields that do not exist in the system
5) A configurable set of rules:
    If you see field X, then output message Y
    If you see pattern X (regex), then output message Y
6) Missing arguments to #INCLUDE or #EXCLUDE when the first argument is a boolean:
    #INCLUDE(OR,value) -> should be #INCLUDE(OR, FIELD, value)
7) Usage of NOT without using parens (NOT takes precidence so the resulting plan could be quite incorrect)
    FIELD1:abc OR FIELD2:def NOT FIELD3:123 -> should be (FIELD1:abc OR FIELD2:def) NOT FIELD3:123
8) #TIME functions used with fields that are not typed as a date
9) Unfielded terms in general should probably be flagged as potential issues
10) Wildcards inside of phrases
    FIELD:"abc\*def" -> either the * is incorrect or they meant FIELD:/abc*def/
11) Using the wrong quote (` instead of ')
12) Proximity wher the number is smaller than the number of terms
    FIELD:"term1 term2 term3"~1 -> the 1 should be 3 or greater
13) Not escaping special characters
    FIELD:abc:def -> should be FIELD:abc\\:def
14) Searching for a range of numbers against a non-numeric field

lbschanno:
Implement the query/{logicName}/validate endpoint. This feature supports the ability to configure validation rules that will validate LUCENE and JEXL queries against a number of criteria and provide meaningful feedback to customers.

Closes #2585

Each validation criteria is implemented via a corresponding `datawave.query.rules.QueryRule implementation`. 

### `datawave.query.rules.AmbiguousNotRule`
Checks LUCENE queries for usage of NOT without using parens (NOT takes precedence so the resulting plan could be quite incorrect):
-	`FIELD1:abc OR FIELD2:def NOT FIELD3:123` -> should be `(FIELD1:abc OR FIELD2:def) NOT FIELD3:123` 

Sample bean config:
```xml
<bean id="AmbiguousNotRule" scope="prototype" class="datawave.query.rules.AmbiguousNotRule">  
    <property name="name" value="Check for Ambiguous Usage of NOT"/>  
</bean>
```
Sample message for query `FIELD1:abc FIELD2:def NOT FIELD:ghi`:
```
Ambiguous usage of NOT detected with multiple unwrapped preceding terms: "FIELD1:abc AND FIELD2:def NOT" should be "(FIELD1:abc AND FIELD2:def) NOT".
```
### `datawave.query.rules.AmbiguousOrPhrasesRule`
Checks LUCENE queries for any parens before field or completely missing parens when using ORs:
- `FIELD:1234 OR 5678` -> should be `FIELD:(1234 OR 5678)`  
- `(FIELD:1234 OR 5678)` -> should be `FIELD:(1243 OR 5687)`

Sample bean config:
```xml
<bean id="AmbiguousOrPhrasesRule" scope="prototype" class="datawave.query.rules.AmbiguousOrPhrasesRule">  
    <property name="name" value="Check for Unfielded Terms That Could Be Wrapped"/>  
</bean>
```
Sample message for query `FOO:abc OR def OR efg`:
```
Ambiguous unfielded terms OR'd with fielded term detected: FOO:abc OR def OR efg Recommended: FOO:(abc OR def OR efg)
```
### `datawave.query.rules.AmbiguousUnquotedPhrasesRule`
Checks LUCENE queries for any phrases not quoted (where we would automatically AND) anywhere that the boolean operator is missing  
 - `FIELD:term1 term2` -> should be `FIELD:"term1 term2"`

Sample bean config:
```xml
<bean id="AmbiguousUnquotedPhrasesRule" scope="prototype" class="datawave.query.rules.AmbiguousUnquotedPhrasesRule">  
    <property name="name" value="Check for Unfielded Terms That Could Be Quoted"/>  
</bean>
```
Sample message for query `FOO:abc def ghi`:
```
Ambiguous unfielded terms AND'd with fielded term detected: FOO:abc AND def AND ghi. Recommended: FOO:"abc def ghi"
```
### `datawave.query.rules.FieldExistenceRule`
Checks the query for any fields that do not exist in the data dictionary. Special fields may be specified that will not be flagged if they do not exist in the data dictionary.
Sample bean config:
```xml
<bean id="FieldExistenceRule" scope="prototype" class="datawave.query.rules.FieldExistenceRule">  
    <property name="name" value="Check Field Existence"/>  
    <property name="specialFields">  
        <list value-type="java.lang.String">   
            <value>_ANYFIELD_</value>  
            <value>_NOFIELD_</value>  
        </list>  
    </property>  
</bean>
```
Sample message for query `FOO:abc AND MISSING1:def AND MISSING2:efg`:
```
Fields not found in data dictionary: MISSING1, MISSING2
```
### `datawave.query.rules.FieldPatternPresenceRule`
Returns configured messages if we see field X or Y.
Sample bean config:
```xml
<bean id="FieldPatternPresenceRule" scope="prototype" class="datawave.query.rules.FieldPatternPresenceRule">  
    <property name="name" value="Check Presence of Field or Pattern"/>  
    <property name="fieldMessages">  
        <map key-type="java.lang.String" value-type="java.lang.String">  
            <entry key="_ANYFIELD_" value="Detected presence of _ANYFIELD_"/>  
        </map>  
    </property>  
    <property name="patternMessages">  
        <map key-type="java.lang.String" value-type="java.lang.String">  
            <entry key=".*" value="Detected pattern '.*' that will match everything"/>  
        </map>  
    </property>  
</bean>
```
Sample messages for query `_ANYFIELD_:abc || FOO:.*`
```
Detected presence of _ANYFIELD_
Detected pattern '.*' that will match everything
```
### `datawave.query.rules.IncludeExcludeArgsRule`
Checks LUCENE queries for any `#INCLUDE` or `#EXCLUDE` functions with an incorrect number of arguments, taking into account whether a boolean was given as the first argument.
Sample bean config:
```xml
<bean id="IncludeExcludeArgsRule" scope="prototype" class="datawave.query.rules.IncludeExcludeArgsRule">  
    <property name="name" value="Validate Args of #INCLUDE and #EXCLUDE"/>  
</bean>
```
Sample message for query: `#INCLUDE(OR, 'value')`
```
Function #INCLUDE supplied with uneven number of arguments after the first boolean arg OR. Must supply field/value after the boolean, e.g. #INCLUDE(OR, FIELD, 'value') or #INCLUDE(OR, FIELD1, 'value1',' FIELD2, 'value2').
```
### `datawave.query.rules.IncludeExcludeIndexFieldsRule`
Checks queries for any usage of `#INCLUDE` or `#EXCLUDE` with index fields in them.
Sample bean config:
```xml
<bean id="IncludeExcludeIndexFieldsRule" scope="prototype" class="datawave.query.rules.IncludeExcludeIndexFieldsRule">  
    <property name="name" value="Check #INCLUDE and #EXCLUDE for Indexed Fields"/>  
</bean>
```
Sample message for query: `FOO:abc AND #INCLUDE(INDEXED1, 'abc')`:
```
Indexed fields found within the function filter:includeRegex: INDEXED1
```
### `datawave.query.rules.InvalidQuoteRule`
Checks LUCENE queries for the usage of the quote \` instead of '.
Sample bean config:
```xml
<bean id="InvalidQuoteRule" scope="prototype" class="datawave.query.rules.InvalidQuoteRule">  
    <property name="name" value="Check for Invalid Quote"/>  
</bean>
``` 
Sample message for query ``FOO:'abc' AND BAR:`def` ``:
```
Invalid quote ` used in phrase "BAR:`abc`". Use ' instead.
```
### `datawave.query.rules.MinimumSlopProximityRule`
Checks queries for any cases where the proximity is smaller than the number of terms.
- `FIELD:"term1 term2 term3"~1` -> the 1 should be 3 or greater.

Sample bean config:
```xml
<bean id="MinimumSlopProximityRule" scope="prototype" class="datawave.query.rules.MinimumSlopProximityRule">  
    <property name="name" value="Validate Slop Proximity"/>  
</bean>
```
Sample message for query `FOO:"term1 term2 term3"~2`:
```
Invalid slop proximity, the 2 should be 3 or greater: FOO:"term1 term2 term3"~2
```
### `datawave.query.rules.NumericValueRule`
Checks queries for any cases where a numeric value is used against a non-numeric field.
Sample bean config:
```xml
<bean id="NumericValueRule" scope="prototype" class="datawave.query.rules.NumericValueRule">  
    <property name="name" value="Validate Numeric Values Only Given for Numeric Fields"/>  
</bean>
```
Sample message for query `AGE:1 AND NON_NUMERIC_FIELD:4`
```
Numeric values supplied for non-numeric field(s): NON_NUMERIC_FIELD
```
### `datawave.query.rules.TimeFunctionRule`
Checks queries for any usage of `#TIME_FUNCTION` with any non-date fields.
Sample bean config:
```xml
<bean id="TimeFunctionRule" scope="prototype" class="datawave.query.rules.TimeFunctionRule">  
    <property name="name" value="Validate #TIME_FUNCTION has Date Fields"/>  
</bean>
```
Sample message for query `filter:timeFunction(DATE1,NON_DATE2,'-','>',2522880000000L)`:
```
Function #TIME_FUNCTION (filter:timeFunction) found with fields that are not date types: NON_DATE2
```
### `datawave.query.rules.UnescapedSpecialCharsRule`
Checks queries for fields and patterns with unescaped special characters. Characters that should be considered exceptions to the rule may be configured for literals and patterns. The regex-reserved characters `. + * ? ^ $ ( ) [ ] { } | \` will always be treated as exceptions for patterns. Whether whitespace characters should be considered special characters is also configurable.
Sample bean config:
```xml
<bean id="UnescapedSpecialCharsRule" scope="prototype" class="datawave.query.rules.UnescapedSpecialCharsRule">  
    <property name="name" value="Check for Unescaped Special Characters"/>  
    <property name="literalExceptions">  
        <list value-type="java.lang.Character">  
            <value>?</value>  
        </list>  
    </property>  
    <property name="patternExceptions">  
        <list value-type="java.lang.Character">  
            <value>_</value>  
        </list>  
    </property>  
    <property name="escapedWhitespaceRequiredForLiterals" value="false"/>  
    <property name="escapedWhitespaceRequiredForPatterns" value="false"/>  
</bean>
```
Sample message for query `FOO == 'ab^c' || FOO =~ 'ab=cd'`:
```
Literal string "ab^c" has the following unescaped special character(s): '^'");  
Regex pattern "ab=cd" has the following unescaped special character(s): '='"
```
As configured in the bean example above, the character `?` would not be required to be escaped for literal strings, and the character `=` would not be required to be escaped.

### `datawave.query.rules.UnescapedWildcardsInPhrasesRule`
Checks LUCENE queries for any quoted phrases that have an unescaped wildcard in them.
Sample bean config:
```xml
<bean id="UnescapedWildcardsInPhrasesRule" scope="prototype" class="datawave.query.rules.UnescapedWildcardsInPhrasesRule">  
    <property name="name" value="Check Quoted Phrases for Unescaped Wildcard"/>  
</bean>
```
Sample message for `FOO:"ab*"`:
```
Unescaped wildcard found in phrase FOO:"ab*". Wildcard is incorrect, or phrase should be FOO:/ab*/
```
### `datawave.query.rules.UnfieldedTermsRule`
Checks LUCENE queries for any unfielded terms.
Sample bean config:
```xml
<bean id="UnfieldedTermsRule" scope="prototype" class="datawave.query.rules.UnfieldedTermsRule">  
    <property name="name" value="Check for Unfielded Terms"/>  
</bean>
```
Sample message for `FOO:abc def "efg"`
```
Unfielded term def found.
Unfielded term "efg" found.
```

## Activating Rules
Rules must be configured in the QueryLogicFactory.xml and added to the list of rules configured for the `validationRules` property of a ShardQueryLogic.
Sample bean config:
```xml
	<!-- Rules added to validationRules via referencing bean definition samples. -->
    <property name="docAggregationThresholdMs" value="-1" />  
    <property name="tfAggregationThresholdMs" value="-1" />  
    <property name="validationRules">  
        <list value-type="datawave.query.rules.QueryRule">  
            <ref bean="InvalidQuoteRule"/>  
            <ref bean="UnfieldedTermsRule"/>  
            <ref bean="UnescapedWildcardsInPhrasesRule"/>  
            <ref bean="AmbiguousNotRule"/>  
            <ref bean="AmbiguousOrPhrasesRule"/>  
            <ref bean="AmbiguousUnquotedPhrasesRule"/>  
            <ref bean="MinimumSlopProximityRule"/>  
            <ref bean="IncludeExcludeArgsRule"/>  
            <ref bean="FieldExistenceRule"/>  
            <ref bean="UnescapedSpecialCharsRule"/>  
            <ref bean="FieldPatternPresenceRule"/>  
            <ref bean="IncludeExcludeIndexFieldsRule"/>  
            <ref bean="NumericValueRule"/>  
            <ref bean="TimeFunctionRule"/>  
        </list>  
    </property>  
</bean>
```
Some notes on rules:
- All rules may be configured with a name. Names are not required to be unique. If a name is not specified, the name will be the name of the rule class.
- Rules will be executed in the order that they are listed in the `validationRules` property.
- Rules will never be executed twice for a validation operation.
- Some rules will only execute against LUCENE queries. The rest will execute on the resulting parsed JEXL query. LUCENE-only rules:
```
AmbiguousNotRule
AmbiguousOrRule
AmbiguousUnquotedPhrasesRule
IncludeExcludeArgsRule
InvalidQuoteRule
MinimumSlopProximityRule
UnescapedWildcardsInPhrasesRule
UnfieldedTermsRule
```	
 
## REST Response Schema
The REST response schema will be as follows:
```json5
{
	// A list of names of rules that successfully executed 
	// against the query, in the order that they executed.
	"ExecutedRules": [
		"<RuleName>",
		// ...
	],
	// Whether any issues were flagged.
	"HasResults": true,
	"LogicName": "<logicName>",
	"OperationTimeMS": 10,
	"QueryId": "<uuid>",
	// The flagged issues in the order that they were flagged.
	"Results": [
		{
			"RuleName":"<RuleName>",
			"Messages":[
				"A message about a flagged issue",
				// ...
			],
			// This is populated only if an exception occurred 
			// while a rule was attempting to validate a query.
			"Exception": {
				"Cause": "",
				"Message":""
			}
		},
		// ...
	],
	"Exceptions": [
		{
			"Cause":"",
			"Code":"",
			"Message":""
		},
		// ...
	]
}
```


warehouse/query-core/src/main/java/datawave/query/language/parser/QueryParser.java
```diff
@@ -1,5 +1,7 @@
 package datawave.query.language.parser;
 
+import org.apache.lucene.queryparser.flexible.core.parser.SyntaxParser;
```
ivakegg: Is this needed?

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/JexlControlledQueryParser.java
```diff
@@ -10,6 +10,7 @@
 import org.apache.commons.jexl3.parser.ASTIdentifier;
 import org.apache.commons.jexl3.parser.JexlNode;
 import org.apache.commons.lang.StringUtils;
+import org.apache.lucene.queryparser.flexible.core.parser.SyntaxParser;
```
ivakegg: Is this needed?

core/query/src/main/java/datawave/core/query/logic/QueryLogic.java
```diff
@@ -7,8 +7,11 @@
 
 import org.apache.accumulo.core.client.AccumuloClient;
 import org.apache.accumulo.core.security.Authorizations;
+import org.apache.commons.collections4.Transformer;
 import org.apache.commons.collections4.iterators.TransformIterator;
 
+import com.google.common.collect.HashMultimap;
```
ivakegg: Unused import

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -1344,7 +1395,163 @@ public void close() {
...
+                            + (this.getSettings() == null ? "empty" : this.getSettings().getId()) + ')');
+        }
+
+        // todo - maybe unnecessary, we could just return early if no rules configured.
+        if (validationRules == null || validationRules.isEmpty()) {
+            throw new IllegalStateException("Query validation rules not configured.");
+        }
+
+        // todo - verify if we should allow these to be configured, or always stick to an established default.
+        this.config.setExpandFields(expandFields);
```
ivakegg: I am confused here.  What does expanding fields and values have to do with the query validation?  I do not believe those concepts are used at all.  They only apply to query planning which we are not doing here.  We are only validating the initial query.

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -1344,7 +1395,163 @@ public void close() {
...
+        }
+
+        // Ensure any nodes with the literal on the left and the identifier on the right are re-ordered.
+        config.setQueryTree(InvertNodeVisitor.invertSwappedNodes(config.getQueryTree()));
+        // Uppercase all identifiers.
+        config.setQueryTree(ShardQueryUtils.upperCaseIdentifiers(metadataHelper, config, config.getQueryTree()));
+        // Flatten the tree.
+        config.setQueryTree(TreeFlatteningRebuilder.flatten(config.getQueryTree()));
+        // Apply the query model.
+        config.setQueryTree(ShardQueryUtils.applyQueryModel(config.getQueryTree(), config, metadataHelper.getAllFields(config.getDatatypeFilter()),
```
ivakegg, line 1527: Do we need to apply the query model for query validation?  There may be a good reason, I am just unaware of which rule(s) require this.

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -3010,11 +3012,121 @@ private void updateQueryParams(Query q, String queryLogicName, String query, Dat
     @Path("/{logicName}/validate")
     @Interceptors({RequiredInterceptor.class, ResponseInterceptor.class})
     @Timed(name = "dw.query.validateQuery", absolute = true)
-    public GenericResponse<String> validateQuery(@Required("logicName") @PathParam("logicName") String queryLogicName,
+    public QueryValidationResponse validateQuery(@Required("logicName") @PathParam("logicName") String queryLogicName,
                     MultivaluedMap<String,String> queryParameters) {
-        GenericResponse<String> response = new GenericResponse<>();
-        response.setMessages(Collections.singletonList("Query validator coming soon."));
-        throw new DatawaveWebApplicationException(new UnsupportedOperationException("Query validator not implemented"), response, 501);
+        QueryData queryData = validateQuery(queryLogicName, queryParameters, null);
```
ivakegg: I am realizing that this validateQuery method can be confused with the existing private validateQuery method being call herein.  I am thinking that we keep your method as is, but change this underlying method to something like validateQueryParameters as that is really all it is doing.

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -3010,11 +3012,121 @@ private void updateQueryParams(Query q, String queryLogicName, String query, Dat
...
+            try {
+                // The query should be transient.
+                qp.setPersistenceMode(QueryPersistence.TRANSIENT);
+                Map<String,List<String>> optionalQueryParameters = qp.getUnknownParameters(MapUtils.toMultivaluedMap(queryParameters));
+                query = persister.create(queryData.userDn, queryData.dnList, marking, queryLogicName, qp, MapUtils.toMultivaluedMap(optionalQueryParameters));
+                auditType = queryData.logic.getAuditType();
+            } finally {
+                queryParameters.add(PrivateAuditConstants.AUDIT_TYPE, auditType.name());
+
+                if (!auditType.equals(AuditType.NONE)) {
```
ivakegg, line 3047: Why are we auditing this request?  This is not needed.

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -3010,11 +3012,121 @@ private void updateQueryParams(Query q, String queryLogicName, String query, Dat
...
+            if (queryParameters.containsKey(EXPAND_VALUES)) {
+                expandValues = Boolean.valueOf(queryParameters.getFirst(EXPAND_VALUES));
+            }
+
+            AuditType auditType = queryData.logic.getAuditType();
+            try {
+                // The query should be transient.
+                qp.setPersistenceMode(QueryPersistence.TRANSIENT);
+                Map<String,List<String>> optionalQueryParameters = qp.getUnknownParameters(MapUtils.toMultivaluedMap(queryParameters));
+                query = persister.create(queryData.userDn, queryData.dnList, marking, queryLogicName, qp, MapUtils.toMultivaluedMap(optionalQueryParameters));
```
ivakegg, line 3042: Why are we persisting this query?  This is not needed.

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -3010,11 +3012,121 @@ private void updateQueryParams(Query q, String queryLogicName, String query, Dat
...
+        Query query = null;
+        AccumuloClient client = null;
+        AccumuloConnectionFactory.Priority priority;
+        RunningQuery runningQuery = null;
+
+        try {
+            // by default we will expand the fields but not the values.
+            boolean expandFields = true;
+            boolean expandValues = false;
+            if (queryParameters.containsKey(EXPAND_FIELDS)) {
```
ivakegg: I don't think expanding fields or values has anything to do with query validation.  Those are concepts specific to retrieving the query plan.

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -3010,11 +3012,121 @@ private void updateQueryParams(Query q, String queryLogicName, String query, Dat
...
+                        log.error("Error auditing query", e);
+                        QueryException qe = new QueryException(DatawaveErrorCode.QUERY_AUDITING_ERROR, e);
+                        response = new QueryValidationResponse();
+                        response.addException(qe);
+                        throw qe;
+                    }
+                }
+            }
+
+            priority = queryData.logic.getConnectionPriority();
```
ivakegg, line 3079: Do we really need a connection factory client setup for this validation?

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
```diff
@@ -3010,11 +3012,121 @@ private void updateQueryParams(Query q, String queryLogicName, String query, Dat
...
+            boolean expandFields = true;
+            boolean expandValues = false;
+            if (queryParameters.containsKey(EXPAND_FIELDS)) {
+                expandFields = Boolean.valueOf(queryParameters.getFirst(EXPAND_FIELDS));
+            }
+            if (queryParameters.containsKey(EXPAND_VALUES)) {
+                expandValues = Boolean.valueOf(queryParameters.getFirst(EXPAND_VALUES));
+            }
+
+            AuditType auditType = queryData.logic.getAuditType();
```
ivakegg: There is no need to audit this request.  I believe this entire block can be removed (through line 3077)

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -1344,7 +1395,163 @@ public void close() {
...
+    public Object validateQuery(AccumuloClient client, Query settings, Set<Authorizations> auths, boolean expandFields, boolean expandValues) throws Exception {
+        this.config = ShardQueryConfiguration.create(this, settings);
+        if (log.isTraceEnabled()) {
+            log.trace("Initializing ShardQueryLogic for query validation: " + System.identityHashCode(this) + '('
+                            + (this.getSettings() == null ? "empty" : this.getSettings().getId()) + ')');
+        }
+
+        // todo - maybe unnecessary, we could just return early if no rules configured.
+        if (validationRules == null || validationRules.isEmpty()) {
+            throw new IllegalStateException("Query validation rules not configured.");
```
ivakegg: I would think in this case that query validation is not supported.  Call super.validateQuery instead to throw the sane UnsupportedOperationException.

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
ivakegg: Nevermind.... I guess this would be needed for the field existence rule to work correctly.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldsWithNumericValuesVisitor.java
```diff
@@ -0,0 +1,91 @@
...
+    public Object visit(ASTGENode node, Object data) {
+        checkSingleField(node, data);
+        return data;
+    }
+
+    private void checkSingleField(JexlNode node, Object data) {
+        String field = JexlASTHelper.getIdentifier(node);
+        if (field != null) {
+            Object literal = JexlASTHelper.getLiteralValue(node);
+            if (literal instanceof Number) {
```
ivakegg, line 83: I think we need to pull string values that contain numbers as well.  I believe that all numbers end up getting converted to strings literals in the process of parsing lucene.  I could be wrong  but I am pretty sure that is the case.

warehouse/query-core/src/main/java/datawave/query/lucene/visitors/AmbiguousUnfieldedTermsVisitor.java
```diff
@@ -0,0 +1,273 @@
...
+import java.util.List;
+import java.util.function.Function;
+
+import org.apache.lucene.queryparser.flexible.core.nodes.AndQueryNode;
+import org.apache.lucene.queryparser.flexible.core.nodes.FieldQueryNode;
+import org.apache.lucene.queryparser.flexible.core.nodes.GroupQueryNode;
+import org.apache.lucene.queryparser.flexible.core.nodes.OrQueryNode;
+import org.apache.lucene.queryparser.flexible.core.nodes.QueryNode;
+
+public class AmbiguousUnfieldedTermsVisitor extends BaseVisitor {
```
ivakegg, line 18: Add a class comment as to what this visitor is looking for.

warehouse/query-core/src/main/java/datawave/query/lucene/visitors/BaseVisitor.java
```diff
@@ -0,0 +1,301 @@
...
+import org.apache.lucene.queryparser.flexible.standard.nodes.PrefixWildcardQueryNode;
+import org.apache.lucene.queryparser.flexible.standard.nodes.RegexpQueryNode;
+import org.apache.lucene.queryparser.flexible.standard.nodes.SynonymQueryNode;
+import org.apache.lucene.queryparser.flexible.standard.nodes.TermRangeQueryNode;
+import org.apache.lucene.queryparser.flexible.standard.nodes.WildcardQueryNode;
+import org.apache.lucene.search.WildcardQuery;
+
+import datawave.query.language.parser.lucene.EscapeQuerySyntaxImpl;
+
+public class BaseVisitor {
```
ivakegg, line 39: Nice addition creating a visitor pattern like the JEXL one.

warehouse/query-core/src/main/java/datawave/query/lucene/visitors/PrintingVisitor.java
```diff
@@ -0,0 +1,325 @@
...
+import org.apache.lucene.queryparser.flexible.standard.nodes.PointRangeQueryNode;
+import org.apache.lucene.queryparser.flexible.standard.nodes.PrefixWildcardQueryNode;
+import org.apache.lucene.queryparser.flexible.standard.nodes.RegexpQueryNode;
+import org.apache.lucene.queryparser.flexible.standard.nodes.SynonymQueryNode;
+import org.apache.lucene.queryparser.flexible.standard.nodes.TermRangeQueryNode;
+import org.apache.lucene.queryparser.flexible.standard.nodes.WildcardQueryNode;
+
+import datawave.query.Constants;
+
+public class PrintingVisitor extends BaseVisitor {
```
ivakegg, line 43: I am not seeing this class being used anywhere, what am I missing.  Also what functionality does this provide that the LuceneQueryStringBuildingVisitor does not already supply?

ivakegg (changes requested):
A few comments to handle, but all in all this looks pretty good!  Created https://github.com/NationalSecurityAgency/datawave/pull/2656 with some of the suggested changes.

web-services/query/src/main/java/datawave/webservice/query/runner/QueryExecutorBean.java
lbschanno: Fixed by your PR, but I added back in the call to the persister (with transient mode so that the query is not actually persisted) to initialize the query var. If you prefer this to be initialized a different way, let me know.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldsWithNumericValuesVisitor.java
lbschanno: Fixed, and updated tests to test it.

warehouse/query-core/src/main/java/datawave/query/lucene/visitors/AmbiguousUnfieldedTermsVisitor.java
lbschanno: Flxed.

warehouse/query-core/src/main/java/datawave/query/lucene/visitors/BaseVisitor.java
lbschanno: Thanks!

warehouse/query-core/src/main/java/datawave/query/lucene/visitors/PrintingVisitor.java
lbschanno: This class is analogous to the PrintingVisitor for JEXL queries. It prints the overall structure of LUCENE query trees so that we can the hierarchy of nodes and their details. We used it extensively when examining trees while writing tests. It is currently now used in at least one place to log the LUCENE query after the initial parse in `ShardQueryLogic.validateQuery().`

The LuceneQueryStringBuildingVisitor is analogous to JexlStringBuildingVisitor. It returns a LUCENE query tree as a query string, but does not show the actual structure of the tree.

ivakegg (dismissed):

ivakegg (dismissed):

ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/QueryFunctionsDescriptor.java
```diff
@@ -33,9 +33,11 @@
 import datawave.query.jexl.functions.arguments.JexlArgumentDescriptor;
 import datawave.query.jexl.nodes.QueryPropertyMarker;
 import datawave.query.jexl.visitors.EventDataQueryExpressionVisitor;
+import datawave.query.jexl.visitors.PrintingVisitor;
```
alerman: is this used? I dont see it in the changes to this file, but I may have just missed it.

lbschanno: It is not being used. Just pushed a commit to remove it.

ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/LuceneToJexlQueryParser.java
```diff
@@ -86,6 +81,20 @@ public JexlNode convertToJexlNode(String query) throws ParseException {
         return parsedQuery;
     }
 
+    @Override
+    public org.apache.lucene.queryparser.flexible.core.nodes.QueryNode parseToLuceneQueryNode(String query) throws QueryNodeParseException {
+        query = query.replaceAll("\\u0093", "\""); // replace open smart quote 147
```
hlgp: Consider adding these to Constants as well?

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -1850,135 +1852,22 @@ private void loadDataTypeMetadata(Multimap<String,Type<?>> fieldToDatatypeMap, S
         }
     }
 
-    protected Set<String> upcase(Set<String> fields) {
```
hlgp, line 1853: While in general I am a big fan of slimming down the DefaultQueryPlanner, it's hard to see how much of this refactoring is necessary for the validate endpoint.   Is this necessary?

warehouse/query-core/src/test/java/datawave/query/rules/IncludeExcludeIndexFieldsRuleTest.java
```diff
@@ -0,0 +1,102 @@
...
+        expectMessage("Indexed fields found within the function filter:" + name + ": INDEXED1");
+
+        assertResult();
+    }
+
+    /**
+     * Test a query with both the includeRegex and excludeRegex functions with indexed fields.
+     */
+    @Test
+    void testMultipleFunctionWithIndexedField() throws Exception {
```
hlgp, line 74: Lots of great tests!

warehouse/query-core/src/test/java/datawave/query/tables/ShardQueryLogicTest.java
```diff
@@ -219,197 +209,209 @@ protected void runTestQuery(Set<Set<String>> expected, String querystr, Date sta
...
     public void testFieldMappingTransformViaProfile() throws Exception {
-        Map<String,String> extraParameters = new HashMap<>();
-        extraParameters.put("include.grouping.context", "false");
-        extraParameters.put("query.profile", "copyFieldEventQuery");
-
-        Date startDate = format.parse("20091231");
-        Date endDate = format.parse("20150101");
-
-        String queryString = "UUID =~ '^[CS].*'";
+        givenQuery("UUID =~ '^[CS].*'");
```
hlgp, line 212: Very pretty refactoring.  I like the change a lot, but in future MRs can we keep this kind of refactoring separate from adding functionality?  Hard to validate the validate while going on side quests.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
lbschanno: The functionality for uppercasing identifiers and applying the query model was extracted to `datawave.query.util.ShardQueryUtils` so that it could be used in `ShardQueryLogic.validateQuery()` when applying some basic normalization on the JEXL query before running it through any JEXL-specific validation. I did this because I didn't want `ShardQueryLogic.validateQuery()` to be too tightly bound DefaultQueryPlanner.

warehouse/query-core/src/test/java/datawave/query/rules/IncludeExcludeIndexFieldsRuleTest.java
lbschanno: Thanks!

warehouse/query-core/src/test/java/datawave/query/tables/ShardQueryLogicTest.java
lbschanno: Fair comment, I'll keep that in mind for future MRs.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
hlgp: Got it!  Thanks!

hlgp (dismissed):
Impressive amount of work and tests!

warehouse/query-core/src/main/java/datawave/query/language/parser/jexl/LuceneToJexlQueryParser.java
lbschanno: No problem, will do.

hlgp (approved):

ivakegg (approved):

PR #2655, opened 2024-11-25
Consolidate Ivarator Config

### LINKED ISSUE #2644: Consolidate Ivarator Configs
It would be good to consolidate several ivarator configs together into a single config object. This would improve code readability and reduce the chance that a simple change in one area isn't propagated. 

As a first step, consolidate the following section of the ShardQueryLogic into a single IvaratorConfig object. For simplicity we can reuse the object mapper code from the IvaratorCacheDirConfig. 

For the sake of backwards compatibility and downstream integration, leave the getter and setter methods for each of these variables, but delegate them to the ivarator config object. 

```
    private List<IvaratorCacheDirConfig> ivaratorCacheDirConfigs = Collections.emptyList();
    private String ivaratorFstHdfsBaseURIs = null;
    private int ivaratorCacheBufferSize = 10000;
    private long ivaratorCacheScanPersistThreshold = 100000L;
    private long ivaratorCacheScanTimeout = 1000L * 60 * 60;
    private int maxFieldIndexRangeSplit = 11;
    private int ivaratorMaxOpenFiles = 100;
    private int ivaratorNumRetries = 2;
    private boolean ivaratorPersistVerify = true;
    private int ivaratorPersistVerifyCount = 100;
    private int maxIvaratorSources = 33;
    private long maxIvaratorSourceWait = 1000L * 60 * 30;
    private long maxIvaratorResults = -1;
```

SethSmucker:
Merges IvaratorConfigs into one central object to be used in ~ShardQueryLogic~ ShardQueryConfiguration. Setters/Getters remain the same.
#2644


warehouse/query-core/src/test/java/datawave/query/config/IvaratorConfigTest.java
```diff
@@ -0,0 +1,91 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+
+import datawave.query.iterator.QueryOptions;
+
+class IvaratorConfigTest {
+
+    @Test
+    public void testSerializeDeserializeEquivalence() throws JsonProcessingException {
+        IvaratorConfig conf1 = new IvaratorConfig();
+        conf1.setIvaratorFstHdfsBaseURIs("hdfs://example.com/ivarator");
```
apmoriarty: The URI won't have a dot com in it, a reasonable value for tests would be something like "file://tmp/ivarator" or "hdfs://tmp/ivarator/"

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -1507,17 +1495,6 @@ public boolean validateOptions(Map<String,String> options) {
             this.termCounts = getMapSerDe().deserializeFromString(serializedMap);
         }
 
-        // parse out cardinality threshold
```
apmoriarty, line 1573: It looks like cardinality threshold stuff was reverted as part of this work, please correct this

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -1114,27 +1093,35 @@ public void setIvaratorPersistOptions(FileSortedSet.PersistOptions ivaratorPersi
...
     }
 
     public void setMaxIvaratorSourceWait(long maxIvaratorSourceWait) {
-        this.maxIvaratorSourceWait = maxIvaratorSourceWait;
+        ivaratorConfig.setMaxIvaratorSourceWait(maxIvaratorSourceWait);
     }
 
     public long getMaxIvaratorResults() {
-        return maxIvaratorResults;
+        return ivaratorConfig.getMaxIvaratorResults();
```
apmoriarty: Let's follow a "get or create" pattern. This allows lazy initialization of the ivarator config. This means a query that does not require an ivarator can avoid some extra object initialization (and later, some potentially expensive I/O calls to the configured filesystem).

This would become 
```java
getIvaratorConfig().setMaxIvaratorResults(maxIvaratorResults)
..
return getIvaratorConfig().getMaxIvaratorResults()
..
public IvaratorConfig getIvaratorConfig(){
  if (ivaratorConfig == null) {
    // create a fresh config with default values
    ivaratorConfig = new IvaratorConfig();
  }
  return ivaratorConfig;
}
```

warehouse/query-core/src/main/java/datawave/query/config/IvaratorConfig.java
```diff
@@ -0,0 +1,217 @@
...
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+
+import datawave.query.iterator.ivarator.IvaratorCacheDirConfig;
+import datawave.query.util.sortedset.FileSortedSet;
+
+public class IvaratorConfig implements Serializable {
```
apmoriarty, line 35: This looks great! Add a class level comment that this class encapsulates all ivarator configs in a single object to simplify serialization of configs between the webservice and tablet servers

apmoriarty (changes requested):

warehouse/query-core/src/test/java/datawave/query/config/ShardQueryConfigurationTest.java
```diff
@@ -394,34 +394,34 @@ public void setUp() throws Exception {
...
+        // defaultValues.put("maxIvaratorSources", 33);
+        // updatedValues.put("maxIvaratorSources", 16);
+        // defaultValues.put("maxIvaratorResults", -1L);
+        // updatedValues.put("maxIvaratorResults", 10000L);
         defaultValues.put("maxIvaratorTerms", -1);
         updatedValues.put("maxIvaratorTerms", 50);
-        defaultValues.put("maxIvaratorSourceWait", 1000L * 60 * 30);
-        updatedValues.put("maxIvaratorSourceWait", 1000L * 60 * 10);
+        // defaultValues.put("maxIvaratorSourceWait", 1000L * 60 * 30);
+        // updatedValues.put("maxIvaratorSourceWait", 1000L * 60 * 10);
```
apmoriarty: remove commented out code please

warehouse/query-core/src/main/java/datawave/query/config/IvaratorConfig.java
apmoriarty: Still need a class level comment explaining the motivation

warehouse/query-core/src/test/java/datawave/query/config/IvaratorConfigTest.java
```diff
@@ -0,0 +1,93 @@
+package datawave.query.config;
+
+import static org.junit.jupiter.api.assertEquals;
+import static org.junit.jupiter.api.assertNotEquals;
```
foster33: These are not valid imports and are causing compilation failures.

```suggestion
import static org.junit.jupiter.Assertions.api.assertEquals;
import static org.junit.jupiter.api.Assertions.assertNotEquals;
```

foster33 (changes requested):

PR #2658, opened 2024-12-03
Log the results of index expansion in a standardized way

apmoriarty:
Standardize the logging for index expansion so that at the debug level the result of the expansion is recorded, and if the expansion fails we know which type (threshold exceeded or no hits)

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/BoundedRangeIndexExpansionVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/BoundedRangeIndexExpansionVisitor.java
@@ -11,9 +11,9 @@ import static datawave.query.jexl.nodes.QueryPropertyMarker.MarkerType.INDEX_HOL
 import org.apache.accumulo.core.client.TableNotFoundException;
 import org.apache.commons.jexl3.parser.ASTAndNode;
 import org.apache.commons.jexl3.parser.JexlNode;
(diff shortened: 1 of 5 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/BoundedRangeIndexExpansionVisitor.java
```diff
@@ -104,7 +104,39 @@ protected void rebuildFutureJexlNode(FutureJexlNode futureJexlNode) {
...
+     */
+    protected void logResult(JexlNode node, IndexLookupMap lookupMap) {
+        String field = JexlASTHelper.getIdentifier(node);
+
+        if (lookupMap == null) {
+            log.debug("Lookup map was null for term [{}]", JexlStringBuildingVisitor.buildQuery(node));
+            return;
+        }
+
+        if (lookupMap.containsKey(field)) {
```
hlgp: Love it.  My only thought is since the logResult methods in the visitors are extremely similar, this could probably be bumped to a base Visitor and supply the strings that are different, i.e., range, regex, or field expansion

apmoriarty: easy enough

ivakegg (approved):

awrigh11 (approved):

PR #2660, opened 2024-12-04
Add FieldExpansionIterator that replaces the ANYFIELD scanner

apmoriarty:
Add iterator that expands fields for literal values.


warehouse/query-core/src/main/java/datawave/query/data/parsers/ShardIndexKey.java
FineAndDandy, line 1: should this be folded into DatawaveKey?

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/FieldExpansionIndexLookup.java
```diff
@@ -0,0 +1,168 @@
...
+                    scanner.fetchColumnFamily(new Text(field));
+                }
+
+                IteratorSetting setting = createIteratorSetting();
+                scanner.addScanIterator(setting);
+
+                timedScanFuture = execService.submit(createTimedCallable(scanner));
+            } catch (Exception e) {
+                log.error("Error expanding term into discrete fields", e);
+                throw new RuntimeException(e);
```
FineAndDandy, line 78: cleanup the scanner

FineAndDandy (changes requested):

warehouse/query-core/src/main/java/datawave/query/data/parsers/ShardIndexKey.java
apmoriarty: No. The datawave key general purpose and not reusable, the ShardIndex key and other key parsers like it (EventKey, FieldIndexKey, TermFrequencyKey) are lightweight and reusable.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/FieldExpansionIndexLookup.java
apmoriarty: Even if an exception is thrown in the catch block, the finally statement will still execute, thus closing the scanner.

FineAndDandy: If you get an exception inside the future that is true, but if you get an exception in the above method after creating the scanner, I don't see that there is a finally that will catch anything. None of the scanner methods have throws on them, but you are catching Exception, which covers a lot more than expected throwables.

apmoriarty: Taking a second pass through this. The `createTimedCallable` method that handles executing the scanner will close it via `scannerFactory.close(scanner)`. I think that's appropriate because otherwise the finally block might close the scanner before the executor service has had a chance to execute the timed callable. 

I thought I had a finally block in a local commit that never made it up but that not correct. 

Just for fun I closed the scanner in a finally block for regex and bounded range and several tests blew up. I'll add comments to each lookup that scanners should never be closed here.

warehouse/query-core/src/main/java/datawave/core/iterators/FieldExpansionIterator.java
```diff
@@ -0,0 +1,152 @@
...
+        endDate = options.get(END_DATE) + Constants.MAX_UNICODE_STRING;
+
+        super.init(source, options, env);
+    }
+
+    @Override
+    public FilterResult filter(Key k, Value v) {
+
+        // keep it simple for now
+        log.trace("tk: {}", k.toStringNoTime());
```
ivakegg: wrap with if (log.isTraceEnabled() as the toStringNoTime could be expensive

warehouse/query-core/src/main/java/datawave/core/iterators/FieldExpansionIterator.java
```diff
@@ -0,0 +1,152 @@
...
+            log.trace("field not in set of expansion fields, or already seen this field. advancing to next field");
+            return new FilterResult(false, AdvanceResult.NEXT_CF);
+        }
+
+        // ensure key falls within the date range
+        String date = parser.getShard();
+        if (date.compareTo(startDate) < 0) {
+            // advance to start date
+            log.trace("Key before start date: {} < {}", date, startDate);
+            return new FilterResult(false, AdvanceResult.NEXT);
```
ivakegg: is there a NEXT_CQ ?

warehouse/query-core/src/main/java/datawave/query/data/parsers/ShardIndexKey.java
```diff
@@ -0,0 +1,98 @@
...
+            if (cq == null) {
+                cq = key.getColumnQualifierData();
+                for (int i = 0; i < cq.length(); i++) {
+                    if (cq.byteAt(i) == 0x00) {
+                        cqSplit = i;
+                        break;
+                    }
+                }
+            }
+            shard = cq.subSequence(0, cqSplit).toString();
```
ivakegg, line 63: is it worth setting up the datatype here as well?

warehouse/query-core/src/main/java/datawave/query/data/parsers/ShardIndexKey.java
```diff
@@ -0,0 +1,98 @@
...
+            if (cq == null) {
+                cq = key.getColumnQualifierData();
+                for (int i = cq.length() - 1; i > 0; i--) {
+                    if (cq.byteAt(i) == 0x00) {
+                        cqSplit = i;
+                        break;
+                    }
+                }
+            }
+            datatype = cq.subSequence(cqSplit + 1, cq.length()).toString();
```
ivakegg, line 47: is it worth setting up the shard as well here?

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/FieldExpansionIndexLookup.java
```diff
@@ -0,0 +1,172 @@
...
+    protected AtomicLong lookupStartTimeMillis = new AtomicLong(Long.MAX_VALUE);
+    protected CountDownLatch lookupStartedLatch;
+    protected CountDownLatch lookupStoppedLatch;
+
+    private Scanner scanner;
+
+    public FieldExpansionIndexLookup(ShardQueryConfiguration config, ScannerFactory scannerFactory, String term, Set<String> fields,
+                    ExecutorService execService) {
+        super(config, scannerFactory, true, execService);
+        this.term = term.toLowerCase();
```
ivakegg: why lower case here?  normalizers do not necessarily lower case values.

warehouse/query-core/src/main/java/datawave/core/iterators/FieldExpansionIterator.java
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        if (date.compareTo(endDate) > 0) {
+            // advance to next field
+            log.trace("Key after end date: {} > {}", date, endDate);
+            return new FilterResult(false, AdvanceResult.NEXT_CF);
+        }
+
+        if (datatypes != null && !datatypes.contains(parser.getDatatype())) {
+            // figure out if we are before the first datatype -- seek to first datatype
```
ivakegg: This comment does not seem relevant here

ivakegg: can we seek to the next datatype?

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/data/parsers/ShardIndexKey.java
apmoriarty: KeyParsers are lazily parsed, omitting the datatype is by design.

warehouse/query-core/src/main/java/datawave/query/data/parsers/ShardIndexKey.java
apmoriarty: KeyParsers are lazily parsed, omitting the shard is by design.

warehouse/query-core/src/main/java/datawave/core/iterators/FieldExpansionIterator.java
apmoriarty: Due to the structure of the shard index keys a seek based on the datatype is never practical. I'll correct the comment.

warehouse/query-core/src/main/java/datawave/core/iterators/FieldExpansionIterator.java
apmoriarty: Instead of NEXT_CQ maybe this should be a custom hint that triggers a seek to the start date..

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/FieldExpansionIndexLookup.java
apmoriarty: Right you are...I need to revisit this case. The previous implementation was doing a just-in-time expansion for normalization which implies the ExpandMultiNormalizedTerms doesn't handle unfielded terms well..

apmoriarty: I went on a journey through the query planner...

warehouse/query-core/src/main/java/datawave/core/iterators/FieldExpansionIterator.java
apmoriarty: Revisited this. If any datatype in the filter sorts AFTER the current datatype then we can advance via a next call. If no datatype in the filter sorts after the current datatype then no match is possible and the iterator can advance to the next field.

alerman (dismissed):

avgAGB (dismissed):

ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/core/iterators/FieldExpansionIterator.java
```diff
@@ -0,0 +1,182 @@
...
+        // keep it simple for now
+        if (log.isTraceEnabled()) {
+            log.trace("tk: {}", k.toStringNoTime());
+        }
+
+        parser.parse(k);
+
+        // if field does not match, skip to next field
+        if ((fields != null && !fields.contains(parser.getField())) || found.contains(parser.getField())) {
+            log.trace("field not in set of expansion fields, or already seen this field. advancing to next field");
```
FineAndDandy, line 89: this also needs log.isTraceEnabled()

FineAndDandy, line 90: consider also adding trace enabled check here and other places in the class

warehouse/query-core/src/main/java/datawave/core/iterators/FieldExpansionIterator.java
```diff
@@ -0,0 +1,182 @@
...
+    @Override
+    public boolean validateOptions(Map<String,String> options) {
+        return options.containsKey(START_DATE) && options.containsKey(END_DATE);
+    }
+
+    @Override
+    public void seek(Range range, Collection<ByteSequence> columnFamilies, boolean inclusive) throws IOException {
+        if (!range.isStartKeyInclusive()) {
+            // need to skip to next column family
+            Key skip = range.getStartKey().followingKey(PartialKey.ROW_COLFAM);
```
FineAndDandy, line 172: if the start key was null (infinite -) this will throw an NPE

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/FieldExpansionIndexLookup.java
```diff
@@ -0,0 +1,172 @@
...
+
+                        entry.getKey().getRow(holder);
+                        String row = holder.toString();
+
+                        entry.getKey().getColumnFamily(holder);
+                        String columnFamily = holder.toString();
+
+                        // We are only returning a mapping of field name to field value, no need to
+                        // determine cardinality and such at this point.
+                        log.trace("put {}:{}", columnFamily, row);
```
FineAndDandy: should this check log level? This will be called a lot potentially?

warehouse/query-core/src/test/java/datawave/query/UnindexedNumericQueryTest.java
```diff
@@ -68,7 +68,7 @@ public void testNumericTerm() throws Exception {
 
         String min = "115";
         String iowa = "'indiana'";
```
FineAndDandy: these are no longer referenced

warehouse/query-core/src/main/java/datawave/core/iterators/FieldExpansionIterator.java
apmoriarty: slf4j will make sure trace is enabled before building complex objects for logging, so this method call should be safe. I will check the other instances in this class to make sure I'm using it correctly

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/FieldExpansionIndexLookup.java
apmoriarty: can't hurt

warehouse/query-core/src/main/java/datawave/core/iterators/FieldExpansionIterator.java
apmoriarty: adding some precondition checks..

ivakegg (approved):

FineAndDandy (approved):

PR #2661, opened 2024-12-05
Add WritesQueryMetrics to EventQueryTransformer 

### LINKED ISSUE #2605: EventQueryTransformer does not implement WritesQueryMetrics
The EventQueryTransformer does not implement the WritesQueryMetrics, interface, which is preventing the seek, next, yield, source, etc, metrics from the tserver from being written to the query metric

SethSmucker:
Update EventQueryTransformerSupport to implement WritesQueryMetrics so EventQueryTransformer has access to the methods.

Follows the structure of DocumentTransformerSupport implementing WritesQueryMetrics instead of implementing it directly into DocumentTransformer.

fixes #2605


billoley:
I think the reason that EventQueryTransformer does not implement WritesQueryMetrics is that the logics using this transformer do not use the QueryIterator, do not track or return these metrics, and do not have a Document to call extractMetrics.

Recommend closing issue.

PR #2662, opened 2024-12-06
Add ability to configure date types for no expansion

### LINKED ISSUE #2636: Update the ability to configure a date type to not expand if end date is current date
Current users can specify a "date.type" which will use the dateIndex to  subsequently create a "SHARDS_AND_DAYS" variable in the query to subset the shards to be actually searched.  There are situations where if the end date of the date range is trhe current date, and a specific date type is requested, then there is no need to do that expansions.  This needs to be configurable for specified date types.

lbschanno:
Add the ability to configure date types that will not result in date filter and SHARD_AND_DAYS hint addition if the query's date type is one that should not be expanded when the query's end date is the current date.

Fixes #2636


hlgp (dismissed):

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
```diff
@@ -515,6 +515,18 @@ public class ShardQueryConfiguration extends GenericQueryConfiguration implement
      */
     private double fieldIndexHoleMinThreshold = 1.0d;
 
+    /**
+     * The set of date types that, if the query's end date is the current date, will NOT result in any date range adjustments or the addition of a
+     * SHARDS_AND_DAYS hint.
+     */
+    private Set<String> noExpansionIfCurrent = Collections.emptySet();
```
ivakegg: Can this be renamed to denote it is a list of date types somehow?  Perhaps noExpansionIfCurrentDateTypes

lbschanno: Done.

ivakegg (approved):

alerman (approved):

PR #2664, opened 2024-12-09
Draft: Updating protobuf version

avgAGB:


web-services/compile_protos.sh
```diff
@@ -1,10 +1,10 @@
 #!/bin/bash 
 
 # test to see if we have protoc installed
-VERSION=`protoc --version 2>/dev/null | grep "2.5.0" |  wc -l`
+VERSION=`protoc --version 2>/dev/null | grep "3.16.3" |  wc -l`
 if [ "$VERSION" -ne 1 ] ; then
    # Nope: bail
-   echo "::protoc is not available or incorrect version. Requires libprotoc 2.5.0"
+   echo "::protoc is not available or incorrect version. Requires libprotoc 3.16.0"
```
foster33: Should this be `3.16.3` instead of `3.16.0`?

Maybe the version number could be a configurable variable that is then used in the check and the echo statement.

avgAGB: good catch, will fix.

hlgp:
OBE

PR #2678, opened 2024-12-31
Add MAX_UNIQUE_COUNT functionality

### LINKED ISSUE #2635: True Unique capability
The request is to be able to modify the unique capability to return documents that are truely unique within the context of the results.  This means that only those unique results that actually only existed once in the non-uniquified result set are returned.  I am thinking that this can be extended to specify a max-count on the unique results such that those unique results that only occured up to the specfied max-count are actually returned.

lbschanno:
Add the ability to specify a max count on results from the unique
operation such that only the unique results that occurred up to the
specified max count are returned.

This max count can be specified either via the function
MAX_UNIQUE_COUNT or via the query parameter max.unique.count.

Fixes #2635 

NOTE: This work is based off existing work in PR #2570 and should only be merged after that PR.


warehouse/query-core/src/main/java/datawave/query/util/sortedmap/BufferedFileBackedSortedMap.java
```diff
@@ -0,0 +1,560 @@
...
+
+    protected List<SortedMapFileHandlerFactory> handlerFactories;
+    protected int bufferPersistThreshold = DEFAULT_BUFFER_PERSIST_THRESHOLD;
+
+    /**
+     * A factory for SortedMapFileHandlers
+     *
+     *
+     *
+     */
```
foster33, line 51: Extra lines
```suggestion
    /**
     * A factory for SortedMapFileHandlers
     */
```

warehouse/query-core/src/main/java/datawave/query/attributes/TypeAttribute.java
```diff
@@ -99,7 +99,7 @@ public boolean equals(Object o) {
 
         if (o instanceof TypeAttribute) {
             TypeAttribute other = (TypeAttribute) o;
-            return this.getType().equals(other.getType()) && (0 == this.compareMetadata(other));
+            return this.getType().equals(other.getType());
```
awrigh11: Do you mind explaining the reasoning behind this change?

lbschanno: My branch was originally based off another feature branch rather than integration, and this change originated from that feature branch. That feature branch was later merged into integration, but I need to drop the earliest two commits in this PR, which should remove this change.

lbschanno: This change has been removed.

PR #2680, opened 2025-01-06
FederatedQueryPlanner.getSubQueryDateRanges can return an illegal range and is skipping some time periods

### LINKED ISSUE #2679: FederatedQueryPlanner.getSubQueryDateRanges can return an illegal range and is skipping some time periods
Because the index holes are all time midnight, when the final index hole is on the same day as the query endDate but the query endDate is not midnight, the FederatedQueryPlanner is returning a Pair<Date,Date> where the beginDate > endDate

The current logic is also uses oneDayBefore and oneDayAfter in places where oneMsBefore and oneMsAfter are required.  This is causing whole days to be missed when generating subRanges to cover a query date range.

The oneDayBefore and oneDayAfter logic is also causing small periods on the edge of the subRanges to violate the purpose of the method -- that all sub ranges contain all dates with no index holes or all dates with index holes.

More robust testing should be added to prevent regression

billoley:
FederatedQueryPlanner.getSubQueryDateRanges can return an illegal range and is skipping some time periods (#2679)

Because the index holes are all time midnight, when the final index hole is on the same day as the query endDate but the query endDate is not midnight, the FederatedQueryPlanner is returning a Pair<Date,Date> where the beginDate > endDate

The current logic is also uses oneDayBefore and oneDayAfter in places where oneMsBefore and oneMsAfter are required. This is causing whole days to be missed when generating subRanges to cover a query date range.

The oneDayBefore and oneDayAfter logic is also causing small periods on the edge of the subRanges to violate the purpose of the method -- that all sub ranges contain all dates with no index holes or all dates with index holes.

In FederatedQueryPlanner, usages of Calendar were moved to local variables rather than a class variable to avoid the possibility of thread un-safe access.

Added more robust testing to verify fixes and prevent regression

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/planner/FederatedQueryPlanner.java
+++ b/warehouse/query-core/src/main/java/datawave/query/planner/FederatedQueryPlanner.java
@@ -484,7 +484,7 @@ public class FederatedQueryPlanner extends QueryPlanner implements Cloneable {
             Iterator<Pair<Date,Date>> it = mergedHoles.iterator();
 
             // If the start of the first index hole occurs after the configured start date, add a range spanning from
-            // the beginDate to one day before the start of the first index hole.
+            // the beginDate to one millisecond before the start of the first index hole.
             Pair<Date,Date> firstHole = it.next();
             // Track the end of the previous range.
             Date endOfPrevRange;
@@ -711,7 +711,7 @@ public class FederatedQueryPlanner extends QueryPlanner implements Cloneable {
     }
 
     /**
-     * Return one day before the given date.
+     * Return one millisecond before the given date.
      */
     private Date oneMsBefore(Date date) {
         Calendar calendar = Calendar.getInstance();
--- a/warehouse/query-core/src/main/java/datawave/query/planner/FederatedQueryPlanner.java
+++ b/warehouse/query-core/src/main/java/datawave/query/planner/FederatedQueryPlanner.java
@@ -1,7 +1,6 @@
 package datawave.query.planner;
 
 import java.io.IOException;
-import java.text.MessageFormat;
 import java.text.SimpleDateFormat;
 import java.util.Calendar;
 import java.util.Collection;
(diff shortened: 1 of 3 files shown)


warehouse/query-core/src/main/java/datawave/query/planner/FederatedQueryPlanner.java
```diff
@@ -482,79 +479,110 @@ private SortedSet<Pair<Date,Date>> getSubQueryDateRanges(ShardQueryConfiguration
...
             // Otherwise, get the valid date ranges. First, Merge any overlaps.
             SortedSet<Pair<Date,Date>> mergedHoles = mergeRanges(relevantHoles);
+            // Adjust the index holes so that they span from 00:00:00.000 on the first day to 23:59:59.999 on the last day
+            mergedHoles = spanEntireDays(mergedHoles);
             Iterator<Pair<Date,Date>> it = mergedHoles.iterator();
 
-            // If the start of the first hole occurs after the configured start date, add a range spanning from the start date to one day before the start
-            // of the first hole.
+            // If the start of the first index hole occurs after the configured start date, add a range spanning from
+            // the beginDate to one day before the start of the first index hole.
```
ivakegg: Should this be "one ms before the start of the first index hole" ?

ivakegg (dismissed):

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/planner/FederatedQueryPlanner.java
billoley: Good catch.  Thanks.

apmoriarty (approved):

ivakegg (approved):

billoley:
Tested fix by generating metadata i column index hole at the end of the query range.  Before fix - illegal range where end > start.  After fix - correctly generate sub-ranges with no exception

PR #2681, opened 2025-01-07
Store active job cache pointer in Zookeeper

jschmidt10:
We want to store the active job cache somewhere accessible to services not running on ingestmaster. The team decided to go with zookeeper.

I updated load-job-cache.sh to call a java utility to set the active job cache as doing it in bash was non-trivial as you have to recursively create ZK paths if they do not exist.

Honestly, we may want to consider moving the whole load-job-cache functionality into java as that bash script is pretty hefty. We'd be able to test it and probably read it easier. But for now, I just set the active job cache in ZK.


warehouse/ingest-scripts/src/main/resources/bin/ingest/findJars.sh
```diff
@@ -22,8 +22,13 @@ findAccumuloJar (){
 findZookeeperJar(){
   result=$(ls -1 $ZOOKEEPER_HOME/zookeeper-*.jar 2>/dev/null | head -1)
   [[ -f $result ]] || result=$(ls -1 $ZOOKEEPER_HOME/lib/zookeeper-*.jar | head -1)
+  echo $result
+}
+findZookeeperJuteJar(){
```
alerman: This new method only differs from the one above by the name of the jat its looking for. Should we instead pass in a parameter and use one function

jalphonso:
I like it, nice work!

jalphonso (dismissed):

warehouse/ingest-scripts/src/main/resources/bin/ingest/findJars.sh
jschmidt10: Will do!

jschmidt10: Done!

awrigh11 (approved):

alerman (approved):

apmoriarty (approved):

hlgp (approved):

PR #2684, opened 2025-01-09
NestedIterator provides native support for a seek method

apmoriarty:
- added native support for seeking to the NestedIterator interface
- deprecated the SeekableIterator interface and ArrayIterator class
- some unit test cleanup


alerman (dismissed):

warehouse/query-core/src/main/java/datawave/query/iterator/SeekableIterator.java
```diff
@@ -9,6 +9,7 @@
 /**
  *
  */
```
avgAGB, line 11: Small note. Would you consider adding a little note in the javadoc to link a reference to the replacing class? like:

"use {@link #new()} instead." - something like that

jalphonso (dismissed):

warehouse/query-core/src/main/java/datawave/query/iterator/SeekableIterator.java
ivakegg: Is this interface still being used somewhere?

ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/iterator/SeekableIterator.java
apmoriarty: I don't think this interface is still used anywhere, but just in case it is I've deprecated it and marked it for removal so that any project that does use it will see build warnings.

ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/iterator/NestedQueryIterator.java
```diff
@@ -81,23 +83,27 @@ public T next() {
...
     }
 
     @Override
     public T move(T minimum) {
         return currentNest.move(minimum);
     }
 
+    @Override
+    public void seek(Range range, Collection<ByteSequence> columnFamilies, boolean inclusive) throws IOException {
+        currentNest.seek(range, columnFamilies, inclusive);
```
FineAndDandy, line 117: might this need to go beyond currentNest if currentNest is exhausted but there are more keys?

*Edit*

After some code searching, it doesn't look like we ever set a nest when using this iterator. Maybe we should agree what this should do, document it, and add a test for it? As-is both move and this method will act on the currentNest only, but hasNext and next will potentially move to the next nest if available.

FineAndDandy (changes requested):
I'd like some clarity/docs/tests on the NestedQueryIterator then I'm good.

ivakegg (approved):

FineAndDandy (approved):

PR #2697, opened 2025-01-16
improve GroupBy performance

apmoriarty:
- Numeric now checks for encoded values to avoid expensive exception-based error handling.
- GroupingIterator and GroupingTransform no longer keep a list of document keys, instead opting to retain the most recent key.
- GroupingIterator and GroupingTransform now log how many documents were seen at debug level when no more documents exist in the source iterator
- Deprecated GroupingUtils method that builds a document from a list of keys and added a method that builds a document from a single key.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/attributes/Numeric.java
+++ b/warehouse/query-core/src/main/java/datawave/query/attributes/Numeric.java
@@ -69,10 +69,23 @@ public class Numeric extends Attribute<Numeric> implements Serializable {
     }
 
     private Number parseToNumber(String value) {
-        Number number = null;
+        // some documents will contain numeric data that is already encoded
+        // in those cases it is faster to check for encoding up front instead
+        // of finding out the hard way via exception.
+        try {
+            if (NumericalEncoder.isPossiblyEncoded(value)) {
+                BigDecimal bigDecimal = NumericalEncoder.decode(value);
+                return bigDecimal.doubleValue();
+            }
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+
+        Number number;
         try {
             number = NumberUtils.createNumber(value);
         } catch (Exception ex) {
+            // alternatively, throw a shallow exception to avoid expensive calls to 'fill in stacktrace'
             BigDecimal bigDecimal = NumericalEncoder.decode(value);
             number = bigDecimal.doubleValue();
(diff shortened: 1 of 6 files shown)


warehouse/query-core/src/main/java/datawave/query/iterator/GroupingIterator.java
```diff
@@ -78,18 +78,20 @@ public boolean hasNext() {
...
                 log.trace("hasNext is false because yield was called");
                 if (!groups.isEmpty()) {
                     // reset the yield and use its key in the flattened document prepared below
-                    keys.add(yieldCallback.getPositionAndReset());
+                    mostRecentKey = yieldCallback.getPositionAndReset();
                 }
                 break;
             } else {
                 // in.hasNext() was false and there was no yield
+                log.debug("GroupingIterator saw {} documents producing {} groups", documentCount, groups.getGroups().size());
```
dtspence: Would it be possible to update this log message to be at a trace level?

jalphonso (approved):

warehouse/query-core/src/main/java/datawave/query/attributes/Numeric.java
```diff
@@ -69,10 +69,23 @@ public Object getData() {
...
+        // some documents will contain numeric data that is already encoded
+        // in those cases it is faster to check for encoding up front instead
+        // of finding out the hard way via exception.
+        try {
+            if (NumericalEncoder.isPossiblyEncoded(value)) {
+                BigDecimal bigDecimal = NumericalEncoder.decode(value);
+                return bigDecimal.doubleValue();
+            }
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
alerman, line 81: Should we add a message to this RE, or is the thrown exception specific enough?

If its a specific exception, should we only catch the specific exception rather than Exception?

alerman (approved):

warehouse/query-core/src/main/java/datawave/query/attributes/Numeric.java
apmoriarty: The thrown exception should always be an IllegalArgumentException which is specific enough for me. The NumericalEncoder.decode method catches Exception to throw the RuntimeException so I think replicating that pattern here is also okay.

PR #2700, opened 2025-01-17
Colvis cache

### LINKED ISSUE #2699: ColumnVisibility objects should be cached in Attribute
ColumnVisibility objects that are created in a tight loop can benefit from caching to improve performance

jalphonso:
Closes #2699  
Updates Attribute class to use Caffeine cache to store ColumnVisibility objects.  
When the cache is used a copy of the original ColumnVisibility is made and returned to the caller.  
This is made possible by the new copy constructor in Accumulo 2.1.4 so this PR is gated by that release.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -38,7 +38,7 @@
         <maven.compiler.target>11</maven.compiler.target>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <surefire.forkCount>1C</surefire.forkCount>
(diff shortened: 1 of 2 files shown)


jalphonso:
Note: build succeeded in my local environment when I built a 2.1.2 accumulo with my changes on top. There are breaking changes unrelated to my code that will need to be addressed before we can use 2.1.4

dtspence (approved):

PR #2701, opened 2025-01-21
Fix Event Field Filter

apmoriarty:
Actually set the evaluation filter so that filtering works for standard event queries


warehouse/query-core/src/main/java/datawave/query/predicate/EventDataQueryFieldFilter.java
```diff
@@ -54,6 +54,7 @@ public EventDataQueryFieldFilter(EventDataQueryFieldFilter other) {
         this.maxNextCount = other.maxNextCount;
         this.fields = other.fields;
         this.parser = other.parser;
+        this.parser.clearState();
```
jalphonso, line 57: Is this the only field we need to worry about being shared across EventDataQueryFieldFilter objects? Should we instead be performing a deepCopy or does it not matter for this use case?

jalphonso (commented):
Good catch, seems pretty straightforward. I only had one comment. Let me know what you think and I'll approve after any changes are made if needed.

warehouse/query-core/src/main/java/datawave/query/predicate/EventDataQueryFieldFilter.java
apmoriarty: I don't think there are any other objects that need to be copied over, this clear state call is defensive -- I couldn't find evidence that it affected anything.

jack-gitdev (approved):

jalphonso (approved):

PR #2710, opened 2025-01-23
fixes unhealthy query metrics container

### LINKED ISSUE #2666: Docker Compose Query Metrics Container is Often 'Unhealthy'
I have been running into an issue with the docker compose. When I have `--web` set in `docker-compose.yml` and then run `docker compose --profile quickstart --profile full up -d` I almost always get the following error:

`dependency failed to start: container docker-metrics-1 is unhealthy`

I am thinking this has something to do with the timings in the health check of query metrics (whether that be start_period, interval, or retries). Regardless, the health check should be able to account for the amount of time it takes to initialize the container no matter if we are using the webservice or not. I have also found the metrics container is sometimes unhealthy with `--accumulo` set, so this might be a general timing issue.

skyler-kv:
increases the start period for the healthcheck to account for the time to initialize the container

Fixes: #2666


cogross (approved):
Yes -- I've been having issues with the smaller start period as well.

mineralntl:
Purely out of curiosity - how did we arrive at 105 for the new start period?

foster33 (approved):
Tested on my end and looks good.

skyler-kv:
> Purely out of curiosity - how did we arrive at 105 for the new start period?

I did 105 because that was the time where the metric container usually timed out/said it was unhealthy and after testing it looked like the average time to finish was 135s on `--web` and 128s on `--accumulo` so I left it as is

mineralntl (approved):

PR #1968, opened 2023-05-18
consolidate buildDocument logic into the IteratorBuildingVisitor

apmoriarty:
- buildDocument logic now located in the IteratorBuildingVisitor
- 'fieldsToAggregate' set no longer needed, so removed from all index iterator builders
- boolean 'forceDocumentBuild' renamed to 'buildDocument'

PR #2712, opened 2025-01-28
Refactor build document logic

apmoriarty:
Refactor document building logic so that it's all in one place and not distributed throughout many different classes.

Redo of https://github.com/NationalSecurityAgency/datawave/pull/1968

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/iterator/builder/IndexRegexIteratorBuilder.java
+++ b/warehouse/query-core/src/main/java/datawave/query/iterator/builder/IndexRegexIteratorBuilder.java
@@ -19,7 +19,7 @@ import datawave.query.iterator.logic.IndexIteratorBridge;
  *
  */
 public class IndexRegexIteratorBuilder extends IvaratorBuilder implements IteratorBuilder {
-    private static Logger log = Logger.getLogger(IndexRegexIteratorBuilder.class);
+    private static final Logger log = Logger.getLogger(IndexRegexIteratorBuilder.class);
 
     protected Boolean negated;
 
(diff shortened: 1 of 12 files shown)


warehouse/query-core/src/test/java/datawave/query/tables/facets/FacetedQueryLogicTest.java
```diff
@@ -164,11 +164,9 @@ public void testQueryDynamicFacets() throws Exception {
         // TODO: this test isn't working properly. I would expect a query for Italy that is configured to facet
         // the CITY field - to return a facet for rome and paris, but also return a field name.
         Set<String> expected = new TreeSet<>();
-
-        // @formatter:off
         expected.add("null; paris -- paris//1");
         expected.add("null; rome -- rome//2");
-        // @formatter:on
+        expected.add("null; italy -- italy//10");
```
apmoriarty: @drewfarris any thoughts on why this would come back? I was a little concerned that what is effectively a no-op refactor caused more data to come back on this test, but then I saw the comment that it wasn't working properly. Is this enough to stop a merge?

apmoriarty: Nevermind, I found the answer and added a comment to the code.

avgAGB (approved):

warehouse/query-core/src/main/java/datawave/query/iterator/builder/IndexRegexIteratorBuilder.java
```diff
@@ -19,7 +19,7 @@
  *
  */
 public class IndexRegexIteratorBuilder extends IvaratorBuilder implements IteratorBuilder {
-    private static Logger log = Logger.getLogger(IndexRegexIteratorBuilder.class);
+    private static final Logger log = Logger.getLogger(IndexRegexIteratorBuilder.class);
```
avgAGB, line 22: there's an MR out to refactor the Logger references to the slf4j LoggerFactory instead, so it may be worth updating this to be in line with what we expect the loggers to be built as going forward

hlgp (approved):

PR #2717, opened 2025-01-29
Redesign of the DatePartitionedQueryPlanner (formerly know as the FederatedQueryPlanner)

### LINKED ISSUE #2702: FieldIndexHoles not being used by DefaultQueryPlanner
FederatedQueryPlanner issues:

1) The FederateQueryPlanner shall be renamed to the DatePartitionedQueryPlanner
2) The original IndexHole needs to be renamed to a IndexValueGap, and FieldIndexHole to IndexFieldGap
3) The FederatedQueryPlanner.isInDateRange is not handling overlaps of the endDate correctly
4) The date type is not being applied prior to determining the index holes which results is the initial partitioning being incorrect
5) The FederateQueryPlanner is expanding unfielded terms but the query tree is not preservered which means that the
    default query planners are doing the same work all over again
6) The DefaultQueryPlanner is using a static set of indexed fields which basically means FieldIndexHoles are being ignored
   6a) The DefaultQueryPlanner should apply FieldIndexHoles in a mannor similar to the ValueIndexHoles
   6b) OR the set of indexed and reverse indexed fields should be pulled based on the date range.

Consider doing the default query planning completely up front, then partition the query plans and push down terms in each partition based on the field index holes

ivakegg:


warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
```diff
@@ -0,0 +1,820 @@
...
+        original.copyFrom(subConfig);
+        // Ensure the original begin date, end date, and timers are restored after copying over all changes from the sub-config.
+        original.setBeginDate(originalBeginDate);
+        original.setEndDate(originalEndDate);
+        original.setTimers(originalTimers);
+    }
+
+    /**
+     * Return the set of date ranges that sub-queries should be created for. Each date range will have a consistent index state, meaning that within each date
+     * range there will be zero or more fields that are completely indexed in that range. Also it is expected that the date ranges will complete cover the query
```
hlgp: nit: completely cover

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
```diff
@@ -0,0 +1,820 @@
...
+        }
+
+        ensureConsistency(reducedTimeline, config.getBeginDate(), config.getEndDate());
+
+        return reducedTimeline;
+    }
+
+    /**
+     * This method is intended to ensure some fault tolerance in out production of the timeline. The date range from beginDate to endDate should be completely
+     * covered, there should be no gaps, no overlapping date ranges, no negative length date ranges, and every date range should have a different set of
```
hlgp, line 519: yes!  love the validation

warehouse/query-core/src/test/java/datawave/query/IndexHoleQueryTest.java
```diff
@@ -15,7 +15,7 @@
 import org.junit.ClassRule;
 import org.junit.Test;
 
-import datawave.query.config.IndexHole;
+import datawave.query.config.IndexValueHole;
 import datawave.query.exceptions.FullTableScansDisallowedException;
```
hlgp, line 19: Rename test class to IndexValueHoleQueryTest?

hlgp:
We should probably adjust the marker concept to be one day earlier than the last old index entry timestamp.  Since we usually deploy during the day, any deployment where we ceased indexing something would only be a partial day.  I replicated this with a test case and any data from the last indexed date that did not get indexed (i.e. post hypothetical deployment) would be missed.

hlgp:
In other tests, I'm getting empty ranges.  We're still checking the fi for the field with the _Hole_ marker, so the AndIterator appears to be defeating.  Haven't fully stepped through, though, but we need more end-to-end tests up in here (up in here).

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
```diff
@@ -0,0 +1,171 @@
...
+import org.apache.log4j.Logger;
+
+import datawave.query.config.ShardQueryConfiguration;
+import datawave.query.jexl.JexlASTHelper;
+import datawave.query.jexl.LiteralRange;
+import datawave.query.jexl.nodes.QueryPropertyMarker;
+import datawave.query.util.MetadataHelper;
+
+/**
+ * Visitor meant to 'push down' predicates for expressions that are not executable against the index.
```
apmoriarty: index!? global index or field index.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
```diff
@@ -0,0 +1,171 @@
...
+ */
+public class PushdownUnindexedFieldsVisitor extends RebuildingVisitor {
```
apmoriarty, line 34: RebuildingVisitor isn't necessary. If you have 100 terms in a query and only one unindexed term, you're copying 100 terms to swap in a single new term. Extend base visitor and modify the query tree in place, with confidence!

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    public boolean isIndexed(JexlNode node) {
+        String field = JexlASTHelper.getIdentifier(node);
+        try {
+            return (field != null && this.helper.isIndexed(field, this.dataTypeFilter));
+        } catch (TableNotFoundException e) {
+            throw new IllegalStateException("Unable to find metadata table", e);
+        }
+    }
```
apmoriarty: I question whether the metadata helper is actually necessary for this. You already have a set of non-indexed fields...by definition isn't every field not in that set indexed? If you do need to determine the indexed state of a field, why not remove the metadata helper and simply add the set of indexed fields? Then you can remove the set of non-indexed fields.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+    }
+
+    @Override
+    public Object visit(ASTReferenceExpression node, Object data) {
+        // if not already delayed somehow
+        if (!QueryPropertyMarker.findInstance(node).isAnyTypeExcept(BOUNDED_RANGE)) {
+            return super.visit(node, data);
+        }
```
apmoriarty: Marker nodes are handled at the level of an ASTAndNode, you are not guaranteed to have a reference expression. You already handle this case above, so this method isn't necessary

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
```diff
@@ -0,0 +1,171 @@
...
+    public Object visit(ASTEQNode node, Object data) {
+        if (isIndexed(node) && missingIndexRange(node)) {
+            return QueryPropertyMarker.create(node, INDEX_HOLE);
+        }
+        return node;
+    }
+
+    @Override
+    public Object visit(ASTERNode node, Object data) {
+        if (isIndexed(node) && missingIndexRange(node)) {
```
apmoriarty: extract the field once and pass it to the helper methods. helper methods should operate on the field.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
```diff
@@ -0,0 +1,171 @@
...
+        // if not already delayed somehow
+        if (!QueryPropertyMarker.findInstance(node).isAnyTypeExcept(BOUNDED_RANGE)) {
+            return super.visit(node, data);
+        }
+        return node;
+    }
+
+    @Override
+    public Object visit(ASTEQNode node, Object data) {
+        if (isIndexed(node) && missingIndexRange(node)) {
```
apmoriarty: I'm not sure that this logic will handle queries with arithmetic or methods very well. The field or the value may be null.
See the FunctionalSetTest for examples

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
```diff
@@ -0,0 +1,171 @@
...
+import datawave.query.util.MetadataHelper;
+
+/**
+ * Visitor meant to 'push down' predicates for expressions that are not executable against the index.
+ */
+public class PushdownUnindexedFieldsVisitor extends RebuildingVisitor {
+
+    private static final Logger log = Logger.getLogger(PushdownUnindexedFieldsVisitor.class);
+
+    protected ShardQueryConfiguration config;
```
apmoriarty: config isn't necessary

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
ivakegg: reworded

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
ivakegg: thank you

warehouse/query-core/src/test/java/datawave/query/IndexHoleQueryTest.java
ivakegg: done

ivakegg:
> In other tests, I'm getting empty ranges. We're still checking the fi for the field with the _Hole_ marker, so the AndIterator appears to be defeating. Haven't fully stepped through, though, but we need more end-to-end tests up in here (up in here).

I changed this to use evaluation only instead of index value hole markers in the query to avoid the field index lookups for unindexed fields.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
ivakegg: updated comment.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
ivakegg: I have to use the rebuilding visitor.  Each set of unindexed fields is being applied to the original query plan tree so I cannot modify the tree inline.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
ivakegg: I will see what I can do here.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
ivakegg: you are probably correct.  I will validate.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
ivakegg: cool

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
ivakegg: but those methods will not be used in the index lookup so do they matter?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
ivakegg: ok, fair point.  thank you!

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
ivakegg: Actually there is no need to check for indexed fields as the unindexed field list is enough for this visitor..... fixing.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
ivakegg: Sorry, I think, I see what you mean.... I will reevaluate what I am missing.

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
```diff
@@ -0,0 +1,801 @@
...
+
+            return configCopy;
+        } finally {
+            stopwatch.stop();
+        }
+    }
+
+    /**
+     * Return the set of date ranges that sub-queries should be created for. Each date range will have a consistent index state, meaning that within each date
+     * range all query fields are either indexed or not-indexed across the entire range. It is expected that the date ranges will complete cover the original
```
lbschanno: I think you intended `completely cover`, not `complete cover`.

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
```diff
@@ -0,0 +1,801 @@
...
+
+    private static final Logger log = ThreadConfigurableLogger.getLogger(DatePartitionedQueryPlanner.class);
+
+    private final SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy/MM/dd");
+
+    // we want a unique set of plans, but maintain insertion order (facilitates easier testing)
+    private final Set<String> plans = new LinkedHashSet<>();
+    private DefaultQueryPlanner queryPlanner;
+    private String plannedScript;
+    private SortedSet<Pair<Date,Date>> relevantHoles;
```
lbschanno: As far as I can tell, this isn't being used anywhere.

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
```diff
@@ -0,0 +1,801 @@
...
+
+    // handles boilerplate operations that surround a visitor's execution (e.g., timers, logging, validating)
+    private TimedVisitorManager visitorManager = new TimedVisitorManager();
```
lbschanno: This can be final.

warehouse/query-core/src/main/java/datawave/query/planner/FederatedQueryIterable.java
```diff
@@ -9,7 +9,7 @@
 import datawave.query.CloseableIterable;
 
 /**
- * Implementation of {@link CloseableIterable} intended to be used by {@link FederatedQueryPlanner}. This iterable
+ * Implementation of {@link CloseableIterable} intended to be used by {@link DatePartitionedQueryPlanner}. This iterable
  */
 public class FederatedQueryIterable implements CloseableIterable<QueryData> {
```
lbschanno: Should this be renamed to `DatePartitionedQueryIterable`?

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
```diff
@@ -0,0 +1,801 @@
...
+     * Update the planned script to represent a concatenation of the planned scripts from all sub-plans of the most recently executed call to
+     * {@link #process(GenericQueryConfiguration, String, Query, ScannerFactory)}.
+     */
+    private void updatePlannedScript() {
+        if (plans.isEmpty()) {
+            this.plannedScript = "";
+        } else if (this.plans.size() == 1) {
+            this.plannedScript = this.plans.iterator().next();
+        } else {
+            int lastIndex = plans.size() - 1;
```
lbschanno: `lastIndex` is never used and can be deleted.

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
```diff
@@ -0,0 +1,801 @@
...
+                                .isEquals();
+            }
+            return false;
+        }
+
+        @Override
+        public int compareTo(IndexFieldHoleBoundary other) {
+            int comparison = date.compareTo(other.date);
+            if (comparison == 0) {
+                comparison = Boolean.valueOf(other.indexFieldHoleStart).compareTo(Boolean.valueOf(indexFieldHoleStart));
```
lbschanno: This can be simplified to `Boolean.compare(boolean, boolean)`.

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
```diff
@@ -0,0 +1,801 @@
...
+     * unindexed fields.
+     *
+     * @param timeline
+     *            The timeline to verify
+     * @param beginDate
+     *            The begin date
+     * @param endDate
+     *            The end date
+     */
+    private void ensureConsistency(SortedMap<Pair<Date,Date>,Set<String>> timeline, Date beginDate, Date endDate) throws DatawaveFatalQueryException {
```
lbschanno, line 529: I really like this.

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
```diff
@@ -0,0 +1,801 @@
...
+    private SortedSet<IndexFieldHoleBoundary> createTimeline(Map<String,IndexFieldHole> fieldIndexHoles, Date beginDate, Date endDate) {
+        // We want to create a timeline of index hole begin and end dates
+        // that overlap the query's target date range
+        // and map to the fields for which holes are beginning and ending
+        SortedSet<IndexFieldHoleBoundary> timeline = new TreeSet<>();
+        for (Map.Entry<String,IndexFieldHole> hole : fieldIndexHoles.entrySet()) {
+            String field = hole.getKey();
+            IndexFieldHole indexHole = hole.getValue();
+            for (Pair<Date,Date> range : getHolesOverlappingOriginalQueryDateRange(beginDate, endDate, indexHole)) {
+                timeline.add(new IndexFieldHoleBoundary(range.getLeft(), true, field));
```
lbschanno, line 667: `IndexFieldHoleBoundary` can hold a set of fields, however only the single field constructor `IndexFieldHoleBoundary(Date date, boolean start, String field)` is used here, and `IndexFieldHoleBoundary.addFields()` is never called. 

Did you intend for `createTimeline()` to return a set of boundaries that contain singular fields only, or are the fields supposed to be collapsed into any boundaries where they overlap?

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
```diff
@@ -0,0 +1,801 @@
...
+                timeline.add(new IndexFieldHoleBoundary(range.getRight(), false, field));
+            }
+        }
+        if (timeline.isEmpty()) {
+            timeline.add(new IndexFieldHoleBoundary(beginDate, true));
+            timeline.add(new IndexFieldHoleBoundary(endDate, false));
+        } else {
+            if (timeline.first().getBoundary().after(beginDate)) {
+                // start with and beginning boundary that starts at the beginDate with no unindex field
```
lbschanno: Minor nitpick: `start with a beginning`, `with no unindexed field`

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
```diff
@@ -0,0 +1,801 @@
...
+    }
+
+    /**
+     * Return one millisecond before the given date.
+     */
+    private Date oneMsBefore(Date date) {
+        return new Date(date.getTime() - 1);
+    }
+
+    public static class IndexFieldHoleBoundary implements Comparable<IndexFieldHoleBoundary> {
```
lbschanno, line 751: Can you add a comment here describing that this represents a bounding endpoint for the dates that a field index hole encompasses? It becomes apparent after some inspection, but a description never hurts.

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
```diff
@@ -0,0 +1,801 @@
...
+        private final Date date;
+        private final boolean indexFieldHoleStart;
```
lbschanno: I don't think this needs to include `indexFieldHole` in the name, `isStart` or `start` would be fine, or perhaps `boundaryStart`.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
```diff
@@ -0,0 +1,131 @@
...
+     *
+     * @param data
+     *            the node data
+     * @param range
+     *            the range
+     * @param currentNode
+     *            the current node
+     * @return a jexl node
+     */
+    protected JexlNode delayBoundedIndexHole(LiteralRange range, ASTAndNode currentNode, Object data) {
```
lbschanno: The parameter `data` is not used and can be removed.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
```diff
@@ -0,0 +1,131 @@
...
+import datawave.query.jexl.JexlASTHelper;
+import datawave.query.jexl.LiteralRange;
+import datawave.query.jexl.nodes.QueryPropertyMarker;
+
+/**
+ * Visitor meant to 'push down' predicates for expressions that are not executable against the global OR field index.
+ */
+public class PushdownUnindexedFieldsVisitor extends RebuildingVisitor {
+
+    private static final Logger log = Logger.getLogger(PushdownUnindexedFieldsVisitor.class);
```
lbschanno, line 36: `log` is never used and can be deleted.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
```diff
@@ -0,0 +1,131 @@
...
+    public static <T extends JexlNode> T pushdownPredicates(T queryTree, Set<String> unindexedFields) {
+        PushdownUnindexedFieldsVisitor visitor = new PushdownUnindexedFieldsVisitor(unindexedFields);
+        return (T) (queryTree.jjtAccept(visitor, null));
+    }
+
+    @Override
+    public Object visit(ASTAndNode node, Object data) {
+        // if not already delayed somehow
+        if (!QueryPropertyMarker.findInstance(node).isAnyTypeExcept(BOUNDED_RANGE)) {
+            LiteralRange range = JexlASTHelper.findRange().getRange(node);
```
lbschanno: This can be simplified to `if(QueryPropertyMarker.findInstance(node).isType(BOUNDED_RANGE)) {`

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
```diff
@@ -0,0 +1,131 @@
...
+import org.apache.log4j.Logger;
+
+import datawave.query.jexl.JexlASTHelper;
+import datawave.query.jexl.LiteralRange;
+import datawave.query.jexl.nodes.QueryPropertyMarker;
+
+/**
+ * Visitor meant to 'push down' predicates for expressions that are not executable against the global OR field index.
+ */
+public class PushdownUnindexedFieldsVisitor extends RebuildingVisitor {
```
lbschanno, line 34: Will you be adding any tests for this visitor?

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
```diff
@@ -0,0 +1,801 @@
...
+
+                    // Get the range stream for the new date range and query
+                    results.addIterable(subPlan.reprocess(configCopy, configCopy.getQuery(), scannerFactory));
+
+                    if (log.isDebugEnabled()) {
+                        log.debug("Query string for config of sub-plan against date range (" + subBeginDate + "-" + subEndDate + ") with unindexed fields "
+                                        + dateRange.getValue() + ": " + configCopy.getQueryString());
+                    }
+                } catch (DatawaveQueryException e) {
+                    log.warn("Exception occured when processing sub-plan against date range (" + subBeginDate + "-" + subEndDate + ")", e);
```
lbschanno: Spelling: `occured` should be `occurred`.

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
ivakegg: You are right, I was thinking of collapsing the boundaries but I ended up not needing/wanting to do that.....  removing obsolete functionality.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
ivakegg: TBD

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
ivakegg: yeah, totally rewritting this one.  I had copied it from the other pushdown visitor we had and did not think hard enough about it.  Also writting some tests.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
ivakegg: Yes, will do

avgAGB (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
```diff
@@ -0,0 +1,226 @@
...
+
+    @Override
+    public Object visit(ASTFunctionNode node, Object data) {
+        // no need to delay functions
+        return copy(node);
+    }
+
+    @Override
+    public Object visit(ASTMethodNode node, Object data) {
+        // no need to delay functions
```
apmoriarty: nit pick: methods

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/PushdownUnindexedFieldsVisitor.java
```diff
@@ -0,0 +1,226 @@
...
+            }
+        }
+        return false;
+    }
+
+    private boolean missingIndexRange(LiteralRange range) {
+        return (unindexedFields.contains(range.getFieldName()));
+    }
+
+}
```
apmoriarty, line 226: This looks much cleaner

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/JexlStringBuildingVisitor.java
```diff
@@ -188,156 +191,54 @@ public Object visit(ASTAndNode node, Object data) {
...
-            QueryException qe = new QueryException(DatawaveErrorCode.NODE_PROCESSING_ERROR, "An ASTNENode has more than two children");
-            throw new IllegalArgumentException(qe);
-        }
-
-        node.jjtGetChild(0).jjtAccept(this, sb);
-
-        sb.append(" != ");
-
-        node.jjtGetChild(1).jjtAccept(this, sb);
+    public Object visit(ASTEQNode node, Object data) {
```
apmoriarty, line 212: This class has some great cleanup

warehouse/query-core/src/test/java/datawave/query/CompositeFunctionsTest.java
```diff
@@ -670,6 +671,28 @@ public void testTLDWithLuceneAndIdentifierToLiteralEQJexl() throws Exception {
         }
     }
 
+
+    @Test
+    public void testWithHoles(){
+        eventQueryLogic.setQueryPlanner(new DatePartitionedQueryPlanner());
+        Map<String,String> extraParameters = new HashMap<>();
+        extraParameters.put("include.grouping.context", "true");
+        //TODO: why is HOLE not being pushed down with _Hole_ marker?!
```
apmoriarty: Is this still an issue?

warehouse/query-core/src/test/java/datawave/query/planner/DatePartitionedQueryPlannerTest.java
```diff
@@ -638,7 +635,6 @@ public void testFieldIndexMinThresholdWithSomeNotMeeting() throws Exception {
         expectEvents("20130104", FieldIndexHoleDataIngest.corleoneUID, FieldIndexHoleDataIngest.caponeUID, FieldIndexHoleDataIngest.sopranoUID);
         expectEvents("20130105", FieldIndexHoleDataIngest.corleoneUID, FieldIndexHoleDataIngest.caponeUID, FieldIndexHoleDataIngest.sopranoUID);
 
-        assertQueryResults();
-        assertSubrangesCorrect();
+        assertSubrangesCorrect(assertQueryResults());
```
apmoriarty: It took a bit of reading to understand how this test was handling asserting the correct. I would like to see the final query plans asserted, just so it's easy to visually inspect the final results. Something like
```
expectPlan("20130101", "UUID == 'uuid'")
expectPlan("20130101", "(_Hole_ = true) && (UUID == 'uuid')")
```
Doesn't have to be done before a merge, but should be done relatively soon.

apmoriarty (dismissed):

warehouse/query-core/src/test/java/datawave/query/CompositeFunctionsTest.java
hlgp: nope.  corresponding MR: https://github.com/NationalSecurityAgency/datawave/pull/2759

I know you like MRs, so I put some MRs in your MRs.

warehouse/query-core/src/test/java/datawave/query/planner/DatePartitionedQueryPlannerTest.java
ivakegg: done

warehouse/query-core/src/test/java/datawave/query/util/FieldIndexHoleDataIngest.java
hlgp, line 1: IndexFieldHoleDataIngest

ivakegg: done

avgAGB (approved):

PR #2709, opened 2025-01-22
Updated the GlobalIndexUidAggregator to truncate timestamps on keys

ivakegg:
down to the beginning of the day before grouping values with the same key.

PR #2718, opened 2025-01-30
Demonstrate count converstion from Aggregator to RangeStream

FineAndDandy:
only needs to be applied to 7.10.x. https://github.com/NationalSecurityAgency/datawave/pull/2709 should cover all newer versions

### DIFF
--- a/warehouse/query-core/src/test/java/datawave/query/util/WiseGuysIngest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/util/WiseGuysIngest.java
@@ -1055,7 +1055,7 @@ public class WiseGuysIngest {
      */
     private static Value getValueForNuthinAndYourHitsForFree() {
         Uid.List.Builder builder = Uid.List.newBuilder();
-        builder.setCOUNT(50); // better not be zero!!!!
+        builder.setCOUNT(0); // better not be zero!!!!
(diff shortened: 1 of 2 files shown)


apmoriarty (approved):

avgAGB (approved):

warehouse/query-core/src/test/java/datawave/query/util/WiseGuysIngest.java
```diff
@@ -1055,7 +1055,7 @@ private static Value getValueForBuilderFor(String... in) {
      */
     private static Value getValueForNuthinAndYourHitsForFree() {
         Uid.List.Builder builder = Uid.List.newBuilder();
-        builder.setCOUNT(50); // better not be zero!!!!
+        builder.setCOUNT(0); // better not be zero!!!!
```
ivakegg, line 1058: I was tempted to force you to fix this comment, but I already did that in the future versions ;)

ivakegg (approved):

warehouse/query-core/src/test/java/datawave/query/util/WiseGuysIngest.java
FineAndDandy: I was trying to keep the same commit hash!

PR #2726, opened 2025-02-04
SatisfactionVisitor now considers bounded ranges to be satisfiable

apmoriarty:


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/SatisfactionVisitor.java
```diff
@@ -114,16 +114,32 @@ public Object visit(ASTEQNode node, Object data) {
...
+                case EVALUATION_ONLY:
+                    isQueryFullySatisfied = false;
+                    return null;
+                case EXCEEDED_TERM:
+                    log.warn("cannot support exceeded term marker");
+                    isQueryFullySatisfied = false;
+                    return null;
+                case STRICT:
+                case LENIENT:
+                case INDEX_HOLE:
```
ivakegg: index hole means cannot use global index, but it can use the field index.  I thing that one should set satisfied to false.

ivakegg: also please verify the strict and lenient.  Those had to do with model expansion, not index lookup capabilities.

ivakegg (changes requested):
see comments

ivakegg (approved):

avgAGB (approved):

PR #2729, opened 2025-02-05
Attempt to stop duplicate merge pipelines

alerman:


.github/workflows/tests.yml
```diff
@@ -17,10 +17,53 @@ env:
...
+            echo "SHOULD_RUN_FOR_MERGE_QUEUE=true" >> $GITHUB_OUTPUT
+          else
+            echo "SHOULD_RUN_FOR_MERGE_QUEUE=false" >> $GITHUB_OUTPUT
+          fi
   # Runs the pom sorter and code formatter to ensure that the code
   # is formatted and poms are sorted according to project rules. If changes are found,
   # they are committed back to the branch
   check-code-formatting:
+    needs: check-should-run
+    if: github.event_name != 'merge_group' || needs.check-should-run.outputs.SHOULD_RUN_FOR_MERGE_QUEUE == 'false'
```
foster33:
```suggestion
    if: github.event_name == 'merge_group' && needs.check-should-run.outputs.SHOULD_RUN_FOR_MERGE_QUEUE == 'true'
```
At the moment this logic seems to be doing the opposite of what we would want. It is skipping the tests when they shouldn't be. With the suggested change, I think it would run on pull requests and pushes to PRs & branches, just not when the condition `[ "$MERGE_BASE" != "$integration_HEAD" ] && [ "$QUEUE_SIZE" -le 1 ];` is met and the event is `merge_group`. If this assessment is correct, this will have to change for all of the jobs using the `needs: check-should-run`

alerman: Good catch. Originally i had done BRANCH_IS_UP_TO_DATE, but I changed it when I added checking for merge queue length. Applying your suggestion

ivakegg (dismissed):

foster33 (dismissed):

hlgp (approved):

foster33 (approved):

foster33 (commented):
After working on some things relating to skipping the "check-code-formatting" job when the event is merge_group, it seems as though these changes will not work. 

I suspect `if: needs.check-should-run.outputs.SHOULD_RUN == 'true'` will cause some issues because all jobs are required to run and cannot be skipped. As of right now the merge queue will kick the merge "due to no response for status checks".

foster33 (changes requested):
See comment above

alerman:
@foster33 These did work, but created a loophole that allowed things to merge sans pipeline if you clicked at the wrong time. Closing and may revisit other options

PR #2730, opened 2025-02-05
Add batch support and streaming deduping to SSDeepChainedDiscoveryQueryLogic

FineAndDandy:
* Shift state to config where possible
* Add configurable hash deduping at both similarity and chaining layers
* Lazily stream off the similarity iterator when batching is enabled
* Add batch options to chain strategy

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/config/SSDeepSimilarityQueryConfiguration.java
+++ b/warehouse/query-core/src/main/java/datawave/query/config/SSDeepSimilarityQueryConfiguration.java
@@ -37,6 +37,8 @@ public class SSDeepSimilarityQueryConfiguration extends GenericQueryConfiguratio
 
     private Multimap<NGramTuple,SSDeepHash> queryMap;
 
+    private boolean dedupe = true;
+
     public SSDeepSimilarityQueryConfiguration() {
         super();
         setQuery(new QueryImpl());
@@ -68,6 +70,7 @@ public class SSDeepSimilarityQueryConfiguration extends GenericQueryConfiguratio
         setQueryMap(other.getQueryMap());
         setQueryThreads(other.getQueryThreads());
         setRanges(other.getRanges());
+        setDedupe(other.isDedupe());
     }
 
     public Collection<Range> getRanges() {
@@ -140,5 +143,13 @@ public class SSDeepSimilarityQueryConfiguration extends GenericQueryConfiguratio
 
     public void setBucketEncodingLength(int bucketEncodingLength) {
         this.bucketEncodingLength = bucketEncodingLength;
+    }
+
+    public boolean isDedupe() {
+        return this.dedupe;
+    }
+
+    public void setDedupe(boolean dedupe) {
+        this.dedupe = dedupe;
     }
 }
--- a/warehouse/query-core/src/main/java/datawave/query/config/SSDeepSimilarityQueryConfiguration.java
+++ b/warehouse/query-core/src/main/java/datawave/query/config/SSDeepSimilarityQueryConfiguration.java
@@ -1,15 +1,5 @@
 package datawave.query.config;
 
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Map;
-import java.util.Set;
-
-import org.apache.accumulo.core.data.Range;
-
-import com.google.common.collect.Multimap;
-
 import datawave.core.query.configuration.GenericQueryConfiguration;
 import datawave.core.query.logic.BaseQueryLogic;
 import datawave.microservice.query.QueryImpl;
@@ -18,7 +8,6 @@ import datawave.util.ssdeep.BucketAccumuloKeyGenerator;
 import datawave.util.ssdeep.ChunkSizeEncoding;
 import datawave.util.ssdeep.IntegerEncoding;
 import datawave.util.ssdeep.NGramGenerator;
-import datawave.util.ssdeep.NGramTuple;
 import datawave.util.ssdeep.SSDeepHash;
 
 public class SSDeepSimilarityQueryConfiguration extends GenericQueryConfiguration {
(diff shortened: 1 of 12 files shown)


warehouse/query-core/src/main/java/datawave/query/config/SSDeepSimilarityQueryConfiguration.java
```diff
@@ -33,25 +36,45 @@ public class SSDeepSimilarityQueryConfiguration extends GenericQueryConfiguratio
...
-    private Multimap<NGramTuple,SSDeepHash> queryMap;
-
     /**
      * The number of hashes the query logic will accept, -1 indicates unlimited
      */
     private int maxHashes = -1;
 
     /**
-     * The max number of hashes to be generated per ngram, -1 indicates unlimited
+     * Dedupe matching hashes matched inside the SSDeepScoringFunction. When true only process a matching hash one regardless of how many times an ngram may
```
drewfarris:
```suggestion
     * Dedupe matching hashes matched inside the SSDeepScoringFunction. When true only process a matching hash once regardless of how many times an ngram may
```

warehouse/query-core/src/main/java/datawave/query/config/SSDeepSimilarityQueryConfiguration.java
```diff
@@ -33,25 +36,45 @@ public class SSDeepSimilarityQueryConfiguration extends GenericQueryConfiguratio
...
+     * match it
+     */
+    private boolean dedupeSimilarityHashes = true;
+
+    /**
+     * The max number of hashes to be retrieved per ngram, -1 indicates unlimited
      */
     private int maxHashesPerNGram = -1;
 
+    // start of query state objections
```
drewfarris:
```suggestion
    // start of query state objects
```

drewfarris: Is there any value in managing these four instance variables that maintain state (e.g., ranges, queryMap, seenHashes and ngramCountMap) in a single class named `SSDeepSimilarityQueryState` independently from the `SSDeepSimilarityQueryConfiguration`? Would that allow us to avoid cloning the Configuration?

warehouse/query-core/src/main/java/datawave/query/tables/ssdeep/SSDeepSimilarityQueryLogic.java
```diff
@@ -86,9 +87,11 @@ public void setupQuery(GenericQueryConfiguration genericConfig) throws Exception
             scanner.setRanges(config.getRanges());
 
             // must be called after setRanges so that we get the query map from the config.
-            final SSDeepScoringFunction scoringFunction = new SSDeepScoringFunction(config, getMaxResults(), getMaxHashesPerNGram());
+            final SSDeepScoringFunction scoringFunction = new SSDeepScoringFunction(config);
 
-            this.iterator = scanner.stream().flatMap(scoringFunction).distinct().iterator();
+            Stream<ScoredSSDeepPair> stream = scanner.stream().flatMap(scoringFunction);
```
drewfarris: Consider whether we want to implement a separate `Function`s for deduplication and hash count limits that get inserted into the stream prior to the scoring function. This way we maintain a separation of concerns between the functions of deduping, limiting and scoring (and the scoring function remains effectively stateless)

warehouse/query-core/src/main/java/datawave/query/tables/ssdeep/SSDeepScoringFunction.java
```diff
@@ -143,38 +128,42 @@ public Stream<ScoredSSDeepPair> apply(Map.Entry<Key,Value> entry) {
...
+            if (config.getSeenHashes().contains(hashcode)) {
+                return Stream.empty();
+            }
+            config.getSeenHashes().add(hashcode);
+        }
+
+        if (config.getMaxHashesPerNGram() != -1) {
+            Long count = config.getNgramCountMap().get(ngram);
             if (count == null) {
+                // first seen
```
drewfarris: See comment in `SSDeepSimilarityQueryLogic` re: considering implementing separate `Function`s for deduping and limiting.

drewfarris (commented):
Initial pass with some comments - I need to spend more time looking at the streaming chained query and test code.

jack-gitdev:
I read through everything except the unit tests so far. Will read through again tomorrow and run/go through the unit tests as well.

warehouse/query-core/src/main/java/datawave/query/config/SSDeepSimilarityQueryConfiguration.java
FineAndDandy: This is tricky. The Configuration object always contains a mix of state and actual logic config depending on the QueryLogic it is configuring. Since the Config is the only thing that is passed between the init/setupQuery methods and the underlying logic is reused by the framework, the Config necessarily has to carry state info. I like the idea of creating an explicit area of the config that is clearly separated from the logic config to carry this state though.

warehouse/query-core/src/main/java/datawave/query/tables/ssdeep/SSDeepSimilarityQueryLogic.java
FineAndDandy: This is a good idea. I will add this in the next commit

warehouse/query-core/src/main/java/datawave/query/config/SSDeepSimilarityQueryConfiguration.java
```diff
@@ -1,15 +1,19 @@
 package datawave.query.config;
 
 import java.util.Collection;
```
drewfarris: It looks like there may be some unused imports here.

warehouse/query-core/src/main/java/datawave/query/tables/ssdeep/SSDeepScoringFunction.java
```diff
@@ -1,7 +1,7 @@
 package datawave.query.tables.ssdeep;
 
 import java.util.Collection;
```
drewfarris, line 3: A couple unused imports here as well.

drewfarris (dismissed):

jack-gitdev (dismissed):

drewfarris (approved):

jack-gitdev (approved):

PR #2733, opened 2025-02-05
Refactoring StringUtils calls to java builtin

avgAGB:


apmoriarty (dismissed):

apmoriarty (dismissed):

jalphonso (dismissed):

warehouse/index-stats/src/main/java/datawave/mapreduce/shardStats/StatsHyperLogMapper.java
```diff
@@ -46,6 +45,7 @@ class StatsHyperLogMapper extends Mapper<Key,Value,BulkIngestKey,Value> {
     static final int DEFAULT_OUTPUT_INTERVAL = 100;
 
     private static final char NULL_CHAR = '\0';
```
awrigh11: This doesn't appear to be used anymore. Do you mind removing it?

awrigh11 (dismissed):

warehouse/index-stats/src/main/java/datawave/mapreduce/shardStats/StatsHyperLogMapper.java
avgAGB: sure, will do on next commit

awrigh11 (dismissed):

apmoriarty (approved):

foster33 (approved):

core/base-rest-responses/src/main/java/datawave/webservice/query/exception/DatawaveErrorCode.java
```diff
@@ -293,7 +291,7 @@ public String getErrorCode() {
     }
 
     public static DatawaveErrorCode findCode(String errorCode) {
-        String[] parts = StringUtils.split(errorCode, '-');
+        String[] parts = errorCode.split("-");
```
jalphonso, line 294: I'm going to ask John to look at this. I know he did some recent work with optimizing certain things so he may have opinions on the single vs double quote usage and possibly other suggestions.

avgAGB: In some of these cases the split method that he switched to needed to take in strings instead of characters for the other refactors, but I'll double check with him on this as well.

avgAGB: Talked to John a few days ago about this and he seemed to be ok with this implementation, since we're looking to use the string parameters as input across the board on the native calls. I'll rebase and re-test with queries before requesting re-reviews.

PR #2740, opened 2025-02-06
Refactor AccumuloSyntaxParserConstants so it is not an interface

### LINKED ISSUE #2739: Refactor AccumuloSyntaxParserConstants class so it is not an interface
Refactor AccumuloSyntaxParserConstants so that it is not an interface, but rather a final, uninstantiable class with defined constants.

lbschanno:
Refactor AccumuloSyntaxParserConstants so that it is not an interface that consists solely of constants, but is rather a final, uninstantiable class with constants. While not strictly an issue, it is odd for AccumuloSyntaxParser and AccumuloSyntaxParserTokenManager to both be considered implementations of AccumuloSyntaxParserConstants.

Closes #2739


apmoriarty (approved):

hlgp (approved):
lgtm

PR #2742, opened 2025-02-07
Updates needed for compatibility with Accumulo 2.1.4

ddanielr:
Accumulo 2.1.3 started using AbstractLogger which is a newer method from slf4j provided in version 2. 
zookeeper-jute was also causing Method Not Found errors when attempting to connect to curator test instances. 
The jute dependency was 3.9 (from accumulo) while the rest of the deps used a 3.8 zookeeper version.  

These changes support slfj42 being used while excluding transitive dependencies. 

Tested changes against accumulo 2.1.3 first, then updated to 2.1.4 as a separate commit.

Opening as a draft PR to discuss the merits of these changes.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -38,8 +38,7 @@
         <maven.compiler.target>11</maven.compiler.target>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <surefire.forkCount>1C</surefire.forkCount>
-        <!--version.accumulo>2.1.4-97e4684860</version.accumulo-->
-        <version.accumulo>2.1.3</version.accumulo>
+        <version.accumulo>2.1.4-97e4684860</version.accumulo>
         <version.arquillian>1.4.1.Final</version.arquillian>
         <version.arquillian-weld-ee-embedded>1.0.0.Final</version.arquillian-weld-ee-embedded>
         <version.assertj>3.20.2</version.assertj>
--- a/pom.xml
+++ b/pom.xml
@@ -38,7 +38,8 @@
         <maven.compiler.target>11</maven.compiler.target>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <surefire.forkCount>1C</surefire.forkCount>
-        <version.accumulo>2.1.2</version.accumulo>
+        <!--version.accumulo>2.1.4-97e4684860</version.accumulo-->
+        <version.accumulo>2.1.3</version.accumulo>
         <version.arquillian>1.4.1.Final</version.arquillian>
         <version.arquillian-weld-ee-embedded>1.0.0.Final</version.arquillian-weld-ee-embedded>
         <version.assertj>3.20.2</version.assertj>
@@ -121,7 +122,8 @@
         <version.powermock>2.0.9</version.powermock>
         <version.protobuf>3.16.3</version.protobuf>
         <version.protostuff>1.6.2</version.protostuff>
-        <version.slf4j>1.7.36</version.slf4j>
+        <!--version.slf4j>1.7.36</version.slf4j -->
+        <version.slf4j>2.0.12</version.slf4j>
         <version.spotify-dns>3.1.5</version.spotify-dns>
         <version.spring>5.2.2.RELEASE</version.spring>
         <version.springframework>${version.spring}</version.springframework>
@@ -858,6 +860,10 @@
                         <groupId>ch.qos.logback</groupId>
                         <artifactId>logback-classic</artifactId>
                     </exclusion>
+                    <exclusion>
+                        <groupId>ch.qos.logback</groupId>
+                        <artifactId>logback-core</artifactId>
+                    </exclusion>
                 </exclusions>
             </dependency>
             <dependency>
@@ -1034,11 +1040,6 @@
                 <groupId>org.locationtech.jts</groupId>
                 <artifactId>jts-core</artifactId>
                 <version>${version.jts}</version>
-            </dependency>
-            <dependency>
-                <groupId>org.slf4j</groupId>
-                <artifactId>jul-to-slf4j</artifactId>
-                <version>${version.slf4j}</version>
             </dependency>
             <dependency>
                 <groupId>org.slf4j</groupId>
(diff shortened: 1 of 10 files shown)


warehouse/accumulo-extensions/src/test/java/datawave/ingest/table/balancer/ShardedTableTabletBalancerTest.java
```diff
@@ -257,7 +258,7 @@ public void testNoRebalanceWithPendingMigrationsForServer() {
                 new TabletIdImpl(new KeyExtent(bar, new Text("2"), new Text("1"))),
                 new TabletIdImpl(new KeyExtent(TNAME, new Text("2"), new Text("1"))));
         //@formatter:on
-        long balanceWaitTime = testBalancer.balance(new BalanceParamsImpl(testTServers.getCurrent(), migrations, migrationsOut));
+        long balanceWaitTime = testBalancer.balance(new BalanceParamsImpl(testTServers.getCurrent(), migrations, migrationsOut, Ample.DataLevel.USER));
```
keith-turner, line 261: Accumulo PR 5195 would break this when merged as it is now.  Need to get 5195 merged before 2.1.4 release as it fixes a few balancing bugs.  The bugs it fixes are referenced in a later comment on the PR, I wrote an IT after opening the PR and found a few bugs from running that IT.

warehouse/core/pom.xml
```diff
@@ -55,6 +55,16 @@
         <dependency>
             <groupId>gov.nsa.datawave.microservice</groupId>
             <artifactId>accumulo-utils</artifactId>
+            <exclusions>
```
keith-turner, line 58: I wonder if the accumulo-utils pom marking certain deps as scope provided would allow dropping these exclusions.  For example if accumulo-utils marked zookeeper as provided in its pom then maybe zookeeper and zookeepers deps would not show up as a transitive dep for everything that uses accumulo-utils.  Not something that has to be done for this PR.

pom.xml
```diff
@@ -1557,10 +1557,20 @@
                                 <rules>
                                     <bannedDependencies>
                                         <excludes>
-                                            <exclude>log4j:log4j</exclude>
-                                            <exclude>org.slf4j:slf4j-log4j12</exclude>
-                                            <exclude>ch.qos.logback:logback-classic</exclude>
+                                            <!-- we redirect logging to log4j2, so we should have those bridges instead -->
+                                            <!-- commons-logging is allowed because it natively sends to log4j2 or slf4j -->
+                                            <exclude>ch.qos.logback:*</exclude>
```
keith-turner, line 1562: Seems like the expectation is that slf4j-api has zero dependencies.  Wonder if adding something like `<exclude>*:*</exclude>` here as the only exclude would still build and would catch if slf4j-api ever did add any dep.

warehouse/accumulo-extensions/src/test/java/datawave/ingest/table/balancer/ShardedTableTabletBalancerTest.java
ddanielr: Okay, that's good to know. I'll add a blocker label to #5195

warehouse/core/pom.xml
ddanielr: Zookeeper is listed as a dependency in the parent datawave pom so I think that would work.
I'd rather stop the transitive dep from being included initially vs adding all of these exclusions.

pom.xml
ddanielr: Adding `<exclude>*:*</exclude` bans all dependencies. 
 
Regarding slf4j adding a dependency, I think that's covered by the exclude line on 1568 `<exclude>org.slf4j:*</exclude>` and then the individual include line for only slf4j-api on line 1572 `<include>org.slf4j:slf4j-api</include>`

keith-turner: ah, I thought this section was related to just the sl4j-api but its way more than that.  Need to expand the diff.

pom.xml
```diff
@@ -121,7 +121,8 @@
         <version.powermock>2.0.9</version.powermock>
         <version.protobuf>3.16.3</version.protobuf>
         <version.protostuff>1.6.2</version.protostuff>
-        <version.slf4j>1.7.36</version.slf4j>
+        <!--version.slf4j>1.7.36</version.slf4j -->
+        <version.slf4j>2.0.12</version.slf4j>
```
keith-ratcliffe, line 125: Thanks for looking at this @ddanielr

When I was looking at fixing up the broken tests on [BulkV2-accumulo-4898 (#2727)](https://github.com/NationalSecurityAgency/datawave/pull/2727), I ran into the same issue you did here, which is that slf4j 2.x won't use the **log4j-slf4j-impl** bindings jar, requiring instead **log4j-slf4j2-impl**

FWIW, if you take a look at my 1582a112 commit there, you'll see that, rather than add the new bindings jar as a test dependency, I went ahead and replaced all the existing **log4j-slf4j-impl** deps with **log4j-slf4j2-impl**, and that wound up fixing pretty much everything

You'll also note that I updated log4j to 2.19.0, which I guess wasn't strictly necessary, but it seemed worthwhile since the log4j BOM only includes **log4j-slf4j2-impl** in 2.19.0 and up. Anyway, the log4j update forced the change to our `TestLogCollector`, and my removal of the `V2_LOCAL_MAPPING` bulk import option forced the change to `BulkIngestMapFileLoaderTest`

The only things still broken there are the two docker compose tests, which I'm pretty sure will be fixed as soon as  we can update to a rel/2.1.4 binary sourced from an apache mirror rather than from maven.pkg.github.com

warehouse/accumulo-extensions/src/test/java/datawave/ingest/table/balancer/ShardedTableTabletBalancerTest.java
alerman: @ddanielr so should this pr be blocked on waiting for 5195 to be merged and this getting updated?

ddanielr: 5195 is merged. 
Will address breaking changes as they pop up in the build

pom.xml
ddanielr: Great! I'm going to close this in favor of just merging #2727 

I had originally created this PR as a way to get some of the changes merged in without the bulk v2 changes. But at this point I think it's easier to just merge the original PR.

ddanielr:
Closing in favor of merging #2727

PR #2743, opened 2025-02-10
Hint to Shards Iterator

apmoriarty:
Add support for transforming a day hint into a scanner stream of all shards for the day

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/index/lookup/HintToShardIterator.java
+++ b/warehouse/query-core/src/main/java/datawave/query/index/lookup/HintToShardIterator.java
@@ -63,6 +63,10 @@ public class HintToShardIterator implements Iterator<Tuple2<String,IndexInfo>> {
 
     @Override
     public Tuple2<String,IndexInfo> next() {
+        if (top == null) {
+            throw new RuntimeException("Tried calling next when no top element existed");
+        }
+
         String hint;
         if (isTopDay) {
             // get next shard
--- a/warehouse/query-core/src/main/java/datawave/query/index/lookup/HintToShardIterator.java
+++ b/warehouse/query-core/src/main/java/datawave/query/index/lookup/HintToShardIterator.java
@@ -1,0 +1,102 @@
+package datawave.query.index.lookup;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.TreeSet;
+
+import org.apache.commons.jexl3.parser.JexlNode;
+
+import datawave.query.Constants;
+import datawave.query.index.lookup.RangeStream.NumShardFinder;
+import datawave.query.jexl.JexlNodeFactory;
+import datawave.query.util.Tuple2;
+
+/**
+ * Presents a 'shard-only' view of the underlying list of shard and day hints
+ * <p>
+ * This iterator accepts an arbitrary number of hints (day or shard) along with a mechanism to determine the number of shards in a given day.
+ * <p>
+ * When a day range is present and the number of shards is known this iterator will generate all possible shards for that day. This preserves the ability to
+ * conduct searches across a day in parallel.
+ */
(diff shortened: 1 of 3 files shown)


warehouse/query-core/src/main/java/datawave/query/index/lookup/HintToShardIterator.java
```diff
@@ -0,0 +1,102 @@
...
+            if (!elements.isEmpty()) {
+                top = elements.remove(0);
+                isTopDay = isDayHint(top);
+
+                if (isTopDay) {
+                    // populate shards
+                    int num = numShardFinder.getNumShards(top);
+                    for (int i = 0; i < num; i++) {
+                        shards.add(top + "_" + i);
+                    }
```
ivakegg, line 52: Do we need to reset top to be the first element in the shards list ?

warehouse/query-core/src/main/java/datawave/query/index/lookup/HintToShardIterator.java
```diff
@@ -0,0 +1,102 @@
...
+        return top != null;
+    }
+
+    private boolean isDayHint(String hint) {
+        return !hint.contains("_");
+    }
+
+    @Override
+    public Tuple2<String,IndexInfo> next() {
+        String hint;
```
ivakegg, line 70: Should we verify that top is set before continuing ?

warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
```diff
@@ -809,15 +818,101 @@ public Object visit(ASTAssignment node, Object data) {
...
-            } else {
+
+            if (shardsAndDays.length == 0) {
                 return ScannerStream.noData(myNode);
             }
+
+            // it is important that we check for a day range. in that case we need to build a different iterator
+            // to preserve search parallelism.
+            boolean hintContainsDays = checkHintForDays(shardsAndDays);
+            if (hintContainsDays) {
```
ivakegg, line 829: A little confused why we can't simply use the HintToShardIterator in any case?  How would that deny us parallelism ?

warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
```diff
@@ -809,15 +818,101 @@ public Object visit(ASTAssignment node, Object data) {
...
+                return 0;
+            }
+        }
+
+        private void populateCache() {
+            if (client == null) {
+                log.warn("no client configured, will not populate num shards");
+                return;
+            }
+            try (Scanner scanner = client.createScanner(TableName.METADATA)) {
```
ivakegg, line 894: Can we reuse any code between this and the generateNumShardsCache ?

ivakegg (changes requested):
see comments

warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
apmoriarty: If the hint or hints are all shards then we can use the existing mechanism in the range stream. This only uses the complex logic if it's necessary.

warehouse/query-core/src/main/java/datawave/query/index/lookup/HintToShardIterator.java
apmoriarty: Done in next commit

warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
apmoriarty: Almost certainly. Wrote https://github.com/NationalSecurityAgency/datawave/issues/2750 to address this.

FineAndDandy (approved):

ivakegg (approved):

PR #2745, opened 2025-02-10
Handle case when the set of fields to keep is empty

apmoriarty:
The intent in the aggregator logic is to only keep a field if the field set contains elements. The aggregator will turn into a pass-through if the field set is null. Now that logic is expanded to "pass through if null or empty"

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/functions/IdentityAggregator.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/functions/IdentityAggregator.java
@@ -106,7 +106,8 @@ public class IdentityAggregator extends SeekingAggregator implements FieldIndexA
             for (Tuple2<String,String> fieldNameValue : fieldNameValues) {
                 Attribute<?> attr = attrs.create(fieldNameValue.first(), fieldNameValue.second(), topKey, true);
                 // only keep fields that are index only and pass the attribute filter
-                boolean toKeep = (fieldsToKeep == null || fieldsToKeep.contains(JexlASTHelper.removeGroupingContext(fieldNameValue.first())))
+                boolean toKeep = (fieldsToKeep == null || fieldsToKeep.isEmpty()
+                                || fieldsToKeep.contains(JexlASTHelper.removeGroupingContext(fieldNameValue.first())))
                                 && (filter == null || filter.keep(topKey));
                 attr.setToKeep(toKeep);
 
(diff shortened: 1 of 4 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/functions/IdentityAggregator.java
```diff
@@ -106,7 +106,8 @@ public Key apply(SortedKeyValueIterator<Key,Value> itr, Document doc, AttributeF
             for (Tuple2<String,String> fieldNameValue : fieldNameValues) {
                 Attribute<?> attr = attrs.create(fieldNameValue.first(), fieldNameValue.second(), topKey, true);
                 // only keep fields that are index only and pass the attribute filter
-                boolean toKeep = (fieldsToKeep == null || fieldsToKeep.contains(JexlASTHelper.removeGroupingContext(fieldNameValue.first())))
+                boolean toKeep = (fieldsToKeep == null || fieldsToKeep.isEmpty()
```
ivakegg, line 109: The list of "indexOnlyFields" which includes the non-event fields is pased in as the fieldsToKeep from the QueryIterator.  If is is empty, then we should not be keeping any fields.  I believe this change is incorrect.

warehouse/query-core/src/main/java/datawave/query/tld/TLDFieldIndexAggregator.java
```diff
@@ -51,7 +51,7 @@ public Key apply(SortedKeyValueIterator<Key,Value> itr, Document d, AttributeFac
             Attribute<?> attr = af.create(field, value, key, true);
             // in addition to keeping fields that the filter indicates should be kept, also keep fields that the filter applies. This is due to inconsistent
             // behavior between event/tld queries where an index only field index will be kept except when it is a child of a tld
-            attr.setToKeep((fieldsToAggregate == null || fieldsToAggregate.contains(JexlASTHelper.removeGroupingContext(field)))
+            attr.setToKeep((fieldsToAggregate == null || fieldsToAggregate.isEmpty() || fieldsToAggregate.contains(JexlASTHelper.removeGroupingContext(field)))
```
ivakegg, line 54: This class is used in the TLDQueryIterator and the list of non-event fields is explicitly passed in here.  If it is empty, then we should not be keeping any of them.  I believe this change is incorrect.

ivakegg (changes requested):
See comments

apmoriarty:
Good call -- thanks!

PR #1657, opened 2022-07-29
Update Jackson to Fix Vulnerability

cmitchell:
Had to update Wildfly as well in order to update jackson, so the version of jackson is the same used by wildfly (2.12.3) .  Wildfly ships with its own version of jackson that will be used no matter how hard you try to over ride it (unless you excldue the jaxrs subsystem which is what brings in wildfly's jackson, but doing so is its own rabbit hole)

Wildfly required the upgrade of cdi, weld and arquillian dependencies as well.

This branch and PR is in place of the dependabot branch

PR #2748, opened 2025-02-11
Adding jackson version with downstream fix

avgAGB:


hlgp (approved):

hlgp:
Wonder if this will encounter the same issues as https://github.com/NationalSecurityAgency/datawave/pull/1657 wrt wildfly dependencies.  If so, additional changes may be needed and we should close https://github.com/NationalSecurityAgency/datawave/pull/1657 in lieu of this one.

avgAGB:
> Wonder if this will encounter the same issues as #1657 wrt wildfly dependencies. If so, additional changes may be needed and we should close #1657 in lieu of this one.

good point, didn't know that the older MR had done some investigation already. I'll check it out

PR #2749, opened 2025-02-11
LimitFields Updates

apmoriarty:
- LimitFields now considers the backing data when determining if an attribute is a hit attribute. Previously a PreNormalizedAttribute would fail comparison against a TypeAttribute despite having the same metadata key and backing value
- Added LimitFieldsTest and some basic test cases


warehouse/query-core/src/test/java/datawave/query/function/LimitFieldsTest.java
jalphonso, line 1: Nicely organized and readable tests!

jalphonso (approved):

dtspence (approved):

PR #2760, opened 2025-02-13
Update generate splits

alerman:


warehouse/ingest-core/src/main/java/datawave/ingest/util/GenerateShardSplits.java
```diff
@@ -87,6 +115,20 @@ public static void main(String[] args) throws Exception {
...
+            } else if (i == 4) {
+                try {
+                    splitsPerBatch = Integer.parseInt(args[i]);
+                } catch (NumberFormatException e) {
+                    System.out.println("Splits Per Batch argument is not an integer:" + e.getMessage());
+                    System.exit(-2);
+                }
+            } else if (i == 5) {
+                try {
+                    balancerDelay = Integer.parseInt(args[i]);
```
FineAndDandy: check >= 0

warehouse/ingest-core/src/main/java/datawave/ingest/util/GenerateShardSplits.java
```diff
@@ -87,6 +115,20 @@ public static void main(String[] args) throws Exception {
                     System.out.println("Split Step argument is not an integer:" + e.getMessage());
                     System.exit(-2);
                 }
+            } else if (i == 4) {
+                try {
+                    splitsPerBatch = Integer.parseInt(args[i]);
```
FineAndDandy, line 109: check >= 0

FineAndDandy (dismissed):
input validation then looks good

warehouse/ingest-core/src/main/java/datawave/ingest/util/GenerateShardSplits.java
```diff
@@ -186,4 +287,43 @@ public static void main(String[] args) throws Exception {
...
+        Set<String> tabletLocations = new HashSet<>();
+        for (Range range : rangesToWaitFor) {
+            TabletId id = locations.groupByRange().get(range).get(0);
+            tabletLocations.add(locations.getTabletLocation(id));
+            System.out.println(range.getStartKey() + "," + range.getEndKey() + "," + locations.getTabletLocation(id));
+        }
+        return tabletLocations;
+    }
+
+    protected static List<Text> calculateMidpoints(List<Text> splits) {
```
keith-turner: Could potentially pass down the output list and keep adding to it.

```suggestion
    protected static void calculateMidpoints(List<Text> splits, List<Text> midpoints) {
```

FineAndDandy (dismissed):

FineAndDandy (dismissed):

warehouse/ingest-core/src/main/java/datawave/ingest/util/GenerateShardSplits.java
```diff
@@ -153,13 +216,57 @@ public static void main(String[] args) throws Exception {
 
             startDate = DateUtils.addDays(startDate, 1);
         }
+        List<Text> results = new ArrayList(splits.size());
+        calculateMidpoints(splits, results);
```
ivakegg, line 220: Since the splits are no longer a sorted set, this calculate midpoints mechanism may not be doing what you expect?  I think you are trying to maximize the distance between splits somehow?  Or are we simply trying to create something that is no longer sorted (i.e randomized)?

jalphonso (dismissed):

jalphonso (dismissed):
discussed purpose with Adam. reapproving

warehouse/ingest-core/src/main/java/datawave/ingest/util/GenerateShardSplits.java
alerman: The batch that gets sent in is still a sorted set. This is breaking the pattern we have now, which results in accumulo generating the splits all on one tserver into a batched mechanism where the batches are in the midpoints. See the test to see the difference.

The batches are still sent as a sortedset, but are pulling from a sublist of the result of calulated midpoints such that it will result in more than one tserver participating in the split when the number of splits to be created is greater than splits per batch.

jalphonso (approved):

warehouse/ingest-core/src/main/java/datawave/ingest/util/GenerateShardSplits.java
ivakegg: "splits" you changed from a sorted set to a list.  It is hence no longer sorted as you are creating splits in numerical order.  20250606_10 should come before 20250606_2 if sorted lexographically.  That being said this PR is still doing better than what we had before so I will add an approval.

ivakegg (approved):

PR #2765, opened 2025-02-18
Remove remote only users from listEffectiveAuthorization response

jack-gitdev:


billoley (dismissed):

FineAndDandy (changes requested):
can you add a unit test for this please?

web-services/security/src/main/java/datawave/security/user/UserOperationsBean.java
```diff
@@ -133,7 +134,19 @@ private AuthorizationsListBase listEffectiveAuthorizations(Object p, boolean inc
                 for (UserOperations remote : remoteUserOperationsList) {
                     try {
                         DatawavePrincipal remotePrincipal = remote.getRemoteUser(datawavePrincipal);
-                        datawavePrincipal = WSAuthorizationsUtil.mergePrincipals(datawavePrincipal, remotePrincipal);
+                        Set<DatawaveUser> reducedRemoteProxiedUsers = new HashSet<>();
```
ivakegg: Does the user need to exist across ALL of the remove user operations AND the local principal to be used?  For our example we only have one remote system but there could be multiple in the future.

ivakegg (changes requested):
multiple remote users could be a problem

web-services/security/src/main/java/datawave/security/user/UserOperationsBean.java
jack-gitdev: I tried redoing the for loop from the perspective of the local principal, but realized we lose the ability to detect remote only users. I think we need both perspectives to get the desired result. Remote principal perspective allows us to detect remote only users. Local principal perspective allows us to check that each remote principal contains all local principal users.

jack-gitdev:
Working on unit tests.

web-services/security/src/main/java/datawave/security/user/UserOperationsBean.java
```diff
@@ -130,15 +134,37 @@ private AuthorizationsListBase listEffectiveAuthorizations(Object p, boolean inc
...
+                        if (!remoteDns.containsAll(localDns)) {
+                            log.error(localDns + " was not contained by " + remoteDns);
+                            throw new IllegalStateException("Failed to merge authorizations from remote service");
+                        }
+
+                        for (DatawaveUser user : remotePrincipal.getProxiedUsers()) {
+                            if (localDns.contains(user.getDn().subjectDN())) {
+                                reducedRemoteProxiedUsers.add(user);
+                            } else {
+                                log.debug("{} was a remote only user and has been removed", user);
```
ivakegg: maybe only use user.getDn().subjectDN() in this debug message

FineAndDandy: can we also include the auths of the remote only user that were removed?

jack-gitdev: I will make these changes.

ivakegg (approved):

FineAndDandy (approved):

PR #2776, opened 2025-02-27
Actually exercise document range tests

apmoriarty:
The test QueryLogicFactory.xml has the collapse uids option set to true. This means that even if the underlying test data contains a uid, that uid will not be persisted as a document range -- effectively turning every document range test into a shard range test. 

All document range tests now explicitly disable the collapse uids option so that document ranges are generated. 

One unit test was ignored because it now returns zero results when using a document range.

### DIFF
--- a/warehouse/query-core/src/test/java/datawave/query/IfThisTestFailsThenHitTermsAreBroken.java
+++ b/warehouse/query-core/src/test/java/datawave/query/IfThisTestFailsThenHitTermsAreBroken.java
@@ -113,6 +113,12 @@ public abstract class IfThisTestFailsThenHitTermsAreBroken {
                 PrintUtility.printTable(client, auths, QueryTestTableHelper.MODEL_TABLE_NAME);
             }
         }
+
+        @Before
+        public void setup() throws Exception {
+            super.setup();
+            logic.setCollapseUids(true);
+        }
     }
 
     @RunWith(Arquillian.class)
@@ -130,6 +136,12 @@ public abstract class IfThisTestFailsThenHitTermsAreBroken {
                 PrintUtility.printTable(client, auths, TableName.SHARD_INDEX);
                 PrintUtility.printTable(client, auths, QueryTestTableHelper.MODEL_TABLE_NAME);
             }
+        }
+
+        @Before
+        public void setup() throws Exception {
+            super.setup();
+            logic.setCollapseUids(true);
         }
     }
 
--- a/warehouse/query-core/src/test/java/datawave/query/IfThisTestFailsThenHitTermsAreBroken.java
+++ b/warehouse/query-core/src/test/java/datawave/query/IfThisTestFailsThenHitTermsAreBroken.java
@@ -141,7 +141,7 @@ public abstract class IfThisTestFailsThenHitTermsAreBroken {
         @Before
         public void setup() throws Exception {
             super.setup();
-            logic.setCollapseUids(true);
+            logic.setCollapseUids(false);
         }
     }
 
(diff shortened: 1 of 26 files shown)


hlgp:
Documenting what we talked about:  Looks like the ignored test is subject to the InMemoryAccumulo problem.  There are multiple index entries for the term and the UID where it satisfies the query gets clobbered by another.

warehouse/query-core/src/test/java/datawave/query/TestLimitReturnedGroupsToHitTermGroups.java
```diff
@@ -276,6 +300,7 @@ public void testOneGroup() throws Exception {
         Set<String> goodResults = Sets.newHashSet("CANINE.PET.13:shepherd", "CAT.PET.13:ragdoll", "FISH.PET.13:tetra", "BIRD.PET.13:lovebird",
                         "REPTILE.PET.1:snake", "DOG.WILD.1:coyote", "SIZE.CANINE.3:20,12.5", "SIZE.CANINE.WILD.1:90,26.5");
 
+        // TODO: when executing as a document range no results are found
```
hlgp: meant to delete this line since it works now

apmoriarty: done

hlgp (dismissed):

warehouse/query-core/src/test/java/datawave/query/IfThisTestFailsThenHitTermsAreBroken.java
```diff
@@ -131,6 +137,12 @@ public static void beforeAll() throws Exception {
                 PrintUtility.printTable(client, auths, QueryTestTableHelper.MODEL_TABLE_NAME);
             }
         }
+
+        @Before
+        public void setup() throws Exception {
+            super.setup();
+            logic.setCollapseUids(true);
```
FineAndDandy: false?

warehouse/query-core/src/test/java/datawave/query/util/ShapesIngest.java
```diff
@@ -134,6 +136,18 @@ public static void writeData(AccumuloClient client, RangeType type) throws Excep
         tops.create(SHARD_RINDEX);
         tops.create(METADATA);
 
+        IteratorUtil.IteratorScope[] scopes = IteratorUtil.IteratorScope.values();
+        for (IteratorUtil.IteratorScope scope : scopes) {
+            String name = "table.iterator." + scope.name() + ".UIDAggregator";
+            String opt = "table.iterator." + scope.name() + ".UIDAggregator.opt.*";
+
+            client.tableOperations().setProperty(SHARD_INDEX, name, "19,datawave.iterators.TotalAggregatingIterator");
+            client.tableOperations().setProperty(SHARD_INDEX, opt, "datawave.ingest.table.aggregator.KeepCountOnlyUidAggregator");
```
FineAndDandy, line 145: this will effectively only produce shard ranges won't it?

FineAndDandy (changes requested):

warehouse/query-core/src/test/java/datawave/query/util/ShapesIngest.java
apmoriarty: You're thinking of the `KeepCountOnlyNoUidAggregator`

warehouse/query-core/src/test/java/datawave/query/IfThisTestFailsThenHitTermsAreBroken.java
apmoriarty: Good catch

FineAndDandy (approved):

hlgp (approved):

PR #2778, opened 2025-02-28
We need the F Columns as well to enable field index hole calculations

ivakegg:


hlgp (dismissed):
Will be curious to see the memory implications of this

core/connection-pool/src/main/java/datawave/core/common/cache/BaseTableCache.java
```diff
@@ -248,11 +245,6 @@ public Boolean call() throws Exception {
 
     public void setupScanner(BatchScanner scanner) {
         scanner.setRanges(Lists.newArrayList(new Range()));
-        Map<String,String> options = new HashMap<>();
-        options.put(RegExFilter.COLF_REGEX, "^f$");
-        options.put("negate", "true");
-        IteratorSetting settings = new IteratorSetting(100, "skipFColumn", RegExFilter.class, options);
-        scanner.addScanIterator(settings);
```
apmoriarty: This should be behind a flag so that only the exact callers that need the F column get it. Otherwise all callers will start pulling far more data than necessary.

apmoriarty (changes requested):

core/connection-pool/src/main/java/datawave/core/common/cache/BaseTableCache.java
ivakegg: Yeah, I think this is the wrong approach to the problem.  I am going to rework this to cache the index holes instead of caching the entire metadata table.

ivakegg: used a different approach.  see updates.

apmoriarty (approved):

avgAGB (approved):

PR #2789, opened 2025-03-04
Add IndexScanList that generates shard specific plans

apmoriarty:
Value markers and list markers currently generate a 'full scan list' of day ranges. Day ranges are problematic, so instead these terms should create an iterator that presents a shard-specific view.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexScanList.java
+++ b/warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexScanList.java
@@ -1,0 +1,97 @@
+package datawave.query.index.lookup;
+
+import java.util.Calendar;
+import java.util.Date;
+import java.util.Iterator;
+import java.util.TreeSet;
+
+import org.apache.commons.jexl3.parser.JexlNode;
+
+import datawave.query.index.lookup.RangeStream.NumShardFinder;
+import datawave.query.util.Tuple2;
+import datawave.util.time.DateHelper;
+
+/**
+ * This iterator presents a shard-specific view of an index scan, given the date range and the number of shards per day
+ */
+public class IndexScanList implements Iterator<Tuple2<String,IndexInfo>> {
+
+    private final JexlNode node;
+    private final NumShardFinder numShardFinder;
+    private final Calendar start;
+    private final Calendar stop;
+
+    private String top = null;
+    private String day = null;
+    private final TreeSet<String> shards = new TreeSet<>();
+
+    public IndexScanList(JexlNode node, NumShardFinder numShardFinder, Date start, Date stop) {
+        this.node = node;
+        this.numShardFinder = numShardFinder;
+        this.start = getCalendarStartOfDay(start);
+        this.stop = getCalendarStartOfDay(stop);
+
+        day = DateHelper.format(start);
+        getShardsForDay();
+    }
+
+    @Override
+    public boolean hasNext() {
+        if (top == null) {
+            if (day == null) {
+                getNextDay();
+                if (day != null) {
+                    getShardsForDay();
+                }
+            }
+
+            if (!shards.isEmpty()) {
+                top = shards.pollFirst();
+                if (shards.isEmpty()) {
+                    day = null;
+                }
+            }
+        }
+        return top != null;
+    }
+
+    private void getNextDay() {
+        if (day == null) {
+            if (start.compareTo(stop) < 0) {
+                start.add(Calendar.DAY_OF_YEAR, 1);
+                day = DateHelper.format(start.getTime());
+            }
+        }
+    }
+
+    private void getShardsForDay() {
+        shards.clear();
+        int num = numShardFinder.getNumShards(day);
+        for (int i = 0; i < num; i++) {
+            shards.add(day + "_" + i);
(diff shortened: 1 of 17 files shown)


warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexScanList.java
```diff
@@ -0,0 +1,97 @@
...
+import org.apache.commons.jexl3.parser.JexlNode;
+
+import datawave.query.index.lookup.RangeStream.NumShardFinder;
+import datawave.query.util.Tuple2;
+import datawave.util.time.DateHelper;
+
+/**
+ * This iterator presents a shard-specific view of an index scan, given the date range and the number of shards per day
+ */
+public class IndexScanList implements Iterator<Tuple2<String,IndexInfo>> {
```
FineAndDandy: can this be renamed to better reflect what the class is doing?

FineAndDandy (changes requested):
just the name update please!

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexScanList.java
apmoriarty: Sure thing

FineAndDandy (dismissed):

hlgp:
Would be cool to have a full query test where the underlying numShards changes within the begin/end date range.

warehouse/query-core/src/main/java/datawave/query/index/lookup/ShardSpecificIndexIterator.java
```diff
@@ -0,0 +1,103 @@
...
+                day = DateHelper.format(start.getTime());
+            }
+        }
+    }
+
+    private void getShardsForDay() {
+        shards.clear();
+        int num = numShardFinder.getNumShards(day);
+        for (int i = 0; i < num; i++) {
+            shards.add(day + "_" + i);
```
hlgp, line 77: this is fine so long as we're never planning on starting with an arbitrary shard.  otherwise we'll need to incorporate some lexicographical arithmetic.
heaven forbid we ever have the wrong numshards, too.

apmoriarty: discussed in person

hlgp (approved):

warehouse/query-core/src/test/java/datawave/query/ColorsTest.java
```diff
@@ -0,0 +1,574 @@
...
+        withQuery("COLOR == 'red' || COLOR == 'yellow' || COLOR == 'blue'");
+        withOptionalAnyOf("COLOR:red", "COLOR:yellow", "COLOR:blue");
+        withExpectedCount(3 * getTotalEventCount());
+        planAndExecuteQuery();
+    }
+
+    @Test
+    public void testSearchAllShardsDefeatedAtFieldIndex() throws Exception {
+        withQuery("COLOR == 'red' && !COLOR == 'red'");
+        planAndExecuteQuery();
```
FineAndDandy, line 528: with expected count 0?

warehouse/query-core/src/test/java/datawave/query/ColorsTest.java
```diff
@@ -0,0 +1,574 @@
...
+    }
+
+    @Test
+    public void testSearchAllShardsDefeatedAtEvaluation() throws Exception {
+        withQuery("COLOR == 'red' && filter:includeRegex(COLOR, 'yellow')");
+        planAndExecuteQuery();
```
FineAndDandy, line 534: can you add a count expect here?

warehouse/query-core/src/test/java/datawave/query/ColorsTest.java
```diff
@@ -0,0 +1,574 @@
...
+        withExpectedCount(3);
+        withExpectedDays("20250326", "20250327");
+        withExpectedShards("20250326", ColorsIngest.getNumShards());
+        withExpectedShards("20250327", ColorsIngest.getNewShards());
+        planAndExecuteQuery();
+    }
+
+    // TODO: unique
+
+    // TODO: grouping
```
FineAndDandy, line 573: are these coming in this mr?

FineAndDandy (changes requested):
a few test questions

warehouse/query-core/src/test/java/datawave/query/ColorsTest.java
apmoriarty: the expected count defaults to zero, no need to explicitly set it.

warehouse/query-core/src/test/java/datawave/query/ColorsTest.java
apmoriarty: No, those are coming later

warehouse/query-core/src/test/java/datawave/query/ColorsTest.java
apmoriarty: same comment as above

FineAndDandy (approved):

PR #2790, opened 2025-03-05
Document Scheduler

apmoriarty:
All work here is experimental and subject to change. Comments, suggestions, etc are welcome.

The DocumentScheduler splits query execution into two phases, candidate search and candidate retrieval. The candidate search phase is a scan of the field index and the resulting record ids are pushed onto a queue. The candidate retrieval phase takes the record id and performs a document range scan. 

This approach assumes that many short-lived scans are preferable to fewer long-running scans.

```
DocumentScheduler
- QueryData iterator (consumer pushes QueryData to candidate search executor)
- candidate search executor pool (field index scan)
- record id queue (consumer pushes to record ids to the candidate retrieval executor)
- candidate retrieval executor pool (document range scan)
- result queue
```

The DocumentScheduler is configurable. 

There are two internal queues, one for record ids (candidates) and one for results. Buffering more record ids can cause candidate search scans to complete more quickly. Buffering more results can lower the page time. 

The number of threads for the candidate search and retrieval are independently configurable depending on the need. For example, one might want to lower the number of concurrent search scans but increase the number of concurrent retrieval scans. 

Each executor can also configure a different number of 'max tasks submitted' to prevent wait time. For example, the candidate retrieval executor might have 10 threads, but the maximum number of submitted tasks is 15. This ensures that the executor is always operating at capacity for maximum throughput.

Example scheduler stats
```
=== DocumentScheduler stats ===
query data seen: 1
doc/shard scans submitted: (0/1)
total next/seek calls: (12/5)
total datatype/regex/time misses: (0/0/0)
candidate scan stats: min: 10.3, avg: 10.3, max: 10.3, p50: 10.3, p95: 10.3, p99: 10.3
retrieval scan stats: min: 5, avg: 17.7, max: 141.8, p50: 6.6, p95: 141.8, p99: 141.8
total candidates/doc scans/results: (0/12/12)
```

Scheduler stats possibly interesting and hopefully useful.


warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexInfo.java
```diff
@@ -187,7 +187,7 @@ public IndexInfo union(IndexInfo first, IndexInfo o, List<JexlNode> delayedNodes
             JexlNode topLevelOr = getOrNode(sourceNode);
 
             if (null == topLevelOr) {
-                nodeSet.add(sourceNode);
+                nodeSet.add(first.myNode);
```
FineAndDandy: this is a fix regardless right?

warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
```diff
@@ -460,7 +460,9 @@ public IndexStream visit(ASTAndNode node, Object data) {
                 }
             }
             JexlNode wrappedNode = JexlNodes.wrap(node);
-            return ScannerStream.exceededValueThreshold(createFullFieldIndexScanList(config, wrappedNode).iterator(), wrappedNode);
+            IndexScanList iter = new IndexScanList(wrappedNode, getNumShardFinder(), config.getBeginDate(), config.getEndDate());
+            return ScannerStream.exceededValueThreshold(iter, wrappedNode);
+
```
FineAndDandy: this looks like the work from 2789

warehouse/query-core/src/main/java/datawave/query/index/lookup/IndexScanList.java
```diff
@@ -0,0 +1,97 @@
...
+import org.apache.commons.jexl3.parser.JexlNode;
+
+import datawave.query.index.lookup.RangeStream.NumShardFinder;
+import datawave.query.util.Tuple2;
+import datawave.util.time.DateHelper;
+
+/**
+ * This iterator presents a shard-specific view of an index scan, given the date range and the number of shards per day
+ */
+public class IndexScanList implements Iterator<Tuple2<String,IndexInfo>> {
```
FineAndDandy: this looks like the work from 2789, ignoring here

warehouse/query-core/src/main/java/datawave/next/stats/StatUtil.java
```diff
@@ -0,0 +1,22 @@
...
+     *            the elapsed time in nanoseconds
+     * @return the formatted time
+     */
+    public static String formatNanos(long ns) {
+        long millis = ns / 1_000_000;
+        if (millis > 0) {
+
+        }
+        return "formatted string";
+    }
```
FineAndDandy, line 21: just highlighting this so it isn't forgotten

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -1350,9 +1355,32 @@ protected void loadQueryModel(MetadataHelper helper, ShardQueryConfiguration con
     }
 
     protected Scheduler getScheduler(ShardQueryConfiguration config, ScannerFactory scannerFactory) {
+        // ony optimize if unique fields are not requested
```
FineAndDandy: typo

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -1350,9 +1355,32 @@ protected void loadQueryModel(MetadataHelper helper, ShardQueryConfiguration con
...
+        if (isUseDocumentScheduler() && config.getUniqueFields().isEmpty() && !isCheckpointable() && !isFullTableScanEnabled()) {
```
FineAndDandy: can you add a blurb here highlighting why this can't work with unique fields? It looks like ideally this could be part of the validation of the SimpleQueryVisitor?

FineAndDandy, line 1359: it looks like there may be other nodes besides unique fields (like functions and query markers?) that won't be supported by the SimpleQueryVisitor. Add defeats for those as well?

warehouse/query-core/src/main/java/datawave/next/SimpleQueryVisitor.java
```diff
@@ -0,0 +1,198 @@
...
+        valid = false;
+        negationsExist = true;
+        return data;
+    }
+
+    @Override
+    public Object visit(ASTFunctionNode node, Object data) {
+        // be aggressive when invalidating a query. For example, query functions are okay. But we're going to ignore that for now.
+        // valid = false;
+        // assume all functions are okay when punting to the query iterator in a doc range
```
FineAndDandy: // TODO

warehouse/query-core/src/main/java/datawave/next/SimpleQueryVisitor.java
```diff
@@ -0,0 +1,198 @@
...
+    private boolean atLeastOneFieldIndexed = false;
+
+    public SimpleQueryVisitor(Set<String> indexedFields, Set<String> indexOnlyFields, Set<String> nonEventFields) {
+        this.indexedFields = indexedFields;
+        this.indexOnlyFields = indexOnlyFields;
+        this.nonEventFields = nonEventFields;
+    }
+
+    public boolean isValid() {
+        return this.valid && this.atLeastOneFieldIndexed && !negationsExist;
```
FineAndDandy: is this condition okay with an or? Consider FRUIT:APPLE OR PROTIEN:STEAK OR CARB:FRENCHFRIES if one of these fields is indexed or index only this will return true.

warehouse/query-core/src/main/java/datawave/next/async/ContextThreadFactory.java
```diff
@@ -0,0 +1,34 @@
...
+public class ContextThreadFactory implements ThreadFactory {
+
+    private final String context;
+    private final ThreadFactory threadFactory;
+    private final ContextualUncaughtExceptionHandler uncaughtExceptionHandler;
+
+    public ContextThreadFactory(String context) {
+        this.context = context;
+        threadFactory = Executors.defaultThreadFactory();
+        uncaughtExceptionHandler = new ContextualUncaughtExceptionHandler();
```
FineAndDandy: I think this has to be linked up with the existing uncaughtExceptionHandler associated with the query

warehouse/query-core/src/main/java/datawave/next/async/ContextThreadFactory.java
```diff
@@ -0,0 +1,34 @@
...
+ * <p>
+ * It is expected that the {@link QueryDataConsumer} or {@link DocumentRangeScan} will update their thread name with something more specific.
+ */
+public class ContextThreadFactory implements ThreadFactory {
+
+    private final String context;
+    private final ThreadFactory threadFactory;
+    private final ContextualUncaughtExceptionHandler uncaughtExceptionHandler;
+
+    public ContextThreadFactory(String context) {
```
FineAndDandy, line 28: probably need an alternate constructor that lets you pass along along an existing threadFactroy and uncaughtExceptionHandler

warehouse/query-core/src/main/java/datawave/next/async/ContextualUncaughtExceptionHandler.java
```diff
@@ -0,0 +1,16 @@
+package datawave.next.async;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ContextualUncaughtExceptionHandler implements Thread.UncaughtExceptionHandler {
```
FineAndDandy, line 6: not sure this is necessary if linked to the existing uncaughtExceptionHandler associated with the query

warehouse/query-core/src/main/java/datawave/next/scanner/DocumentScheduler.java
```diff
@@ -0,0 +1,112 @@
...
+
+            config.setRecordIdFactory(new ContextThreadFactory("fi scan"));
+            config.setDocumentIdFactory(new ContextThreadFactory("doc scan"));
+
+            config.setDocIdExecutorPool(new ThreadPoolExecutor(0, config.getMaxDocIdThreads(), 15, TimeUnit.SECONDS, new LinkedBlockingQueue<>(),
+                            config.getRecordIdFactory()));
+            config.setDocumentExecutorPool(new ThreadPoolExecutor(0, config.getMaxDocumentThreads(), 15, TimeUnit.SECONDS, new LinkedBlockingQueue<>(),
+                            config.getDocumentIdFactory()));
+            config.getQueryDataConsumerExecuting().set(true);
+            config.getDocumentIdConsumerExecuting().set(true);
```
FineAndDandy: is this temporary state?

warehouse/query-core/src/main/java/datawave/next/scanner/DocumentScanner.java
```diff
@@ -0,0 +1,127 @@
...
+        hasNext();
+        return this;
+    }
+
+    @Override
+    public boolean hasNext() {
+        while ((config.getDocumentIdConsumerExecuting().get() || !config.getDocIdQueue().isEmpty() || config.getNumFiScans().get() > 0
+                        || config.getNumDocScans().get() > 0 || !config.getResults().isEmpty()) && result == null) {
+            try {
+                result = results.poll(1, TimeUnit.MILLISECONDS);
```
FineAndDandy: fast polling is potentially painful for the cpu... self promotion, the ProducerConsumerBuffer might be a better fit here. 
https://github.com/NationalSecurityAgency/datawave/blob/feature/producerConsumerBuffer/warehouse/query-core/src/main/java/datawave/query/tables/ProducerConsumerBuffer.java

FineAndDandy: or as an Iterator

https://github.com/NationalSecurityAgency/datawave/blob/feature/producerConsumerBuffer/warehouse/query-core/src/main/java/datawave/query/tables/ConsumerIterator.java

warehouse/query-core/src/main/java/datawave/next/scanner/QueryDataConsumer.java
```diff
@@ -0,0 +1,161 @@
...
+        IteratorSetting next = new IteratorSetting(settings.getPriority(), "DocIdQueryIterator", DocIdQueryIterator.class);
+        next.addOption(QueryOptions.QUERY, queryData.getQuery());
+        next.addOption(QueryOptions.START_TIME, settings.getOptions().get(QueryOptions.START_TIME));
+        next.addOption(QueryOptions.END_TIME, settings.getOptions().get(QueryOptions.END_TIME));
+        next.addOption(QueryOptions.INDEXED_FIELDS, settings.getOptions().get(QueryOptions.INDEXED_FIELDS));
+        if (settings.getOptions().containsKey(QueryOptions.DATATYPE_FILTER)) {
+            next.addOption(QueryOptions.DATATYPE_FILTER, settings.getOptions().get(QueryOptions.DATATYPE_FILTER));
+        }
+
+        Scanner scanner = config.getClient().createScanner(queryData.getTableName(), config.getAuthorizations());
```
FineAndDandy: shift this to our friend the scannerfactory or otherwise provide a way to configure/use configured hints and consistency levels

warehouse/query-core/src/main/java/datawave/next/scanner/QueryDataConsumer.java
```diff
@@ -0,0 +1,161 @@
...
+            throw new RuntimeException("Scan range differed from input range");
+        }
+
+        scanner.setRange(range);
+        scanner.addScanIterator(next);
+
+        // wait until there's room to run
+        while (numFiScans.get() >= maxFiScans) {
+            // Note: the max field index tasks submitted may exceed the number of executor threads. This
+            // effectively queues work and ensures the executor is always running at capacity.
```
FineAndDandy, line 114: this is going to lock the cpu while looping

warehouse/query-core/src/main/java/datawave/next/scanner/QueryDataConsumer.java
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        numFiScans.incrementAndGet();
+        DocumentIdProducer fiScan = new DocumentIdProducer(config, scanner, queryData);
+
+        String context = range.getStartKey().getRow().toString();
+        fiScan.setContext("fi scan " + ++fiScansSubmitted + " - " + context);
+        executor.submit(fiScan);
```
FineAndDandy: is there a good reason to thread this off from the thread its already in? So far the work this thread has done is just create a scanner, then its creating another thread.

warehouse/query-core/src/main/java/datawave/next/scanner/QueryDataConsumer.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    private void putDocId(QueryData queryData, Range range) {
+        stats.incrementNumDocScans();
+        KeyWithContext keyWithContext = new KeyWithContext(range.getStartKey(), queryData);
+
+        boolean accepted = false;
+        while (!accepted) {
+            try {
+                accepted = config.getDocIdQueue().offer(keyWithContext, 5_000, TimeUnit.MILLISECONDS);
```
FineAndDandy: another plug for the ProducerConsumerBuffer

warehouse/query-core/src/main/java/datawave/next/scanner/DocumentIdProducer.java
```diff
@@ -0,0 +1,93 @@
...
+
+                    if (key.getColumnFamily().toString().equals("STATS")) {
+                        // skip this key
+                        continue;
+                    }
+                }
+
+                offered = false;
+                while (!offered) {
+                    offered = results.offer(new KeyWithContext(key, context), 500, TimeUnit.MILLISECONDS);
```
FineAndDandy: ProduerConsumerBuffer

warehouse/query-core/src/main/java/datawave/next/scanner/DocumentIdConsumer.java
```diff
@@ -0,0 +1,91 @@
...
+            KeyWithContext keyWithContext;
+            while (producerExecuting.get() || !documentIdQueue.isEmpty() || numFiScans.get() > 0) {
+                try {
+                    keyWithContext = documentIdQueue.poll(1, TimeUnit.MILLISECONDS);
+                    if (keyWithContext != null) {
+
+                        // wait until there's room to run
+                        while (numDocScans.get() >= maxDocScans) {
+                            // Note: the max document tasks submitted may exceed the number of executor threads. This
+                            // effectively queues work and ensures the executor is always running at capacity.
```
FineAndDandy: potential cpu spinning

warehouse/query-core/src/main/java/datawave/next/scanner/DocumentIdConsumer.java
```diff
@@ -0,0 +1,91 @@
...
+                        // construct query iterator
+                        config.getStats().incrementTotalDocumentScansSubmitted();
+                        numDocScans.getAndIncrement();
+
+                        DocumentRangeScan scan = new DocumentRangeScan(keyWithContext, config);
+
+                        Key key = keyWithContext.getKey();
+                        String context = key.getRow().toString() + "-" + key.getColumnFamily().toString();
+                        scan.setContext("doc scan " + ++totalIdsConsumed + " - " + context);
+                        executor.submit(scan);
```
FineAndDandy: does this have a reason to be a new thread?

warehouse/query-core/src/main/java/datawave/next/scanner/DocumentRangeScan.java
```diff
@@ -0,0 +1,143 @@
...
+        } catch (Exception e) {
+            log.error("error executing document range {}", keyWithContext.getKey().toStringNoTime(), e);
+        } finally {
+            config.getStats().merge(stats);
+        }
+    }
+
+    private Range createRange() {
+        Key start = keyWithContext.getKey();
+        Key stop = new Key(start.getRow(), new Text(start.getColumnFamily().toString() + Constants.MAX_UNICODE_STRING));
```
FineAndDandy: this may always pull tlds

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
apmoriarty: This is kind of interesting. I originally suspected that the unique transform on the query iterator would still be necessary to avoid overwhelming the unique transform in the webservice, but I ran a test where the document scheduler would handle a query with a unique function and saw no discernable difference. 

I would like to do a more focused test with unique functions in the near future. Ideally the SimpleQueryVisitor doesn't have to deal with external state like the ShardQueryConfig. Given that, I think it's appropriate to check for unique fields here. Although I hope that check goes away entirely soon.

warehouse/query-core/src/main/java/datawave/next/SimpleQueryVisitor.java
apmoriarty: I'll add a todo here..

warehouse/query-core/src/main/java/datawave/next/scanner/DocumentRangeScan.java
apmoriarty: heh, old habits die hard.

warehouse/query-core/src/main/java/datawave/next/scanner/QueryDataConsumer.java
apmoriarty: Adding a TODO..

warehouse/query-core/src/main/java/datawave/next/scanner/QueryDataConsumer.java
FineAndDandy: this is controlled via another thread pool and allows concurrent fetching

warehouse/query-core/src/main/java/datawave/next/BaseDocIdIterator.java
```diff
@@ -0,0 +1,284 @@
...
+    protected String maxDatatypeUid;
+
+    protected String field;
+
+    protected Key tk;
+    protected final SortedKeyValueIterator<Key,Value> source;
+
+    protected boolean seenSeek = false;
+    protected final FieldIndexKey parser = new FieldIndexKey();
+    protected final DocumentIteratorStats stats = new DocumentIteratorStats();
```
FineAndDandy, line 63: might it be valuable to be able to disable stats?

warehouse/query-core/src/main/java/datawave/next/CountQueryIterator.java
```diff
@@ -0,0 +1,170 @@
...
+
+        if (options.containsKey(QueryOptions.QUERY)) {
+            String opt = options.get(QueryOptions.QUERY);
+
+            try {
+                this.script = JexlASTHelper.parseAndFlattenJexlQuery(opt);
+            } catch (ParseException e) {
+                throw new RuntimeException(e);
+            }
+        }
```
FineAndDandy: If no QueryOptions.QUERY is passed this should throw an error

warehouse/query-core/src/main/java/datawave/next/BaseDocIdIterator.java
apmoriarty: I plan to make this configurable in the near future

warehouse/query-core/src/main/java/datawave/next/CountQueryIterator.java
apmoriarty: correct! done in the next commit, thanks

warehouse/query-core/src/main/java/datawave/next/retrieval/DocumentIteratorOptions.java
```diff
@@ -0,0 +1,220 @@
...
+     *            the character set
+     * @return the decompressed data
+     * @throws IOException
+     *             if there is a deserialization exception
+     */
+    protected String decompressOption(final String buffer, Charset characterSet) throws IOException {
+        final byte[] inBase64 = Base64.decodeBase64(buffer.getBytes());
+
+        ByteArrayInputStream byteInputStream = new ByteArrayInputStream(inBase64);
+        GZIPInputStream gzipInputStream = new GZIPInputStream(byteInputStream);
```
dtspence, line 208: Would it be possible to make the compression algorithm configurable or in a follow-on ticket?

warehouse/query-core/src/main/java/datawave/next/CountScanner.java
```diff
@@ -0,0 +1,76 @@
...
+
+    private Value serializeCount(long count, ColumnVisibility cv) {
+        log.info("count query found {} documents", count);
+
+        Kryo kryo = new Kryo();
+        ResultCountingIterator.ResultCountTuple result = new ResultCountingIterator.ResultCountTuple(count, cv);
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        Output kryoOutput = new Output(baos);
+        kryo.writeObject(kryoOutput, result);
+        kryoOutput.close();
```
FineAndDandy, line 72: this looks like its copied/pasted a few times. Move to a utility class?

warehouse/query-core/src/main/java/datawave/next/DocIdIterator.java
```diff
@@ -0,0 +1,222 @@
...
+     *            a JexlNode
+     */
+    protected void parseNode(JexlNode node) {
+        this.field = JexlASTHelper.getIdentifier(node);
+        Preconditions.checkNotNull(field, "Expected field to not be null for node: " + JexlStringBuildingVisitor.buildQuery(node));
+
+        Object literal = JexlASTHelper.getLiteralValue(node);
+        Preconditions.checkNotNull(literal, "Expected value to not be null for node: " + JexlStringBuildingVisitor.buildQuery(node));
+        this.value = String.valueOf(literal);
+    }
```
FineAndDandy, line 59: should this validate an EqNode?

warehouse/query-core/src/main/java/datawave/next/DocIdIterator.java
```diff
@@ -0,0 +1,222 @@
...
+        }
+
+        while (tk == null && source.hasTop()) {
+            Key top = source.getTopKey();
+            parser.parse(top);
+
+            // apply time and datatype filter
+            if (!acceptDatatype(parser.getDatatype())) {
+                handleDatatypeMiss();
+                continue;
```
FineAndDandy, line 74:
```suggestion
                handleDatatypeMiss();
                safeNext();
                continue;
```

warehouse/query-core/src/main/java/datawave/next/DocIdIterator.java
```diff
@@ -0,0 +1,222 @@
...
+            Range range = new Range(start, false, stop, true);
+            safeSeek(range, true);
+        } else if (datatype.compareTo(maxDatatype) > 0) {
+            // rollover seek. For the EQ case generate a rollover range which causes hasNext() to be false
+            Key start = new Key(row, prefix, value + '\u0000' + maxDatatype + '\uffff');
+            Key stop = new Key(row, prefix, value + '\u0000' + '\uffff' + '\uffff');
+            Range range = new Range(start, false, stop, true);
+            safeSeek(range, true);
+        }
+    }
```
FineAndDandy, line 221: pull the common code into methods

warehouse/query-core/src/main/java/datawave/next/DocIdIteratorVisitor.java
```diff
@@ -0,0 +1,521 @@
...
+
+    protected DocIdIteratorVisitor(SortedKeyValueIterator<Key,Value> source, Range range, Set<String> datatypeFilter, LongRange timeFilter,
+                    Set<String> indexedFields) {
+        this.source = source;
+        this.range = range;
+        this.datatypeFilter = datatypeFilter;
+        this.timeFilter = timeFilter;
+        this.indexedFields = indexedFields;
+
+        this.row = range.getStartKey().getRow().toString();
```
FineAndDandy, line 114: check precondition that start key in the range is not null?

warehouse/query-core/src/main/java/datawave/next/DocIdIteratorVisitor.java
```diff
@@ -0,0 +1,521 @@
...
+                    log.debug("short circuit intersection, child returned zero hits");
+                }
+                return new HashSet<>();
+            }
+
+            if (result == null) {
+                result = scanResult;
+            } else if (scanResult.isTimeout()) {
+                log.debug("scan timed out");
+                // TODO: partial intersection with sorted sets
```
FineAndDandy, line 225: probably need to do something here?

warehouse/query-core/src/main/java/datawave/next/DocIdIteratorVisitor.java
```diff
@@ -0,0 +1,521 @@
...
+            case INDEX_HOLE:
+                log.info("found an index hole");
+                return data;
+            case EVALUATION_ONLY:
+            case DELAYED:
+            case EXCEEDED_TERM:
+            case DROPPED:
+            case STRICT:
+            case LENIENT:
+                log.debug("not handling marker of type: {}", instance.getType().getLabel());
```
FineAndDandy, line 313: should these throw an error since they won't be handled? Or is it expected they will be handled later in the evaluation just not yet?

warehouse/query-core/src/main/java/datawave/next/DocIdIteratorVisitor.java
```diff
@@ -0,0 +1,521 @@
...
+        while (iterator.hasNext()) {
+            count++;
+            result.addKey(iterator.next());
+            if (checkForTimeout && count % resultInterval == 0) {
+                elapsedScanTime = System.currentTimeMillis() - scanStart;
+                if (elapsedScanTime >= maxScanTimeMillis) {
+                    result.setTimeout(true);
+                    log.warn("term: [{}] founds {} hits before hitting timeout threshold: {}", iterator.getNode(), result.getResults().size(),
+                                    maxScanTimeMillis);
+                    break;
```
FineAndDandy, line 394: what is the impact of this? Maybe this is actually debug or trace?

warehouse/query-core/src/main/java/datawave/next/DocIdIteratorVisitor.java
```diff
@@ -0,0 +1,521 @@
...
+    }
+
+    public void setMaxScanTimeMillis(long maxScanTimeMillis) {
+        this.maxScanTimeMillis = maxScanTimeMillis;
+    }
+
+    /**
+     * This class allows us to retain knowledge of the min and max keys from a scan without requiring an expensive sorted set. Future scan ranges are restricted
+     * by the min and max
+     */
```
FineAndDandy, line 458: when doing an intersection?

warehouse/query-core/src/main/java/datawave/next/DocIdQueryIterator.java
```diff
@@ -0,0 +1,324 @@
...
+
+        if (options.containsKey(QueryOptions.QUERY)) {
+            String opt = options.get(QueryOptions.QUERY);
+
+            try {
+                this.script = JexlASTHelper.parseAndFlattenJexlQuery(opt);
+            } catch (ParseException e) {
+                throw new RuntimeException(e);
+            }
+        }
```
FineAndDandy, line 96: else throw exception

warehouse/query-core/src/main/java/datawave/next/DocIdQueryIterator.java
```diff
@@ -0,0 +1,324 @@
...
+        } else {
+            throw new RuntimeException("date not set");
+        }
+
+        if (options.containsKey(QueryOptions.INDEXED_FIELDS)) {
+            String opt = options.get(QueryOptions.INDEXED_FIELDS);
+            this.indexedFields = new HashSet<>(Splitter.on(',').splitToList(opt));
+        } else {
+            throw new RuntimeException("indexed fields not set");
+        }
```
FineAndDandy, line 120: maybe pull this code into a reusable place since its common to all the iterators?

warehouse/query-core/src/main/java/datawave/next/ListDocIdIterator.java
```diff
@@ -0,0 +1,177 @@
...
+        for (String value : values) {
+            if (datatypeFilter != null && !datatypeFilter.isEmpty()) {
+                Key start = new Key(this.row, this.prefix, value + "\u0000" + datatypeFilter.first() + "\u0000");
+                Key end = new Key(this.row, this.prefix, value + "\u0000" + datatypeFilter.last() + "\u0000" + Constants.MAX_UNICODE_STRING);
+                ranges.add(new Range(start, true, end, false));
+            } else {
+                Key start = new Key(this.row, this.prefix, value + "\u0000");
+                Key end = new Key(this.row, this.prefix, value + "\u0000" + Constants.MAX_UNICODE_STRING);
+                ranges.add(new Range(start, true, end, false));
+            }
```
FineAndDandy, line 73: seen these ranges generated in a few different classes. Move to common code

warehouse/query-core/src/main/java/datawave/next/ListDocIdIterator.java
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        if (tk == null) {
+            // iterator exhausted, log stats
+            if (log.isDebugEnabled()) {
+                log.debug("stats for term: [{}]: {}", getNode(), stats);
+            }
+        }
+
+        return tk != null;
```
FineAndDandy, line 125: common code with DocIdIterator

warehouse/query-core/src/main/java/datawave/next/ListDocIdIterator.java
```diff
@@ -0,0 +1,177 @@
...
+            Range range = new Range(start, false, stop, true);
+            safeSeek(range, true);
+            return; // no further datatypes exist, so we're done
+        }
+
+        // otherwise seek to the next datatype and continue iterating
+        Key start = new Key(row, prefix, value + '\u0000' + nextDatatype);
+        Key stop = new Key(row, prefix, value + '\u0000' + '\uffff');
+        Range range = new Range(start, false, stop, true);
+        safeSeek(range, true);
```
FineAndDandy, line 175: extract common code with other iterators

warehouse/query-core/src/main/java/datawave/next/RegexDocIdIterator.java
```diff
@@ -0,0 +1,260 @@
...
+            if (!matchesMinMaxDatatypeUid()) {
+                handleMinMaxDatatypeUidMiss();
+                continue;
+            }
+
+            // apply datatype filter
+            if (!acceptDatatype(parser.getDatatype())) {
+                handleDatatypeMiss();
+                continue;
+            }
```
FineAndDandy, line 65: missing safeNext();

warehouse/query-core/src/main/java/datawave/next/ListDocIdIterator.java
```diff
@@ -0,0 +1,177 @@
...
+            }
+
+            Key top = source.getTopKey();
+            parser.parse(top);
+
+            // apply time and datatype filter
+            if (datatypeFilter != null && !datatypeFilter.isEmpty() && !datatypeFilter.contains(parser.getDatatype())) {
+                handleDatatypeMiss();
+                continue;
+            }
```
FineAndDandy, line 104:
```suggestion
            // apply time and datatype filter
            if (datatypeFilter != null && !datatypeFilter.isEmpty() && !datatypeFilter.contains(parser.getDatatype())) {
                handleDatatypeMiss();
                safeNext();
                continue;
            }
```

warehouse/query-core/src/main/java/datawave/next/retrieval/DocumentIterator.java
```diff
@@ -0,0 +1,201 @@
...
+        checkForScanRebuild();
+
+        // aggregate document
+        AttributeFactory attributeFactory = new AttributeFactory(typeMetadata);
+        EventKey parser = new EventKey();
+
+        for (String candidate : candidates) {
+
+            Range candidateRange = rangeForCandidate(candidate);
+            source.seek(candidateRange, excludeCFs, false);
```
FineAndDandy, line 100: calling seek will not only seek, but also aggregate all the documents and evaluate them without waiting for a next call. This seems to break the normal iterator pattern? Also could there be memory concerns here with adding n documents regardless of size into memory?

warehouse/query-core/src/main/java/datawave/next/scanner/DocumentIdConsumer.java
```diff
@@ -0,0 +1,100 @@
...
+                keyWithContext = documentIdQueue.poll(candidateQueuePollTimeMillis, TimeUnit.MILLISECONDS);
+                if (keyWithContext != null) {
+
+                    // wait until there's room to run
+                    while (numRetrievalScans.get() >= maxRetrievalTasks) {
+                        // Note: the max document tasks submitted may exceed the number of executor threads. This
+                        // effectively queues work and ensures the executor is always running at capacity.
+                        Thread.onSpinWait();
+                        Thread.sleep(10);
+                    }
```
FineAndDandy, line 68: track the total time waiting?

warehouse/query-core/src/main/java/datawave/next/scanner/DocumentRangeScan.java
```diff
@@ -0,0 +1,250 @@
...
+
+        // set the query from the query data
+        if (setting.getOptions().containsKey(QueryOptions.QUERY)) {
+            setting.addOption(QueryOptions.QUERY, queryData.getQuery());
+        }
+
+        // serialize query
+        setting.addOption(QueryOptions.SERIAL_EVALUATION_PIPELINE, "true");
+        setting.addOption(QueryOptions.MAX_EVALUATION_PIPELINES, "1");
+        setting.addOption(QueryOptions.MAX_PIPELINE_CACHED_RESULTS, "1");
```
FineAndDandy, line 232: I think these are only used by the original iterator? Probably don't need to be set if using the document iterator

warehouse/query-core/src/main/java/datawave/next/scanner/KeyWithContext.java
```diff
@@ -0,0 +1,58 @@
...
+
+    public KeyWithContext(Key key, QueryData context, boolean buildCompareKey) {
+        this.key = key;
+        this.context = context;
+
+        if (buildCompareKey) {
+            // the ColumnFamily is datatype \0 uid, invert this for priority queue
+            String cf = key.getColumnFamily().toString();
+            int index = cf.indexOf('\u0000');
+            compareKey = cf.substring(index + 1) + cf.substring(0, index);
```
FineAndDandy, line 36: this is dropping the null byte from the key, is this desirable?

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
FineAndDandy, line 1: can you describe what these changes are attempting to do?

FineAndDandy (changes requested):
lots of good tests. Some minor things and some suggestions on reusable code you can take or leave.

warehouse/query-core/src/main/java/datawave/next/stats/StatUtil.java
apmoriarty: Created https://github.com/NationalSecurityAgency/datawave/issues/2909

warehouse/query-core/src/main/java/datawave/next/scanner/DocumentScheduler.java
apmoriarty: This was removed at one point, so I will assume it was in fact temporary.

warehouse/query-core/src/main/java/datawave/next/async/ContextThreadFactory.java
apmoriarty: Created https://github.com/NationalSecurityAgency/datawave/issues/2910

warehouse/query-core/src/main/java/datawave/next/async/ContextThreadFactory.java
apmoriarty: Created https://github.com/NationalSecurityAgency/datawave/issues/2910

warehouse/query-core/src/main/java/datawave/next/async/ContextualUncaughtExceptionHandler.java
apmoriarty: Created https://github.com/NationalSecurityAgency/datawave/issues/2910

warehouse/query-core/src/main/java/datawave/next/scanner/DocumentScanner.java
apmoriarty: Created https://github.com/NationalSecurityAgency/datawave/issues/2912

warehouse/query-core/src/main/java/datawave/next/scanner/QueryDataConsumer.java
apmoriarty: Created https://github.com/NationalSecurityAgency/datawave/issues/2912

warehouse/query-core/src/main/java/datawave/next/scanner/DocumentIdProducer.java
apmoriarty: Created https://github.com/NationalSecurityAgency/datawave/issues/2912

warehouse/query-core/src/main/java/datawave/next/scanner/DocumentIdConsumer.java
apmoriarty: This could be addressed by https://github.com/NationalSecurityAgency/datawave/issues/2912

warehouse/query-core/src/main/java/datawave/next/retrieval/DocumentIteratorOptions.java
apmoriarty: Absolutely https://github.com/NationalSecurityAgency/datawave/issues/2913

warehouse/query-core/src/main/java/datawave/next/CountScanner.java
apmoriarty: Expanded the scope of https://github.com/NationalSecurityAgency/datawave/issues/2913 to include this.

warehouse/query-core/src/main/java/datawave/next/DocIdIterator.java
apmoriarty: The validation happens externally in the calling class, DocIdIteratorVisitor

warehouse/query-core/src/main/java/datawave/next/DocIdIterator.java
apmoriarty: I will look at this

warehouse/query-core/src/main/java/datawave/next/DocIdIteratorVisitor.java
apmoriarty: Created https://github.com/NationalSecurityAgency/datawave/issues/2914

warehouse/query-core/src/main/java/datawave/next/DocIdIteratorVisitor.java
apmoriarty: These are markers that are not handled today, and in most cases they should be removed from the query during planning. I wanted to record the existence of these markers in case we ever miss something. 

I don't think it's worth outright failing the query.

warehouse/query-core/src/main/java/datawave/next/DocIdIteratorVisitor.java
apmoriarty: This is what allows future scans to have a more restrictive start and end key, when applicable.

warehouse/query-core/src/main/java/datawave/next/DocIdQueryIterator.java
apmoriarty: Yes. Will include under: https://github.com/NationalSecurityAgency/datawave/issues/2910

warehouse/query-core/src/main/java/datawave/next/retrieval/DocumentIterator.java
apmoriarty: As it exists today this absolutely breaks the iterator pattern and is a memory concern. The goal for this class is to do async aggregation/evaluation while returning candidates. Probably a good use case for the producer consumer buffer.

apmoriarty: Created https://github.com/NationalSecurityAgency/datawave/issues/2915

warehouse/query-core/src/main/java/datawave/next/DocIdIterator.java
apmoriarty: General cleanup here https://github.com/NationalSecurityAgency/datawave/issues/2916

warehouse/query-core/src/main/java/datawave/next/DocIdIterator.java
apmoriarty: Noted in https://github.com/NationalSecurityAgency/datawave/issues/2916

warehouse/query-core/src/main/java/datawave/next/DocIdIteratorVisitor.java
apmoriarty: That's an open question, noted as part of https://github.com/NationalSecurityAgency/datawave/issues/2916

warehouse/query-core/src/main/java/datawave/next/DocIdQueryIterator.java
apmoriarty: Noted as part of https://github.com/NationalSecurityAgency/datawave/issues/2916

warehouse/query-core/src/main/java/datawave/next/ListDocIdIterator.java
apmoriarty: Yup: https://github.com/NationalSecurityAgency/datawave/issues/2916

warehouse/query-core/src/main/java/datawave/next/ListDocIdIterator.java
apmoriarty: Yep: https://github.com/NationalSecurityAgency/datawave/issues/2916

warehouse/query-core/src/main/java/datawave/next/ListDocIdIterator.java
apmoriarty: Yup: https://github.com/NationalSecurityAgency/datawave/issues/2916

warehouse/query-core/src/main/java/datawave/next/ListDocIdIterator.java
apmoriarty: Part of https://github.com/NationalSecurityAgency/datawave/issues/2916

warehouse/query-core/src/main/java/datawave/next/scanner/DocumentIdConsumer.java
apmoriarty: I was going to do that at one point, but at I think the producer/consumer work will supersede this https://github.com/NationalSecurityAgency/datawave/issues/2912

warehouse/query-core/src/main/java/datawave/next/scanner/DocumentRangeScan.java
apmoriarty: Correct. Updated the description of https://github.com/NationalSecurityAgency/datawave/issues/2916

warehouse/query-core/src/main/java/datawave/next/scanner/KeyWithContext.java
apmoriarty: Unintended, but doesn't have any bad side effects

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
apmoriarty: It allows the visitor function to apply to a query data in addition to a scanner chunk

warehouse/query-core/src/main/java/datawave/next/RegexDocIdIterator.java
apmoriarty: https://github.com/NationalSecurityAgency/datawave/issues/2916

warehouse/query-core/src/main/java/datawave/next/DocIdIteratorVisitor.java
apmoriarty: https://github.com/NationalSecurityAgency/datawave/issues/2916

warehouse/query-core/src/main/java/datawave/next/CountScanner.java
```diff
@@ -0,0 +1,76 @@
...
+            return new Result<>(sum.getKey(), value);
+        }
+
+        return null;
+    }
+
+    private Value serializeCount(long count, ColumnVisibility cv) {
+        log.info("count query found {} documents", count);
+
+        Kryo kryo = new Kryo();
```
dtspence, line 67: Minor optimization - could we move `Kryo` to member variable, if thread-safe? There is some overhead for allocating the `Kryo` object.

dtspence (approved):

warehouse/query-core/src/main/java/datawave/next/CountScanner.java
apmoriarty: Tracked under https://github.com/NationalSecurityAgency/datawave/issues/2938

FineAndDandy (approved):

PR #2795, opened 2025-03-07
rename field CONTENT_SUMMARY to SUMMARY

austin007008:
rename field CONTENT_SUMMARY to SUMMARY

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
+++ b/warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
@@ -269,6 +269,8 @@ public class QueryOptions implements OptionDescriber {
 
     public static final String SUMMARY_ITERATOR = "summary.iterator.class";
 
+    public static final String SUMMARY_FIELD_NAME = "summary.field.name";
+
     // field and next thresholds before a seek is issued
     public static final String FI_FIELD_SEEK = "fi.field.seek";
     public static final String FI_NEXT_SEEK = "fi.next.seek";
@@ -448,6 +450,8 @@ public class QueryOptions implements OptionDescriber {
 
     protected Class<? extends SortedKeyValueIterator<Key,Value>> summaryIterator = ContentSummaryIterator.class;
 
+    protected String summaryFieldname;
+
     // off by default, controls when to issue a seek
     private int fiFieldSeek = -1;
     private int fiNextSeek = -1;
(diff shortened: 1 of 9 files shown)


avgAGB (dismissed):

emiliodskinner (dismissed):

FineAndDandy (changes requested):
can this be configurable?

drewfarris (dismissed):

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -448,6 +450,8 @@ public class QueryOptions implements OptionDescriber {
 
     protected Class<? extends SortedKeyValueIterator<Key,Value>> summaryIterator = ContentSummaryIterator.class;
 
+    protected String summaryFieldname;
```
ivakegg, line 453: Do we want a default fieldname set here?

ivakegg (dismissed):

hlgp (dismissed):

FineAndDandy (dismissed):

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -1865,6 +1865,14 @@ public void setSummaryIterator(String iteratorClass) {
         }
     }
 
+    public String getSummaryFieldname() {
```
keith-ratcliffe: When I run the quickstart locally, I'm able to replicate the test failure.
In Query.log, there are a ton of these...
```
Error setting property values; nested exception is org.springframework.beans.NotWritablePropertyException: Invalid property 'summaryFieldName' of bean class [datawave.query.tables.ShardQueryLogic]: Bean property 'summaryFieldName' is not writable or has an invalid setter method. Did you mean 'summaryFieldname'?
```
So, you'll need to either change the setter/getter names here to match what Spring expects based on your current QLF.xml (`get/setSummaryFieldName`) or change QLF.xml to use `"summaryFieldname"` instead. Seems like the former is probably what you intended

keith-ratcliffe (changes requested):

keith-ratcliffe (approved):

avgAGB (approved):

ivakegg (approved):

ivakegg (approved):

PR #2797, opened 2025-03-11
Address race condition where expansion is successful but more values …

apmoriarty:
…are added before the result is logged

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/BaseIndexExpansionVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/BaseIndexExpansionVisitor.java
@@ -278,19 +278,19 @@ public abstract class BaseIndexExpansionVisitor extends RebuildingVisitor {
             return;
         }
 
-        if (lookupMap.containsKey(field)) {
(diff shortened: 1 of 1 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/BaseIndexExpansionVisitor.java
```diff
@@ -284,7 +285,11 @@ protected void logResult(String stage, JexlNode node, IndexLookupMap lookupMap)
             } else if (lookupMap.get(field).isThresholdExceeded()) {
                 log.debug("{} expansion for term [{}] failed (threshold)", stage, term);
             } else {
-                log.debug("{} expansion for term [{}] success ({} values)", stage, term, lookupMap.get(field).size());
+                try {
+                    log.debug("{} expansion for term [{}] success ({} values)", stage, term, lookupMap.get(field).size());
```
jschmidt10, line 289: The exception we hit was originating from the `lookupMap.containsKey()` call above. We might want to wrap from like 281 down in the try/catch.

jschmidt10 (approved):

hlgp (approved):

PR #2803, opened 2025-03-14
Pom updates

jalphonso:
Update plugin versions  
Ensure all modules use maven enforcer  
Remove deprecated plugins  
Clean up dependency tree
Reworked docker builds using jkube kubernetes-maven-plugin
Formatting changes are mostly isolated in one commit


core/pom.xml
```diff
@@ -69,18 +58,6 @@
...
                         <runOrder>random</runOrder>
                     </configuration>
-                    <dependencies>
-                        <dependency>
-                            <groupId>org.junit.jupiter</groupId>
-                            <artifactId>junit-jupiter-engine</artifactId>
-                            <version>${version.junit.bom}</version>
-                        </dependency>
-                        <dependency>
-                            <groupId>org.junit.vintage</groupId>
```
ivakegg, line 79: Are you sure we do not need this?  If it is not included then you will find that not all of the test cases are run.  Same comment for all places you are removing this dependency.

jalphonso: Thank you for catching this. I added/moved the junit/jupiter dependencies to the top level parents -> datawave-parent and datawave-microservice-parent and verified the tests are being run now. Eventually we might want to migrate the tests to JUnit5. IntelliJ can mostly do this for us but I'll leave that for another ticket. https://www.jetbrains.com/guide/java/tutorials/migrating-junit4-junit5/automatically-migrate-all-tests/

warehouse/assemble/datawave/pom.xml
```diff
@@ -475,33 +459,61 @@
...
-                            </resources>
+                            <verbose>true</verbose>
+                            <images>
+                                <image>
+                                    <name>${docker.image.prefix}datawave/ingest-${build.env}</name>
+                                    <build>
+                                        <!-- defaults to ${project.basedir}/src/main/docker when using relative paths -->
+                                        <contextDir>.</contextDir>
+                                        <buildOptions>
+                                            <memory>8589934592</memory>
```
alerman, line 474: Does this require the build machine have at least 8.5Gb of memory available? Should we add a note to the readme?

alerman (dismissed):

warehouse/assemble/datawave/pom.xml
jalphonso: This is a memory limit for the build container only (default is unlimited); it doesn't affect the actual container ran later. I forget where I got this from but figured it can't hurt to keep it.

alerman (approved):

avgAGB (approved):

PR #2804, opened 2025-03-14
Fix scan hints and consistency levels

jack-gitdev:


apmoriarty:
Could you update the description with a brief overview of what use cases were missed?

PR #2805, opened 2025-03-14
RegexIndexExpansion not compiling

avgAGB:


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RegexIndexExpansionVisitor.java
```diff
@@ -85,12 +84,8 @@ protected RegexIndexExpansionVisitor(ShardQueryConfiguration config, ScannerFact
...
 
         if (config.isLimitTermExpansionToModel()) {
-            try {
-                QueryModel queryModel = helper.getQueryModel(config.getModelTableName(), config.getModelName());
-                this.onlyUseThese = queryModel.getForwardQueryMapping().values();
-            } catch (ExecutionException e) {
-                this.onlyUseThese = null;
-            }
+            QueryModel queryModel = helper.getQueryModel(config.getModelTableName(), config.getModelName());
+            this.onlyUseThese = queryModel.getForwardQueryMapping().values();
```
keith-ratcliffe, line 88: Thanks! I just noticed the compile breakage too. Wondering if we should do something like this, to ensure we still continue processing the query, same as before. Thoughts?
```suggestion
            try {
                QueryModel queryModel = helper.getQueryModel(config.getModelTableName(), config.getModelName());
                this.onlyUseThese = queryModel.getForwardQueryMapping().values();
            } catch (Exception e) {
                this.onlyUseThese = null;
            }
```

ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RegexIndexExpansionVisitor.java
ivakegg: Actuall I reverted this back out.  The underlying query model was only throwing ExecutionException because of backward compatibility.  That exception was never actualy thrown.  I merely pulled it off of the method signature which subsequently caused this issue.

ivakegg (approved):

PR #2807, opened 2025-03-18
datawave-code-style stop indenting blank lines

jalphonso:
we are removing trailing whitespace in the formatter config which occurs after the formatter rules have been applied so we should stop indenting blank lines in those rules. The `contrib/datawave-utils/code-style/src/main/resources/eclipse/Eclipse-Datawave-Codestyle.xml` can be imported into IntelliJ so the IDE formats the same way as Maven on the command line.  
`Settings -> Editor -> Code Style -> Scheme -> Gear icon -> Import Scheme -> Eclipse XML Profile`

### DIFF
--- a/contrib/datawave-utils/code-style/pom.xml
+++ b/contrib/datawave-utils/code-style/pom.xml
@@ -3,7 +3,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>gov.nsa.datawave</groupId>
     <artifactId>datawave-code-style</artifactId>
-    <version>1.1-SNAPSHOT</version>
+    <version>1.0.1-SNAPSHOT</version>
     <packaging>jar</packaging>
     <name>DataWave Code Formatter</name>
     <description>This pom configures the formatter-maven-plugin to format the code
--- a/contrib/datawave-utils/code-style/src/main/resources/eclipse/Eclipse-Datawave-Codestyle.xml
+++ b/contrib/datawave-utils/code-style/src/main/resources/eclipse/Eclipse-Datawave-Codestyle.xml
@@ -159,7 +159,7 @@
 <setting id="org.eclipse.jdt.core.formatter.alignment_for_throws_clause_in_constructor_declaration" value="16"/>
 <setting id="org.eclipse.jdt.core.formatter.insert_space_after_assignment_operator" value="insert"/>
 <setting id="org.eclipse.jdt.core.formatter.insert_space_before_assignment_operator" value="insert"/>
-<setting id="org.eclipse.jdt.core.formatter.indent_empty_lines" value="true"/>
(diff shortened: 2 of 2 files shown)


contrib/datawave-utils/code-style/pom.xml
```diff
@@ -3,7 +3,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>gov.nsa.datawave</groupId>
     <artifactId>datawave-code-style</artifactId>
-    <version>1.1-SNAPSHOT</version>
+    <version>1.0.1-SNAPSHOT</version>
```
alerman, line 6: Why is this changng? Did we release a 1.0? is this in preparation for releasing a 1.0.1?

jalphonso: semantic versioning. 1.0 is our only release. I will release 1.0.1 right after this gets merged in so I can update the formatter to use it.

contrib/datawave-utils/code-style/src/main/resources/eclipse/Eclipse-Datawave-Codestyle.xml
```diff
@@ -159,7 +159,7 @@
 <setting id="org.eclipse.jdt.core.formatter.alignment_for_throws_clause_in_constructor_declaration" value="16"/>
 <setting id="org.eclipse.jdt.core.formatter.insert_space_after_assignment_operator" value="insert"/>
 <setting id="org.eclipse.jdt.core.formatter.insert_space_before_assignment_operator" value="insert"/>
-<setting id="org.eclipse.jdt.core.formatter.indent_empty_lines" value="true"/>
+<setting id="org.eclipse.jdt.core.formatter.indent_empty_lines" value="false"/>
```
alerman, line 162: we do preserve empty lines (up to 1) below. Do these rules get executed in order? Should this be true but only placed after the rule preserving only 1 empty line is run?

jalphonso: Order does not seem to matter. I tested with a locally installed SNAPSHOT and when `indent_empty_lines` is `true` and the `removeTrailingWhiteSpace` is `false`, it indents all the blank lines even if I move it after the preserve line. I opened this MR because I use the formatter's Eclipse profile in IntelliJ and noticed the IDE was fighting maven on the command line. The decision to remove trailing whitespace occurred in this MR: https://github.com/NationalSecurityAgency/datawave/pull/1988

jalphonso: also with `indent_empty_lines` set to `false` and `removeTrailingWhiteSpace` set to `false` no new formatting changes are present

apmoriarty (approved):

foster33 (approved):

PR #2812, opened 2025-03-19
Preserve token context for term frequency entries to avoid position overlap across multiple field instances

FineAndDandy:
In some cases, we receive multiple values that get mapped to the same field name, e.g., CONTENT.  When these fields are received independently in separate ingest passes and then tokenized, we wind up with tokens in the term frequency portion of the shard table that have offsets that overlap. This will become more frequent as we implement annotation indexing.

For example, one CONTENT field value might be CONTENT:"cat mat" and another value might be "dog bog". In this case, it's possible under certain circumstances that a query CONTENT:"cat bog" would hit on this field and require further filtering to determine that it's a false positive. 

This feature adds the ability to employ grouping notation, similar to how it's used on the event fields, to maintain a distinction between the contexts within which a token appears as a part of the term frequency entry.  It will involve modifications to ingest and query to preserve this extra context in the form of dot notation.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionEvaluator.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionEvaluator.java
@@ -14,6 +14,7 @@ import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Lists;
 
 import datawave.ingest.protobuf.TermWeightPosition;
+import datawave.query.jexl.JexlASTHelper;
 import datawave.query.postprocessing.tf.TermOffsetMap;
 
 /**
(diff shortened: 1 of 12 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/functions/ContentFunctionEvaluator.java
```diff
@@ -190,8 +219,16 @@ public Collection<String> evaluate() {
...
                 for (String field : offsetsByField.keySet()) {
-                    if (!fields.isEmpty() && !fields.contains(field)) {
-                        continue;
+                    if (!fields.isEmpty()) {
+                        // account for trailing dot notation
+                        String compareField = field;
+                        int dotIndex = field.indexOf(".");
+                        if (dotIndex > -1) {
+                            compareField = field.substring(0, dotIndex);
+                        }
```
apmoriarty: This feels ripe for a static utility method, similar to the ASTHelper's deconstruct identifier method.

apmoriarty (dismissed):
LGTM, one minor comment

drewfarris:
No downstream effects without modifications to downstream codebase to produce tokens with context.

drewfarris:
Going to close this one and reopen with me as the owner.

PR #2820, opened 2025-03-27
Fix incorrect step in docker README

lbschanno:
Fix step in README that instructs how to start the wildfly deployment via Docker.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -201,10 +201,20 @@ This command also prevents the microservice services from building with `-DskipS
 If you ever need to rebuild the Datawave quickstart docker image, but don't want to ingest the sample data you can add `-DskipIngest` to 
 your build command.  This can save you some time, since the docker compose configuration stores ingested data in a persistent volume.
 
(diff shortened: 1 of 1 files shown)


foster33 (dismissed):

foster33:
Just a note. Starting wildfly can also be controlled via `/docker/docker-compose.yml`.

By default, when the docker quickstart is started it runs `datawave-bootstrap.sh` and sends it `--accumulo` ([See here](https://github.com/NationalSecurityAgency/datawave/blob/6ff69d6dcb57d86852abf0296ea4faa8adbda719/docker/docker-compose.yml#L11)). This will not start wildfly. However, if you change `--accumulo` to `--web` in the docker-compose file, it will run the wildfly webservice when the docker services are deployed. Not sure if that is worth mentioning in the docs as well.

lbschanno:
@foster33 that is true, good catch. I added information about modifying docker-compose.yml to start wildfly to the steps.

foster33 (approved):

avgAGB (approved):

PR #2825, opened 2025-03-31
Implements keyword extraction for content stored in Datawave

drewfarris:
* Added YAKE! keyword extraction algorithm in warehouse/keyword-common module.
* Enhanced baseline YAKE implementation with multilingual stopwords and sentence breaking.
* First cut at KeywordQueryLogic to run keyword extraction on stored content.


warehouse/keyword-common/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+    <parent>
+        <groupId>gov.nsa.datawave</groupId>
+        <artifactId>datawave-warehouse-parent</artifactId>
+        <version>7.18.0-SNAPSHOT</version>
+    </parent>
+    <artifactId>datawave-keyword-common</artifactId>
+    <properties>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
drewfarris: I should probably remove these and inherit whatever the parent sets.

warehouse/keyword-common/src/main/java/datawave/util/keyword/language/YakeLanguage.java
```diff
@@ -0,0 +1,85 @@
...
+        static final Map<String,YakeLanguage> languageRegistry = new HashMap<>();
+
+        public static void add(YakeLanguage language) {
+            languageRegistry.put(language.getLanguageCode(), language);
+        }
+
+        public static YakeLanguage find(String languageCode) {
+            return languageRegistry.get(languageCode);
+        }
+    }
```
FineAndDandy: any concern about thread safety?

warehouse/keyword-common/src/main/java/datawave/util/keyword/YakeKeywordExtractor.java
```diff
@@ -0,0 +1,718 @@
...
+     * @param input
+     *            input to break into sentences.
+     * @return a list of sentences.
+     */
+    public List<String> breakSentences(String input) {
+        // todo: investigate better sentence breaking
+        List<String> sentences = new ArrayList<>();
+        synchronized (sentenceBreakIterator) {
+            sentenceBreakIterator.setText(input);
+            sentenceBreakIterator.setText(input);
```
FineAndDandy: duplicate call?

warehouse/keyword-common/src/main/java/datawave/util/keyword/YakeKeywordExtractor.java
```diff
@@ -0,0 +1,718 @@
...
+        //@formatter:off
+        final List<String> sentenceTokens = Arrays.stream(sentence.split("\\s+")) // split on space
+                .filter(Predicate.not(String::isEmpty))
+                .map(s -> s.endsWith(".") ? s.substring(0, s.length()-1) : s) // remove trailing periods
+                .collect(Collectors.toList());
+        //@formatter:on
+
+        if (!sentenceTokens.isEmpty()) {
+            sentenceTokens.add("."); // always as a placeholder for end-of-sentence.
+        }
```
FineAndDandy, line 670: might be answered later, but why do we hold perids?

warehouse/keyword-common/src/main/java/datawave/util/keyword/YakeKeywordExtractor.java
```diff
@@ -0,0 +1,718 @@
...
+        final List<TokenValue> resultsFlattenedIndexed = new ArrayList<>();
+
+        //@formatter:off
+        IntStream.range(0, tokenizedSentences.size())
+                .mapToObj(i ->
+                        tokenizedSentences.get(i)
+                                .stream()
+                                .map(t -> new TokenValue(t, i)))
+                .reduce(Stream.empty(), Stream::concat)
+                .collect(Collectors.toCollection(() -> resultsFlattenedIndexed));
```
FineAndDandy: this is fairly complicated for what it does. This is a little simpler if you want to use streams

```suggestion
        AtomicInteger sentenceIndex = new AtomicInteger(0);
        return tokenizedSentences.stream()
                .flatMap(sentence -> {
                    int currentIndex = sentenceIndex.getAndIncrement();
                    return sentence.stream().map(token -> new TokenValue(token, currentIndex));
                })
                .collect(Collectors.toCollection(ArrayList::new));
```

warehouse/keyword-common/src/main/java/datawave/util/keyword/YakeKeywordExtractor.java
```diff
@@ -0,0 +1,718 @@
...
+                            }
+
+                            // otherwise there are not enough tokens to form a full window, return nothing.
+                            return IntStream.empty();
+                        })) // return a stream of streams of all windows
+                .reduce(Stream.empty(), Stream::concat)// reduce them to a single stream
+                .map(k -> k.mapToObj(tokens::get) // get a string containing the tokens for each position in each list of positions.
+                        .collect(Collectors.toList()))
+                .filter(s -> !s.isEmpty())).collect(Collectors.toList());
+    }
```
FineAndDandy, line 362: if you want to keep it using streams this method could also be simplified a bit like calculateBasicStats

warehouse/keyword-common/src/main/java/datawave/util/keyword/YakeKeywordExtractor.java
```diff
@@ -0,0 +1,718 @@
...
+                                return IntStream.range(start, end);
+                            }
+
+                            // otherwise there are not enough tokens to form a full window, return nothing.
+                            return IntStream.empty();
+                        })) // return a stream of streams of all windows
+                .reduce(Stream.empty(), Stream::concat)// reduce them to a single stream
+                .map(k -> k.mapToObj(tokens::get) // get a string containing the tokens for each position in each list of positions.
+                        .collect(Collectors.toList()))
+                .filter(s -> !s.isEmpty())).collect(Collectors.toList());
```
FineAndDandy:
```suggestion
    List<List<String>> ngrams = new ArrayList<>();

    for (List<String> sentence : sentences) {
        int sentenceSize = sentence.size();
        for (int start = 0; start < sentenceSize; start++) {
            for (int size = minNGrams; size <= Math.min(maxNGrams, sentenceSize); size++) {
                if (start + size <= sentenceSize) {
                    ngrams.add(sentence.subList(start, start + size));
                }
            }
        }
    }
    return ngrams;
```

warehouse/keyword-common/src/main/java/datawave/util/keyword/YakeKeywordExtractor.java
```diff
@@ -0,0 +1,718 @@
...
+    /** Build a co-occurrence matrix of words based on a set of ngrams. Words are down-cased prior to adding them
+     *  to the matrix. We track the max sentence of co-occurrence instances for each pair of words.
+     *
+     * @param ngrams the set of ngrams to analyze
+     * @param left if true, the matrix is based on preceding words (e.g., those that appear to the left of the target word.)
+     *             if false, the matrix is based on following words (e.g., those that appear to the right of the target word.)
+     * @return a map where the key is a word and the value is the list of words that co-occur with that word. For each
+     *   co-occurrence, we track the max sentence id where that co-occurrence happened.
+     */
+    protected Map<String, Map<String, Integer>> buildCoOccurrence(List<List<String>> ngrams, boolean left) {
```
FineAndDandy: is the intent to compare only to the first word/last word of each sentence? or pairs of adjacent words? This will only compare each word in the sentence to the leading or trailing word of each sentence

warehouse/keyword-common/src/main/java/datawave/util/keyword/language/YakeLanguage.java
drewfarris: Not really, I'll document the expected usage here in javadoc that should explain why we might not need to be as concerned about thread safety here.

warehouse/keyword-common/src/main/java/datawave/util/keyword/YakeKeywordExtractor.java
```diff
@@ -0,0 +1,718 @@
...
+     *  specified ngram sizes. Ngrams are generated using a sliding window over the input text. We will always return
+     *  at least one ngram per sentence.
+     *
+     * @param sentences a list of sentences each split into words (should be down-cased).
+     * @param minNGrams the minimum ngram size.
+     * @param maxNGrams the maximum ngram size.
+     * @return a list of ngrams generated from the input sentences ordered based on where they appeared in the input
+     *    sentences. These ngrams are not deduplicated and the words in these ngrams are not normalized in any way.
+     */
+    protected static List<List<String>> calculateNGrams(List<List<String>> sentences, int minNGrams, int maxNGrams) {
```
FineAndDandy, line 348: This also generates every ngram between min and max, rather than just the largest ngram possible for each start word.

warehouse/keyword-common/src/main/java/datawave/util/keyword/YakeKeywordExtractor.java
FineAndDandy: I realize now what is going in here is not sentences, but the ngrams, but when we are dealing with an ngram of length N I think this will only compare the first and Nth ngram with the other ngrams

FineAndDandy: do we want to do proper normalization here or is lowercase sufficient?

FineAndDandy: we spoke about punctuation specifically

warehouse/keyword-common/src/main/java/datawave/util/keyword/TaggedToken.java
FineAndDandy, line 1: very much appreciate these and the other similar classes for readability

warehouse/query-core/src/main/java/datawave/query/config/KeywordQueryConfiguration.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    public synchronized void setRanges(final Collection<Range> ranges) {
+        // As a single atomic operation, clear the range and add all of the
+        // specified ranges
+        this.ranges.clear();
+        if (null != ranges) {
+            this.ranges.addAll(ranges);
+        }
+    }
```
FineAndDandy: I don't see any threaded access to this, probably doesn't need to be synchronized?

warehouse/query-core/src/main/java/datawave/query/transformer/KeywordQueryTransformer.java
```diff
@@ -0,0 +1,123 @@
...
+        // capture the metadata that identifies the field.
+        Metadata m = new Metadata();
+        m.setRow(kkv.getShardId());
+        m.setDataType(kkv.getDatatype());
+        m.setInternalId(kkv.getUid());
+        e.setMetadata(m);
+
+        // store the content in a field based on its view name.
+        field.setMarkings(kkv.getMarkings());
+        field.setName(kkv.getViewName() + "_KEYWORDS");
```
FineAndDandy, line 139: consider making this configurable

warehouse/query-core/src/main/java/datawave/query/iterator/logic/KeywordExtractingIterator.java
```diff
@@ -0,0 +1,388 @@
...
+    public boolean hasTop() {
+        return tk != null;
+    }
+
+    @Override
+    public SortedKeyValueIterator<Key,Value> deepCopy(IteratorEnvironment env) {
+        KeywordExtractingIterator it = new KeywordExtractingIterator();
+        it.source = source.deepCopy(env);
+        return it;
+    }
```
FineAndDandy: this should copy everything that was set or configured in init() into the new iterator. I think there is more to do here.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/KeywordExtractingIterator.java
```diff
@@ -0,0 +1,388 @@
...
+            for (String name : preferredViews) {
+                if (name.endsWith("*")) {
+                    name = name.substring(0, name.length() - 1);
+                    if (currentViewName.startsWith(name)) {
+                        addFoundContent(foundContent, currentViewName, source);
+                    }
+                } else {
+                    if (currentViewName.equalsIgnoreCase(name)) {
+                        addFoundContent(foundContent, currentViewName, source);
+                    }
```
FineAndDandy: one of these ignores the view case, the other does not

warehouse/query-core/src/main/java/datawave/query/util/keyword/KeywordExtractor.java
```diff
@@ -0,0 +1,121 @@
...
+            maxContentLength = Integer.parseInt(iteratorOptions.get(MAX_CONTENT_CHARS));
+        }
+    }
+
+    public KeywordResults extractKeywords() {
+        KeywordResults results = EMPTY_RESULTS;
+        for (String viewName : preferredViews) {
+            for (Map.Entry<String,byte[]> foundEntry : foundContent.entrySet()) {
+                String[] parts = foundEntry.getKey().split(":");
+                if (parts.length > 0 && viewName.equals(parts[0])) {
```
FineAndDandy, line 100: Please document the expected form of the entries.

nothing is done with the true/false set in the KeywordExtractingIterator. I'm assuming this was to track if the document was already uncomressed and decoded or not? But it also could be used for tracking if the doc was truncated.

warehouse/query-core/src/main/java/datawave/query/util/keyword/KeywordExtractor.java
```diff
@@ -0,0 +1,121 @@
...
+                    final byte[] encodedContent = foundEntry.getValue();
+                    final byte[] decodedContent = ContentKeyValueFactory.decodeAndDecompressContent(encodedContent);
```
FineAndDandy, line 102: I think this is safe to call even if previously decoded/decompressed

warehouse/keyword-common/src/test/java/datawave/util/keyword/SimpleRegressionUtil.java
```diff
@@ -0,0 +1,190 @@
...
+import java.util.List;
+
+import org.apache.commons.io.IOUtils;
+
+//@formatter:off
+/** A set of simple utilities for creating parameterized tests based on a series of input files stored in a directory
+ *  on disk. {@link #getTestsForClassData(Class)} uses the test class name to find input and expected output files
+ *  in a directory under <pre>target/test-classes</pre> based on the class' package and class names. Additionally
+ *  the methods {@link #getInputFileForTest(String, Class)}, {@link #getExpectedFileForTest(String, Class)}, and
+ *  {@link #getOutputFileForTest(String, Class)} can be used to get File objects used for inidividual test outputs.
```
FineAndDandy:
```suggestion
 *  {@link #getOutputFileForTest(String, Class)} can be used to get File objects used for individual test outputs.
```

warehouse/keyword-common/src/test/java/datawave/util/keyword/SimpleRegressionUtil.java
```diff
@@ -0,0 +1,190 @@
...
+ *  <p/>
+ *  A skeleton for a test that would used this class could look like:
```
FineAndDandy:
```suggestion
 *  A skeleton for a test that would use this class could look like:
```

warehouse/keyword-common/src/test/java/datawave/util/keyword/SimpleRegressionUtil.java
```diff
@@ -0,0 +1,190 @@
...
+ *     }
+ *
+ *     public static List<String> findTests() {
+ *         return getTestsForClassData(YakeKeywordRegressionTest.class);
+ *     }
+ *  }
+ *  </pre>
+ */
+// @formatter:on
+public class SimpleRegressionUtil {
```
FineAndDandy, line 41: I like this

warehouse/keyword-common/src/test/java/datawave/util/keyword/YakeKeywordExtractorTest.java
```diff
@@ -0,0 +1,118 @@
...
+    static {
+        LinkedHashMap<String,Double> m = EXPECTED_NEWS_OUTPUT;
+        m.put("acquiring kaggle", 0.4602);
+        m.put("cloud next", 0.3884);
+        m.put("cloud next conference", 0.5552);
+        m.put("san francisco", 0.3884);
+    }
+
+    @Test
+    public void testSteps() {
```
FineAndDandy, line 44: this was helpful in validating understanding of the code

warehouse/query-core/src/main/java/datawave/query/tables/keyword/KeywordQueryLogic.java
```diff
@@ -0,0 +1,413 @@
...
+        config.setState(new KeywordQueryState());
+
+        // Get the BYPASS_ACCUMULO parameter if given
+        String bypassAccumuloString = settings.findParameter(BYPASS_ACCUMULO).getParameterValue().trim();
+        if (StringUtils.isNotBlank(bypassAccumuloString)) {
+            boolean bypassAccumuloBool = Boolean.parseBoolean(bypassAccumuloString);
+            config.setBypassAccumulo(bypassAccumuloBool);
+        }
+
+        this.scannerFactory = new ScannerFactory(config);
```
FineAndDandy, line 140: I led you wrong here. There is actually more work to be done to support this. The only code that currently supports the bypass flag is used via the PushdownScheduler: https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/query-core/src/main/java/datawave/query/scheduler/PushdownScheduler.java#L166

They key is instead of using a Scanner you need to use a ScannerSession. This allows you to set the setDelegatedInitializer() to RFileResource.class when bypassAccumulo is configured. Since you already have a ScannerFactory, you can get access to a ScannerSession as-is would require a fair bit more code.

warehouse/query-core/src/main/java/datawave/query/tables/keyword/KeywordQueryLogic.java
```diff
@@ -0,0 +1,413 @@
...
+            KeywordExtractingIterator.setOptions(cfg, config.getMinNgrams(), config.getMaxNgrams(), config.getMaxKeywords(), config.getMaxScore(),
+                            config.getMaxContentChars(), config.getState().getViewNames(), config.getState().getDocumentLanguageMap(), false);
+            scanner.addScanIterator(cfg);
+
+            this.iterator = scanner.iterator();
+            this.scanner = scanner;
+
+        } catch (TableNotFoundException e) {
+            throw new RuntimeException("Table not found: " + this.getTableName(), e);
+        }
```
FineAndDandy, line 198:
```suggestion
        final ScannerSession scannerSession = this.scannerFactory.newLimitedScanner(ScannerSession.class, config.getTableName(), config.getAuthorizations(),
                        config.getQuery());
        scannerSession.setRanges(config.getState().getRanges());

        if (config.getBypassAccumulo()) {
            scannerSession.setDelegatedInitializer(RfileResource.class);
        }

        SessionOptions options = scannerSession.getOptions();

        final IteratorSetting cfg = new IteratorSetting(60, "keyword-extractor", KeywordExtractingIterator.class);
        KeywordExtractingIterator.setOptions(cfg, config.getMinNgrams(), config.getMaxNgrams(), config.getMaxKeywords(), config.getMaxScore(),
                        config.getMaxContentChars(), config.getState().getViewNames(), config.getState().getDocumentLanguageMap(), false);
        options.addScanIterator(cfg);

        this.iterator = new Iterator<>() {
            @Override
            public boolean hasNext() {
                return scannerSession != null && scannerSession.hasNext();
            }

            @Override
            public Entry<Key,Value> next() {
                Result result = scannerSession.next();
                return new AbstractMap.SimpleEntry<>(result.getKey(), result.getValue());
            }
        };
```

FineAndDandy: this should allow you to use the bypass accumulo option. There will be test updates and a few other changes required, but this gives the jist.

warehouse/query-core/src/main/java/datawave/query/transformer/KeywordQueryTransformer.java
```diff
@@ -0,0 +1,191 @@
...
+     * </p>
+     *
+     * @param querySettings
+     *            the current query for which we are transforming results.
+     * @param identifierMap
+     *            used to store mappings between document UIDs and identifiers
+     * @param languageMap
+     *            used to store mappings between document UIDs and languages
+     */
+    public static void extractIdentifiersAndLanguages(Query querySettings, Map<Metadata,String> identifierMap, Map<Metadata,String> languageMap) {
```
FineAndDandy, line 70: this needs test coverage

warehouse/query-core/src/main/java/datawave/query/util/keyword/KeywordResults.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    public LinkedHashMap<String,Double> get() {
+        return results;
+    }
+
+    @Override
+    public void readFields(DataInput dataInput) throws IOException {
+        int sz = dataInput.readInt();
+        this.source = dataInput.readUTF();
```
FineAndDandy, line 47: if source is an empty string reset to null

warehouse/keyword-common/src/main/java/datawave/util/keyword/YakeKeywordExtractor.java
drewfarris: Yes, too complicated indeed. This is even simpler:
```java
        final List<TokenValue> resultsFlattenedIndexed = new ArrayList<>();
        for (int i=0; i < tokenizedSentences.size(); i++) {
            for (String token: tokenizedSentences.get(i)) {
                resultsFlattenedIndexed.add(new TokenValue(token, i));
            }
        }
        return resultsFlattenedIndexed;
```

warehouse/keyword-common/src/main/java/datawave/util/keyword/YakeKeywordExtractor.java
drewfarris: Correct, that is the intent.

warehouse/query-core/src/main/java/datawave/query/config/KeywordQueryConfiguration.java
drewfarris: Moved these into the KeywordQueryState class.

warehouse/query-core/src/main/java/datawave/query/tables/keyword/KeywordQueryLogic.java
drewfarris: Thanks this is helpful - I will tackle this in the next PR that adds the chained queries.

warehouse/query-core/src/main/java/datawave/query/transformer/KeywordQueryTransformer.java
drewfarris: Will look into this in the next round of changes

warehouse/query-core/src/main/java/datawave/query/util/keyword/KeywordExtractor.java
drewfarris: Will resolve this in the next round of changes

warehouse/query-core/src/main/java/datawave/query/util/keyword/KeywordExtractor.java
drewfarris: Will resolve this in the next round of changes

FineAndDandy (approved):

warehouse/query-core/src/test/java/datawave/query/tables/keyword/KeywordQueryTransformerTest.java
```diff
@@ -0,0 +1,110 @@
...
+    String[] input = {"DOCUMENT:shard1/datatype/uid", "DOCUMENT:shard2/datatype/uid!PAGE_ID:12345",
+            "DOCUMENT:shard3/datatype/uid!PAGE_ID:12345%LANGUAGE:ENGLISH", "DOCUMENT:shard4/datatype/uid%LANGUAGE:ENGLISH",
+
+    };
+
+    String[][] expected = {{"DOCUMENT", "shard1", "datatype", "uid", "", ""}, {"DOCUMENT", "shard2", "datatype", "uid", "PAGE_ID:12345", ""},
+            {"DOCUMENT", "shard3", "datatype", "uid", "PAGE_ID:12345", "ENGLISH"}, {"DOCUMENT", "shard4", "datatype", "uid", "", "ENGLISH"},};
+
+    String[] malformedInput = {"DOCUMENT:shard1/datatype", "shard2/datatype", "DOCUMENT:shard3/data!type/uid!PAGE_ID:12345%LANGUAGE:ENGLISH",
+            "shard5/datatype/uid!PAGE_ID:54321", "shard6/datatype/uid%LANGUAGE:ENGLISH", "shard7/datatype/uid!PAGE_ID:54321%LANGUAGE:ENGLISH"};
```
apmoriarty, line 29: Would be nice to see these on separate lines via disabling the formatter

drewfarris: This is fixed in the feature/keyword-extraction-chain branch. Thanks.

FineAndDandy (dismissed):

warehouse/keyword-common/src/main/java/datawave/util/keyword/YakeKeywordExtractor.java
```diff
@@ -0,0 +1,751 @@
...
+                final List<String> tagList = y.stream().map(TaggedToken::getTag).collect(Collectors.toList());
+                return !(tagList.contains("u") || tagList.contains("d"));
+            }).filter(y -> {
+                // remove any candidates where the first or last word is a stopwords
+                final String firstWord = y.get(0).getLowercaseToken();
+                final String lastWord = y.get(y.size() - 1).getLowercaseToken();
+                return !(stopwords.contains(firstWord) || stopwords.contains(lastWord));
+            })
+                            // transform remaining candidates into a comma-delimited string
+                            .map(y -> y.stream().map(TaggedToken::getToken).map(String::toLowerCase).collect(Collectors.joining(",")));
```
apmoriarty, line 539: weird formatting to fix later (or now)

apmoriarty (dismissed):

drewfarris:
Squashed and rebased - resolves pom version issue that was blocking merge.

apmoriarty (approved):

FineAndDandy (approved):

PR #2826, opened 2025-04-01
Automate Whindex Creation Date Marker Generation and Rule Parsing

### LINKED ISSUE #2671: Whindex Marker Deployment
The whindex concept was built a composite index based on the a subset of values in one field in a document.  The whindex start date marker currently needs to be created manually in the metadata table.  We would like this start date to be determined more automatically based on what is ingested.  This could be by having ingest automatically add the markers, or perhaps the dated "i" entries and "f" entries can be used in the same way the federated query planner determines index holes.

SethSmucker:
Whindex Creation Date Markers (WCD's) have been manually generated in the past. This PR adds automatic WCD creation via the `WhindexIngest` interface and `WhindexNormalizer` class. 

`WhindexFieldIngestHelper` has been added to make interacting with `WhindexIngest` easier, as well as align with the implementations of `VirtualIngest` and `CompositeIngest`. 
`MetadataWithEarliestDate` has been added to help with identifying and adding IndexedFields with the earliest date to the `EventMetadata`.

`WhindexFieldIngestHelperTest` and `WhindexCreationDateFilterTest` test classes have also been included.

The following classes have been updated to accommodate the new Whindex implementation:
- `BaseIngestHelper`
- `IngestHelperInterface`
- `EventMapper`
- `EventMetadata`
- `MetadataTableConfigHelper`

fixes #2671


warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexIngest.java
```diff
@@ -0,0 +1,212 @@
...
+import com.google.common.collect.LinkedListMultimap;
+import com.google.common.collect.Multimap;
+
+import datawave.ingest.data.Type;
+import datawave.ingest.data.config.NormalizedContentInterface;
+import datawave.ingest.data.config.NormalizedFieldAndValue;
+import datawave.marking.MarkingFunctions;
+import datawave.marking.MarkingFunctionsFactory;
+
+public interface WhindexIngest {
```
ivakegg, line 12: Can we include an overall comment which describes what a whindex is, and includes the overall configuration syntax please

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexIngest.java
```diff
@@ -0,0 +1,212 @@
...
+        /**
+         * @param field
+         *            the field to check.
+         * @return {@code true} if {@code field} is a whindex field.
+         */
+        public boolean isWhindexField(String field) {
+            return whindexFieldDefinitions.containsKey(field);
+        }
+
+        // todo: discuss what this actually needs to do
```
ivakegg: So this method is not actually used and the same goes for the CompositeIngest normalizer.  This method is a hold over from being a "Normalizer" implementation at one point.  You can remove this method.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexIngest.java
```diff
@@ -0,0 +1,212 @@
...
+        public Multimap<String,NormalizedContentInterface> normalize(Multimap<String,String> fields) {
+            Multimap<String,NormalizedContentInterface> eventFields = HashMultimap.create();
+            for (Map.Entry<String,String> field : fields.entries()) {
+                eventFields.put(field.getKey(), new NormalizedFieldAndValue(field.getKey(), field.getValue()));
+            }
+            return normalizeMap(eventFields);
+        }
+
+        // todo: discuss what this actually needs to do
```
ivakegg, line 175: What is there to discuss?

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexIngest.java
ivakegg: See pinned comment in our chat:

So the configuration needs to be able to encode the following:
1) The field from which the values will be pulled.
2) A mapping of fieldname + value -> whindex fieldname
The separator configuration is nice but perhaps not really needed.  The current configuration is
RULES := RULE;RULE
RULE := FIELDS::VALUE{,VALUE}
FIELDS := FIELD{,FIELD}
FIELD := SRC_FIELD:DST_FIELD[:DELETE_SRC]
DELETE_SRC := 'true' if the SCR_FIELD should be removed, 'false' otherwise - defaults to 'false'
SRC_FIELD: the field which should be renamed/remapped
DST_FIELD := the new name for the field (a.k.a the whindex field)
VALUE := The value pulled from the field defined in #1 above.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexIngest.java
```diff
@@ -0,0 +1,212 @@
...
+         *            the config instance that holds the rules.
+         */
+        public void setup(Type type, Configuration config) {
+            markingFunctions = MarkingFunctionsFactory.createMarkingFunctions();
+
+            String whindexRules = config.get(type.typeName() + "." + RULE, "");
+            if (whindexRules.isEmpty()) {
+                return;
+            }
+
```
ivakegg: We are missing configuration (1) from the description above; namely the fields from which those values apply.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexIngest.java
ivakegg: Lets call the field defined in 1. above to be the VALUE_FIELD

ivakegg: So this ingester needs to be able to create the following fields:

DST_FIELD = <value of SRC_FIELD> IFF VALUE_FIELD contains one of the VALUEs

And the ingester needs to remove the following fields:

SRC_FIELD IFF DELETE_SRC is true and its value is being used in the DST_FIELD created above.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexIngest.java
ivakegg: We need to be able to create multiple whindex configuration

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexIngest.java
ivakegg: This method is intended to take the input map of fields and values, and is to apply the whindex configuration.  This means creating the whindex fields per the configuration, dropping the source fields (of overloaded), etc.  This is intended to be business logic behind the creating of whindex fields.

See the comment above as to the fields that this method needs to create and fields this method needs to remove.  The map returned is a modification of the eventFields map passed in.

ivakegg (changes requested):
see comments

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexIngest.java
lbschanno: How should we expect the different whindex configurations to be distinguished? Will they have an instance value in their configuration property, or will we have multiple configurations mapped to the same `<type>.rules` property?

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexIngest.java
lbschanno: Just to confirm, is VALUE_FIELD equivalent to SRC_FIELD within the rule configuration syntax?

lbschanno: If VALUE_FIELD is not equivalent to SRC_FIELD, then I assume we need another configuration property such as `<type>.whindex.valueField` to allow for this to be configured.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexIngest.java
lbschanno: For example, do we need to support something akin to 
```
wiki.rules.1=`...`
wiki.rules.2=`...`
```
or something different?

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexIngest.java
lbschanno: VALUE_FIELD clarified in separate chat.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexConfig.java
```diff
@@ -0,0 +1,70 @@
+package datawave.ingest.data.config.ingest;
+
+import java.util.List;
+import java.util.Objects;
+
+import datawave.ingest.data.config.ingest.WhindexIngest;
+
```
lbschanno, line 7: Would like to see a class-level comment here describing the purpose of this class.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelper.java
```diff
@@ -0,0 +1,138 @@
...
+import org.apache.log4j.Logger;
+
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.LinkedListMultimap;
+import com.google.common.collect.Multimap;
+
+import datawave.ingest.data.Type;
+import datawave.ingest.data.config.NormalizedContentInterface;
+import datawave.ingest.data.config.NormalizedFieldAndValue;
+
```
lbschanno, line 22: Need a class-level comment here as well describing the overall purpose of this class.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelper.java
```diff
@@ -0,0 +1,138 @@
...
+    private final Type type;
+    private final Multimap<String,WhindexConfig> valueFieldsToWhindexConfigs = HashMultimap.create();
+    private final Set<String> overloadedFields = new HashSet<>();
+
+    public WhindexFieldIngestHelper(Type type) {
+        this.type = type;
+    }
+
+    @Override
```
lbschanno, line 91: Please add a method level comment here describing what properties will be examined for whindex configuration values.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelper.java
```diff
@@ -0,0 +1,138 @@
...
+                case VALUES:
+                    whindexConfig.setValues(List.of(entry.getValue().split(",")));
+                    break;
+                default:
+                    String originalProperty = commonPrefix + groupID + "." + property;
+                    log.warn("Unexpected whindex property given:" + originalProperty + "=" + entry.getValue());
+            }
+        }
+
+        // add to fieldDefinitions and VFToWC
```
lbschanno: This comment is not very clear, there is no variable called fieldDefinitions, and you should avoid abbreviating variable like you did for VFToWC in documentation. Please reference by the full variable name, or an understandable concept like "map of value fields to whindex configs".

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelper.java
```diff
@@ -0,0 +1,138 @@
...
+        groupingsToConfigs.values().forEach((wc) -> {
+            valueFieldsToWhindexConfigs.put(wc.getValueField(), wc);
+        });
+
+    }
+
+    @Override
```
lbschanno: Please add a method-level comment here describing what this method returns. The documentation should provide a summary of the overall workflow for how we determine if/when a whindex entry will be included in the results.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelper.java
```diff
@@ -0,0 +1,138 @@
...
+    public Multimap<String,NormalizedContentInterface> getWhindexFields(Multimap<String,NormalizedContentInterface> eventMap) {
+
+        Multimap<String,NormalizedContentInterface> whindicesInEventMap = HashMultimap.create();
+
+        // Get all the wc that have both vf and sf in the eventMap
```
lbschanno: Abbreviation of variable names here in comment that should be avoided.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelper.java
```diff
@@ -0,0 +1,138 @@
...
+        List<WhindexConfig> matchingConfigs = valueFieldsToWhindexConfigs.entries().stream()
+                        .filter(entry -> eventMap.containsKey(entry.getValue().getValueField()) && eventMap.containsKey(entry.getValue().getSourceField()))
+                        .map(Map.Entry::getValue).collect(Collectors.toList());
+
+        // Check that the eventMap entry has EITHER an EventField or IndexedField in common with the wc's VALUES
```
lbschanno: abbreviation of variable name here in comment that should be avoided.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelper.java
```diff
@@ -0,0 +1,138 @@
...
+        for (WhindexConfig curWhindexConfig : matchingConfigs) {
+            Collection<NormalizedContentInterface> relatedValueEventContents = eventMap.get(curWhindexConfig.getValueField());
+            // if any of the NCI's have either EF or IF that's GOOD!
```
lbschanno: Abbreviation of variable name/class name in comment that should be avoided.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelper.java
```diff
@@ -0,0 +1,138 @@
...
+        groupingsToConfigs.values().forEach((wc) -> {
+            valueFieldsToWhindexConfigs.put(wc.getValueField(), wc);
+        });
+
+    }
+
+    @Override
+    public Multimap<String,NormalizedContentInterface> getWhindexFields(Multimap<String,NormalizedContentInterface> eventMap) {
+
+        Multimap<String,NormalizedContentInterface> whindicesInEventMap = HashMultimap.create();
```
lbschanno: I would rename this to `whindexEntries`, or `newEntries` to make the overall purpose a bit clearer.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelper.java
```diff
@@ -0,0 +1,138 @@
...
+        // Get all the wc that have both vf and sf in the eventMap
+        List<WhindexConfig> matchingConfigs = valueFieldsToWhindexConfigs.entries().stream()
+                        .filter(entry -> eventMap.containsKey(entry.getValue().getValueField()) && eventMap.containsKey(entry.getValue().getSourceField()))
+                        .map(Map.Entry::getValue).collect(Collectors.toList());
+
+        // Check that the eventMap entry has EITHER an EventField or IndexedField in common with the wc's VALUES
+        for (WhindexConfig curWhindexConfig : matchingConfigs) {
+            Collection<NormalizedContentInterface> relatedValueEventContents = eventMap.get(curWhindexConfig.getValueField());
+            // if any of the NCI's have either EF or IF that's GOOD!
+            boolean containsAnyMatchingValue = relatedValueEventContents.stream()
```
lbschanno: You can create a local variable that contains the result of `curWhindexConfig.getValues()`, and then reference that when checking if the set contains either the event field value or index field value. This will avoid multiple calls to `curWhindexConfig.getValues().`

Additionally, please use `curr` instead of `cur` when using as an abbreviation for current.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelper.java
```diff
@@ -0,0 +1,138 @@
...
+                            .anyMatch(nci -> curWhindexConfig.getValues().contains(nci.getEventFieldValue())
+                                            || curWhindexConfig.getValues().contains(nci.getIndexedFieldValue()));
+
+            if (containsAnyMatchingValue) {
```
lbschanno: Add a comment here describing why we're doing this step.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelper.java
```diff
@@ -0,0 +1,138 @@
...
+
+    }
+
+    @Override
+    public Multimap<String,WhindexConfig> getValueFieldsToWhindexConfigs() {
+        return valueFieldsToWhindexConfigs;
+    }
+
+    @Override
+    public boolean isWhindexField(String field) {
```
lbschanno, line 213: I think it would be good for us to have a separate class level `Set<String>` variable that contains the dest fields that we populate during a call to `setup(Configuration config)`. It would make the implementation of this method simpler. 

Additionally, we should make this new set, the overloadedFields set, and the valueFieldsToWhindexConfigs map immutable/unmodifiable so that they cannot be modified after they've been retrieved via their getter methods. We should only support modification of them through calls to setup(), or through explicit setter methods.

warehouse/ingest-core/src/main/java/datawave/ingest/metadata/MetadataWithEarliestDate.java
```diff
@@ -0,0 +1,92 @@
+package datawave.ingest.metadata;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.apache.hadoop.io.Text;
+
+import com.google.common.base.Objects;
+
```
lbschanno, line 10: Please add a class-level comment describing the purpose of the class, as well as some more method-level documentation throughout the class.

warehouse/ingest-core/src/test/java/datawave/ingest/data/config/ingest/WhindexConfigTest.java
```diff
@@ -0,0 +1,70 @@
+package datawave.ingest.data.config.ingest;
+
+import static org.junit.jupiter.api.Assertions.*;
```
lbschanno: Starred imports should not be used.

warehouse/ingest-core/src/test/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelperTest.java
```diff
@@ -0,0 +1,259 @@
...
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.LinkedListMultimap;
+import com.google.common.collect.Multimap;
+
+import datawave.ingest.data.Type;
+
+class WhindexFieldIngestHelperTest {
+
+    /*
+     * Tests that a configuration without a <type>.rules property does not have whindex field definitions.
```
lbschanno: The documentation is out of date. Whindex field configurations are now supplied via properties with the  `<type>.whindex.rules` prefix.

warehouse/ingest-core/src/test/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelperTest.java
```diff
@@ -0,0 +1,259 @@
...
+        HashMultimap<String,WhindexConfig> expectedValues = HashMultimap.create();
+        expectedValues.put("AppleVF", expectedConfigA);
+        expectedValues.put("aeiouVF", expectedConfigB);
+        expectedValues.put("hahaVF", expectedConfigC);
+
+        Assertions.assertEquals(expectedValues, wHelper.getValueFieldsToWhindexConfigs());
+    }
+
+     @Test
+     void testGetWhindexFields() {
```
lbschanno: This test contains assertions for three different scenarios. I recommend breaking up this test into three separate test cases, one for each scenario. It makes it clearer and easier to document what scenarios are being tested. Please do the same for the other tests in this class that do a similar thing.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelper.java
```diff
@@ -0,0 +1,247 @@
...
+ * Implements methods provided by the WhindexIngest interface.
+ * <p>
+ * The WhindexFieldIngestHelper is responsible for parsing whindex rules from a Hadoop Configuration and generating a mapping between event map fields and
+ * corresponding whindex configurations. The configuration rules determine how and when new whindex fields are created from an input event map.
+ * <ul>
+ * <li><strong>valueFieldsToWhindexConfigs</strong>: Associates an event map's value field with a corresponding WhindexConfig that specifies the source field,
+ * destination field, values, and whether the source field is overloaded (i.e. to be removed).</li>
+ * <li><strong>overloadedFields</strong>: A set of source fields that are marked as overloaded (i.e., should be deleted) based on the configuration.</li>
+ * <li><strong>destinationFields</strong>: A set of fields that are defined as new whindex (destination) fields. (Note: In this implementation, destination
+ * fields are added during setup but the getter for destination fields currently returns a copy of the overloadedFields. Verify if this behavior meets your
```
ivakegg: I really do not get this comment.  Overloaded fields would be the source fields that are dropped in lieu of the destination/whindex fields.  How can overloaded fields be the same as the destination fields?  I think we have another misunderstanding here unfortunately.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexConfig.java
```diff
@@ -0,0 +1,167 @@
...
+ * Container class for keeping track of configuration rules related to Whindex entries.
+ * <p>
+ * A whindex rule configuration is defined by several parameters:
+ * <ul>
+ * <li><strong>valueField</strong> - The field in which an event map's values are checked. The event map must contain a matching entry for this field in order
+ * to create a Whindex entry.</li>
+ * <li><strong>values</strong> - A list of values that need to be present in the event map's valueField to create a whindex entry.</li>
+ * <li><strong>sourceField</strong> - A field that must be present in the event map for a Whindex entry to be created. Depending on configuration, this field
+ * may be replaced with the destination field.</li>
+ * <li><strong>destField</strong> - Also known as the "WhindexField". This is the name of the new field created when a whindex entry is generated.</li>
```
ivakegg: Please add that the destField value will be that of the sourceField when generated.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelper.java
```diff
@@ -0,0 +1,247 @@
...
+                    destinationFields.add(entry.getValue());
+                    break;
+                case VALUES:
+                    // Split the comma-separated list of values and set them.
+                    whindexConfig.setValues(List.of(entry.getValue().split(",")));
+                    break;
+                default:
+                    // Log a warning for any unexpected property found in the configuration.
+                    String originalProperty = commonPrefix + groupID + "." + property;
+                    log.warn("Unexpected whindex property given:" + originalProperty + "=" + entry.getValue());
```
ivakegg: I think this should be an error, and throw an exception.  If a property was specified that is unexpected, then it was an error/typo that needs to be fixed.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelper.java
```diff
@@ -0,0 +1,247 @@
...
+
+        // Get all properties from the configuration that start with the common prefix.
+        Map<String,String> properties = config.getPropsWithPrefix(commonPrefix);
+        // Map to temporarily group properties by their group ID.
+        Map<String,WhindexConfig> groupingsToConfigs = new HashMap<>();
+
+        // Process each configuration property.
+        for (Map.Entry<String,String> entry : properties.entrySet()) {
+            // Split the key by '.' to extract the group ID and property name.
+            String[] parts = entry.getKey().split("\\.");
```
ivakegg: A little validation here.... only 2 parts and neither part is empty/0 length.  Also ensure value is non-empty.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelper.java
```diff
@@ -0,0 +1,247 @@
...
+        List<WhindexConfig> matchingConfigs = valueFieldsToWhindexConfigs.entries().stream()
+                        .filter(entry -> eventMap.containsKey(entry.getValue().getValueField()) && eventMap.containsKey(entry.getValue().getSourceField()))
+                        .map(Map.Entry::getValue).collect(Collectors.toList());
+
+        // Iterate over the matching configurations to generate the whindex fields.
+        for (WhindexConfig currWhindexConfig : matchingConfigs) {
+            // Retrieve all normalized content corresponding to the value field.
+            Collection<NormalizedContentInterface> relatedValueEventContents = eventMap.get(currWhindexConfig.getValueField());
+            // Check if any of these contents match the allowed values (either event field or indexed field).
+            boolean containsAnyMatchingValue = relatedValueEventContents.stream().anyMatch(nci -> {
```
ivakegg, line 172: This could be very inefficient, lets make one pass through the values as opposed to two when creating the whindex entries.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/EventMapper.java
```diff
@@ -792,21 +793,36 @@ public Multimap<String,NormalizedContentInterface> getFields(RawRecordContainer
...
                 newFields.putAll(fieldName, compositeFields.get(fieldName));
             }
         }
 
+        // Also get whindex fields, if applicable
+        if (handler.getHelper(value.getDataType()) instanceof WhindexIngest) {
+            WhindexIngest vHelper = (WhindexIngest) handler.getHelper(value.getDataType());
+            Multimap<String,NormalizedContentInterface> whindexFields = vHelper.getWhindexFields(newFields);
+            for (String fieldName : whindexFields.keySet()) {
+                if (vHelper.isOverloadedWhindexField(fieldName)) {
```
ivakegg: I think we need to rework how overloaded is implemented here.  We should only be droppping source fields IFF they were turned into whindex/destination fields.
Perhaps the getWhindexFields should instead replace the "newFields" map and implement the "overloaded" concept there-in.  So this block would simply become something like:
newFields = vHelper.processWhindexFields(newFields)

ivakegg: The problem here is that you are copying the composite field concept of overloaded which is apparently not the same.  Now that I think about it the composite field overloaded implementation may be incorrect but that will be a separate ticket.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelper.java
```diff
@@ -0,0 +1,242 @@
...
+     */
+    public WhindexFieldIngestHelper(Type type) {
+        this.type = type;
+    }
+
+    /**
+     * Parses the whindex rules from the provided Hadoop Configuration.
+     * <p>
+     * The configuration is expected to have properties in the following form: <code>typeName.whindex.rules.[groupID].[property]=value</code>, where the
+     * property is one of VALUE_FIELD, SRC_FIELD, DELETE_SRC_FIELD, DST_FIELD, or VALUES. Each groupID represents a separate whindex rule.
```
lbschanno: I would use {@value #VALUE_FIELD} to inject the actual property name into the javadoc. Same for SRC_FIELD, DELETE_SRC_FIELD, DST_FIELD, and VALUES.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelper.java
```diff
@@ -0,0 +1,242 @@
...
+
+                // Removes the entry from the eventMap IFF a whindex field is created and the srcField is overloaded.
+                if (currWhindexConfig.isOverloaded()) {
+                    eventMap.removeAll(currWhindexConfig.getSourceField());
+                }
+            }
+        }
+
+        // Add the generated whindex fields to the eventMap
+        eventMap.putAll(whindicesInEventMap);
```
lbschanno: I would make a copy of eventMap that then has the whindex entries added to it, and return the copy. This avoids modifying the input map.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelper.java
```diff
@@ -0,0 +1,242 @@
...
+ * fields are added during setup. Verify if this behavior meets your requirements.)</li>
+ * </ul>
+ * <p>
+ * After setup, getWhindexFields() processes an event map to produce new whindex fields based on the defined rules.
+ * </p>
+ */
+public class WhindexFieldIngestHelper implements WhindexIngest {
+
+    // Logger for logging warnings and debug information.
+    private static final Logger log = Logger.getLogger(WhindexFieldIngestHelper.class);
```
lbschanno, line 41: This logger is never used. I would add some debug logging for at least the setup() method to log the final whindex configurations that the ingest helper was set up with.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexConfig.java
```diff
@@ -0,0 +1,168 @@
+package datawave.ingest.data.config.ingest;
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Container class for keeping track of configuration rules related to Whindex entries.
```
lbschanno, line 9: A class-level comment should not contain this much detail about the internal variables of the class unless there is a good reason for doing so, such as explaining a complex process that is handled by the class. We already have documentation about the purpose of the internal members through the comments directly above them. A class-level summary should focus more on being a summary of the purpose of the class. It is often fine for this to be a single sentence for a simple class like this.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexConfig.java
```diff
@@ -0,0 +1,168 @@
...
+ * <li><strong>destField</strong> - Also known as the "WhindexField". This is the name of the new field created when a whindex entry is generated. The value of
+ * this new field will be that of the sourceField when generated.</li>
+ * <li><strong>overloaded</strong> - A boolean flag indicating whether the original source field should be deleted (true) or retained (false) once the whindex
+ * entry has been processed. Inclusion of this is optional, the default value is assumed false.</li>
+ * </ul>
+ * </p>
+ */
+public class WhindexConfig {
+
+    // The field in which the event map's values are checked.
```
lbschanno: The comments in this class should not reference an 'event map'. There is no actual event map in this class, only in WhindexFieldIngestHelper. This can lead to confusion about why an event map is mentioned if someone is only looking at the documentation of this class. I would change the comments here (and elsewhere in the class) to focus more on the purpose of each member. For instance,
```java
    /**
     * The field name parsed from an instance of the property {@link WhindexFieldIngestHelper#VALUE_FIELD}.
     */
    private String valueField;

    /**
     * The field values of {@link #valueField} parsed from an instance of the property {@link WhindexFieldIngestHelper#VALUES}.  
     */
    private List<String> values;
```

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexConfig.java
```diff
@@ -0,0 +1,168 @@
...
+ * <li><strong>sourceField</strong> - A field that must be present in the event map for a Whindex entry to be created. Depending on configuration, this field
+ * may be replaced with the destination field.</li>
+ * <li><strong>destField</strong> - Also known as the "WhindexField". This is the name of the new field created when a whindex entry is generated. The value of
+ * this new field will be that of the sourceField when generated.</li>
+ * <li><strong>overloaded</strong> - A boolean flag indicating whether the original source field should be deleted (true) or retained (false) once the whindex
+ * entry has been processed. Inclusion of this is optional, the default value is assumed false.</li>
+ * </ul>
+ * </p>
+ */
+public class WhindexConfig {
```
lbschanno, line 11: I think we should make this class and its internal properties immutable.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelper.java
```diff
@@ -0,0 +1,242 @@
...
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.ImmutableMultimap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Multimap;
+
+import datawave.ingest.data.Type;
+import datawave.ingest.data.config.NormalizedContentInterface;
+import datawave.ingest.data.config.NormalizedFieldAndValue;
+
+/**
```
lbschanno, line 23: Similar comment here about simplifying the class-level documentation. The internal properties shouldn't necessarily be described here since they already have documentation at their definition. You should include some documentation about how setup() must be called first to initialize the whindex configurations before processWhindexFields() can be used to generate the whindex entries.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelper.java
```diff
@@ -0,0 +1,242 @@
...
+                for (NormalizedContentInterface content : relatedSourceEventContents) {
+                    NormalizedFieldAndValue copy = new NormalizedFieldAndValue(content);
+                    copies.add(copy);
+                }
+                // Map the copies to the destination field (whindex field).
+                whindicesInEventMap.putAll(currWhindexConfig.getDestField(), copies);
+
+                // Removes the entry from the eventMap IFF a whindex field is created and the srcField is overloaded.
+                if (currWhindexConfig.isOverloaded()) {
+                    eventMap.removeAll(currWhindexConfig.getSourceField());
```
lbschanno: You should wait until you've iterated through each whindex config before deleting any overloaded source fields. It is possible for multiple whindex configs to reference the same source field, and deleting the source fields too early can lead to missing whindex entries. Something like the following should work:

```java
Set<String> fieldsToRemove = new HashSet<>();
for(WhindexConfig currWhindexConfig : matchingConfigs) {
    // Create whindex entries. Add overloaded src fields to fieldsToRemove when a corresponding whindex entry is added.
}

for (String field : fieldsToRemove) {
    // Now delete the src fields.
}
```

lbschanno: This would also be a good edge case to add a test for: if multiple configs with the same overloaded src fields are called, do we still get all the expected whindex entries?

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexIngest.java
```diff
@@ -0,0 +1,36 @@
...
+import org.apache.hadoop.conf.Configuration;
+
+import com.google.common.collect.Multimap;
+
+import datawave.ingest.data.config.NormalizedContentInterface;
+
+public interface WhindexIngest {
+
+    /**
+     * Used to allow external scopes to interface with the WhindexIngest's WhindexFieldNormalizer's .setup() method. Initializes the WhindexIngest from a
```
lbschanno: This comment is outdated.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexIngest.java
```diff
@@ -0,0 +1,36 @@
...
+
+    /**
+     * @param field
+     *            the field to check.
+     * @return {@code true} if {@code field} is a whindex field.
+     */
+    boolean isWhindexField(String field);
+
+    /**
+     * // todo Given a "{@code RULE}", return a {@code Multimap<String, NormalizedContentInterface>} of whindex fields ("{@code DST_FIELD}") mapped to the
```
lbschanno: Leftover todo here.

warehouse/ingest-core/src/test/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelperTest.java
```diff
@@ -0,0 +1,364 @@
...
+        WhindexFieldIngestHelper wHelper = new WhindexFieldIngestHelper(type);
+        wHelper.setup(config);
+
+        // Verify that only the destination field is recognized as a whindex field.
+        Assertions.assertTrue(wHelper.isWhindexField("zDestField"));
+        Assertions.assertFalse(wHelper.isWhindexField("zSourceField"));
+        Assertions.assertFalse(wHelper.isWhindexField("zTestValue1"));
+    }
+
+    // /**
```
lbschanno: All of this commented out code should be removed, or uncommented if they are tests that need to be added.

lbschanno (changes requested):
On the right track, but some clean up is needed.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexConfig.java
SethSmucker: Does this mean that we should wait to see if all properties per WhindexConfig are present in the configuration _before_ creating the WhindexConfig? The setters are used during the setup process. Alternatively, we could just re-construct the WhindexConfig with the current values + the new one from the config at each iteration. What do you think?

lbschanno: I think you could instead use a Builder pattern.

.maven-dockerignore
```diff
@@ -0,0 +1,59 @@
+# This .dockerignore gets symlinked temporarily from the datawave repo's
```
lbschanno: Please remove this file.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexConfig.java
```diff
@@ -0,0 +1,193 @@
...
+
+    /**
+     * The overloaded field parsed from an instance of the property {@link WhindexFieldIngestHelper#DELETE_SRC_FIELD}. A flag indicating whether the original
+     * source field should be deleted (overloaded=true) or retained (overloaded=false). The default is assumed false.
+     */
+    private final boolean overloaded;
+
+    private WhindexConfig(Builder builder) {
+        this.valueField = builder.valueField;
+        this.values = builder.values;
```
lbschanno: The values list needs to be created in an immutable fashion here.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelper.java
```diff
@@ -0,0 +1,227 @@
...
+
+            log.debug("Loaded Whindex configs: " + valueFieldsToWhindexConfigs.entries());
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to setup WhindexFieldIngestHelper", e);
+        }
+    }
+
+    /**
+     * Applies all configured Whindex rules to the given event map. For each config whose <code>valueField</code> has a triggering value, copies all entries
+     * from the <code>sourceField</code> into the new <code>destField</code>, optionally deleting the source.
```
lbschanno: Optionally deleting the source when? Let's be a bit more specific here about what triggers the deletion of the source field (i.e. when the source is marked as overloaded in the config, and only when a whindex entry was created.)

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/WhindexFieldIngestHelper.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    /**
+     * Retrieves an immutable view of the mapping from trigger fields to configs.
+     *
+     * @return a Guava ImmutableMultimap of valueField -> WhindexConfig
+     */
+    @Override
+    public ImmutableMultimap<String,WhindexConfig> getValueFieldsToWhindexConfigs() {
+        return ImmutableMultimap.copyOf(valueFieldsToWhindexConfigs);
```
lbschanno: Rather than returning an immutable copy here, let's make valueFieldsToWhindexConfig and destinationFields always an immutable lists type. If you make them non-final, then you can replace them every time setup() is called, and we can avoid copying more often then we need to.

warehouse/ingest-core/src/main/java/datawave/ingest/metadata/EventMetadata.java
```diff
@@ -416,6 +423,8 @@ public Multimap<BulkIngestKey,Value> getBulkMetadata() {
         addIndexedFieldToMetadata(bulkData, this.compositeFieldsInfo);
         addIndexedFieldToMetadata(bulkData, this.compositeSeparators);
 
+        addIndexedFieldToMetadata(bulkData, this.whindexFieldsInfo);
```
lbschanno: Please add tests to EventMetadataTests to test the new additions here.

lbschanno (changes requested):

ivakegg (dismissed):

lbschanno (dismissed):

warehouse/ingest-core/src/main/java/datawave/ingest/metadata/MetadataWithEarliestDate.java
```diff
@@ -0,0 +1,210 @@
...
+/**
+ * Given a column family (COLF), this class tracks the related metadata's earliest entry per datatype.
+ * <p>
+ * It maintains a mapping between a unique key (constructed from the field name and datatype output name) and a {@link Components} instance, which holds the
+ * metadata field name, its datatype, and the earliest event date encountered.
+ * </p>
+ */
+public class MetadataWithEarliestDate {
+
+    public static final String IGNORED_NORMALIZER_CLASS = null;
```
lbschanno, line 20: This field is never used. Can it be removed?

warehouse/ingest-csv/src/test/java/datawave/ingest/csv/TableConfigurationUtilTest.java
```diff
@@ -107,7 +107,8 @@ public void testCacheFile() throws AccumuloSecurityException, AccumuloException,
 
         Assert.assertEquals(0, tempCacheFile.length());
         tcu.updateCacheFile();
-        Assert.assertEquals(7485, tempCacheFile.length());
+        Assert.assertEquals(7607, tempCacheFile.length());
+        // Assert.assertEquals(7485, tempCacheFile.length());
```
lbschanno, line 111: We can delete this comment.

lbschanno (changes requested):

PR #2829, opened 2025-04-02
dataTypeBug: Fixed Data Type Column Overlap Bug

dws4:
This fixes a bug where there is a Data Type Column Overlap that bleeds into the next column.

### DIFF
--- a/microservices/services/dictionary/service/src/main/frontend/src/functions/formatters.ts
+++ b/microservices/services/dictionary/service/src/main/frontend/src/functions/formatters.ts
@@ -16,21 +16,24 @@ export function parseVal(colName: any, colValue: any): string {
 
 // Produces the max substring for the table, adds '...' if above 34 chars.
 export function maxSubstring(str: any, colName: any): any {
-  if (str == undefined) {
+  if (str === undefined) {
     return;
-  } else if ((colName === 'fieldName' || colName === 'internalFieldName') && str.length > 32) {
-    return str.substring(0, 30) + ' ...';
-  } else if ((colName === 'Types') && str.length > 14) {
-    // Types is offset by 2 to prevent overlapping in 'Tokenized' Column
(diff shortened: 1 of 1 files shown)


microservices/services/dictionary/service/src/main/frontend/src/functions/formatters.ts
```diff
@@ -27,6 +27,8 @@ export function maxSubstring(str: any, colName: any): any {
     return str.substring(0, 22) + ' ...';
   } else if ((colName === 'CopyPaste') && str.length > 42) {
     return str.substring(0, 40) + ' ...';
+  } else if ((colName === 'dataType') && str.length > 12) {
```
jsduncan2: This is getting a little long, what do you think about using a switch instead? Should improve readability and performance a bit.

foster33 (dismissed):

microservices/services/dictionary/service/src/main/frontend/src/functions/formatters.ts
```diff
@@ -16,19 +16,24 @@ export function parseVal(colName: any, colValue: any): string {
...
-  } else {
-    return str;
+  }
+
+  switch (colName) {
+    case 'fieldName':
+    case 'internalFieldName':
+      return str.length > 32 ? str.substring(0, 30) + ' ...' : str;
+    case 'Types':
+      return str.length > 14 ? str.substring(0, 9) + ' ...' : str;
```
jsduncan2, line 29: Can you add back the comment here?
```suggestion
      // Types is offset by 2 to prevent overlapping in 'Tokenized' Column
      return str.length > 14 ? str.substring(0, 9) + ' ...' : str;
```

dws4:
Sorry @foster33 I pushed something else

jsduncan2 (approved):

foster33 (approved):

PR #2835, opened 2025-04-03
* cap begin date at the user specified begin date for configured date…

ivakegg:
… types

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
+++ b/warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
@@ -2083,7 +2083,7 @@ public class DefaultQueryPlanner extends QueryPlanner implements Cloneable {
                     // the Query settings object
                     config.setEndDate(dateIndexData.getEndDate());
                 } else {
-                    log.info("No Remapped dates for " + dateType + " because " + dateIndexData.getEndDate() + " not after " + config.getEndDate());
+                    log.info("No Remapped dates for " + dateType + " because " + config.getEndDate() + " is today");
                 }
             } else {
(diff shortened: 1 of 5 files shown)


warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -2072,15 +2072,26 @@ public ASTJexlScript addDateFilters(final ASTJexlScript queryTree, ScannerFactor
...
+                // now lets update the query parameters with the correct start and
+                // end dates
+                log.info("Remapped " + dateType + " dates [" + config.getBeginDate() + "," + config.getEndDate() + "] to EVENT dates "
+                                + dateIndexData.getBeginDate() + "," + dateIndexData.getEndDate());
+
+                // reset the dates in the configuration, no need to reset them in
+                // the Query settings object
+                config.setBeginDate(dateIndexData.getBeginDate());
+                config.setEndDate(dateIndexData.getEndDate());
+            }
```
apmoriarty: The query planner used to blindly set the begin and end date from the date index data, and now we're respecting the user-supplied behind date if the date type matches the no expansion thing.

Should the query planner also respect the user-supplied end date?

ivakegg: This is the query planner.  Setting the dates in the configuration serve that purpose.  Am I missing the question here?

ivakegg: ok, readjusted to avoid updating begin date altogether if our special type, and to only update end date if not current date.

apmoriarty (approved):

PR #2837, opened 2025-04-04
Delay bulk loads by pattern

alerman:


warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/job/BulkIngestMapFileLoader.java
```diff
@@ -120,7 +123,8 @@ public static void main(String[] args) throws AccumuloSecurityException, IOExcep
             log.error("usage: BulkIngestMapFileLoader hdfsWorkDir jobDirPattern instanceName zooKeepers username password "
                             + "[-sleepTime sleepTime] [-majcThreshold threshold] [-majcCheckInterval count] [-majcDelay majcDelay] "
                             + " [-seqFileHdfs seqFileSystemUri] [-srcHdfs srcFileSystemURI] [-destHdfs destFileSystemURI] [-jt jobTracker] "
-                            + "[-ingestMetricsDisabled] [-jobObservers jobObserverClasses] [-shutdownPort portNum] confFile [{confFile}]");
+                            + "[-ingestMetricsDisabled] [-jobObservers jobObserverClasses] [-shutdownPort portNum] [-delayPathPattern pattern] "
```
ivakegg: -delayPathPattern or -delayPattern ?  change usage or if statement below.

ivakegg: So delay path pattern is a glob against the files under the bulk load directory if I understand the code correctly.  This should be made clear in the comments.  At least add a log.info line to the output around line 340 below.  How it relates to the jobDirPattern is not obvious

ivakegg (commented):
functionally I think it is good, but needs some better parameter help output

alerman:
> functionally I think it is good, but needs some better parameter help output

Updated the usage print to be more verbose.

ivakegg (approved):

awrigh11 (approved):

PR #2839, opened 2025-04-04
type-utils update to support better serialization

apmoriarty:
Update Types so that setter methods do not make additional calls that involve normalizers.

It is now the callers responsibility to select the correct methods when normalization or denormalization is required to build correct Types.

See https://github.com/NationalSecurityAgency/datawave/pull/2840 for integration

### DIFF
--- a/core/utils/type-utils/src/main/java/datawave/data/type/BaseType.java
+++ b/core/utils/type-utils/src/main/java/datawave/data/type/BaseType.java
@@ -5,6 +5,7 @@ import java.util.Collection;
 import java.util.List;
 
 import com.google.common.base.Preconditions;
+
 import datawave.data.normalizer.Normalizer;
 import datawave.webservice.query.data.ObjectSizeOf;
 
--- a/core/utils/type-utils/src/main/java/datawave/data/type/BaseType.java
+++ b/core/utils/type-utils/src/main/java/datawave/data/type/BaseType.java
@@ -4,6 +4,7 @@ import java.io.Serializable;
 import java.util.Collection;
 import java.util.List;
 
+import com.google.common.base.Preconditions;
 import datawave.data.normalizer.Normalizer;
 import datawave.webservice.query.data.ObjectSizeOf;
 
@@ -56,6 +57,7 @@ public class BaseType<T extends Comparable<T> & Serializable> implements Seriali
      * Use the normalizer and delegate to set the normalized value. Assumes that {@link #setDelegate(Comparable)} has been called.
      */
     public void setNormalizedValueFromDelegate() {
+        Preconditions.checkNotNull(delegate, "Tried to set normalized value from delegate, but delegate was null");
         this.normalizedValue = normalizer.normalizeDelegateType(this.delegate);
     }
     
--- a/core/utils/type-utils/src/main/java/datawave/data/type/BaseType.java
+++ b/core/utils/type-utils/src/main/java/datawave/data/type/BaseType.java
@@ -9,16 +9,25 @@ import datawave.webservice.query.data.ObjectSizeOf;
 
 public class BaseType<T extends Comparable<T> & Serializable> implements Serializable, Type<T>, ObjectSizeOf {
     
-    private static final long serialVersionUID = 5354270429891763693L;
+    private static final long serialVersionUID = 1261391800060720465L;
     private static final long STATIC_SIZE = PrecomputedSizes.STRING_STATIC_REF + Sizer.REFERENCE + Sizer.REFERENCE;
     
     protected T delegate;
     protected String normalizedValue;
     protected final Normalizer<T> normalizer;
     
-    public BaseType(String delegateString, Normalizer<T> normalizer) {
+    /**
+     * Default constructor. Generates the normalized and non-normalized variants of the provided value
+     * 
+     * @param value
+     *            the value
+     * @param normalizer
+     *            the normalizer
+     */
+    public BaseType(String value, Normalizer<T> normalizer) {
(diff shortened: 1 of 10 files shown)


core/utils/type-utils/src/main/java/datawave/data/type/BaseType.java
```diff
@@ -9,16 +9,25 @@
...
+     * @param value
+     *            the value
+     * @param normalizer
+     *            the normalizer
+     */
+    public BaseType(String value, Normalizer<T> normalizer) {
         this.normalizer = normalizer;
-        setDelegate(normalizer.denormalize(delegateString));
+        this.delegate = normalizer.denormalize(value);
+        this.normalizedValue = normalizer.normalize(value);
```
ivakegg: So previously we set the normalized value to the normalization of the denormalized value.  This change is slightly different and I wonder if there may be an unintended consequence.

apmoriarty: Good catch, next commit functionally restores the original code path. Rebuilt against main datawave and all tests pass.

core/utils/type-utils/src/main/java/datawave/data/type/BaseType.java
```diff
@@ -9,16 +9,25 @@
...
+     * @param value
+     *            the value
+     * @param normalizer
+     *            the normalizer
+     */
+    public BaseType(String value, Normalizer<T> normalizer) {
         this.normalizer = normalizer;
-        setDelegate(normalizer.denormalize(delegateString));
+        this.delegate = normalizer.denormalize(value);
+        this.normalizedValue = normalizer.normalizeDelegateType(delegate);
```
alerman, line 32: Ok, maybe this makes sense to someone that works in this module a lot, but to me I see "set the delegate to the denormalized, then nomalizedValue= normalize the denormalized". Maybe I'm reading that wrong, or maybe our naming just isn't great.

core/utils/type-utils/src/main/java/datawave/data/type/BaseType.java
```diff
@@ -33,9 +42,21 @@ public void setDelegateFromString(String in) {
...
     public void setDelegate(T delegate) {
         this.delegate = delegate;
-        normalizeAndSetNormalizedValue(this.delegate);
+    }
+    
+    /**
+     * Use the normalizer and delegate to set the normalized value. Assumes that {@link #setDelegate(Comparable)} has been called.
+     */
+    public void setNormalizedValueFromDelegate() {
+        this.normalizedValue = normalizer.normalizeDelegateType(this.delegate);
```
alerman, line 62: does normaliceDelegateType handle nulls if setDelegate wasnt called?

core/utils/type-utils/src/main/java/datawave/data/type/BaseType.java
apmoriarty: The input string may or may not be normalized.
The input string is then denormalized into a delegate (non-normalized)
The the delegate is then run through the normalizer to produce a normalized value. 

I believe you read that correctly, and you are also correct that the naming is not great.

core/utils/type-utils/src/main/java/datawave/data/type/ListType.java
```diff
@@ -39,6 +39,14 @@ public void setDelegateFromString(String in) {
         setNormalizedValue(in);
     }
     
+    /**
+     * Assumes that {@link #setDelegate(Comparable)} has been called
+     */
+    public void setNormalizedValueFromDelegate() {
+        this.normalizedValue = getDelegateAsString();
+        this.normalizedValues = normalizeToMany(getDelegateAsString());
```
dtspence: Minor suggestion, depending on what `getDelegateAsString()` executes (i.e. the type may keep repeated calls), using `normalizedValue` as the input would avoid second call to `getDelegateAsString()`.
EDIT: please resolve if overhead should be minimal.

```suggestion
        this.normalizedValues = normalizeToMany(this.normalizedValue);
```

core/utils/type-utils/src/main/java/datawave/data/type/BaseType.java
apmoriarty: I'll review the normalizers to be certain, but I can also add a preconditions check

apmoriarty: Normalizers assume the input is not null.

core/utils/type-utils/src/main/java/datawave/data/type/ListType.java
```diff
@@ -39,6 +36,16 @@ public void setDelegateFromString(String in) {
         setNormalizedValue(in);
     }
     
+    /**
+     * Assumes that {@link #setDelegate(Comparable)} has been called
+     */
+    public void setNormalizedValueFromDelegate() {
+        if (normalizedValue == null) {
+            normalizedValue = getDelegateAsString();
+        }
```
dtspence, line 45: Would there be a risk of a call to `setNormalizedValue()` or a change to delegate; and missing an update to `normalizedValue` here?

apmoriarty: I don't believe so. That method call is effectively an internal method call. List types use the more explicit normalizeToMany method

ivakegg (approved):

dtspence (approved):

apmoriarty:
Closed in favor of https://github.com/NationalSecurityAgency/datawave/pull/2904

PR #2860, opened 2025-04-15
Simplify Range Stream

apmoriarty:
The RangeStream, affectionately known as the RageStream due to it's complexity, has been the source of much consternation over the years. This merge request attempts to simplify the number of states associated with an index stream, which in turn simplifies the logic that handles unions, intersections, and the final index stream.


alerman (dismissed):

drewfarris:
I'm in favor of simplifying this code.

When looking at this I keep wondering whether the states are actually useful in understanding the nature of query execution or cause of query failures. I've run into a couple cases where a field was unknown, for example that indicated an operational issue. 

Bottom line - is there a way to simplify the code while preserving diagnostic information? (Does this already do that?)

apmoriarty:
> I'm in favor of simplifying this code.
> 
> When looking at this I keep wondering whether the states are actually useful in understanding the nature of query execution or cause of query failures. I've run into a couple cases where a field was unknown, for example that indicated an operational issue.
> 
> Bottom line - is there a way to simplify the code while preserving diagnostic information? (Does this already do that?)

Short answer: no. 

Long answer: it isn't the RangeStream's responsibility to determine executability. There are whole visitors dedicated to that particular problem and now that those visitors have been improved there's no need for the RangeStream to handle multiple responsibilities at once.

warehouse/query-core/src/main/java/datawave/query/index/lookup/Intersection.java
```diff
@@ -171,25 +164,12 @@ public Intersection(Collection<? extends IndexStream> streams, UidIntersector ui
...
-            this.context = StreamContext.EXCEEDED_TERM_THRESHOLD;
-            this.contextDebug = "all children exceeded term threshold";
-        } else if (areAllChildrenSameContext(streams, StreamContext.EXCEEDED_VALUE_THRESHOLD)) {
-            this.context = StreamContext.EXCEEDED_VALUE_THRESHOLD;
-            this.contextDebug = "all children exceeded value threshold";
-            next(); // call next to populate the top key. this is a stream of days generated by the range stream
         } else if (this.children.isEmpty() && !delayedNodes.isEmpty()) {
             // we have a mix of delayed marker nodes
-            this.context = StreamContext.DELAYED_FIELD;
+            this.context = StreamContext.DELAYED;
```
FineAndDandy, line 172: if we intersect a delay and a non-delay as long as there is at least one PRESENT or VARIABLE I think for the sake of intersection is actually PRESENT/VARIABLE?

warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
```diff
@@ -250,17 +250,22 @@ public Iterator<QueryPlan> iterator() {
...
+                        }
+                    case DELAYED:
+                        if (queryStream instanceof Union) {
+                            // all terms in a top level union must be executable
+                            this.context = StreamContext.ABSENT;
+                            this.itr = Collections.emptyIterator();
+                            return itr;
+                        }
+                    default:
+                        context = queryStream.context();
```
FineAndDandy: Revaluate this switch statement. The VARIABLE case will fall through to the DELAYED case. Not sure if that is the intent or not.

warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
```diff
@@ -496,26 +493,19 @@ public IndexStream visit(ASTAndNode node, Object data) {
...
+                    case DELAYED:
                         return build;
-                    case UNINDEXED:
-                        return ScannerStream.unindexed(build.currentNode(), build);
-                    case UNKNOWN_FIELD:
                     case INITIALIZED:
-                        return ScannerStream.unknownField(build.currentNode(), build);
                     default:
-                        return ScannerStream.unknownField(node, build);
+                        throw new RuntimeException("unhandled stream context: " + build.context());
```
FineAndDandy, line 527: to confirm intent, INITIALIZED at this point is an error condition?

warehouse/query-core/src/test/java/datawave/query/index/lookup/IndexStreamComparatorTest.java
```diff
@@ -154,32 +134,20 @@ private ScannerStream s3() {
...
     }
 
     private ScannerStream exceededTerm() {
-        return buildScannerStream("F8", "h", EXCEEDED_TERM_THRESHOLD);
+        return buildScannerStream("F8", "h", DELAYED);
     }
 
     private ScannerStream exceededValue() {
-        return buildScannerStream("F9", "i", EXCEEDED_VALUE_THRESHOLD);
+        return buildScannerStream("F9", "i", PRESENT);
```
FineAndDandy, line 150: why wouldn't an EXCEEDED_VALUE_THRESHOLD be converted to DELAYED?

FineAndDandy (changes requested):
a few questions to resolve

warehouse/query-core/src/main/java/datawave/query/index/lookup/Intersection.java
apmoriarty: That is correct, but in this case the set of executable terms is empty leaving only delayed terms, thus the delay. I'll update the comment to be correct. The comment was trying to say "all terms are delayed, but none of them are of a single type so I cannot surface the most specific context". If that sounds confusing, well, this is why we're simplifying things.

warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
apmoriarty: I debated the best way to write this. A top level union that is variable (a mix of executable and non-executable terms) is not actually executable. So it would fall through the variable case into the delayed case, where it would get marked as absent. 

I think I'll redo this so that the rules of handling unions and intersections are more clear

warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
apmoriarty: Yes, at this point the concurrent scanner initializer has run so if the intersection is still in the initialized state then something went very wrong. I'll add a comment so it's clear why an exception is being thrown.

warehouse/query-core/src/test/java/datawave/query/index/lookup/IndexStreamComparatorTest.java
apmoriarty: Because we cheat. Exceeded value and exceeded or markers are given special treatment where we construct a fake stream of shards. Now that I think about it...we should do the same thing for bounded range markers...

FineAndDandy (approved):

drewfarris (approved):

alerman (approved):

PR #2864, opened 2025-04-16
Allow fields to be excluded from the error index

### LINKED ISSUE #2845: Need a mechanism to exclude fields from the errorIndex
The error index process started in the EventMapper needs a mechanism to exclude fields from being indexed.

lbschanno:
Do not override isIndexedField() and isReversedIndexField() in ErrorShardedIngestHelper. Add unit tests for BaseIngestHelper.isIndexedField() and BaseIngestHelper.isReversedIndexedField(). Verify that disallow lists can be configured for the ErrorShardedIngestHelper.

Fixes #2845


ivakegg (dismissed):

avgAGB (dismissed):

warehouse/ingest-core/src/test/java/datawave/ingest/data/config/ingest/ErrorShardedIngestHelperTest.java
```diff
@@ -0,0 +1,64 @@
...
+    @Test
+    void testSetupGivenDisallowedIndexedFieldLists() {
+        Configuration config = getBaseConfig();
+        config.set(DATA_TYPE_NAME + BaseIngestHelper.DISALLOWLIST_INDEX_FIELDS, "FOO,BAR,HAT");
+        config.set(DATA_TYPE_NAME + BaseIngestHelper.DISALLOWLIST_REVERSE_INDEX_FIELDS, "APPLE,BANANA,KIWI");
+
+        ErrorShardedIngestHelper helper = new ErrorShardedIngestHelper();
+        helper.setup(config);
+
+        Assertions.assertEquals(Set.of("FOO", "BAR", "HAT"), helper.getIndexedFields());
```
hlgp, line 59: I was deeply confused when I read this, but now I see what's going on and I'm sorry we did this to you, and to ourselves.  

It looks like when we have a disallowed list of index fields, we still populate the indexedFields Set in the BaseIngestHelper.  isIndexedField will check the hasIndexDisallowlist and negate the result.  However, getIndexedFields is still public and anything accessing that would need to know to check the disallow flag.  Jimmies.  Rustled.

But, as none of that is your fault, I think it would be less heart-attack inducing to see the an assertion that FOO, BAR, etc. return false from the isIndexedField method until we can get the aforementioned hokiness resolved via a different ticket,

hlgp (dismissed):

warehouse/ingest-core/src/test/java/datawave/ingest/data/config/ingest/ErrorShardedIngestHelperTest.java
lbschanno: Added the requested sanity check in the tests to keep an eye on said hokiness.

ivakegg (approved):

apmoriarty (approved):

PR #2868, opened 2025-04-23
Free ubuntu space

alerman:


apmoriarty (dismissed):

.github/workflows/tests.yml
```diff
@@ -196,6 +204,10 @@ jobs:
           java-version: ${{env.JAVA_VERSION}}
           maven-version: 3.9.5
           cache: 'maven'
+      # Clean up free space
+      - uses: ./.github/actions/free-space
+        with:
+          tool-cache: false
```
jalphonso: Can we also add this to `compose-build-and-test`

.github/actions/free-space/action.yml
```diff
@@ -0,0 +1,242 @@
...
+        printDH "BEFORE CLEAN-UP:"
+        echo ""
+
+
+        # Option: Remove Android library
+
+        if [[ ${{ inputs.android }} == 'true' ]]; then
+          BEFORE=$(getAvailableSpace)
+          
+          sudo rm -rf /usr/local/lib/android || true
```
ivakegg, line 131: Interesting.... can the rpms not be removed instead ?

ivakegg: well, perhaps /usr/local/lib is not from an rpm...

.github/workflows/tests.yml
alerman: added and running pipelines

apmoriarty (approved):

jalphonso (approved):
Thanks for the help, this worked great!

PR #2879, opened 2025-04-28
Bounded Range Validation

apmoriarty:
Bounded range now verify that the lower bound sorts before the upper bound.

Because an invalid bounded range may be considered valid after normalization we cannot check bounded ranges immediately. A visitor was added to validate bounded ranges and that is run after terms are expanded from normalizers. 

- Added convenience methods to LiteralRange that compare the lower and upper bounds to test equivalency or error conditions
- UnmarkedBoundedRangeDetectionVisitor no longer operates on reference nodes per Jexl3 changes.
- Corrected many test queries and expected query plans that had improper bounded terms.
- Corrected a condition where a bounded range expands into a single term and would leave an improper node lineage (ref-and-singleterm)
- Added visitor to validate bounded range terms. It will throw an exception for an illegal range, but only warn about ranges with equivalent bounds that could be rewritten as equality terms.

While working this I noticed ExpandCompositeTerms was generating node trees with an invalid lineage, so opened https://github.com/NationalSecurityAgency/datawave/pull/2883 for that.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
+++ b/warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
@@ -486,6 +486,17 @@ public class RangeStream extends BaseVisitor implements CloseableIterable<QueryP
         } else if (instance.isType(INDEX_HOLE)) {
             return ScannerStream.delayed(node);
         } else if (instance.isType(BOUNDED_RANGE)) {
+            LiteralRange<?> range = JexlASTHelper.findRange().getRange(node);
+            if (range == null) {
+                throw new RuntimeException("BoundedRange was null");
+            }
+
+            if (isIndexed(range.getFieldName(), config.getIndexedFields())) {
+                JexlNode wrapped = JexlNodes.wrap(node);
+                ShardSpecificIndexIterator iter = new ShardSpecificIndexIterator(wrapped, getNumShardFinder(), config.getBeginDate(), config.getEndDate());
+                return ScannerStream.withData(iter, wrapped);
+            }
+
             // here we must have a bounded range that was not expanded, so it must not be expandable via the index
             return ScannerStream.delayed(node);
         } else {
@@ -908,6 +919,7 @@ public class RangeStream extends BaseVisitor implements CloseableIterable<QueryP
             }
             try (Scanner scanner = client.createScanner(TableName.METADATA)) {
                 scanner.setRange(Range.exact(NumShards.NUM_SHARDS, NumShards.NUM_SHARDS_CF));
+                int scannedKeys = 0;
                 for (Map.Entry<Key,Value> entry : scanner) {
                     // num_shards ns:date_shards
                     // num_shards ns:20050207_17
@@ -917,8 +929,13 @@ public class RangeStream extends BaseVisitor implements CloseableIterable<QueryP
                         continue;
                     }
 
+                    scannedKeys++;
                     String[] parts = cq.split("_");
                     cache.put(parts[0], Integer.parseInt(parts[1]));
+                }
+
+                if (scannedKeys == 0) {
+                    log.error("no entries in num_shards cache");
                 }
             } catch (TableNotFoundException | AccumuloException | AccumuloSecurityException e) {
                 // an exception here shouldn't kill the query
--- a/warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
+++ b/warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
@@ -935,7 +935,7 @@ public class RangeStream extends BaseVisitor implements CloseableIterable<QueryP
                 }
 
                 if (scannedKeys == 0) {
-                    log.error("no entries in num_shards cache");
+                    log.fatal("no entries in num_shards cache");
                 }
             } catch (TableNotFoundException | AccumuloException | AccumuloSecurityException e) {
                 // an exception here shouldn't kill the query
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
@@ -983,6 +983,159 @@ public class JexlASTHelper {
     }
 
     /**
+     * Swaps the original node with the replacement, following rules for reference expressions, single terms and junctions
+     * <p>
+     * This method prevents reference expressions from having a single child.
+     * <p>
+     * This method also allows for merging similar junctions.
+     *
+     * @param original
+     *            the original node
+     * @param replacement
+     *            the replacement node
+     */
+    public static void replaceNodeSafely(JexlNode original, JexlNode replacement) {
+        JexlNode parent = original.jjtGetParent();
+
+        if (parent == null) {
+            return; // cannot do anything in this case
+        }
+
+        boolean multiValue = isJunction(replacement);
+        if (multiValue) {
+            replaceNodeWithMultiValue(parent, original, replacement);
+        } else {
+            replaceNodeWithSingleValue(parent, original, replacement);
+        }
+    }
+
+    /**
+     * This method accepts a multivalued replacement node (ASTAndNode or ASTOrNode) and attempts to merge it cleanly into the parent or grandparent node with a
+     * preference for a reference expression as the new parent.
+     *
+     * @param parent
+     *            the parent node
+     * @param original
+     *            the original node
+     * @param replacement
+     *            the replacement node
+     */
+    private static void replaceNodeWithMultiValue(JexlNode parent, JexlNode original, JexlNode replacement) {
+
+        // similar junctions, merge children
+        if (isEquivalentJunctions(parent, replacement)) {
+            mergeJunctions(parent, original, replacement);
+            return;
+        }
+
+        // if no 'legal' rewrite could occur with the parent, try the grandparent
+        JexlNode grandParent = parent.jjtGetParent();
+        if (grandParent != null) {
+            // reference expression, straight replacement
+            if (grandParent instanceof ASTReferenceExpression) {
+                JexlNodes.replaceChild(grandParent, parent, replacement);
+                return;
+            }
+
+            // similar junctions, merge children
+            if (isEquivalentJunctions(grandParent, replacement)) {
+                mergeJunctions(grandParent, parent, replacement);
+                return;
+            }
+        }
+
+        // if no optimal rewrite occurred, fall back to a simple replacement
+        JexlNodes.replaceChild(parent, original, replacement);
+    }
+
+    /**
+     * This method accepts a single replacement node (leaf node) and attempts to merge it cleanly into the parent or grandparent while avoiding a reference
+     * expression as the new parent.
+     *
+     * @param parent
+     *            the parent node
+     * @param original
+     *            the original node
+     * @param replacement
+     *            the replacement node
+     */
+    private static void replaceNodeWithSingleValue(JexlNode parent, JexlNode original, JexlNode replacement) {
+
+        if (!(parent instanceof ASTReferenceExpression)) {
+            // straight replacement
+            JexlNodes.replaceChild(parent, original, replacement);
+            return;
+        }
+
+        // otherwise we have a reference expression and we need to go up a layer
+        JexlNode grandParent = parent.jjtGetParent();
+
+        if (grandParent != null) {
+            // check for the negated single term case: !(term)
+            // it is also unlikely that two reference expressions would sit next to each other in the tree, but it's
+            // not impossible
+            if (!(grandParent instanceof ASTNotNode) && !(grandParent instanceof ASTReferenceExpression)) {
+                JexlNodes.replaceChild(grandParent, parent, replacement);
+                return;
+            }
+        }
+
+        // if no optimal rewrite occurred fallback to a simple replacement
+        JexlNodes.replaceChild(parent, original, replacement);
+    }
+
+    /**
+     * Determines if the provided JexlNode is a junction
+     *
+     * @param node
+     *            the JexlNode
+     * @return true if the node is a junction
+     */
+    public static boolean isJunction(JexlNode node) {
+        return !QueryPropertyMarker.findInstance(node).isAnyType() && (node instanceof ASTOrNode || node instanceof ASTAndNode);
+    }
+
+    /**
+     * Determines if the two JexlNodes are similar junction types (both And nodes or both Or nodes)
+     *
+     * @param a
+     *            first JexlNode
+     * @param b
+     *            second JexlNode
+     * @return true if the two nodes are equivalent junctions
+     */
+    public static boolean isEquivalentJunctions(JexlNode a, JexlNode b) {
+        if (QueryPropertyMarker.findInstance(a).isAnyType() || QueryPropertyMarker.findInstance(b).isAnyType()) {
+            return false;
+        }
+
+        return (a instanceof ASTOrNode && b instanceof ASTOrNode) || (a instanceof ASTAndNode && b instanceof ASTAndNode);
+    }
+
+    /**
+     * Given two equivalent junctions per {@link #isEquivalentJunctions(JexlNode, JexlNode)}, merges the children for node B into the children of node A
+     *
+     * @param parent
+     *            the parent
+     * @param a
+     *            first JexlNode
+     * @param b
+     *            second JexlNode
+     */
+    public static void mergeJunctions(JexlNode parent, JexlNode a, JexlNode b) {
+        List<JexlNode> nodes = new ArrayList<>();
+        for (JexlNode child : JexlNodes.getChildren(parent)) {
+            if (child == a) {
+                // preserves node order
+                nodes.addAll(List.of(JexlNodes.getChildren(b)));
+            } else {
+                nodes.add(child);
+            }
+        }
+        JexlNodes.setChildren(parent, nodes.toArray(new JexlNode[0]));
+    }
+
+    /**
      * Ranges: A range prior to being "tagged" must be of the form "(term1 &amp;&amp; term2)" where term1 and term2 refer to the same field and denote two sides
      * of the range ((LE or LT) and (GE or GT)). A tagged range is of the form "(BoundedRange=true) &amp;&amp; (term1 &amp;&amp; term2))"
      *
@@ -1796,7 +1949,8 @@ public class JexlASTHelper {
                     JexlNode child = node.jjtGetChild(i);
                     if (child != null) {
                         if (child.jjtGetParent() == null) {
-                            String message = "Tree included child " + child + " with a null parent";
+                            String nodeString = JexlStringBuildingVisitor.buildQuery(child);
+                            String message = "Tree included child " + child + " [" + nodeString + "] with a null parent";
                             recordViolation(message, failHard, validation);
                         } else if (child.jjtGetParent() != node) {
                             String message = "Included a child " + child + " with conflicting parent. Expected " + node + " but was " + child.jjtGetParent();
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
@@ -1124,7 +1124,7 @@ public class JexlASTHelper {
      * @param b
      *            second JexlNode
      */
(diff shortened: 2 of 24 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/BoundedRangeIndexExpansionVisitor.java
```diff
@@ -83,6 +83,16 @@ public Object visit(ASTAndNode node, Object data) {
         else if (instance.isType(BOUNDED_RANGE)) {
             LiteralRange<?> range = rangeFinder.getRange(node);
             if (range != null) {
+
+                if (range.areBoundsEquivalent()) {
+                    log.warn("lower and upper bound are equivalent");
```
jschmidt10: Can you add the range to this log as well?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/BaseIndexExpansionVisitor.java
```diff
@@ -187,9 +190,37 @@ protected void rebuildFutureJexlNodes() {
             // if the parent is not null, replace the child
             // if the parent is null, this is the root node, and we will handle that in the expand method
             if (futureJexlNode.jjtGetParent() != null) {
-                JexlNodes.replaceChild(futureJexlNode.jjtGetParent(), futureJexlNode, newNode);
+                if (!(newNode instanceof ASTOrNode || newNode instanceof ASTAndNode)) {
+                    if (isWrappedJunction(futureJexlNode.getOrigNode())) {
+                        JexlNode parent = futureJexlNode.jjtGetParent();
+                        JexlNode grandParent = parent.jjtGetParent();
```
FineAndDandy: could the grandparent be null?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/BaseIndexExpansionVisitor.java
```diff
@@ -187,9 +190,37 @@ protected void rebuildFutureJexlNodes() {
             // if the parent is not null, replace the child
             // if the parent is null, this is the root node, and we will handle that in the expand method
             if (futureJexlNode.jjtGetParent() != null) {
-                JexlNodes.replaceChild(futureJexlNode.jjtGetParent(), futureJexlNode, newNode);
+                if (!(newNode instanceof ASTOrNode || newNode instanceof ASTAndNode)) {
```
FineAndDandy: would love to see a bit more on why this is necessary and what you're correcting.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/UnmarkedBoundedRangeDetectionVisitor.java
```diff
@@ -22,26 +34,30 @@ public static boolean findUnmarkedBoundedRanges(JexlNode script) {
...
+    public Object visit(ASTAndNode node, Object data) {
+        if (data == null) {
+            return null;
+        }
+
+        // check for a bounded marker where the source node is not a range
+        Instance instance = QueryPropertyMarker.findInstance(node);
+        if (instance.isType(BOUNDED_RANGE)) {
+            LiteralRange<?> range = JexlASTHelper.findRange().getRange(node);
+            if (range == null) {
```
FineAndDandy, line 49: this doesn't look quite right. If we have a bounded range and the range is null set hasBounded = true?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/UnmarkedBoundedRangeDetectionVisitor.java
```diff
@@ -22,26 +34,30 @@ public static boolean findUnmarkedBoundedRanges(JexlNode script) {
...
-        else if (JexlASTHelper.findRange().notDelayed().notMarked().isRange(node)) {
-            if (null != data) {
-                AtomicBoolean hasBounded = (AtomicBoolean) data;
-                hasBounded.set(true);
-            }
 
+        // check for a range that is not marked
+        LiteralRange<?> range = JexlASTHelper.findRange().notDelayed().notMarked().getRange(node);
+        if (range != null && range.isBounded()) {
+            AtomicBoolean hasBounded = (AtomicBoolean) data;
```
FineAndDandy: if its not marked is there any concern about the data type cast?

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/UnmarkedBoundedRangeDetectionVisitorTest.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    /**
+     * This illegal range is detected by the {@link ValidateBoundedRangeVisitor}
+     */
+    @Test
+    public void testBoundedMarkerWithInvalidRange() {
+        String query = "((_Bounded_ = true) && (FIELD >= '5' && FIELD <= '3'))";
+        test(false, query);
+    }
```
FineAndDandy, line 30: shouldn't this also throw an exception?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/UnmarkedBoundedRangeDetectionVisitor.java
```diff
@@ -22,26 +34,30 @@ public static boolean findUnmarkedBoundedRanges(JexlNode script) {
     }
 
     @Override
-    public Object visit(ASTReference node, Object data) {
-        // determine if we have a marked range that is not actually a range
-        if (QueryPropertyMarker.findInstance(node).isType(BOUNDED_RANGE)) {
-            if (null != data && !JexlASTHelper.findRange().isRange(node)) {
+    public Object visit(ASTAndNode node, Object data) {
```
FineAndDandy, line 38: I'd like a couple of unit tests that hit all the code paths here.

FineAndDandy (changes requested):
a few questions and requests

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/BaseIndexExpansionVisitor.java
apmoriarty: It could, I'll add a check

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/UnmarkedBoundedRangeDetectionVisitor.java
apmoriarty: This is equivalent to the old code. The LiteralRange.isRange() method returns true if the range is not equals to null. The variables are poorly named in this class, if you think a larger refactor would be better for clarity I'm happy to do that as part of this merge request. In essence, this visitor returns true if an unmarked bounded range was found.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/UnmarkedBoundedRangeDetectionVisitor.java
apmoriarty: No. But now I'm leaning towards a larger refactor for clarity.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/UnmarkedBoundedRangeDetectionVisitorTest.java
apmoriarty: This visitor only checks to see if a bounded range is marked correctly. The BoundedRangeValidationVisitor will handle throwing an exception.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/UnmarkedBoundedRangeDetectionVisitor.java
apmoriarty: Sure thing

apmoriarty:
Removing the integration tested label as the scope of changes grows

lbschanno (changes requested):
A couple comments and requests.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/ValidateBoundedRangeVisitor.java
```diff
@@ -0,0 +1,52 @@
...
+import datawave.query.jexl.nodes.QueryPropertyMarker;
+import datawave.query.jexl.nodes.QueryPropertyMarker.Instance;
+import datawave.query.jexl.visitors.ShortCircuitBaseVisitor;
+
+/**
+ * Verifies the lower bound sorts before the upper bound for all bounded ranges.
+ * <p>
+ * This visitor must be run after bounded range expansion due to how normalization happens.
+ */
+public class ValidateBoundedRangeVisitor extends ShortCircuitBaseVisitor {
```
lbschanno, line 21: Will you be adding tests for this visitor?

warehouse/query-core/src/main/java/datawave/query/jexl/LiteralRange.java
```diff
@@ -4,17 +4,21 @@
...
  * The values of T should already be normalized
  *
- *
- *
  * @param <T>
  *            type of the range
  */
 public class LiteralRange<T extends Comparable<T>> implements Comparable<LiteralRange<T>> {
 
+    private static final Logger log = LoggerFactory.getLogger(LiteralRange.class);
```
lbschanno, line 20: As far as I can tell, this Logger instance isn't used anywhere.

lbschanno (changes requested):
Review attempt 2, hopefully with the actual comments added this time.

warehouse/query-core/src/main/java/datawave/query/jexl/LiteralRange.java
apmoriarty: removed in next commit

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/ValidateBoundedRangeVisitor.java
apmoriarty: added in next commit

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/BaseIndexExpansionVisitor.java
apmoriarty: removed in next commit

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -982,6 +982,159 @@ private static Map<String,Object> getAssignments(JexlNode node, Map<String,Objec
...
+     * This method prevents reference expressions from having a single child.
+     * <p>
+     * This method also allows for merging similar junctions.
+     *
+     * @param original
+     *            the original node
+     * @param replacement
+     *            the replacement node
+     */
+    public static void replaceNodeSafely(JexlNode original, JexlNode replacement) {
```
lbschanno, line 999: Would you be willing to add tests for the new functions you added here?

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/validate/ValidateBoundedRangeVisitorTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Test
+    public void testBoundedMarkerWithInvalidSource() {
+        String query = "((_Bounded_ = true) && (FIELD == '5' && FIELD == '3'))";
+        // unmarked bounded range will throw an exception
+        assertThrows(DatawaveFatalQueryException.class, () -> test(query));
+    }
+
+    @Test
+    public void testNoMarkerValidSource() {
```
lbschanno: `testNoMarkerValidSource()` and `testNoMakerInvalidSource()` are functionally equivalent since there is no marker in either of them, and I didn't see any logic in ValidateBoundedRangeVisitor that cares about EQ and LE/GE in non-marker nodes. I think these can be replaced with a single `testNoMarker()` test.

lbschanno (changes requested):
A couple extra comments. I like the test coverage!

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/validate/ValidateBoundedRangeVisitorTest.java
apmoriarty: Can do

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
apmoriarty: Absolutely

lbschanno (dismissed):

warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
```diff
@@ -917,9 +929,14 @@ private void populateCache() {
                         continue;
                     }
 
+                    scannedKeys++;
                     String[] parts = cq.split("_");
                     cache.put(parts[0], Integer.parseInt(parts[1]));
                 }
+
+                if (scannedKeys == 0) {
+                    log.error("no entries in num_shards cache");
```
FineAndDandy: should this be fatal?

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -982,6 +982,161 @@ private static Map<String,Object> getAssignments(JexlNode node, Map<String,Objec
...
+                JexlNodes.replaceChild(grandParent, parent, replacement);
+                return;
+            }
+
+            // similar junctions, merge children
+            if (isEquivalentJunctions(grandParent, replacement)) {
+                mergeJunctions(grandParent, parent, replacement);
+                return;
+            }
+        }
```
FineAndDandy, line 1047: does this need to consider the type of node of the parent since the method requires any JexlNode?

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -982,6 +982,161 @@ private static Map<String,Object> getAssignments(JexlNode node, Map<String,Objec
...
+
+        if (grandParent != null) {
+            // check for the negated single term case: !(term)
+            // it is also unlikely that two reference expressions would sit next to each other in the tree, but it's
+            // not impossible
+            if (!(grandParent instanceof ASTNotNode) && !(grandParent instanceof ASTReferenceExpression)) {
+                JexlNodes.replaceChild(grandParent, parent, replacement);
+                return;
+            }
+        }
```
FineAndDandy, line 1083: probably also need to explicitly check the type of the parent here?

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -982,6 +982,161 @@ private static Map<String,Object> getAssignments(JexlNode node, Map<String,Objec
...
+     * Given two equivalent junctions per {@link #isEquivalentJunctions(JexlNode, JexlNode)}, merges the children for node B into the children of node A
+     *
+     * @param parent
+     *            the parent
+     * @param a
+     *            first JexlNode
+     * @param b
+     *            second JexlNode
+     */
+    public static void mergeJunctions(JexlNode parent, JexlNode a, JexlNode b) {
```
FineAndDandy: This has some preconditions that need to be checked. No guarantees that parent is related to a or b

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
```diff
@@ -982,6 +982,161 @@ private static Map<String,Object> getAssignments(JexlNode node, Map<String,Objec
...
+        List<JexlNode> nodes = new ArrayList<>();
+        for (JexlNode child : JexlNodes.getChildren(parent)) {
+            if (child == a) {
+                // preserves node order
+                nodes.addAll(List.of(JexlNodes.getChildren(b)));
+            } else {
+                nodes.add(child);
+            }
+        }
```
FineAndDandy, line 1136: this looks like it could lose nodes if parent contains more than just a? Also in that case might add a's nodes multiple times.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/BaseIndexExpansionVisitor.java
```diff
@@ -182,14 +180,31 @@ protected void rebuildFutureJexlNodes() {
 
             rebuildFutureJexlNode(futureJexlNode);
 
-            JexlNode newNode = futureJexlNode.getRebuiltNode();
+            if (futureJexlNode.jjtGetParent() == null) {
```
FineAndDandy, line 183: a future jexl node can never be the root of the tree?

warehouse/query-core/src/test/java/datawave/query/jexl/JexlASTHelperTest.java
FineAndDandy, line 1: can you add some non-happy path tests please?

FineAndDandy (changes requested):
few more things

warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
apmoriarty: sure

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
apmoriarty: No, it does not. The parent method can be used to replace reference expressions, and nodes, or nodes, or any leaf node. The check you see here simply ensure that any replacement produces a logically correct query tree, i.e., no adjacent junctions or reference expressions.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
apmoriarty: Going back to the calling method, the node being replaced (the parent) can be any node.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
apmoriarty: The relationship check is handled in the calling method, but I see you point. I'll change this to private to enforce the preconditions.

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
apmoriarty: Nope. The two places this method is called look like this:
```
mergeJunctions(parent, original, replacement)
mergeJunctions(grandparent, parent, replacement)
```
This method is merging two junctions together while preserving node order. You can see this in the tests added for the JexlASTHelper, two notable examples are below

```
# replacement node preserved and original node order preserved
original: A == '1' && X == 'x' && D == '4'
replacement: B == '2' || C == '3'
A == '1' && (B == '2' || C == '3') && D == '4'

# replacement node is NOT preserved, it's children are inserted in-line into the parent node's children, original node order preserved
A == '1' && X == 'x' && D == '4'
B == '2' && C == '3'
A == '1' && B == '2' && C == '3' && D == '4'
```

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/BaseIndexExpansionVisitor.java
apmoriarty: That is correct. They were previously embedded in the node tree but not anymore because that was dumb.

warehouse/query-core/src/test/java/datawave/query/jexl/JexlASTHelperTest.java
apmoriarty: done

warehouse/query-core/src/main/java/datawave/query/jexl/JexlASTHelper.java
FineAndDandy: this is a public static method you can't rely on the calling method to validate the internals

FineAndDandy (approved):

lbschanno (approved):

PR #2882, opened 2025-04-30
generate edges depending on type category

hlgp:
Requires https://github.com/NationalSecurityAgency/datawave/pull/2885


warehouse/ingest-configuration/src/main/resources/config/edge-definitions.xml
```diff
@@ -35,6 +35,11 @@ http://www.springframework.org/schema/util/spring-util-4.0.xsd">
         <property name="edgeAttribute2" value="NAME"/>
         <property name="edgeAttribute3" value="ID"/>
         <property name="activityDateField" value = "PREMIERED"/>
+        <property name="allowedTypeCategories">
```
ivakegg, line 38: Do we want separate allowed type categories for the source and the sink ?

hlgp: I would let the bidirectional config dictate that.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
```diff
@@ -14,6 +14,7 @@
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import org.apache.accumulo.core.util.Pair;
```
apmoriarty: I thought we were trying to avoid using this class, maybe replace with apache common's Pair?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgeDataTypeHandler.java
```diff
@@ -157,6 +158,8 @@ public class ProtobufEdgeDataTypeHandler<KEYIN,KEYOUT,VALUEOUT> implements Exten
     long futureDelta, pastDelta;
     long newFormatStartDate;
 
+    protected List<datawave.data.type.Type.Category> allowedCategories = Lists.newArrayList();
```
apmoriarty, line 162: I would prefer to see a static import for Type.Category

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgePreconditionTest.java
```diff
@@ -478,6 +480,40 @@ public void testAwarePreconDifferentGroup() {
...
+        ProtobufEdgeDataTypeHandler<Text,BulkIngestKey,Value> edgeHandler = new ProtobufEdgeDataTypeHandler<>();
+        TaskAttemptContext context = new TaskAttemptContextImpl(conf, new TaskAttemptID());
+        edgeHandler.setup(context);
+
+        Set<String> expectedKeys = new HashSet<>();
+
+        RawRecordContainer myEvent = getEvent(conf);
+
+        EdgeHandlerTestUtil.processEvent(fields, edgeHandler, myEvent, 0, true, false);
+        Assert.assertEquals(expectedKeys, EdgeHandlerTestUtil.edgeKeyResults.keySet());
```
apmoriarty, line 513: If this test is demonstrating that no edge keys are produced I would rename the method and rewrite the assertions to be clear. Someone who isn't familiar with this code like myself might wonder what edges are produced when the preconditions belong to different groups, see that we create a set of expected keys, and then go looking for where we updated the expected collection.

warehouse/query-core/src/test/java/datawave/query/util/IndexFieldHoleDataIngest.java
```diff
@@ -368,7 +368,8 @@ private static void writeShardIndexTable(AccumuloClient client, Range range, Lis
                 writeShardIndexEntry(bw, indexCounts, "ETA", "24", true, shard, timeStamp, range.getValue(corleoneUID));
                 writeShardIndexEntry(bw, indexCounts, "ETA", "40", true, shard, timeStamp, range.getValue(corleoneUID));
                 // geo
-                for (String normalized : ((OneToManyNormalizerType<Geometry>) geoType).normalizeToMany("POINT(10 10)")) {
+                for (String normalized : ((OneToManyNormalizerType<Geometry>) geoType).normalizeToMany("POINT(10 10)").stream()
+                                .map(org.apache.accumulo.core.util.Pair::getFirst).collect(Collectors.toList())) {
```
apmoriarty: Use apache common's Pair and use static import

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
hlgp: ah, right.  slipped my mind

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgePreconditionTest.java
hlgp: "// test edge config only allows FULL category. LIST_ELEMENTS edges should not be produced in this case"

No problem renaming the test.  When I hit these in other MRs I usually expand the rest of the Test class to see what is unaffected and other cases currently handled.

"When the preconditions belong to different groups" ?  Not sure what you mean there.  You have one precondition per edge, each edge is aware or unaware of the group, and the fields belong to different groups or the same group.  The test names try to indicate which combination of attributes 
is being tested, which, I agree, is not intuitive. 
 e.g., https://github.com/NationalSecurityAgency/datawave/blob/320f29cb95c3c4e70fa8c8131b0dd2c9230f16c5/warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/handler/edge/ProtobufEdgePreconditionTest.java#L313

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
alerman: AbstractMap.SimplEntry is also a fine replacement.

hlgp:
@gemini-cli /review

PR #2889, opened 2025-05-06
Cleanup ASTs

apmoriarty:
- Add visitor that removes adjacent junctions
- Add visitor that removes extra reference expressions
- Enabled ASTValidation for most tests
- Integrated new visitors to fix invalid query trees
- Updated test assertions where appropriate


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/MergeAdjacentJunctionsVisitor.java
```diff
@@ -0,0 +1,124 @@
...
+import org.apache.commons.jexl3.parser.ParserTreeConstants;
+
+import datawave.query.jexl.nodes.QueryPropertyMarker;
+
+/**
+ * A visitor that merges adjacent {@link ASTAndNode} and {@link ASTOrNode}
+ * <p>
+ * It is easier to run multiple visitors in sequence than it is to merge all functionality into a single visitor.
+ */
+public class MergeAdjacentJunctionsVisitor extends BaseVisitor {
```
lbschanno, line 20: I noticed you're extending BaseVisitor here instead of RebuildingVisitor. It would probably be safer to avoid modifying the input node, and instead return a modified copy like we do in other RebuildingVisitor implementations.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RemoveExtraReferenceExpressionsVisitor.java
```diff
@@ -0,0 +1,60 @@
...
+
+import datawave.query.jexl.nodes.QueryPropertyMarker;
+
+/**
+ * Visitors that modify the query tree will frequently leave behind extra reference expressions.
+ * <p>
+ * This visitor detects and eliminates adjacent reference expressions (double parens) in addition to eliminating references with a single child except marker
+ * nodes and negations.
+ */
+public class RemoveExtraReferenceExpressionsVisitor extends BaseVisitor {
```
lbschanno, line 18: Similar comment here as what I said for MergeAdjacentJunctionsVisitor about extending RebuildingVisitor instead and returning a modified copy.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/MergeAdjacentJunctionsVisitor.java
```diff
@@ -0,0 +1,124 @@
...
+            for (int i = 0; i < node.jjtGetNumChildren(); i++) {
+                JexlNode child = node.jjtGetChild(i);
+                if (JexlNodes.id(child) == target && !QueryPropertyMarker.findInstance(child).isAnyType()) {
+                    children.addAll(List.of(JexlNodes.getChildren(child)));
+                } else {
+                    children.add(child);
+                }
+            }
+            JexlNodes.setChildren(node, children.toArray(new JexlNode[0]));
+        } else if (node.jjtGetNumChildren() == 1 && node.jjtGetParent() != null) {
```
lbschanno, line 84: I didn't see a test for this edge case. Could you add one?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/MergeAdjacentJunctionsVisitor.java
```diff
@@ -0,0 +1,124 @@
...
+
+        if (node.jjtGetParent() != null && node.jjtGetNumChildren() == 1) {
+
+            JexlNode parent = node.jjtGetParent();
+            JexlNode child = node.jjtGetChild(0);
+
+            if (areParentAndChildSameJunctions(parent, child)) {
+                List<JexlNode> children = new ArrayList<>();
+                for (int i = 0; i < parent.jjtGetNumChildren(); i++) {
+                    if (parent.jjtGetChild(i) == node) {
```
lbschanno: Noticed multiple calls to `parent.jjtGetChild(i)`. Would probably be easier to read if you had a scoped variable such as `currChild` that is instantiated once and then used.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/MergeAdjacentJunctionsVisitor.java
```diff
@@ -0,0 +1,124 @@
...
+            }
+            JexlNodes.setChildren(node, children.toArray(new JexlNode[0]));
+        } else if (node.jjtGetNumChildren() == 1 && node.jjtGetParent() != null) {
+            // junction with single child should be eliminated
+            JexlNodes.swap(node.jjtGetParent(), node, node.jjtGetChild(0));
+        }
+    }
+
+    @Override
+    public Object visit(ASTReferenceExpression node, Object data) {
```
lbschanno, line 107: Would like to see a comment here detailing how this merging similar junctions between parents and wrapped junction grandchildren. If that's not what's happening here, then please correct me.

lbschanno (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/MergeAdjacentJunctionsVisitor.java
apmoriarty: Usage of the rebuilding visitor should be discouraged as a general principle. For larger queries you end up with a heavier memory footprint. Any visitor that extends the rebuilding visitor cannot trust the node lineage above or below the current node, especially if nodes could be modified or pruned. For those reason, I do not like to use the rebuilding visitor.

If we care about preserving the original query tree, this visitor can operate on a copy. In the specific cases where this visitor is used, I do not have any reservations about preserving the original.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/RemoveExtraReferenceExpressionsVisitor.java
apmoriarty: See above comment

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/MergeAdjacentJunctionsVisitor.java
apmoriarty: Certainly

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/MergeAdjacentJunctionsVisitor.java
apmoriarty: Agree

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/MergeAdjacentJunctionsVisitor.java
lbschanno: That's totally fair, I'm fine leaving this as is then.

lbschanno (approved):

avgAGB (approved):

PR #2894, opened 2025-05-06
Fixed to actually use the dev profile by default.

ivakegg:
The one concession is that the "default.properties" will need to be added to any other other "read-properties" environments as done in the top level pom.xml

As it turns out the dev profile was never actually activated by default because we always have other profiles activated based on undefined properties.  Now that the build requires some set of read-properties files to be defined, this will fix that glitch.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2061,9 +2061,6 @@
         </profile>
         <profile>
             <id>kubernetes</id>
-            <activation>
-                <activeByDefault>false</activeByDefault>
-            </activation>
             <properties>
                 <build.env>kubernetes</build.env>
                 <env>kubernetes</env>
--- a/pom.xml
+++ b/pom.xml
@@ -1433,8 +1433,10 @@
                             <file>${maven.multiModuleProjectDirectory}/properties</file>
                             <file>${user.home}/.m2/datawave/properties</file>
                         </directories>
-                        <files combine.children="append">
+                        <files>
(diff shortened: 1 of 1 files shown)


alerman (dismissed):

pom.xml
```diff
@@ -2031,32 +2033,6 @@
                 <module>microservices</module>
             </modules>
         </profile>
-        <profile>
-            <id>dev</id>
-            <activation>
-                <activeByDefault>true</activeByDefault>
-            </activation>
-            <properties>
-                <build.env>dev</build.env>
```
jalphonso, line 2040: should we add `<build.env>` as a property up top with the default value?

ivakegg: Good catch.  Done.

pom.xml
```diff
@@ -2031,32 +2036,6 @@
                 <module>microservices</module>
             </modules>
         </profile>
-        <profile>
-            <id>dev</id>
-            <activation>
-                <activeByDefault>true</activeByDefault>
-            </activation>
```
jalphonso, line 2038: We now know that `activeByDefault` is essentially useless for activating profiles since any other profile activation including those activated by a property will disable it. We should remove all instances of its usage

ivakegg: oh, crud, I forgot.
on it

jalphonso (approved):

PR #2898, opened 2025-05-08
Github CI changes to support pom-updates

jalphonso:
Bunch of CI updates to address the following:  
- dev profile is no-more so removed
- update mvn goals and arguments to support recent changes and try and make things a bit more uniform
- ingest image was failing to build because back to back COPYs will not work when the first one is a no-op. I diffed the directories and contents were the same before/after. When the OverlayFS Native Overlay Diff is false (which on the Github runner is) then it errors out in this scenario. Solution was to insert a command between the two COPYs so the layering in OverlayFS can work.
- All DataWave Images (including microservices) and Packages on integration or tags will automatically get deployed to their respective repositories
- Clean out the old submodule .github directories and unused workflows
- Speed up testing by using surefire and failsafe built in retry mechanism when encountering flaky test cases. Example: https://github.com/NationalSecurityAgency/datawave/actions/runs/15052904512/job/42311795202#step:6:4547
- separate out unit tests from integration tests for easier identification


warehouse/assemble/datawave/src/main/docker/Dockerfile
```diff
@@ -5,6 +5,7 @@ FROM ${docker.image.prefix}datawave-stack-accumulo:${docker.image.accumulo.tag}
 USER root
 
 COPY --from=hadoop  /usr/local/hadoop/  /usr/local/hadoop/
+RUN true
```
alerman: Rather than adding a noop maybe move the yum install from below up. Also add a comment on the line after the first copy cautioning not to add a second after

alerman (changes requested):
See comment on noop. 

Also not sure how I feel about requiring the image builds on each build as they are resource intensive and time consuming. I do understand the why just need to think through if I'm ok with this.

alerman:
> See comment on noop.
> 
> Also not sure how I feel about requiring the image builds on each build as they are resource intensive and time consuming. I do understand the why just need to think through if I'm ok with this.

Maybe leave it only on integration and tags but mark it as failure allowed.

alerman:
> > See comment on noop.
> > Also not sure how I feel about requiring the image builds on each build as they are resource intensive and time consuming. I do understand the why just need to think through if I'm ok with this.
> 
> Maybe leave it only on integration and tags but mark it as failure allowed.

I think what you want is to add continue-on-error: true on the build job in build-images

warehouse/assemble/datawave/src/main/docker/Dockerfile
jalphonso: So one question I have is why even have this line if the hadoop files are already present in the base image (datawave-stack-accumulo)?
```
COPY --from=hadoop  /usr/local/hadoop/  /usr/local/hadoop/
```

jalphonso: fixed but question remains. let me know if you want me to remove the COPY line for hadoop

jalphonso:
> See comment on noop.
> 
> Also not sure how I feel about requiring the image builds on each build as they are resource intensive and time consuming. I do understand the why just need to think through if I'm ok with this.

time wasn't too bad, although I suppose it can vary. I was seeing around 21m for the job in question. went ahead and removed it though. but keep in mind that we never build web or ingest containers on a pull request so any issues with those won't be seen until after changes are merged into integration

jalphonso:
I was thinking maybe we should break up the workflows so we can better target what is being regression tested based on what files changed. i.e. don't build microservices or utils if nothing changed there
Might be difficult to know though. May leave this for another ticket

jalphonso:
> See comment on noop.
> 
> Also not sure how I feel about requiring the image builds on each build as they are resource intensive and time consuming. I do understand the why just need to think through if I'm ok with this.

removed

jalphonso:
> > See comment on noop.
> > Also not sure how I feel about requiring the image builds on each build as they are resource intensive and time consuming. I do understand the why just need to think through if I'm ok with this.
> 
> Maybe leave it only on integration and tags but mark it as failure allowed.

removed

jalphonso:
> > > See comment on noop.
> > > Also not sure how I feel about requiring the image builds on each build as they are resource intensive and time consuming. I do understand the why just need to think through if I'm ok with this.
> > 
> > 
> > Maybe leave it only on integration and tags but mark it as failure allowed.
> 
> I think what you want is to add continue-on-error: true on the build job in build-images

added. also added the maven arg `--fail-at-end` for the deploy workflow so we get all the images and packages pushed out that we can.

.github/workflows/deploy-datawave.yml
```diff
@@ -1,5 +1,4 @@
-#
-name: Create and publish a Docker image
+name: Deploy
```
apmoriarty, line 1: Change to "Deploy packages and images" ?

warehouse/assemble/datawave/src/main/docker/Dockerfile
```diff
@@ -4,11 +4,13 @@ FROM ${docker.image.prefix}datawave-stack-accumulo:${docker.image.accumulo.tag}
 
 USER root
 
+# Keep one non-COPY operation between COPY operations. A failure may occur if the destination dir is unchanged and
+# is followed by a second COPY operation. This is related to OverlayFS and its native overlay diff feature.
```
apmoriarty, line 8: +1

apmoriarty (dismissed):

.github/workflows/deploy-datawave.yml
jalphonso: So our other workflow name is Tests. I shortened it so it would be easier to see what was going on in the Checks. The job names within the workflow have more descriptive names. And within the jobs the steps also have names. At this level, I view this as analogous to a stage name in Gitlab CI. Below you'll see a snippet from the Checks in the form of `<workflow name>/<job name>`
```
Deploy / deploy-packages-and-images (push) Successful in 44m
Tests / build-and-test (pull_request) Successful in 60m
...
```

awrigh11 (dismissed):

apmoriarty (approved):

alerman (approved):

PR #2907, opened 2025-05-19
Adds a quickstart test for keyword extraction

drewfarris:


drewfarris:
Oops. Duplicate created

PR #2924, opened 2025-05-22
Fixes to resuming Ivarators after yield; always persist after fillSortedSets; log ivaratorCacheDirs

billoley:
Fixes to resuming Ivarators after yield; always persist after fillSortedSets; log ivaratorCacheDirs

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
+++ b/warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
@@ -16,6 +16,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
+import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
 import java.util.concurrent.TimeUnit;
@@ -1274,6 +1275,16 @@ public abstract class DatawaveFieldIndexCachingIteratorJexl extends WrappingIter
                     canResume = false;
                     break;
                 }
+            }
+        }
+
+        if (previousIvarator != null) {
+            Set<String> currentUriSet = this.ivaratorCacheDirs.stream().map(d -> d.getPathURI()).collect(Collectors.toSet());
+            Set<String> previousUriSet = previousIvarator.ivaratorCacheDirs.stream().map(d -> d.getPathURI()).collect(Collectors.toSet());
+            // this condition doesn't fail the resumeFromIvaratorFutures call, but is cause for concern and should be investigated
+            if (!currentUriSet.equals(previousUriSet)) {
+                log.warn(String.format("Resuming Ivarator %s suspicious condition - currentUriSet != previousUriSet, %s != %s", ivaratorInfo, currentUriSet,
+                                previousUriSet));
             }
         }
 
--- a/warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
+++ b/warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
@@ -16,7 +16,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
-import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
 import java.util.concurrent.TimeUnit;
@@ -1288,18 +1287,6 @@ public abstract class DatawaveFieldIndexCachingIteratorJexl extends WrappingIter
                     canResume = false;
                     break;
                 }
-            }
-        }
-
-        if (previousIvarator != null) {
-            Set<String> currentUriSet = this.ivaratorCacheDirs.stream().map(d -> d.getPathURI()).collect(Collectors.toSet());
-            Set<String> previousUriSet = previousIvarator.ivaratorCacheDirs.stream().map(d -> d.getPathURI()).collect(Collectors.toSet());
-            // ivaratorCacheDirs must be the same. If they are different, we probably have multiple copies of identical Ivarators from
-            // different sequenced terms in the same query. If this happens and that is not the case, then it should be investigated.
-            if (!currentUriSet.equals(previousUriSet)) {
-                log.debug(String.format("Resuming Ivarator %s failed - currentUriSet != previousUriSet, %s != %s", ivaratorInfo, currentUriSet,
-                                previousUriSet));
-                canResume = false;
             }
         }
 
--- a/warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
+++ b/warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
@@ -16,7 +16,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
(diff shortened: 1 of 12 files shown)


warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -921,24 +920,17 @@ else if (key.compareTo(this.lastRangeSeeked.getStartKey()) < 0) {
...
-                // to force setup of a new HDFS backed set
-                usePreviousSortedSet = false;
-            }
-        }
-        if (!usePreviousSortedSet) {
+        // If we are running fillSortedSets as part of a re-seek after yield, then the fillSet threads would not have
+        // completed when a WaitWindowOverrunException was thrown. Therefore, the set will not have been marked as
+        // complete when setupRowBasedHdfsBackedSet is called. We will try to copy the RowBasedHdfsBackedSet from
+        // the previously used Ivarator and resume processing.
+        boolean resumeFromIvaratorFutures = resumeFromIvaratorFutures();
```
billoley, line 907: No need to check seekRange startKey first.  If the previously used IvaratorFutures are not found then either that Ivarator wasn't run before or it was completed and we should find the persisted data in the HdfsBackedSortedSet

resumeFromIvaratorFutures is a better name for this variable

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -1043,11 +1034,6 @@ private void fillSortedSets() throws IOException {
             throw new IvaratorException("Failed Ivarator fillSortedSets: " + result, exception);
         }
 
-        if (usePreviousSortedSet) {
```
billoley, line 1015: the set gets persisted in findTop when this method completes successfully

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -1324,16 +1308,6 @@ protected boolean resumeFromPreviousCall() throws IOException {
             }
         }
```
billoley, line 1310: This check isn't necessary since we are assigning the necessary parts of the HDFSBackedSortedSet from the previous Ivarator to the current Ivarator -- i.e. ivaratorCacheDirs was used the first time around in setupRowBasedHdfsBackedSet

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -1392,15 +1366,16 @@ protected void setupRowBasedHdfsBackedSet(String row) throws IOException {
 
         try {
             // for each of the ivarator cache dirs
```
billoley, line 1350: Nothing really happens in this part unless this.allowDirReuse == false, so check that first

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -1392,15 +1366,16 @@ protected void setupRowBasedHdfsBackedSet(String row) throws IOException {
...
+                    // get the row specific dir
+                    Path rowDir = getRowDir(new Path(ivaratorCacheDir.getPathURI()), row);
+                    FileSystem fs = ivaratorCacheDir.getFs();
+
+                    // if we are not allowing reuse of directories, then delete it
+                    if (fs.exists(rowDir)) {
+                        fs.delete(rowDir, true);
+                    }
                 }
             }
```
billoley, line 1363: log information on correlation between ivaratorCacheDirs and ivaratorInfo for better debugging

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -376,8 +376,9 @@ public void seek(Range range, Collection<ByteSequence> columnFamilies, boolean i
         // preserve the original range for use with the Final Document tracking iterator because it is placed after the ResultCountingIterator
         // so the FinalDocumentTracking iterator needs the start key with the count already appended
         this.originalRange = range;
```
billoley, line 378: separate this.waitWindowObserver.setSeekRange(range) from this.waitWindowObserver.start() since we bypass the latter call when maxYields is exceeded

ivakegg (approved):

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -896,8 +896,8 @@ else if (key.compareTo(this.lastRangeSeeked.getStartKey()) < 0) {
                     throw new RuntimeException("Ivarator query was cancelled");
                 }
 
-                // if we have any persisted data or we have scanned a significant number of keys, then persist it completely
-                if (this.set != null && (this.set.hasPersistedData() || (this.scannedKeys.get() >= this.scanThreshold))) {
+                // persist set to disk so we can reuse it on a rebuild or yield
+                if (this.set != null) {
                     forcePersistence();
```
ivakegg, line 880: Is it possible for the ivarator to have found no results leaving the set null?  If that is the case then we need to make sure the "complete" flag is generated regardless whether we have a set.  I think it is correct but I need to make sure..... investigating.

ivakegg: Ah, I remember now.  The set will be non-null in the normal case.  Its the "unsorted" case where it is null so this is correct.

ivakegg (dismissed):

ivakegg (dismissed):

avgAGB (dismissed):

ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/core/iterators/DatawaveFieldIndexCachingIteratorJexl.java
```diff
@@ -1296,9 +1294,11 @@ protected boolean resumeFromPreviousCall() throws IOException {
         if (previousIvarator != null) {
             Set<String> currentUriSet = this.ivaratorCacheDirs.stream().map(d -> d.getPathURI()).collect(Collectors.toSet());
             Set<String> previousUriSet = previousIvarator.ivaratorCacheDirs.stream().map(d -> d.getPathURI()).collect(Collectors.toSet());
-            // ivaratorCacheDirs is declared final, but must be the same
+            // ivaratorCacheDirs must be the same. If they are different, we probably have multiple copies of identical Ivarators from
+            // different sequenced terms in the same query. If this happens and that is not the case, then it should be investigated.
             if (!currentUriSet.equals(previousUriSet)) {
-                log.error(String.format("Resuming Ivarator %s failed - currentUriSet != previousUriSet", ivaratorInfo));
+                log.debug(String.format("Resuming Ivarator %s failed - currentUriSet != previousUriSet, %s != %s", ivaratorInfo, currentUriSet,
```
ivakegg: Why is this now debug?

apmoriarty (dismissed):

ivakegg (approved):

avgAGB (approved):

PR #2930, opened 2025-05-27
Pull out NumShard's metadata interaction to metadata-utils

SethSmucker:


core/utils/metadata-utils/src/main/java/datawave/query/util/NumShardMetadataHelper.java
```diff
@@ -0,0 +1,57 @@
...
+import org.apache.accumulo.core.data.Value;
+import org.apache.accumulo.core.security.Authorizations;
+import org.apache.hadoop.io.Text;
+import org.apache.log4j.Logger;
+
+public class NumShardMetadataHelper {
+
+    private static final Logger log = Logger.getLogger(NumShardMetadataHelper.class);
+
+    public static ArrayList<String> getNumShardEntries(AccumuloClient accumuloClient, String metadataTableName, Text numShards, Text numShardsCF)
```
ivakegg, line 25: Please look at the metadata helper for examples:
1) should not be static
2) add caching like we do for a mess of the other calls
3) The auths should use the "root auths" like we do on other scanners against the metadata table.

ivakegg (changes requested):

hlgp:
Reopen when comments addressed and ready for reviews

PR #2933, opened 2025-05-30
Serialize Type Class Index

apmoriarty:
Serialize the integer index of a Type class name instead of the full class name for every attribute. 

For attributes that are not found in the DatawaveTypeIndex, such as custom attributes in external projects, the Type class name will still be serialized.

### DIFF
--- a/core/utils/type-utils/src/main/java/datawave/data/type/DatawaveTypeIndex.java
+++ b/core/utils/type-utils/src/main/java/datawave/data/type/DatawaveTypeIndex.java
@@ -1,0 +1,66 @@
+package datawave.data.type;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * A utility class that returns an index for a given Datawave {@link Type}
+ */
+public class DatawaveTypeIndex {
+
+    private DatawaveTypeIndex() {
+        // static utility
+    }
+
+    private static final Map<String,Integer> classNameIndex = new HashMap<>();
+    static {
+        classNameIndex.put(DateType.class.getTypeName(), 1);
+        classNameIndex.put(GeoLatType.class.getTypeName(), 2);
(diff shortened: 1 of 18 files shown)


warehouse/query-core/src/test/java/datawave/query/attributes/TypeAttributeIT.java
```diff
@@ -30,10 +31,12 @@
  */
 public abstract class TypeAttributeIT {
 
+    private static final Logger log = LoggerFactory.getLogger(TypeAttributeIT.class);
+
     private static final int MAX_ITERATIONS = 100_000;
 
     // normalization context
-    protected static String NORMALIZED = "normalized";
+    protected static String NORMALIZED = "    normalized";
```
avgAGB, line 39: was the spacing here intended?

core/utils/type-utils/src/main/java/datawave/data/type/DatawaveTypeIndex.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    /**
+     * Get the index for the provided {@link Type} name
+     *
+     * @param className
+     *            the Type name
+     * @return the index for the Type name, or zero if no such Type exists
+     */
+    public static int getIndexForTypeName(String className) {
```
dtspence, line 52: Minor semantic, could use `OptionalInt` or define a constant for `NOT_FOUND`. No need to change, please feel free to resolve.

warehouse/query-core/src/test/java/datawave/query/attributes/TypeAttributeIT.java
apmoriarty: Yes, this allows you to view pretty A/B numbers in unit tests

dtspence (approved):

hlgp (approved):

avgAGB (approved):

PR #2955, opened 2025-06-09
task/descriptionPopulate: Fixed Column for Description issue

dws4:
MR addresses the following changes:
1. Fixed a bug where the Description Columns do not populate or error out.
2. Updates node module `caniuse-lite`
3. Ignores `.maven-dockerignore` in git-ignore


microservices/services/dictionary/service/src/main/frontend/src/functions/formatters.ts
```diff
@@ -1,16 +1,47 @@
...
+// Ensure that the undefined/null values are loosely compared (i.e '==' not '===')
+export function parseVal(colName: string, colValue: any) : string {
+  if (colName === 'Types') {
     if (colValue == undefined) {
       return '';
     } else {
       return colValue.toString();
     }
+  } else if (colName === 'Descriptions') {
+    if (colValue == undefined || colValue.length === 0) {
```
jsduncan2, line 32: Are we assuming `colValue` is an array here? Can we check `Array.isArray(colValue)` here before the `.length` call to at least make this a bit more readable? When reading this `colValue` is an `any` so it's hard to know what the expectation is.

dws4: So generally in this function, `colValue` can either be a populated array, an array that is null, or a string. But since this is the Descriptions column, the value here will always be an array. This checks to see if it is undefined/null (doesn't have a description) or has a length of 0 (if something was pulled incorrectly in the JSON). If we checked to see if its an array (i.e `isArray(colValue)`) this will always return true here.

foster33 (dismissed):

microservices/services/dictionary/service/src/main/frontend/src/functions/formatters.ts
jsduncan2: Ok then let's at least add a comment here to state that.

dws4: Will do

foster33 (approved):

microservices/services/dictionary/service/src/main/frontend/src/functions/formatters.ts
jsduncan2: Looks great thank you!

jsduncan2 (approved):

PR #2956, opened 2025-06-10
ScanResult now supports partial intersections.

apmoriarty:
A partial intersection is possible when the left hand set is complete and the right hand side is guaranteed to be in sorted order (i.e., an Equality term).

Feature flag partial intersection logic

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/next/ScanResult.java
+++ b/warehouse/query-core/src/main/java/datawave/next/ScanResult.java
@@ -149,7 +149,7 @@ public class ScanResult {
      * @return true if the candidate should be removed
      */
     private boolean nominate(Key candidate, Key min, Key max, Set<Key> otherResults) {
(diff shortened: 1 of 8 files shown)


warehouse/query-core/src/main/java/datawave/next/ScanResult.java
```diff
@@ -0,0 +1,247 @@
...
+     * @param min
+     *            the minimum value of the external set
+     * @param max
+     *            the maximum value of the external set
+     * @param otherResults
+     *            the external set
+     * @return true if the candidate should be removed
+     */
+    private boolean nominate(Key candidate, Key min, Key max, Set<Key> otherResults) {
+        return candidate.compareTo(min) < 0 || candidate.compareTo(max) <= 0 && !otherResults.contains(candidate);
```
ivakegg: Use parens to force operation ordering please.

apmoriarty: updated

avgAGB (approved):

alerman (approved):

PR #2966, opened 2025-06-12
Stop redundant workflow jobs from running after merge queue

foster33:
Currently, when PRs are merged via the "merge queue" they run jobs within the `test.yml` workflow file twice. Once while in the merge queue, and once when it is then pushed to integration. This is wasteful and takes a long time. 

This PR aims to resolve this issue by checking to see if the user who is triggering the workflow, as a result of a push to certain branches, is the github-merge-queue bot. If it is, it will skip over the intensive parts of the jobs and pass the test (since it has already ran and passed in the merge_group event triggered workflow).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -94,13 +94,21 @@ jobs:
         USER_NAME: ${{ secrets.USER_NAME }}
         ACCESS_TOKEN: ${{ secrets.ACCESS_TOKEN }}
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -83,13 +83,21 @@ jobs:
         USER_NAME: ${{ secrets.USER_NAME }}
         ACCESS_TOKEN: ${{ secrets.ACCESS_TOKEN }}
       run: |
-        mvn -s $GITHUB_WORKSPACE/.github/workflows/settings.xml --show-version --batch-mode --errors --no-transfer-progress "-Dstyle.color=always" -Pexamples,assemble,spotbugs -Dmaven.build.cache.enabled=false -Ddeploy -Ddist -T1C clean install -Dsurefire.rerunFailingTestsCount=3 -Dutils -Dservices -Dstarters -DskipITs -DskipFormat
+        if [[ "${{ github.event_name }}" == "push" && "${{ github.triggering_actor }}" == "github-merge-queue" ]]; then
```
jalphonso, line 86: I think for all these we can pull the if logic up to the workflow or job level instead of having to launch the runner just to find out there's no work to be done.

foster33:
This will need to be reworked against the new workflows

PR #2976, opened 2025-06-20
Improve performance of TypeAttribute hashcode method

apmoriarty:
The result of a hashcode calculation is now cached, along with the delegate string

The hashcode is now serialized to avoid expensive calls when deserializing TypeAttributes

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/attributes/TypeAttribute.java
+++ b/warehouse/query-core/src/main/java/datawave/query/attributes/TypeAttribute.java
@@ -37,6 +37,9 @@ public class TypeAttribute<T extends Comparable<T>> extends Attribute<TypeAttrib
 
     private Type<T> datawaveType;
 
+    private int hashCode = Integer.MIN_VALUE;
+    private String delegateString = null;
+
     protected TypeAttribute() {
         super(null, true);
     }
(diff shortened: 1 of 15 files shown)


warehouse/query-core/src/main/java/datawave/query/attributes/TypeAttribute.java
```diff
@@ -128,9 +135,15 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        HashCodeBuilder hcb = new HashCodeBuilder(2099, 2129);
-        hcb.append(datawaveType.getDelegateAsString()).append(super.hashCode());
-        return hcb.toHashCode();
+        if (this.hashCode == Integer.MIN_VALUE) {
+            //  @formatter:off
+            this.hashCode = new HashCodeBuilder(2099, 2129)
+                    .append(datawaveType.getDelegateAsString())
```
dtspence, line 141: Do you know if we would be able to append the `datawaveType` directly (i.e. calling the types hashcode) or is there a reason we hash the string representation?

apmoriarty: I think it's worth looking into that in the near future. I don't know what the trade off is for hashing the string vs. hashing a complex type like a Geometry. It's also worth taking a pass through all Types to make sure they implement equals and hashcode...

dtspence: Thank you, I did not consider the complex types. The `DateType` ran a normalizer when calling the `getDelegateAsString()`, but reviewing more that may be the only one. It is alright to resolve the thread -- unsure if there's a way to avoid the normalizer in `DateType`.

alerman (approved):

dtspence (approved):

PR #2979, opened 2025-06-20
Add release through CI

alerman:
This is meant as a first pass. It will need to get into integration before it can be tested as it uses workflow_dispatch which will not get executed until it exists in the default branch.

### DIFF
--- a/.github/workflows/microservice-release.yml
+++ b/.github/workflows/microservice-release.yml
@@ -109,5 +109,29 @@ jobs:
           -Dtag="${{ steps.version.outputs.VERSION_TAG }}" \
           -DpushChanges=true \
           -DlocalCheckout=true
+    - name: Build and Push packages and images 
+      run: |
+        mvn -s $GITHUB_WORKSPACE/.github/workflows/settings.xml 
+          --show-version 
+          --batch-mode
+          --errors 
+          --no-transfer-progress 
+          --fail-at-end 
+          clean deploy 
+          -Pkubernetes 
+          -Dmaven.build.cache.enabled=false 
+          -Dtar 
+          -Ddeploy 
+          -Ddist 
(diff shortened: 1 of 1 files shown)


.github/workflows/microservice-release.yml
```diff
@@ -0,0 +1,139 @@
...
+        mvn -s $GITHUB_WORKSPACE/.github/workflows/settings.xml --batch-mode release:prepare release:perform \
+          -DreleaseVersion="${{ steps.version.outputs.FINAL_VERSION }}" \
+          -DdevelopmentVersion="${{ steps.version.outputs.NEXT_DEV_VERSION }}" \
+          -Dtag="${{ steps.version.outputs.VERSION_TAG }}" \
+          -DpushChanges=true \
+          -DlocalCheckout=true
+    - name: Build and Push packages and images 
+      run: |
+        git checkout ${{ steps.version.outputs.VERSION_TAG }}
+          cd "${{ github.event.inputs.projectToRelease }}"
```
foster33: Indentation seems to be off, will this affect anything since its yaml?

.github/workflows/microservice-release.yml
```diff
@@ -0,0 +1,139 @@
...
+          -Ddocker-release 
+          -Dutils 
+          -Dservices 
+          -Dstarters 
+          -DskipTests 
+          -DskipFormat 
+          -DskipSpotbugs 
+          -Denforcer.skip=true 
+          -Dcheckstyle.skip=true  
+          -Dmicroservice-docker
```
foster33: Would you need to add `\` for each of these lines for proper line continuation? Like you did in the section above

apmoriarty (approved):

kmtoth (approved):

PR #2991, opened 2025-06-25
Implements additional fixes for tag cloud generation

drewfarris:
Fixed LookupUUIDUtil.generateTagCloud to close query executor when done.

Improved number detection in YakeKeywordExtractor
* Avoids exceptions in Float.parseFloat and an expensive string replace to remove number separators
* Improves number separator detection for multiple locales (e.g., those that done use commas as place separators in numbers)

Adds caching for stopword lists iterator side.
* Avoids multiple reloads of a classpath resource for stopwords by caching the lists on a per-language basis. These don't change often, so no reloading mechanism is required.

Fixed issue where a tag's document frequency score didn't match the unique source count.

### DIFF
--- a/web-services/query/src/main/java/datawave/webservice/query/util/LookupUUIDUtil.java
+++ b/web-services/query/src/main/java/datawave/webservice/query/util/LookupUUIDUtil.java
@@ -773,49 +773,53 @@ public class LookupUUIDUtil {
 
         BaseQueryResponse lastResponse = null;
 
-        do {
-            try {
-                // Get the first/next results
-                BaseQueryResponse response = this.queryExecutor.next(keywordQueryId);
-                lastResponse = response;
+        try {
+            do {
+                try {
+                    // Get the first/next results
+                    BaseQueryResponse response = this.queryExecutor.next(keywordQueryId);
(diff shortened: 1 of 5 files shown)


web-services/query/src/main/java/datawave/webservice/query/util/LookupUUIDUtil.java
```diff
@@ -773,49 +773,53 @@ public <T> T generateTagCloud(final AbstractUUIDLookupCriteria unvalidatedCriter
...
-                lastResponse = response;
-
-                if (TagCloudResponseBase.class.isAssignableFrom(response.getClass())) {
-                    // merge tag cloud responses
-                    final TagCloudResponseBase tagCloudResponse = (TagCloudResponseBase) response;
-                    if (mergedResponse == null) {
-                        mergedResponse = tagCloudResponse;
+        try {
+            do {
+                try {
```
drewfarris, line 778: try/do/try feels weird, but this mirrors what we do elsewhere in this class for consistency.

apmoriarty (approved):

avgAGB (approved):

PR #2996, opened 2025-06-26
change microservices to see checks only run on ms

alerman:


apmoriarty (approved):

alerman:
This should not be merged. This one is just validating the correct jobs execute when changes are localized under microservices build.

PR #2559, opened 2024-09-12
Updating hadoop to 3.4.1

alerman:

PR #2998, opened 2025-06-26
cleanup quickstart and jkube docker configuration

jalphonso:
- Aligned jkube configuration to resolve several warnings
- No longer retrieving packages from CDNs, Maven will download them from our Github packages repository so we can control when they change. Prevents our pipelines from breaking when the next minor rev of a binary is published on the CDN and our current one is removed.
- Maven itself is copied from the `maven:3-amazoncorretto-11` container image in the Dockerfile so no additional bootstrapping is needed
- Sourcing bashrc no longer triggers downloads and installs, merely loads functions and variables into the shell env
- Installation of binaries only occurs at container build time
- bash cleanup, respect function return codes and exit with a corresponding code when appropriate; the image stops building upon first failure
- relocated jdk check to the install scripts since its a prerequisite for the installation of the binaries
- consolidated quickstart and compose testing. compose jobs will run the query and web tests
- docker compose now volume mounts .m2 dir so a user can execute builds if they desire
- Single maintenance point. Update all binary downloads, versions, checksums in the quickstart docker pom


contrib/datawave-quickstart/docker/pom.xml
```diff
@@ -25,7 +25,7 @@
         <!-- These distributions and URLs should match what is listed in each service's bootstrap.sh script -->
         <version.accumulo.tar.gz>2.1.3</version.accumulo.tar.gz>
         <version.hadoop.tar.gz>3.3.6</version.hadoop.tar.gz>
-        <version.maven.tar.gz>3.9.10</version.maven.tar.gz>
+        <version.maven.tar.gz>3.8.8</version.maven.tar.gz>
```
jalphonso: testing

jalphonso: removed so no longer relevant

.github/workflows/build-accumulo.yml
```diff
@@ -80,20 +80,22 @@ jobs:
...
+        username: ${{ env.GITHUB_USER_NAME }}
+        password: ${{ env.GITHUB_ACCESS_TOKEN }}
 
     - name: Update DataWave Dependency Version
       run: |
 
         export JAVA_HOME="$DATAWAVE_JAVA"
         cd "$GITHUB_WORKSPACE/datawave"
-        mvn -s "$GITHUB_WORKSPACE/datawave/.github/workflows/settings.xml" versions:set-property -Dproperty=version.accumulo -DnewVersion=${{ steps.get-accumulo-version.outputs.accumuloVersion }} -DgenerateBackupPoms=false
+        cp $GITHUB_WORKSPACE/.github/workflows/settings.xml ~/.m2/settings.xml
```
alerman: Why copy the file instead of just using the -s? I am not necessarily opposed just curious why.

warehouse/assemble/datawave/src/main/docker/Dockerfile
```diff
@@ -1,6 +1,6 @@
-FROM ${docker.image.prefix}datawave-stack-hadoop:3.3.6 AS hadoop
-FROM bitnami/zookeeper:3.6.3 AS zookeeper
-FROM ${docker.image.prefix}datawave-stack-accumulo:${docker.image.accumulo.tag}
+FROM ${docker.image.prefix}/datawave-stack-hadoop:${docker.image.hadoop.tag} AS hadoop
```
alerman: Let's use version.hadoop, version.accumulo, and version.zookeeper. That way if we are updating those we also force an update to those images.

Note: Also add a comment with a link to the datawave-stack-docker-images so someone who finds a failing docker build wil know where to go.

.github/workflows/build-accumulo.yml
jalphonso: can go either way. originally I was trying to get auth to work inside the quickstart image at build time but later on decided it wasn't worth the trouble. ultimately left it to make the mvn commands a little shorter. can switch it back if you want.

warehouse/assemble/datawave/src/main/docker/Dockerfile
jalphonso: thanks for the suggestion, will do!

.github/workflows/build-accumulo.yml
jalphonso: I'll switch it back. No need to keep it separate now

warehouse/assemble/datawave/src/main/docker/Dockerfile
jalphonso: going to use these from the quickstart/docker pom
```xml
        <version.quickstart.accumulo>2.1.3</version.quickstart.accumulo>
        <version.quickstart.hadoop>3.3.6</version.quickstart.hadoop>
        <version.quickstart.wildfly>17.0.1</version.quickstart.wildfly>
        <version.quickstart.zookeeper>3.8.4</version.quickstart.zookeeper>
```

jalphonso: added these to the `warehouse/assemble/datawave` pom
```xml
        <version.accumulo>2.1.3</version.accumulo>
        <version.hadoop>3.3.6</version.hadoop>
        <version.zookeeper>3.6.3</version.zookeeper>
 ```

File: .github/workflows/compose-and-quickstart.yml [hunk 4]
```diff
@@ -161,46 +155,31 @@ jobs:
             fi
           done
           cd scripts
-          ./testAll.sh
+#          ./testAll.sh
```
jalphonso: temporary: commented out to test the follow on task faster

contrib/datawave-quickstart/bin/services/accumulo/bootstrap.sh
```diff
@@ -49,14 +41,8 @@ admin.enableServer=false"
 
 # Accumulo config
 
-# You may override DW_ACCUMULO_DIST_URI in your env ahead of time, and set as file:///path/to/file.tar.gz for local tarball, if needed
-# DW_ACCUMULO_DIST_URI should, if possible, be using https. There are potential security risks by using http.
-DW_ACCUMULO_VERSION="2.1.3"
```
jalphonso, line 54: Need a way to read the version in from the pom for users running these command locally

contrib/datawave-quickstart/bin/services/accumulo/bootstrap.sh
```diff
@@ -49,14 +41,8 @@ admin.enableServer=false"
...
 DW_ACCUMULO_DIST_URI="${DW_ACCUMULO_DIST_URI:-https://dlcdn.apache.org/accumulo/${DW_ACCUMULO_VERSION}/accumulo-${DW_ACCUMULO_VERSION}-bin.tar.gz}"
-# The sha512 checksum for the tarball. Value should be the hash value only and does not include the file name. Cannot be left blank.
-DW_ACCUMULO_DIST_SHA512_CHECKSUM="${DW_ACCUMULO_DIST_SHA512_CHECKSUM:-1a27a144dc31f55ccc8e081b6c1bc6cc0362a8391838c53c166cb45291ff8f35867fd8e4729aa7b2c540f8b721f8c6953281bf589fc7fe320e4dc4d20b87abc4}"
```
jalphonso, line 57: same here, lets get this from the pom

contrib/datawave-quickstart/bin/services/datawave/install-ingest.sh
```diff
@@ -121,9 +129,9 @@ else
    error "Unable to write ingest-passwd.sh, missing ${DW_DATAWAVE_INGEST_CONFIG_HOME} directory"
 fi
 
-[ ! -d "${DW_DATAWAVE_INGEST_LOG_DIR}" ] && assertCreateDir "${DW_DATAWAVE_INGEST_LOG_DIR}"
-[ ! -d "${DW_DATAWAVE_INGEST_FLAGFILE_DIR}" ] && assertCreateDir "${DW_DATAWAVE_INGEST_FLAGFILE_DIR}"
-[ ! -d "${DW_DATAWAVE_INGEST_LOCKFILE_DIR}" ] && assertCreateDir "${DW_DATAWAVE_INGEST_LOCKFILE_DIR}"
+[ ! -d "${DW_DATAWAVE_INGEST_LOG_DIR}" ] && assertCreateDir "${DW_DATAWAVE_INGEST_LOG_DIR}" || exit 1
+[ ! -d "${DW_DATAWAVE_INGEST_FLAGFILE_DIR}" ] && assertCreateDir "${DW_DATAWAVE_INGEST_FLAGFILE_DIR}" || exit 1
+[ ! -d "${DW_DATAWAVE_INGEST_LOCKFILE_DIR}" ] && assertCreateDir "${DW_DATAWAVE_INGEST_LOCKFILE_DIR}" || exit 1
```
drewfarris, line 134: One of these seems to have tripped me up when re-running a `datawaveIngestInstall` after re-initializing Accumulo. I suspect this was due to the fact that directory already existed. 

Consider whether attempting to create a directory that already exists should be considered a hard failure? It looks like there's a check for that in `assertCreateDir` even though `mkdir -p` is used later in that function.

drewfarris (approved):
This is great. 

I ran into one minor issue with the improved error checking in install-ingest.sh (see comment), but that shoudn't be a blocker to merging this. 

The calls to maven to read variables add a bit of overhead, so we might consider how to amortize that cost in the future but it also shouldn't be a blocker. 

I was able to successfully use the quickstart with my regular workflow that doesn't employ docker. I did not test the docker images.

avgAGB (approved):

warehouse/assemble/datawave/pom.xml
```diff
@@ -12,6 +12,9 @@
     <properties>
         <!-- jkube kubernetes-maven-plugin only has access to project properties so we redefine them to properly render the Dockerfile-->
         <project.version>${project.version}</project.version>
+        <version.accumulo>2.1.3</version.accumulo>
+        <version.hadoop>3.3.6</version.hadoop>
+        <version.zookeeper>3.6.3</version.zookeeper>
```
apmoriarty, line 17: We recently upgraded to hadoop 3.4.1 in https://github.com/NationalSecurityAgency/datawave/pull/2559. Should this be updated in a follow on merge request?

contrib/datawave-quickstart/docker/pom.xml
```diff
@@ -11,23 +11,30 @@
...
-        <version.wildfly.tar.gz>17.0.1</version.wildfly.tar.gz>
-        <version.zookeeper.tar.gz>3.8.4</version.zookeeper.tar.gz>
+
+        <url.accumulo>file://${DW_CLOUD_PLUGINS}/accumulo/${dist.accumulo}</url.accumulo>
+        <url.hadoop>file://${DW_CLOUD_PLUGINS}/hadoop/${dist.hadoop}</url.hadoop>
+        <url.wildfly>file://${DW_CLOUD_PLUGINS}/datawave/${dist.wildfly}</url.wildfly>
+        <url.zookeeper>file://${DW_CLOUD_PLUGINS}/accumulo/${dist.zookeeper}</url.zookeeper>
+
+        <version.quickstart.accumulo>2.1.3</version.quickstart.accumulo>
+        <version.quickstart.hadoop>3.3.6</version.quickstart.hadoop>
```
apmoriarty, line 34: Might be worth adding a note to a readme somewhere about the different places someone will need change when updating the version of hadoop/accumulo/zk/wildfly. I can see these numbers drifting quickly so that a simple find and replace doesn't catch it.

apmoriarty (approved):

contrib/datawave-quickstart/bin/services/datawave/install-ingest.sh
jalphonso: Thanks again for testing. Really appreciate that. I'll keep this in mind for the next PR.

warehouse/assemble/datawave/pom.xml
jalphonso: Thank you. For now I just preserved the versions that were being used for this purpose. I'll test out aligning the versions again on another PR. I know I had issues when I tried to force the same accumulo version globally before.

contrib/datawave-quickstart/docker/pom.xml
jalphonso: I'll keep this in mind for the next PR.

jalphonso:
> This is great.
> 
> I ran into one minor issue with the improved error checking in install-ingest.sh (see comment), but that shoudn't be a blocker to merging this.
> 
> The calls to maven to read variables add a bit of overhead, so we might consider how to amortize that cost in the future but it also shouldn't be a blocker.
> 
> I was able to successfully use the quickstart with my regular workflow that doesn't employ docker. I did not test the docker images.

Great suggestion, I'll look at adding these variables to the env file that gets written out and use that in the next PR. I'm glad everything worked for you!

contrib/datawave-quickstart/bin/services/datawave/install-ingest.sh
jalphonso: There was a bug (return with no return code in the assertCreateDir function just before the mkdir -p)

PR #3007, opened 2025-06-30
Improve speed of document creation by avoiding expensive text decode

apmoriarty:
Hits AttributeFactory, KeyToFieldName, and ValueToAttributes

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/attributes/AttributeFactory.java
+++ b/warehouse/query-core/src/main/java/datawave/query/attributes/AttributeFactory.java
@@ -1,6 +1,5 @@
 package datawave.query.attributes;
 
-import java.nio.ByteBuffer;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


alerman (approved):

warehouse/query-core/src/main/java/datawave/query/attributes/AttributeFactory.java
```diff
@@ -56,17 +55,15 @@ public AttributeFactory(TypeMetadata typeMetadata, String defaultType) {
...
+        for (int i = 0; i < bytes.length(); i++) {
+            if (bytes.byteAt(i) == 0x00) {
+                index = i;
                 break;
-            endPos++;
+            }
         }
-        cf.set(cf.getBytes(), 0, endPos);
-        return cf.toString();
+        return bytes.subSequence(0, index).toString();
```
drewfarris, line 66: I see that `ByteSequence` in the 2.1 or main Accumulo branch doesn't seem to implement `toString()`, does this mean that this should really something like:
```
return new String(bs.subSequence(0, index).toArray(), StandardCharsets.UTF_8)
```

apmoriarty: `ByteSequence` is an abstract class, in practice we are always working with an `ArrayByteSequence` which does implement `toString()`

drewfarris (approved):

warehouse/query-core/src/main/java/datawave/query/attributes/AttributeFactory.java
drewfarris: Thanks for pointing this out - this is why looking at code in GitHub doesn't always work the best.

PR #3009, opened 2025-06-30
Ssdeep overlap scorer update

jack-gitdev:
During ngram overlap scoring, the chunk size compatibility is checked first. If not compatible, then short circuit. If the hashes' chunk/double chunk sizes are compatible, then substring comparison is used to computer overlapping ngrams.

### DIFF
--- a/warehouse/ssdeep-common/src/main/java/datawave/util/ssdeep/NGramGenerator.java
+++ b/warehouse/ssdeep-common/src/main/java/datawave/util/ssdeep/NGramGenerator.java
@@ -79,18 +79,28 @@ public class NGramGenerator implements Serializable {
 
     /**
      *
-     * @param queries
+     * @param queryHashNGramMap
      *            expected to be a collection of SSDeep hashes in chunkSize:chunk:doubleChunk format
      * @return a multimap of NGramTuples mapped to the SSDeepHash from which they originated.
      */
-    public Multimap<NGramTuple,SSDeepHash> preprocessQueries(Collection<SSDeepHash> queries) {
+    public Multimap<NGramTuple,SSDeepHash> preprocessQueries(Multimap<SSDeepHash,NGramTuple> queryHashNGramMap) {
         Multimap<NGramTuple,SSDeepHash> queryMap = TreeMultimap.create();
 
-        for (SSDeepHash queryHash : queries) {
-            generateNgrams(queryHash).forEach(t -> queryMap.put(t, queryHash));
+        for (SSDeepHash queryHash : queryHashNGramMap.keys()) {
+            queryHashNGramMap.get(queryHash).forEach(t -> queryMap.put(t,queryHash));
         }
 
         return queryMap;
+    }
+
+    public Multimap<SSDeepHash,NGramTuple> createHashNGramMap(Collection<SSDeepHash> queries) {
+        Multimap<SSDeepHash,NGramTuple> hashNGramMap = TreeMultimap.create();
+
+        for (SSDeepHash queryHash : queries) {
+            generateNgrams(queryHash).forEach(t -> hashNGramMap.put(queryHash, t));
+        }
+
+        return hashNGramMap;
     }
 
     /**
--- a/warehouse/ssdeep-common/src/main/java/datawave/util/ssdeep/NGramGenerator.java
+++ b/warehouse/ssdeep-common/src/main/java/datawave/util/ssdeep/NGramGenerator.java
@@ -157,18 +157,15 @@ public class NGramGenerator implements Serializable {
         }
     }
 
-    public Set<NGramTuple> calculateOverlappingNGrams(String chunk1, String chunk2, int chunkSize) {
+    public Set<NGramTuple> calculateOverlappingNGrams(int chunkSize, String chunk1, String chunk2) {
         Set<NGramTuple> overlappingNGrams = new HashSet<>();
 
-        // I think min hash size needs to be calculated here
+        final int ngramCount = chunk2.length() - ngramSize;
 
-        String normalizedChunk1 = SSDeepHash.normalizeSSDeepChunk(chunk1, maxRepeatedChars);
-        String normalizedChunk2 = SSDeepHash.normalizeSSDeepChunk(chunk2, maxRepeatedChars);
-
-        for(int i = 0; i < chunk2.length()-6; i++) {
-            String ngramString = chunk2.substring(i, i+7);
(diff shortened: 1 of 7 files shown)


jack-gitdev:
Getting an error when style check runs. Working to address that.

warehouse/ssdeep-common/src/main/java/datawave/util/ssdeep/NGramGenerator.java
```diff
@@ -156,4 +156,19 @@ public void generateNgrams(int chunkSize, String chunk, Set<NGramTuple> output)
...
+
+        for (int i = 0; i <= ngramCount; i++) {
+            String ngram = chunk2.substring(i, i + ngramSize);
+            if (chunk1.contains(ngram)) {
+                overlappingNGrams.add(new NGramTuple(chunkSize, ngram));
+            }
+        }
+
+        return overlappingNGrams;
+    }
```
drewfarris: Consider moving `calculateOverlappingNGrams` to `SSDeepNgramOverlapScorer` because it is most functionally relevant to the scorer class as opposed to the generator. Including it here could be be considered a leaky abstraction because the concern of calculating overlaps winds up in the generator.

This could be achieved by adding an argument `int ngramSize` to `calculateOverlappingNGrams` to accept the configured ngram size from the generator`. Additionally, `calculateOverlappingNGrams` could be `private static` (optionally visible for testing)

warehouse/ssdeep-common/src/main/java/datawave/util/ssdeep/SSDeepNGramOverlapScorer.java
```diff
@@ -15,10 +16,22 @@ public SSDeepNGramOverlapScorer(int ngramSize, int maxRepeatedChars, int minHash
...
+            }
+
+            return ngrams;
+        } else if (signature1.hasDoubleChunk() && (signature1.getDoubleChunkSize() == signature2.getChunkSize())) {
+            return generator.calculateOverlappingNGrams(signature1.getDoubleChunkSize(), signature1.getDoubleChunk(), signature2.getChunk());
+        } else if (signature2.hasDoubleChunk() && (signature1.getChunkSize() == signature2.getDoubleChunkSize())) {
+            return generator.calculateOverlappingNGrams(signature1.getChunkSize(), signature1.getChunk(), signature2.getDoubleChunk());
+        } else {
+            return new HashSet<NGramTuple>();
+        }
```
drewfarris: After the changes suggested above, this implementation might look something like:
```java
    /** Calculate the number of ngrams shared between a part of ssdeep hashes. Considers chunk sizes along with
     *  the single chunk / double chunk of each has as appropriate.
     * 
     * @param signature1 the first ssdeep hash for comparison
     * @param signature2 the second ssdeep hash for comparison
     * @return the set of ngrams shared between each ssdeep hash, labeled with chunk size.
     */
    public Set<NGramTuple> apply(SSDeepHash signature1, SSDeepHash signature2) {
        final int ngramSize = generator.getNgramSize();
        if (signature1.getChunkSize() == signature2.getChunkSize()) {
            // generate overlaps for both the single chunk and double chunk, if present.
            final Set<NGramTuple> ngrams = calculateOverlappingNGrams(ngramSize, signature1.getChunkSize(), signature1.getChunk(), signature2.getChunk());
            if (signature1.hasDoubleChunk() && signature2.hasDoubleChunk()) {
                ngrams.addAll(calculateOverlappingNGrams(ngramSize, signature1.getDoubleChunkSize(), signature1.getDoubleChunk(), signature2.getDoubleChunk()));
            }
            return ngrams;
        } else if (signature1.hasDoubleChunk() && (signature1.getDoubleChunkSize() == signature2.getChunkSize())) {
            return calculateOverlappingNGrams(ngramSize, signature1.getDoubleChunkSize(), signature1.getDoubleChunk(), signature2.getChunk());
        } else if (signature2.hasDoubleChunk() && (signature1.getChunkSize() == signature2.getDoubleChunkSize())) {
            return calculateOverlappingNGrams(ngramSize, signature1.getChunkSize(), signature1.getChunk(), signature2.getDoubleChunk());
        } else {
            return Collections.emptySet();
        }
    }

    /** Calculate the number of ngrams shared amongst both chunks. Each chunk must have the same ngram size or the
     *  comparison will be invalid. For efficiency, the shorter chunk will drive the comparison.
     * 
     * @param ngramSize the size of ngrams to compare
     * @param chunkSize the chunk size of the ngrams we are comparing, both chunks must have the same size.
     * @param chunk1 the first chunk to compare
     * @param chunk2 the second chunk to compare
     * @return a set of ngrams shared between each chunk, labeled with chunk size.
     */
    public static Set<NGramTuple> calculateOverlappingNGrams(int ngramSize, int chunkSize, String chunk1, String chunk2) {
        final Set<NGramTuple> overlappingNGrams = new HashSet<>();
        final String smallChunk = chunk1.length() <= chunk2.length() ? chunk1 : chunk2;
        final String largeChunk = chunk2.equals(smallChunk) ? chunk1 : chunk2;
        final int ngramCount = smallChunk.length() - ngramSize;
        
        for (int i = 0; i <= ngramCount; i++) {
            String ngram = smallChunk.substring(i, i + ngramSize);
            if (largeChunk.contains(ngram)) {
                overlappingNGrams.add(new NGramTuple(chunkSize, ngram));
            }
        }
        return overlappingNGrams;
    }
```

drewfarris (changes requested):

warehouse/ssdeep-common/src/main/java/datawave/util/ssdeep/NGramGenerator.java
jack-gitdev: I moved it to SSDeepNGramOverlapScorer and added the change for using the shortest chunk to guide the comparison.

warehouse/ssdeep-common/src/main/java/datawave/util/ssdeep/SSDeepNGramOverlapScorer.java
jack-gitdev: I was able to remove the ngram generator object and only provide the class the ngram size. I did end up making the calculateOverlappingNGrams private for now since it is only used by this class.

drewfarris (approved):

drewfarris:
When someone merges this, please squash as a part of the merge.

apmoriarty (approved):

PR #3016, opened 2025-07-03
fixes #2935: Invoke IngestTypePruningVisitor in the VisitorFunction

### LINKED ISSUE #2935: IngestTypePruningVisitor can produce invalid Jexl Syntax
When the IngestTypePruning visitor is run after the RangeStream it can produce a query string with invalid Jexl

One example is the token `( && (A == '1'))`. 

The root cause of this is unknown, but some possibilities include:
- The query tree pre-RangeStream has invalid node parentage
- The query tree post-RangeStream has invalid node parentage
- The IngestTypePruningVisitor breaks node parentage during it's operation
- Some unhandled edge case in the IngestTypePruningVisitor

ivakegg:


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/validate/MinimalReferenceExpressionsVisitor.java
```diff
@@ -132,14 +132,15 @@ private boolean isAnyChildAWrappedSingleTerm(JexlNode node) {
...
+                invalidate(node, reason.WRAPPED_SINGLETON);
                 return true;
             }
         }
         return false;
     }
 
     private boolean isWrappedSingleTerm(JexlNode node) {
-        if (node instanceof ASTReferenceExpression) {
+        if (node instanceof ASTReferenceExpression && !(node.jjtGetParent() instanceof ASTNotNode)) {
```
apmoriarty, line 143: Great catch

apmoriarty (dismissed):

hlgp (dismissed):

hlgp (approved):

avgAGB (approved):

PR #3018, opened 2025-07-03
fixes #3015: updated to avoid tokenizing regex expressions

### LINKED ISSUE #3015: LuceneToJexl will incorrectly interpret regex on tokenized fields
If you supply a regex against a tokenized field (e.g. FIELD:(/my?regex/)), then the lucene to jexl conversion will attempt to tokenize the regex and add within or phrase functions into the parsed tree.  The CustomAnalyzerQueryNodeProcessor is the culprit.

ivakegg:


ivakegg:
Test case added.  Verified that it failed without fix in place.

alerman (approved):

alerman (approved):

hlgp (approved):

PR #3019, opened 2025-07-07
Support grouping functions in TLD

apmoriarty:
Add support for grouping functions in TLD, prevent matching across children

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/functions/GroupingRequiredFilterFunctions.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/functions/GroupingRequiredFilterFunctions.java
@@ -225,10 +225,9 @@ public class GroupingRequiredFilterFunctions {
      */
     private static void manageMatchesInGroupRemainingArgs(Object fieldValue, String regex, String context, Collection<ValueTuple> allMatches,
                     ValueTuple currentMatch, int positionFromRight) {
-
         Key candidateMetadata = ValueTuple.getSourceMetadata(fieldValue);
         Key currentMetadata = ValueTuple.getSourceMetadata(currentMatch);
-        if (!currentMetadata.equals(candidateMetadata)) {
+        if (candidateMetadata != null && currentMetadata != null && !currentMetadata.equals(candidateMetadata)) {
             return; // only allow matching within the same document key
         }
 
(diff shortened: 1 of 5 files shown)


ivakegg (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GroupingRequiredFilterFunctions.java
```diff
@@ -224,6 +225,13 @@ public static Collection<?> matchesInGroup(Object... args) {
      */
     private static void manageMatchesInGroupRemainingArgs(Object fieldValue, String regex, String context, Collection<ValueTuple> allMatches,
                     ValueTuple currentMatch, int positionFromRight) {
+
+        Key candidateMetadata = ValueTuple.getSourceMetadata(fieldValue);
+        Key currentMetadata = ValueTuple.getSourceMetadata(currentMatch);
+        if (!currentMetadata.equals(candidateMetadata)) {
```
FineAndDandy: currentMetadata could be null I think?

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GroupingRequiredFilterFunctions.java
```diff
@@ -338,6 +346,12 @@ public static Collection<?> matchesInGroupLeft(Object... args) {
     private static void manageMatchesInGroupLeftRemainingArgs(Object fieldValue, String regex, Collection<ValueTuple> allMatches, String theFirstMatch,
                     String theNextMatch, ValueTuple currentMatch) {
 
+        Key candidateMetadata = ValueTuple.getSourceMetadata(fieldValue);
+        Key currentMetadata = ValueTuple.getSourceMetadata(currentMatch);
+        if (!currentMetadata.equals(candidateMetadata)) {
```
FineAndDandy: same concern here

FineAndDandy (changes requested):
need a few null checks I think

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GroupingRequiredFilterFunctions.java
apmoriarty: Yes, good catch. Fixed in next commit.

FineAndDandy (approved):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/GroupingRequiredFilterFunctions.java
```diff
@@ -224,6 +225,12 @@ public static Collection<?> matchesInGroup(Object... args) {
      */
     private static void manageMatchesInGroupRemainingArgs(Object fieldValue, String regex, String context, Collection<ValueTuple> allMatches,
                     ValueTuple currentMatch, int positionFromRight) {
+        Key candidateMetadata = ValueTuple.getSourceMetadata(fieldValue);
+        Key currentMetadata = ValueTuple.getSourceMetadata(currentMatch);
+        if (candidateMetadata != null && currentMetadata != null && !currentMetadata.equals(candidateMetadata)) {
```
ivakegg, line 230: So if they are both null, then we cannot match?  I guess in general they should never be null but ok....

ivakegg (approved):

PR #3022, opened 2025-07-08
Validate Unique and Grouping temporal fields

### LINKED ISSUE #3010: Improve Unique Granularity Error Handling
The unique function supports different granularities for date normalized fields. However, there is no error handling to prevent someone from requesting granularity on a non-date normalized field.

For example, adding this to the UniqueTest produces some interesting behavior. The field value pair AGE:30 is run through the DateTimeValueFormatter and the numeric value 30 is normalized to the start of epoch time. 

In other cases such as a text field the normalizer will fail with a thrown exception. While the original value is returned this could incur a performance hit as every value for every non-date field could throw an exception.
```
    @Test
    public void testNonDateFieldWithUniqueGranularity() throws Exception {
        Map<String,String> extraParameters = new HashMap<>();
        extraParameters.put("include.grouping.context", "true");
        extraParameters.put("query.syntax", "LUCENE");

        Set<Set<String>> expected = new HashSet<>();
        expected.add(Sets.newHashSet(WiseGuysIngest.sopranoUID, WiseGuysIngest.corleoneUID, WiseGuysIngest.caponeUID));
        Date startDate = format.parse("20091231");
        Date endDate = format.parse("20150101");

        String queryString = "UUID:/^[CS].*/ AND #UNIQUE(BIRTH_DATE[DAY],AGE[DAY])";
    }
```

It would be good to check the requested unique fields against the TypeMetadata and throw an informative exception if a user requests granularity on a non-date normalized field.

lbschanno:
Ensure that when fields used with the #UNIQUE or #GROUP_BY function, if they are paired with a temporal granularity other than ALL, that they are datetime fields.

Fixes #3010


apmoriarty (dismissed):

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -963,7 +970,10 @@ protected ASTJexlScript updateQueryTree(ScannerFactory scannerFactory, MetadataH
         // groom the query so that any nodes with the literal on the left and the identifier on
         // the right will be re-ordered to simplify subsequent processing
 
-        config.setQueryTree(timedInvertSwappedNodes(timers, config.getQueryTree()));
+        {
```
avgAGB: just wondering, were the braces here meant to block out the method call?

lbschanno: No, not sure how that got left in there. Removing it.

lbschanno: Now removed.

avgAGB (approved):

apmoriarty (approved):

PR #3025, opened 2025-07-08
Task/switch builtin calls

avgAGB:


apmoriarty (dismissed):

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/IngestFieldFilter.java
```diff
@@ -331,9 +331,9 @@ public FieldConfiguration(String fieldStr, boolean fieldCountMustMatch) {
          */
         public void load(String fieldsStr, boolean fieldCountMustMatch) throws IllegalArgumentException {
             if (StringUtils.isNotBlank(fieldsStr)) {
-                for (String pair : StringUtils.split(fieldsStr, PAIR_DELIM)) {
+                for (String pair : fieldsStr.split(",")) {
```
ivakegg: What happened to PAIR_DELIM ?

ivakegg (changes requested):
1 comment.

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/IngestFieldFilter.java
avgAGB: The string specific split call doesn't take characters as parameters, so I ended up using string representations of them. To make it a little neater it probably makes more sense to add String specific constants and sub those in, so I'll refactor this.

apmoriarty (approved):

ivakegg (approved):

Issue #43, opened 2018-03-08
The NumShards cache needs to handle multiple hadoop filesystems


ivakegg:
Currently the NumShards cache can only handle URIs that correspond to the configured defaultFS. However we have instances where we may have multiple available filesystems. Instead of using new FileSystem(conf) we should be using new FileSystem(uri, conf) to create the appropriate one.

PR #3026, opened 2025-07-08
 Replace Deprecated Calls (#43 from datawave-metadata-utils)

SethSmucker:
Replace deprecated Long(long) with suggested alternative.

part of #2443


ivakegg (dismissed):

lbschanno (dismissed):

avgAGB (commented):
I know this hasn't been rebased in a little bit, but when you come back to it I think I found two areas with older Long references in the codebase. They're located in Cardinality.java and DashboardSummary.java. Would those references also be worth updating?

avgAGB (approved):

foster33 (approved):

PR #3028, opened 2025-07-09
Adding URL Search Query Implementation

dws4:
**This PR adds a search query implementation that does the following:**
1. User can input '?search=<query>' to search anything in the table.
2. It is reactive, can be added/removed/changed to a different query, and changes the filter in the search bar as well.
3. Adds comments to help readability for the search functionality.

### DIFF
--- a/microservices/services/dictionary/service/src/main/frontend/src/pages/DataDictionary.vue
+++ b/microservices/services/dictionary/service/src/main/frontend/src/pages/DataDictionary.vue
@@ -268,54 +268,28 @@ onMounted(() => {
     console.log('Error fetching and formatting rows: ' + reason);
   });
 
-  // This block is similar to watch() methods below, but handles the search when initially set via the URL.
-  const q = route.query.search;
-
-  // Converts to valid string and then queries if filter is changed.
-  if (Array.isArray(q)) {
-    changeFilter.value = q[0] || '';
-  } else if (q !== null && q !== undefined) {
-    changeFilter.value = q;
-  }
+  // This line is similar to the one in the watch() method below, but handles the search when initially set via the URL.
+  changeFilter.value = Formatters.filterSearch(route.query.search, changeFilter.value);
 
   if (changeFilter.value) {
     queryTable();
   }
 });
 
-// This watch() handles initial changes (if a search query was added to loaded page).
-// Logic: Input -> URL (UI -> URL)
-watch(changeFilter, (val) => {
-  router.replace({
-    query: {
-      ...route.query,
-      search: val || undefined,
-    },
-  });
-});
-
 // This watch() handles a URL Change from a previous query.
 // Logic: Input + Table (reactive URL -> UI + Filters)
 watch(
   () => route.query.search,
-  (val) => {
-    let newVal = '';
-
+  (searchVal) => {
     // Converts the input into a valid string to be queried.
-    if (Array.isArray(val)) {
-      newVal = val[0] || '';
-    } else if (val !== null && val !== undefined) {
-      newVal = val;
-    }
+    let searchValNew = Formatters.filterSearch(searchVal, '');
 
-    if (newVal !== changeFilter.value) {
-      changeFilter.value = newVal;
-      if (newVal) {
-
+    if (searchValNew !== changeFilter.value) {
+      changeFilter.value = searchValNew;
+      if (searchValNew) {
         // Triggers a re-query if the user has changed to a new value.
         queryTable();
       } else {
-
         // This retriggers back to the original state if user clears.
         filter.value = '';
         const originalRows = rows;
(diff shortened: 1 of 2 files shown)


microservices/services/dictionary/service/src/main/frontend/src/pages/DataDictionary.vue
```diff
@@ -264,8 +267,64 @@ onMounted(() => {
   .catch((reason) => {
     console.log('Error fetching and formatting rows: ' + reason);
   });
+
+  // This block is similar to watch() methods below, but handles the search when initially set not added.
+  const q = route.query.search;
+
+  // Converts to valid string and ten queries if filter is changed.
```
jsduncan2:
```suggestion
  // Converts to valid string and then queries if filter is changed.
```

dws4: oops, thank you 🤦

microservices/services/dictionary/service/src/main/frontend/src/pages/DataDictionary.vue
```diff
@@ -264,8 +267,64 @@ onMounted(() => {
   .catch((reason) => {
     console.log('Error fetching and formatting rows: ' + reason);
   });
+
+  // This block is similar to watch() methods below, but handles the search when initially set not added.
```
jsduncan2: I'm not following this comment fully, I think this is what you meant?
```suggestion
  // This block is similar to watch() methods below, but handles the search when initially set via the URL.
```

microservices/services/dictionary/service/src/main/frontend/src/pages/DataDictionary.vue
```diff
@@ -264,8 +267,64 @@ onMounted(() => {
...
+  const q = route.query.search;
+
+  // Converts to valid string and ten queries if filter is changed.
+  if (Array.isArray(q)) {
```
jsduncan2: Out of curiosity, when would this be an array?

microservices/services/dictionary/service/src/main/frontend/src/pages/DataDictionary.vue
dws4: Yeah that works, thanks

microservices/services/dictionary/service/src/main/frontend/src/pages/DataDictionary.vue
```diff
@@ -264,8 +267,64 @@ onMounted(() => {
...
+    changeFilter.value = q[0] || '';
```
jsduncan2: I noticed you went through the process to not check for "falsy" one line below and instead check for null and undefined. Is that worth doing here too? Also you default to an empty string here but not below, is that intentional?

microservices/services/dictionary/service/src/main/frontend/src/pages/DataDictionary.vue
```diff
@@ -264,8 +267,64 @@ onMounted(() => {
...
+  () => route.query.search,
+  (val) => {
+    let newVal = '';
+
+    // Converts the input into a valid string to be queried.
+    if (Array.isArray(val)) {
+      newVal = val[0] || '';
+    } else if (val !== null && val !== undefined) {
+      newVal = val;
+    }
```
jsduncan2: I've seen this code block duplicated above. Can we make this a function and reference it in both locations?

microservices/services/dictionary/service/src/main/frontend/src/pages/DataDictionary.vue
dws4: So when I tested it was initially failing with a type error. If a query somehow does contain multiple values like `?=search=<something>&search=<something-else>` it wont crash it with an exception, it will just take the first one. I am just safeguarding it here in case something else was added to the search.

microservices/services/dictionary/service/src/main/frontend/src/pages/DataDictionary.vue
dws4: Yeah so q[0] might be undefined or null or a blank string it isn't a boolean. This checks to make sure that the value inputted will not be null/undefined. So if it is an empty string, we do not need to do it below, since it won't make a difference when it is searched. Meaning if the user searches for a blank string it will just default to the base table.

microservices/services/dictionary/service/src/main/frontend/src/pages/DataDictionary.vue
dws4: Yeah I'll do that.

microservices/services/dictionary/service/src/main/frontend/src/pages/DataDictionary.vue
jsduncan2: Ah makes sense

foster33 (approved):
Tested, LGTM

jalphonso:
Has this been integration tested?

jalphonso (approved):

dws4:
Merge this when #3043 is approved

PR #3030, opened 2025-07-09
Dependabot dnsjava-2 version updates

avgAGB:


core/connection-pool/pom.xml
```diff
@@ -9,17 +9,8 @@
     <artifactId>datawave-core-connection-pool</artifactId>
     <name>${project.artifactId}</name>
     <properties>
-        <version.dnsjava>2.1.8</version.dnsjava>
+        <version.dnsjava>3.6.3</version.dnsjava>
     </properties>
```
apmoriarty: This block can be removed as well now that it's a top level dependency.

apmoriarty (approved):

ivakegg (approved):

PR #3033, opened 2025-07-09
Allows JSON POSTs in Datawave Query API

drewfarris:
Currently HTTP POST requests to endpoints exposed by the `QueryExecutorBean` can not be transmitted using `Content-type: application/json`, and there are some circumstances where it is more straightforward to use this format over `application/x-www-form-urlencoded`. This demonstrates one way to support an `application/json` POST payload for the batch lookupUUID, lookupContentUUID and generateTagCloud endpoints

### DIFF
--- a/contrib/datawave-quickstart/bin/services/datawave/test-web/tests/KeywordExtraction.test
+++ b/contrib/datawave-quickstart/bin/services/datawave/test-web/tests/KeywordExtraction.test
@@ -10,6 +10,7 @@
 # unique-ish key for its related documents may be registered in QLF.xml
 # for this purpose
 
+###############################################################
 # First, Extract wikipedia keywords by PAGE_ID
 
 TEST_UID_TYPE="PAGE_ID"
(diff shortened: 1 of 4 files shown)


drewfarris:
There's not great documentation/discussion of having the `@Consumes` annotation handle two different content-types seamlessly. Form data maps well to JAX-RS's `MultivaluedMap`, but the JAX-RS imlemention Datawave is using isn't able to do the same with JSON automatically. I'll capture the error and add it here for posterity.

I found some good references at:
* https://eclipse-ee4j.github.io/jersey.github.io/documentation/latest/message-body-workers.html#d0e7980
* https://eclipse-ee4j.github.io/jersey.github.io/documentation/latest/media.html

alerman (dismissed):

avgAGB (dismissed):

drewfarris:
Capturing the error message here:
```
com.fasterxml.jackson.databind.JsonMappingException: Cannot finda deserializer for non-concrete Map type [map type; class javax.ws.rs.core.MultivaluedMap, simple type, class java.lang.String] -> [collection type; class java.util.List, contains [simple type, class java.lang.String]]]
at [Source: (io.undertow.servlet.spec.ServletInputstreamImpl); line 1, column: 1]",
```

drewfarris:
After considering the error message a bit further, I took another crack at solving this in a generic way. It turns out that we can register a custom JSON deserializer with Jackson that will produce the javax.ws.rs `MultivaluedMap<String, String>` object the QueryExecutor expects.

References that explain how/why this works:

- https://www.baeldung.com/jackson-deserialization
- https://stackoverflow.com/questions/18782313/resteasy-and-contextresolverobjectmapper-for-jackson

avgAGB (approved):

jack-gitdev (approved):

alerman (approved):

PR #3041, opened 2025-07-15
Returned detailed message when query validation rules not configured

lbschanno:
Currently, when query validation rules are not configured for a ShardQueryLogic bean, the resulting exception and message is identical to the default UnsupportedOperationException thrown by the QueryLogic interface. This is misleading, since the operation is supported, but additional configuration is required.

Update the ShardQueryLogic to instead throw a BadRequestQueryException with the error code 412-19 and a message indicating that query validation rules need to be configured for the underlying logic before validation can proceed.

Additionally, add the option `-VQ | --validateQuery` to the quickstart datawaveQuery function to make the command post a query to the validate endpoint.


foster33 (dismissed):
Tested & LGTM

foster33 (dismissed):

core/base-rest-responses/src/main/java/datawave/webservice/query/exception/DatawaveErrorCode.java
```diff
@@ -258,7 +258,7 @@ public enum DatawaveErrorCode {
     STOPWATCH_MISSING(404, 27, "Stopwatch missing."),
     VALUE_OPERATOR_MISSING(404, 28, "No operator for value."),
     SERVICE_NOT_CONFIGURED(404, 29, "Service is not configured."),
-    // 412 Precondition Failed
+    // 412 Precondition Failed,
```
avgAGB:
```suggestion
    // 412 Precondition Failed
```

just noticed it to match the format of the codes above

foster33 (approved):

avgAGB (approved):

PR #3044, opened 2025-07-16
Mitigate build failures by increasing lock timeout

dtspence:
The builds have periodically been failing due to a lock timeout. The following increases a lock timeout so we will wait for longer if it happens.


dtspence:
Closing - changes will be pulled into different merge request.

PR #3053, opened 2025-07-22
Adds warnings for when regex doesn't match

ddanielr:
Adds warning log messages for when tier regex patterns do not match any tservers.

This covers the case where tservers match against one of the tier regexes, but not a second regex.

Removed usage of deprecated date methods.

### DIFF
--- a/warehouse/accumulo-extensions/src/main/java/datawave/ingest/table/balancer/ShardRendezvousHostBalancer.java
+++ b/warehouse/accumulo-extensions/src/main/java/datawave/ingest/table/balancer/ShardRendezvousHostBalancer.java
@@ -8,6 +8,7 @@ import java.time.format.DateTimeParseException;
 import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


alerman (dismissed):

warehouse/accumulo-extensions/src/main/java/datawave/ingest/table/balancer/ShardRendezvousHostBalancer.java
```diff
@@ -159,10 +164,17 @@ protected Function<String,List<TabletServerId>> getServerPartitioner(Collection<
...
         }
 
+        final HashSet<Pattern> unMatchedPatterns = new HashSet<>();
+        unMatchedPatterns.addAll(configuredTiers.stream().map(Map.Entry::getValue).map(Matcher::pattern).collect(Collectors.toSet()));
+
         TreeMap<Long,List<TabletServerId>> serverPartitioningMap = new TreeMap<>();
 
+        // Ensure each tier has a list of tservers. This will ensure that when a tiers regex matches nothing that the tier has an empty list. This will cause
+        // its tablets to not be assigned anywhere.
+        configuredTiers.forEach(e -> serverPartitioningMap.put(e.getKey(), new ArrayList<>()));
```
keith-turner, line 174: Adding this line avoids sending tablets in a tier to a neighboring tier when the tier has no tservers.

ddanielr:
Confirmed that logging warn message shows up as expected
 ```
 [main] WARN  datawave.ingest.table.balancer.ShardRendezvousHostBalancer [] - Regex: willNotMatch[12].* did not match against any tservers
```
I did have to modify some code to work with slf4j, those changes will come across in a different PR as they affected other datawave modules.

keith-turner (approved):

alerman (approved):

PR #3073, opened 2025-07-29
Bringing in createBatchScanner refactor from older MR

avgAGB:


warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
```diff
@@ -2820,6 +2824,30 @@ public void setTfAggregationThresholdMs(int tfAggregationThresholdMs) {
         this.tfAggregationThresholdMs = tfAggregationThresholdMs;
     }
 
+    public BatchScanner createBatchScanner(ScannerFactory scannerFactory, QueryData qd) throws TableNotFoundException {
+        final BatchScanner bs = scannerFactory.newScanner(this.getShardTableName(), this.getAuthorizations(), this.getNumQueryThreads(), this.getQuery());
```
apmoriarty: I don't think we want the config to be creating batch scanners

avgAGB: Yeah, think I ended up porting the wrong commit from Tom Nelson. I'll readjust this and recommit.

warehouse/query-core/src/main/java/datawave/query/tables/ScannerFactory.java
```diff
@@ -515,6 +517,30 @@ public ScannerBase newRfileScanner(String tableName, Set<Authorizations> auths,
         }
     }
 
+    public BatchScanner newScanner(ShardQueryConfiguration config, QueryData qd) throws TableNotFoundException {
```
jschmidt10: So this method used to live in the ShardQueryLogic so the idea of the shard table was implied. Here in ScannerFactory it feels like a more general utility. Thoughts on either passing in the table name or renaming the method to `newShardScanner`?

avgAGB: that makes sense to me, I'll tweak it to pass the table name in.

apmoriarty (approved):

jschmidt10 (approved):

PR #3077, opened 2025-07-30
Updated DatePartitionedQueryPlanner to succeed as long as one

ivakegg:
of the date ranges succeeds


apmoriarty (dismissed):

avgAGB (dismissed):

avgAGB (dismissed):

warehouse/query-core/src/test/java/datawave/query/planner/DatePartitionedQueryPlannerTest.java
```diff
@@ -878,6 +892,37 @@ public void testFieldIndexMinThresholdWithSomeNotMeeting() throws Exception {
...
+        configureEvent(IndexFieldHoleDataIngest.EventConfig.forDate("20130103").withMetadataCount("GENDER", 10L, 2L));
+        configureEvent(IndexFieldHoleDataIngest.EventConfig.forDate("20130104").withMetadataCount("UUID", 10L, 2L));
+        configureEvent(IndexFieldHoleDataIngest.EventConfig.forDate("20130105").withMetadataCount("UUID", 10L, 2L));
+
+        givenQuery("(UUID =~ 'C.*' || GEN == 'MALE')");
+        givenStartDate("20130101");
+        givenEndDate("20130105");
+        givenPlan("(GENDER == 'male' || GENERE == 'male' || UUID == 'capone' || UUID == 'corleone')");
+
+        expectPlan(start("20130101"), end("20130103"), "GENERE == 'male' || UUID == 'capone' || UUID == 'corleone' || ((_Eval_ = true) && (GENDER == 'male'))");
```
hlgp, line 909: In the "Hey Google" meeting, we talked about changing logic where a plan ends up being

EXECUTABLE || FULL TABLE SCAN

to proceed with the executable bit and no longer fail the whole thing.  Did we change our minds there?
Still on partial vacation brain, though, might be missing something....

ivakegg: I think that is exactly what I did here.  As long as one of plans succeeds (i.e. is executable), then the query will run.  Did I fail to do that here?

hlgp: Maybe I'm misreading assertQueryResults, but it looks like you have partialSuccessExpected = false, which reads as no partial success.  You also have fullTableScanRequired = true, which in the real world we do not allow.  Together it reads like this query will only succeed if full table scan is enabled, and I thought you were trying to demonstrate the opposite.

partialSuccessExpected-" This denotes that with full table scan disabled, success is expected anyway (at least one date range will succeed)."  I would expect from this description that your second argument should be true?  To show that at least one date range will succeed?  Is there something I'm missing in here?

ivakegg: Ah, I see the problem here.  You are misreading the assertQueryResults.  The full table scan required is initially set to allow us to fully test the query plans.  If it is true, then we subsequently try it out with full table scan disabled to allow us to test whether we can partially be successful.  In this particular test case, I am testing that we fail because ALL of the date partitions fail.

hlgp: Gotcha, thanks.

(Probably worth renaming the fullTableScanRequired to fullyTestQueryPlans and internally handle the full scan logic)

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
```diff
@@ -555,6 +556,12 @@ public class ShardQueryConfiguration extends GenericQueryConfiguration implement
     private boolean useDocumentScheduler = false;
     private DocumentScannerConfig documentScannerConfig;
 
+    /**
+     * This property is used to avoid the pushdown/pullup portion of the planning. This is used by the DatePartitionedQueryPlanner as the pushdown/pullup is
+     * done by the reprocess planning method.
+     */
+    private boolean deferPushdownPullup = false;
```
apmoriarty, line 564: Add `deferPushdownPullup` and `expandUnfieldedValues` to the equals and hashcode methods

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -328,6 +329,7 @@ public String getPlan(AccumuloClient client, Query settings, Set<Authorizations>
                             + (this.getSettings() == null ? "empty" : this.getSettings().getId()) + ')');
         this.config.setExpandFields(expandFields);
         this.config.setExpandValues(expandValues);
+        this.config.setExpandUnfieldedValues(expandValues);
```
apmoriarty, line 332: Add getter/setter for `expandUnfieldedValues`, I'm not sure if `deferPushdownPullup` requires a getter/setter here.

warehouse/query-core/src/test/java/datawave/query/planner/DatePartitionedQueryPlannerTest.java
```diff
@@ -622,18 +630,17 @@ public void testFieldIndexHolesPartiallyWithinDateRange() throws Exception {
         givenQuery("UUID =~ 'C.*' || UUID =~ 'S.*'");
         givenStartDate("20130101");
         givenEndDate("20130104");
-        givenPlan("UUID == 'capone' || UUID == 'corleone' || UUID == 'soprano'");
+        givenInitialPlan("UUID =~ 'c.*' || UUID =~ 's.*'");
 
         expectPlan(start("20130101"), end("20130102"), "UUID == 'capone' || UUID == 'corleone' || UUID == 'soprano'");
-        expectPlan(start("20130103"), end("20130104"),
-                        "((_Eval_ = true) && (UUID == 'capone')) || ((_Eval_ = true) && (UUID == 'corleone')) || ((_Eval_ = true) && (UUID == 'soprano'))");
+        expectPlan(start("20130103"), end("20130104"), "((_Eval_ = true) && (UUID =~ 'c.*')) || ((_Eval_ = true) && (UUID =~ 's.*'))");
```
apmoriarty, line 632: Not necessary for this change set, consider in the future wrapping this in a method like `expectPartition(start, end, plan)`

apmoriarty (changes requested):
Just the two changes to ShardQueryConfig and possibly ShardQueryLogic.

Query planner changes and test updates look good.

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
ivakegg: Will do.   Also adding tests to the ShardedQueryConfigurationTest to ensure everything is covered in equals and hashcode methods.

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
ivakegg: Not sure we need it for expandUnfieldedValues, but I could see a case being made so I will add them.  Definitely not needed for deferPushdownPullup.

warehouse/query-core/src/test/java/datawave/query/planner/DatePartitionedQueryPlannerTest.java
ivakegg: I kinda like it the way it is if you don't mind.

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
```diff
@@ -3175,7 +3193,23 @@ public boolean equals(Object o) {
...
+                isUseDocumentScheduler() == that.isUseDocumentScheduler() &&
+                Objects.equals(getSummaryFieldName(), that.getSummaryFieldName()) &&
+                Objects.equals(getShardDateFormatter(), that.getShardDateFormatter()) &&
+                Objects.equals(getDocumentScannerConfig(), that.getDocumentScannerConfig()) &&
+                Objects.equals(getExcerptIterator(), that.getExcerptIterator()) &&
+                isUseQueryTreeScanHintRules() == that.isUseQueryTreeScanHintRules() &&
+                getMaxLinesToPrint() == that.getMaxLinesToPrint() &&
+                getMaxAnyFieldScanTimeMillis() == that.getMaxAnyFieldScanTimeMillis() &&
+                isDisableIteratorUniqueFields() == that.isDisableIteratorUniqueFields() &&
+                isUseShardedIndex() == that.isUseShardedIndex() &&
```
apmoriarty, line 3220: good catches

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
```diff
@@ -302,66 +304,87 @@ public CloseableIterable<QueryData> process(GenericQueryConfiguration genericCon
...
         // Let's do the planning with the delegate planner first to ensure we have a final date range
         // and appropriately expanded unfielded terms etc.
+        boolean generatePlanOnly = planningConfig.isGeneratePlanOnly();
+        planningConfig.setGeneratePlanOnly(true);
+        boolean expandValues = planningConfig.isExpandValues();
+        // we do NOT want to expand any values yet as they may not be dependable
+        // note we are expanding unfielded values (different flag)
+        planningConfig.setExpandValues(false);
+        planningConfig.setDeferPushdownPullup(true);
+
```
drewfarris, line 322: Minor nit: if we preserve the original `generatePlanOnly` and `expandValues` value so we can add them back to the config later, why don't we do the same for `deferPushdownPullup`?

drewfarris (dismissed):

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
ivakegg: That is not a value that we change via the query logic, so I did not see a need to do that.

ivakegg: However I will do it to be safe...

apmoriarty (approved):

drewfarris (approved):

PR #3083, opened 2025-08-04
Set balancer delay so create shards will go fast in quickstart

alerman:


dtspence (dismissed):

avgAGB (dismissed):

warehouse/ingest-scripts/src/main/resources/bin/ingest/create-shards-since.sh
```diff
@@ -51,11 +51,17 @@ if [[ -n "$3" ]]; then
     shardsPerSplit=$3
 fi
 
-TABLES="${SHARD_TABLE_NAME},${ERROR_SHARD_TABLE_NAME},${QUERYMETRICS_SHARD_TABLE_NAME}"
+
+balancerDelay=1
```
apmoriarty: would prefer to see this as `balancerDelay$UNIT`, e.g. `balancerDelayMinutes`

apmoriarty (dismissed):

warehouse/ingest-scripts/src/main/resources/bin/ingest/create-shards-since.sh
alerman: Done

avgAGB (approved):

ivakegg (approved):

apmoriarty (approved):

PR #3085, opened 2025-08-05
slf4j and log4j updates

alerman:
Breaking up the Bulkv2 PR to try to limit its scope and identify issues in smaller batches


ddanielr (dismissed):

warehouse/ingest-scripts/src/main/resources/bin/ingest/map-file-bulk-loader.sh
```diff
@@ -64,7 +66,7 @@ if [[ ! -z $JOB_OBSERVERS ]]; then
   OBSERVER_OPTS="-jobObservers $JOB_OBSERVERS $JOB_OBSERVER_EXTRA_OPTS"
 fi
 
-export HADOOP_OPTS=" ${HADOOP_INGEST_OPTS} -Dapp=bulkIngestMapFileLoader -DshutdownPort=$shutdownPort -Dfile.encoding=UTF8 -Duser.timezone=GMT"
+export HADOOP_OPTS=" ${HADOOP_INGEST_OPTS} -Dlog4j.configuration=log4j-bulkloader.xml -Dapp=bulkIngestMapFileLoader -DshutdownPort=$shutdownPort -Dfile.encoding=UTF8 -Duser.timezone=GMT"
```
jalphonso, line 69: if I'm not mistaken I think this property changes to `-Dlog4j2.configurationFile` when using log4j2

alerman: i have a follow on PR that will do updates to the map file loaders. Ill include this change in that pr

warehouse/ingest-scripts/src/main/resources/bin/ingest/map-file-bulk-loader.sh
```diff
@@ -29,6 +29,8 @@ done
 WORKDIR=${BASE_WORK_DIR}/
 PATTERN="'*'"
 
+CLASSPATH=$CLASSPATH:../../config/log4j-bulkloader.xml
```
jalphonso, line 32: does this config exist yet?

warehouse/ingest-scripts/src/main/resources/bin/ingest/map-file-bulk-loader.sh
jalphonso: sounds good. you might also need `slf4j-log4j12` for the v1 configuration to load. see if the application is actually logging or throwing errors in the console output during the logger initialization. alternatively, on the follow on PR just update to use the log4j2 configuration syntax preferably in `yml`

jalphonso (approved):

ddanielr (approved):

apmoriarty (approved):

PR #3090, opened 2025-08-07
Fixed issue with ivarator building visitor where the directory used was not specific enough

ivakegg:
to prevent improper reuse.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
@@ -1070,7 +1070,7 @@ public class IteratorBuildingVisitor extends BaseVisitor {
      * @throws IOException
      *             for issues with read/write
      */
-    private List<IvaratorCacheDir> getIvaratorCacheDirs(int termNumber, Range rangeLimiter, String field, String value) throws IOException {
+    public List<IvaratorCacheDir> getIvaratorCacheDirs(int termNumber, Range rangeLimiter, String field, String value) throws IOException {
         List<IvaratorCacheDir> pathAndFs = new ArrayList<>();
 
         // use the ivaratorCount / term number to create a unique subdirectory
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
@@ -1036,6 +1036,33 @@ public class IteratorBuildingVisitor extends BaseVisitor {
         return null;
     }
 
+    public String getDocument(Range range) {
+        // if the range has the same document in the start and end key, then return the document name.
+        // This is used to create unique ivarator directories for the DelayedNonEventSubTreeVisitor
+        if (range != null && range.getStartKey() != null && range.getEndKey() != null) {
+            String cf1 = range.getStartKey().getColumnFamily().toString();
+            String cf2 = range.getEndKey().getColumnFamily().toString();
+            StringBuilder builder = new StringBuilder();
+            int minLen = Math.min(cf1.length(), cf2.length());
+            for (int i = 0; i < minLen; i++) {
+                char c1 = cf1.charAt(i);
+                char c2 = cf2.charAt(i);
+                if (c1 == c2) {
+                    if (c1 == NULL_DELIMETER.charAt(0)) {
+                        c1 = '_';
+                    }
+                    builder.append(c1);
+                } else {
+                    break;
+                }
+            }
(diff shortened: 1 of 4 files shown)


ivakegg:
Next... create a unit test case

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
```diff
@@ -1036,18 +1036,49 @@ public Object visit(ASTNumberLiteral node, Object o) {
         return null;
     }
 
+    public String getDocument(Range range) {
+        // if the range has the same document in the start and end key, then return the document name.
+        // This is used to create unique ivarator directories for the DelayedNonEventSubTreeVisitor
+        if (range != null && range.getStartKey() != null && range.getEndKey() != null) {
+            String cf1 = range.getStartKey().getColumnFamily().toString();
```
hlgp, line 1043: NPE here

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
```diff
@@ -1036,18 +1036,49 @@ public Object visit(ASTNumberLiteral node, Object o) {
...
+    private List<IvaratorCacheDir> getIvaratorCacheDirs(int termNumber, Range rangeLimiter, String field, String value) throws IOException {
         List<IvaratorCacheDir> pathAndFs = new ArrayList<>();
 
         // use the ivaratorCount / term number to create a unique subdirectory
-        String subdirectory = ivaratorCacheSubDirPrefix + "term" + termNumber;
+        String subdirectory = ivaratorCacheSubDirPrefix + "_term_" + termNumber + "_field_" + field + "_valueHash_" + value.hashCode();
+        String document = getDocument(rangeLimiter);
+        if (document != null) {
+            subdirectory = subdirectory + "_doc_" + document;
+        }
```
apmoriarty, line 1081: I think we still have a problem where two directories are built. When a shard range scan lands on a tablet the first directory created does not have a document as part of the directory. After a teardown/rebuild the scan continues but now had a record id in the column qualifier and the Ivarator will create a new directory. 

In the extreme case where we teardown after every document I think we will end up with a tremendous amount of directories, like this
```
# for FIELD =~ 'val.*'
prefix_term_val_field_FIELD_valueHash_hash_123
prefix_term_val_field_FIELD_valueHash_hash_123_doc_uid1
prefix_term_val_field_FIELD_valueHash_hash_123_doc_uid2
prefix_term_val_field_FIELD_valueHash_hash_123_doc_uid3
```

I think we need an integration test with MiniAccumuloCluster where an InterruptIterator is configured at a lower priority than the query iterator and we assert that minimal ivarator directories are created.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
apmoriarty: This won't throw an NPE, I walked the shard range case with a debugger. The test case could have been more clear but it's covered here: 
```
start = new Key(shard);
end = new Key(shard + '\u0000');
range = new Range(start, false, end, false);
doc = visitor.getDocument(range);
Assert.assertEquals(null, doc);
```

ivakegg: The problem was that the wrong jar was being used by a tserver during the testing.  
This code is correct.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/IteratorBuildingVisitor.java
ivakegg: Absolutely agreee I still need a unit test here that forces the issue.  That is today's chore.  There are many directories that are now created.  I believe we need to rethink how the evaluation phase running of ivarators is being done.  Also I have an older PR that worked on automatically cleaning up ivarator directories after there were completely used that I need to look at again and rework given the latest changes.

ivakegg: I would like to note that the term number has always been different that the term number of the base ivarators.  The document specific evaluation only ivarators NEVER reused the base ivarators (this is the same in the old code).  Now with the increased yielding and ensuring a directory is marked as completed even if no results are found, this over use of the same directory has become a bug/problem.

alerman (approved):

FineAndDandy:
Approving with https://github.com/NationalSecurityAgency/datawave/issues/3093 as a lien

FineAndDandy (approved):

apmoriarty (approved):

PR #3098, opened 2025-08-12
Remove unfielded numeric regex normalization

jack-gitdev:
Removes use of numeric regex normalizer for unfielded terms. Also updates unit tests to reflect this change.


jack-gitdev:
Set as draft as this still needs integration testing.

PR #3106, opened 2025-08-14
Remember previously assigned shardId

jschmidt10:
There are styles of re-processing jobs where we want to preserve a previously assigned shardId. This will be done via a subclass of RawRecordContainer.

### DIFF
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/data/RawRecordContainer.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/data/RawRecordContainer.java
@@ -42,6 +42,15 @@ public interface RawRecordContainer {
     void setDataType(Type dataType);
 
     /**
+     * Gets the previoulsy assigned shardId if present. This is only used for non-standard flows such as re-processing.
+     *
+     * @return Previously assigned shardId or null if one does not exist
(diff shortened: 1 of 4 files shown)


ivakegg (approved):

lbschanno (approved):

alerman (approved):

apmoriarty (approved):

warehouse/ingest-core/src/main/java/datawave/ingest/data/RawRecordContainer.java
```diff
@@ -41,6 +41,15 @@ public interface RawRecordContainer {
 
     void setDataType(Type dataType);
 
+    /**
+     * Gets the previoulsy assigned shardId if present. This is only used for non-standard flows such as re-processing.
```
avgAGB, line 45:
```suggestion
     * Gets the previously assigned shardId if present. This is only used for non-standard flows such as re-processing.
```

jschmidt10: Dang, nice catch. Sorry I totally missed this comment. Thanks for reviewing

PR #3118, opened 2025-08-26
Updated to allow an atomic replacement of a table

ivakegg:


core/in-memory-accumulo/src/main/java/datawave/accumulo/inmemory/InMemoryTableOperations.java
```diff
@@ -173,9 +173,13 @@ public void delete(String tableName) throws AccumuloException, AccumuloSecurityE
     @Override
     public void rename(String oldTableName, String newTableName)
                     throws AccumuloSecurityException, TableNotFoundException, AccumuloException, TableExistsException {
+        rename(oldTableName, newTableName, false);
+    }
+
+    public void rename(String oldTableName, String newTableName, boolean force) throws TableNotFoundException, TableExistsException {
```
apmoriarty, line 194: Method doc that indicates what the force flag will do?

When a table is renamed is there any change to the table name, namespace, or table id? If not, might be worth adding a simple test that verifies that assumption.

core/in-memory-accumulo/src/main/java/datawave/accumulo/inmemory/InMemoryAccumulo.java
```diff
@@ -41,10 +41,10 @@
...
 
 public class InMemoryAccumulo {
-    final Map<String,InMemoryTable> tables = new HashMap<>();
-    final Map<String,InMemoryNamespace> namespaces = new HashMap<>();
-    final Map<String,String> systemProperties = new HashMap<>();
-    Map<String,InMemoryUser> users = new HashMap<>();
+    final Map<String,InMemoryTable> tables = Collections.synchronizedMap(new HashMap<>());
+    final Map<String,InMemoryNamespace> namespaces = Collections.synchronizedMap(new HashMap<>());
+    final Map<String,String> systemProperties = Collections.synchronizedMap(new HashMap<>());
+    Map<String,InMemoryUser> users = Collections.synchronizedMap(new HashMap<>());
```
apmoriarty: I think the `users` collection can be final as well

core/in-memory-accumulo/src/main/java/datawave/accumulo/inmemory/InMemoryTableOperations.java
ivakegg: done

core/in-memory-accumulo/src/main/java/datawave/accumulo/inmemory/InMemoryAccumulo.java
ivakegg: ok

apmoriarty (approved):

PR #3123, opened 2025-08-28
Update Numeric Range Rule

### LINKED ISSUE #2872: validation update - numbers in non-number fields
Currently, the rule returns an error if any number is searched in a non-number field. It should only validate number ranges, not just a single number. For Example [1 TO 10] {50 TO *] should return the error but just searching 12345 should not.

SethSmucker:
Add FieldsWith**NumericRangeValues**Visitor, which is similar to FieldsWith**NumericValues**Visitor except more specific in the fields it returns.

Differences include:
- Only returning field names that have a range value, instead of returning and field that could be interpreted as a Number Type.
- Throw an error if the field has a range value but IS NOT a Number Type

Add tests for the new class.
Updated old tests to accommodate the new class.

Fixes #2872

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldsWithNumericRangeValuesVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldsWithNumericRangeValuesVisitor.java
@@ -15,7 +15,6 @@ import org.apache.commons.jexl3.parser.JexlNode;
 
 import datawave.query.jexl.JexlASTHelper;
 
-
 public class FieldsWithNumericRangeValuesVisitor extends ShortCircuitBaseVisitor {
 
     /**
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldsWithNumericRangeValuesVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldsWithNumericRangeValuesVisitor.java
@@ -2,40 +2,33 @@ package datawave.query.jexl.visitors;
 
 import java.util.Collections;
 import java.util.LinkedHashSet;
-import java.util.List;
 import java.util.Set;
 
-import datawave.query.jexl.nodes.QueryPropertyMarker;
-import datawave.query.lucene.visitors.QueryNodeType;
 import org.apache.commons.jexl3.parser.ASTAndNode;
-import org.apache.commons.jexl3.parser.ASTEQNode;
 import org.apache.commons.jexl3.parser.ASTGENode;
 import org.apache.commons.jexl3.parser.ASTGTNode;
 import org.apache.commons.jexl3.parser.ASTJexlScript;
 import org.apache.commons.jexl3.parser.ASTLENode;
 import org.apache.commons.jexl3.parser.ASTLTNode;
-import org.apache.commons.jexl3.parser.ASTNENode;
-import org.apache.commons.jexl3.parser.ASTOrNode;
 import org.apache.commons.jexl3.parser.ASTReference;
 import org.apache.commons.jexl3.parser.JexlNode;
-import org.apache.commons.lang3.math.NumberUtils;
 
 import datawave.query.jexl.JexlASTHelper;
-import org.apache.lucene.queryparser.flexible.core.nodes.NotBooleanQueryNode;
-
-import static datawave.query.jexl.nodes.QueryPropertyMarker.MarkerType.BOUNDED_RANGE;
 
 /**
- * A visitor that fetches all fields found in the query that have a numeric range as their value.
+ * Visitor that collects fields participating in range comparisons (<, >, <=, >=).
+ * <p>
+ * Intended to be used after Lucene queries are converted to JEXL, where Lucene ranges are represented as JEXL comparison nodes (optionally wrapped by a bounded
+ * marker). This visitor does not perform any type validation; callers can validate field normalizers via {@code TypeMetadata}.
  */
 public class FieldsWithNumericRangeValuesVisitor extends ShortCircuitBaseVisitor {
 
     /**
-     * Fetch all fields that have a numeric range value.
+     * Collect the set of field names that appear in comparison nodes that represent ranges.
      *
      * @param query
-     *            the query
-     * @return the set of fields
+     *            the parsed JEXL query script
+     * @return ordered set of field names encountered in range comparisons
      */
     @SuppressWarnings("unchecked")
     public static Set<String> getFields(ASTJexlScript query) {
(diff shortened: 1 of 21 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldsWithNumericRangeValuesVisitor.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.commons.jexl3.parser.ASTGTNode;
+import org.apache.commons.jexl3.parser.ASTJexlScript;
+import org.apache.commons.jexl3.parser.ASTLENode;
+import org.apache.commons.jexl3.parser.ASTLTNode;
+import org.apache.commons.jexl3.parser.ASTReference;
+import org.apache.commons.jexl3.parser.JexlNode;
+
+import datawave.query.jexl.JexlASTHelper;
+
+public class FieldsWithNumericRangeValuesVisitor extends ShortCircuitBaseVisitor {
```
lbschanno, line 27: I'd like to see a class-level description here describing what the visitor does.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldsWithNumericRangeValuesVisitor.java
```diff
@@ -0,0 +1,86 @@
...
+     * If the supplied node is a comparison node, add its field identifier to the output set.
+     *
+     * @param node
+     *            a JEXL comparison node
+     * @param data
+     *            the accumulating set {@code (LinkedHashSet<String>)}
+     */
+    @SuppressWarnings("unchecked")
+    private void checkComparisonNodeField(JexlNode node, Object data) {
+        String field = JexlASTHelper.getIdentifier(node);
```
lbschanno, line 90: I don't see any validation here that the value of the field is actually a number. See `FieldsWithNumericValueVisitor.checkSingleField()` for an example.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/FieldsWithNumericRangeValuesVisitorTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+    // Toggle printing the parsed AST via PrintingVisitor.
+    // Set to true to see parsed JEXL and AST output during tests.
+    private static final boolean PRINT_VISITOR = false;
+
+    private String query;
+    private String luceneQuery;
+    private final Set<String> expectedFields = new LinkedHashSet<>();
+
+    private static final LuceneToJexlQueryParser LUCENE = new LuceneToJexlQueryParser();
```
lbschanno: Would like to see this variable renamed to `luceneParser`. Camelcase is usually preferred for non-primitive constants.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/FieldsWithNumericRangeValuesVisitorTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+import datawave.query.jexl.JexlASTHelper;
+import datawave.query.language.parser.jexl.LuceneToJexlQueryParser;
+import datawave.query.language.tree.QueryNode;
+
+class FieldsWithNumericRangeValuesVisitorTest {
+
+    // Toggle printing the parsed AST via PrintingVisitor.
+    // Set to true to see parsed JEXL and AST output during tests.
+    private static final boolean PRINT_VISITOR = false;
```
lbschanno, line 23: Would rename this to `ENABLE_PRINTING` to make its purpose clearer.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/FieldsWithNumericRangeValuesVisitorTest.java
```diff
@@ -0,0 +1,146 @@
...
+        assertLuceneResult();
+    }
+
+    @Test
+    void testLuceneMultipleFields() throws ParseException {
+        givenLuceneQuery("A:[1 TO 2] OR B:[* TO 10} AND C:'xyz' OR D:[5 TO 5]");
+        expectFields("A", "B", "D");
+        assertLuceneResult();
+    }
+
```
lbschanno, line 101: Currently your tests only cover numeric values that are within bounded ranges. There should also be tests that cover the following cases:

- Lucene input where FIELD == Value: `A:1`. Should not return match.
- Lucene input where FIELD != Value: `A:1 NOT B:4`. Should not return match.
- Jexl query explicitly with bounded range: `((_Bounded_ = true) && (FOO >= '1' && FOO <= '4'))`. Should return match.
- Jexl query with ==: `FOO == '1'`. Should not return match.
- Jexl query with !=: `FOO != '1'`. Should not return match.
- Jexl query with >: `FOO > '1'`. Should return match.
- Jexl query with >=: `FOO >= '1'`. Should return match.
- Jexl query with <: `FOO < '1'`. Should return match.
- Jexl query with <=: `FOO <= '1'`. Should return match.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/FieldsWithNumericRangeValuesVisitorTest.java
```diff
@@ -0,0 +1,146 @@
...
+    private void givenQuery(String query) {
```
lbschanno: Would like to see this renamed to `givenJexlQuery`.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/FieldsWithNumericRangeValuesVisitorTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+    private void givenLuceneQuery(String lucene) {
+        this.luceneQuery = lucene;
+    }
+
+    private void expectFields(String... fields) {
+        this.expectedFields.addAll(List.of(fields));
+    }
+
+    private void assertResult() throws ParseException {
```
lbschanno: Would like to see this renamed to `assertJexlResult`.

warehouse/query-core/src/main/java/datawave/query/rules/NumericValueRule.java
```diff
@@ -43,27 +41,34 @@ public QueryRuleResult validate(QueryValidationConfiguration ruleConfiguration)
 
         try {
             ASTJexlScript jexlScript = (ASTJexlScript) ruleConfig.getParsedQuery();
-            // Fetch the set of fields that have numeric values.
-            Set<String> fields = FieldsWithNumericValuesVisitor.getFields(jexlScript);
-            // If fields with numeric values were found, check the field types.
-            if (!fields.isEmpty()) {
+            if (log.isDebugEnabled()) {
```
lbschanno: We don't need to log the query here. The query itself will already be logged in ShardQueryLogic.validateQuery() before this rule is ever called.

warehouse/query-core/src/main/java/datawave/query/rules/NumericValueRule.java
```diff
@@ -43,27 +41,34 @@ public QueryRuleResult validate(QueryValidationConfiguration ruleConfiguration)
...
+                PrintingVisitor.printQuery(jexlScript);
+            }
+
+            // Identify fields that participate in range comparisons (<, >, <=, >=), including those derived from Lucene ranges.
+            Set<String> fields = FieldsWithNumericRangeValuesVisitor.getFields(jexlScript);
+
+            // If no range fields were found, or no type metadata is available, nothing to validate.
+            TypeMetadata typeMetadata = ruleConfig.getTypeMetadata();
+            if (!fields.isEmpty() && typeMetadata != null) {
```
lbschanno: I don't think we should do nothing if the typeMetadata is null. This rule requires the typeMetadata, and if it's null, something is wrong. Perhaps throw an IllegalStateException if the typeMetadata is null.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/FieldsWithNumericRangeValuesVisitorTest.java
lbschanno: Also would like to see tests where the above cases where a match should be returned do not return any matches if the value is non-numeric.

lbschanno (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldsWithNumericRangeValuesVisitor.java
SethSmucker: My intention with that was to separate concerns, dividing out:
- Identification of a field with a Numeric Range value (anything that could be _interpreted_ as a Numeric Range, even if the field isn't marked as the Numeric type)
- Checking that the field itself is stored in the TypeRegistry as a Number type.

That way we can reuse the chunks / swap them out down the line. Should I still change it?

lbschanno: The first bullet point is what needs to be implemented here. There is no validation here that the value can be interpreted as a number. Verifying whether a field supports the number type should be done at the rule level.

lbschanno: Your latest push has the changes I was looking for here.

lbschanno (approved):

avgAGB (approved):

PR #3130, opened 2025-09-02
changes IncludeExcludeIndexOnlyFieldsRule message output

skyler-kv:
adds a line to message output for user understanding


warehouse/query-core/src/main/java/datawave/query/rules/IncludeExcludeIndexOnlyFieldsRule.java
```diff
@@ -59,7 +59,8 @@ public QueryRuleResult validate(QueryValidationConfiguration ruleConfiguration)
                     // If the function contains any index only fields, add a message to the result.
                     if (!intersection.isEmpty()) {
                         result.addMessage("Index Only fields found within the filter function " + functionFields.getNamespace() + ":"
-                                        + functionFields.getFunction() + ": " + String.join(", ", intersection));
+                                        + functionFields.getFunction() + ": " + String.join(", ", intersection)
```
avgAGB, line 62: just an idea while its in draft, but it might make the message a little easier to parse to have some sort of separation between the intersection and the 'The field(s)' string; sort of like

"found within the filter function filter: A -- The field(s) cannot be used in this manner."

apmoriarty (approved):

lbschanno (approved):

PR #3133, opened 2025-09-03
remove universal set

apmoriarty:
Remove universal set and update filter logic throughout. Fixed numerous discrepancies throughout the codebase due to faulty assumptions.


core/common-util/src/main/java/datawave/core/common/util/TypeFilter.java
```diff
@@ -0,0 +1,109 @@
...
+            return EMPTY;
+        }
+        return Joiner.on(',').join(elements);
+    }
+
+    public static String writeString(Set<String> data) {
+        return Joiner.on(',').join(data);
+    }
+
+    public static TypeFilter fromString(String data) {
```
lbschanno, line 85: This seems to be the only functionality of TypeFilter that is actually used, and generally just in something along the lines of
```java
TypeFilter filter = TypeFilter.fromString(opt);
if (filter.isEmpty()) {
    datatypes = new TreeSet<>();
} else {
    datatypes = new TreeSet<>(filter.getElements());
}
```
I don't see any usages of the filtering functionality, such as `contains()`, `accepts()`, etc outside of the tests. Are you planning to add usages of this class that leverage the filtering functionality? If not, it may make more sense to have a simple utility function somewhere that handles splitting/formatting datatypes to/from a set.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -199,6 +199,7 @@ public class QueryOptions implements OptionDescriber {
 
     public static final String NON_INDEXED_DATATYPES = "non.indexed.dataTypes";
 
+    @Deprecated
     public static final String EVERYTHING = "*";
```
avgAGB: Would it be plausible to scrap the constant entirely? Or maybe circle back to it after confirming that there were no side effects after merging this in - Not too worried either way.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ShardIndexQueryTableStaticMethods.java
```diff
@@ -115,8 +115,8 @@ public static IndexLookup expandQueryTerms(JexlNode node, ShardQueryConfiguratio
     public static IndexLookup expandQueryTerms(String literal, ShardQueryConfiguration config, ScannerFactory scannerFactory, Set<String> expansionFields,
                     MetadataHelper helperRef, ExecutorService execService) throws TableNotFoundException {
         Set<String> terms = Sets.newHashSet(literal);
-        return new FieldNameIndexLookup(config, scannerFactory, getIndexedExpansionFields(expansionFields, false, config.getDatatypeFilter(), helperRef), terms,
-                        execService);
+        Set<String> indexedExpansionFields = getIndexedExpansionFields(expansionFields, false, config.getDatatypeFilter(), helperRef);
```
avgAGB, line 118: neat pull here

core/common-util/src/main/java/datawave/core/common/util/TypeFilter.java
apmoriarty: Instead of replacing every instance of the UniversalSet with a TypeFilter I chose to only use the TypeFilter at the serialization boundary. A simple static utility was the original plan, but the utility lacked context whereas the TypeFilterTest shows someone exactly how it is intended to be used. 

I am planning to add the TypeFilter to several iterators but opted against doing that in this merge request to keep the change set to a reasonable size.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
apmoriarty: After double checking, yes this can be removed

avgAGB (approved):

core/common-util/src/main/java/datawave/core/common/util/TypeFilter.java
lbschanno: Understood, thanks for clarifying.

lbschanno (approved):

PR #3137, opened 2025-09-04
Pad date with 0s when grouping by subset of date

### LINKED ISSUE #3124: Pad date with 0s when grouping by subset of date
The formatting of dates within `TemporalGranularity` should pad the dates with 0.

For example, if my query is #GROUPBY(FOO[TENTH_OF_HOUR]) the result, as it currently is, would be something like: 2025-08-28T09:1

Make it be the format: 2025-08-28T09:10:00.000Z

foster33:
Fixes #3124


warehouse/query-core/src/main/java/datawave/query/attributes/TemporalGranularity.java
```diff
@@ -153,7 +153,9 @@ public String apply(String value) {
                 if (!isTenth) {
                     return formattedDate;
                 } else {
-                    return formattedDate.substring(0, formattedDate.length() - 1);
+                    String truncatedString = formattedDate.substring(0, formattedDate.length() - 1);
```
lbschanno: I think there's an opportunity here for us to refine this  a bit so that we are not locked into a specific position for where the 10th of a value may be. If we made DateTimeValueFormatter accept a post-format function that is in charge of appending on any additional formatting needed after using the SimpleDateFormat, we can remove the need for the isTenth property, and could instead have something along the lines of:
```java
private static class DateTimeValueFormatter implements Function<String,String> {

    private static final Logger log = Logger.getLogger(DateTimeValueFormatter.class);
    private final SimpleDateFormat formatter;
    private final Function<String,String> postFormatFunction;

    private DateTimeValueFormatter(String pattern) {
        this(pattern, Function.identity());
    }

    private DateTimeValueFormatter(String pattern, Function<String,String> postFormatFunction) {
        this.formatter = new SimpleDateFormat(pattern);
        this.postFormatFunction = postFormatFunction;
    }
        
    @Override
    public String apply(String value) {
        try {
            // Attempt to format the denormalized date value.
            Date date = DateNormalizer.DATE_NORMALIZER.denormalize(value);
            String formattedDate = formatter.format(date);
            // Perform any additional formatting required.
            return postFormatFunction.apply(formattedDate);     
```
Where we could add the following function:
```java
/**
 * Replaces the character at the given index in the string with the character '0'. Useful for truncating values in date strings.
 * @param str the string
 * @param index the character index
 * @return the new string
 */
private static String replaceCharWithZero(String str, int index) {
    return str.substring(0, (index - 1)) + "0" + str.substring(index);
}
```
And subsequently update the TRUNCATE_TEMPORAL_TO_TENTH_OF_HOUR instantiation to
```java
TRUNCATE_TEMPORAL_TO_TENTH_OF_HOUR("TENTH_OF_HOUR", new DateTimeValueFormatter("yyyy-MM-dd'T'HH:mm:00.000", (date) -> replaceCharWithZero(date, 16))),
```

lbschanno (changes requested):
Overall looks good, but I think we could change the approach a bit.

lbschanno (dismissed):

avgAGB (dismissed):

warehouse/query-core/src/main/java/datawave/query/attributes/TemporalGranularity.java
```diff
@@ -126,22 +147,35 @@ public String transform(String value) {
...
+     * Replaces the character at the given index in the string with the character '0'. Useful for truncating values in date strings.
+     *
+     * @param str
+     *            the string
+     * @param index
+     *            the character index
+     * @return the new string
+     */
+    private static String replaceCharWithZero(String str, int index) {
+        return str.substring(0, (index - 1)) + "0" + str.substring(index);
```
hlgp: consider adding error handling in case someone provides a index greater than the string length.  Also wondering if using a char array might be more efficient

foster33: Improved in latest commit

lbschanno (approved):

avgAGB (approved):

PR #3144, opened 2025-09-09
DRAFT: Correct Unfielded Index Expansion

apmoriarty:
Add iterators for fielded and unfielded regex expansion. These iterators provide field, date and datatype filtering in a single layer of the iterator stack.

Discovered issue where the UnfieldedRegexExpansionVisitor would delegate lookups to the RegexIndexExpansionVisitor, thus setting field and value thresholds. If either threshold is exceeded during an unfielded lookup there is the potential for missed results.

Updated the index expansion visitors to avoid this error condition and corrected test assertions throughout.

It should be more clear precisely what kind of lookup is created.

Breakout of tasks
- add regex iterators https://github.com/NationalSecurityAgency/datawave/pull/3154
- cleanup timeout iterator https://github.com/NationalSecurityAgency/datawave/pull/3155
- add index expansion unit tests https://github.com/NationalSecurityAgency/datawave/pull/3157
- add new IndexLookups https://github.com/NationalSecurityAgency/datawave/pull/3161
- integrate new IndexLookups https://github.com/NationalSecurityAgency/datawave/pull/3191

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/core/iterators/FieldedRegexExpansionIterator.java
+++ b/warehouse/query-core/src/main/java/datawave/core/iterators/FieldedRegexExpansionIterator.java
@@ -1,0 +1,233 @@
+package datawave.core.iterators;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.apache.accumulo.core.data.ByteSequence;
+import org.apache.accumulo.core.data.Key;
+import org.apache.accumulo.core.data.Range;
+import org.apache.accumulo.core.data.Value;
+import org.apache.accumulo.core.iterators.IteratorEnvironment;
+import org.apache.accumulo.core.iterators.OptionDescriber;
+import org.apache.accumulo.core.iterators.SortedKeyValueIterator;
+import org.apache.accumulo.core.iterators.user.SeekingFilter;
+import org.apache.hadoop.io.Text;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Splitter;
+
+import datawave.query.Constants;
+import datawave.query.data.parsers.ShardIndexKey;
+
+/**
+ * Attempts to expand a fielded regex into discrete values
+ * <p>
+ * Supports date and datatype filtering by default
+ */
+public class FieldedRegexExpansionIterator extends SeekingFilter implements OptionDescriber {
+
+    public static final Logger log = LoggerFactory.getLogger(FieldedRegexExpansionIterator.class);
+
+    public static final String START_DATE = "start.date";
+    public static final String END_DATE = "end.date";
+    public static final String DATATYPES = "dts";
+    public static final String FIELD = "field";
+    public static final String PATTERN = "pattern";
+    public static final String REVERSE = "reverse";
+
+    // TODO: max value threshold, max keys seen threshold?
+
+    private String startDate;
+    private String endDate;
+    private String field;
+    private Pattern pattern;
+    private Matcher matcher;
+    private Set<String> datatypes;
+
+    private boolean reverse = false;
+    private final StringBuilder sb = new StringBuilder();
+
+    private Text columnQualifierDate;
+    private Text columnQualifierDateAndDatatype;
+
+    private final ShardIndexKey parser = new ShardIndexKey();
+
+    enum HINT_TYPE {
+        NONE, FIELD, DATE
+    }
+
+    private HINT_TYPE hint = HINT_TYPE.NONE;
+
+    @Override
+    public void init(SortedKeyValueIterator<Key,Value> source, Map<String,String> options, IteratorEnvironment env) throws IOException {
+
+        if (!validateOptions(options)) {
+            throw new IllegalArgumentException("Iterator was not configured correctly");
+        }
+
+        super.init(source, options, env);
+
+        if (options.containsKey(START_DATE)) {
+            this.startDate = options.get(START_DATE);
+            this.columnQualifierDate = new Text(startDate + "_0");
+        } else {
+            throw new IllegalArgumentException("Iterator requires START_DATE option");
+        }
+
+        if (options.containsKey(END_DATE)) {
+            this.endDate = options.get(END_DATE) + Constants.MAX_UNICODE_STRING;
+        } else {
+            throw new IllegalArgumentException("Iterator requires END_DATE option");
+        }
+
+        if (options.containsKey(FIELD)) {
+            this.field = options.get(FIELD);
+        } else {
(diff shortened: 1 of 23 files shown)


warehouse/query-core/src/test/java/datawave/query/AnyFieldQueryTest.java
```diff
@@ -548,21 +551,24 @@ public void testReverseIndexMissesRemovedIndexEntries() throws Exception {
         // Test the plan with all expansions
         // test running the query
         String expect = this.dataManager.convertAnyField(phrase);
-        runTest(query, expect);
-
-        // remove the metadata entries
-        Multimap<String,KeyValue> metadata = removeMetadataEntries(JexlASTHelper.getIdentifierNames(JexlASTHelper.parseJexlQuery(expect)),
-                        ColumnFamilyConstants.COLF_RI);
+        // runTest(query, expect);
```
foster33, line 554:
```suggestion
```
Can this be removed? Or should it be uncommented?

warehouse/query-core/src/main/java/datawave/core/iterators/FieldedRegexExpansionIterator.java
```diff
@@ -0,0 +1,233 @@
...
+        if (!matcher.matches()) {
+            // advance to next row
+            log.info("pattern does not match, advance to next row");
+            return new FilterResult(false, AdvanceResult.NEXT_ROW);
+        }
+
+        int result = parser.getField().compareTo(field);
+        if (result < 0) {
+            // advance to field
+            log.info("field {} sorts before target {}, advance to {}", parser.getField(), field, field);
```
lbschanno, line 167: I think this sort of logging (along with the rest in the class) would be better done at the DEBUG or TRACE level.

warehouse/query-core/src/main/java/datawave/core/iterators/UnfieldedRegexExpansionIterator.java
```diff
@@ -0,0 +1,227 @@
...
+                super.seek(skipRange, columnFamilies, inclusive);
+            }
+        } else {
+            super.seek(range, columnFamilies, inclusive);
+        }
+    }
+
+    @Override
+    public FilterResult filter(Key k, Value v) {
+        if (log.isInfoEnabled()) {
```
lbschanno, line 136: Similar suggestion here that perhaps DEBUG or TRACE level would be more appropriate.

warehouse/query-core/src/main/java/datawave/core/iterators/TimeoutExceptionIterator.java
```diff
@@ -107,21 +101,23 @@ public void next() throws IOException {
         } catch (IteratorTimeoutException e) {
             setReturnKey();
         } catch (RuntimeException e) {
-            if (e.getCause() instanceof IteratorTimeoutException)
+            if (e.getCause() instanceof IteratorTimeoutException) {
                 setReturnKey();
-            else
+            } else {
```
lbschanno, line 106: Love to see explicit bracing :tada:

warehouse/query-core/src/main/java/datawave/core/iterators/FieldedRegexExpansionIterator.java
```diff
@@ -0,0 +1,233 @@
...
+    public static final String PATTERN = "pattern";
+    public static final String REVERSE = "reverse";
+
+    // TODO: max value threshold, max keys seen threshold?
+
+    private String startDate;
+    private String endDate;
+    private String field;
+    private Pattern pattern;
+    private Matcher matcher;
```
lbschanno, line 53: This could be a local variable, I don't see `matcher` being used anywhere outside of the filter function.

warehouse/query-core/src/main/java/datawave/core/iterators/UnfieldedRegexExpansionIterator.java
```diff
@@ -0,0 +1,227 @@
...
+    public static final String PATTERN = "pattern";
+    public static final String REVERSE = "reverse";
+
+    private String startDate;
+    private String endDate;
+    private Pattern pattern;
+    private boolean reverse;
+    private final StringBuilder sb = new StringBuilder();
+
+    private Matcher matcher;
```
lbschanno, line 53: `matcher` here can also be a local variable.

warehouse/query-core/src/main/java/datawave/core/iterators/FieldedRegexExpansionIterator.java
```diff
@@ -0,0 +1,233 @@
...
+
+        // parse key and reset hint
+        parser.parse(k);
+        hint = HINT_TYPE.NONE;
+
+        if (reverse) {
+            sb.setLength(0);
+            sb.append(parser.getValue());
+            sb.reverse();
+            matcher = pattern.matcher(sb);
```
lbschanno, line 153: Would it be worth it at all to have a cache of previously evaluated strings to do a quick check if we've already run the string against the matcher and know it matches/doesn't match? Could make the cache options configurable via the iterator options to tweak memory usage.

warehouse/query-core/src/main/java/datawave/core/iterators/UnfieldedRegexExpansionIterator.java
```diff
@@ -0,0 +1,227 @@
...
+
+        // parse key and reset hint
+        parser.parse(k);
+        hint = HINT_TYPE.NONE;
+
+        if (reverse) {
+            sb.setLength(0);
+            sb.append(parser.getValue());
+            sb.reverse();
+            matcher = pattern.matcher(sb);
```
lbschanno, line 148: Similar question here about the viability of having a cache to check before calling `pattern.matcher()`.

lbschanno (changes requested):
Nice test coverage. Some minor changes requested and one question re. caching as food for thought.

warehouse/query-core/src/main/java/datawave/core/iterators/FieldedRegexExpansionIterator.java
apmoriarty: All logging will be moved to trace or debug

warehouse/query-core/src/test/java/datawave/query/AnyFieldQueryTest.java
apmoriarty: That is uncommented in a following commit, but then that commit grew quite large, and now I'm going to break this into smaller, mor manageable units of work

warehouse/query-core/src/main/java/datawave/core/iterators/FieldedRegexExpansionIterator.java
apmoriarty: Hmm, I really like this idea...due to the structure of the inverted index we don't even need a cache, we just need to record the previously matched value. I'm going to break this merge request up -- and add a previously evaluated value check.

lbschanno: Sweet!

apmoriarty:
All work was broken out into smaller chunks and should now be merged.

PR #3149, opened 2025-09-09
Bugfix/root pointer

ivakegg:
This fixes a bug where more than the root / TLD document was being returned to the user with a luceneUUIDLookup.  The problem was that the UID of the document in question was actually smaller than normal which broke some optimizations.  Now that the event data filters can maintain context of the document, I rewrote this to more efficiently determine whether a key is for a root document or not.


FineAndDandy (dismissed):

warehouse/query-core/src/main/java/datawave/query/predicate/RootPointerPredicate.java
```diff
@@ -0,0 +1,149 @@
...
+     *            The shard table key in question
+     * @return true if for the root document
+     */
+    private boolean isRootPointerWithContext(Key k) {
+        ByteSequence cf = k.getColumnFamilyData();
+
+        if (WritableComparator.compareBytes(cf.getBackingArray(), 0, 2, FI_CF, 0, 2) == 0) {
+            ByteSequence seq = k.getColumnQualifierData();
+
+            if (seq.length() <= dtAndUid.getLength()) {
```
jack-gitdev, line 52: Should this method check if a document has been provided? If someone forgets to call startNewDocument first then this dtAndUid.getLength call will throw a NPE.

jack-gitdev: My bad, just realized this is caught in apply.

jack-gitdev (dismissed):

warehouse/query-core/src/main/java/datawave/query/predicate/RootPointerPredicate.java
```diff
@@ -0,0 +1,149 @@
...
+import org.apache.accumulo.core.data.ByteSequence;
+import org.apache.accumulo.core.data.Key;
+import org.apache.hadoop.io.Text;
+import org.apache.hadoop.io.WritableComparator;
+
+import com.google.common.base.Predicate;
+
+import datawave.query.Constants;
+
+public class RootPointerPredicate implements Predicate<Key> {
```
apmoriarty, line 18: Add class level docs explaining why this is necessary

warehouse/query-core/src/test/java/datawave/query/predicate/IsRootPointerPredicateTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.util.Random;
+
+import org.apache.accumulo.core.data.Key;
+import org.junit.Test;
+
+import datawave.data.hash.HashUID;
+import datawave.data.hash.UID;
+
+public class IsRootPointerPredicateTest {
```
apmoriarty: Test name should match the class it is testing, i.e. "RootPointerPredicateTest". 

For extra clarity the tests could be broken out into more explicit names like 
testEventKeyWithContext
testTermFrequencyKeyWithContext
testFieldIndexKeyWithContext
...and then the same sans/without context

Please add a test that demonstrates the failure case for a uid that is too short.

ivakegg: done

warehouse/query-core/src/main/java/datawave/query/predicate/RootPointerPredicate.java
ivakegg: done

apmoriarty (approved):

jack-gitdev (approved):

FineAndDandy (approved):

PR #3150, opened 2025-09-10
Accumulo scan hints and consistency levels

rdhayes68:
This is a conservative list of scanners changed with the future ability to use ScanServers (consistency level set to EVENTUAL).


hlgp:
reopen when ready for integration testing

PR #3151, opened 2025-09-10
Accumulo scan hints and consistency levels aggressive

rdhayes68:
This is a more aggressive list of scanners changed with the future ability to use ScanServers (consistency level set to EVENTUAL) when compared to [Accumulo scan hints and consistency levels #3150.](https://github.com/NationalSecurityAgency/datawave/pull/3150)


hlgp:
reopen when ready for integration testing

PR #3156, opened 2025-09-12
fixes #3152: Updated MostRecentUnique feature to handle TLD documents

### LINKED ISSUE #3152: Most Recent Unique feature not working with TLDEventQuery
When documents from the TLDEventQuery come back, then have multiple document ids in the Document and hence an Attributes value.  This fails to be casted to a DocumentKey at line 573 of the UniqueTransform.getDocKeyAttr method.

ivakegg:


warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -569,12 +568,25 @@ private static long getTimestamp(Document doc) {
         return getDocKeyAttr(doc).getTimestamp();
     }
 
-    private static DocumentKey getDocKeyAttr(Document doc) {
-        return (DocumentKey) (doc.get(Document.DOCKEY_FIELD_NAME));
+    private static Attribute getDocKeyAttr(Document doc) {
+        Attribute<?> attr = doc.get(Document.DOCKEY_FIELD_NAME);
+        if (attr instanceof Attributes) {
+            // if the attr is an instanceof Attributes, then we need to frind the one that best describes
```
foster33:
```suggestion
            // if the attr is an instanceof Attributes, then we need to find the one that best describes
```

hlgp (dismissed):

FineAndDandy (dismissed):

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -569,12 +568,25 @@ private static long getTimestamp(Document doc) {
         return getDocKeyAttr(doc).getTimestamp();
     }
 
-    private static DocumentKey getDocKeyAttr(Document doc) {
-        return (DocumentKey) (doc.get(Document.DOCKEY_FIELD_NAME));
+    private static Attribute getDocKeyAttr(Document doc) {
```
apmoriarty: Consider renaming to `getRootDocumentKeyAttr`. 

Add a method comment.

warehouse/query-core/src/test/java/datawave/query/transformer/UniqueTransformTest.java
```diff
@@ -610,11 +610,17 @@ protected class InputDocumentBuilder {
...
             Key key = new Key("shardid", "datatype\u0000" + getUid(), cq, ts);
+            Key key2 = new Key("shardid", "datatype\u0000" + getUid() + ".1", cq, ts);
+            Key key3 = new Key("shardid", "datatype\u0000" + getUid() + ".5", cq, ts);
             this.document = new Document(key, true);
             inputDocuments.add(document);
             this.document.getMetadata().set(key);
             Attribute<?> docKeyAttributes = new DocumentKey(key, true);
+            Attribute<?> docKeyAttributes2 = new DocumentKey(key2, true);
+            Attribute<?> docKeyAttributes3 = new DocumentKey(key3, true);
             this.document.put(Document.DOCKEY_FIELD_NAME, docKeyAttributes);
```
apmoriarty, line 617: I know from experience that this addition to the test data will exercise the new code that was added, but it would be better to open up the `getDocKey` methods from private to protected and add a test case that exercises TLD and child keys.

apmoriarty (changes requested):
Method docs and a test are still needed. Looks good otherwise.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
ivakegg: ok

warehouse/query-core/src/test/java/datawave/query/transformer/UniqueTransformTest.java
ivakegg: done

apmoriarty (approved):

d-hwang (approved):

PR #3161, opened 2025-09-16
Add new IndexLookups

apmoriarty:
Add IndexLookups for fielded and unfielded regex expansion iterators, add test framework for IndexLookups

IndexLookupMap now has variables to track state. This lays the groundwork for migrating the control flow away from thrown exceptions

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/lookups/UnfieldedRegexIndexLookup.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/lookups/UnfieldedRegexIndexLookup.java
@@ -1,12 +1,9 @@
 package datawave.query.jexl.lookups;
 
-import static datawave.query.jexl.lookups.ShardIndexQueryTableStaticMethods.EXPANSION_HINT_KEY;
-
 import java.util.Collections;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
-import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 
 import org.apache.accumulo.core.client.IteratorSetting;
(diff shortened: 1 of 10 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/lookups/UnfieldedRegexIndexLookup.java
```diff
@@ -0,0 +1,145 @@
...
+import datawave.query.config.ShardQueryConfiguration;
+import datawave.query.tables.ScannerFactory;
+import datawave.util.time.DateHelper;
+
+/**
+ * An asynchronous index lookup which expands a fielded regex into discrete fields and values.
+ * <p>
+ * Because an unfielded term is not executable it is best if this index lookup runs without a time or field threshold.
+ */
+public class UnfieldedRegexIndexLookup extends AsyncIndexLookup {
```
ivakegg: Code is very similar to the fielded regex index lookup.... can we combine somehow or is this really needed?

ivakegg (changes requested):
see comment

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/UnfieldedRegexIndexLookup.java
apmoriarty: These look very similar today, but the unfielded lookup is configured to ignore timeout exceptions...which begs the question, why bother configuring a timeout iterator in the first place? 

That will get cleaned up when the new index lookups are integrated. At that point keeping these as separate classes preserves the intent.

ivakegg: Why not simply use the same class but configure a different timeout ?

ivakegg (approved):

avgAGB (approved):

PR #3170, opened 2025-09-18
Accumulo annotation serializer, data access object and utilities.

drewfarris:
This should be considered a draft until feature/annotation-base-model is branch is finalized in #3169, putting this here so @FineAndDandy can begin reviewing and comment on it.

PR #3173, opened 2025-09-19
Annotation manager web service

drewfarris:
This builds on the framework in #3170 to provide a web-services layer for managing annotations and performing general CRUD operations.


warehouse/ingest-core/pom.xml
drewfarris, line 1: `io.dropwizard.metrics:metrics-core` is a duplicate dependency declared in both the default and provided scope. I removed this to get rid of warnings.

apmoriarty (approved):

web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
```diff
@@ -0,0 +1,550 @@
...
+     *            the type of id provided
+     * @param id
+     *            the id itself.
+     * @return a Metadata object containing the source table, row, datatype and internal document id (uid).
+     * @throws QueryException
+     *             if the id is malformed.
+     */
+    private Metadata lookupDocumentIdentifier(String idType, String id) throws QueryException {
+        // TODO: if the idType is CONTENT or DOCUMENT, assume that the id provided is an internal id, otherwise
+        if (idType.equals("DOCUMENT")) {
```
ivakegg, line 495: We may be overloading too much here.  Should this be "RECORD_ID" which matches the field returned in documents ?

web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
```diff
@@ -0,0 +1,550 @@
...
+
+    @Inject
+    @SpringBean(name = "AnnotationManagerConfig")
+    private AnnotationManagerConfig config;
+
+    // Stateful variables, managed locally.
+    Set<Authorizations> authorizations;
+    AccumuloClient client;
+    LookupUUIDService lookupUUIDService;
+    AnnotationDataAccess annotationDataAccess;
```
ivakegg, line 91: You marked this bean as stateless above, and yet you have state here.  Did you intend these to be static maybe?  Some synchronized access around these may be required.  Or did you intend these to be used on a call by call basis in which case remove these and instead mange them inside of the rest full methods.  Could make the thread local as another possibility.

web-services/pom.xml
```diff
@@ -556,7 +562,7 @@
                         <threadCountClasses>0</threadCountClasses>
                         <threadCountMethods>1</threadCountMethods>
                         <threadCountSuites>0</threadCountSuites>
-                        <argLine>@{argLine} -XX:+TieredCompilation -XX:TieredStopAtLevel=1  -Dfile.encoding=UTF8 -Duser.timezone=GMT -Xmx1024m -Dapple.awt.UIElement=true -Djava.security.krb5.realm= -Djava.security.krb5.kdc=
+                        <argLine>@{argLine} -XX:+TieredCompilation -XX:TieredStopAtLevel=1  -Dfile.encoding=UTF8 -Duser.timezone=GMT -Xmx2g -Dapple.awt.UIElement=true -Djava.security.krb5.realm= -Djava.security.krb5.kdc=
```
drewfarris, line 565: Added #3248 to explain why a 2g heap size is needed for these unit tests.

jalphonso (approved):

warehouse/annotation-core/src/main/java/datawave/annotation/util/v1/AnnotationValidators.java
```diff
@@ -0,0 +1,32 @@
...
+import datawave.annotation.protobuf.v1.Segment;
+import datawave.annotation.util.Validator;
+
+/** Utility class holding validator implementations for Annotation and Segment classes */
+public class AnnotationValidators {
+    //@formatter:off
+    public static final Validator<Annotation> annotationValidator = Validator.<Annotation>create()
+            .addCheck(a -> StringUtils.isNotBlank(a.getShard()), "Annotation shard must not be blank")
+            .addCheck(a -> StringUtils.isNotBlank(a.getDataType()), "Annotation datatype must not be blank")
+            .addCheck(a -> StringUtils.isNotBlank(a.getUid()), "Annotation uid must not be blank")
```
austin007008, line 15: if you want to stay consistent with the checks below this line, Java has a built in "isBlank" method you can use. So instead of "StringUtils.isNotBlank(a.getShard())" it would be "!a.getShard().isBlank".   And then the stringUtils import can be removed if that change is made.

web-services/annotations/pom.xml
```diff
@@ -0,0 +1,237 @@
...
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.picketbox</groupId>
+            <artifactId>picketbox</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-api-easymock</artifactId>
```
austin007008, line 195: Just making a note that powermock is not compatable with Java17+ so tests written in it will need to be updated whenever Datawave is moved off of Java11

warehouse/annotation-core/src/main/java/datawave/annotation/util/v1/AnnotationValidators.java
drewfarris: Thanks for the suggestion. The nice thing about `StringUtils.isNotBlank` here is that it can be used for a null check in addition to a blank string. If we were to use the built in String method, we'd possibly need a null check as well, which would make this code more verbose. However, with Annotation being a protobuf, I believe (but haven't verified) that the String fields are never null.

web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
drewfarris: Ahh, this is a good point - I had used "DOCUMENT" as the field name here to be consistent with the query form used by the [ContentQueryLogic](https://github.com/NationalSecurityAgency/datawave/blob/344e5e26f29143bea8fd552aa61a488be9e91aed/warehouse/query-core/src/main/java/datawave/query/tables/content/ContentQueryLogic.java#L218)

web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
drewfarris: I should clarify that comment because these don't preserve state across requests, but they hold state that is local to a specific, individual request. As a result, they aren't intended to violate the Stateless-ness of the class. How's something like `// Per-request stateful variables, managed internally` instead? Also, there's no reason they shouldn't be private.

PR #3174, opened 2025-09-19
Make /validate check for field existence in date range

foster33:
Updates FieldExistenceRule to also verify that there is data in a field for the given date range.

In the case of an OR node, ALL fields must be missing from the date range for a failure message to be returned.


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldMissingFromDateRangeVisitor.java
```diff
@@ -0,0 +1,249 @@
...
+
+/**
+ * Class to check that each query node contains a field which exists in the schema.
+ *
+ * <pre>
+ * 1. If a datatype filter was specified, then the existence check is limited to only those datatypes
+ * 2. If a datatype filter is NOT specified (null or empty), this implies ALL datatypes.
+ * </pre>
+ */
+public class FieldMissingFromDateRangeVisitor extends ShortCircuitBaseVisitor {
```
lbschanno, line 42: At first glance, I don't see any glaring issues, but I would like to see some tests for this class.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldMissingFromDateRangeVisitor.java
```diff
@@ -0,0 +1,249 @@
...
+
+    // Descend through these nodes
+    @Override
+    public Object visit(ASTJexlScript node, Object data) {
+        node.childrenAccept(this, data);
+        return data;
+    }
+
+    @Override
+    public Object visit(ASTAndNode node, Object data) {
```
lbschanno: This is identical to the super method, and can be removed.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldMissingFromDateRangeVisitor.java
```diff
@@ -0,0 +1,249 @@
...
+        node.childrenAccept(this, data);
+        return data;
+    }
+
+    @Override
+    public Object visit(ASTNotNode node, Object data) {
```
lbschanno: Also identical to super, can be removed.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldMissingFromDateRangeVisitor.java
```diff
@@ -0,0 +1,249 @@
...
+        node.childrenAccept(this, data);
+        return data;
+    }
+
+    @Override
+    public Object visit(ASTReference node, Object data) {
```
lbschanno: Also identical to super, can be removed.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldMissingFromDateRangeVisitor.java
```diff
@@ -0,0 +1,249 @@
...
+        node.childrenAccept(this, data);
+        return data;
+    }
+
+    @Override
+    public Object visit(ASTReferenceExpression node, Object data) {
```
lbschanno: Also identical to super, can be removed.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldMissingFromDateRangeVisitor.java
```diff
@@ -0,0 +1,249 @@
...
+    /**
+     * @param node
+     *            Jexl node
+     * @param data
+     *            The set of names which we have determined do not exist
+     * @return the updated set of names which do not exist
+     */
+    protected Object ingestDateOrNodeVisit(JexlNode node, Object data) {
+        @SuppressWarnings("unchecked")
+        Set<String> nonExistentFieldNames = (null == data) ? new HashSet<>() : (Set<String>) data;
```
lbschanno: Suggest initializing nonExistentFieldNames as a LinkedHashSet to be consistent with `getNonIngestFields()`.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldMissingFromDateRangeVisitor.java
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    /**
+     * @param node
+     *            Jexl node
+     * @param data
+     *            The set of names which we have determined do not exist
+     * @return the updated set of names which do not exist
+     */
+    protected Object ingestDateOrNodeVisit(JexlNode node, Object data) {
```
lbschanno: I would prefer a more descriptive method name here, such as `findMissingFields(...)`. If you change the method signature here to `findMissingFields(ASTOrNode node, Object data`, you can also rename the function `genericIngestDateVisit(...)` to `findMissingFields(...)`.

lbschanno: Additionally, this method and `genericIngestDateVisit` can both be private, unless you are explicitly planning for this class to be extended.

lbschanno (changes requested):

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/FieldMissingFromDateRangeVisitorTest.java
```diff
@@ -0,0 +1,307 @@
...
+    private final Set<String> specialFields = Sets.newHashSet(ANY_FIELD, NO_FIELD);
+    private final List<Mutation> mutations = new ArrayList<>();
+
+    private static final String TABLE_METADATA = "metadata";
+    private static final String[] AUTHS = {"FOO", "BAR"};
+    private static final Set<Authorizations> AUTHS_SET = Collections.singleton(new Authorizations(AUTHS));
+    private static final String NULL_BYTE = "\0";
+
+    @BeforeEach
+    public void setUp() throws Exception {
```
lbschanno: You can avoid the overhead of using the InMemoryAccumuloClient here by either passing in a mocked MetadataHelper or an instance of datawave.query.util.MockMetadataHelper. That way your tests do not care about how MetadatHelper is implemented.

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/job/reindex/ShardReindexMapperTest.java
```diff
@@ -76,8 +76,8 @@
 import datawave.ingest.protobuf.TermWeight;
 import datawave.iterators.FrequencyMetadataAggregator;
 import datawave.query.iterator.SortedListKeyValueIterator;
-import datawave.util.accumulo.RFileUtil;
 import datawave.query.model.DateFrequencyMap;
+import datawave.util.accumulo.RFileUtil;
```
lbschanno, line 80: Why this change?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldMissingFromDateRangeVisitor.java
```diff
@@ -0,0 +1,221 @@
...
+        if (identifiers.isEmpty()) {
+            // Catch cases where we have two literals
+            // essentially everything but identifier op literal
+            return nonIngestedFieldNames;
+        }
+
+        for (ASTIdentifier identifier : identifiers) {
+            String fieldName = JexlASTHelper.deconstructIdentifier(identifier);
+            long occurrences = this.querySettings != null
+                            ? helper.getCountsByFieldForDays(fieldName, this.querySettings.getBeginDate(), this.querySettings.getEndDate(), this.datatypeFilter)
```
lbschanno: You mentioned some concerns with how long this rule might take depending on how wide of a date range a query is targeting. I think there are a couple things we can do to mitigate this:

1. I notice that we can have multiple calls to `MetadataHelper.getCountsByFieldForDays()` and `MetadataHelper.getCountsForFieldsInDateRange()`, which will be the most time-consuming part of this visitor. We could defer a call to MetadataHelper until we have collected all of the unique fields in the query, and then make one call to MetadataHelper to get the total occurrences.
2. If we have a single call to MetadataHelper as suggested above, we can easily submit it as a task to an ExecutorService with a timeout that is both an argument passed into the visitor and a configurable property of the rule. You will likely want to discuss with interested parties what the default timeout should be based on the average performance of this rule. While a timeout would effectively cause the validation to fail, at least it would keep the validation process from taking excessively long.
3. Rather than using `MetadataHelper.getCountsForFieldsInDateRange()`, we could add a new method to MetadataHelper, such as `MetadataHelper.getMissingFields(Set<String> fields, Set<String> datatypes, String beginDate, String endDate)` that will return any fields that do not have a frequency row in the metadata table with a non-zero value. If a frequency row is guaranteed to always have a non-zero value, then we can simply check for the existence of a frequency row within the date range for each field. This would be faster than `MetadataHelper.getCountsForFieldsInDateRange()` because we would avoid the overhead of summing up the frequency counts for each field.

lbschanno (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldMissingFromDateRangeVisitor.java
foster33: I went with option 3 in latest commit.

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/job/reindex/ShardReindexMapperTest.java
foster33: My branch was out of sync with integration, so it did not get the commit that fixed this formatting. Should be resolved now.

core/utils/metadata-utils/src/main/java/datawave/query/util/MetadataHelper.java
```diff
@@ -1692,6 +1692,68 @@ public Map<String,Long> getCountsForFieldsInDateRange(Set<String> fields, Set<St
...
+            bs.fetchColumnFamily(ColumnFamilyConstants.COLF_F);
+
+            IteratorSetting setting = new IteratorSetting(50, "MetadataFrequencySeekingIterator", MetadataFColumnSeekingFilter.class);
+            setting.addOption(MetadataFColumnSeekingFilter.DATATYPES_OPT, Joiner.on(',').join(sortedDatatypes));
+            setting.addOption(MetadataFColumnSeekingFilter.START_DATE, beginDate);
+            setting.addOption(MetadataFColumnSeekingFilter.END_DATE, endDate);
+            bs.addScanIterator(setting);
+
+            for (Entry<Key,Value> entry : bs) {
+                isMissing = false;
```
lbschanno: I think this function would be more useful if it returned the set of fields that are missing. We could then indicate to the user which specific fields are missing in the date range in the rule response.

lbschanno (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldMissingFromDateRangeVisitor.java
lbschanno: I think we can combine option 3 with option 1, and collect all of the fields before making just one call to metadata helper to reduce our overhead.

core/utils/metadata-utils/src/main/java/datawave/query/util/MetadataHelper.java
foster33: Addressed in latest commit

core/utils/metadata-utils/src/main/java/datawave/query/util/MetadataHelper.java
```diff
@@ -1692,6 +1692,73 @@ public Map<String,Long> getCountsForFieldsInDateRange(Set<String> fields, Set<St
...
+            bs.addScanIterator(regexIter);
+
+            for (Entry<Key,Value> entry : bs) {
+                try {
+                    DateFrequencyMap map = new DateFrequencyMap(entry.getValue().get());
+                    if (!map.subMap(beginDate, endDate).isEmpty()) {
+                        foundFields.add(entry.getKey().getRow().toString());
+                    }
+                } catch (IOException e) {
+                    log.trace("Could not convert the Value to a DateFrequencyMap: {}", entry.getValue());
```
lbschanno: I don't think you need the trace message here, the error log message should be sufficient.

core/utils/metadata-utils/src/main/java/datawave/query/util/MetadataHelper.java
```diff
@@ -1692,6 +1692,73 @@ public Map<String,Long> getCountsForFieldsInDateRange(Set<String> fields, Set<St
...
+     *            the datatypes
+     * @param beginDate
+     *            the start date
+     * @param endDate
+     *            the end date
+     * @param specialFields
+     *            special fields to exclude from search
+     * @return a set of missing fields from the given date range
+     */
+    public Set<String> getMissingFieldsInDateRange(Set<String> fields, Set<String> datatypes, String beginDate, String endDate, Set<String> specialFields)
```
lbschanno: Will you be adding any tests for this method? You can review the tests for `getCardinalityForField()` in `MetadataHelperTest` for examples of convenience methods used to generate 'f' rows in both aggregated and non-aggregated form.

core/utils/metadata-utils/src/main/java/datawave/query/util/MetadataHelper.java
```diff
@@ -1692,6 +1692,73 @@ public Map<String,Long> getCountsForFieldsInDateRange(Set<String> fields, Set<St
...
+            IteratorSetting regexIter = new IteratorSetting(50, "regexFilter", RegExFilter.class);
+            if (!dataTypeRegex.toString().isEmpty()) {
+                regexIter.addOption(RegExFilter.COLQ_REGEX, dataTypeRegex.toString());
+            }
+            bs.setRanges(ranges);
+            bs.addScanIterator(regexIter);
+
+            for (Entry<Key,Value> entry : bs) {
+                try {
+                    DateFrequencyMap map = new DateFrequencyMap(entry.getValue().get());
```
lbschanno: It is not guaranteed that all 'f' rows will have an aggregated value. It's possible for there to be rows that have not yet been aggregated, and thus have long values. See `MetadataHelper.getCardinalityForField(String fieldName, String datatype, Date begin, Date end)` for an example of how to determine if we're looking at an aggregated entry vs. a single-value one.

core/utils/metadata-utils/src/main/java/datawave/query/util/MetadataHelper.java
```diff
@@ -1692,6 +1692,73 @@ public Map<String,Long> getCountsForFieldsInDateRange(Set<String> fields, Set<St
...
+            }
+        }
+
+        AccumuloClient client = accumuloClient;
+        if (client instanceof WrappedAccumuloClient) {
+            client = ((WrappedAccumuloClient) client).getReal();
+        }
+
+        try (BatchScanner bs = ScannerHelper.createBatchScanner(client, getMetadataTableName(), getAuths(), fields.size())) {
+            IteratorSetting regexIter = new IteratorSetting(50, "regexFilter", RegExFilter.class);
```
lbschanno: A further optimization we could do here is to also use the FirstEntryInRowIterator in our iterator stack. This would return just one entry per row per tablet, reducing the number of entries to iterate over.

lbschanno (changes requested):

core/utils/metadata-utils/src/main/java/datawave/query/util/MetadataHelper.java
```diff
@@ -1692,6 +1694,109 @@ public Map<String,Long> getCountsForFieldsInDateRange(Set<String> fields, Set<St
...
+        if (client instanceof WrappedAccumuloClient) {
+            client = ((WrappedAccumuloClient) client).getReal();
+        }
+
+        try (BatchScanner bs = ScannerHelper.createBatchScanner(client, getMetadataTableName(), getAuths(), fields.size())) {
+            if (!datatypes.isEmpty()) {
+                settings.add(new IteratorSetting(50, "regexFilter", RegExFilter.class));
+            }
+            settings.add(new IteratorSetting(51, "firstEntryInRow", FirstEntryInRowIterator.class));
+            for (IteratorSetting setting : settings) {
```
lbschanno: Last change request, I (as much as I can) promise. Instead of looping over the settings to configure them, we can configure them directly after instantiation before adding them to the stack, like so:
```java
if (!datatypes.isEmpty()) {
    IteratorSetting regexFilter = new IteratorSetting(50, "regexFilter", RegExFilter.class);
    regexFilter.addOption(RegExFilter.COLQ_REGEX, dataTypeRegex.toString());
    settings.add(regexFilter);
}
```
We can do the same for FirstEntryInRowIterator, and then this loop will no longer be needed.

lbschanno (changes requested):

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldMissingFromDateRangeVisitor.java
```diff
@@ -0,0 +1,262 @@
...
+        if (identifiers.isEmpty()) {
+            // Catch cases where we have two literals
+            // essentially everything but identifier op literal
+            return nonIngestedFieldNames;
+        }
+
+        for (ASTIdentifier identifier : identifiers) {
+            String fieldName = JexlASTHelper.deconstructIdentifier(identifier);
+            if (!specialFields.contains(fieldName)) {
+                nonIngestedFieldNames.addAll(helper.getMissingFieldsInDateRange(Set.of(fieldName), datatypeFilter,
```
ivakegg: Is it worth trying to gather up the fields before we call getMissingFieldsInDateRange?  For very large queries where we repeat the same field hundreds of times, this could get quite expensive.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldMissingFromDateRangeVisitor.java
```diff
@@ -0,0 +1,262 @@
...
+        }
+        this.datatypeFilter = datatypeFilter;
+    }
+
+    @SuppressWarnings("unchecked")
+    public static Set<String> getNonIngestedFields(MetadataHelper helper, ASTJexlScript script, Set<String> datatypes, Set<String> specialFields,
+                    Query querySettings) {
+        FieldMissingFromDateRangeVisitor visitor = new FieldMissingFromDateRangeVisitor(helper, datatypes, specialFields, querySettings);
+        // Maintain insertion order.
+        return (Set<String>) script.jjtAccept(visitor, new LinkedHashSet<>());
```
apmoriarty: Even though the MetadataHelper will cache operations it is probably better to have the visitor collect all fields and then do a single check at the end. So this code block would become:
```java
private final Set<String> fields = new HashSet<>()

public static Set<String> getNonIngestedFields(...) {
    script.accept(visitor, null)
    Set<String> missingFields = checkForMissingFields()
    if (log.isDebugEnabled() && !missingFields.isEmpty()) {
        log.debug("Found missing fields:  {}", missingFields)
    }
    return missingFields;
```
One benefit of collecting all fields first and then checking once at the end is that you only have to create a single SimpleDateFormatter.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/FieldMissingFromDateRangeVisitorTest.java
```diff
@@ -0,0 +1,454 @@
...
+    /**
+     * Test query with ORed fields but ALL do not exist during the date range. Fields will only be returned as NonIngested if ALL ORed fields are not found.
+     */
+    @Test
+    public void testWithOrFieldsThatAllDoNotExist()
+                    throws ParseException, AccumuloException, TableExistsException, AccumuloSecurityException, TableNotFoundException {
+        String query = "AGE == 'foo' || GENDER == 'bar' || JOB == 'foo'";
+        ASTJexlScript script = JexlASTHelper.parseJexlQuery(query);
+
+        Date begin = new Date(1546214400000L); // 12/31/2018 00:00:00 GMT
```
apmoriarty: Consider adding a helper method that converts a `yyyyMMdd hh:mm:ss` format into a date?

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/FieldMissingFromDateRangeVisitorTest.java
```diff
@@ -0,0 +1,454 @@
...
+import com.google.common.collect.Sets;
+
+import datawave.microservice.query.Query;
+import datawave.query.jexl.JexlASTHelper;
+import datawave.query.model.DateFrequencyMap;
+import datawave.query.util.MockMetadataHelper;
+import datawave.util.TableName;
+import datawave.util.time.DateHelper;
+
+public class FieldMissingFromDateRangeVisitorTest {
```
apmoriarty, line 26: Tests look really good, can you add a few more cases?
```
# nested intersection
A || (B && C)

# nested union
A && (B || C)

# double nested intersection
(A && B) || (C && D)

# double nested union
(A || B) && (C || D)

# and a jexl arithmetic case
UUID == 'CORLEONE' AND filter:getAllMatches(NAM,'hubert').size() == 0
```

lbschanno (dismissed):

lbschanno (approved):

avgAGB (approved):

PR #3192, opened 2025-09-22
add wildcard processing and check for EMPTY_RESULTS instead of null and add unit test

austin007008:
add wildcard processing and check for EMPTY_RESULTS instead of null and add unit test

### DIFF
--- a/warehouse/keyword-common/src/main/java/datawave/util/keyword/KeywordExtractor.java
+++ b/warehouse/keyword-common/src/main/java/datawave/util/keyword/KeywordExtractor.java
@@ -4,6 +4,8 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
+import javax.annotation.Nonnull;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


warehouse/keyword-common/src/main/java/datawave/util/keyword/KeywordExtractor.java
```diff
@@ -99,23 +102,37 @@ public void parseOptions(Map<String,String> iteratorOptions) {
         }
     }
 
+    @Nonnull
```
drewfarris, line 113: Consider adding the javadoc:
```suggestion
    /**
     * Extract keywords from the <code>foundContent</code> field in priority order based on the list of views in <code>preferredViews</code> field. Will return
     * a KeywordResults object populated with the keywords extracted from the first view in <code>preferredViews</code> that is found to contain content. If no
     * keywords can be extracted from all preferred views, returns an empty <code>KeywordResults</code> object.
     *
     * @return a KeywordResults object containing keywords from the first of the <code>preferredViews</code> found in <code>foundContent</code> that yields a
     *         non-empty set of keywords or an empty KeywordResults object if no keywords can be extracted.
     */
    @Nonnull
```

warehouse/keyword-common/src/main/java/datawave/util/keyword/KeywordExtractor.java
```diff
@@ -99,23 +102,37 @@ public void parseOptions(Map<String,String> iteratorOptions) {
...
+                if (results != EMPTY_RESULTS) {
+                    return results;
                 }
             }
         }
-        return results;
+        return EMPTY_RESULTS;
+    }
+
+    @Nonnull
```
drewfarris, line 148: Consider adding this javadoc:
```suggestion
    /**
     * Attempts to extract keywords from VisibleContent object provided.
     *
     * @param viewName
     *            the view name we're extracting from.
     * @param content
     *            the content associated with that view.
     * @return a KeywordResult object containing keywords extracted from the provided content, or the canonical <code>EMPTY_RESULTS</code> KeywordResults object
     *         if no keywords could be extracted.
     */
    @Nonnull
```

drewfarris (changes requested):
Just some minor javadoc additions for clarity.

Thanks for catching the use of the guava apis we shouldn't be using any longer.

drewfarris (approved):

FineAndDandy (approved):

PR #3200, opened 2025-09-23
Upgrade to protobuf 3.25.5

drewfarris:


drewfarris:
I _think_ I might need to include the protobufs `web-services/client/src/main/protobuf` in this too, but we should take a look at:

* `./web-services/compile_protos.sh`
* `./web-services/client/src/main/protobuf/compile_protos.sh`

FineAndDandy:
its hard to know if the changes to the generated classes are safe or not... any suggestion on this process?

drewfarris:
This will require some comprehensive integration testing.

avgAGB:
a bit of a random question, but with the change in versions of protobuf do the .proto files in datawave-ws-client and datawave-core need to adjust syntax in any way in order to be compatible?

drewfarris:
> a bit of a random question, but with the change in versions of protobuf do the .proto files in datawave-ws-client and datawave-core need to adjust syntax in any way in order to be compatible?

Typically it's not necessary to upgrade he `.proto` files - especially when going from 3.16 -> 3.25, which are generally compatible versions. One thing that will likely change is the generated java source code.. as often minor version changes may involved changes in generators.

PR #3208, opened 2025-09-25
pre order the views before passing to keyword extractor

austin007008:
pre order the views before passing to keyword extractor

### DIFF
--- a/warehouse/keyword-common/src/main/java/datawave/util/keyword/KeywordExtractor.java
+++ b/warehouse/keyword-common/src/main/java/datawave/util/keyword/KeywordExtractor.java
@@ -55,6 +55,23 @@ public class KeywordExtractor {
 
     YakeKeywordExtractor yakeKeywordExtractor;
 
+    /**
+     * Creates a {@code KeywordExtractor}, a lightweight wrapper for {@link YakeKeywordExtractor} that handles selecting from multiple sources of content,
+     * choosing a language for keyword processing and property-based configuration.
+     *
+     * @param source
+     *            used to name the source for the content from which we'll extract keywords, carried forward into the results object potentially used for
+     *            display or grouping multiple keyword sets.
+     * @param orderedContent
+     *            an ordered list of content. The extractor will attempt to extract keywords from each item and will stop after the first item it successfully
+     *            extracts keywords from.
+     * @param language
+     *            the language to use when extracting content, if the language can't be found in the {@link YakeLanguage.Registry}, it will default to english.
+     *            This language will also be included in the results object.
+     * @param options
+     *            values for the options related to {@code MIN_NGRAMS}, {@code MAX_NGRAMS}, {@code MAX_KEYWORDS}, {@code MAX_SCORE} or
+     *            {@code MAX_CONTENT_CHARS}.
+     */
     public KeywordExtractor(String source, List<Map.Entry<String,VisibleContent>> orderedContent, String language, Map<String,String> options) {
         this.source = source;
         this.orderedContent = orderedContent;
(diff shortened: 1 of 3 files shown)


austin007008:
just an internal change. is a no-op

awrigh11 (dismissed):

warehouse/keyword-common/src/main/java/datawave/util/keyword/KeywordExtractor.java
```diff
@@ -56,10 +55,9 @@ public class KeywordExtractor {
 
     YakeKeywordExtractor yakeKeywordExtractor;
 
-    public KeywordExtractor(String source, List<String> preferredViews, Map<String,VisibleContent> foundContent, String language, Map<String,String> options) {
+    public KeywordExtractor(String source, List<Map.Entry<String,VisibleContent>> orderedContent, String language, Map<String,String> options) {
```
drewfarris, line 75:
```suggestion
    /**
     * Creates a {@code KeywordExtractor}, a lightweight wrapper for {@link YakeKeywordExtractor} that handles selecting from multiple sources of content,
     * choosing a language for keyword processing and property-based configuration.
     *
     * @param source
     *            used to name the source for the content from which we'll extract keywords, carried forward into the results object potentially used for
     *            display or grouping multiple keyword sets.
     * @param orderedContent
     *            an ordered list of content. The extractor will attempt to extract keywords from each item and will stop after the first item it successfully
     *            extracts keywords from.
     * @param language
     *            the language to use when extracting content, if the language can't be found in the {@link YakeLanguage.Registry}, it will default to english.
     *            This language will also be included in the results object.
     * @param options
     *            values for the options related to {@code MIN_NGRAMS}, {@code MAX_NGRAMS}, {@code MAX_KEYWORDS}, {@code MAX_SCORE} or
     *            {@code MAX_CONTENT_CHARS}.
     */
    public KeywordExtractor(String source, List<Map.Entry<String,VisibleContent>> orderedContent, String language, Map<String,String> options) {
```

warehouse/query-core/src/main/java/datawave/query/iterator/logic/KeywordExtractingIterator.java
```diff
@@ -226,10 +226,34 @@ public void next() throws IOException {
...
+                    orderedContent.add(entry);
+                    entriesToRemove.add(entry.getKey());
+                }
+            }
+            if (!entriesToRemove.isEmpty()) {
+                for (String key : entriesToRemove) {
+                    foundContent.remove(key);
+                }
+            }
+        }
```
drewfarris: A slightly different variant that takes advantage of the ability to remove elements from the `foundContent` map using an iterator on the entrySet and handles the wildcards a little more cleanly. A little more concise - take it or leave it.

Regardless of which path we choose, I think you want to add a break after the exact match case to stop iterating over the entrySet once we've found an exact match. We can't do this in the pattern match case because there might be more entries that match the wildcard.

```suggestion
        final List<Map.Entry<String,VisibleContent>> orderedContent = new ArrayList<>();
        for (String view : preferredViews) {
            final String truncatedView = view.endsWith("*") ? view.substring(0, view.length() - 1) : null;
            final Iterator<Map.Entry<String,VisibleContent>> iterator = foundContent.entrySet().iterator();

            while (iterator.hasNext()) {
                final Map.Entry<String,VisibleContent> entry = iterator.next();
                final String foundView = entry.getKey();

                if (truncatedView != null && foundView.startsWith(truncatedView)) {
                    orderedContent.add(entry);
                    iterator.remove();
                } else if (foundView.equals(view)) {
                    orderedContent.add(entry);
                    iterator.remove();
                    break;
                }
            }
        }
```

drewfarris (changes requested):
Just some Javadoc comments to add and a tweak to the loop in KeywordExtractingIterator.

apmoriarty (approved):

drewfarris (approved):

PR #3211, opened 2025-09-26
Create a Field Replacement Visitor for replacing fields in ranges

awrigh11:


warehouse/query-core/src/test/java/datawave/query/planner/replacement/FieldReplacementVisitorTest.java
```diff
@@ -0,0 +1,186 @@
...
+    public void directFieldReplacementTest() throws Exception {
+        // @formatter:off
+        String query = "ABC == 'x'";
+        String expected = "XYZ == 'x'" ;
+        // @formatter:on
+        testReplacement(query, expected, List.of(dfrRule), false);
+    }
+
+    @Test
+    public void multiRuleReplacementTest() throws Exception {
```
hlgp, line 125: really nice tests!

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -1174,6 +1186,11 @@ protected ASTJexlScript processTree(final ASTJexlScript originalQueryTree, Shard
         // need to fetch field to datatype map first
         timedFetchDatatypes(timers, "Fetch Required Datatypes", config.getQueryTree(), config);
 
+        if (!disableFieldReplacementRules) {
```
hlgp, line 1189: perhaps a comment here for our future selves about how this must be called before expand multi normalized terms, so no one accidentally refactors this in the wrong order

warehouse/query-core/src/main/java/datawave/query/planner/replacement/rules/DirectFieldReplacementRule.java
```diff
@@ -0,0 +1,43 @@
+package datawave.query.planner.replacement.rules;
+
+import org.apache.commons.jexl3.parser.ASTIdentifier;
+import org.apache.commons.jexl3.parser.JexlNode;
+
+import datawave.query.jexl.JexlASTHelper;
+
+public class DirectFieldReplacementRule implements FieldReplacementRule {
```
apmoriarty, line 8: Add class level documentation for what the expected behavior is, what types of nodes are affected, etc

warehouse/query-core/src/main/java/datawave/query/planner/replacement/rules/DirectFieldReplacementRule.java
```diff
@@ -0,0 +1,43 @@
...
+    private String field = null;
+    private String replacement = null;
```
apmoriarty, line 10: These can be final

warehouse/query-core/src/main/java/datawave/query/planner/replacement/rules/FieldReplacementRule.java
```diff
@@ -0,0 +1,9 @@
+package datawave.query.planner.replacement.rules;
+
+import org.apache.commons.jexl3.parser.JexlNode;
+
+public interface FieldReplacementRule {
```
apmoriarty, line 5: Add brief class level doc

warehouse/query-core/src/main/java/datawave/query/planner/replacement/rules/RangeFieldReplacementRule.java
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.commons.jexl3.parser.JexlNodes;
+import org.apache.commons.jexl3.parser.ParserTreeConstants;
+
+import datawave.query.jexl.JexlASTHelper;
+import datawave.query.jexl.JexlNodeFactory;
+import datawave.query.jexl.LiteralRange;
+import datawave.query.jexl.nodes.QueryPropertyMarker;
+import datawave.query.jexl.visitors.RebuildingVisitor;
+
+public class RangeFieldReplacementRule implements FieldReplacementRule {
```
apmoriarty, line 21: Visitor name should be updated to include "BoundedRange", range on it's own could be any range operator (LT, LE, GT, GE).

The direct field replacement rule actually replaces the field. This visitor doesn't actually replace the bounded range, consider an alternate name.

warehouse/query-core/src/main/java/datawave/query/planner/replacement/rules/RangeFieldReplacementRule.java
```diff
@@ -0,0 +1,85 @@
...
+    public JexlNode apply(JexlNode node) {
+        LiteralRange<?> range = JexlASTHelper.findRange().getRange(node);
+
+        if (range == null || !fieldMap.containsKey(range.getFieldName())) {
+            return node;
+        }
+
+        // Create a copy and mark it as "evaluationOnly"
+        JexlNode copy = RebuildingVisitor.copy(node);
+        JexlNode evalNode = QueryPropertyMarker.create(copy, EVALUATION_ONLY);
```
apmoriarty, line 46: Why mark the original as evaluation only? Are the replacement fields always indexed?

If you had a case where the original field is indexed and the replacement field is not indexed then the executability of the query was just destroyed, in cases where the bounded range is an anchor term.

warehouse/query-core/src/main/java/datawave/query/planner/replacement/FieldReplacementVisitor.java
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.commons.jexl3.parser.ASTIdentifier;
+import org.apache.commons.jexl3.parser.ASTJexlScript;
+import org.apache.commons.jexl3.parser.JexlNode;
+import org.apache.log4j.Logger;
+
+import datawave.core.common.logging.ThreadConfigurableLogger;
+import datawave.query.jexl.visitors.RebuildingVisitor;
+import datawave.query.planner.replacement.rules.FieldReplacementRule;
+
+public class FieldReplacementVisitor extends RebuildingVisitor {
```
apmoriarty, line 15: Add class level docs

warehouse/query-core/src/main/java/datawave/query/planner/replacement/FieldReplacementVisitor.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Override
+    public Object visit(ASTAndNode node, Object data) {
+        return applyRules(super.visit(node, data));
+    }
+
+    @Override
+    public Object visit(ASTIdentifier node, Object data) {
+        return applyRules(super.visit(node, data));
+    }
```
apmoriarty, line 37: We typically avoid overriding the visit method to an identifier node. Instead you may want to override the specific nodes you care about, EQ, NE, ER, RN, LT, LE, GT, GE, etc.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -227,6 +229,13 @@ public class DefaultQueryPlanner extends QueryPlanner implements Cloneable {
      */
     protected boolean disableTestNonExistentFields = false;
 
+    /**
+     * Disables Field Replacement rules
+     *
+     * @see FieldReplacementVisitor
+     */
+    protected boolean disableFieldReplacementRules = false;
```
apmoriarty, line 237: double negatives have gotten us into trouble in the past. This should be one of the following 
```
fieldReplacementRulesEnabled = false

enableFieldReplacementRules = false
```

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
apmoriarty: That would be extremely helpful

warehouse/query-core/src/test/java/datawave/query/planner/replacement/FieldReplacementVisitorTest.java
```diff
@@ -0,0 +1,186 @@
...
+        testReplacement(query, expected, List.of(rfrRule), true);
+    }
+
+    @Test
+    public void directFieldReplacementTest() throws Exception {
+        // @formatter:off
+        String query = "ABC == 'x'";
+        String expected = "XYZ == 'x'" ;
+        // @formatter:on
+        testReplacement(query, expected, List.of(dfrRule), false);
```
apmoriarty, line 121: If this equality term was part of an intersection it would not get replaced. Add tests for replacing a single child in a union and in an intersection

warehouse/query-core/src/test/java/datawave/query/planner/replacement/FieldReplacementVisitorTest.java
```diff
@@ -0,0 +1,186 @@
...
+import datawave.query.jexl.visitors.RebuildingVisitor;
+import datawave.query.jexl.visitors.TreeEqualityVisitor;
+import datawave.query.planner.replacement.rules.DirectFieldReplacementRule;
+import datawave.query.planner.replacement.rules.FieldReplacementRule;
+import datawave.query.planner.replacement.rules.RangeFieldReplacementRule;
+
+public class FieldReplacementVisitorTest {
+    private static final Logger log = Logger.getLogger(FieldReplacementVisitorTest.class);
+    private static final DirectFieldReplacementRule dfrRule = new DirectFieldReplacementRule("ABC", "XYZ");
+    private static final Map<String,String> rangeMap = Map.of("AA", "BB", "CC", "DD");
```
apmoriarty, line 27: I like the convention you setup in the direct field replacement rule where `ABC` -> `XYZ`. You might consider doing the same with the range map, something like
```
AA -> XX
BB -> YY
CC -> ZZ
```

warehouse/query-core/src/test/java/datawave/query/planner/replacement/FieldReplacementVisitorTest.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    private void assertScriptEquality(ASTJexlScript actualScript, String expected) throws ParseException {
+        ASTJexlScript expectedScript = JexlASTHelper.parseJexlQuery(expected);
+        TreeEqualityVisitor.Comparison comparison = TreeEqualityVisitor.checkEquality(expectedScript, actualScript);
+        if (!comparison.isEqual()) {
+            log.error("Expected " + PrintingVisitor.formattedQueryString(expectedScript));
+            log.error("Actual " + PrintingVisitor.formattedQueryString(actualScript));
+        }
+        assertTrue(comparison.getReason(), comparison.isEqual());
```
apmoriarty, line 65: +1. This is the way to use the TreeEqualityVisitor. Excellent work.

warehouse/query-core/src/test/java/datawave/query/planner/replacement/FieldReplacementVisitorTest.java
```diff
@@ -0,0 +1,186 @@
...
+        // apply the visitor
+        ASTJexlScript resultScript = FieldReplacementVisitor.apply(originalScript, rules);
+
+        // Verify the script is as expected, and has a valid lineage.
+        assertScriptEquality(resultScript, expected);
+        assertLineage(resultScript);
+
+        // Verify the original script was not modified, and still has a valid lineage.
+        assertScriptEquality(originalScript, original);
+        assertLineage(originalScript);
```
apmoriarty, line 43: Asserting script equality AND lineage? Love it.

warehouse/query-core/src/test/java/datawave/query/planner/replacement/FieldReplacementVisitorTest.java
```diff
@@ -0,0 +1,186 @@
...
+
+import datawave.query.jexl.JexlASTHelper;
+import datawave.query.jexl.visitors.PrintingVisitor;
+import datawave.query.jexl.visitors.RebuildingVisitor;
+import datawave.query.jexl.visitors.TreeEqualityVisitor;
+import datawave.query.planner.replacement.rules.DirectFieldReplacementRule;
+import datawave.query.planner.replacement.rules.FieldReplacementRule;
+import datawave.query.planner.replacement.rules.RangeFieldReplacementRule;
+
+public class FieldReplacementVisitorTest {
```
apmoriarty, line 24: These tests are a good start. Given that the visitor supports replacing both single fields and bounded ranges it would be prudent to cover the following cases

- single term
- single term in union
- single term in intersection
- bounded range (covered)
- bounded range in union (covered)
- bounded range in intersection

It would be nice to see a field replacement rule run against each of the basic node types (EQ, NE, ER, NR, LT, GT, LE, GE) just to document what the expected behavior is.

Based on reading the code there is a bug where the bounded range rules will not be applied to an intersection of bounded ranges. Fixing that bug could lead to another bug where a direct field replacement rule is accidentally run on a bounded range. So there should be a test case added that looks like this:
````
direct field rule AA -> XX
(bounded range for AA)
```

warehouse/query-core/src/test/java/datawave/query/planner/replacement/FieldReplacementVisitorTest.java
```diff
@@ -0,0 +1,186 @@
...
+
+    @Test
+    public void rangeFieldReplacementWithMultipleRangesTest() throws Exception {
+        // @formatter:off
+        String query = "((_Bounded_ = true) && (CC >= '2' && CC <= '4')) || ((_Bounded_ = true) && (AA >= '2' && AA <= '4'))";
+        String expected = "(((_Eval_ = true) && ((_Bounded_ = true) && (CC >= '2' && CC <= '4'))) && " +
+                "((_Bounded_ = true) && (DD >= '2' && DD <= '4'))) || " +
+                "(((_Eval_ = true) && ((_Bounded_ = true) && (AA >= '2' && AA <= '4'))) && " +
+                "((_Bounded_ = true) && (BB >= '2' && BB <= '4')))" ;
+        // @formatter:on
```
apmoriarty, line 111: I really like the way this is broken out. Makes the test easy to read.

apmoriarty (changes requested):

PR #3212, opened 2025-09-29
make summary transform work with TLDs

austin007008:
make summary transform work with TLDs

if we have multiple record ids, choose the one for the TLD

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/transformer/SummaryTransform.java
+++ b/warehouse/query-core/src/main/java/datawave/query/transformer/SummaryTransform.java
@@ -17,6 +17,7 @@ import java.util.Set;
 
 import javax.annotation.Nullable;
 
+import datawave.query.attributes.Attribute;
 import org.apache.accumulo.core.data.Key;
 import org.apache.accumulo.core.data.PartialKey;
 import org.apache.accumulo.core.data.Range;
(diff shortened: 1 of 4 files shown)


warehouse/query-core/src/main/java/datawave/query/transformer/SummaryTransform.java
```diff
@@ -106,8 +107,21 @@ public Entry<Key,Document> apply(@Nullable Entry<Key,Document> entry) {
      */
     private static ArrayList<DocumentKey> getEventIds(Document document) {
```
apmoriarty: Open this up to protected and add some unit tests

apmoriarty (changes requested):

warehouse/query-core/src/test/java/datawave/query/util/SummaryTest.java
```diff
@@ -421,6 +430,22 @@ public void testMultiView() throws Exception {
...
+        Map<String,String> extraParameters = new HashMap<>();
+        extraParameters.put("include.grouping.context", "true");
+        extraParameters.put("return.fields", "SUMMARY");
+        extraParameters.put("query.syntax", "LUCENE");
+
+        String queryString = "QUOTE:farther AND QUOTE:child #SUMMARY(gimme)";
+
+        // not sure why the timestamp and delete flag are present
+        Set<String> goodResults = new HashSet<>(Set.of(
+                        "SUMMARY:CONTENT: You can get much farther with a kind word and a gun than you can with a kind word alone: : [] 9223372036854775807 false"));
```
apmoriarty: Accumulo key structure should not be in the final content

warehouse/query-core/src/test/java/datawave/query/util/SummaryTest.java
```diff
@@ -148,15 +155,16 @@ public static void teardown() {
     public void setup() {
         TimeZone.setDefault(TimeZone.getTimeZone("GMT"));
         log.setLevel(Level.TRACE);
-        logic.setFullTableScanEnabled(true);
+        eventLogic.setFullTableScanEnabled(true);
+        TLDLogic.setFullTableScanEnabled(true);
```
apmoriarty: Do we actually need to enable a full table scan? If not, remove these.

warehouse/query-core/src/main/java/datawave/query/transformer/SummaryTransform.java
austin007008: keeping private but added unit test in the current SummaryTest that tests summary functionality for TLDs. Verified that the new test exercises the added code path

warehouse/query-core/src/test/java/datawave/query/util/SummaryTest.java
austin007008: in production it doesn't actually get added. It only does that for all of these kinds of tests and no one knows why (previously tried to investigate when making summary code and also when updating excerpt code).

warehouse/query-core/src/test/java/datawave/query/util/SummaryTest.java
austin007008: I'll check. all the other tests do it and I just copied them

austin007008: we do not need it enabled. removed

apmoriarty (approved):

hlgp (approved):

PR #3226, opened 2025-10-08
Fix IPv6 compression

apmoriarty:
Previous algorithm was incorrectly replacing a single leading hextet of all zeroes with `::` which led to invalid IPv6 addresses that would throw an exception when attempting to parse.

### DIFF
--- a/core/utils/type-utils/src/main/java/datawave/data/type/util/IpV6Address.java
+++ b/core/utils/type-utils/src/main/java/datawave/data/type/util/IpV6Address.java
@@ -126,10 +126,10 @@ public class IpV6Address extends IpAddress {
         int zeroHextetLength = -1;
         if (skipZeros) {
             for (int i = 0; i < hextets.size(); i++) {
-                if (isHextetAllZeroes(hextets.get(i))) {
+                if (hextets.get(i).equals("0")) {
                     int len = 1;
                     for (int j = i + 1; j < hextets.size(); j++) {
-                        if (isHextetAllZeroes(hextets.get(j))) {
+                        if (hextets.get(j).equals("0")) {
                             len++;
                         } else {
                             break;
@@ -140,6 +140,7 @@ public class IpV6Address extends IpAddress {
                         zeroHextetStart = i;
                         zeroHextetLength = len;
                         log.trace("found sequence at index: {} length: {}", zeroHextetStart, zeroHextetLength);
+                        i += (len - 1); // do not look at zero hextets that were already skipped
                     }
                 }
             }
@@ -147,15 +148,6 @@ public class IpV6Address extends IpAddress {
 
         // now build the final address, replacing the longest run of zeroes if necessary
         return buildAddress(hextets, zeroHextetStart, zeroHextetLength);
-    }
-
-    private static boolean isHextetAllZeroes(String hextet) {
-        for (int i = 0; i < hextet.length(); i++) {
-            if (hextet.charAt(i) != '0') {
-                return false;
-            }
-        }
-        return true;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


core/utils/type-utils/src/main/java/datawave/data/type/util/IpV6Address.java
```diff
@@ -98,48 +105,87 @@ public static IpV6Address parse(String address) {
...
-            int count = 0;
-            for (int i = 0; i < 8; i++) {
-                if (address[i] == 0) {
-                    count++;
-                } else {
-                    if (count > length) {
-                        startSkip = i - count;
-                        length = count;
+            for (int i = 0; i < hextets.size(); i++) {
+                if (isHextetAllZeroes(hextets.get(i))) {
```
ivakegg: could this be simply (address[i] == 0)

core/utils/type-utils/src/main/java/datawave/data/type/util/IpV6Address.java
```diff
@@ -98,48 +105,87 @@ public static IpV6Address parse(String address) {
...
+                            len++;
+                        } else {
+                            break;
+                        }
+                    }
+
+                    if (len >= 2 && len > zeroHextetLength) {
+                        zeroHextetStart = i;
+                        zeroHextetLength = len;
+                        log.trace("found sequence at index: {} length: {}", zeroHextetStart, zeroHextetLength);
```
ivakegg, line 144: could advance i here to avoid checking already skipped zeros

core/utils/type-utils/src/main/java/datawave/data/type/util/IpV6Address.java
```diff
@@ -98,48 +105,87 @@ public static IpV6Address parse(String address) {
...
-                    count++;
-                } else {
-                    if (count > length) {
-                        startSkip = i - count;
-                        length = count;
+            for (int i = 0; i < hextets.size(); i++) {
+                if (isHextetAllZeroes(hextets.get(i))) {
+                    int len = 1;
+                    for (int j = i + 1; j < hextets.size(); j++) {
+                        if (isHextetAllZeroes(hextets.get(j))) {
```
ivakegg: could be address[j] == 0

core/utils/type-utils/src/main/java/datawave/data/type/util/IpV6Address.java
```diff
@@ -126,10 +126,10 @@ public static String toString(short[] address, boolean zeroPadded, boolean skipZ
         int zeroHextetLength = -1;
         if (skipZeros) {
             for (int i = 0; i < hextets.size(); i++) {
-                if (isHextetAllZeroes(hextets.get(i))) {
+                if (hextets.get(i).equals("0")) {
```
ivakegg, line 131: if zeropadding, then hextets.get(i).equals("0000")

ivakegg: perhaps we need some test with zero passing on

apmoriarty: per discussion, verified the method is always called with different flags set so this case should never be hit. Added a precondition check to fail if the method is ever erroneously called with true/true or false/false.

ivakegg (approved):

hlgp (approved):

core/utils/type-utils/src/main/java/datawave/data/type/util/IpV6Address.java
```diff
@@ -98,48 +106,80 @@ public static IpV6Address parse(String address) {
     }
 
     public static String toString(short[] address, boolean zeroPadded, boolean skipZeros) {
-        StringBuilder builder = new StringBuilder(39);
-        int startSkip = -1;
-        int length = 0;
+
+        Preconditions.checkArgument(zeroPadded != skipZeros, "cannot zero pad address and skip zeros");
```
alerman, line 110: This would also fail if both not zero padded and not skip zeros. Intended?

Issue #3243, opened 2025-10-23
implement simple annotation ingest for testing


d-hwang:
we need a simple ingest mechanism for annotation data to be used in quickstart

SethSmucker:
there's a warehouse/ingest-annotation module now #3245. Closing

PR #3245, opened 2025-10-23
simple annotation ingest (#3243)

jack-gitdev:
Dan was having github issues so I pushed this for him in case anyone has questions about this PR


pom.xml
```diff
@@ -19,6 +19,7 @@
         <module>core</module>
         <module>web-services</module>
         <module>warehouse</module>
+        <module>warehouse/ingest-annotation</module>
```
drewfarris: Should this be in `warehouse/pom.xml` as an `ingest-annotation` module instead of calling it out from the top level `pom.xml`?

warehouse/assemble/datawave/pom.xml
```diff
@@ -25,6 +25,16 @@
...
+        <dependency>
+            <groupId>com.google.api.grpc</groupId>
+            <artifactId>proto-google-common-protos</artifactId>
+            <version>${version.proto-google-common-protos}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.protobuf</groupId>
+            <artifactId>protobuf-java-util</artifactId>
+            <version>${version.protobuf-java-util}</version>
+        </dependency>
```
drewfarris, line 35: Should we put these in the `dependencyManagement` section of the top level pom and remove the version specification here? It seems we do this a little inconsistently across the poms (e.g., there are other dependencies in this section that call out the version..), but these particular dependencies will also be used in the webservices, so we might want to have them defined in a different way elsewhere. This might also adjust how we reference versions in other poms (e.g., I can't remember if I included the versions like this in the annotation modules, but I probably shouldn't have...)

drewfarris (dismissed):

pom.xml
d-hwang: yep, updated.

warehouse/assemble/datawave/pom.xml
d-hwang: updated. let me know if it looks better now.

drewfarris (approved):

jalphonso (approved):
Looks great!

PR #3249, opened 2025-10-28
Fix issues with visitor

lbschanno:
- Simplify test setup and execution.
- Update visitor to collect fields from top-level nodes in separate subsets.
- Update rule to provide more information about which query subsets are considered immaterial.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldMissingFromDateRangeVisitor.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldMissingFromDateRangeVisitor.java
@@ -1,12 +1,13 @@
 package datawave.query.jexl.visitors;
 
-import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashSet;
+import java.util.Date;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.apache.accumulo.core.client.TableNotFoundException;
 import org.apache.commons.jexl3.parser.ASTEQNode;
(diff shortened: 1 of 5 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldMissingFromDateRangeVisitor.java
```diff
@@ -139,31 +84,41 @@ private Object collectFields(ASTOrNode node, Object data) throws TableNotFoundEx
      * @return the updated set of names which have not been ingested during the date range.
      */
     private Object collectFields(JexlNode node, Object data) throws TableNotFoundException {
```
foster33, line 97: Is the javadoc for this method still accurate?

warehouse/query-core/src/main/java/datawave/query/rules/FieldExistenceRule.java
```diff
@@ -95,6 +101,18 @@ public QueryRuleResult validate(QueryValidationConfiguration ruleConfiguration)
...
+        return new StringBuilder()
+                        .append("Fields ")
+                        .append(irrelevantJunction.getFields())
+                        .append(" found in the query subset '")
+                        .append(JexlStringBuildingVisitor.buildQuery(irrelevantJunction.getNode()))
+                        .append("' were not ingested in the provided date range.")
+                        .toString();
+        // @formatter:on
+    }
+
```
foster33: Would it be better if we just output the individual fields that are missing? Thinking it might be a bunch of output if a specific field is used multiple times and might make finding all of the different fields a bit difficult.

lbschanno: Hmm, I think so. I'll change it to just print out the fields.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldMissingFromDateRangeVisitor.java
lbschanno: Likely not, will update.

warehouse/query-core/src/main/java/datawave/query/rules/FieldExistenceRule.java
lbschanno: Alternatively, it could be helpful for the user to see which snippet of the query is made immaterial by the missing fields. Perhaps this is a decision for the customer to weigh in on.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldMissingFromDateRangeVisitor.java
lbschanno: Updated.

warehouse/query-core/src/main/java/datawave/query/rules/FieldExistenceRule.java
lbschanno: Message now just prints the set of fields.

foster33 (approved):

PR #3257, opened 2025-10-31
Add ability to enforce concurrent query limits

### LINKED ISSUE #3100: Limit the total number of queries a user can run concurrently in the system
The idea is that a group of webservers needs to coordinate how many queries a user can run concurrently.  The webserver zookeeper would be the place to coordinate the number of currently running queries for a user.  The system needs to be able to set limit on a user overall, and set a separate limit on the number of queries for a user per query logic.  The use case is that lookup UUID queries may have a higher threshold but all other queries would fall under a lower concurrent limit per user.

lbschanno:
## Configuration
Add the ability to enforce concurrent query limits across a group of webservers. Zookeeper is used to track active queries and the following data:
- The query ID
- The user who submitted the query
- The system the query was submitted on
- The query logic the query originated from

Query limit enforcement is done through the `QueryLimiter` class. Given a user, system, and query logic, it can determine if any of the following limits have been exceeded:

- The max allowed concurrent queries for the user.
- The max allowed concurrent queries of the query logic for the user.
- The max allowed concurrent queries for the system.
- The max allowed concurrent queries of the query logic for the system.

Limits may be defined and customized on a per-user and per-system basis. They may also be defined for groups of query logics. The classes `UserLimitProvider`, `SystemLimitProvider`, and `QueryLogicGroupLimitProvider` are respectively responsible for identifying the best limits to enforce for a user, system, and query logic. They will be initialized in the `QueryLimiter` after providing a `QueryLimitConfiguration` instance. The following can be configured:

On a system-wide basis:
- The default concurrent user query limit. This applies to the total number of queries a user may run across all systems. May be overridden per user.
- The default concurrent system query limit. Primarily to avoid a system getting overloaded. May be overridden per system.

On a per-system basis:
- The system name/ids the configuration targets. Regex matching is supported. Pattern uniqueness is enforced.
- The concurrent system query limit. Overrides the system-wide value.
- Whether queries submitted to the system count towards a user's concurrent query total. This is always true unless specified otherwise.
- The concurrent system query limit for different query logic groups. Regex matching against group names is supported. Pattern uniqueness is enforced.

On a per-user basis:
- The user DN.
- The user's concurrent query limit. Overrides the system-wide configuration.
- The user's concurrent query limit for different query logic groups. Regex matching against group names is supported.

On a per-query-logic-group basis:
- The group name.
- The query logics included in the group. Regex matching is supported. Pattern uniqueness is enforced.
- The default concurrent user query limit. This applies to the total concurrent queries a user may run that originate from a query logic in the group across all systems.

Given the possibilities for exact matches, partial regex matches, and wildcard regex matches, the determination of the best limit to use for any particular system, query logic, or query logic group is done by sorting matches into the following 'matching buckets' (in best-match priority):

1. Exact match: We attempt to find an exact match first and use the associated limit.
2. Partial regex (non-wildcard-only): If we cannot find an exact match, then we attempt to find all partial matches, and see if any of their limits are met.
3. Wildcard-only regex: In the case of no exact or partial matches, we use the wildcard match with the lowest limit.

## Implementation

Checking limits and marking as active/inactive is done through the `QueryLimiter` class. The three main methods to know are:
- `QueryLimiter.checkForLimits()`
- `QueryLimiter.countQueryTowardsLimits()`
- `QueryLimiter.stopCountingQueryTowardsLimits()`

When a query is marked as active via `QueryLimiter.countQueryTowardsLimits()`, it will delegate to the ActiveQueryTracker, which will in turn create nodes in Zookeeper under the `ActiveQueries` namespace. When `ActiveQueryTracker.trackQuery()` is called, the following nodes are created:

```
# Container nodes
/users/<userDn>/<queryLogic> # Only for queries on systems that count towards user limit
/systems/<system>/<queryLogic>
/distinctQueryLogics/<queryLogic>  [Only created if it does not exist.]

# Ephemeral nodes. These will auto-delete themselves if their associated Zookeeper connection ever goes down.
/users/<userDn>/<queryLogic>/<queryId> # Only for queries on systems that count towards user limit
/systems/<system>/<queryLogic>/<queryId>
```

`ActiveQueryTracker.trackQuery()` will return a `QueryHeartbeat` that contain a list of `PersistentNode` (provided by the Apache Curator library) wrappers around the ephemeral nodes listed above. The `QueryHeartbeat` will maintain the connection  to Zookeeper and attempt to keep the ephemeral nodes present in Zookeeper until `QueryHeartbeat.stop()` is called. If `QueryHeartbeat.stop()` is called, or the webserver crashes, the ephemeral nodes will automatically be deleted by Zookeeper.

The following error codes have been added:
```
412-20  - Concurrent query limit exceeded
500-164 - Error checking concurrent query limits
```
Closes #3100


web-services/query/src/main/java/datawave/webservice/query/limit/QueryLimiter.java
```diff
@@ -0,0 +1,570 @@
...
+         *
+         * @throws Exception
+         *             if an error occurs while fetching the snapshot.
+         */
+        private void loadSnapshot() throws Exception {
+            // When fetching the snapshot of query activity, we need to include queries for any query logics that may count towards a query logic group limit
+            // applicable for the candidate user, system, and query logic. The applicable query logic groups depend on the default query logic group limits,
+            // along with any custom limits defined for the user or system.
+            Set<String> queryLogics = getQueryLogicsToFetchInSnapshot();
+            this.snapshot = getActiveQueryTracker().getSnapshot(userDn, system, queryLogics);
```
ivakegg: Do we need to load everything in the snapshot for the user, system, and query logics?  Perhaps we only need a subset of that depending on what the configured limits are?

web-services/query/src/main/java/datawave/webservice/query/limit/GroupLimitCache.java
```diff
@@ -0,0 +1,98 @@
...
+
+    // The set of group limit overrides, sorted in best-match order.
+    private final SortedSet<QueryLogicGroupLimit> groupLimits;
+
+    // Internal cache for improved lookup efficiency.
+    private final Cache<String,Map<String,Integer>> cache;
+
+    public GroupLimitCache(SortedSet<QueryLogicGroupLimit> groupsToLimits) {
+        if (groupsToLimits != null && !groupsToLimits.isEmpty()) {
+            this.cache = Caffeine.newBuilder().maximumSize(100).build();
```
ivakegg: 100 could actually be too small a maximum to hold all query logics.  Make this 200, and perhaps configurable

lbschanno: Will do.

web-services/query/src/main/java/datawave/webservice/query/limit/QueryLimiter.java
lbschanno: Perhaps so, I'll rework this so that we're checking against limits as we're actively scanning over the nodes, and returning early when we've met a limit.

web-services/query/src/main/java/datawave/webservice/query/limit/GroupLimitCache.java
lbschanno: Addressed in latest commit.

web-services/query/src/main/java/datawave/webservice/query/limit/QueryLimiter.java
lbschanno: This logic has been reworked to incrementally scan and aggregate totals only as needed.

web-services/cached-results/src/main/java/datawave/webservice/results/cached/CachedResultsBean.java
```diff
@@ -722,13 +746,19 @@ else if (maxValueLength > maxLength) {
                 } catch (Exception e) {
                     response.addException(new QueryException(DatawaveErrorCode.QUERY_CLOSE_ERROR, e).getBottomQueryException());
                 }
+                try {
+                    queryLimiter.stopCountingQueryTowardsLimits(query.getSettings().getId().toString());
+                } catch (Exception e) {
+                    log.error("Failed to stop counting query " + query.getSettings().getId().toString() + "towards limits", e);
```
foster33:
```suggestion
                    log.error("Failed to stop counting query " + query.getSettings().getId().toString() + " towards limits", e);
```

web-services/query/src/main/java/datawave/webservice/query/limit/ActiveQueryTracker.java
```diff
@@ -0,0 +1,616 @@
...
+    private long lastClientAccess;
+    private Timer clientCleanupTimer;
+
+    /**
+     * Create and return a new {@link ActiveQueryTracker} instance
+     *
+     * @param zookeeperConfig
+     *            the zookeeper config
+     * @param clientCleanupInterval
+     *            the interval in milliseconds after which the zookeeper client should be cleaned up since its last access\
```
foster33:
```suggestion
     *            the interval in milliseconds after which the zookeeper client should be cleaned up since its last access
```

web-services/query/src/main/java/datawave/webservice/query/limit/ActiveQueryTracker.java
```diff
@@ -0,0 +1,616 @@
...
+import org.apache.hadoop.fs.Path;
+import org.apache.log4j.Logger;
+import org.apache.zookeeper.CreateMode;
+import org.apache.zookeeper.KeeperException;
+import org.apache.zookeeper.data.Stat;
+import org.apache.zookeeper.server.quorum.QuorumPeer;
+import org.apache.zookeeper.server.quorum.QuorumPeerConfig;
+
+/**
+ * This class provides methods for leveraging Zookeeper to track queries and their active status.
```
ivakegg, line 29: All the documentation you put into this ticket could be encoded here perhaps.

ivakegg: alternatively a README in this directory.

lbschanno: Fair point, I'll do that.

lbschanno: Added a README file.

ivakegg (changes requested):
I like what I am seeing in here.  I am thinking about potential error scenarios where things could get out of sync.  Currently we have the following maps related to queries:

1) QueryCache: A map of query id to RunningQuery instances.  Used by QueryExpirationBean to check for expired queries.
2) HeartbeatCache: A map of query id to heartbeat.

It might be worth adding a loop in the QueryExpirationBean that goes through the ids in the HeartbeatCache to verify that those queries are still running per the QueryCache.  This would help to prevent anything getting out of sync if we have some unknown error cases that are not being accounted for.  Other than that I am good to start integration testing this.

lbschanno:
@ivakegg understood, I'll add in a loop to the QueryExpirationBean for synchronization with the heartbeat cache.

lbschanno:
Added synchronization safeguard.

ivakegg:
So I got this up and running and watching the zookeeper entries I realized that there has been one requirements which was not translated correctly and I apologize for not realizing this earlier.  The "system" limit is supposed to be base on the "systemFrom" query parameter and not the hostname on which the query is running.  I will take a shot at modifying the code accordingly.

ivakegg (dismissed):

apmoriarty (dismissed):

apmoriarty (approved):

ivakegg (approved):

avgAGB (approved):

apmoriarty (approved):

hlgp (approved):

PR #3258, opened 2025-10-31
Updated the planning to be concurrent for index holes in the

ivakegg:
date partitioned query planner


warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
```diff
@@ -332,29 +366,48 @@ public CloseableIterable<QueryData> process(GenericQueryConfiguration genericCon
         SortedMap<Pair<Date,Date>,Set<String>> dateRanges = getSubQueryDateRanges(planningConfig);
 
         DatePartitionedQueryIterable results = new DatePartitionedQueryIterable();
-        List<Exception> exceptions = new ArrayList<>();
+        List<Throwable> exceptions = new ArrayList<>();
+        List<Triple<Map.Entry<Pair<Date,Date>,Set<String>>,ShardQueryConfiguration,Future<CloseableIterable<QueryData>>>> futures = new ArrayList<>();
```
apmoriarty: Opportunity for a concrete class instead of a triple?

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
```diff
@@ -291,6 +314,17 @@ public CloseableIterable<QueryData> process(GenericQueryConfiguration genericCon
                     throws DatawaveQueryException {
         visitorManager.setDebugEnabled(log.isDebugEnabled());
 
+        // start up an executor first to give the threads a chance to start
+        final AtomicInteger threadCounter = new AtomicInteger(1);
+        ThreadFactory threadFactory = new ThreadFactory() {
+            @Override
+            public Thread newThread(Runnable r) {
+                return new Thread(r, "DatePartitionedQueryPlanner " + threadCounter.getAndIncrement());
```
jschmidt10: Thoughts on adding the query id to the thread name?

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
jschmidt10: This type declaration is pretty gnarly. Might be worth rolling a class to hold this so we can name the pieces.

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
```diff
@@ -332,29 +366,48 @@ public CloseableIterable<QueryData> process(GenericQueryConfiguration genericCon
...
+                @Override
+                public CloseableIterable<QueryData> call() {
+                    try {
+                        // Create a copy of the original default query planner, and process the query with the new date range.
+                        DefaultQueryPlanner subPlan = DatePartitionedQueryPlanner.this.queryPlanner.clone();
+
+                        // Get the range stream for the new date range and query
+                        return subPlan.reprocess(configCopy, configCopy.getQuery(), scannerFactory);
+                    } catch (Exception e) {
+                        throw new DatawaveAsyncOperationException(e);
```
jschmidt10: We could add a little context here maybe like:

```
  throw new DatawaveAsyncOperationException(queryId + " failed for " + dateRange, e);
```

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
```diff
@@ -332,29 +366,48 @@ public CloseableIterable<QueryData> process(GenericQueryConfiguration genericCon
...
+                    }
+                }
+            })));
+        }
+
+        for (Triple<Map.Entry<Pair<Date,Date>,Set<String>>,ShardQueryConfiguration,Future<CloseableIterable<QueryData>>> future : futures) {
```
jschmidt10: If you do the proposal above with the new class, then this comment is OBE'd. If we don't do a new class, we should rename this variable b/c only the third element of the triple is the future so it's hard to follow. I might name it something like `dateRangeTriple` or `partitionTriple` or something.

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
```diff
@@ -332,29 +366,48 @@ public CloseableIterable<QueryData> process(GenericQueryConfiguration genericCon
...
-
+                results.addIterable(future.getRight().get());
                 if (log.isDebugEnabled()) {
                     log.debug("Query string for config of sub-plan against date range (" + subBeginDate + "-" + subEndDate + ") with unindexed fields "
                                     + dateRange.getValue() + ": " + configCopy.getQueryString());
                 }
-            } catch (DatawaveQueryException e) {
+            } catch (ExecutionException e) {
                 log.warn("Exception occurred when processing sub-plan against date range (" + subBeginDate + "-" + subEndDate + ")", e);
+                if (e.getCause() instanceof DatawaveAsyncOperationException) {
```
jschmidt10: Can we just leave the DatawaveAsyncOperationException and put it in the exceptions list? If we add the context I suggested above, then we'd know which piece of the query the exception came from.

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
```diff
@@ -291,6 +314,17 @@ public CloseableIterable<QueryData> process(GenericQueryConfiguration genericCon
...
 
+        // start up an executor first to give the threads a chance to start
+        final AtomicInteger threadCounter = new AtomicInteger(1);
+        ThreadFactory threadFactory = new ThreadFactory() {
+            @Override
+            public Thread newThread(Runnable r) {
+                return new Thread(r, "DatePartitionedQueryPlanner " + threadCounter.getAndIncrement());
+            }
+        };
+        ExecutorService executor = new ThreadPoolExecutor(1, getMaxConcurrentPlanningThreads(), 60L, TimeUnit.SECONDS, new LinkedBlockingQueue<Runnable>(),
```
jschmidt10: We never `shutdown()` this executor.

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
ivakegg: Awesome idea.... will do

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
ivakegg: fair enough

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
ivakegg: doh

jschmidt10 (dismissed):

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
```diff
@@ -67,6 +76,10 @@ public class DatePartitionedQueryPlanner extends QueryPlanner implements Cloneab
     private DefaultQueryPlanner queryPlanner;
     private String initialPlan;
     private String plannedScript;
+    /**
+     * The max number of concurrent planning threads
+     */
+    private int maxConcurrentPlanningThreads = 10;
```
apmoriarty: This variable name seems like it's the absolute max planning threads, but it can be updated to be any value. This might be better named `numConcurrentPlanningThreads` and reserve `maxConcurrentPlanningThreads` for a final constant.

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
```diff
@@ -410,6 +487,60 @@ private void updatePlannedScript() {
...
+        public CloseableIterable<QueryData> call() throws Exception {
+            try {
+                // Get an updated configuration with the new date range and query tree
+                this.subPlanConfig = getUpdatedConfig(planningConfig, dateRange.getKey(), dateRange.getValue());
+
+                // Create a copy of the original default query planner, and process the query with the new date range.
+                DefaultQueryPlanner subPlan = basePlanner.clone();
+
+                // Get the range stream for the new date range and query
+                return subPlan.reprocess(subPlanConfig, subPlanConfig.getQuery(), scannerFactory);
```
apmoriarty, line 520: I'm curious about the memory impact of starting a range stream and hanging onto the reference. I suppose if this solution isn't tenable we could always grab the range stream, verify it has a hit, and then close it -- marking this subplan as 'has data'.

The point is this: even though the concurrency is limited for how many range streams are executing at one point in time, and even though the scanners close between next calls, we still have the entire object in memory and it ain't cheap.

jschmidt10 (dismissed):

warehouse/query-core/src/main/java/datawave/query/planner/DatePartitionedQueryPlanner.java
ivakegg: So we may need to reduce the concurrency down to 1 if that is an issue.  This is a straight forward tradeoff between memory and speed.

ivakegg: Not fully grasping what you are talking about as a solution.  If it has a hit, and we close it, then we can't get the hits.  What am I missing.

ivakegg: After our discussion I understand what you were getting at.  Essentially the Intersection/Union tree for each partition will be held in memory.  So if we have a very large query and several partitions, that in itself could become quite memory intensive.  
I am now thinking that I should do these subplans within the range stream so that they are processed as needed. instead of all up front.
I am going to keep this PR up but I will work on an alternative solution.

jschmidt10 (approved):

ivakegg:
Going for a streaming the subplans instead of precalculating up front: https://github.com/NationalSecurityAgency/datawave/pull/3283

PR #3260, opened 2025-11-03
Update AnnotationManagerBean to properly handle documents with multiple internal ids

drewfarris:
The major change here is that `lookupDocumentIdentifier` in AnnotationManagerBean now returns a `List<Metadata>` instead of a `Metadata` objects, so all bean methods had to be updated to work with more than one `Metadata` object.

Also implemented a separate annotation-sepecific `Metadata` different from the one in the event package because the two contexts has different needs and it made sense to decouple the two.

### DIFF
--- a/web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
+++ b/web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
@@ -191,7 +191,7 @@ public class AnnotationManagerBean implements AnnotationManager {
                 }
             }
             if (results.isEmpty()) {
-                return jsonNotFound("annotation types", idType, id, Metadata.format(metadata), null, null, null);
+                return jsonNotFound("annotation types", idType, id, metadata.toString(), null, null, null);
             }
             return jsonOk(results);
         } catch (Exception e) {
@@ -222,7 +222,7 @@ public class AnnotationManagerBean implements AnnotationManager {
                 }
             }
             if (results.isEmpty()) {
-                return jsonNotFound("annotations", idType, id, Metadata.format(metadata), null, null, null);
+                return jsonNotFound("annotations", idType, id, metadata.toString(), null, null, null);
             }
             return jsonOk(results);
         } catch (Exception e) {
@@ -254,7 +254,7 @@ public class AnnotationManagerBean implements AnnotationManager {
                 }
             }
             if (results.isEmpty()) {
-                return jsonNotFound("annotations of type", idType, id, Metadata.format(metadata), annotationType, null, null);
+                return jsonNotFound("annotations of type", idType, id, metadata.toString(), annotationType, null, null);
             }
             return jsonOk(results);
         } catch (Exception e) {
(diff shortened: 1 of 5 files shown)


web-services/annotations/src/main/java/datawave/webservice/annotation/Metadata.java
```diff
@@ -0,0 +1,103 @@
...
+
+        this.row = row;
+        this.dataType = dataType;
+        this.internalId = internalId;
+        this.table = table;
+    }
+
+    /** Format metadata for error messages, since it doesn't have a reasonable {@code toString} method */
+    static String format(Metadata metadata) {
+        return String.format("%s/%s/%s [%s]", metadata.getRow(), metadata.getDataType(), metadata.getInternalId(), metadata.getTable());
```
jschmidt10: I think it'd be simpler to put this in `Metadata.toString()` and then whether you are printing a `Metadata` or a `List<Metadata>`, both should print nicely without needing a special call.

web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
```diff
@@ -390,50 +423,42 @@ public Response updateAnnotation(@PathParam("idType") String idType, @PathParam(
...
 
-            // now filter out the segment that was requested
-            Annotation a = annotation.get();
-            List<Segment> matchingSegments = new ArrayList<>();
-            for (Segment s : a.getSegmentsList()) {
-                if (s.getSegmentId().equals(segmentId)) {
-                    matchingSegments.add(s);
+            final Map<Metadata,Collection<Segment>> results = new HashMap<>();
+            for (Map.Entry<Metadata,Annotation> entry : annotationResults.entrySet()) {
+                // now select only the segment that were requested.
```
jschmidt10:
```suggestion
                // now select only the segments that were requested.
```

web-services/annotations/src/main/java/datawave/webservice/annotation/Metadata.java
```diff
@@ -0,0 +1,103 @@
...
+        return internalId;
+    }
+
+    public String getRow() {
+        return row;
+    }
+
+    public String getTable() {
+        return table;
+    }
```
jalphonso, line 68: Since there is no special handling in these getters we can reduce the clutter by annotating the class with Lombok `@Getter`

drewfarris: Thanks for the suggestion. Lombok is availalble in the microservices starter, whch gives it to us in scope provided for all the microservices modules, I'd prefer to avoid doing this work for the webservices for now.

web-services/annotations/src/main/java/datawave/webservice/annotation/Metadata.java
drewfarris: Yes, I was being lazy when refactoring. Thanks for holding me accountable on this.

jschmidt10 (approved):

jalphonso (approved):

PR #3261, opened 2025-11-04
Add ColumnVisibility cache

apmoriarty:
Add several column visibility cache implementations along with an integration test that demonstrates cache performance under various levels of concurrency. 

Implementations include a synchronized LRU map, Guava and Caffeine.

### DIFF
--- a/core/common-util/src/main/java/datawave/core/cache/CaffeineCVCache.java
+++ b/core/common-util/src/main/java/datawave/core/cache/CaffeineCVCache.java
@@ -13,12 +13,20 @@ import com.github.benmanes.caffeine.cache.LoadingCache;
  */
 public class CaffeineCVCache implements CVCache {
 
-    //  @formatter:off
-    private final LoadingCache<ByteSequence,ColumnVisibility> cache = Caffeine.newBuilder()
-            .maximumSize(128)
-            .expireAfterAccess(1, TimeUnit.HOURS)
-            .build(bytes -> new ColumnVisibility(bytes.toArray()));
-    //  @formatter:on
+    private final LoadingCache<ByteSequence,ColumnVisibility> cache;
+
+    public CaffeineCVCache() {
+        this(DEFAULT_SIZE);
+    }
+
+    public CaffeineCVCache(int size) {
+        //  @formatter:off
+        cache = Caffeine.newBuilder()
+                .maximumSize(size)
+                .expireAfterAccess(1, TimeUnit.HOURS)
+                .build(bytes -> new ColumnVisibility(bytes.toArray()));
+        //  @formatter:on
+    }
 
     @Override
     public String name() {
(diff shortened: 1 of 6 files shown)


core/common-util/src/main/java/datawave/core/cache/GuavaCVCache.java
```diff
@@ -0,0 +1,43 @@
...
+import com.google.common.cache.LoadingCache;
+
+/**
+ * A guava {@link LoadingCache} implementation of a {@link CVCache}
+ */
+public class GuavaCVCache implements CVCache {
+
+    //  @formatter:off
+    private final LoadingCache<ByteSequence,ColumnVisibility> cache = CacheBuilder.newBuilder()
+            .maximumSize(128)
```
ivakegg: 128 is based on what?

core/common-util/src/main/java/datawave/core/cache/LRUMapCVCache.java
```diff
@@ -0,0 +1,31 @@
...
+import org.apache.accumulo.core.data.ByteSequence;
+import org.apache.accumulo.core.security.ColumnVisibility;
+import org.apache.commons.collections4.map.LRUMap;
+
+/**
+ * A {@link LRUMap} implementation of a {@link CVCache}
+ */
+public class LRUMapCVCache implements CVCache {
+
+    private final Map<ByteSequence,ColumnVisibility> cache = Collections.synchronizedMap(new LRUMap<>(256));
```
ivakegg: and here we use 256?

core/common-util/src/main/java/datawave/core/cache/CaffeineCVCache.java
```diff
@@ -0,0 +1,32 @@
...
+import com.github.benmanes.caffeine.cache.LoadingCache;
+
+/**
+ * A {@link Caffeine} implementation of a {@link CVCache}
+ */
+public class CaffeineCVCache implements CVCache {
+
+    //  @formatter:off
+    private final LoadingCache<ByteSequence,ColumnVisibility> cache = Caffeine.newBuilder()
+            .maximumSize(128)
```
ivakegg: Shall we set this during construction?

core/common-util/src/main/java/datawave/core/cache/LRUMapCVCache.java
apmoriarty: The original LRUMap implementation in accumulo-utils used 256...I'll make that the standard for all of these.

core/common-util/src/main/java/datawave/core/cache/CaffeineCVCache.java
apmoriarty: I wasn't planning to make the size configurable given that the current implementation is a fixed size.

apmoriarty: Let me know what you think of the updates

ivakegg (approved):

foster33 (approved):

PR #3262, opened 2025-11-04
Add class cache

apmoriarty:
Add several class cache implementations along with an integration test that demonstrates cache performance under various levels of concurrency.

Implementations include a synchronized LRU map, Guava and Caffeine.

### DIFF
--- a/core/common-util/src/main/java/datawave/core/cache/GuavaClassCache.java
+++ b/core/common-util/src/main/java/datawave/core/cache/GuavaClassCache.java
@@ -1,0 +1,40 @@
+package datawave.core.cache;
+
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+
+/**
+ * A guava {@link LoadingCache} implementation of a {@link ClassCache}
+ */
+public class GuavaClassCache implements ClassCache {
+
+    //  @formatter:off
+    private final LoadingCache<String,Class<?>> cache = CacheBuilder.newBuilder()
+            .maximumSize(128)
+            .expireAfterWrite(1, TimeUnit.HOURS)
+            .build(new CacheLoader<>() {
+                @Override
+                public Class<?> load(String name)throws ClassNotFoundException {
+                    return Class.forName(name);
+                }
+            });
+    //  @formatter:on
+
+    @Override
+    public String name() {
+        return "Guava";
+    }
+
+    @Override
+    public Class<?> get(String name) throws ClassNotFoundException {
+        try {
+            return cache.get(name);
+        } catch (ExecutionException e) {
+            throw new ClassNotFoundException("Invalid Class name", e);
+        }
+    }
+}
--- a/core/common-util/src/main/java/datawave/core/cache/GuavaClassCache.java
+++ b/core/common-util/src/main/java/datawave/core/cache/GuavaClassCache.java
@@ -12,17 +12,25 @@ import com.google.common.cache.LoadingCache;
  */
 public class GuavaClassCache implements ClassCache {
 
-    //  @formatter:off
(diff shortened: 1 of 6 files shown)


core/common-util/src/main/java/datawave/core/cache/GuavaClassCache.java
```diff
@@ -0,0 +1,40 @@
...
+import com.google.common.cache.LoadingCache;
+
+/**
+ * A guava {@link LoadingCache} implementation of a {@link ClassCache}
+ */
+public class GuavaClassCache implements ClassCache {
+
+    //  @formatter:off
+    private final LoadingCache<String,Class<?>> cache = CacheBuilder.newBuilder()
+            .maximumSize(128)
```
ivakegg: Same comments as the other PR.... set during construction?

apmoriarty: As with the other merge request, wasn't planning to make this configurable but certainly could. In practice this is used for tracking the implementations of an Attribute or Type and the sum of those is less than 128

core/common-util/src/main/java/datawave/core/cache/LRUMapClassCache.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    @Override
+    public String name() {
+        return "LRUMap";
+    }
+
+    @Override
+    public Class<?> get(String name) throws ClassNotFoundException {
+        Class<?> clazz = cache.get(name);
```
jschmidt10, line 30: Not a big deal but given that we're calling get + set separately, there could be a race condition here. It seems like that's probably fine, we just call the `Class.forName()` twice.

ivakegg (dismissed):

core/common-util/src/main/java/datawave/core/cache/LRUMapClassCache.java
apmoriarty: Ideally we never use this cache, it's here to act as a reference point for the other two implementations. Agree that it's probably fine.

core/common-util/src/test/java/datawave/core/cache/ClassCacheIT.java
```diff
@@ -0,0 +1,139 @@
...
+        ExecutorService executor = Executors.newFixedThreadPool(threads);
+
+        List<Callable<Long>> callables = new ArrayList<>();
+        for (int i = 0; i < threads; i++) {
+            final List<String> copy = new ArrayList<>(names);
+            Collections.shuffle(copy);
+            callables.add(() -> {
+                long total = 0L;
+                for (int k = 0; k < maxIterations; k++) {
+                    for (String name : names) {
```
jschmidt10:
```suggestion
                    for (String name : copy) {
```

warehouse/query-core/src/test/java/datawave/query/function/serializer/DocumentSerializationIT.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    @Test
+    @Order(4)
+    public void testHighThreadsHighVariance() {
+        driveThreads(highThreads, 100);
+    }
+
+    protected void driveThreads(int threads, int numDocs) {
+        ExecutorService executor = Executors.newFixedThreadPool(threads);
```
jschmidt10, line 90: We should call executor.shutdown() at the end to free the worker threads.

core/common-util/src/test/java/datawave/core/cache/ClassCacheIT.java
apmoriarty: Actually this is incorrect. Each callable should be operating on it's own copy of the class names, so this is correct. Fixed in next commit.

warehouse/query-core/src/test/java/datawave/query/function/serializer/DocumentSerializationIT.java
apmoriarty: Fixed in next commit

ivakegg (approved):

jschmidt10 (approved):

PR #3264, opened 2025-11-04
Make FieldExistenceRule check for hidden fields

foster33:
The query validation rule, `FieldExistenceRule`, will now check for hidden fields. If a field is hidden, it will be treated as non-existent.


core/utils/metadata-utils/src/test/java/datawave/query/util/MetadataHelperTest.java
```diff
@@ -104,6 +105,12 @@ private void givenMutation(String row, String columnFamily, String columnQualifi
         givenMutation(mutation);
     }
 
+    private void givenHiddenField(String row, Text colf, String datatype) {
```
jschmidt10: Since the colf must be `h` to be a hidden field, I would remove that argument and just hard-code the colf in this method.

core/utils/metadata-utils/src/main/java/datawave/query/util/MetadataHelper.java
```diff
@@ -717,6 +717,30 @@ public boolean isTokenized(String fieldName, Set<String> ingestTypeFilter) throw
...
+     *
+     * @param fieldName
+     *            the field name
+     * @param ingestTypeFilter
+     *            the ingest type filter
+     * @return true if the field is hidden for the provided ingest types
+     * @throws TableNotFoundException
+     *             if the table does not exist
+     */
+    public boolean isHidden(String fieldName, Set<String> ingestTypeFilter) throws TableNotFoundException {
```
jschmidt10: Can you add this TableNotFoundException to your catch block? Given we do not have valid fallback logic, we might as well bail on it too and that will clean up all of your Visitor methods below so we don't have to try/catch in so many places.

jschmidt10 (dismissed):

core/utils/metadata-utils/src/test/java/datawave/query/util/MetadataHelperTest.java
```diff
@@ -104,6 +104,13 @@ private void givenMutation(String row, String columnFamily, String columnQualifi
         givenMutation(mutation);
     }
 
+    private void givenHiddenField(String row, String datatype) {
+        Text h = new Text("h");
```
jschmidt10: Nit: You could still use `ColumnFamilyConstants.COLF_H` here but I approved. Maybe if you make changes for someone else throw that in?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldMissingFromSchemaVisitor.java
```diff
@@ -94,7 +94,7 @@ protected Object findMissingFields(JexlNode node, Object data) {
 
         for (ASTIdentifier identifier : identifiers) {
             String fieldName = JexlASTHelper.deconstructIdentifier(identifier);
-            if (!this.allFieldsForDatatypes.contains(fieldName) && !specialFields.contains(fieldName)) {
+            if ((!this.allFieldsForDatatypes.contains(fieldName) && !specialFields.contains(fieldName)) || helper.isHidden(fieldName, this.datatypeFilter)) {
```
lbschanno: The documentation for the colf `h` states that it denotes an event that should be hidden from display.  Are you certain that it should be treated as non-existent within the context of the FieldMissingFromSchemaVisitor? This visitor is used in `DefaultQueryPlanner.timedTestForNonExistentFields()`, and will impact results there. If that's intended and correct, then disregard this comment.

warehouse/query-core/src/test/java/datawave/query/rules/FieldExistenceRuleTest.java
```diff
@@ -95,4 +106,12 @@ protected ShardQueryRule getNewRule() {
         rule.setSpecialFields(fieldExceptions);
         return rule;
     }
+
```
lbschanno: Rather than use an AccumuloClient here, it would be simpler to instead mock out the expected interactions with the helper as you did in PR #3174. This way, FieldExistenceRuleTest does not need to care about the implementation details of MetadataHelper.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/FieldMissingFromSchemaVisitor.java
lbschanno: This could result in a call to `helper.isHidden()` for each field that we find in the tree. It would be more efficient to implement something like `helper.getAllHiddenFields()` that can be called just once to get the set of all unique hidden fields, store the fields in a set like we do for `allFieldsForDatatypes`, and then check against that set here.

lbschanno (changes requested):

core/utils/metadata-utils/src/main/java/datawave/query/util/MetadataHelper.java
```diff
@@ -717,6 +717,30 @@ public boolean isTokenized(String fieldName, Set<String> ingestTypeFilter) throw
...
+    /**
+     * Get the set of hidden fields for the provided ingest type filter. A null or empty filter indicates all hidden fields should be returned.
+     *
+     * @param ingestTypeFilter
+     *            the ingest type filter
+     * @return the set of hidden fields given the provided ingest type filter
+     * @throws TableNotFoundException
+     *             if the table does not exist
+     */
+    public Set<String> getHiddenFields(Set<String> ingestTypeFilter) throws TableNotFoundException {
```
lbschanno, line 730: It may be beneficial to make this method cacheable as well with `@Cacheable`.

core/utils/metadata-utils/src/main/java/datawave/query/util/AllFieldMetadataHelper.java
```diff
@@ -1108,6 +1108,36 @@ public Multimap<String,String> loadIndexedFields() throws TableNotFoundException
...
+    /**
+     * Fetch the Set of all fields marked as being hidden, {@link ColumnFamilyConstants#COLF_H}. Returns a multimap of datatype to field
+     *
+     * @return a multimap of datatypes to hidden fields
+     * @throws TableNotFoundException
+     *             if no table exists
+     */
+    @Cacheable(value = "loadHiddenFields", key = "{#root.target.auths,#root.target.metadataTableName}", cacheManager = "metadataHelperCacheManager",
+                    sync = true)
+    public Multimap<String,String> loadHiddenFields() throws TableNotFoundException {
```
lbschanno, line 1120: Can you add a unit test for this method?

lbschanno (changes requested):

lbschanno (approved):

jschmidt10 (approved):

PR #3270, opened 2025-11-10
Integrate class cache

apmoriarty:
- TypeFactory now uses a CaffeineCache
- Document, Attributes, Content use a ThreadLocal Caffeine cache found in the Attribute class
- Updated DocumentSerializationIT for better coverage

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
+++ b/warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
@@ -21,6 +21,8 @@ import com.esotericsoftware.kryo.KryoSerializable;
 import com.esotericsoftware.kryo.io.Input;
 import com.esotericsoftware.kryo.io.Output;
 
+import datawave.core.cache.CaffeineClassCache;
+import datawave.core.cache.ClassCache;
 import datawave.query.Constants;
 import datawave.query.jexl.DatawaveJexlContext;
 
@@ -40,6 +42,9 @@ public abstract class Attribute<T extends Comparable<T>> implements WritableComp
     // cache computation to avoid repeated calculation
     protected int hashcode = Integer.MIN_VALUE;
     protected long sizeInBytes = Long.MIN_VALUE;
+
+    // used by Document, Attributes and TypeAttribute
+    protected static final ThreadLocal<ClassCache> classCache = ThreadLocal.withInitial(CaffeineClassCache::new);
 
     public Attribute() {}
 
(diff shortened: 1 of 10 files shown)


warehouse/query-core/src/main/java/datawave/query/attributes/Content.java
```diff
@@ -157,12 +158,15 @@ public void read(Kryo kryo, Input input) {
...
             try {
-                sourceClass = Class.forName(clazz);
-                source = (Attribute<?>) sourceClass.newInstance();
-            } catch (ClassNotFoundException | InstantiationException | IllegalAccessException e) {
+                String className = input.readString();
+                Class<?> clazz = classCache.get().get(className);
+                if(Attribute.class.isAssignableFrom(clazz)) {
+                    source = (Attribute<?>) clazz.getDeclaredConstructor().newInstance();
+                } else {
+                    throw new RuntimeException("could not instantiate " + className);
```
jschmidt10: Maybe: `throw new RuntimeException(className " does not extend " + Attribute.getClass().getName());`

warehouse/query-core/src/main/java/datawave/query/attributes/Content.java
```diff
@@ -157,12 +158,15 @@ public void read(Kryo kryo, Input input) {
...
+                }
+            } catch (ClassNotFoundException | InstantiationException | IllegalAccessException | NoSuchMethodException | InvocationTargetException e) {
                 throw new RuntimeException("could not parse source", e);
```
jschmidt10: This runtime exception is really "We couldn't instantiate className".

warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
```diff
@@ -41,6 +43,9 @@ public abstract class Attribute<T extends Comparable<T>> implements WritableComp
     protected int hashcode = Integer.MIN_VALUE;
     protected long sizeInBytes = Long.MIN_VALUE;
 
+    // used by Document, Attributes and TypeAttribute
+    protected static final ThreadLocal<ClassCache> classCache = ThreadLocal.withInitial(CaffeineClassCache::new);
```
jschmidt10, line 47: Just curious why its thread local. Off the top of my head it sounds like it'd be better to share across all threads since we're still gonna end up doing reflection on each thread. Is it to remove contention?

apmoriarty: Correct. Even though the Caffeine cache won out on performance there was still a lot of contention in an environment with lots of threads. Making this ThreadLocal improved the time to parse each document and reduced the overall wall clock time.

jschmidt10 (dismissed):

lbschanno (dismissed):

foster33 (dismissed):

lbschanno (dismissed):

foster33 (approved):

lbschanno (approved):

PR #3273, opened 2025-11-12
Eliminate unnecessary column visibility creation

apmoriarty:
- Attribute has method that allows access to the raw ColumnVisibility bytes without reparsing the visibility
- Attribute has method that allows setting a ColumnVisibility from the raw bytes without reparsing the visibility

Practically this eliminates creating two ColumnVisibility objects when reading an Attribute's metadata.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
+++ b/warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
@@ -28,6 +28,8 @@ public abstract class Attribute<T extends Comparable<T>> implements WritableComp
 
     private static final Logger log = Logger.getLogger(Attribute.class);
     private static final Text EMPTY_TEXT = new Text();
+    private static final byte[] EMPTY_BYTES = new byte[0];
+    private static final ByteSequence EMPTY_BYTE_SEQUENCE = new ArrayByteSequence(new byte[0]);
 
     /**
      * The metadata for this attribute. Really only the column visibility and timestamp are preserved in this metadata when serializing and deserializing.
@@ -57,6 +59,18 @@ public abstract class Attribute<T extends Comparable<T>> implements WritableComp
             return metadata.getColumnVisibilityParsed();
         }
         return Constants.EMPTY_VISIBILITY;
+    }
+
+    /**
+     * Some callers only require the raw bytes
+     *
+     * @return the byte array that backs the column visibility
+     */
+    public byte[] getColumnVisibilityBytes() {
+        if (isMetadataSet()) {
+            return metadata.getColumnVisibilityData().getBackingArray();
+        }
+        return Constants.EMPTY_BYTES;
     }
 
     public void setColumnVisibility(ColumnVisibility columnVisibility) {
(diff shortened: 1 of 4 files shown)


warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
```diff
@@ -59,6 +61,18 @@ public ColumnVisibility getColumnVisibility() {
...
+    /**
+     * Some callers only require the raw bytes
+     *
+     * @return the byte array that backs the column visibility
+     */
+    public byte[] getColumnVisibilityBytes() {
+        if (isMetadataSet()) {
+            return metadata.getColumnVisibilityData().getBackingArray();
+        }
+        return Constants.EMPTY_BYTES;
```
jschmidt10: You defined your own EMPTY_BYTES above but then you pull this from Constants. I think either is fine - maybe a slight preference to just use the one in Constants and delete the one above.

warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
```diff
@@ -59,6 +61,18 @@ public ColumnVisibility getColumnVisibility() {
...
     }
 
+    /**
+     * Some callers only require the raw bytes
+     *
+     * @return the byte array that backs the column visibility
+     */
+    public byte[] getColumnVisibilityBytes() {
+        if (isMetadataSet()) {
+            return metadata.getColumnVisibilityData().getBackingArray();
```
jschmidt10: I think we want `toArray()` here. I'm pretty sure we've hit issues before where the internal byte[] is reused so you'd need to account for the starting offset (which `toArray()` does).

warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
apmoriarty: I like pulling from the existing Constants

warehouse/query-core/src/main/java/datawave/query/attributes/Attribute.java
apmoriarty: I started to look the difference between `getBackingArray` and `toArray`, and noticed a few unsettling things. Ultimately neither of us are doing the correct thing if we care about immutability.

For an ArrayByteSequence the offset will always be zero. The call for `Key.getColumnVisibilityData()` returns a new `ArrayByteSequence` using this constructor.
```java
  public ArrayByteSequence(byte[] data) {
    this.data = data;
    this.offset = 0;
    this.length = data.length;
  }
```

I looked at the differences between `ArrayByteSequence.getBackingArray()` and `ArrayByteSequence.toArray()` and found they return the same object. The same byte array can also be accessed via `Key.getColumnVisibilityParsed().getExpression()`. Would have thought that was an immutable call but I guess not.
```
ColumnVisibility cv = new ColumnVisibility("VIZ-A");
Key key =  new Key("row", "datatype\0uid", "", cv, 10L);

ByteSequence columnVisibilityData = key.getColumnVisibilityData();
byte[] backing = columnVisibilityData.getBackingArray();
byte[] toArray = columnVisibilityData.toArray();

ColumnVisibility parsedCV = key.getColumnVisibilityParsed();
byte[] parsedExpression = parsedCV.getExpression();

System.out.println("1: " + (backing == toArray));
System.out.println("2: " + (backing == parsedExpression));
System.out.println("3: " + (toArray == parsedExpression));
// prints 'true' each time
```

Another test to confirm that two separate ColumnVisibility objects share the same byte array...
```java
ColumnVisibility left = key.getColumnVisibilityParsed();
ColumnVisibility right = key.getColumnVisibilityParsed();
System.out.println("x: " + (left.getExpression() == right.getExpression()));
// this prints 'true'
```

Given that the ColumnVisibility is less immutable than expected, I think this actually accomplishes what we both want
```java
// get a copy of the backing byte array without re-parsing the ColumnVisibility
byte[] data = metadata.getColumnVisibilityData().getBackingArray();
return Arrays.copyOf(data, data.length);
```

jschmidt10: Thanks for digging. Yeah, i think in the case where a ColumnVisibility object only ever holds one value, that getBackingArray and toArray are equivalent. If we were to do a readFields() using the same object and the new value is shorter, then data.length would be less than the backing.length and using getBackingArray() would give us some garbage at the end of the array.

jschmidt10 (approved):

lbschanno (approved):

PR #3275, opened 2025-11-14
Re-allow normalization within an evaluation-only clause, and update the

ivakegg:
lossy-regex to avoid adding the lossy regex if already nested in an evaluation-only clause.


ivakegg:
A test case for the whindex bug that resulted in this PR needs to be created eventually: https://github.com/NationalSecurityAgency/datawave/issues/3277

apmoriarty (approved):

drewfarris (approved):

PR #3276, opened 2025-11-14
Adds endpoints for retrieving annotation source data

drewfarris:
* Adds `/DataWave/Annotations/v1/source/{analyticHash}` to look up analytic sources based on their analyticHash.
* Calls like `/DataWave/Annotations/v1/{idType}/{id}` will now lookup the their `analyticHash` in the `annotationSource` table, to include it in the returned annotation, if the source data is available.


warehouse/assemble/datawave/src/main/docker/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # https://github.com/NationalSecurityAgency/datawave-stack-docker-images
 FROM ghcr.io/nationalsecurityagency/datawave-stack-hadoop:${version.hadoop} AS hadoop
-FROM bitnami/zookeeper:${version.zookeeper} AS zookeeper
```
drewfarris, line 3: unrelated, but added to make Docker builds successful.

jalphonso (approved):

d-hwang (approved):

PR #3291, opened 2025-12-08
Set event sizeInBytes

apmoriarty:
Set an approximate event size for the DiscoveryTransformer and ShardQueryCountTableTransformer

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryTransformer.java
+++ b/warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryTransformer.java
@@ -80,6 +80,7 @@ public class DiscoveryTransformer extends BaseQueryLogicTransformer<DiscoveredTh
         }
 
         event.setFields(fields);
+        event.setSizeInBytes(fields.size() * 6L);
 
         Metadata metadata = new Metadata();
         metadata.setInternalId(""); // there is no UUID for a single index pointer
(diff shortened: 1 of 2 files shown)


warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryTransformer.java
```diff
@@ -80,6 +80,7 @@ public EventBase transform(DiscoveredThing thing) {
         }
 
         event.setFields(fields);
+        event.setSizeInBytes(fields.size() * 6L);
```
alerman, line 83: is 6 an average?

apmoriarty: It's a place holder until I figure out a better way to calculate the size.

apmoriarty: After revisiting this I believe it is okay to leave the code as-is. The transformer is used in conjunction with the ShardIndexQueryTable which only returns a single page. Because the size in bytes is used to determine when to paginate early there's no point in doing extra calculations here. 

I would like to see the size in bytes calculated dynamically as the event is built up. That would solve the entire problem.

foster33 (approved):

ivakegg (approved):

PR #3302, opened 2025-12-11
Force additional cache cleanup to avoid intermittent test failure

apmoriarty:
The fix is an additional call to `factory.cleanup()`

### DIFF
--- a/core/utils/type-utils/src/test/java/datawave/data/type/TypeFactoryTest.java
+++ b/core/utils/type-utils/src/test/java/datawave/data/type/TypeFactoryTest.java
@@ -52,7 +52,9 @@ public class TypeFactoryTest {
         // same type created in a row with a cache size of one will return the same type instance
         assertSame(left, right);
 
-        // trigger maintenance task to evict the LcType
+        // at this point the cache could contain both the LcType and the IpAddressType if the eviction thread
+        // has not run. Triggering the eviction thread will evict the oldest type (the LcType) leaving just
+        // the IpAddressType. This sets the condition for creating a new instance of the LcType below.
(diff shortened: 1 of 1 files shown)


core/utils/type-utils/src/test/java/datawave/data/type/TypeFactoryTest.java
```diff
@@ -47,13 +47,13 @@ public void testTypeFactoryCustomSize() {
...
-        }
+        Type<?> left = factory.createType(IpAddressType.class.getName());
+        Type<?> right = factory.createType(IpAddressType.class.getName());
+        // same type created in a row with a cache size of one will return the same type instance
+        assertSame(left, right);
+
+        // trigger maintenance task to evict the LcType
+        factory.cleanup();
 
         // creating a new LcType should return a new instance due to the low cache size
```
ivakegg, line 60: Is this reason correct?  I think its because you have a fresh factory.

apmoriarty: The reason is correct, but could be clarified. The most recent type is the IpAddressType, so when the eviction tasks are run it will evict the oldes type, which is the LcType. Updated comment coming.

ivakegg: ah, ok

ivakegg: misunderstood what factory.cleanup does

ivakegg (approved):

hlgp (approved):

PR #3318, opened 2025-12-23
Replace non-public TabletClosedException and ScanTimedOutException

### LINKED ISSUE #3317: Replace non-public TabletClosedException and ScanTimedOutException
## Summary
Replace direct usage of non-public Accumulo exceptions with class name checks for Accumulo 4 compatibility.

## Non-Public Classes
- `org.apache.accumulo.tserver.tablet.TabletClosedException`
- `org.apache.accumulo.core.clientImpl.ThriftScanner.ScanTimedOutException`

## Files Affected
- `QueryIterator.java` - TabletClosedException
- `DocIdQueryIterator.java` - TabletClosedException
- `DocumentIterator.java` - TabletClosedException
- `Scan.java` - ScanTimedOutException

## Approach
Replace `instanceof` checks with class name string comparison:
```java
private static boolean isTabletClosedException(Throwable t) {
    return t != null && t.getClass().getName().equals(
        "org.apache.accumulo.tserver.tablet.TabletClosedException");
}
```

Part of #2443

SethSmucker:
## Summary
- Replace `instanceof TabletClosedException` with class name check
- Replace `catch (ScanTimedOutException)` with class name check inside generic catch block

## Files Changed
- `QueryIterator.java` - TabletClosedException
- `DocIdQueryIterator.java` - TabletClosedException  
- `DocumentIterator.java` - TabletClosedException
- `Scan.java` - ScanTimedOutException

## Approach
Since these are non-public Accumulo classes that won't exist in Accumulo 4, we use class name comparison:
```java
private static boolean isTabletClosedException(Throwable t) {
    return t != null && t.getClass().getName().equals(
        "org.apache.accumulo.tserver.tablet.TabletClosedException");
}
```

Fixes #3317
Part of #2443


warehouse/query-core/src/main/java/datawave/query/tables/async/Scan.java
```diff
@@ -289,10 +288,13 @@ public Scan call() throws Exception {
...
             } while (!finished());
-        } catch (ScanTimedOutException e) {
-            // this is okay. This means that we are being timesliced.
-            myScan.addRange(currentRange);
         } catch (Exception e) {
+            // Check for ScanTimedOutException by class name to avoid non-public API import
+            if (isScanTimedOutException(e)) {
+                // this is okay. This means that we are being timesliced.
+                myScan.addRange(currentRange);
+                return this;
```
foster33: Should we be returning this here? Just curious because it was not in the original catch block you removed

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -576,22 +575,29 @@ else if (documentRange != null && (!this.isContainsIndexOnlyTerms() && this.getT
      * @throws IOException
      *             for read/write issues
      */
+    /**
+     * Check if the throwable is a TabletClosedException by class name to avoid importing non-public Accumulo API.
```
lbschanno: I think it would be better for this function (along with the one for ScanTimedOutException to be extracted to a utility class so that we can avoid repeating code and better document why we're using string class names to check the class type.

apmoriarty (approved):

ddanielr (changes requested):
Sorry but I would not recommend merging these changes.

While this current approach would work to eliminate the import, it's extremely fragile and will cause issues later.

#### main concern
This code is trying to ensure that if specific accumulo exceptions have been thrown they are raised to accumulo for proper handling instead of being handled by datawave code.

When/If these exception classes change in the future then these checks will always return false and accumulo exceptions will be swallowed during runtime.

#### Possible Alternative
A better approach might be to consolidate these accumulo exception type checks to a single class with a `checkExceptionType()` method and then just adding import control statements to limit the use of these exceptions to that class. 

This doesn't eliminate the import, but for this specific case I believe that's a better approach as it results in the failure condition being discovered at compile time instead of runtime.

warehouse/query-core/src/main/java/datawave/query/tables/async/Scan.java
```diff
@@ -308,6 +310,13 @@ public Scan call() throws Exception {
 
     }
 
+    /**
+     * Check if the throwable is a ScanTimedOutException by class name to avoid importing non-public Accumulo API.
+     */
+    private boolean isScanTimedOutException(Throwable t) {
+        return t != null && t.getClass().getName().equals("org.apache.accumulo.core.clientImpl.ThriftScanner$ScanTimedOutException");
```
ddanielr, line 317: This class name doesn't look correct.

PR #3325, opened 2025-12-29
Add MAC test utility

apmoriarty:
Add utility for interacting with MiniAccumuloCluster and integrate into several tests

### DIFF
--- a/warehouse/query-core/src/test/java/datawave/query/util/ShapesIngest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/util/ShapesIngest.java
@@ -6,9 +6,9 @@ import static datawave.util.TableName.SHARD_INDEX;
 import static datawave.util.TableName.SHARD_RINDEX;
 
 import java.util.Date;
-import java.util.HashSet;
+import java.util.HashMap;
 import java.util.List;
-import java.util.Set;
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.accumulo.core.client.AccumuloClient;
@@ -142,18 +142,16 @@ public class ShapesIngest {
         tops.create(SHARD_RINDEX);
         tops.create(METADATA);
 
-        Set<String> additions = new HashSet<>();
+        Map<String,String> additions = new HashMap<>();
         IteratorUtil.IteratorScope[] scopes = IteratorUtil.IteratorScope.values();
         for (IteratorUtil.IteratorScope scope : scopes) {
             String name = "table.iterator." + scope.name() + ".UIDAggregator";
             String opt = "table.iterator." + scope.name() + ".UIDAggregator.opt.*";
 
-            tops.setProperty(SHARD_INDEX, name, "19,datawave.iterators.TotalAggregatingIterator");
-            tops.setProperty(SHARD_INDEX, opt, "datawave.ingest.table.aggregator.KeepCountOnlyUidAggregator");
-            additions.add(name);
-            additions.add(opt);
+            additions.put(name, "19,datawave.iterators.TotalAggregatingIterator");
+            additions.put(opt, "datawave.ingest.table.aggregator.KeepCountOnlyUidAggregator");
         }
-        MacTestUtil.waitForPropertyAddition(tops, SHARD_INDEX, additions);
+        MacTestUtil.addPropertiesAndWait(tops, SHARD_INDEX, additions);
 
         // grant root user all auths so they can scan the tables
         client.securityOperations().changeUserAuthorizations("root", new Authorizations("ALL"));
--- a/warehouse/query-core/src/test/java/datawave/query/util/ShapesIngest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/util/ShapesIngest.java
@@ -5,6 +5,7 @@ import static datawave.util.TableName.SHARD;
 import static datawave.util.TableName.SHARD_INDEX;
 import static datawave.util.TableName.SHARD_RINDEX;
 
+import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
@@ -29,6 +30,7 @@ import datawave.data.type.ListType;
 import datawave.data.type.NumberType;
 import datawave.ingest.protobuf.Uid;
 import datawave.query.index.day.IndexIngestUtil;
+import datawave.test.MacTestUtil;
 import datawave.util.TableName;
 
 /**
@@ -139,14 +141,18 @@ public class ShapesIngest {
         tops.create(SHARD_RINDEX);
         tops.create(METADATA);
 
+        List<String> additions = new ArrayList<>();
         IteratorUtil.IteratorScope[] scopes = IteratorUtil.IteratorScope.values();
         for (IteratorUtil.IteratorScope scope : scopes) {
             String name = "table.iterator." + scope.name() + ".UIDAggregator";
             String opt = "table.iterator." + scope.name() + ".UIDAggregator.opt.*";
 
-            client.tableOperations().setProperty(SHARD_INDEX, name, "19,datawave.iterators.TotalAggregatingIterator");
-            client.tableOperations().setProperty(SHARD_INDEX, opt, "datawave.ingest.table.aggregator.KeepCountOnlyUidAggregator");
+            tops.setProperty(SHARD_INDEX, name, "19,datawave.iterators.TotalAggregatingIterator");
+            tops.setProperty(SHARD_INDEX, opt, "datawave.ingest.table.aggregator.KeepCountOnlyUidAggregator");
+            additions.add(name);
+            additions.add(opt);
         }
+        MacTestUtil.waitForPropertyAddition(tops, SHARD_INDEX, additions);
 
         // grant root user all auths so they can scan the tables
         client.securityOperations().changeUserAuthorizations("root", new Authorizations("ALL"));
--- a/warehouse/query-core/src/test/java/datawave/query/util/ShapesIngest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/util/ShapesIngest.java
@@ -5,9 +5,10 @@ import static datawave.util.TableName.SHARD;
 import static datawave.util.TableName.SHARD_INDEX;
 import static datawave.util.TableName.SHARD_RINDEX;
 
-import java.util.ArrayList;
 import java.util.Date;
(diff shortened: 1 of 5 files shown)


warehouse/query-core/src/test/java/datawave/test/MacTestUtil.java
```diff
@@ -0,0 +1,164 @@
...
+     */
+    public static void createOrRecreate(TableOperations tops, String tableName) {
+        try {
+            if (tops.exists(tableName)) {
+                tops.delete(tableName);
+            }
+            tops.create(tableName);
+        } catch (AccumuloException | AccumuloSecurityException | TableNotFoundException | TableExistsException e) {
+            fail("Failed to delete/create table");
+            throw new RuntimeException(e);
```
apmoriarty: Runtime exception will catch tests that do not use JUnit5

ivakegg: How about
throw new RuntimeException("Failed to delete/create table", e);

warehouse/query-core/src/test/java/datawave/test/MacTestUtil.java
```diff
@@ -0,0 +1,164 @@
...
+                    if (properties.contains(prop.getKey())) {
+                        allRemoved = false;
+                        break;
+                    }
+                }
+            }
+            long elapsed = System.currentTimeMillis() - start;
+            log.trace("removed {} properties in {} ms", properties.size(), elapsed);
+        } catch (AccumuloException | TableNotFoundException e) {
+            throw new RuntimeException("Exception while verifying property removal", e);
```
ivakegg, line 115: did you want to try fail(...) first here ?

warehouse/query-core/src/test/java/datawave/test/MacTestUtil.java
```diff
@@ -0,0 +1,164 @@
...
+                    }
+                }
+                Collections.sort(additions);
+                Collections.sort(properties);
+                allAdded = additions.equals(properties);
+            }
+            long elapsed = System.currentTimeMillis() - start;
+            log.trace("added {} properties in {} ms", properties.size(), elapsed);
+        } catch (AccumuloException | TableNotFoundException e) {
+            throw new RuntimeException("Exception while verifying property addition", e);
```
ivakegg, line 170: did you want to try fail(...) first here?

warehouse/query-core/src/test/java/datawave/test/MacTestUtil.java
```diff
@@ -0,0 +1,166 @@
...
+     *            the table name
+     */
+    public static void createOrRecreate(TableOperations tops, String tableName) {
+        try {
+            if (tops.exists(tableName)) {
+                tops.delete(tableName);
+            }
+            tops.create(tableName);
+        } catch (AccumuloException | AccumuloSecurityException | TableNotFoundException | TableExistsException e) {
+            fail("Failed to delete/create table");
```
jschmidt10: I think `fail()` will throw an AssertException or something like that so the RuntimeExceptions will never be reached. But given that this utility isn't about assertions - it's more about shortcuts for Accumulo - I would just throw the runtime exceptions here and remove all the `fail()`s.

warehouse/query-core/src/test/java/datawave/test/MacTestUtil.java
```diff
@@ -0,0 +1,166 @@
...
+     *            the table name
+     * @param properties
+     *            the list of removed properties
+     */
+    public static void waitForPropertyAddition(TableOperations tops, String tableName, List<String> properties) {
+        try {
+            long start = System.currentTimeMillis();
+            boolean allAdded = false;
+            while (!allAdded) {
+                List<String> additions = new ArrayList<>();
```
jschmidt10: If you used Set instead of List, you could just check for `additions.equals(properties)` without the sort calls.

warehouse/query-core/src/test/java/datawave/test/MacTestUtil.java
```diff
@@ -0,0 +1,166 @@
...
+     *            the table name
+     * @param properties
+     *            the list of removed properties
+     */
+    public static void waitForPropertyRemoval(TableOperations tops, String tableName, List<String> properties) {
+        try {
+            long start = System.currentTimeMillis();
+            boolean allRemoved = false;
+            while (!allRemoved) {
+                allRemoved = true;
```
jschmidt10, line 103: These waiting while loops are going to be "busy waits". There's nothing in the loop that will yield the CPU. I probably would inject a sleep, even if its only a few millis. Do you have a feel for how many iterations we hit in these loops?

warehouse/query-core/src/test/java/datawave/test/MacTestUtil.java
```diff
@@ -0,0 +1,166 @@
...
+     * assumptions.
+     *
+     * @param tops
+     *            an instance of {@link TableOperations}
+     * @param tableName
+     *            the table name
+     * @param properties
+     *            the list of removed properties
+     */
+    public static void waitForPropertyAddition(TableOperations tops, String tableName, List<String> properties) {
```
jschmidt10: Just a proposal - from an API point of view, a lot of your methods will take the TableOperations and do the handy stuff for you - like the create/recreate table method. Instead of having the users directly use `tops` to set/remove properties and then have to remember to call these wait methods, we could pass in the additions/deletions into these methods and do both the adding/removing + waiting. 

Something like:

```
Map<String,String> additions = ...
MacTestUtil.addPropertiesAndWait(tops, tableName, additions);
```

warehouse/query-core/src/test/java/datawave/test/MacTestUtil.java
apmoriarty: When turning up the log level on a MAC test the time is between 0 and 5 ms (on my machine) and always completes in one cycle. Added additional logging in the next commit to track how many loops it takes for all property changes to take effect.

warehouse/query-core/src/test/java/datawave/test/MacTestUtil.java
apmoriarty: I like this.

warehouse/query-core/src/test/java/datawave/test/MacTestUtil.java
jschmidt10: Okay, dont sweat it then. If it was iterating a bunch, then I was worried it might peg a CPU doing nothing but if it's only a few loops, then we should be okay.

apmoriarty: The only time we might see more than one cycle is when a test machine is under load, but even then the trip across the network to grab the properties out of ZK act like a functional sleep statement.

warehouse/query-core/src/test/java/datawave/test/MacTestUtil.java
```diff
@@ -0,0 +1,203 @@
...
+     */
+    public static void addPropertiesAndWait(TableOperations tops, String tableName, Map<String,String> properties) {
+        for (Map.Entry<String,String> prop : properties.entrySet()) {
+            try {
+                tops.setProperty(tableName, prop.getKey(), prop.getValue());
+            } catch (AccumuloException | AccumuloSecurityException e) {
+                throw new RuntimeException("Failed to add property", e);
+            }
+        }
+        waitForPropertyAddition(tops, tableName, properties.keySet());
```
ivakegg, line 137: Wow!  is the setProperty mechanism actually async relative to top.getProperties?

apmoriarty: When using MiniAccumuloCluster, yes. It can take a moment for the property change to persist in the tablet server and ZK. I've had a few tests fail due to an iterator or combiner not getting set before a scan.

warehouse/query-core/src/test/java/datawave/test/index/NoUidIndexTest.java
```diff
@@ -264,13 +267,18 @@ protected String getCloneTableName() {
...
     protected void configureClonedTable(TableOperations tops, String tableName) {
         try {
+            Set<String> additions = new HashSet<>();
             for (var scope : IteratorUtil.IteratorScope.values()) {
                 String name = "table.iterator." + scope.name() + ".agg";
                 String opt = "table.iterator." + scope.name() + ".agg.opt.*";
 
                 tops.setProperty(getCloneTableName(), name, "19,datawave.iterators.TotalAggregatingIterator");
                 tops.setProperty(getCloneTableName(), opt, "datawave.ingest.table.aggregator.KeepCountOnlyNoUidAggregator");
+                additions.add(name);
```
ivakegg: So why are we not filling a hashmap and then calling addPropAndWait ?  Also if you are setting the properties on the clone table name, then why are we waiting for properties on the passed in table name?

warehouse/query-core/src/test/java/datawave/query/util/ShapesIngest.java
```diff
@@ -139,14 +142,18 @@ public static void writeData(AccumuloClient client, RangeType type) throws Excep
...
+        Set<String> additions = new HashSet<>();
         IteratorUtil.IteratorScope[] scopes = IteratorUtil.IteratorScope.values();
         for (IteratorUtil.IteratorScope scope : scopes) {
             String name = "table.iterator." + scope.name() + ".UIDAggregator";
             String opt = "table.iterator." + scope.name() + ".UIDAggregator.opt.*";
 
-            client.tableOperations().setProperty(SHARD_INDEX, name, "19,datawave.iterators.TotalAggregatingIterator");
-            client.tableOperations().setProperty(SHARD_INDEX, opt, "datawave.ingest.table.aggregator.KeepCountOnlyUidAggregator");
+            tops.setProperty(SHARD_INDEX, name, "19,datawave.iterators.TotalAggregatingIterator");
+            tops.setProperty(SHARD_INDEX, opt, "datawave.ingest.table.aggregator.KeepCountOnlyUidAggregator");
```
ivakegg: Why not use the hashmap and then call addPropAndWait instead?

warehouse/query-core/src/test/java/datawave/test/index/NoUidIndexTest.java
apmoriarty: That's in the next commit

jschmidt10 (approved):

ivakegg (approved):

PR #3330, opened 2026-01-05
Feature/all keywords

FineAndDandy:


warehouse/query-core/src/main/java/datawave/query/transformer/annotation/BoundaryComparator.java
```diff
@@ -0,0 +1,46 @@
+package datawave.query.transformer.annotation;
+
+import java.util.Comparator;
+
+import datawave.annotation.protobuf.v1.SegmentBoundary;
+
+/**
+ * Sort by ascending boundary type value > 1 first Sort by descending boundary type < 1 always after all boundary type greater than 0 Secondary sort for equal
```
drewfarris: The bare `<` and `>` will cause javadoc errors, replace with `&gt;` and `&lt;` - Consider giving an example using the names first (e.g., "the intent is to have the UNKONWN boundary type always sort to last..")

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/SegmentValueComparator.java
```diff
@@ -0,0 +1,23 @@
+package datawave.query.transformer.annotation;
+
+import java.util.Comparator;
+
+import datawave.annotation.protobuf.v1.SegmentValue;
+
+/**
+ * Sort ascending by score, then by value
+ */
+public class SegmentValueComparator implements Comparator<SegmentValue> {
```
drewfarris: Consider renaming to `SegmentValueByScoreComparator`?

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/model/TermHit.java
drewfarris, line 1: `Term` and `TermHit` are nearly identical - do you need both? Is there any advantage in just using a boolean `isHit` flag? Maybe just having a `Term` object and membership in the `oneBestContext` vs. `termHits` is sufficient to distinguish the role of the term?

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -776,6 +785,26 @@ private void addConfigBasedTransformers() throws QueryException {
...
+                AnnotationDataAccess annotationDataAccess = new AnnotationDataAccess(this.getConfig().getClient(), getConfig().getAuthorizations(),
+                                getAnnotationTableName(), getAnnotationSourceTableName(), annotationSerializer, annotationSourceSerializer);
+                AllHitsFactory allHitsFactory = null;
+                try {
+                    Class<?> annotationHitsFactoryClass = Class.forName(getAnnotationHitsFactoryClass());
+                    Class<? extends AllHitsFactory> subClass = annotationHitsFactoryClass.asSubclass(AllHitsFactory.class);
+                    allHitsFactory = subClass.getDeclaredConstructor().newInstance();
+                } catch (ClassNotFoundException | InvocationTargetException | InstantiationException | IllegalAccessException | NoSuchMethodException e) {
+                    throw new QueryException("cannot instantiate allHitsFactory", e);
+                }
```
drewfarris: Consider encapsulating this set of lines in a method, so you can just simply call something like `getAnnotationHitsFactory()` in the call to the AnnotationHitsTransformer constructor?

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -776,6 +785,26 @@ private void addConfigBasedTransformers() throws QueryException {
             ((DocumentTransformer) this.transformerInstance).setProjectFields(getConfig().getProjectFields());
             ((DocumentTransformer) this.transformerInstance).setDisallowlistedFields(getConfig().getDisallowlistedFields());
 
+            if (getConfig().isAnnotationHitsEnabled()) {
+                // since this may be called multiple times always rebuild
+                // TODO may need to inject options to serializers
+                AnnotationSerializer<Iterator<Entry<Key,Value>>,Annotation> annotationSerializer = new AccumuloAnnotationSerializer();
+                AnnotationSerializer<Iterator<Entry<Key,Value>>,AnnotationSource> annotationSourceSerializer = new AccumuloAnnotationSourceSerializer();
+                AnnotationDataAccess annotationDataAccess = new AnnotationDataAccess(this.getConfig().getClient(), getConfig().getAuthorizations(),
+                                getAnnotationTableName(), getAnnotationSourceTableName(), annotationSerializer, annotationSourceSerializer);
```
drewfarris: Oh man, this makes me want to refactor `AnnotationDataAccess` construction to be a bit more friendly.

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AnnotationHitsTransformer.java
```diff
@@ -0,0 +1,389 @@
...
+     * Lookup the annotation for a document and enrich all hits if it exists
+     *
+     * @param keyDocumentEntry
+     *            the function argument
+     * @return
+     */
+    @Nullable
+    @Override
+    public Entry<Key,Document> apply(@Nullable Entry<Key,Document> keyDocumentEntry) {
+        if (!this.enabled || this.searchHitTerms == null || this.searchHitTerms.isEmpty()) {
```
drewfarris: It's strange that `keyDocumentEntry` is annotated with `@Nullable`, but we'll need a check for null as a result. In the suggestion below I swapped to `CollectionUtils.isEmpty(..)` to make the if statement more concise. I won't be disappointed if you break this into multiple if statement either.

```suggestion
        if (!this.enabled || keyDocumentEntry == null || CollectionUtils.isEmpty(searchHitTerms)) {
```

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AnnotationHitsTransformer.java
```diff
@@ -0,0 +1,389 @@
...
+            // no search terms, no-op
+            return keyDocumentEntry;
+        }
+
+        Key key = keyDocumentEntry.getKey();
+        if (key != null) {
```
drewfarris: Minor style nit. Instead of nesting a bunch of checks, consider checking for the error case and then bailing out early, e.g.:

```
if (key == null) {
  // handle null key and return or throw
}

// parse out the key parts..
String shard = key.getRow().toString();
String cf = key.getColumnFamily().toString();
String[] parts = cf.split("\u0000");

if (parts.length != 2) {
   // handle malformed cf error and return or throw.
}

// finish up.. 
String dataType = parts[0];
String uid = parts[1];
List<Annotation> annotations = annotationDataAccess.getAnnotations(shard, dataType, uid);
for (Annotation annotation : annotations) {
 // ...
 if (!validTypes.contains(annotationType)) {
    /// handle invalid annotation type and continue to next annotation?
  }

  // handle valid annotation types..
}
```

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AnnotationHitsTransformer.java
```diff
@@ -0,0 +1,389 @@
...
+                // TODO
+            }
+        } else {
+            // TODO
+        }
+
+        return keyDocumentEntry;
+    }
+
+    private List<AllHits> getCurrentAllHitsValue(Entry<Key,Document> entry) {
```
drewfarris: consider moving this function getCurrentAllHitsValue below updateDocument, b/c updateDocument is called from the apply method directory and then in turn calls getCurrentAllHitsValue - slightly easier for readability.

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AnnotationHitsTransformer.java
```diff
@@ -0,0 +1,389 @@
...
+        Content attr = (Content) entry.getValue().get(this.targetField);
+        List<AllHits> rollup = null;
+        if (attr != null) {
+            try {
+                rollup = this.objectMapper.readValue(attr.getContent(), new TypeReference<>() {});
+            } catch (JsonProcessingException e) {
+                throw new RuntimeException(e);
+            }
+        }
```
drewfarris: - Minor consistency nit, consider avoiding using `this` unless it's required? (see suggestion)
- Consider doing more checking to ensure that the `Document.get()` call doesn't return some weird Attribute that's not Content? 
- Consider avoiding transient documents where `isToKeep()` returns false.

```suggestion
        Content attr = (Content) entry.getValue().get(targetField);
        List<AllHits> rollup = null;
        if (attr != null) {
            try {
                rollup = objectMapper.readValue(attr.getContent(), new TypeReference<>() {});
            } catch (JsonProcessingException e) {
                throw new RuntimeException(e);
            }
        }
```

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AnnotationHitsTransformer.java
```diff
@@ -0,0 +1,389 @@
...
+     *
+     * @param segments
+     * @return
+     */
+    private TreeMap<SegmentBoundary,List<SegmentValue>> sort(List<Segment> segments) {
+        TreeMap<SegmentBoundary,List<SegmentValue>> orderedSegments = new TreeMap<>(new BoundaryComparator());
+        for (Segment segment : segments) {
+            // make a copy so they can be sorted
+            List<SegmentValue> segmentValues = new ArrayList<>(segment.getValuesList());
+            Collections.sort(segmentValues, new SegmentValueComparator());
```
drewfarris: Slightly more concise...
```suggestion
            segmentValues.sort(new SegmentValueComparator());
```

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AnnotationHitsTransformer.java
```diff
@@ -0,0 +1,389 @@
...
+
+            List<SegmentValue> values = sortedSegments.get(boundary);
+            for (int i = 0; i < values.size(); i++) {
+                SegmentValue segmentValue = values.get(i);
+                String normalizedTerm = normalize(segmentValue.getValue());
+                if (segmentValue.getScore() >= minScore && searchHitTerms.contains(normalizedTerm)) {
+                    // partial hits index is the location in the window where the hit is complete
+                    List<SegmentHit> hits = partialHits.computeIfAbsent(segmentIndex + contextSize, x -> new ArrayList<>());
+                    hits.add(new SegmentHit(window.getFirst(), boundary, i));
+                }
```
drewfarris: - The `search` method could use just a tad more explanation about how it works. I had to stare at it a bit to figure out why `segmentIndex + contextSize` make sense, and I still need to spend some time playing with the tests to fully understand it.
- In theory we could avoid calling `normalize` if the score isn't large enough, but I'm not certain I like the result b/c the if statement is pretty complex.
- We could use a method reference in the `computeIfAbsentCall`.

```suggestion
                if (segmentValue.getScore() >= minScore && searchHitTerms.contains(normalize(segmentValue.getValue()))) {
                    // partial hits index is the location in the window where the hit is complete
                    List<SegmentHit> hits = partialHits.computeIfAbsent(segmentIndex + contextSize, ArrayList::new);
                    hits.add(new SegmentHit(window.getFirst(), boundary, i));
                }
```

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AnnotationHitsTransformer.java
```diff
@@ -0,0 +1,389 @@
...
+        return finishedHits;
+    }
+
+    private Set<String> extractSearchHitTerms(Query settings) throws ParseException {
+        Set<String> searchTerms = new HashSet<>();
+
+        String query = settings.getQuery();
+        if (query != null) {
+            ASTJexlScript script = JexlASTHelper.parseJexlQuery(query);
+            List<ASTEQNode> eqNodes = JexlASTHelper.getEQNodes(script);
```
drewfarris: Do we need to support anything other than EQ nodes? (e.g., ranges, wildcard, regexps, functions)

warehouse/query-core/src/test/java/datawave/query/transformer/annotation/AllHitsFactoryTest.java
```diff
@@ -0,0 +1,277 @@
...
+    @Before
+    public void setup() {
+        factory = new AllHitsFactory();
+        objectMapper = new ObjectMapper();
+    }
+
+    @Test
+    public void noHitsTest() throws JsonProcessingException, AllHitsException {
+        AllHits result = factory.create("123", List.of(), new TreeMap<>(new BoundaryComparator()));
+        assertEquals(objectMapper.writeValueAsString(new AllHits()), objectMapper.writeValueAsString(result));
```
drewfarris: Consider making `objectMapper.writeValueAsString(new AllHits())` a constant like `EMPTY_HITS_VALUE` and calcuating it once.

warehouse/query-core/src/test/java/datawave/query/transformer/annotation/AllHitsFactoryTest.java
```diff
@@ -0,0 +1,277 @@
+package datawave.query.transformer.annotation;
+
+import static org.junit.Assert.assertEquals;
+
+import java.util.List;
+import java.util.TreeMap;
+
+import org.junit.Before;
+import org.junit.Test;
```
drewfarris: Consider Junit5 instead of Junit4? (e.g., `org.junit.jupiter.api.Test`). It will change the names of the test annotations and the methods for asserting that exceptions are thown (e.g., Before -> BeforeEach)

and..

```
    @Test
    void testExceptionIsThrown() {
        // Call assertThrows with the expected exception type and a lambda
        Exception exception = assertThrows(IllegalArgumentException.class, () -> {
            throwExceptionMethod();
        });

        // Optionally, assert the exception message or other properties
        String expectedMessage = "Invalid argument provided";
        String actualMessage = exception.getMessage();

        assertEquals(expectedMessage, actualMessage);
    }
```

drewfarris (commented):
This is looking good - it's pretty straightforward, but I need to spend a little more time with it to fully understand how parts work. Nevertheless here are a few of the comments so far, many nits; overall the design looks great.

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
```diff
@@ -558,6 +559,36 @@ public class ShardQueryConfiguration extends GenericQueryConfiguration implement
...
+
+    /**
+     * field to write all hits data to
+     */
+    private String annotationHitsTargetField;
+
+    private String annotationHitsFactoryClass = AllHitsFactory.class.getCanonicalName();
+
+    private String annotationTableName = "annotation";
+    private String annotationSourceTableName = "annotationSource";
```
apmoriarty: Consider extracting annotation specific variables in their own config object like the DocumentScannerConfig, or have an AnnotationQueryConfig that extends ShardQueryConfig.

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/model/AllHitsError.java
```diff
@@ -0,0 +1,19 @@
+package datawave.query.transformer.annotation.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * AllHits that includes an additional errorMessage field to communicate problems
```
apmoriarty, line 6: "An optional error message field to communicate that a problem occurred"

warehouse/query-core/src/test/java/datawave/query/transformer/annotation/AllHitsFactoryTest.java
apmoriarty: Second this. Please use JUnit5 for new code as much as possible.

warehouse/query-core/src/test/java/datawave/query/transformer/annotation/AllHitsFactoryTest.java
```diff
@@ -0,0 +1,277 @@
...
+        context.put(hitBoundary, List.of());
+        factory.create("123", List.of(hit), context);
+    }
+
+    @Test(expected = AllHitsException.class)
+    public void hitWithOutOfBoundsIndexTest() throws AllHitsException {
+        SegmentBoundary hitBoundary = SegmentBoundary.newBuilder().setBoundaryType(BoundaryType.TIME_MILLI).setStart(3).setEnd(4).build();
+        AnnotationHitsTransformer.SegmentHit hit = new AnnotationHitsTransformer.SegmentHit(hitBoundary, hitBoundary, 1);
+        hit.setContextEnd(hitBoundary);
+        TreeMap<SegmentBoundary,List<SegmentValue>> context = new TreeMap<>(new BoundaryComparator());
```
apmoriarty, line 102: Consider extracting this to a factory-like method

```java
private TreeMap<> createContext() {
    return new TreeMap<>(new BoundaryComparator());
}
```

warehouse/query-core/src/test/java/datawave/query/tables/ShardQueryLogicTest.java
```diff
@@ -411,6 +527,705 @@ public void testExcludeDataTypesIncludeOneTypeExcludeOneType() throws Exception
...
+
+        givenQuery("UUID=='CAPONE'");
+        givenStartDate("20091231");
+        givenEndDate("20150101");
+
+        expectNoField(WiseGuysIngest.caponeUID, "ALL_HITS_RESULTS");
+
+        Set<Set<String>> expected = new HashSet<>();
+        expected.add(Sets.newHashSet("UID:" + WiseGuysIngest.caponeUID));
+        runTestQuery(expected);
```
apmoriarty, line 1029: Nice and declarative, I like it

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AnnotationHitsTransformer.java
```diff
@@ -0,0 +1,403 @@
...
+     *
+     * @param segments
+     * @return
+     */
+    private TreeMap<SegmentBoundary,List<SegmentValue>> sort(List<Segment> segments) {
+        TreeMap<SegmentBoundary,List<SegmentValue>> orderedSegments = new TreeMap<>(new BoundaryComparator());
+        for (Segment segment : segments) {
+            // make a copy so they can be sorted
+            List<SegmentValue> segmentValues = new ArrayList<>(segment.getValuesList());
+            Collections.sort(segmentValues, new SegmentValueComparator());
```
apmoriarty: private final Comparator instead of a new instance each time sort is called.

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AnnotationHitsTransformer.java
apmoriarty: I have a strong preference for avoiding deep nesting wherever possible.

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AnnotationHitsTransformer.java
```diff
@@ -0,0 +1,403 @@
...
+                    }
+                }
+            }
+        }
+
+        return searchTerms;
+    }
+
+    private String normalize(String toNormalize) {
+        return toNormalize.toLowerCase().trim();
```
apmoriarty: Is this a "just in case" safety check or is there a failure to normalize the query during planning?

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AnnotationHitsTransformer.java
```diff
@@ -0,0 +1,403 @@
...
+import datawave.query.attributes.Content;
+import datawave.query.attributes.Document;
+import datawave.query.config.ShardQueryConfiguration;
+import datawave.query.jexl.JexlASTHelper;
+import datawave.query.transformer.DocumentTransform;
+import datawave.query.transformer.annotation.model.AllHits;
+import datawave.query.transformer.annotation.model.AllHitsError;
+
+/**
+ * This iterator will lookup and search annotations for hits as well as provide context
```
apmoriarty: Sir, this is a ~~Wendy's~~ transformer

apmoriarty (commented):
Overall design looks good to me, a few stylistic preferences but nothing major. Will defer to @drewfarris on what's an actual blocker for merge.

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/model/TermHit.java
FineAndDandy: I think this may be possible. I'm trying to adhere to an existing data model, and I may have cut enough from that model that there is no longer a reason for these to be different

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AnnotationHitsTransformer.java
FineAndDandy: I had originally thought the answer was no, because this is going to be post query planning, however after more thought I believe there are cases where a regex or range can go non-expanded against a field. I will have to create some test cases that mirror these conditions to verify this.

warehouse/query-core/src/test/java/datawave/query/transformer/annotation/AllHitsFactoryTest.java
FineAndDandy: old habbits die hard

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AnnotationHitsTransformer.java
FineAndDandy: this is because things in the annotation table have not been normalized

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
FineAndDandy: I will move this to a small method as well just to keep this method readable

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AnnotationHitsTransformer.java
FineAndDandy: yeah... I convinced myself the code was more readable using all the this. it definitely isn't. Thanks for the nudge.

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AnnotationHitsTransformer.java
FineAndDandy: This was rescoped today. This code will be revamped in an upcoming update

apmoriarty (dismissed):

warehouse/query-core/src/test/java/datawave/query/tables/ShardQueryLogicTest.java
```diff
@@ -589,6 +590,42 @@ public void annotationHitsNotEnabledByQueryParamTest() throws Exception {
...
     }
 
+    @Test(expected = IllegalStateException.class)
+    public void annotationHitsNullTermExtractorTest() throws Exception {
+        withAnnotationHits();
+        logic.getAllHitsQueryConfig().setQueryTermExtractor(null);
+
+        givenAnnotation(buildAnnotation(S1));
+
+        givenQuery("UUID=='CAPONE'");
```
apmoriarty, line 602: small nit pick, queries in unit tests should be well formed when possible. Replace with `UUID == 'CAPONE'`

apmoriarty (dismissed):

warehouse/query-core/src/test/java/datawave/query/transformer/annotation/AllHitsFactoryTest.java
FineAndDandy: I've actually updated this to return a null rather than an uninitialized AllHits

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AnnotationHitsTransformer.java
FineAndDandy: this has been revamped in the new code

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/model/TermHit.java
FineAndDandy: Due to the relabeling of the object depending on the context and it all going into a single object unfortunately these need to remain separate.

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -768,6 +779,43 @@ public boolean isLongRunningQuery() {
         return getConfig().getGroupFields().hasGroupByFields() || !getUniqueFields().isEmpty();
     }
 
+    /**
+     *
+     * @return a configured AnnotationDataAccess object given the table names
+     */
+    private AnnotationDataAccess getAnnotationDataAccess() {
+        AnnotationSerializer<Iterator<Entry<Key,Value>>,Annotation> annotationSerializer = new AccumuloAnnotationSerializer();
+        AnnotationSerializer<Iterator<Entry<Key,Value>>,AnnotationSource> annotationSourceSerializer = new AccumuloAnnotationSourceSerializer();
```
drewfarris: We should use the variant of the `AccumuloAnnotationSerializer` and `AccumuloAnnotationSourceSerialzier` that takes the `visibilityTransformer` and `timestampTransformer` as arguments. The implementation for the transformers should be specified in the config (`AnnotationConfig`?) and ultimately set in `QueryLogicFactory.xml` - although we currently don't have different implementations we plug in for different runtime contexts, I anticipate that we will in the near future.

See also `AnnotationManagerConfig`. Maybe we reuse `AnnotationConfig` for the `AnnotationManagerBean` too? If so, we can remove anything already specified in `AnnotationConfig` from `AnnotationManagerConfig`.

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -768,6 +779,43 @@ public boolean isLongRunningQuery() {
...
+     *
+     * @return
+     * @throws QueryException
+     */
+    private AllHitsFactory getAnnotationHitsFactory() throws QueryException {
+        AllHitsFactory allHitsFactory;
+        try {
+            Class<?> annotationHitsFactoryClass = Class.forName(getAllHitsQueryConfig().getAllHitsFactoryClass());
+            Class<? extends AllHitsFactory> subClass = annotationHitsFactoryClass.asSubclass(AllHitsFactory.class);
+            allHitsFactory = subClass.getDeclaredConstructor().newInstance();
```
drewfarris, line 815: If `AllHitsFactory` is stateless and thread-safe, we might consider creating the instance of the `AllHitsFactory` in the spring xml and just injecting it here instead of instantiating it here from a class name.

warehouse/query-core/src/test/java/datawave/query/transformer/annotation/AllHitsFactoryTest.java
```diff
@@ -0,0 +1,279 @@
+package datawave.query.transformer.annotation;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThrows;
```
drewfarris: I'm not 100% certain this matters, but these should possibly be replace with the jupiter equivalents. 
```suggestion
import static org.junit.jupiter.api.Assertions.assertEquals;
import static org.junit.jupiter.api.Assertions.assertNull;
import static org.junit.jupiter.api.Assertions.assertThrows;
```

warehouse/query-core/src/test/java/datawave/query/transformer/annotation/AnnotationHitsTransformerTest.java
```diff
@@ -0,0 +1,911 @@
...
+        lenient().when(normalizer.normalize("lllLLLL")).thenReturn("lllllll");
+        lenient().when(normalizer.normalize("mmmMMMM")).thenReturn("mmmmmmm");
+        lenient().when(normalizer.normalize("nnnNNNN")).thenReturn("nnnnnnn");
+    }
+
+    private void withParameter(String key, String value) {
+        settings.addParameter(key, value);
+    }
+
+    private Annotation buildAnnotation(String annotationType, String shard, String dataType, String documentId, String sourceHash, Segment... segments) {
```
drewfarris, line 876:
```suggestion
    @SuppressWarnings("SameParameterValue")
    private Annotation buildAnnotation(String annotationType, String shard, String dataType, String documentId, String sourceHash, Segment... segments) {
```

warehouse/query-core/src/test/java/datawave/query/transformer/annotation/AnnotationHitsTransformerTest.java
```diff
@@ -0,0 +1,911 @@
...
+    private void withHits(String id, List<AnnotationHitsTransformer.SegmentHit> expectedHits) throws AllHitsException {
+        when(allHitsFactory.create(any(), any(), any(), any())).thenAnswer(invocation -> {
+            List<AnnotationHitsTransformer.SegmentHit> hits = invocation.getArgument(1);
+            assertEquals(expectedHits.size(), hits.size());
+
+            for (AnnotationHitsTransformer.SegmentHit hit : hits) {
+                boolean found = false;
+                for (AnnotationHitsTransformer.SegmentHit expectedHit : expectedHits) {
+                    if (expectedHit.equals(hit)) {
+                        found = true;
```
drewfarris, line 840:
```suggestion
                        found = true;
                        break;
```

warehouse/query-core/src/test/java/datawave/query/transformer/annotation/AnnotationHitsTransformerTest.java
```diff
@@ -0,0 +1,911 @@
...
+            .setBoundary(SegmentBoundary.newBuilder().setBoundaryType(BoundaryType.TIME_MILLI).setStart(32).setEnd(34).build())
+            .build();
+    private static final Segment S7 = Segment.newBuilder()
+            .addValues(SegmentValue.newBuilder().setValue("mmmMMMM").setScore(1f).build())
+            .addValues(SegmentValue.newBuilder().setValue("nnnNNNN").setScore(.5f).build())
+            .setBoundary(SegmentBoundary.newBuilder().setBoundaryType(BoundaryType.TIME_MILLI).setStart(35).setEnd(50).build())
+            .build();
+    // @formatter:on
+
+    private AnnotationHitsTransformer transformer;
```
drewfarris, line 85: Consider making `AnnotationHitsTransformer` a local variable in the `test` method and removing this?

warehouse/query-core/src/test/java/datawave/query/transformer/annotation/TermExtractorTest.java
```diff
@@ -0,0 +1,135 @@
+package datawave.query.transformer.annotation;
+
+import static org.junit.Assert.assertEquals;
```
drewfarris, line 3:
```suggestion
import static org.junit.jupiter.api.Assertions.assertEquals;
```

warehouse/query-core/src/test/java/datawave/query/transformer/annotation/TermExtractorTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+    @BeforeEach
+    public void setup() {
+        termExtractor = new TermExtractor(null);
+        normalizer = new LcNoDiacriticsNormalizer();
+    }
+
+    // test term extraction of eq nodes
+    // @formatter:off
+    @CsvSource({
```
drewfarris, line 32: @CsvSource is cool!

warehouse/query-core/src/test/java/datawave/query/transformer/annotation/TermExtractorTest.java
drewfarris, line 1: I read this test wondering if we need to support term extraction from content functions - but I think I convinced myself that we don't need to do this because I think most content functions also include a `FIELD == VALUE` part too IIRC. Worth thinking about further?

warehouse/query-core/src/test/java/datawave/query/transformer/annotation/SegmentValueByScoreComparatorTest.java
```diff
@@ -0,0 +1,42 @@
+package datawave.query.transformer.annotation;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
```
drewfarris:
```suggestion
import static org.junit.jupiter.api.Assertions.assertEquals;
import static org.junit.jupiter.api.Assertions.assertTrue;
```

warehouse/query-core/src/test/java/datawave/query/transformer/annotation/TermExtractorTest.java
```diff
@@ -0,0 +1,135 @@
...
+        String[] splits;
+        if (terms != null) {
+            splits = terms.split(";");
+        } else {
+            splits = new String[0];
+        }
+        Set<String> expectedSet = new HashSet<>();
+        for (String term : splits) {
+            expectedSet.add(term);
+        }
```
drewfarris:
```suggestion
        Set<String> expectedSet = new HashSet<>(Arrays.asList(splits));
```

warehouse/query-core/src/test/java/datawave/query/transformer/annotation/BoundaryComparatorTest.java
```diff
@@ -0,0 +1,65 @@
+package datawave.query.transformer.annotation;
+
+import static org.junit.Assert.assertTrue;
```
drewfarris:
```suggestion
import static org.junit.jupiter.api.Assertions.assertTrue;
```

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AllHitsFactory.java
```diff
@@ -0,0 +1,184 @@
...
+import java.util.TreeMap;
+import java.util.concurrent.TimeUnit;
+
+import datawave.annotation.protobuf.v1.SegmentBoundary;
+import datawave.annotation.protobuf.v1.SegmentValue;
+import datawave.query.transformer.annotation.model.AllHit;
+import datawave.query.transformer.annotation.model.AllHits;
+import datawave.query.transformer.annotation.model.Term;
+import datawave.query.transformer.annotation.model.TermHit;
+
```
drewfarris, line 16: Consider adding some brief class-level javadoc?

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AnnotationHitsTransformer.java
```diff
@@ -0,0 +1,498 @@
...
+     *
+     * @param sortedSegments
+     *            the sorted set of all SegmentBoundary to consider for the search. All SegmentValue will appear in ascending order
+     * @param contextSize
+     *            the window of adjacent terms +/- from the hit term to include in the hit
+     * @param minScore
+     *            the minimum score a term must have to register a hit
+     * @return non-null List of hits ordered by the segmentBoundary they hit on. Hit order guaranteed to be ascending SegmentBoundary, no second order sort is
+     *         applied. Hits for the same SegmentBoundary will appear in the order they were found.
+     */
```
drewfarris, line 349: Love it!

drewfarris (approved):
Just minor nits that can be cleaned up later.

drewfarris (changes requested):
Only one change required: making the timestamp and visibility transformers for the annotation serializers pluggable.

warehouse/query-core/src/test/java/datawave/query/transformer/annotation/TermExtractorTest.java
FineAndDandy: when converting phrases from lucene to jexl the functions should always include the individual term. If the query is written in jexl, the user could technical just write the function... but things like indexing would never work correctly this way in the rest of query. We use the EQNodes to anchor the indexed terms.

drewfarris (approved):

warehouse/query-core/src/test/java/datawave/query/transformer/annotation/TermExtractorTest.java
```diff
@@ -0,0 +1,133 @@
...
+            "UUID=~'a' || !(UUID=~'b' && UUID=~'c'),a",
+            "!(UUID=~'a' || UUID=~'b') && (UUID=~'c'),c",
+            "!(!((UUID=~'a'))),a",
+            "!(!(!(UUID=~'a'))),",
+            "UUID!='abc',",
+            "!UUID=='abc',",
+            "!(!(UUID=='abc')),abc",
+            "!(!(UUID=='abc')||(UUID=='def')),abc",
+            "!(UUID!='abc'),abc"
+    })
```
apmoriarty, line 77: Might be worth a check inside your test method that pushes down negations and asserts no changes to the query. Not required for merge but definitely worth consideration.

web-services/annotations/src/test/resources/datawave/webservice/annotation/AnnotationManagerFactory.xml
```diff
@@ -10,9 +10,19 @@
          because it will get discovered by CDI -->
     <bean id="AnnotationManager" scope="prototype"  class="datawave.webservice.annotation.AnnotationManagerBean"/>
 
-    <bean id="AnnotationManagerConfig" scope="prototype"  class="datawave.webservice.annotation.AnnotationManagerConfig">
+    <bean id="AnnotationConfig" scope="prototype" class="datawave.query.config.annotation.AnnotationConfig">
         <property name="annotationTableName" value="annotation"/>
         <property name="annotationSourceTableName" value="annotationSource"/>
+        <property name="visibilityTransformer">
```
apmoriarty, line 16: Thank you for providing examples in test configs

apmoriarty (approved):

PR #3331, opened 2026-01-05
Remove non-public Accumulo API usage from ScannerBaseDelegate

### LINKED ISSUE #3334: Remove non-public API usage from ScannerBaseDelegate
## Summary
Remove dependencies on non-public Accumulo APIs in ScannerBaseDelegate.

## Non-Public Classes
- `org.apache.accumulo.core.clientImpl.ScannerOptions`
- `org.apache.accumulo.core.dataImpl.thrift.IterInfo`

## Solution
Track iterator names locally instead of introspecting the delegate via non-public APIs:
- Add `systemIteratorNames` and `userIteratorNames` Sets
- Update add/remove methods to maintain the local tracking sets
- Simplify `clearScanIterators()` to iterate over tracked names
- Remove `ScannerOptionsHelper` inner class

## Files Affected
- `core/utils/accumulo-utils/src/main/java/datawave/webservice/common/connection/ScannerBaseDelegate.java`

Part of #2443

SethSmucker:
## Summary
- Remove dependency on non-public Accumulo APIs: `ScannerOptions`, `IterInfo`
- Track iterator names locally instead of introspecting the delegate
- Remove `ScannerOptionsHelper` inner class
- Simplify `clearScanIterators()` implementation

Fixes #3334
Part of #2443


core/utils/accumulo-utils/src/main/java/datawave/webservice/common/connection/ScannerBaseDelegate.java
```diff
@@ -140,23 +148,20 @@ public void clearColumns() {
 
     @Override
     public void clearScanIterators() {
-        if (delegate instanceof ScannerOptions) {
```
lbschanno, line 143: I'm not 100% sure it's safe to remove this check until we know we're not using or creating instances of ScannerOptions anywhere (and know that accumulo isn't creating any instances of them anywhere). There was some initial work done to replace our usage of ScannerOptions with `datawave.query.tables.SessionOptions`, so that could be an avenue for replacement here instead.

apmoriarty: I think I would like to see something added back like this
```java
if ( !(delegate instanceof ScannerOptions )) {
 throw new UnsupportedOperationException("Cannot clear scan iterators on a non-ScannerOptions class! (" + delegate.getClass() + ")");
}
```

apmoriarty (changes requested):

avgAGB (approved):

apmoriarty (approved):

PR #3340, opened 2026-01-05
Replace ClientContext.getTableId with public API

### LINKED ISSUE #3339: Replace ClientContext.getTableId with public API
## Summary
Replace usage of non-public ClientContext.getTableId() with public AccumuloClient API.

## Non-Public Class
- `org.apache.accumulo.core.clientImpl.ClientContext`

## Problem
Code uses `ClientContext.getTableId(tableName)` which requires casting AccumuloClient to the non-public ClientContext class.

## Solution
Replace with public API:
```java
// Before
ClientContext ctx = AccumuloConnectionFactory.getClientContext(client);
tableId = ctx.getTableId(tableName);

// After
tableId = TableId.of(client.tableOperations().tableIdMap().get(tableName));
```

Also remove the now-unused `AccumuloConnectionFactory.getClientContext()` utility method.

## Files Affected
- `core/connection-pool/src/main/java/datawave/core/common/connection/AccumuloConnectionFactory.java`
- `warehouse/query-core/src/main/java/datawave/query/scheduler/PushdownScheduler.java`

Part of #2443

SethSmucker:
## Summary
- Remove `getClientContext` utility from `AccumuloConnectionFactory`
- Update `PushdownScheduler` to use `tableOperations().tableIdMap()`

Fixes #3339
Part of #2443

### DIFF
--- a/core/connection-pool/src/main/java/datawave/core/common/connection/AccumuloConnectionFactory.java
+++ b/core/connection-pool/src/main/java/datawave/core/common/connection/AccumuloConnectionFactory.java
@@ -5,10 +5,8 @@ import java.util.List;
 import java.util.Map;
 
 import org.apache.accumulo.core.client.AccumuloClient;
-import org.apache.accumulo.core.clientImpl.ClientContext;
 
 import datawave.core.common.result.ConnectionPool;
-import datawave.webservice.common.connection.WrappedAccumuloClient;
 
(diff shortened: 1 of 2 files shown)


foster33:
It seems that `ClientContext.getTableId` is also used within `BulkInputFormat`

PR #3346, opened 2026-01-06
Migrate MultiNormalizerTest to use AbstractQueryTest

apmoriarty:


jalphonso (approved):

warehouse/query-core/src/test/java/datawave/query/util/AbstractQueryTest.java
```diff
@@ -67,7 +63,7 @@ public enum RangeType {
 
     protected final HitTermAssertions hitTermAssertions = new HitTermAssertions();
 
-    protected AccumuloClient clientForTest;
+    private AccumuloClient clientForTest;
```
ivakegg, line 66: Are we sure nothing on the other side extends this and uses?

ivakegg (approved):

warehouse/query-core/src/test/java/datawave/query/util/AbstractQueryTest.java
apmoriarty: Yes. The static setup methods require a static client, and then in the non-static 'before' methods the static client is assigned to this private instance.

PR #3361, opened 2026-01-08
[DO NOT MERGE] Make webservice use query starter query logic factory xml files and the config service for properties

billoley:
Make webservice use query starter query logic factory xml files and the config service for properties

The reason for some changes may not be immediately obvious.  Here are some hints:

1) Changes in SpringCDIExtension to create a three layer applicationContext from configuration while remaining backward compatible
2) Additional spring Configuration classes to inject beans that are expected in XML configuration.  Primary reasons are when collections (lists, maps, sets, class structures) are created using yaml but these can't just be substitututed using a PropertiesPlaceholderConfigurer.
3) Changes to XML or yaml to being the configuration (old vs new) to parity
4) Changes to unit and integration tests to make them use the new configuration
5) Tests to compare the compile-time property substitution to the runtime property substitution method of creating a spring application context


web-services/deploy/configuration/src/main/resources/datawave-compile-time-properties/modification/ExampleModificationServices.xml
```diff
@@ -0,0 +1,54 @@
...
+    <!-- Database properties are retrieved from the database.properties file but can be overridden by system properties
+    <bean id="propertyPlaceholderConfigurer" class="org.springframework.beans.factory.config.PropertyPlaceholderConfigurer">
+        <property name="location">
+            <value>classpath:/datawave/query/database.properties</value>
+        </property>
+        <property name="systemPropertiesModeName">
+            <value>SYSTEM_PROPERTIES_MODE_OVERRIDE</value>
+        </property>
+    </bean>
+    -->
```
apmoriarty: Should this be removed?

apmoriarty (dismissed):

web-services/deploy/configuration/src/main/resources/datawave-compile-time-properties/modification/ExampleModificationServices.xml
billoley: removed

PR #3362, opened 2026-01-08
consolidate abstract query test

apmoriarty:
- consolidate naming convention and method flow for AbstractQueryTest
 - native support for running test cases against each type of index (uid, no uid, and bitset)
 - upgraded from JUnit4 to JUnit5
 - migrated from Arquillian to Spring

### DIFF
--- a/warehouse/query-core/src/test/java/datawave/query/util/AbstractQueryTest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/util/AbstractQueryTest.java
@@ -1,8 +1,9 @@
 package datawave.query.util;
 
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 import java.text.DateFormat;
@@ -13,6 +14,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeSet;
 import java.util.UUID;
 
 import org.apache.accumulo.core.client.AccumuloClient;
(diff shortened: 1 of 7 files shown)


alerman (dismissed):

warehouse/query-core/src/test/java/datawave/query/util/AbstractQueryTest.java
```diff
@@ -186,25 +261,70 @@ public void withOptionalAllOf(String... hitTerms) {
...
+            switch (indexTableName) {
+                case TestIndexTableNames.SHARD_INDEX:
+                case TestIndexTableNames.NO_UID_INDEX:
+                    break;
+                case TestIndexTableNames.TRUNCATED_INDEX:
+                    getLogic().setUseTruncatedIndex(false);
+                    break;
+                default:
+                    throw new IllegalStateException("Unknown index table name: " + indexTableName);
+            }
```
FineAndDandy: It looks like the point of this is to undo the sets that were made above. Move this to a finally block in case the planAndExecuteIndividualQuery throws an exception

warehouse/query-core/src/test/java/datawave/query/util/AbstractQueryTest.java
```diff
@@ -272,22 +384,68 @@ public void assertPlannedQuery(String query) {
...
 
-    protected String getQuery() {
-        Preconditions.checkNotNull(query, "query is null");
-        return query;
+    /**
+     * An optional assertion for the UUID field
+     */
+    public void assertUuids() {
+        if (expectedUUIDs.isEmpty()) {
+            return;
```
FineAndDandy, line 421: assertTrue(results.isEmpty())

apmoriarty: uuid assertion is optional, so that won't work. There are cases where just the count of results is asserted without ever expecting a specific UUID.

warehouse/query-core/src/test/java/datawave/query/util/AbstractQueryTest.java
apmoriarty: Updated to use try-finally in https://github.com/NationalSecurityAgency/datawave/pull/3362/changes/af0d540bfd093ce926d5e901ce99835d1e1460c5

ivakegg (approved):

dtspence (approved):

PR #3363, opened 2026-01-12
Index expansion lookups will now use the bitset index if configured

apmoriarty:


alerman (approved):

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BaseRegexIndexLookup.java
```diff
@@ -41,7 +41,11 @@ public BaseRegexIndexLookup(ShardQueryConfiguration config, ScannerFactory scann
     }
 
     protected String getTableName() {
-        return reverse ? config.getReverseIndexTableName() : config.getIndexTableName();
+        if (reverse) {
+            // there is only one supported reverse index
+            return config.getReverseIndexTableName();
```
jschmidt10, line 46: Are we planning to eventually support bitsets on the reverse index? It feels like the ideal would be both shard index and shard rev index would respect the truncated index flag together, so we don't have mixed schemas.

apmoriarty: Applying the bitset iterators/combiners to the shard reverse index is planned for the near future. Agree that mixed schemas isn't great.

jschmidt10 (approved):

PR #3371, opened 2026-01-13
Replace non-public DateFormatSupplier with SimpleDateFormat

### LINKED ISSUE #3358: Replace non-public DateFormatSupplier
## Summary
Replace non-public `DateFormatSupplier` with standard SimpleDateFormat.

## Non-Public Class
- `org.apache.accumulo.core.util.format.DateFormatSupplier`

## Solution
Replace `DateFormatSupplier.createDefaultFormatSupplier()` with `ThreadLocal.withInitial(() -> new SimpleDateFormat("yyyy/MM/dd HH:mm:ss.SSS"))`.

## Files Affected
- warehouse/core/src/main/java/datawave/mr/bulk/BulkInputFormat.java

Part of #2443

SethSmucker:
## Summary
- Replace non-public `org.apache.accumulo.core.util.format.DateFormatSupplier` with a ThreadLocal initialized using `SimpleDateFormat` with the same date format pattern

Fixes #3358
Part of #2443


warehouse/core/src/main/java/datawave/mr/bulk/BulkInputFormat.java
```diff
@@ -96,7 +96,7 @@ public class BulkInputFormat extends InputFormat<Key,Value> {
     protected static final Logger log = LoggerFactory.getLogger(BulkInputFormat.class);
 
     private static final ThreadLocal<Date> tmpDate = ThreadLocal.withInitial(Date::new);
-    private static final ThreadLocal<DateFormat> formatter = DateFormatSupplier.createDefaultFormatSupplier();
+    private static final ThreadLocal<DateFormat> formatter = ThreadLocal.withInitial(() -> new SimpleDateFormat("yyyy/MM/dd HH:mm:ss.SSS"));
```
jschmidt10: This is a reasonable drop-in. The best fix here would be to use a java8+ DateTimeFormatter as its threadsafe/immutable (and I believe even potentially faster).

jschmidt10: Hmm... it looks like the old formatter is formatting the date as epoch millis so I'm not sure the expression you put in is correct.

warehouse/core/src/main/java/datawave/mr/bulk/BulkInputFormat.java
```diff
@@ -1316,8 +1315,7 @@ public boolean nextKeyValue() throws IOException, InterruptedException {
                         sb.append(new ColumnVisibility(currentK.getColumnVisibility(buffer)));
 
                         // append timestamp
-                        tmpDate.get().setTime(entry.getKey().getTimestamp());
-                        sb.append(" ").append(formatter.get().format(tmpDate.get()));
+                        sb.append(" ").append(formatter.format(Instant.ofEpochMilli(entry.getKey().getTimestamp())));
```
lbschanno: It appears that `DateFormatSupplier.createDefaultFormatSupplier()` returns a DateFormat that simply appends the result of `Date.getTime()` to the resulting string. I confirmed this with the following test:

```java
    @Test
    void testFormatter() {
        ThreadLocal<DateFormat> formatter = DateFormatSupplier.createDefaultFormatSupplier();
        Date date = new Date();
        date.setTime(44);
        Assertions.assertEquals("44", formatter.get().format(date));
    }
    
```
So we should be able to delete the formatter and simply do
```java
sb.append(" ").append(entry.getKey().getTimestamp());
```

lbschanno (changes requested):

jschmidt10 (dismissed):

foster33 (approved):

apmoriarty (approved):

lbschanno (approved):

PR #3384, opened 2026-01-16
Ingest supports writing bitset index keys

apmoriarty:
- ingest supports writing bitset global index keys
- add ability to suppress standard global index keys

### DIFF
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
@@ -159,8 +159,8 @@ public abstract class ShardedDataTypeHandler<KEYIN> extends StatsDEnabledDataTyp
 
     public static final String SHARD_INDEX_ENABLED = "shard.index.enabled";
     public static final String BITSET_INDEX_ENABLED = "bitset.index.enabled";
-    public static final String DAY_INDEX_ENABLED = "year.index.enabled";
-    public static final String YEAR_INDEX_ENABLED = "day.index.enabled";
+    public static final String DAY_INDEX_ENABLED = "day.index.enabled";
+    public static final String YEAR_INDEX_ENABLED = "year.index.enabled";
 
     /**
      * name of ACCUMULO table to store DATAWAVE metadata
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
@@ -144,6 +144,7 @@ public abstract class ShardedDataTypeHandler<KEYIN> extends StatsDEnabledDataTyp
     public static final String SHARD_ININDEX_BLOOM_OPTIMUM_MAX_FILTER_SIZE = "shard.table.index.bloom.optimum.max.filter.size"; // Bytes
     public static final String SHARD_STATS_TNAME = "shard.stats.table.name";
     public static final String SHARD_GIDX_TNAME = "shard.global.index.table.name";
+    public static final String SHARD_BITSET_INDEX_TABLE_NAME = "shard.bitset.index.table.name";
     public static final String SHARD_DAY_INDEX_TABLE_NAME = "shard.global.day.index.table.name";
     public static final String SHARD_YEAR_INDEX_TABLE_NAME = "shard.global.year.index.table.name";
     public static final String SHARD_GRIDX_TNAME = "shard.global.rindex.table.name";
@@ -156,6 +157,8 @@ public abstract class ShardedDataTypeHandler<KEYIN> extends StatsDEnabledDataTyp
     public static final String IS_REINDEX_ENABLED = "ingest.reindex.enabled";
     public static final String FIELDS_TO_REINDEX = "ingest.reindex.fields";
 
+    public static final String SHARD_INDEX_ENABLED = "shard.index.enabled";
+    public static final String BITSET_INDEX_ENABLED = "bitset.index.enabled";
     public static final String DAY_INDEX_ENABLED = "year.index.enabled";
     public static final String YEAR_INDEX_ENABLED = "day.index.enabled";
 
@@ -205,6 +208,7 @@ public abstract class ShardedDataTypeHandler<KEYIN> extends StatsDEnabledDataTyp
     private float bloomFilteringTimeoutThreshold;
     private Text shardTableName = null;
     private Text shardIndexTableName = null;
+    private Text shardBitsetIndexTableName = null;
     private Text shardDayIndexTableName = null;
     private Text shardYearIndexTableName = null;
     private Text indexStatsTableName = null;
@@ -218,8 +222,12 @@ public abstract class ShardedDataTypeHandler<KEYIN> extends StatsDEnabledDataTyp
     protected MarkingFunctions markingFunctions;
     protected IngestConfiguration ingestConfig = IngestConfigurationFactory.getIngestConfiguration();
 
+    // default setting is a standard index with uid and event keys
+    private boolean shardIndexEnabled = true;
     private boolean shardIndexCreateUids = true;
     private boolean suppressEventKeys = false;
+    // alternate global index implementations
+    private boolean bitsetIndexEnabled = false;
     private boolean dayIndexEnabled = false;
     private boolean yearIndexEnabled = false;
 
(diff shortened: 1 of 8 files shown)


warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
```diff
@@ -371,6 +394,11 @@ public String[] getTableNames(Configuration conf) {
         if (null != tableName)
             tableNames.add(tableName);
 
+        tableName = conf.get(SHARD_BITSET_INDEX_TABLE_NAME, null);
+        if (null != tableName) {
+            tableNames.add(tableName);
+        }
+
```
FineAndDandy, line 401: only add this if enabled

FineAndDandy: I see that isn't the previous pattern of the method, so feel free to take it or leave it

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
```diff
@@ -898,6 +995,23 @@ public void writeShardDayIndexKey(Multimap<BulkIngestKey,Value> values, RawRecor
...
+     *            the full shard
+     * @return a Value containing a BitSet
+     */
+    public Value getValueForBitsetIndex(String shard) {
+        int underscoreIndex = shard.indexOf('_');
+        Preconditions.checkArgument(underscoreIndex > 0, "shard did not contain an underscore: " + shard);
+        String shardId = shard.substring(underscoreIndex + 1);
+        int offset = Integer.parseInt(shardId);
+        BitSet bitset = new BitSet();
+        bitset.set(offset);
```
FineAndDandy, line 1011: This will cause the underlying BitSet to be a dynamic size. This may or may not be a problem, just noting it here in case its relevant where the BitSet come together for different Values

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandlerTest.java
```diff
@@ -491,6 +529,11 @@ public void testSimpleReverseIndex() {
         assertTrue(values.isEmpty());
     }
 
+    @Test
+    public void testSimpleReverseIndexWithBitsetIndex() {
+
```
FineAndDandy, line 534: TODO

warehouse/ingest-core/src/test/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandlerTest.java
```diff
@@ -511,6 +554,11 @@ public void testMaskedSimpleReverseIndex() {
 
     }
 
+    @Test
+    public void testMaskedSimpleReverseIndexWithBitsetIndex() {
+
```
FineAndDandy, line 559: TODO

apmoriarty: Support for reverse index is being done in a separate unit of work

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
apmoriarty: This is by design. It allows for a dynamic number of shards per day.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
apmoriarty: I would like to do that later, unsure the consequences if that change were made in this unit of work

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
```diff
@@ -898,6 +995,23 @@ public void writeShardDayIndexKey(Multimap<BulkIngestKey,Value> values, RawRecor
...
+     *
+     * @param shard
+     *            the full shard
+     * @return a Value containing a BitSet
+     */
+    public Value getValueForBitsetIndex(String shard) {
+        int underscoreIndex = shard.indexOf('_');
+        Preconditions.checkArgument(underscoreIndex > 0, "shard did not contain an underscore: " + shard);
+        String shardId = shard.substring(underscoreIndex + 1);
+        int offset = Integer.parseInt(shardId);
```
ravisastryk, line 1009: Add defensive validation has this can potentially throw `NumberFormatException`

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
```diff
@@ -898,6 +995,23 @@ public void writeShardDayIndexKey(Multimap<BulkIngestKey,Value> values, RawRecor
...
+     * Constructs a {@link Value} that contains a {@link BitSet}'s backing byte array. The bitset index is set to the shard offset.
+     *
+     * @param shard
+     *            the full shard
+     * @return a Value containing a BitSet
+     */
+    public Value getValueForBitsetIndex(String shard) {
+        int underscoreIndex = shard.indexOf('_');
+        Preconditions.checkArgument(underscoreIndex > 0, "shard did not contain an underscore: " + shard);
+        String shardId = shard.substring(underscoreIndex + 1);
```
ravisastryk, line 1008: Consider adding defensive validation
```suggestion
        String shardOffsetStr = shard.substring(underscoreIndex + 1);
    int offset;
    try {
        offset = Integer.parseInt(shardOffsetStr);
    } catch (NumberFormatException e) {
        throw new IllegalArgumentException(
            "Shard offset is not a valid integer: " + shardOffsetStr, e);
    }
    
    Preconditions.checkArgument(offset >= 0 && offset < MAX_SHARD_OFFSET,
        "Shard offset out of valid range [0, %s): %s", MAX_SHARD_OFFSET, offset)
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
```diff
@@ -156,6 +157,8 @@ public abstract class ShardedDataTypeHandler<KEYIN> extends StatsDEnabledDataTyp
     public static final String IS_REINDEX_ENABLED = "ingest.reindex.enabled";
     public static final String FIELDS_TO_REINDEX = "ingest.reindex.fields";
 
+    public static final String SHARD_INDEX_ENABLED = "shard.index.enabled";
+    public static final String BITSET_INDEX_ENABLED = "bitset.index.enabled";
     public static final String DAY_INDEX_ENABLED = "year.index.enabled";
     public static final String YEAR_INDEX_ENABLED = "day.index.enabled";
```
ravisastryk: Something weird I have noted here by just skimming through. Although unrelated, let's make this minor fix.

These are swapped, the constant names do not match their string values. This will cause configuration confusion and potential runtime issues. Let's fix this proactively here.
```suggestion
    public static final String DAY_INDEX_ENABLED = "day.index.enabled";
    public static final String YEAR_INDEX_ENABLED = "year.index.enabled";

```

ravisastryk (commented):
Observations but not a blocker. You do not necessarily have to implement these. But if someone can get these addressed in a separate PR, I will be happy to review.

1. Something unrelated. But I found the constants swapped. This will cause configuration confusion and potential runtime issues. 
2. There could be potential architectural improvements such as:
applying the Strategy Pattern to make index implementations pluggable. These need not be implemented here. I am open to make those changes if architects feel a refactor along these lines using the Open/Closed Principle where new index types don't require modifying the handler
```
public interface IndexWriteStrategy {
    boolean isEnabled();
    void writeIndexKey(IndexWriteContext ctx);
}

// In setup():
List<IndexWriteStrategy> activeStrategies = new ArrayList<>();
if (conf.getBoolean(SHARD_INDEX_ENABLED, true)) {
    activeStrategies.add(new StandardIndexStrategy(shardIndexTableName));
}
if (conf.getBoolean(BITSET_INDEX_ENABLED, false)) {
    activeStrategies.add(new BitsetIndexStrategy(bitsetIndexTableName));
}

// In createTermIndexColumn():
for (IndexWriteStrategy strategy : activeStrategies) {
    strategy.writeIndexKey(context);
}
```
3. Consolidate Duplicate BitSet Logic. The current problem is that the bitset index writing logic is duplicated in two places:
```
ShardedDataTypeHandler.writeBitsetIndexKey()
ExtendedContentIndexingColumnBasedHandler.createIndexColumn()
```

Both have similar patterns:
```
String shard = new String(shardId);
String cq = shard.substring(0, 8) + '\u0000' + event.getDataType().outputName();
Key key = new Key(...);
Value value = getValueForBitsetIndex(shard);
```
I would recommend updating to keep in a shared utility class:
```
public class BitsetIndexKeyBuilder {
    
    public static BulkIngestKey buildKey(
            Text tableName, 
            String fieldValue, 
            String fieldName, 
            byte[] shardId, 
            String dataTypeName,
            byte[] visibility,
            long timestamp) {
        
        String shard = new String(shardId);
        String cq = shard.substring(0, 8) + '\u0000' + dataTypeName;
        Key key = new Key(fieldValue, fieldName, cq, new String(visibility), timestamp);
        return new BulkIngestKey(tableName, key);
    }
    
    public static Value buildValue(String fullShard) {
        int offset = extractShardOffset(fullShard);
        BitSet bitset = new BitSet();
        bitset.set(offset);
        return new Value(bitset.toByteArray());
    }
    
    private static int extractShardOffset(String shard) {
        int idx = shard.indexOf('_');
        Preconditions.checkArgument(idx > 0, "Invalid shard format: " + shard);
        return Integer.parseInt(shard.substring(idx + 1));
    }
}
```

ravisastryk (commented):
Thank you for the fixes. It looks good to me

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/tokenize/ExtendedContentIndexingColumnBasedHandler.java
```diff
@@ -860,6 +860,15 @@ protected void createIndexColumn(RawRecordContainer event, ContextWriter<KEYOUT,
         BulkIngestKey bKey = new BulkIngestKey(tableName, k);
         contextWriter.write(bKey, val, context);
 
+        if (getBitSetIndexEnabled()) {
+            String shard = new String(shardId);
```
ivakegg, line 864: Interesting, do we not use the maskedFieldHelper in this class?

ivakegg: Can we reuse the code in the ShardedDataTypeHandler ?

apmoriarty: We do not use the masked field helper in that class, and I cannot easily reuse the code from the ShardedDataTypeHandler. The ShardedDataTypeHandler is adding keys and values to a multimap while the ExtendedContentIndexingColumnBasedHandler is writing keys and values directly to the context. 

There are a few helper methods in the ShardedDataTypeHandler for ingesting bitset index keys, and those are being reused in the ExtendedContentIndexingColumnBasedHandler.

ivakegg (approved):

FineAndDandy (approved):

PR #3394, opened 2026-01-23
Replace non-public IterationInterruptedException in next/ package

### LINKED ISSUE #3393: Replace non-public IterationInterruptedException in next/ package
## Summary
Replace non-public `org.apache.accumulo.core.iteratorsImpl.system.IterationInterruptedException` with class name checks in the `datawave.next` package.

## Non-Public Class
- `org.apache.accumulo.core.iteratorsImpl.system.IterationInterruptedException`

## Solution
Use class name check pattern (same as TabletClosedException in PR #3318):
- Add `isIterationInterruptedException(Throwable t)` helper method
- Replace `instanceof` checks with helper method calls
- Change catch blocks to catch `RuntimeException` and check class name

## Files Affected
- warehouse/query-core/src/main/java/datawave/next/DocIdQueryIterator.java
- warehouse/query-core/src/main/java/datawave/next/retrieval/DocumentIterator.java
- warehouse/query-core/src/main/java/datawave/next/scanner/DocumentIdProducer.java
- warehouse/query-core/src/main/java/datawave/next/scanner/DocumentRangeScan.java

Part of #2443

SethSmucker:
## Summary
- Replace non-public `IterationInterruptedException` with class name checks in the `datawave.next` package
- Add `isIterationInterruptedException(Throwable t)` helper method to each affected file
- Use same pattern as PR #3318 for `TabletClosedException`

Fixes #3393
Part of #2443


warehouse/query-core/src/main/java/datawave/next/DocIdQueryIterator.java
```diff
@@ -271,6 +270,13 @@ public SortedKeyValueIterator<Key,Value> deepCopy(IteratorEnvironment env) {
         return new DocIdQueryIterator(this, env);
     }
 
+    /**
+     * Check if the throwable is an IterationInterruptedException by class name to avoid importing non-public Accumulo API.
+     */
+    private static boolean isIterationInterruptedException(Throwable t) {
+        return t != null && t.getClass().getName().equals("org.apache.accumulo.core.iteratorsImpl.system.IterationInterruptedException");
```
keith-turner: This is very brittle. Its still using the class, but there will be no failure if the class goes away.  The following would fail if the class no longer exists.

```java
var c = getClass().getClassLoader().loadClass("org.apache.accumulo.core.iteratorsImpl.system.IterationInterruptedException");
return t != null && t.getClass().equals(c);
```

PR #3400, opened 2026-01-26
Initial implementation for return child fields

ivakegg:


warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1272,13 +1272,25 @@ protected DocumentProjection getProjection() {
...
             return new DocumentProjection(this.isIncludeGroupingContext(), this.isReducedResponse(), isTrackSizes(),
                             new Projection(this.allowListedFields, Projection.ProjectionType.INCLUDES));
         } else if (this.useDisallowListedFields) {
             // make sure we are not excluding any fields being matched in the limit fields mechanism
             if (!this.matchingFieldSets.isEmpty()) {
                 this.disallowListedFields.removeAll(getMatchingFieldList());
             }
+
+            // make sure we are not excluding any of the child fields
+            if (!this.childFields.isEmpty()) {
```
jschmidt10, line 1291: Just going off the pattern so far where we are treating childFields similarly to matchingFieldList, do we need to add this same code block to `getCompositeProjection()` where we prune out the childFields?

jschmidt10: Also potentially in `getLimitFields()` as well?

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -2756,6 +2756,20 @@ protected void setCommonIteratorOptions(ShardQueryConfiguration config, Iterator
             addOption(cfg, QueryOptions.DISALLOWLISTED_FIELDS, config.getDisallowlistedFieldsAsString(), false);
         }
 
+        // include the child projected fields
+        if (null != config.getProjectChildFields() && !config.getProjectChildFields().isEmpty()) {
+            if (log.isDebugEnabled()) {
+                final int maxLen = 100;
```
jschmidt10: Not a big deal but this is probably better as a static/final constant at the top of the class.

warehouse/query-core/src/main/java/datawave/query/predicate/TLDEventDataFilter.java
```diff
@@ -615,6 +623,25 @@ private void setSortedLists(Set<String> allowlist, Set<String> disallowlist) {
             Collections.sort(sortedDisallowlist);
             sortedDisallowlist = Collections.unmodifiableList(sortedDisallowlist);
         }
+
+        if (childFields != null && !childFields.isEmpty()) {
+            childList = new ArrayList<>(childFields);
```
jschmidt10: A readability win here could be to make a method like:

```
  childList = sortedImmutableList(childFields);
```

That could be used in all the other code blocks above and then if you add varargs, we could use it for the convenience list too:

```
  childSeekList = sortedImmutableList(queryFields, childList);
```

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
```diff
@@ -514,20 +514,17 @@ protected void reduceTypeMetadata(ASTJexlScript script, IteratorSetting newItera
             // sum query fields, projection fields, and composite fields
             fieldsToRetain.addAll(ReduceFields.getQueryFields(script));
 
-            if (options.containsKey(QueryOptions.PROJECTION_FIELDS)) {
-                String option = options.get(QueryOptions.PROJECTION_FIELDS);
-                if (org.apache.commons.lang3.StringUtils.isNotBlank(option)) {
-                    fieldsToRetain.addAll(Splitter.on(',').splitToList(option));
-                }
+            String option = options.get(QueryOptions.PROJECTION_FIELDS);
+            if (org.apache.commons.lang3.StringUtils.isNotBlank(option)) {
```
jschmidt10: I realize you are following the existing pattern but it looks like we can just import `StringUtils` to get rid of the full package path. I didn't see any usages of a non-lang3 StringUtils.

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
```diff
@@ -540,6 +537,14 @@ protected void reduceTypeMetadata(ASTJexlScript script, IteratorSetting newItera
             log.trace("Could not reduce type metadata per shard");
         }
 
+        // ensure children fields are included
+        if (options.containsKey(QueryOptions.PROJECTION_CHILD_FIELDS)) {
```
jschmidt10: The pattern you did above was to pull the option and then check for `isBlank()`. Can we do that here too? I dig eliminating the double Map lookups when possible, too.

warehouse/query-core/src/main/java/datawave/query/util/ShardQueryUtils.java
```diff
@@ -116,6 +121,14 @@ public static ASTJexlScript applyQueryModel(ASTJexlScript script, ShardQueryConf
             config.setProjectFields(Sets.newHashSet(projectFields));
         }
 
+        if (projectChildFields != null && !projectChildFields.isEmpty()) {
+            projectChildFields = queryModel.remapParameter(projectChildFields, inverseReverseModel);
+            if (log != null && log.isTraceEnabled()) {
```
jschmidt10: Its not possible for `log` to be null. Looks like a copy of the pattern above but the calls later in the class eliminate it.

warehouse/query-core/src/test/java/datawave/query/predicate/TLDEventDataFilterTest.java
```diff
@@ -673,4 +675,44 @@ public void testGetSeekRangeForChildWithMultipleValuesPerField() throws ParseExc
...
+        Map<String,ExpressionFilter> expressionFilters = getExpressionFilters(script, new AttributeFactory(mockTypeMetadata));
+        filter = new TLDEventDataFilter(script, Set.of("FOO", "FOO2"), expressionFilters, null, null, childFields, 1, 1);
+
+        Key k1 = new Key("row", "datatype\u0000d8zay2.-3pnndm.-anolok", "FOO\0bar");
+        Key k2 = new Key("row", "datatype\u0000d8zay2.-3pnndm.-anolok.1", "FOO2\0value1");
+        Key k3 = new Key("row", "datatype\u0000d8zay2.-3pnndm.-anolok.2", "FOO2\0value2");
+        Key k4 = new Key("row", "datatype\u0000d8zay2.-3pnndm.-anolok.3", "FOO2\0value3");
+
+        // TLD field is applied, kept, no seek range generated
+        assertTrue(filter.apply(new AbstractMap.SimpleEntry<>(k1, null)));
```
jschmidt10: A little prettier, in Java 11 you can do `Map.entry(k1, null)`.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
ivakegg: getCompositeProjection yes.  If fact that should remove the allow field list as well.
The getLimitFields() is good however.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
ivakegg: sure, done.

warehouse/query-core/src/main/java/datawave/query/predicate/TLDEventDataFilter.java
ivakegg: good idea

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
ivakegg: yea, I never checked for that.  Curious how it got into this state.  fixing....

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
ivakegg: right, got it

warehouse/query-core/src/main/java/datawave/query/util/ShardQueryUtils.java
ivakegg: yea, that is weird.  removing null checks.

warehouse/query-core/src/test/java/datawave/query/predicate/TLDEventDataFilterTest.java
ivakegg: cool, done

warehouse/query-core/src/main/java/datawave/query/tld/TLDQueryIterator.java
```diff
@@ -133,7 +133,7 @@ public EventDataQueryFilter getFiEvaluationFilter() {
                 // document specific case.
                 fiEvaluationFilter = getEventEvaluationFilter();
             } else {
-                fiEvaluationFilter = new TLDFieldIndexQueryFilter(getIndexOnlyFields());
+                fiEvaluationFilter = new TLDFieldIndexQueryFilter(getNonEventFields());
```
ivakegg: Debating whether this is correct.  This includes the TF fields in addition to what getIndexOnlyFields returned.

PR #3414, opened 2026-02-03
Github allow for readme/doc only changes

jalphonso:
Use a coordinator pattern to call workflows and process the results.  
Tested README.md change in a few places.  
Tested code change in main.  
Tested code change in microservices.  
Tested .github changes.

### DIFF
--- a/.github/workflows/check-formatting.yml
+++ b/.github/workflows/check-formatting.yml
@@ -3,10 +3,6 @@ name: Check Code Formatting
 on:
   workflow_call:
 
-concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  cancel-in-progress: true
-  
 env:
   JAVA_VERSION: '11'
   JAVA_DISTRIBUTION: 'zulu' #This is the default on v1 of the action for 1.8
--- a/.github/workflows/check-formatting.yml
+++ b/.github/workflows/check-formatting.yml
@@ -1,15 +1,7 @@
 name: Check Code Formatting
 
 on:
-  push:
-    paths-ignore: ['**/*.md', '.github/**', 'CODEOWNERS', 'LICENSE']
-    branches:
-    - 'integration'
-    - 'release/version*'
-  pull_request:
-    paths-ignore: ['**/*.md', '.github/**', 'CODEOWNERS', 'LICENSE']
-  merge_group:
-  workflow_dispatch:
+  workflow_call:
 
 concurrency:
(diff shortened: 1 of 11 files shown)


apmoriarty (approved):

alerman (changes requested):
I haven’t had a chance to fully review this. But in a quick glance I have some concerns. Marking as needing changes but may change this after I have a chance to fully review. Feel free to dismiss this if I haven’t commented in a week.

.github/workflows/check-formatting.yml
```diff
@@ -1,20 +1,8 @@
...
-    branches:
-    - 'integration'
-    - 'release/version*'
-  pull_request:
-    paths-ignore: ['**/*.md', '.github/**', 'CODEOWNERS', 'LICENSE']
-  merge_group:
-  workflow_dispatch:
+  workflow_call:
 
-concurrency:
```
alerman, line 14: Removing this will remove the ability for subsequent runs of the same PR to cancel in progress runs. Now each commit will require the build to run to completion.

Its possible adding this in required-checks will cancel that pipeline, but im not sure that called workflows will cancel the same way. Perhaps do some testing with making multiple commits to ensure no longer relevant workflows get cancelled

.github/workflows/required-checks.yml
```diff
@@ -0,0 +1,182 @@
...
+    secrets: inherit
+  
+  build-and-test-main:
+    runs-on: ubuntu-latest
+    needs: [check-changes, call-main-tests]
+    if: always()
+    steps:
+    - name: Report skipped
+      if: needs.check-changes.outputs.should_run_main == 'false'
+      run: echo "✓ Skipped - only documentation files changed"
```
alerman, line 111: This isnt 100% accurate. It should tell you if there were no changes to files in the main build

alerman (changes requested):
Just a couple minor changes i think besided the concurrency testing. Will mark as approved with demonstrated testing of the concurrency isue

alerman (approved):

.github/workflows/check-formatting.yml
jalphonso: concurrency is handled in the top level workflow `required-checks`. I tested and verified pushing a new commit on the PR cancels all of its called workflows

.github/workflows/required-checks.yml
jalphonso: good catch, I'll update in a new PR since this one got merged

PR #3417, opened 2026-02-05
Add IndexExpansionQueryTest

apmoriarty:
Documented current behavior for fielded regex expansion, unfielded regex expansion, unfielded literal expansion and bounded range expansion under a variet of conditions.

Update DelayIterator, IOExceptionIterator and RuntimeExceptionIterator so internal state survives a teardown-rebuild.

Updated MacTestUtil with a compaction method.

Updated AbstractQueryTest and extending classes to support unique authorizations/visibilities per test.

Updated test log4j.properties to suppress noisy messages from MiniAccumuloCluster setup/execution.

Added an EventQuery and DefaultQueryPlanner to the test QueryLogicFactory.xml that does not use pushdown rules. This was done to support the IndexExpansionQueryTest.

### DIFF
--- a/warehouse/query-core/src/test/java/datawave/query/CompositeFunctionsTest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/CompositeFunctionsTest.java
@@ -13,6 +13,7 @@ import java.util.Set;
 import java.util.TimeZone;
 
 import org.apache.accumulo.core.client.AccumuloClient;
+import org.apache.accumulo.core.security.Authorizations;
 import org.apache.log4j.Logger;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
@@ -54,6 +55,8 @@ public class CompositeFunctionsTest extends AbstractQueryTest {
 
     @TempDir
     public static Path folder;
+
+    private static final Authorizations auths = new Authorizations("ALL");
 
     protected static AccumuloClient client = null;
 
@@ -104,6 +107,11 @@ public class CompositeFunctionsTest extends AbstractQueryTest {
     @Override
     public ShardQueryLogic getLogic() {
         return eventQueryLogic;
+    }
+
+    @Override
+    public Authorizations getAuths() {
+        return auths;
     }
 
     @Override
--- a/warehouse/query-core/src/test/java/datawave/query/IndexExpansionQueryTest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/IndexExpansionQueryTest.java
@@ -409,8 +409,6 @@ public class IndexExpansionQueryTest extends AbstractQueryTest {
         try {
             addDelayIterator(timeoutDelayMS);
             givenQuery("_ANYFIELD_ =~ 'a.*'");
-            expectPlan("((_Value_ = true) && (FIELD_A =~ 'a.*'))");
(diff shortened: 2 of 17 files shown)


ivakegg:
Your comment of "Update DelayIterator, IOExceptionIterator and RuntimeExceptionIterator so internal state survives a teardown-rebuild" makes me think.  What if the TimeoutExceptionIterator catches the exception, sets the flag, returns the exceeded value, and then is torn down at that moment.  If lookup attempts to do a "next", the state will no longer be held in the iterator and will subsequently happily continue returning keys.  The lookup mechanism MUST acknowledge the timeout and no longer attempt to invoke that scan any longer.  Might be worth verifying that is the case.

ivakegg (dismissed):

ivakegg (dismissed):

warehouse/query-core/src/test/java/datawave/query/CompositeFunctionsTest.java
```diff
@@ -106,6 +109,11 @@ public ShardQueryLogic getLogic() {
         return eventQueryLogic;
     }
 
+    @Override
+    public Authorizations getAuths() {
+        return auths;
+    }
+
```
FineAndDandy, line 116: not a big deal at all, but these could probably go in the abstract class. They are never different in any impl

warehouse/query-core/src/test/java/datawave/query/IndexExpansionQueryTest.java
```diff
@@ -0,0 +1,1021 @@
...
+        "classpath:MarkingFunctionsContext.xml",
+        "classpath:MetadataHelperContext.xml",
+        "classpath:CacheContext.xml"})
+// @formatter:on
+public class IndexExpansionQueryTest extends AbstractQueryTest {
+
+    private static final Logger log = LoggerFactory.getLogger(IndexExpansionQueryTest.class);
+
+    protected static final String PASSWORD = "password";
+    protected static final Authorizations auths = new Authorizations("VIZ-A", "VIZ-B", "VIZ-C");
```
FineAndDandy, line 99: I stand corrected

warehouse/query-core/src/test/java/datawave/query/IndexExpansionQueryTest.java
```diff
@@ -0,0 +1,1021 @@
...
+            mac = new MiniAccumuloCluster(cfg);
+            mac.start();
+
+            client = mac.createAccumuloClient("root", new PasswordToken(PASSWORD));
+        } else {
+            // do not use a rebuilding scanner because the test requires exceptions to propagate
+            // all the way back to the client.
+            InMemoryInstance instance = new InMemoryInstance(IndexExpansionQueryTest.class.getName());
+            client = new InMemoryAccumuloClient("root", instance);
+        }
```
FineAndDandy, line 210: this could be pulled up to the base class

FineAndDandy (dismissed):

warehouse/query-core/src/test/java/datawave/query/IndexExpansionQueryTest.java
apmoriarty: That's what a smart person would do. Will put up a separate merge request for that soon.

FineAndDandy (approved):

ivakegg (approved):

PR #3438, opened 2026-02-19
Add ScanMonitor for external scan monitoring

apmoriarty:
Add ScanMonitor for external scan monitoring

BoundedRangeIndexLookup now delegates scan termination responsibility to ScanMonitor.

Add EmptyCloseableIterator and EmptyQueryPlanStream for ease of testing

Update bounded range tests in IndexExpansionQueryTest that previously documented incorrect behavior with the correct final state, i.e. bounded ranges marked as value exceeded in the case of an exception.

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BoundedRangeIndexLookup.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BoundedRangeIndexLookup.java
@@ -4,41 +4,33 @@ import static datawave.query.jexl.lookups.ShardIndexQueryTableStaticMethods.EXPA
 
 import java.text.MessageFormat;
 import java.util.Collections;
-import java.util.Iterator;
-import java.util.Map.Entry;
-import java.util.concurrent.Callable;
+import java.util.Map;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
-import java.util.concurrent.atomic.AtomicLong;
 
-import org.apache.accumulo.core.client.BatchScanner;
 import org.apache.accumulo.core.client.IteratorSetting;
-import org.apache.accumulo.core.client.TableNotFoundException;
 import org.apache.accumulo.core.data.Key;
 import org.apache.accumulo.core.data.Range;
 import org.apache.accumulo.core.data.Value;
+import org.apache.accumulo.core.security.Authorizations;
 import org.apache.hadoop.io.Text;
-import org.apache.log4j.Logger;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.util.StringUtils;
 
 import com.google.common.base.Preconditions;
 
-import datawave.core.common.logging.ThreadConfigurableLogger;
 import datawave.core.iterators.BoundedRangeExpansionIterator;
 import datawave.core.iterators.CompositeSeekingIterator;
-import datawave.core.iterators.TimeoutExceptionIterator;
-import datawave.core.iterators.TimeoutIterator;
 import datawave.data.type.DiscreteIndexType;
 import datawave.query.Constants;
 import datawave.query.config.ShardQueryConfiguration;
-import datawave.query.exceptions.DatawaveFatalQueryException;
 import datawave.query.exceptions.IllegalRangeArgumentException;
 import datawave.query.jexl.LiteralRange;
 import datawave.query.tables.ScannerFactory;
 import datawave.util.time.DateHelper;
 import datawave.webservice.query.exception.DatawaveErrorCode;
-import datawave.webservice.query.exception.NotFoundQueryException;
 import datawave.webservice.query.exception.QueryException;
 
 /**
@@ -47,16 +39,21 @@ import datawave.webservice.query.exception.QueryException;
  * A fielded bounded range is already executable so this lookup is allowed to hit timeout or value thresholds.
  */
 public class BoundedRangeIndexLookup extends AsyncIndexLookup {
-    private static final Logger log = ThreadConfigurableLogger.getLogger(BoundedRangeIndexLookup.class);
+
+    private static final Logger log = LoggerFactory.getLogger(BoundedRangeIndexLookup.class);
 
     private final LiteralRange<?> literalRange;
+    protected final String field;
+    protected final CountDownLatch startLatch;
+    protected final CountDownLatch stopLatch;
 
-    protected Future<Boolean> timedScanFuture;
-    protected AtomicLong lookupStartTimeMillis = new AtomicLong(Long.MAX_VALUE);
-    protected CountDownLatch lookupStartedLatch;
-    protected CountDownLatch lookupStoppedLatch;
+    // variables to support range creation
+    private final String startDay;
+    private final String endDay;
+    private final String lower;
+    private final String upper;
 
-    protected BatchScanner bs;
+    private final Range scanRange;
 
     /**
      *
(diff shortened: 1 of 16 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ScanMonitor.java
```diff
@@ -0,0 +1,126 @@
...
+     * Registers the future along with the timeout
+     *
+     * @param future
+     *            the future
+     * @param timeout
+     *            the timeout
+     */
+    public void registerTask(Future<?> future, long timeout) {
+        synchronized (tasks) {
+            log.info("registering task: {}", taskID);
```
jschmidt10: Its best to keep the synchronized block as lean as possible. The logging and the IndexScanTask constructor can both be called before grabbing the lock. Same applies the function below.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ScanMonitor.java
```diff
@@ -0,0 +1,126 @@
...
+                            tasks.remove(key);
+                        }
+                    }
+                }
+            }
+
+            try {
+                Thread.sleep(monitorIntervalMillis);
+            } catch (InterruptedException e) {
+                log.warn("interrupted");
```
jschmidt10: Since we don't warn on the interrupt above, I would just log the same info. This isn't a problem, just the thread being stopped.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ScanMonitor.java
```diff
@@ -0,0 +1,126 @@
...
+
+            // always check for interrupts first
+            if (Thread.currentThread().isInterrupted()) {
+                log.info("thread interrupted, stopping");
+                break;
+            }
+
+            synchronized (tasks) {
+                long currentTime = System.currentTimeMillis();
+                Set<String> keys = new HashSet<>(tasks.keySet());
```
jschmidt10: I'm guessing that we are making this Set b/c we are removing keys below. If you use the Maps iterator method, you don't need to create this 2nd Set.

```
var iter = tasks.keySet().iterator();
while (iter.hasNext()) 
  if (...) 
    cancel task
    iter.remove();
}
```

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ScanMonitor.java
```diff
@@ -0,0 +1,126 @@
...
+                for (String key : keys) {
+                    IndexScanTask task = tasks.get(key);
+                    if (task.isDone(currentTime)) {
+                        log.info("closing task {}", key);
+                        task.cancelFuture();
+                        synchronized (tasks) {
```
jschmidt10: Since we already grabbed the lock above, do we need the 2nd synchronized here?

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ScanMonitor.java
```diff
@@ -0,0 +1,126 @@
...
+
+    private static final Logger log = LoggerFactory.getLogger(ScanMonitor.class);
+
+    private final Map<String,IndexScanTask> tasks = new HashMap<>();
+
+    private int taskID = 0;
+    private final long monitorIntervalMillis;
+
+    public ScanMonitor() {
+        this(5L);
```
jschmidt10: Every 5 millis is probably too tight of a loop. It's a lock on the tasks Map which means we're delaying inserts every 5 millis. I know we'll probably set this higher in config or something but I'd increase this default as well. If this is just cleaning up lingering scans, it seems like something on the order of seconds is probably reasonable?

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ScanMonitor.java
```diff
@@ -0,0 +1,126 @@
...
+        public IndexScanTask(Future<?> future, CountDownLatch latch, long timeout) {
+            this.future = future;
+            this.start = System.currentTimeMillis();
+            this.timeout = timeout;
+            this.latch = latch;
+        }
+
+        public boolean isDone(long current) {
+            long elapsed = current - start;
+            return elapsed >= timeout;
```
jschmidt10: Wondering if this should be OR'd with `|| future.isDone()`. I haven't fully worked thru the code but this appears to mean if the task finished before the timeout, that we will still wait for the timeout.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BoundedRangeIndexLookup.java
```diff
@@ -73,210 +70,224 @@ public BoundedRangeIndexLookup(ShardQueryConfiguration config, ScannerFactory sc
...
-                Text holder = new Text();
-                try {
-                    if (log.isTraceEnabled()) {
-                        log.trace("Do we have next? " + iter.hasNext());
+    /**
+     * Wait for the future to complete before returning the {@link IndexLookupMap}
+     */
+    protected void await() {
+
+        synchronized (startLatch) {
```
jschmidt10: I'm not totally following this part yet. Synchronizing on the CountDownLatch feels dubious. The CountDownLatch internally uses a Sync object to control locking so synchronizing on its object monitor doesn't prevent any methods from being called, just someone else synchronizing on the same object monitor. 

For the cases below, `startLatch.await()` will immediately return if the latch has already been decremented to 0 so I think both of your cases can be treated the same. We can just call `startLatch.await()` and continue.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BoundedRangeIndexLookup.java
```diff
@@ -73,210 +70,224 @@ public BoundedRangeIndexLookup(ShardQueryConfiguration config, ScannerFactory sc
...
-                        indexLookupMap.setKeyThresholdExceeded();
+        synchronized (stopLatch) {
+            switch ((int) stopLatch.getCount()) {
+                case 0:
+                    // scan task completed prior to calling await(), nothing left to do
+                    break;
+                case 1:
+                    // scan task executing, wait for completion
+                    try {
+                        stopLatch.await();
```
jschmidt10: Given that we wait for the startLatch and then immediately wait for the stopLatch makes it look like we do not care about the startLatch at all. Can we get rid of it and just have the stopLatch?

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ScanMonitor.java
```diff
@@ -0,0 +1,126 @@
...
+     *            the timeout
+     */
+    public void registerTask(Future<?> future, long timeout) {
+        synchronized (tasks) {
+            log.info("registering task: {}", taskID);
+            tasks.put(String.valueOf(taskID++), new IndexScanTask(future, timeout));
+        }
+    }
+
+    public void registerTask(Future<?> future, CountDownLatch latch, long timeout) {
```
jschmidt10: I like the try/finally structure with the stopLatch in BoundedRangeIndexLookup rather than passing in a latch here. Are you planning on using this method? I'm just thinking it might get tricky if both this class and the calling class are fighting over the latch object.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/BaseIndexExpansionVisitor.java
```diff
@@ -94,6 +99,21 @@ protected void shutdownExecutor() {
         }
     }
 
+    protected void setupMonitorExecutor() {
+        String id = "(unknown)";
+        if (config.getQuery() != null && config.getQuery().getId() != null) {
+            id = config.getQuery().getId().toString();
+        }
+        this.monitorExecutor = Executors.newSingleThreadExecutor(new ScanMonitorThreadFactory(id, config.getQuery().getUncaughtExceptionHandler()));
```
jschmidt10: Imagine we want to use a ScanMonitor in another class eventually. That new class has to setup this executor and make the same ThreadFactory. Wondering if we can make ScanMonitor do that stuff internally so using it requires less boiler plate. We could move the ThreadFactory definition in ScanMonitor and give it a static method:

```
this.monitorExecutor = ScanMonitor.createMonitorExecutor(id, uncaughtExceptionHandler);
```
Another thought that would hide even more implementation detail from the callers would be if the ScanMonitor was changed to manage the executor itself as well. So the code you have in ScanMonitor now could be put in ScanMonitorTask and then your code here might look like:

```
// create it
this.scanMonitor = new ScanMonitor(id, uncaughtExceptionHandler);

// tear down 
scanMonitor.shutdown();
```

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ScanMonitor.java
apmoriarty: Doubtful. This was leftover from an earlier attempt that had locking to grab the set of tasks and then locking per removal. This whole section can be refactored with your suggestions.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ScanMonitor.java
apmoriarty: Nope, this can be removed

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BoundedRangeIndexLookup.java
apmoriarty: Removing the synchronization on the latch is on my list, there's no reason for that.

The problem with simply calling `latch.await()` is that the latch might already be decremented. In that case the call to `await` will hang.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BoundedRangeIndexLookup.java
apmoriarty: There's an error condition that exists because the existing code did not differentiate enough between a task that fails on startup (or fails before returning the first key) and a task that fails after seeing one or more keys.

I need to add context for when a failure occurs, something like `markExceededWithContext(String context)` where context is "START" or "STOP"

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/BaseIndexExpansionVisitor.java
apmoriarty: I like it.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BoundedRangeIndexLookup.java
jschmidt10: If it was already decremented, this method shouldn't hang. It should return immediately.

apmoriarty: Verified that removing it had no ill effect

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ScanMonitor.java
apmoriarty: Changed this to 25 ms. Because index expansion is often the longest phase of query planning I don't want to increase the default interval any longer than that.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ScanMonitor.java
apmoriarty: Done. Should have been that way from the beginning. As it currently stands for expansion tasks that succeed the executor will be shutdown forcibly when the visitor finishes executing (that is, no opportunity to remove tasks at all).

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/BaseIndexExpansionVisitor.java
apmoriarty: Updated

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ScanMonitor.java
```diff
@@ -0,0 +1,194 @@
...
+ * No further configuration is necessary after calling {@link ScanMonitor#of(String, QueryUncaughtExceptionHandler)}
+ */
+public class ScanMonitor implements Runnable, Closeable {
+
+    private static final Logger log = LoggerFactory.getLogger(ScanMonitor.class);
+
+    private static final long DEFAULT_INTERVAL_MILLIS = 25L;
+
+    private int taskID = 0;
+    private final Map<String,IndexScanTask> tasks = new HashMap<>();
```
FineAndDandy: This is a good candidate for a ConcurrentHashMap. It would simplify the code and also improve performance.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ScanMonitor.java
```diff
@@ -0,0 +1,194 @@
...
+            if (currentThread().isInterrupted()) {
+                if (log.isDebugEnabled()) {
+                    log.debug("thread interrupted, stopping");
+                }
+                break;
+            }
+
+            synchronized (tasks) {
+                long currentTime = System.currentTimeMillis();
+                var iter = tasks.keySet().iterator();
```
FineAndDandy: use strong typing

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ScanMonitor.java
```diff
@@ -0,0 +1,194 @@
...
+        private final long start;
+        private final long timeout;
+
+        public IndexScanTask(Future<?> future, long timeout) {
+            this.future = future;
+            this.start = System.currentTimeMillis();
+            this.timeout = timeout;
+        }
+
+        public boolean isDone(long current) {
```
FineAndDandy: This name is a bit misleading, mostly because of the future api overlap. Would you be open to renaming this cleanup() or something similar?

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BoundedRangeIndexLookup.java
```diff
@@ -73,210 +70,226 @@ public BoundedRangeIndexLookup(ShardQueryConfiguration config, ScannerFactory sc
...
+    }
 
-            log.debug("Range: " + range);
-            bs = null;
-            try {
-                // the 'newScanner' method in the ScannerFactory has no knowledge about the 'expansion' hint, so determine hint here
+    protected Runnable createRunnable(String tableName, Authorizations auths) {
+        return () -> {
+            startLatch.countDown();
+            try (var scanner = config.getClient().createScanner(tableName, auths)) {
```
FineAndDandy: add strongly typed variable

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BoundedRangeIndexLookup.java
```diff
@@ -73,210 +70,226 @@ public BoundedRangeIndexLookup(ShardQueryConfiguration config, ScannerFactory sc
...
+        setting.addOption(CompositeSeekingIterator.SEPARATOR, compositeSeparator);
+        return setting;
+    }
 
-                        // safety check...
-                        Preconditions.checkState(field.equals(literalRange.getFieldName()),
-                                        "Got an unexpected field name when expanding range" + field + " " + literalRange.getFieldName());
+    @Override
+    public synchronized IndexLookupMap lookup() {
+        await();
```
FineAndDandy, line 257: the synchronized, await(), and latches add a layer of complexity I don't believe is necessary here. Each BoundedRangeIndexLookup is used exactly once. This method can be updated to wait on the future if it exists. 

See https://github.com/NationalSecurityAgency/datawave/commit/b941fab9456207e805d14f887b88f3692d80fcab

branch: feature/index-expansion-scan-monitor-future

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ScanMonitor.java
apmoriarty: Done in next commit

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ScanMonitor.java
apmoriarty: Done in next commit

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BoundedRangeIndexLookup.java
apmoriarty: _fine_

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/ScanMonitor.java
apmoriarty: Sure

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BoundedRangeIndexLookup.java
apmoriarty: Yup

avgAGB (approved):

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BoundedRangeIndexLookup.java
```diff
@@ -73,210 +70,214 @@ public BoundedRangeIndexLookup(ShardQueryConfiguration config, ScannerFactory sc
...
+
+        // create the scan range here to surface an illegal range exception
+        this.scanRange = createScanRange(lower, upper, endDay);
     }
 
+    /**
+     * This method is responsible for creating a Runnable, submitting it to the executor and registering the future with the {@link ScanMonitor}.
+     */
     @Override
     public synchronized void submit() {
```
FineAndDandy, line 90: this synchronized can be droppped

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BoundedRangeIndexLookup.java
```diff
@@ -73,210 +70,214 @@ public BoundedRangeIndexLookup(ShardQueryConfiguration config, ScannerFactory sc
...
+        setting.addOption(CompositeSeekingIterator.SEPARATOR, compositeSeparator);
+        return setting;
+    }
 
-                        // obtaining the size of a map can be expensive,
-                        // instead
-                        // track the count of each unique item added.
-                        indexLookupMap.put(field, uniqueTerm);
+    @Override
+    public synchronized IndexLookupMap lookup() {
```
FineAndDandy, line 256: synchronized can be dropped

warehouse/query-core/src/test/java/datawave/query/jexl/lookups/BoundedRangeIndexLookupTest.java
```diff
@@ -282,6 +290,7 @@ private Set<String> createAllValues(int start, int stop) {
         return values;
     }
 
+    @Ignore
```
FineAndDandy, line 293: was this accidentally left ignored?

warehouse/query-core/src/test/java/datawave/query/planner/GeoSortedQueryDataTest.java
```diff
@@ -272,15 +272,21 @@ private static void writeKeyValues(AccumuloClient client, Multimap<BulkIngestKey
             }
             writer.close();
         }
+
+        try (var bw = client.createBatchWriter(TableName.METADATA, new BatchWriterConfig())) {
```
FineAndDandy, line 276: strongly typed variables please

FineAndDandy (approved):
minor things, not going to block merge

PR #3440, opened 2026-02-23
Use the HIT_TERM_FIELD to set the FieldedTagCloudInputExtractor source

FineAndDandy:


apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/tables/keyword/extractor/FieldedTagCloudInputExtractor.java
```diff
@@ -38,6 +38,11 @@ public FieldedTagCloudInputExtractor() {}
 
     @Override
     public void extract(Key source, Map<String,Attribute<? extends Comparable<?>>> documentData) {
+        if (fields == null) {
```
drewfarris, line 41: Consider an empty check here too? (shouldn't block merge in itself..)

warehouse/query-core/src/main/java/datawave/query/tables/keyword/extractor/FieldedTagCloudInputExtractor.java
```diff
@@ -127,6 +142,12 @@ public int hashCode() {
         return Objects.hash(fields, fieldToScoreField, minScore, category, subType);
     }
 
+    private String getDocIdFromHitTerm(Map<String,Attribute<? extends Comparable<?>>> documentData) {
+        Attribute<?> hitTermAttribute = documentData.get(HIT_TERM_FIELD);
+        List<String> identifiers = KeywordQueryUtil.getStringValuesFromAttribute(hitTermAttribute);
+        return KeywordQueryUtil.chooseBestIdentifier(identifiers);
```
drewfarris, line 152: It would be strange that `hitTermAttribue` would be null, but it probably doesn't hurt to be a bit more defensive, especially because `getStringValuesFromAttribute` doesn't do a null check and would probably throw a `NullPointerException` if the `inputAttribute` parameter was `null`. (sorry!)

FWIW, it appears that `identifiers` returned by `getStringValuesFromAttribute` won't ever be null, but it might be empty (and `chooseBestIdentifier` does do a null check).
```suggestion
        Attribute<?> hitTermAttribute = documentData.get(HIT_TERM_FIELD);
        if (hitTermAttribute == null) {
            return null;
        }
        List<String> identifiers = KeywordQueryUtil.getStringValuesFromAttribute(hitTermAttribute);
        return KeywordQueryUtil.chooseBestIdentifier(identifiers);
```

drewfarris (dismissed):
Minor nits that shouldn't block taking these changes.

warehouse/query-core/src/main/java/datawave/query/tables/keyword/extractor/FieldedTagCloudInputExtractor.java
FineAndDandy: agree. I will add a null checks and verify test coverage.

drewfarris (approved):

d-hwang (approved):

PR #3443, opened 2026-02-24
Fixes per-request AnnotationManager resource leakage by introducing a request context object

drewfarris:
The AnnotationManagerBean's lifecycle as a stateless bean meant we were caching some objects across requests. This updates the AnnotationManagerBean to maintain per-reuqest state explicitly instead of assuming the bean framework will do it for us.

### DIFF
--- a/web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
+++ b/web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
@@ -88,106 +88,9 @@ public class AnnotationManagerBean implements AnnotationManager {
     @SpringBean(name = "AnnotationManagerConfig")
     private AnnotationManagerConfig config;
 
-    // Per-request stateful variables, managed internally.
-    private Set<Authorizations> authorizations;
-    private AccumuloClient client;
-    private LookupUUIDService lookupUUIDService;
-    private AnnotationDataAccess annotationDataAccess;
-
-    // Per-request, cache lookups for unique analytic source hashes so we don't do them more than once.
-    // TODO: make this a proper cross-request cache?
-    private final Map<String,Optional<AnnotationSource>> retrievedSourcesCache = new HashMap<>();
-
     @VisibleForTesting
     public void setEJBContext(EJBContext ctx) {
         this.ctx = ctx;
-    }
-
-    /**
-     * Calculate the auths for this query based on the principal and the queryAuths if any.
-     *
-     * @param queryAuths
-     *            the auths to downgrade to, or null if no downgrade is desired
-     * @param principal
-     *            the caller principal
-     * @return a Set of authorizations to use for this query
-     * @throws QueryException
-     *             and exception if there was a problem calculating the auths
-     */
-    private static Set<Authorizations> getUserAuthorizations(@SuppressWarnings("SameParameterValue") String queryAuths, DatawavePrincipal principal)
-                    throws QueryException {
-        try {
-            if (queryAuths == null) {
-                return AuthorizationsUtil.buildAuthorizations(principal.getAuthorizations());
-            } else {
-                final String downgradedAuths = AuthorizationsUtil.downgradeUserAuths(queryAuths, principal, principal);
-                return AuthorizationsUtil.buildAuthorizations(Collections.singleton(AuthorizationsUtil.splitAuths(downgradedAuths)));
-            }
-
-        } catch (Exception e) {
-            throw new QueryException("Failed to get user query authorizations", e);
-        }
-    }
-
-    public AccumuloClient initializeAccumuloClient() throws QueryException {
-        if (client == null || authorizations == null) {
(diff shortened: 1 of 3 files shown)


web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
```diff
@@ -708,4 +617,142 @@ private static Response jsonOk(Object responseObject) {
...
+         * @return a Set of authorizations to use for this query
+         * @throws QueryException
+         *             and exception if there was a problem calculating the auths
+         */
+        private Set<Authorizations> initializeAuthorizations(@SuppressWarnings("SameParameterValue") String queryAuths, DatawavePrincipal principal)
+                        throws QueryException {
+            if (authorizations == null) {
+                try {
+                    if (queryAuths == null) {
+                        return AuthorizationsUtil.buildAuthorizations(principal.getAuthorizations());
```
FineAndDandy: principal may be null here

web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
```diff
@@ -708,4 +617,142 @@ private static Response jsonOk(Object responseObject) {
...
+                    } else {
+                        final String downgradedAuths = AuthorizationsUtil.downgradeUserAuths(queryAuths, principal, principal);
```
FineAndDandy: This will throw an NPE if principal is null

web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
```diff
@@ -708,4 +617,142 @@ private static Response jsonOk(Object responseObject) {
...
+                    trackingMap.put("query.query", "something else");
+                }
+                accumuloConnectionRequestBean.requestBegin(transactionUUID.toString(), userDn, trackingMap);
+                try {
+                    client = connectionFactory.getClient(userDn, proxyServers, config.getConnPoolName(), config.getPriority(), trackingMap);
+                } catch (Exception e) {
+                    throw new QueryException("Unable to get Accumulo client, exception encountered: ", e);
+                } finally {
+                    accumuloConnectionRequestBean.requestEnd(transactionUUID.toString());
+                }
```
FineAndDandy, line 763: if the client is not null, this will borrow a new one without returning it. I don't see a way for this to happen, but consider only replacing the client if null

web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
```diff
@@ -708,4 +617,142 @@ private static Response jsonOk(Object responseObject) {
...
+        private AccumuloClient client;
+
+        /** used to lookup uuids and obtain internal identifiers, scoped to the caller's authorizations */
+        private LookupUUIDService lookupUUIDService;
+
+        /** used to _read_ annotations directly from accumulo, scoped to the caller's authorizations */
+        private AnnotationDataAccess annotationDataAccess;
+
+        /** Cache lookups for unique analytic source hashes so we don't perform lookups more than once. TODO: make this a proper cross-request cache? */
+        private final Map<String,Optional<AnnotationSource>> retrievedSourcesCache = new HashMap<>();
```
d-hwang, line 654: I know this isn't a related to this change, but sources cache could have extra visibility that could be pulled based on which gets pulled first. FWIW, I'm working on a ticket, so we just inherit the source visibility from the annotation's.

d-hwang (commented):
one comment about annotationSource cache, but LGTM otherwise

web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
drewfarris: Cleaned up initialization code so that entry into the block only checks whether the `client` is `null`.

FineAndDandy (dismissed):

web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
drewfarris: added null check + exception

web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
drewfarris: added null check + exception

web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
drewfarris: Sounds good, I'll leave this alone for now.

FineAndDandy (approved):

d-hwang (approved):
LGTM!

PR #3446, opened 2026-02-27
Integrate ScanMonitor into RegexIndexLookups

apmoriarty:
Integrate ScanMonitor into RegexIndexLookups

Deprecate RegexIndexLookup

Fix error case when an unfielded term fails before returning any results. Previously exception handling logic required at least one field to be present and that is not always the case.

Update unit tests with correct assertions

Remove exception tracking from AsyncIndexLookup

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BaseRegexIndexLookup.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BaseRegexIndexLookup.java
@@ -2,16 +2,14 @@ package datawave.query.jexl.lookups;
 
 import static datawave.query.jexl.lookups.ShardIndexQueryTableStaticMethods.EXPANSION_HINT_KEY;
 
-import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
 
 import org.apache.accumulo.core.client.IteratorSetting;
 import org.apache.accumulo.core.data.Range;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import datawave.core.iterators.TimeoutExceptionIterator;
-import datawave.core.iterators.TimeoutIterator;
 import datawave.query.config.ShardQueryConfiguration;
 import datawave.query.tables.ScannerFactory;
 
@@ -26,7 +24,7 @@ public abstract class BaseRegexIndexLookup extends AsyncIndexLookup {
     protected final Range range;
     protected final boolean reverse;
 
-    protected final CountDownLatch latch;
+    protected Future<?> future;
 
     // used when scanning the shard reverse index
     private final StringBuilder sb = new StringBuilder();
@@ -37,7 +35,6 @@ public abstract class BaseRegexIndexLookup extends AsyncIndexLookup {
         this.pattern = pattern;
         this.range = range;
         this.reverse = reverse;
-        this.latch = new CountDownLatch(1);
     }
 
     protected String getTableName() {
@@ -52,13 +49,6 @@ public abstract class BaseRegexIndexLookup extends AsyncIndexLookup {
         return config.getTableHints().containsKey(EXPANSION_HINT_KEY) ? EXPANSION_HINT_KEY : tableName;
     }
 
-    protected IteratorSetting createTimeoutIterator() {
-        long maxTime = (long) (config.getMaxIndexScanTimeMillis() * 1.25);
-        IteratorSetting iterator = new IteratorSetting(5, TimeoutIterator.class);
-        iterator.addOption(TimeoutIterator.MAX_SESSION_TIME, Long.valueOf(maxTime).toString());
-        return iterator;
-    }
-
     /**
      * Extending classes decide how to build a regex iterator depending on if the regex is fielded or not
      *
@@ -66,26 +56,23 @@ public abstract class BaseRegexIndexLookup extends AsyncIndexLookup {
      */
     protected abstract IteratorSetting createRegexIterator();
 
-    protected IteratorSetting createTimeoutExceptionIterator() {
-        return new IteratorSetting(config.getBaseIteratorPriority() + 100, TimeoutExceptionIterator.class);
-    }
-
     /**
      * Waits for the future to complete before returning the index lookup map
      */
     protected void await() {
-        synchronized (latch) {
-            if (latch.getCount() == 1) {
-                try {
-                    latch.await();
-                } catch (InterruptedException e) {
-                    log.error(e.getMessage(), e);
-                    Thread.currentThread().interrupt();
-                    throw new RuntimeException(e);
-                }
+        try {
+            if (future != null) {
+                future.get();
             }
+        } catch (Exception e) {
+            handleException();
         }
     }
+
+    /**
+     * Extending classes deal with exceptions differently
+     */
+    protected abstract void handleException();
 
     /**
      * Reverses the value coming off the shard reverse index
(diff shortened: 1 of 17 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/lookups/UnfieldedRegexIndexLookup.java
```diff
@@ -119,4 +120,28 @@ public IndexLookupMap lookup() {
...
+     */
+    protected void await() {
+        try {
+            if (future != null) {
+                future.get();
+            }
+        } catch (Exception e) {
+            handleException();
+        }
+    }
```
FineAndDandy: this can be removed (same as base class)

FineAndDandy (changes requested):
one minor thing

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BaseRegexIndexLookup.java
```diff
@@ -26,7 +24,7 @@ public abstract class BaseRegexIndexLookup extends AsyncIndexLookup {
     protected final Range range;
     protected final boolean reverse;
 
-    protected final CountDownLatch latch;
+    protected Future<?> future;
```
drewfarris, line 27: Did you consider pushing the whole Future monitoring mechanism here into `AsyncIndexLookup` instead of `BaseRegexIndexLookup`? The `monitor` is defined there, so perhaps it makes sense to do so there? (Not a blocker for approval, just wanted to ask..)

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/FieldedRegexIndexLookup.java
```diff
@@ -46,63 +50,53 @@ public void submit() {
...
-                        }
+            Preconditions.checkNotNull(monitor, "FieldedRegexIndexLookup requires a ScanMonitor");
+            Runnable runnable = createRunnable();
 
-                        String value = key.getRow().toString();
-                        if (reverse) {
-                            value = reverse(value);
-                        }
+            future = execService.submit(runnable);
+            monitor.registerTask(future, config.getMaxIndexScanTimeMillis());
```
drewfarris, line 60: I expected to see code related to cleaning up the monitor if tasks fail or complete, but from what I can tell, the Monitor is created in the visitor and created each time expand is called, so this is likely not necessary.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/UnfieldedRegexIndexLookup.java
```diff
@@ -51,53 +52,53 @@ public void submit() {
         if (indexLookupMap == null) {
             indexLookupMap = new IndexLookupMap(keyThreshold, valueThreshold);
 
-            execService.submit(() -> {
-                String tableName = reverse ? config.getReverseIndexTableName() : getTableName();
-                try (var scanner = config.getClient().createScanner(tableName, config.getAuthorizations().iterator().next())) {
-                    String hintKey = getHintKey(tableName);
-                    scanner.setExecutionHints(Map.of(tableName, hintKey));
+            Preconditions.checkNotNull(monitor, "UnfieldedRegexIndexLookup requires a ScanMonitor");
```
drewfarris, line 55: Consider comparing the `submit()` method in `FieldedRegexIndexLookup` and `UnfieldedRegexIndexLookup` - perhaps `UnfieldedRegexIndexLookup` should have the same short circuit that `FieldedRegexIndexLookup` has? Perhaps the submit method could be imlemented in `BaserRegexIndexLookup`?

There are a few other methods that might be good candidates to push to the Base class like `handleException`, `createRegexIterator` and `createRunnable`. 

Bottom line, `FieldedRegexIndexLookup` and `UnfieldedRegexIndexLookup` are quite similar, the former being a special case of the latter with a list the size of 1, so maybe there's an option to collapse these two into a single class? (Not a blocker for acceptance of this PR, food for thought..)

warehouse/query-core/src/test/java/datawave/query/IndexExpansionQueryTest.java
```diff
@@ -106,7 +106,7 @@ public class IndexExpansionQueryTest extends AbstractQueryTest {
     private static final boolean useMAC = false;
 
     // switch between old code and new code
-    private final boolean useNewIndexLookups = false;
+    private final boolean useNewIndexLookups = true;
```
drewfarris, line 109: Do we want to test both the old and new code so we're sure that each continues to work as expected?

drewfarris (approved):
Added some food for thought, nothing found to block taking this change as-is.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BaseRegexIndexLookup.java
```diff
@@ -52,41 +49,31 @@ protected String getHintKey(String tableName) {
...
-                    log.error(e.getMessage(), e);
-                    Thread.currentThread().interrupt();
-                    throw new RuntimeException(e);
-                }
+        try {
+            if (future != null) {
+                future.get();
             }
+        } catch (Exception e) {
+            handleException();
```
ivakegg, line 68: Should we not be passing the exception into handleException() ?  At least so that we can log it ?

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/UnfieldedRegexIndexLookup.java
apmoriarty: I would love to implement those suggestions, but not as part of this merge request. The goal with this merge request to migrate the exception handling for the two specific index lookups without any changes to behavior. This means carrying along some of the existing inconsistencies.

There is a forthcoming change that will consolidate common code in the index lookups and I will make a note to move the `submit()` method into the base class. After taking a second look it should be fairly straightforward.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BaseRegexIndexLookup.java
apmoriarty: That's what a smart person would do. I'll add that in the next unit of work when I migrate the field expansion lookup to using the scan monitor.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/FieldedRegexIndexLookup.java
apmoriarty: The monitor is closed when the visitor is done executing. A thread inside the monitor handles cancelling and removing tasks if they succeed, fail, or exceed the allotted run time.

warehouse/query-core/src/test/java/datawave/query/IndexExpansionQueryTest.java
apmoriarty: That would be ideal. I'll do that in the next unit of work if that's alright.

warehouse/query-core/src/main/java/datawave/query/jexl/lookups/BaseRegexIndexLookup.java
apmoriarty: Yes, that's part of the work to consolidate index lookups. Current index looking (both existing and new) handle futures in each lookup and those should absolutely be pushed into the base class.

FineAndDandy (approved):

PR #3447, opened 2026-02-27
Add subtype filtering for tag clouds

FineAndDandy:


warehouse/query-core/src/main/java/datawave/query/tables/keyword/extractor/TagCloudInputExtractor.java
```diff
@@ -28,6 +28,8 @@ default void initialize(Query settings) {}
 
     String getName();
 
+    String getSubType();
```
ivakegg, line 89: Just a minor nit, but given this is an interface the methods should be documented if not obvious.  Can do on a separate PR of course.

warehouse/query-core/src/main/java/datawave/query/tables/keyword/KeywordQueryUtil.java
```diff
@@ -52,7 +58,12 @@ public static String chooseBestLanguage(List<String> languages) {
 
         // if we get here, we couldn't find an ideal language, just return the first value, yake will default
         // to processing the data as if it were English.
-        return languages.get(0);
+        if (preferredLanguages == null || preferredLanguages.isEmpty()) {
+            return languages.get(0);
+        } else {
+            Optional<String> first = languages.stream().filter(preferredLanguages::contains).findFirst();
```
ivakegg: We could have gotten here because there was no intersection between the languages and the preferred languages.  Do we want to test the languages against the registry here?

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/tables/keyword/extractor/TagCloudInputExtractor.java
FineAndDandy: added javadoc

warehouse/query-core/src/main/java/datawave/query/tables/keyword/KeywordQueryUtil.java
FineAndDandy: added small refactor to make this more clear

ivakegg (approved):

apmoriarty (approved):

drewfarris (approved):

PR #3449, opened 2026-03-02
Replace ClientContext/TabletLocator/MetadataServicer with AccumuloTableInfoFetcher facade

SethSmucker:
## Summary
- Create `AccumuloTableInfoFetcher` facade in core/connection-pool centralizing Accumulo metadata operations behind public APIs
- Replace Thrift RPC in `BulkIngestMapFileLoader.getMajorCompactionCount()` with `getActiveCompactions()` API
- Replace `MetadataServicer` in `TableSplitsCache` with `locate()` API
- Replace `TabletLocator`/`ClientContext` in `BulkInputFormat` online path with `locate()` API
- Remove `TabletLocator` from import-control-accumulo.xml allowlist

Part of #2443


core/connection-pool/src/main/java/datawave/core/common/connection/AccumuloTableInfoFetcher.java
```diff
@@ -0,0 +1,117 @@
...
+public class AccumuloTableInfoFetcher {
+
+    private final AccumuloClient client;
+
+    public AccumuloTableInfoFetcher(AccumuloClient client) {
+        this.client = client;
+    }
+
+    /**
+     * Get the TableId for a table name using the public {@code tableIdMap()} API.
```
jschmidt10: I'd drop all the "using the public ... API" in your method javadocs. In OO programming, we want to insulate the user from the details. They just need to know that it'll give them the table id.

core/connection-pool/src/main/java/datawave/core/common/connection/AccumuloTableInfoFetcher.java
```diff
@@ -0,0 +1,117 @@
...
+    public Locations getTabletLocations(String tableName, Collection<Range> ranges)
+                    throws TableNotFoundException, AccumuloException, AccumuloSecurityException {
+        return client.tableOperations().locate(tableName, ranges);
+    }
+
+    /**
+     * Get the count of running major compactions across all tablet servers using the public {@code getActiveCompactions()} API.
+     * <p>
+     * Note: This counts only running compactions (not queued), which differs slightly from the original Thrift-based implementation that also counted queued
+     * compactions. This is acceptable because the MAJC_THRESHOLD default is 3000 (a high safety margin) and this is polled on each bulk load cycle.
```
jschmidt10: I'm not sure relying solely on active will work for us. On some of the instances, the tservers never really stop compacting so having a threshold on the active would either always trigger or never trigger. The queued really illustrates the backlog. Thoughts @ivakegg @hlgp ?

warehouse/core/src/main/java/datawave/mr/bulk/BulkInputFormat.java
```diff
@@ -1101,28 +1071,18 @@ public List<InputSplit> getSplits(JobContext job) throws IOException {
...
-                        }
-                        binnedRanges.clear();
-                        log.warn("Unable to locate bins for specified ranges. Retrying.");
-                        TimeUnit.MILLISECONDS.sleep(ThreadLocalRandom.current().nextInt(100, 200));
-                        tl.invalidateCache();
+                    AccumuloTableInfoFetcher fetcher = new AccumuloTableInfoFetcher(client);
+                    // Validate table state
+                    if (!(client instanceof InMemoryAccumuloClient)) {
+                        if (!fetcher.tableExists(tableName))
+                            throw new TableDeletedException(tableName);
```
jschmidt10: Maybe throw a `TableNotFoundException` here. We do not know if its just a typo and never existed or if it was deleted.

warehouse/core/src/main/java/datawave/mr/bulk/BulkInputFormat.java
```diff
@@ -1101,28 +1071,18 @@ public List<InputSplit> getSplits(JobContext job) throws IOException {
...
-                            if (context.getTableState(tableId) == TableState.OFFLINE)
-                                throw new TableOfflineException("Table (" + tableId.canonical() + ") is offline");
-                        }
-                        binnedRanges.clear();
-                        log.warn("Unable to locate bins for specified ranges. Retrying.");
-                        TimeUnit.MILLISECONDS.sleep(ThreadLocalRandom.current().nextInt(100, 200));
-                        tl.invalidateCache();
+                    AccumuloTableInfoFetcher fetcher = new AccumuloTableInfoFetcher(client);
+                    // Validate table state
+                    if (!(client instanceof InMemoryAccumuloClient)) {
```
jschmidt10, line 1066: Do we need to check this? It seems like the table existence would still apply.

jschmidt10: Ahh I missed that was brought over from the old code. You can disregard.

warehouse/core/pom.xml
```diff
@@ -60,6 +60,11 @@
             <groupId>gov.nsa.datawave.core</groupId>
             <artifactId>common-utils</artifactId>
         </dependency>
+        <dependency>
+            <groupId>gov.nsa.datawave.core</groupId>
+            <artifactId>datawave-core-connection-pool</artifactId>
+            <version>${project.version}</version>
```
jschmidt10: Can you add this dependency to the top level pom dependencyManagement and then that'll let us just inherit the version from the parent. Check out common-utils right above for an example.

core/connection-pool/src/main/java/datawave/core/common/connection/AccumuloTableInfoFetcher.java
jschmidt10: @SethSmucker are there any available APIs to get queued compactions? If so, can we just add it to maintain the historical behavior?

SethSmucker: Not that I could find for 2.1.* at least, but it looks like it will be available once we move to 4.0. For now I'll use the same facade pattern and we can swap out the internals once we hit 4.0, removing the facades one by one so we can test them on the high side. I'll push a draft for that soon, but if there's another way to do it that would work better I'm all ears

SethSmucker: Also, it looked like MetadataServicer might not need a class replacement, the public API versions of its usecase are fairly straight forward, so I think we can directly call them instead of having the original facade. We'll still have the one I mentioned above for the compactions, but just not the original one from the ticket. If it would be better to keep it as a centralized place (closer to what MetadataServicer was doing) I can revert it, but I'll go ahead and try out the direct method calls for now.

core/connection-pool/src/main/java/datawave/core/common/connection/AccumuloTableInfoFetcher.java
```diff
@@ -0,0 +1,102 @@
...
+    /**
+     * Get the count of queued and running major compactions across all tablet servers.
+     *
+     * @return the number of queued and running major compactions
+     * @throws AccumuloException
+     *             if a general Accumulo error occurs
+     * @throws AccumuloSecurityException
+     *             if a security error occurs
+     */
+    public int getMajorCompactionCount() throws AccumuloException, AccumuloSecurityException {
```
lbschanno: AccumuloSecurityException is never actually thrown here, and can be removed.

warehouse/core/src/main/java/datawave/mr/bulk/BulkInputFormat.java
```diff
@@ -1095,28 +1063,17 @@ public List<InputSplit> getSplits(JobContext job) throws IOException {
...
-                            if (context.getTableState(tableId) == TableState.OFFLINE)
-                                throw new TableOfflineException("Table (" + tableId.canonical() + ") is offline");
-                        }
-                        binnedRanges.clear();
-                        log.warn("Unable to locate bins for specified ranges. Retrying.");
-                        TimeUnit.MILLISECONDS.sleep(ThreadLocalRandom.current().nextInt(100, 200));
-                        tl.invalidateCache();
+                    // Validate table state using public API
+                    if (!(client instanceof InMemoryAccumuloClient)) {
+                        if (!client.tableOperations().exists(tableName))
```
lbschanno: I would prefer to see the one line if statements here with explicit braces for readability.

warehouse/core/src/main/java/datawave/mr/bulk/BulkInputFormat.java
```diff
@@ -1195,6 +1152,20 @@ private void clipRanges(Map<String,Map<KeyExtent,List<Range>>> binnedRanges) {
 
     }
 
+    private Map<String,Map<KeyExtent,List<Range>>> buildBinnedRanges(Locations locations) {
```
lbschanno: I would like to see a method-level comment describing what this function is doing/returning.

warehouse/core/src/main/java/datawave/mr/bulk/BulkInputFormat.java
```diff
@@ -1195,6 +1152,20 @@ private void clipRanges(Map<String,Map<KeyExtent,List<Range>>> binnedRanges) {
...
+        Map<String,Map<KeyExtent,List<Range>>> binnedRanges = new HashMap<>();
+        for (Entry<TabletId,List<Range>> entry : locations.groupByTablet().entrySet()) {
+            TabletId tabletId = entry.getKey();
+            String location = locations.getTabletLocation(tabletId);
+            if (location == null)
```
lbschanno: Similar request here for explicit braces for the one line if statement.

core/connection-pool/src/main/java/datawave/core/common/connection/AccumuloTableInfoFetcher.java
```diff
@@ -0,0 +1,102 @@
...
+ * <p>
+ * This class exists solely to contain Accumulo internals that have no public API equivalent. Operations that can use public APIs (e.g., table existence, online
+ * status, tablet locations) should call {@link org.apache.accumulo.core.client.admin.TableOperations} or
+ * {@link org.apache.accumulo.core.client.admin.InstanceOperations} directly — they do not belong here.
+ * <p>
+ * See the import block above for the incremental migration plan.
+ *
+ * @see <a href="https://github.com/NationalSecurityAgency/datawave/issues/2443">Issue #2443</a>
+ */
+public class AccumuloTableInfoFetcher {
```
lbschanno: I like the idea of isolating non-public API usage behind a facade, but it might make more sense for this class to follow a utility-class style implementation with static methods. There may be times when we want to isolate non-public Accumulo API usage for an operation that requires an argument that is not an AccumuloClient.

lbschanno (changes requested):

core/connection-pool/src/main/java/datawave/core/common/connection/AccumuloTableInfoFetcher.java
```diff
@@ -0,0 +1,168 @@
...
+ * <p>
+ * This class replaces direct usage of non-public Accumulo internals ({@code ClientContext}, {@code MetadataServicer}, {@code TabletLocator}, etc.) with public
+ * API equivalents where possible. Methods that still require non-public APIs are marked with {@code NON-PUBLIC API} comments.
+ *
+ * @see <a href="https://github.com/NationalSecurityAgency/datawave/issues/2443">Issue #2443</a>
+ */
+public final class AccumuloTableInfoFetcher {
+
+    private AccumuloTableInfoFetcher() {
+        // utility class
```
lbschanno: Would throw an UnsupportedOperationException here.

core/connection-pool/src/main/java/datawave/core/common/connection/AccumuloTableInfoFetcher.java
```diff
@@ -0,0 +1,168 @@
...
+
+/**
+ * Utility class that centralizes Accumulo table metadata operations.
+ * <p>
+ * This class replaces direct usage of non-public Accumulo internals ({@code ClientContext}, {@code MetadataServicer}, {@code TabletLocator}, etc.) with public
+ * API equivalents where possible. Methods that still require non-public APIs are marked with {@code NON-PUBLIC API} comments.
+ *
+ * @see <a href="https://github.com/NationalSecurityAgency/datawave/issues/2443">Issue #2443</a>
+ */
+public final class AccumuloTableInfoFetcher {
```
lbschanno: Could call this AccumuloTableUtils, or something along those lines perhaps.

lbschanno (changes requested):

PR #3450, opened 2026-03-03
Updates to annotation validation and validation test additions

drewfarris:
* Now provides separate validators for checking annotation data vs. annotation identity, so that writes of unidentified annotations can be performend and validated prior to assigning ids.
* Updates ingest tests to add segment hashes


core/annotation/src/main/java/datawave/annotation/util/v1/AnnotationValidators.java
```diff
@@ -55,27 +50,76 @@ public class AnnotationValidators {
...
             .addCheck(s -> !s.getValuesList().isEmpty(), "Segment must have at least one value")
             .addCheck(Segment::hasBoundary, "Segment must have a boundary")
             .addMemberValidator(Segment::getValuesList, segmentValueValidator)
             .addMemberValidator(s -> Collections.singletonList(s.getBoundary()), getBoundaryValidatorByType);
 
+    // ------- annotation identity validation ------ //
+
+    private static final Validator<Annotation> annotationIdentityValidator = Validator.<Annotation>create()
+            .addCheck(a -> StringUtils.isNotBlank(a.getAnnotationId()), "Annotation id must not be blank")
+            .addCheck(a -> StringUtils.isNotBlank(a.getAnalyticSourceHash()) || StringUtils.isNotBlank(a.getSource().getAnalyticHash()), "Annotation source hash must not be blank")
```
FineAndDandy, line 77: should this check that both are not blank? It also probably needs to check that a.getSource() is not null or this could produce an NPE

core/annotation/src/main/java/datawave/annotation/util/v1/AnnotationUtils.java
```diff
@@ -108,6 +137,21 @@ public static SegmentValue injectSegmentValueHash(SegmentValue segmentValue) {
...
+     * these references are used to maintain linkages between the two annotations.
+     *
+     * @param update
+     *            the annotation updating the target
+     * @param updateTargetId
+     *            the identifier of the target being updated.
+     * @return
+     */
+    public static Annotation injectUpdateReference(Annotation update, String updateTargetId) {
+        // TODO: implement me - currently a no-op
```
FineAndDandy, line 151: does this need to be done prior to merge?

core/annotation/src/main/java/datawave/annotation/data/v1/AnnotationDataAccess.java
```diff
@@ -318,52 +337,35 @@ public Optional<Annotation> addAnnotation(Annotation annotation) {
...
-                                + ") are present for document: " + shard + "/" + datatype + "/" + uid + " and annotation id: " + annotationId);
-            }
+        Optional<Annotation> targetAnnotation = getAnnotation(shard, datatype, uid, targetAnnotationId);
+        if (targetAnnotation.isEmpty()) {
+            throw new AnnotationUpdateException("Unable to find annotation to update for document: " + shard + "/" + datatype + "/" + uid
+                            + " and annotation id: " + targetAnnotationId);
+        }
 
-            // TODO: implement update logic here.
+        Annotation referenceAnnotation = AnnotationUtils.injectUpdateReference(annotation, targetAnnotationId);
```
FineAndDandy, line 360: since this is not currently implemented as-is I don't believe this code can execute successfully

FineAndDandy (changes requested):
a few questions otherwise looks good

core/annotation/src/main/java/datawave/annotation/util/v1/AnnotationValidators.java
drewfarris: Either one or the other can be blank. We don't always load the source, so protobuf may stick an 'empty' object there as a result.

Bottom line, the analytic source hash must be present in one of two places:
1. The Annotation's analyticSourceHash field in cases when the annotation source isn't loaded
2. The Annotation's source's analyticHash if the source is loaded

A null check is extraneous here because protobuf has interesting semantics. An Annotation without a source will return a non-null value for source. For example, this test will pass:
```java
Annotation a = Annotation.newBuilder().build();
assertFalse(a.hasSource());
AnnotationSource source = a.getSource();
assertNotNull(source);
```
In this case, the fields of the AnnotationSource will all be 'empty'.

Annotation construction is strictly done with the Protobuf builder pattern, so the following fails with a NPE: 
```java
Annotation a = Annotation.newBuilder().setSource((AnnotationSource) null).build();
```

So it's not possible for an Annotation to have a null source.

FineAndDandy: thanks for the detailed explanation

core/annotation/src/test/java/datawave/annotation/data/v1/AnnotationDataAccessTest.java
```diff
@@ -287,27 +292,123 @@ public void testGetAnnotationSource() {
...
+
+        // This should not throw an exception (delete handles empty iterator gracefully)
+        // Note: if delete returns null mutation, writer.addMutation is called with null which throws
+        try {
+            dao.delete(shard, dataType, uid, annotationId);
+            // If we get here, the delete handled empty results correctly
+        } catch (IllegalArgumentException e) {
+            // This is expected if mutationAdapter returns null and writer.addMutation is called with null
+            // The implementation should be fixed to check for null mutations
+            assertTrue(e.getMessage().contains("m is null"));
```
FineAndDandy: which case are you expecting here? The silent pass through or the catch on the IllegalArgumentException?

drewfarris: Thanks for catching this. I meant to modify the logic to throw an exception when the object to delete is not found - matching the semantics for an update operation.

FineAndDandy (dismissed):

core/annotation/src/main/java/datawave/annotation/data/v1/AnnotationDataAccess.java
```diff
@@ -516,9 +522,13 @@ protected void checkAnnotationSourceForConflicts(AnnotationSource annotationSour
 
         Optional<AnnotationSource> conflicting = getAnnotationSource(annotationSource.getAnalyticHash());
         if (conflicting.isPresent()) {
-            throw new AnnotationWriteException("Cannot add annotation because an annotation with the same id already exists. New annotation: "
-                            + annotationSource + ", Existing annotation: " + conflicting.get());
+            if (blockAnnotationSourceOverwrites) {
+                throw new AnnotationWriteException("Cannot add annotation source because an annotation source with the same id already exists.");
+            } else {
+                log.debug("Allowing annotation source overwrite: {}", annotationSourceString(annotationSource));
```
avgAGB, line 524: maybe irrelevant, but possibly wrap this in a logger.isDebugEnabled check?

core/annotation/src/main/java/datawave/annotation/data/v1/AnnotationDataAccess.java
```diff
@@ -594,11 +604,13 @@ protected void checkAnnotationForConflicts(Annotation annotation) {
...
         }
-
         Optional<Annotation> conflicting = getAnnotation(annotation.getShard(), annotation.getDataType(), annotation.getUid(), annotation.getAnnotationId());
         if (conflicting.isPresent()) {
-            throw new AnnotationWriteException("Cannot add annotation because an annotation with the same id already exists. New annotation: " + annotation
-                            + ", Existing annotation: " + conflicting.get());
+            if (blockAnnotationOverwrites) {
+                throw new AnnotationWriteException("Cannot add annotation because an annotation with the same id already exists.");
+            } else {
+                log.debug("Allowing annotation overwrite: {}", annotationIdContext(conflicting.get()));
```
avgAGB, line 608: similar note here if applicable

core/annotation/src/main/java/datawave/annotation/data/v1/AnnotationDataAccess.java
drewfarris: This is good feedback because this line is generating a string and passing the result in as a parameter. In the standard case where we're just passing in a string that's already created it wouldn't matter as much, but in this case we can prevent the extra string generation for a log message that's not going to be written. Same for the comment regarding line 610 below.

core/annotation/src/main/java/datawave/annotation/util/v1/AnnotationValidators.java
```diff
@@ -55,27 +50,76 @@ public class AnnotationValidators {
...
+            .addCheck(s -> StringUtils.isNotBlank(s.getSegmentHash()), "Segment hash (identity) must not be blank")
+            .addMemberValidator(Segment::getValuesList, segmentValueIdentityValidator);
+
+    // ------ segment data validation ------ //
+
+    private static final Validator<SegmentValue> segmentValueValidator = Validator.<SegmentValue>create()
+            .addCheck(v -> StringUtils.isNotBlank(v.getValue()), "Segment value must not be blank");
+
     private static final Validator<Segment> segmentValidator = Validator.<Segment>create()
             .addCheck(s -> !s.getValuesList().isEmpty(), "Segment must have at least one value")
```
tharke13: I think we will need to remove the constraint of having at least one segment value to indicate in an annotation update that the boundary should be removed/masked.

drewfarris: We'll want to conditionally override this validation in certain write cases, but for the primary case we don't want blank segment values. Let me think about this a bit further, but will likely add a separate method for validating removal or masking.

avgAGB (approved):

tharke13 (approved):
Looks good!

FineAndDandy (approved):

PR #3452, opened 2026-03-05
Integrate ScanMonitor into UnfieldedLiteralIndexLookup

apmoriarty:
- integrate ScanMonitor into UnfieldedLiteralIndexLookup
- rename FieldExpansionIndexLookup to UnfieldedLiteralIndexLookup so it matches naming convention of other lookups
- standardize exception handling per review comment from https://github.com/NationalSecurityAgency/datawave/pull/3446

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/jexl/lookups/FieldedRegexIndexLookup.java
+++ b/warehouse/query-core/src/main/java/datawave/query/jexl/lookups/FieldedRegexIndexLookup.java
@@ -93,8 +93,7 @@ public class FieldedRegexIndexLookup extends BaseRegexIndexLookup {
                 }
 
             } catch (Exception e) {
-                log.error("Unexpected exception seen", e);
-                handleException();
+                handleException(e);
             }
         };
     }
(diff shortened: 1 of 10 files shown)


warehouse/query-core/src/main/java/datawave/query/jexl/lookups/FieldedRegexIndexLookup.java
```diff
@@ -124,10 +123,14 @@ public IndexLookupMap lookup() {
...
      * The entry is then marked as threshold exceeded.
+     *
+     * @param e
+     *            the exception
      */
     @Override
-    protected void handleException() {
-        log.debug("marking regex as exceeded");
+    protected void handleException(Exception e) {
+        log.warn("FieldedRegexIndexLookup saw exception: {}", e.getMessage());
```
dtspence, line 132: Comment applies to multiple locations - do we have anywhere that the full exception is logged? If not, should we try to capture (for the stack trace) or are these exceptions not likely to need that detail?

apmoriarty: I don't think the full stack trace is necessary here. If there's wider disagreement I'm happy to add that in.

dtspence (approved):

avgAGB (approved):

PR #3453, opened 2026-03-06
Code changes to support main datawave using accumulo 2.1.4 instead of 2.1.4-5792fed3-bulkv2

billoley:
Accumulo changes from 2.1.4-5792fed3-bulkv2 (pre-2.1.4 release) to 2.1.4 reuire datawave code changes.

accumulo class ColumnFamily changed fields columnFamily, columnQualifier, etc from public to private.  Wes houldn't have been using them anyway.
IteratorConfigUtil.loadIterators() declares throws ReflectiveOperationException even though that exception is caught internally and is thrown as a RuntimeException.  Also, this declaration is removed in accumulo 3.0.0.  

The microservices can be upgraded to accumulo 2.1.4 after there is a version of datawave that won't break at runtime while using accumulo 2.1.4. (illegal access)

### DIFF
--- a/warehouse/assemble/datawave/pom.xml
+++ b/warehouse/assemble/datawave/pom.xml
@@ -13,7 +13,7 @@
         <!-- jkube kubernetes-maven-plugin only has access to project properties so we redefine them to properly render the Dockerfile-->
         <project.version>${project.version}</project.version>
         <!--   used for image ghcr.io/nationalsecurityagency/datawave-stack-accumulo:${version.accumulo}     -->
-        <version.accumulo>2.1.3</version.accumulo>
+        <version.accumulo>2.1.4</version.accumulo>
     </properties>
     <dependencies>
         <!-- To ensure a consistent version is loaded -->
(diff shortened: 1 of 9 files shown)


warehouse/assemble/datawave/pom.xml
```diff
@@ -13,7 +13,7 @@
         <!-- jkube kubernetes-maven-plugin only has access to project properties so we redefine them to properly render the Dockerfile-->
         <project.version>${project.version}</project.version>
         <!--   used for image ghcr.io/nationalsecurityagency/datawave-stack-accumulo:${version.accumulo}     -->
-        <version.accumulo>2.1.3</version.accumulo>
+        <version.accumulo>2.1.4</version.accumulo>
```
jalphonso: can remove and inherit from top level parent

jalphonso (approved):

avgAGB (approved):

apmoriarty (approved):

Issue #3455, opened 2026-03-06
UniqueTransform failing with NPE when it gets a WaitWindowOverrun document


billoley:
After the SerialIterator reports a WaitWindowOverrun (yield time exceeded) with collectTimingMetrics=true which sends a document containing key WAIT_WINDOW_OVERRUN.  This document will not contain DOCKEY_FIELD_NAME = "RECORD_ID" which is what UniqueTransform is looking for in getTimestamp / getRootDocKeyAttr.

The GroupingIterator also gets documents that the SerialIterator produces, so we should also make sure that it is passing the WAIT_WINDOW_OVERRUN document through and not causing an exception.

Ideally, we can add a test to the QueryIteratorIT so that the WaitWindowQueryIteratorSerialIT and WaitWindowQueryIteratorPipelineIT tests can run it.

```
private static long getTimestamp(Document doc) {
        return getRootDocKeyAttr(doc).getTimestamp();    // NPE here
}

```

```
    public static Attribute getRootDocKeyAttr(Document doc) {
        Attribute<?> attr = doc.get(Document.DOCKEY_FIELD_NAME);
   .....
```

```
java.lang.NullPointerException: null
        at datawave.query.transformer.UniqueTransform.getTimestamp(UniqueTransform.java:568)
        at datawave.query.transformer.UniqueTransform$Builder.lambda$new$0(UniqueTransform.java:441)
        at datawave.query.util.sortedmap.FileSortedMap.put(FileSortedMap.java:490)
        at datawave.query.util.sortedmap.BufferedFileBackedSortedMap.put(BufferedFileBackedSortedMap.java:385)
        at datawave.query.transformer.UniqueTransform.apply(UniqueTransform.java:145)
        at datawave.query.transformer.UniqueTransform$UniqueTransformIterator.getNext(UniqueTransform.java:409)
        at datawave.query.transformer.UniqueTransform$UniqueTransformIterator.hasNext(UniqueTransform.java:389)
.....
        at datawave.query.jexl.functions.KeyAdjudicator.hasNext(KeyAdjudicator:60)
        at datawave.query.iterator.profile.FinalDocumentTrackingIterator.hasNext(FinalDocumentTrackingIterator:85)
```

PR #3456, opened 2026-03-06
UniqueTransform should not try to transform a WAIT_WINDOW_OVERRUN document

billoley:
see #3455

UniqueTransformIterator should not try to transform a WAIT_WINDOW_OVERRUN document


billoley:
The new test was initially passing in QueryIteratorIT, WaitWindowQueryIteratorSerialIT, and WaitWindowQueryIteratorPipelineIT but failing in the TLD and Ancestor versions of the same test.  

The new code in UniformTransform fixes the NPE in the case of a WaitWindowOverrun but there was still an issue with the UniformTransform not correctly maintaining the map and returnSet during repeated yields.  The TLD and Ancestor versions increase the frequency of random yields and were ending up with more than one unique document when they should only end up with the last one.

When a WaitWindowOverrun document is seen in UniqueTransform, we should persist the map rather than relying on the persist thresholds.  The other issue was that when collectTimingDetails was false (TLDQueryIteratorIT), the underlying iterator was returning null (due to yield) and the UniqueTransformIterator was calling flush which persisted the map and the returnSet.   Both get restored on a subsequent seek, but that meant that in setupIterator, the returnSet was not empty to start with.  This PR fixes this by clearing the returnSet first.

Please verify that my logic is correct.

ivakegg:
I believe your assessment of the issues is correct here Bill.  I will review this and integration test today.

billoley:
I'm concerned about clearing the returnSet.  Why pesist it if we clear what is read-in before populating it again.  Another approach that I thought of is to avoid populating returnSet if we know that we are yielding.  FYI some of the QueryIterator tests have collectTimingDetails=true and some have collectTimingDetails=false.

ivakegg:
So persisting the map simply adds another file for the sort afterwards.  There is no harm in that.  Clearing it does not make sense at all to me.  The return set should only start being flushed out after all of the underlying keys have been aggregated.

billoley:
The UniqueTransformIterator.getNext()

```
        private Map.Entry<Key,Document> getNext() {
            Map.Entry<Key,Document> o = null;
            while (o == null && iterator.hasNext()) {
                o = apply(iterator.next());
            }
            // see if there are any results cached by the transform
            if (o == null) {
                o = flush();
            }
            return o;
        }

```
in the collectTimingDetails=false case when a yield is happening in iterator (SerialIterator, etc), I saw hasNext return false, o is null, and yield flag is set.  That is when flush is called, which among other things, populates returnSet with the current batch that is in set.  This was happening in the TLD and Ancestor QueryIteratorIT and therefore the returnSet ended up with more than one unique result when it should have only had the most recent.  That's why I put the clear in, but we could also avoid putting the values from set into returnSet if a yield is happening.  i.e. wait until we are really done with the underlying iterator.  Calling flush in this case DOES ensure that set is persisted, which we still want to do -- just not populate resultSet.

ivakegg:
I think I like it better to simply not populate the returnSet when we are yielding.  That is just extra work that is not needed.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -133,6 +136,13 @@ public Entry<Key,Document> apply(@Nullable Entry<Key,Document> keyDocumentEntry)
...
             }
+            Document document = keyDocumentEntry.getValue();
+            if (document != null && document.containsKey(WAIT_WINDOW_OVERRUN)) {
+                if (log.isDebugEnabled()) {
+                    log.debug("Got a WAIT_WINDOW_OVERRUN key in unique transform: " + keyDocumentEntry.getKey());
+                }
+                return keyDocumentEntry;
+            }
 
             if (keyDocumentEntry.getValue().isIntermediateResult()) {
```
apmoriarty, line 147: Final Document, Intermediate result, and now a yielding key inside a document...seems like we're close to needing a way to simply return special documents.

ivakegg: Not for this patch but that would be a great refactoring

apmoriarty (approved):

apmoriarty (approved):

ivakegg (approved):

ivakegg:
Merged manually

ivakegg:
Merged manually

PR #3458, opened 2026-03-12
Refactor Cost class to avoid Long overflows and underflow

apmoriarty:
- Add tests for Cost class
- Fix overflow and underflow errors by enforcing boundary checks
- Enforce access to Cost constructor via static entry points

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/planner/pushdown/Cost.java
+++ b/warehouse/query-core/src/main/java/datawave/query/planner/pushdown/Cost.java
@@ -237,15 +237,8 @@ public class Cost implements Comparable<Cost> {
         return result;
     }
 
-    /*
-     * (non-Javadoc)
-     *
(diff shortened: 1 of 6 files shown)


warehouse/query-core/src/main/java/datawave/query/planner/pushdown/Cost.java
```diff
@@ -1,73 +1,216 @@
...
+            regexCost = weightedCost;
+        }
     }
 
+    /**
+     * Get the regex cost
+     *
+     * @return the regex cost
+     */
     public long getERCost() {
```
avgAGB: would it complicate the refactor if we changed the name of the getter to getRegexCost or something instead?

apmoriarty: Not at all, happy to do that work here and now

avgAGB (approved):

hlgp (approved):

PR #3460, opened 2026-03-13
fixes #3459: added disableIteratorMostRecentUniqueFields mechanism

### LINKED ISSUE #3459: Create a feature flag for most-recent-unique
We have been seeing some issues with the server-side most-recent-unique functionality combined with yielding.  Because of this we need a feature flag to be able to turn off the tserver side unique transform (in the case of most-recent-uniq).  This will be a flag on the ShardQueryLogic that gets passed through to the DefaultQueryPlanner.

ivakegg:
(true by default pending some addigiona yielding work) to disable
    using the tserver side most-recent-unique functionality and rely
    completely on the webserver side filtering

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/util/sortedmap/BufferedFileBackedSortedMap.java
+++ b/warehouse/query-core/src/main/java/datawave/query/util/sortedmap/BufferedFileBackedSortedMap.java
@@ -305,13 +305,13 @@ public class BufferedFileBackedSortedMap<K,V> implements SortedMap<K,V>, Rewrita
 
             // compact it
             if (log.isDebugEnabled()) {
-                log.debug("Starting compaction for " + mapToCompact);
+                log.debug("Starting compaction for " + mapsPerCompaction + " maps");
             }
             long start = System.currentTimeMillis();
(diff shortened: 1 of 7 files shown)


apmoriarty (dismissed):

apmoriarty (approved):

avgAGB (approved):

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/BufferedFileBackedSortedMap.java
```diff
@@ -305,13 +305,13 @@ public void compact(int maxFiles) throws IOException {
 
             // compact it
             if (log.isDebugEnabled()) {
-                log.debug("Starting compaction for " + mapToCompact);
```
avgAGB, line 308: is there a use case where we would still want to see the contents of the map being compacted, even if it was locked behind a trace level block? Just wondering.

ivakegg: The output of that statement is ridiculously long and in general we do not output the results in the logs so I am going to stick with this change.  Thank you.

PR #3469, opened 2026-03-23
Replace Markings Map with POJO

jalphonso:
update accumulo-utils to use a POJO for storing markings and then updated the rest of the codebase that referenced it.  
replaced some of the ColumnVisibility usage with AccessExpression to improve performance.


core/base-rest-responses/src/main/java/datawave/marking/ColumnVisibilitySecurityMarking.java
```diff
@@ -13,14 +13,16 @@
...
+import org.apache.accumulo.access.impl.AccessExpressionImpl;
 
 @XmlAccessorType(XmlAccessType.NONE)
 @XmlAccessorOrder(XmlAccessOrder.ALPHABETICAL)
 public class ColumnVisibilitySecurityMarking implements SecurityMarking {
 
     public static final String VISIBILITY_MARKING = "columnVisibility";
 
-    private static final Access ACCESS = Access.builder().build();
+    private static final Access ACCESS = ParseExamples.ACCESS;
```
ddanielr: ParseExamples.ACCESS is just `Access.builder().build();` 

Can we make a CONSTANTS class and just create the ACCESS and the AccessExpressionImpl classes there? That would reduce the deps on non-public code. 

AccessExpressionImpl is like 98 lines, so it would be pretty trivial to just internalize.

core/base-rest-responses/src/main/java/datawave/marking/ColumnVisibilitySecurityMarking.java
```diff
@@ -47,8 +49,16 @@ public void setColumnVisibility(String columnVisibility) {
     }
 
     @Override
-    public ColumnVisibility toColumnVisibility() {
-        return new ColumnVisibility(columnVisibility);
+    public AccessExpression toAccessExpression() {
+        if (columnVisibility == null || columnVisibility.isEmpty()) {
+            return AccessExpressionImpl.EMPTY;
```
keith-turner: Would be best to avoid classes w/ Impl in the name as these are not public API and could change at any time.  For this case passing any empty string should be efficient and avoid any object allocations.

```suggestion
            return ACCESS.newExpression("");
```

core/cached-results/src/main/java/datawave/core/query/cachedresults/CacheableQueryRowReader.java
```diff
@@ -1,5 +1,7 @@
 package datawave.core.query.cachedresults;
 
+import static org.apache.accumulo.access.examples.ParseExamples.ACCESS;
```
keith-turner: The accumulo access examples are not part of the accumulo access public API.  Would be good to avoid depending on them.  Could create an ACCESS constant within datawave.

core/cached-results/src/main/java/datawave/core/query/cachedresults/CacheableQueryRowImpl.java
```diff
@@ -7,26 +7,28 @@
...
+import datawave.marking.Markings;
 import datawave.webservice.query.cachedresults.CacheableQueryRow;
 import datawave.webservice.query.data.ObjectSizeOf;
 import datawave.webservice.query.util.TypedValue;
 
 public class CacheableQueryRowImpl extends CacheableQueryRow implements ObjectSizeOf {
 
-    private static Logger log = LoggerFactory.getLogger(CacheableQueryRowImpl.class);
+    private static final Logger LOGGER = LoggerFactory.getLogger(CacheableQueryRowImpl.class);
+    private static final Markings<?> EMPTY_MARKINGS = AccessExpressionMarkings.builder().accessExpression(AccessExpressionImpl.EMPTY).build();
```
keith-turner:
```suggestion
    private static final Markings<?> EMPTY_MARKINGS = AccessExpressionMarkings.builder().accessExpression(ACCESS.newExpression("")).build();
```

core/common-util/src/main/java/datawave/core/common/util/AccessExpressionUtil.java
```diff
@@ -0,0 +1,195 @@
...
+ * Utility for bridging between Accumulo's {@link ColumnVisibility} (from accumulo-core) and {@link AccessExpression} (from accumulo-access).
+ * <p>
+ * Many modules still require {@link ColumnVisibility} for the native Accumulo Key/Mutation API, while the marking functions layer has migrated to
+ * {@link AccessExpression}. This utility provides zero-copy-where-possible conversions between the two.
+ */
+public final class AccessExpressionUtil {
+
+    private static final Access ACCESS = ParseExamples.ACCESS;
+
+    private static final AccessExpression EMPTY_EXPRESSION = AccessExpressionImpl.EMPTY;
```
keith-turner:
```suggestion
    private static final AccessExpression EMPTY_EXPRESSION = ACCESS.newExpression("");
```

core/utils/accumulo-utils/src/main/java/datawave/marking/AccessExpressionUtil.java
```diff
@@ -0,0 +1,320 @@
...
+     * @param cv
+     *            the column visibility
+     * @return the equivalent access expression
+     */
+    public static AccessExpression toAccessExpression(ColumnVisibility cv) {
+        byte[] expr = cv.getExpression();
+        if (expr.length == 0) {
+            return EMPTY_EXPRESSION;
+        }
+        return ACCESS.newExpression(new String(expr, UTF_8));
```
jsduncan2, line 59: Could we delegate to `toAccessExpression(byte[] visibilityBytes)`?

```suggestion
        return toAccessExpression(expr);
```

core/utils/accumulo-utils/src/main/java/datawave/marking/AccessExpressionUtil.java
```diff
@@ -0,0 +1,320 @@
...
+     * @param markings
+     *            the markings to combine
+     * @return the combined markings
+     * @throws MarkingFunctions.Exception
+     *             if the markings cannot be combined
+     */
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    public static Markings<?> combine(MarkingFunctions<?> markingFunctions, Collection<? extends Markings<?>> markings) throws MarkingFunctions.Exception {
+        return ((MarkingFunctions) markingFunctions).combine(markings);
+    }
```
jsduncan2, line 148: I think we can be a little more specific with generics here:

```suggestion
    public static <T extends Markings<?>, U extends MarkingFunctions<T>> T combine(U markingFunctions, Collection<T> markings) throws MarkingFunctions.Exception {
        return markingFunctions.combine(markings);
    }
```

core/utils/accumulo-utils/src/main/java/datawave/marking/AccessExpressionUtil.java
```diff
@@ -0,0 +1,320 @@
...
+    public static Markings<?> mergeMarkings(MarkingFunctions<?> markingFunctions, Markings<?> markings1, Markings<?> markings2) {
+        if (markings2 != null) {
+            if (markings1 == null) {
+                markings1 = markings2;
+            } else {
+                try {
+                    if (markings1 instanceof AccessExpressionMarkings && markings2 instanceof AccessExpressionMarkings) {
+                        AccessExpressionMarkings aem1 = (AccessExpressionMarkings) markings1;
+                        AccessExpressionMarkings aem2 = (AccessExpressionMarkings) markings2;
+                        return combine(markingFunctions, List.of(aem1, aem2));
```
jsduncan2: If the above generic suggestion works, i think you'll need to modify this slightly:

```suggestion
                    if (markingFunctions instanceof MarkingFunctions.Default && markings1 instanceof AccessExpressionMarkings && markings2 instanceof AccessExpressionMarkings) {
                        MarkingFunctions<AccessExpressionMarkings> defaultMarkingFunctions = (MarkingFunctions.Default) markingFunctions;
                        AccessExpressionMarkings aem1 = (AccessExpressionMarkings) markings1;
                        AccessExpressionMarkings aem2 = (AccessExpressionMarkings) markings2;
                        return combine(defaultMarkingFunctions, List.of(aem1, aem2));
```

core/utils/accumulo-utils/src/main/java/datawave/marking/AccessExpressionUtil.java
```diff
@@ -0,0 +1,320 @@
...
+import org.apache.accumulo.access.ParsedAccessExpression;
+import org.apache.accumulo.core.security.ColumnVisibility;
+
+/**
+ * Utility for bridging between Accumulo's {@link ColumnVisibility} (from accumulo-core) and {@link AccessExpression} (from accumulo-access).
+ * <p>
+ * Many modules still require {@link ColumnVisibility} for the native Accumulo Key/Mutation API, while the marking functions layer has migrated to
+ * {@link AccessExpression}. This utility provides zero-copy-where-possible conversions between the two.
+ */
+public final class AccessExpressionUtil {
```
jsduncan2, line 25: This class could use some unit tests. Let me know if you'd like some help making them.

core/utils/accumulo-utils/src/main/java/datawave/marking/Markings.java
```diff
@@ -0,0 +1,13 @@
+package datawave.marking;
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS)
+public interface Markings<T> {
```
jsduncan2, line 9: Can we restrict T here at all?

core/base-rest-responses/src/main/java/datawave/webservice/query/result/event/HasMarkings.java
```diff
@@ -1,10 +1,10 @@
...
+import datawave.marking.Markings;
 
 public interface HasMarkings {
 
-    void setMarkings(Map<String,String> markings);
+    void setMarkings(Markings<?> markings);
 
-    Map<String,String> getMarkings();
+    Markings<?> getMarkings();
 }
```
jsduncan2, line 10: I'm guessing there's some issue you'll eventually run into, but here's a naive suggestion (which will require a lot of associated changes, some of which may be impossible):
```java
public interface HasMarkings <T> {

    void setMarkings(Markings<T> markings);

    Markings<T> getMarkings();
}
```

core/utils/accumulo-utils/src/main/java/datawave/marking/MarkingFunctions.java
```diff
@@ -72,106 +63,71 @@ public Exception(Throwable cause) {
...
-        @Override
-        public ColumnVisibility combine(Collection<ColumnVisibility> expressions) {
+        private static final AccessExpressionMarkings EMPTY_MARKINGS = AccessExpressionMarkings.builder().accessExpression(ACCESS.newExpression("")).build();
 
-            // filter out any empty expressions, then flatten each one (to de-dupe) and concatenate with '&'
-            // flatten the final combined ColumnVisibility and use that to make the ColumnVisibility to return
-            return new ColumnVisibility(new ColumnVisibility(expressions.stream().map(ColumnVisibility::flatten).filter(b -> b.length > 0)
-                            .map(b -> "(" + new String(b, UTF_8) + ")").collect(Collectors.joining("&")).getBytes(UTF_8)).flatten());
+        @Override
+        public ColumnVisibility combineVisibilities(Collection<ColumnVisibility> visibilities) {
```
jschmidt10, line 71: Just curious, why maintain this method vs deleting?

jalphonso: There is extensive usage in the HS code

core/utils/accumulo-utils/src/main/java/datawave/marking/Markings.java
jalphonso: lets see if we can address this down the road on another MR

core/utils/accumulo-utils/src/main/java/datawave/marking/AccessExpressionUtil.java
jalphonso: done

core/utils/accumulo-utils/src/main/java/datawave/marking/AccessExpressionUtil.java
jalphonso: lets see if we can address this down the road on another MR

core/base-rest-responses/src/main/java/datawave/webservice/query/result/event/HasMarkings.java
jalphonso: lets see if we can address this down the road on another MR

core/utils/accumulo-utils/src/main/java/datawave/marking/AccessExpressionUtil.java
jalphonso: lets see if we can address this down the road on another MR

core/utils/accumulo-utils/src/main/java/datawave/marking/AccessExpressionUtil.java
jalphonso: done

warehouse/query-core/src/main/java/datawave/query/transformer/FacetedTransformer.java
```diff
@@ -177,18 +184,27 @@ public BaseQueryResponse createResponse(List<Object> resultList) {
...
+        } catch (Exception e) {
+            log.error("Could not create combined columnVisibility for the count", e);
+            return null;
+        }
 
+        try {
+            response.setMarkings(markingFunctions.translateFromColumnVisibility(cv));
         } catch (MarkingFunctions.Exception e) {
-            log.warn(e);
-            // original ignored these exceptions
```
jalphonso, line 191: should these exceptions be ignored?

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/tokenize/ExtendedContentIngestHelper.java
```diff
@@ -170,89 +169,6 @@ public void addMetadataFromId(RawRecordContainer event, String id) {
...
-
-    protected void addSecurityMetadataFromParms(CharSequence key, Object value, RawRecordContainer event) {
-        // If fieldName is a security marking field (as configured by EVENT_SECURITY_MARKING_FIELD_NAMES),
-        // then put the marking value into this.securityMarkings, where 'key' maps to the domain for the marking
-        // (as configured by EVENT_SECURITY_MARKING_FIELD_DOMAINS)
-        if (!StringUtils.isEmpty(key.toString()) && !StringUtils.isEmpty(value.toString())) {
-            event.addSecurityMarking(this.helper.getSecurityMarkingFieldDomainMap().get(key.toString()), value.toString());
-        }
-    }
-
```
jalphonso, line 255: I don't think this is used anywhere but please doublecheck. There are other references to EVENT_SECURITY_MARKING_FIELD_NAMES and EVENT_SECURITY_MARKING_FIELD_DOMAINS that can be cleaned up once confirmed

web-services/client/src/main/java/datawave/webservice/result/DefaultEdgeQueryResponse.java
```diff
@@ -53,12 +53,13 @@ public String getSecurityMarkings() {
         return securityMarkings;
     }
 
-    public void setMarkings(Map<String,String> markings) {
+    @Override
+    public void setMarkings(Markings<?> markings) {
         this.markings = markings;
-        this.setSecurityMarkings(markings.get("security"));
```
jalphonso, line 58: what is the key "security" used for?

core/cached-results/src/main/java/datawave/core/query/cachedresults/CacheableQueryRowImpl.java
```diff
@@ -289,17 +276,13 @@ public void setColFam(String colFam) {
...
-    public void setMarkings(Map<String,String> markings) {
+    public void setMarkings(Markings<?> markings) {
         // validate the markings
-        try {
-            markingFunctions.translateToColumnVisibility(markings);
-            this.markings = markings;
-        } catch (MarkingFunctions.Exception e) {
-            log.error("Invalid markings {}", markings, e);
-        }
+        markings.toAccessExpression();
```
jschmidt10: This call looks a little strange here - is this needed? If its doing some kind of internal init, wondering if we can make it more obvious.

core/utils/accumulo-utils/src/main/java/datawave/marking/AccessExpressionMarkings.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    @JsonProperty(COLUMN_VISIBILITY_KEY)
+    @XmlElement(name = COLUMN_VISIBILITY_KEY)
+    public String getColumnVisibilityString() {
+        return accessExpression != null ? accessExpression.getExpression() : "";
+    }
+
+    public void setColumnVisibilityString(String expr) {
+        this.accessExpression = ACCESS.newExpression(expr != null ? expr : "");
```
jschmidt10: Nit: The getters all handle when accessExpression is null but here we specifically check for a null param and then default it to empty String. I think the "undefined" state should just be one or the other - either null or empty String.

core/utils/accumulo-utils/src/main/java/datawave/marking/AccessExpressionMarkings.java
```diff
@@ -0,0 +1,174 @@
...
+
+    public static AccessExpressionMarkings createMarkings(ColumnVisibility cv) {
+        if (null == cv) {
+            return null;
+        }
+        return create(new String(cv.getExpression(), StandardCharsets.UTF_8));
+    }
+
+    public static AccessExpressionMarkings createMarkings(AccessExpression ae) {
+        if (null == ae) {
```
jschmidt10, line 89: There's a slight API inconsistency here. If I call `createMarkings(null)`, I get null back but if I call `create(null)` I get an empty AccessExpressionMarkings back.

core/modification/src/main/java/datawave/modification/MutableMetadataHandler.java
```diff
@@ -310,6 +312,24 @@ public void process(AccumuloClient client, ModificationRequestBase request, Map<
...
+        if (rawMarkings == null || rawMarkings.isEmpty()) {
+            return null;
+        }
+        String vis = rawMarkings.getOrDefault("columnVisibility", "");
+        if (vis.isEmpty()) {
+            return null;
+        }
+        try {
+            return markingFunctions.translateFromColumnVisibility(new ColumnVisibility(vis));
+        } catch (MarkingFunctions.Exception e) {
```
jschmidt10, line 328: If we're always converting MarkingsFunctionsExceptions into Runtimes, should we just make it extend Runtime?

core/cached-results/src/main/java/datawave/core/query/cachedresults/CacheableQueryRowImpl.java
jalphonso: agreed, but similar check preexisted. there is a also a preexisting comment above it that says "validate the markings". all the markingFunctions.translateToColumnVisibility(markings) got translated to markings.toColumnVisibility() or markings.toAccessExpression() as appropriate with neither throwing an exception

jalphonso: wait I might have talked myself into removing it since there's no exception being thrown

core/modification/src/main/java/datawave/modification/MutableMetadataHandler.java
jalphonso: its inconsistent for sure but I kept the same behavior as before for each location

core/utils/accumulo-utils/src/main/java/datawave/marking/AccessExpressionMarkings.java
jalphonso: thanks, I'll fix

core/utils/accumulo-utils/src/main/java/datawave/marking/AccessExpressionMarkings.java
jalphonso: thanks, I'll fix

core/base-rest-responses/src/main/java/datawave/marking/ColumnVisibilitySecurityMarking.java
```diff
@@ -31,7 +30,7 @@ public void validate(Map<String,List<String>> parameters) throws IllegalArgument
         if (null == values) {
             throw new IllegalArgumentException("Required parameter " + VISIBILITY_MARKING + " not found");
         }
-        if (values.isEmpty() || values.size() > 1) {
+        if (values.size() != 1) {
```
FineAndDandy, line 33: this will pass through with 0

jalphonso: This was an IntelliJ suggestion. `0 || > 1` simplified to `!= 1`

core/cached-results/src/main/java/datawave/core/query/cachedresults/CacheableQueryRowImpl.java
```diff
@@ -35,51 +34,50 @@ public class CacheableQueryRowImpl extends CacheableQueryRow implements ObjectSi
...
 
-    public void addColumn(String columnName, String columnValueString, Map<String,String> markings, String columnVisibility, Long timestamp) {
+    public void addColumn(String columnName, String columnValueString, Markings<?> markings, String columnVisibility, Long timestamp) {
         addColumn(columnName, new TypedValue(columnValueString), markings, columnVisibility, timestamp);
     }
 
-    public void addColumn(String columnName, TypedValue columnTypedValue, Map<String,String> markings, String columnVisibility, Long timestamp) {
+    public void addColumn(String columnName, TypedValue columnTypedValue, Markings<?> markings, String columnVisibility, Long timestamp) {
 
         columnName = columnName.replaceAll(" ", "_");
```
drewfarris: I know this isn't part of your change set, but consider the fact that for single character strings, the following is generally more performant because the character to replace isn't treated as a regular expression. Also, generally I prefer to assign it to a new immutable variable instead of reusing a variable name, the impact to the stack is minimal.
```suggestion
        final String cleanColumnName = columnName.replace(' ', '_');
```
There's a second `replaceAll` on single characters later in this method as well

core/cached-results/src/main/java/datawave/core/query/cachedresults/CacheableQueryRowImpl.java
```diff
@@ -35,51 +34,50 @@ public class CacheableQueryRowImpl extends CacheableQueryRow implements ObjectSi
...
+                            columnVisibilities.add(markings.toColumnVisibility());
+                            ColumnVisibility combinedVisibility = markingFunctions.combineVisibilities(columnVisibilities);
+
+                            // use combined marking as new markings
+                            this.markings = markingFunctions.translateFromColumnVisibility(combinedVisibility);
+                        } catch (MarkingFunctions.Exception e) {
+                            LOGGER.error("Invalid markings {} skipping column {} = {}", markings, columnName, columnTypedValue, e);
+                            return;
+                        }
                     }
```
drewfarris: Perhaps something like the following instead?
```java
try {
    this.markings = markingFunctions.combine(this.markings, markings);
}
catch (MarkingFunctions.Exception e) {
    LOGGER.error("Invalid markings {} skipping column {} = {}", markings, cleanColumnName, columnTypedValue, e);
    return;
}
```

core/modification/src/test/java/datawave/modification/MutableMetadataHandlerTestSupport.java
```diff
@@ -126,7 +126,7 @@ DefaultModificationRequest newMarkingsInsertRequest(String fieldName, String fie
         request.setFieldName(fieldName);
         request.setFieldValue(fieldValue);
         Map<String,String> markings = new HashMap<>();
-        markings.put(MarkingFunctions.Default.COLUMN_VISIBILITY, columnVisibility);
+        markings.put("columnVisibility", columnVisibility);
```
drewfarris, line 129:
```suggestion
        markings.put(AccessExpressionMarkings.COLUMN_VISIBILITY_KEY, columnVisibility);
```

core/utils/accumulo-utils/src/main/java/datawave/marking/MarkingFunctions.java
```diff
@@ -72,106 +65,101 @@ public Exception(Throwable cause) {
...
+                        AccessExpressionMarkings aem1 = (AccessExpressionMarkings) markings1;
+                        AccessExpressionMarkings aem2 = (AccessExpressionMarkings) markings2;
+                        return combine(List.of(aem1, aem2));
+                    } else {
+                        throw new RuntimeException(
+                                        String.format("Unknown markings class %s or %s", markings1.getClass().getName(), markings2.getClass().getName()));
+                    }
+                }
+            }
+            return markings1 == null ? null : (AccessExpressionMarkings) markings1;
```
drewfarris: Consider something a little less complex? Still, not a fan of the potential `ClassCastExceptions` here, so this suggestion might need more thought if we want to avoid or handle those more cleanly.
```suggestion
public AccessExpressionMarkings combine(Markings<?> markings1, Markings<?> markings2) {
            if (markings2 == null) {
                return (AccessExpressionMarkings) markings1;
            }

            if (markings1 == null) {
                return (AccessExpressionMarkings) markings2;
            }
            
            if (markings1 instanceof AccessExpressionMarkings && markings2 instanceof AccessExpressionMarkings) {
                AccessExpressionMarkings aem1 = (AccessExpressionMarkings) markings1;
                AccessExpressionMarkings aem2 = (AccessExpressionMarkings) markings2;
                return combine(List.of(aem1, aem2));
            } 
            
            throw new RuntimeException(String.format("Unknown markings class %s or %s", markings1.getClass().getName(), markings2.getClass().getName()));
```

core/utils/accumulo-utils/src/test/java/datawave/marking/AccessExpressionMarkingsTest.java
drewfarris, line 1: Nice coverage, but it would be cool to see a few more unhappy-path tests here (e.g., mismatched parenthesis, illegal operators like `&&` and `||` or `+` or `,`)

core/utils/accumulo-utils/src/test/java/datawave/marking/AccessExpressionUtilTest.java
drewfarris, line 1: Likewise, maybe a couple more bizarre tests here like `A&A|B&B|(A&A&A)` or `(A|B|C)&(B|C|D)` can normalize clean things like this up?

core/utils/metadata-utils/src/main/java/datawave/iterators/FrequencyMetadataAggregator.java
```diff
@@ -54,7 +54,7 @@ public class FrequencyMetadataAggregator extends WrappingIterator implements Opt
 
     private static final Logger log = Logger.getLogger(FrequencyMetadataAggregator.class);
     private static final String NULL_BYTE = "\0";
-    private static final MarkingFunctions markingFunctions = MarkingFunctions.Factory.createMarkingFunctions();
+    private final MarkingFunctions<?> markingFunctions;
```
drewfarris, line 57: I'm wondering if the prevalence of `MarkingFunctions<?>` and other wildcard generic types throughout this PR is suggesting that we should consider just using plain sub-classing instead of generics the code. Replacing the `<?>` everywhere it's used it going to be very difficult, but maybe that's not the end goal.

microservices/services/dictionary/api/pom.xml
drewfarris, line 1: Consider adding a direct dependency to accumulo-utils in the dictionary api pom?

core/cached-results/src/main/java/datawave/core/query/cachedresults/CacheableQueryRowImpl.java
```diff
@@ -133,18 +130,14 @@ private void manageColumnInsert(Type<?> datawaveType, String columnName, TypedVa
             } else {
                 columnValues.get(columnName).add(columnTypedValue.getValue().toString());
             }
-            Map<String,String> currMarkings = columnMarkingsMap.get(columnName);
+            Markings<?> currMarkings = columnMarkingsMap.get(columnName);
 
-            if (currMarkings.equals(markings) == false) {
+            if (!currMarkings.equals(markings)) {
```
drewfarris: Consider a null check here?

core/cached-results/src/main/java/datawave/core/query/cachedresults/CacheableQueryRowReader.java
```diff
@@ -86,22 +89,25 @@ public static CacheableQueryRow createRow(ResultSet cachedRowSet, Set<String> fi
...
             if (columnToIndexMap.get("_column_markings_") != null) {
                 String columnMarkings = cachedRowSet.getString(columnToIndexMap.get("_column_markings_"));
                 Map<String,String> combinedColumnMarkings = parseColumnMarkings(columnMarkings, columnToIndexMap);
-                Map<String,Map<String,String>> columnMarkingsMap = new HashMap<>();
+                Map<String,Markings<?>> columnMarkingsMap = new HashMap<>();
                 Map<String,String> columnVisibilityMap = new HashMap<>();
                 for (Map.Entry<String,String> entry : combinedColumnMarkings.entrySet()) {
                     String columnName = entry.getKey();
                     String combinedString = entry.getValue();
                     int x = combinedString.lastIndexOf(':');
```
drewfarris, line 104: Consider whether the combined string might have colons in it from the values created in `CacheableQueryRowImpl` where we encode `accessExpression + ':' + columnVisibility' and this get this split incorrect.

core/utils/accumulo-utils/src/main/java/datawave/marking/Markings.java
```diff
@@ -0,0 +1,20 @@
...
+
+import org.apache.accumulo.access.AccessExpression;
+import org.apache.accumulo.core.security.ColumnVisibility;
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS)
+public interface Markings<T> {
+
+    T getMarkings();
```
drewfarris, line 11: Consider whether we can just drop this method entirely from the interface definition and just remove the generic type on Markings.

core/utils/accumulo-utils/src/test/java/datawave/marking/AccessExpressionMarkingsTest.java
```diff
@@ -0,0 +1,372 @@
...
+        AccessExpressionMarkings b = AccessExpressionMarkings.create("A&B");
+        assertEquals(a, b);
+        assertEquals(a.hashCode(), b.hashCode());
+    }
+
+    // -- JAXB round-trip --
+
+    @Test
+    public void testJaxbMarshal() throws Exception {
+        JAXBContext ctx = JAXBContext.newInstance(AccessExpressionMarkings.class);
```
jschmidt10: You could move all the JAXB boilerplate into a method:

```
AccessExpressionMarkings original = // ... test setup
AccessExpressionMarkings copy = makeJaxbRoundtrip(original);

// assertions
```

core/utils/accumulo-utils/src/test/java/datawave/marking/AccessExpressionMarkingsTest.java
```diff
@@ -0,0 +1,372 @@
...
+        ProtostuffIOUtil.mergeFrom(bytes, deserialized, SCHEMA);
+
+        assertEquals(original.getColumnVisibilityString(), deserialized.getColumnVisibilityString());
+    }
+
+    @Test
+    public void testProtostuffRoundTripEmpty() {
+        AccessExpressionMarkings original = AccessExpressionMarkings.builder().accessExpression(null).build();
+
+        LinkedBuffer buffer = LinkedBuffer.allocate();
```
jschmidt10: Same comment as JAXB roundtrip boilerplate.

core/utils/accumulo-utils/pom.xml
```diff
@@ -93,16 +93,33 @@
             <groupId>org.xerial.snappy</groupId>
             <artifactId>snappy-java</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>1.18.36</version>
```
jschmidt10: Can we bubble these versions up to the top level parent and just inherit here?

microservices/services/accumulo/api/pom.xml
```diff
@@ -19,7 +19,7 @@
     <properties>
         <datawave.webservice.namespace>http://webservice.datawave.nsa/v1</datawave.webservice.namespace>
         <version.commons-codec>1.17.1</version.commons-codec>
-        <version.datawave>7.33.1</version.datawave>
+        <version.datawave>7.39.0-SNAPSHOT</version.datawave>
```
jschmidt10: Do you have changes in other repos that we'll need before we can merge this?

microservices/services/dictionary/api/src/main/java/datawave/webservice/dictionary/edge/DefaultEdgeDictionary.java
```diff
@@ -24,6 +24,7 @@
 import io.protostuff.Message;
 import io.protostuff.Output;
 import io.protostuff.Schema;
+import lombok.Setter;
```
jschmidt10, line 27: Is this the only Lombok addition? If the plan is to use it across the board, I'm cool but if this is the only usage, I'd suggest we drop the dependency and leave the original setter.

core/modification/src/main/java/datawave/modification/MutableMetadataHandler.java
```diff
@@ -847,19 +866,24 @@ protected List<Pair<Key,Value>> getField(AccumuloClient client, Set<Authorizatio
...
 
                 if (null != oldColumnVisibility) {
-                    // need to compare the flattened values for equivalence. It's possible for the visibility to be in a different order
-                    String oldColViz = new String(oldColumnVisibility.flatten(), "UTF-8");
-                    String thisVis = new String(thisViz.flatten(), "UTF-8");
+                    // need to compare the normalized values for equivalence. It's possible for the visibility to be in a different order
+                    String oldColViz = AccessExpressionUtil.normalize(AccessExpressionUtil.toAccessExpression(oldColumnVisibility)).getExpression();
+                    String thisVis = AccessExpressionUtil.normalize(AccessExpressionUtil.toAccessExpression(thisViz)).getExpression();
                     if (!oldColViz.equals(thisVis)) {
                         log.trace("Skipping key that does not match with column visibility: {}", e.getKey());
```
avgAGB: small touch, but could check if trace is enabled before the log statement. also on line 880

microservices/services/dictionary/service/src/main/java/datawave/microservice/dictionary/DataDictionaryControllerLogic.java
```diff
@@ -167,8 +171,7 @@ public VoidResponse setDescriptionPut(String fieldName, String datatype, String
     public VoidResponse setDescriptionPost(String fieldName, String datatype, String description, String modelName, String modelTable, String columnVisibility,
                     DatawaveUserDetails currentUser) throws Exception {
         DESC desc = this.responseObjectFactory.getDescription();
-        Map<String,String> markings = Maps.newHashMap();
-        markings.put("columnVisibility", columnVisibility);
+        Markings<?> markings = AccessExpressionMarkings.builder().accessExpression(ACCESS.newExpression(columnVisibility)).build();
```
jschmidt10: It looks like a large portion of AccessExpressionMarkings are being built from columnVisibilities. Thoughts on adding that to the builder? 

```
AccessExpressionMarkings.builder().colvis(columnVisibility).build();
```

microservices/services/dictionary/service/src/test/java/datawave/microservice/dictionary/DataDictionaryControllerTest.java
```diff
@@ -129,11 +128,10 @@ public void testPostDescriptions() {
                 .build();
         // @formatter:on
 
-        HashMap<String,String> markings = new HashMap<>();
-        markings.put(MarkingFunctions.Default.COLUMN_VISIBILITY, "USER|ADMIN");
         Multimap<Map.Entry<String,String>,DefaultDescription> descriptions = HashMultimap.create();
-        descriptions.put(new AbstractMap.SimpleEntry<>("fooField", "fooType"), new DefaultDescription("my foo field", markings));
-        descriptions.put(new AbstractMap.SimpleEntry<>("barField", "barType"), new DefaultDescription("my bar field", markings));
+        descriptions.put(new AbstractMap.SimpleEntry<>("fooField", "fooType"), new DefaultDescription("my foo field", AccessExpressionMarkings.create("USER")));
```
jschmidt10: Not your code but a nice improvement over the entry creation is `Map.entry(key, value)` instead of the inner SimpleEntry thing here.

web-services/client/src/main/java/datawave/webservice/result/DefaultEdgeQueryResponse.java
jschmidt10: Seems like a bug but is likely unused. This is ultimately called by RemoteEdgeQueryLogic which I don't see any runtime uses of.

microservices/services/query-metric/service/src/main/java/datawave/microservice/querymetric/handler/ShardTableQueryMetricHandler.java
```diff
@@ -365,14 +366,9 @@ protected Multimap<BulkIngestKey,Value> getEntries(ContentIndexingColumnBasedHan
...
-        Map<String,String> markings = updatedQueryMetric.getMarkings();
+        Markings<?> markings = updatedQueryMetric.getMarkings();
         if (markings != null && !markings.isEmpty()) {
-            try {
-                event.setVisibility(this.markingFunctions.translateToColumnVisibility(updatedQueryMetric.getMarkings()));
-            } catch (MarkingFunctions.Exception e) {
-                log.error(e.getMessage(), e);
-                event.setVisibility(this.queryMetricHandlerProperties.getDefaultMetricVisibility());
-            }
+            event.setVisibility(markings.toColumnVisibility());
```
jschmidt10, line 371: Is there error handling somewhere else in this case? I think these query metric markings are influenced by users which is why we had the default fallback.

warehouse/ingest-csv/src/main/java/datawave/ingest/csv/mr/input/CSVRecordReader.java
```diff
@@ -174,14 +171,16 @@ protected void checkField(String fieldName, String fieldValue) {
...
-                this.securityMarkings = new HashMap<>();
-            }
             if (!StringUtils.isEmpty(fieldValue)) {
-                this.securityMarkings.put(this.csvHelper.getSecurityMarkingFieldDomainMap().get(fieldName), fieldValue);
+                AccessExpression ae = AccessExpressionUtil.normalize(AccessExpressionMarkings.ACCESS.newExpression(fieldValue));
+                AccessExpressionMarkings newMarking = AccessExpressionMarkings.builder().accessExpression(ae).build();
+                if (securityMarkings == null) {
+                    securityMarkings = newMarking;
+                } else {
+                    securityMarkings = new MarkingFunctions.Default().combine(securityMarkings, newMarking);
```
jschmidt10, line 182: Maybe not worth addressing right this second but this seems to lock anyone using CSVRecordReader into the default MarkingFunctions impl.

warehouse/query-core/src/main/java/datawave/core/iterators/ResultCountingIterator.java
```diff
@@ -206,16 +207,19 @@ public Value getTopValue() {
...
+        ColumnVisibility columnVisibility;
         try {
-            cv = MarkingFunctions.Factory.createMarkingFunctions().combine(columnVisibilities);
+            columnVisibility = markingFunctions.combineVisibilities(columnVisibilities);
         } catch (MarkingFunctions.Exception e) {
-            log.error("Could not create combined columnVisibility for the count", e);
-            return null;
+            String message = "Could not create combined ColumnVisibility";
+            log.error(message, e);
+            throw new RuntimeException(message, e);
```
jschmidt10: This is a behavior change. Was that intentional?

warehouse/query-core/src/main/java/datawave/query/discovery/TermInfoAggregation.java
```diff
@@ -73,7 +75,8 @@ public DiscoveredThing apply(Collection<TermInfo> from) {
                 chosenCount = showReferenceCountInsteadOfTermCount ? referenceCount : termCount;
 
                 try {
-                    MarkingFunctions.Factory.createMarkingFunctions().translateFromColumnVisibility(ti.vis); // just to test parsing
+                    markingFunctions.translateFromColumnVisibility(ti.vis); // just to test
```
jschmidt10: strange word wrap on the comment here

warehouse/query-core/src/main/java/datawave/query/discovery/TermInfoAggregation.java
```diff
@@ -119,7 +123,9 @@ public DiscoveredThing apply(Collection<TermInfo> from) {
                     countsByVis.put(new Text(entry.getKey()), new VLongWritable(entry.getValue()));
                 }
 
-                return new DiscoveredThing(term, field, type, date, new String(columnVisibility.flatten()), count, countsByVis);
+                assert columnVisibility != null;
```
jschmidt10: Curious about the assert usage being added. I haven't seen it utilized anywhere else in the code. Any particular reason vs letting it fall to a NPE or whatever exception will be generated later?

warehouse/query-core/src/test/java/datawave/query/transformer/GroupingTest.java
```diff
@@ -604,13 +605,17 @@ public void testGroupByWithReducedResponse() throws Exception {
         for (QueryResult result : queryResults) {
             // noinspection rawtypes
             for (EventBase event : result.response.getEvents()) {
-                String eventCV = event.getMarkings().get(COLVIS_MARKING).toString();
+                Markings<?> eventMarkings = event.getMarkings();
```
jschmidt10: For these telescoping null check pattern, you can use java Optional:

```
String eventCV = Optional.ofNulable(event.getMarkings()).map(Markings::getMarkings).map(Object::toString).orElse(null);
```

core/cached-results/src/main/java/datawave/core/query/cachedresults/CacheableQueryRowImpl.java
jalphonso: updated

core/cached-results/src/main/java/datawave/core/query/cachedresults/CacheableQueryRowImpl.java
jalphonso: updated

core/modification/src/test/java/datawave/modification/MutableMetadataHandlerTestSupport.java
jalphonso: We have this as a String in so many places I decided not to attempt to use the constant everywhere as it may have resulted in pulling in dependencies solely for the constant.

core/utils/accumulo-utils/src/main/java/datawave/marking/MarkingFunctions.java
jalphonso: updated

microservices/services/dictionary/api/pom.xml
jalphonso: done

core/cached-results/src/main/java/datawave/core/query/cachedresults/CacheableQueryRowImpl.java
jalphonso: thanks, I added a case to handle when the currMarking is null

core/cached-results/src/main/java/datawave/core/query/cachedresults/CacheableQueryRowReader.java
jalphonso: I think this looks ok. I reviewed the logic in `CacheableQueryRowImpl` and it seems it only creates this combination when both sides are non-null. Also this logic is the same on integration.

core/utils/accumulo-utils/src/main/java/datawave/marking/Markings.java
jalphonso: may review on follow on tickets, for now want to see how the rest of our implementation shakes out

core/utils/metadata-utils/src/main/java/datawave/iterators/FrequencyMetadataAggregator.java
jalphonso: For now since this works, I'd like to leave it as is but am willing to test out alternative implementations if this code remains.

core/utils/accumulo-utils/src/test/java/datawave/marking/AccessExpressionMarkingsTest.java
jalphonso: done

core/utils/accumulo-utils/src/test/java/datawave/marking/AccessExpressionUtilTest.java
jalphonso: added some more tests. you can also see the library's tests here -> https://github.com/apache/accumulo-access/blob/main/modules/core/src/test/java/org/apache/accumulo/access/impl/AccessExpressionTest.java

core/utils/accumulo-utils/src/test/java/datawave/marking/AccessExpressionMarkingsTest.java
jalphonso: done

core/utils/accumulo-utils/src/test/java/datawave/marking/AccessExpressionMarkingsTest.java
jalphonso: done

core/utils/accumulo-utils/pom.xml
jalphonso: yep

microservices/services/accumulo/api/pom.xml
jalphonso: yes integration testing will take care of the validation

microservices/services/dictionary/api/src/main/java/datawave/webservice/dictionary/edge/DefaultEdgeDictionary.java
jalphonso: No there are a few other files in this module that use other lombok annotations

microservices/services/dictionary/service/src/main/java/datawave/microservice/dictionary/DataDictionaryControllerLogic.java
jalphonso: thanks for pointing this out, I added this and it looks much cleaner now

microservices/services/dictionary/service/src/test/java/datawave/microservice/dictionary/DataDictionaryControllerTest.java
jalphonso: easy fix

core/modification/src/main/java/datawave/modification/MutableMetadataHandler.java
jalphonso: done

microservices/services/query-metric/service/src/main/java/datawave/microservice/querymetric/handler/ShardTableQueryMetricHandler.java
jalphonso: I'll double check tomorrow but I believe all the functionality that existed in `MarkingFunctions.translateToColumnVisibility` was moved to the `Markings.toColumnVisibility` method.
Thanks for pointing this out.

warehouse/query-core/src/main/java/datawave/query/discovery/TermInfoAggregation.java
jalphonso: fixed

warehouse/query-core/src/main/java/datawave/query/discovery/TermInfoAggregation.java
jalphonso: it can be removed, it was an IntelliJ suggestion. probably minimal value. only advantage I thought of was it just caught it earlier in an obvious way

warehouse/query-core/src/test/java/datawave/query/transformer/GroupingTest.java
jalphonso: Thanks! updated

warehouse/query-core/src/main/java/datawave/core/iterators/ResultCountingIterator.java
jalphonso: it was not, reverted the behavior, thanks!

warehouse/ingest-csv/src/main/java/datawave/ingest/csv/mr/input/CSVRecordReader.java
jalphonso: this is an interesting one. I'll discuss with you.

core/utils/metadata-utils/src/main/java/datawave/query/model/ModelKeyParser.java
```diff
@@ -208,6 +214,9 @@ public static Mutation createMutation(FieldMapping mapping, String modelName) {
     }
 
     public static Mutation createDeleteMutation(FieldMapping mapping, String modelName) {
+        if (null == mapping.getColumnVisibility()) {
```
hlgp, line 217: If the visibility is null on a mapping, perhaps we intended for the vis to be empty?  Should this really throw an exception?

jalphonso: This is just catching the error earlier in a cleaner way. If the getColumnVisibility method returns a null string, then the next call to the ColumnVisibility constructor would result in an NPE when it calls expression.getBytes()
```java
    public ColumnVisibility(String expression) {
        this(expression.getBytes(StandardCharsets.UTF_8));
    }
```

core/utils/accumulo-utils/src/test/java/datawave/marking/AccessExpressionMarkingsTest.java
```diff
@@ -0,0 +1,385 @@
...
+    @Test
+    public void testJaxbUnmarshal() throws Exception {
+        String expression = "X|Y";
+        AccessExpressionMarkings markings = makeJaxbUnmarshall(expression);
+        assertEquals("X|Y", markings.getColumnVisibilityString());
+        assertFalse(markings.isEmpty());
+    }
+
+    @Test
+    public void testJaxbRoundTrip() throws Exception {
```
apmoriarty, line 235: Consider making the JSON/JAXB round trip methods that are called by every test in this class.

core/utils/accumulo-utils/src/test/java/datawave/marking/AccessExpressionUtilTest.java
```diff
@@ -0,0 +1,215 @@
...
+
+    @Test
+    public void testGetAccess() {
+        assertEquals(ACCESS, AccessExpressionUtil.getAccess());
+    }
+
+    // -- round-trip --
+
+    @Test
+    public void testRoundTripColumnVisibilityToAccessExpression() {
```
apmoriarty, line 191: same comment as before, consider moving this to a method that is called by every test with a valid expression

warehouse/query-core/src/main/java/datawave/query/attributes/AttributeBag.java
```diff
@@ -19,11 +18,15 @@ public abstract class AttributeBag<T extends Comparable<T>> extends Attribute<T>
     private static final Logger log = Logger.getLogger(AttributeBag.class);
     protected long shardTimestamp = Long.MAX_VALUE;
     protected boolean validMetadata = false;
+    protected MarkingFunctions<?> markingFunctions;
```
apmoriarty, line 21: Highlighting a major shift from static marking functions to per-thread marking functions.

warehouse/query-core/src/main/java/datawave/query/attributes/AttributeBag.java
```diff
@@ -78,11 +81,9 @@ private void updateMetadata() {
...
 
     protected ColumnVisibility combineAndSetColumnVisibilities(Collection<Attribute<? extends Comparable<?>>> attributes) throws Exception {
-        Collection<ColumnVisibility> columnVisibilities = Sets.newHashSet();
-        for (Attribute<?> attr : attributes) {
-            columnVisibilities.add(attr.getColumnVisibility());
-        }
-        return MarkingFunctions.Factory.createMarkingFunctions().combine(columnVisibilities);
+        Collection<ColumnVisibility> visibilities = attributes.stream().map(Attribute::getColumnVisibility).collect(Collectors.toList());
+        markingFunctions = getMarkingFunctions();
+        return markingFunctions.combineVisibilities(visibilities);
```
apmoriarty: If marking functions is null the call to `getMarkingFunctions` will set it. So this can be simplified to:
```java
return getMarkingFunctions().combineVisibilities(visibilities);
```

warehouse/query-core/src/main/java/datawave/query/predicate/ValueToAttributes.java
```diff
@@ -290,8 +290,18 @@ public Attribute<?> joinAttributes(String compositeName, Collection<Attribute<?>
...
         }
-        ColumnVisibility combinedColumnVisibility = this.markingFunctions.combine(columnVisibilities);
-        metadata = new Key(metadata.getRow(), metadata.getColumnFamily(), new Text(), combinedColumnVisibility, timestamp);
+        ColumnVisibility columnVisibility;
+        try {
+            if (null == markingFunctions) {
+                markingFunctions = MarkingFunctions.Factory.createMarkingFunctions();
+            }
+            columnVisibility = markingFunctions.combineVisibilities(columnVisibilities);
+        } catch (java.lang.Exception e) {
```
apmoriarty: no fully qualified import for exception

warehouse/query-core/src/test/java/datawave/query/transformer/GroupingTest.java
```diff
@@ -604,13 +606,15 @@ public void testGroupByWithReducedResponse() throws Exception {
...
                 // noinspection unchecked
                 for (FieldBase<?> field : (List<FieldBase<?>>) event.getFields()) {
-                    String fieldCV = field.getMarkings().get(COLVIS_MARKING);
-                    assertThat(fieldCV).describedAs("Assert null field cv for field: %s, teardown: %s, interrupt: %s", field.getName(), result.teardown,
-                                    result.interrupt).isNull();
+                    Markings<?> fieldMarkings = field.getMarkings();
+                    assertThat(fieldMarkings == null || fieldMarkings.isEmpty())
+                                    .describedAs("Assert empty field markings for field: %s, teardown: %s, interrupt: %s", field.getName(), result.teardown,
+                                                    result.interrupt)
+                                    .isTrue();
```
apmoriarty, line 617: did the formatter break?

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/attributes/AttributeBag.java
jalphonso: updated

warehouse/query-core/src/main/java/datawave/query/predicate/ValueToAttributes.java
jalphonso: updated

warehouse/query-core/src/test/java/datawave/query/transformer/GroupingTest.java
jalphonso: this one looks ok to me

core/utils/accumulo-utils/src/test/java/datawave/marking/AccessExpressionMarkingsTest.java
jalphonso: I'd like to understand this one more but can take care of it on a follow on MR

core/utils/accumulo-utils/src/test/java/datawave/marking/AccessExpressionUtilTest.java
jalphonso: I'd like to understand this one more but can take care of it on a follow on MR

apmoriarty (dismissed):

jschmidt10 (approved):

apmoriarty (approved):

jack-gitdev (approved):

PR #3470, opened 2026-03-23
Adds a `document:match` query function for substring matching against `d` column entries

drewfarris:
… `d` column entries

* Adds the `document:match(viewname, string)` and `document:match(string)` query functions that will scan the `d` columns of candidate documents at evaluation time and filter those candidates whose values do not contain the string specified.
  * Exposed via Lucene syntax using the `#DOCUMENT_MATCH` operator.
  * If no view name is included as a function parameter all 'd' columns will be scanned.
  * The viewname can be a prefix that ends with '*' to search all views with the specified prefix.
* If the specified string is found, the view name and start offsets for matches will be stored as a JSON map in the DOCUMENT_MATCHES field in the result.

This change includes:

* Lucene-to-JEXL translation
* Planner/iterator wiring
* Runtime document-match evaluation
* Configurable limits for `d` column sizes to prevent evaluation of large documents
* Unit and integration tests

While useful in its own right, this is a predecessor for more advanced matching functions on `d` column payloads.


warehouse/query-core/src/main/java/datawave/query/function/JexlEvaluation.java
```diff
@@ -97,7 +98,12 @@ public boolean apply(Tuple3<Key,Document,DatawaveJexlContext> input) {
             log.trace("Evaluating " + query + " against document " + input.second().getMetadata() + " with context " + input.third());
         }
 
-        Object o = script.execute(input.third());
+        DocumentMatchContext documentMatchContext = (DocumentMatchContext) input.third().get(DocumentFunctions.DOCUMENT_MATCH_CONTEXT_JEXL_VARIABLE_NAME);
+        if (documentMatchContext != null) {
+            documentMatchContext.clearMergedMatches();
+        }
+        Object o;
+        o = script.execute(input.third());
```
apmoriarty: This can be one line, no need to split it up.

warehouse/query-core/src/main/java/datawave/query/function/KeyToDocumentData.java
```diff
@@ -254,15 +254,52 @@ public List<Entry<Key,Value>> collectDocumentAttributes(final Key documentStartK
...
     }
 
+    /**
+     * extracts the proper column family byte sequence from a key regardless of whether it is an event key or a 'd' column key.
+     *
+     * @param key
+     *            the key to process
+     * @return the column family, consisting of datatype and uid.
+     */
+    private static ByteSequence getDocColumnFamily(Key key) {
```
apmoriarty: rename to `getDocumentKeyColumnFamily`?

warehouse/query-core/src/main/java/datawave/query/function/DocumentMatchContextFunction.java
```diff
@@ -0,0 +1,150 @@
...
+            collectDocumentColumnAttributes(documentKey, documentColumns);
+        }
+        return documentColumns;
+    }
+
+    private void collectDocumentColumnAttributes(Key documentKey, List<Entry<Key,Value>> documentColumns) throws IOException {
+        String row = documentKey.getRow().toString();
+        String datatypeAndUid = documentKey.getColumnFamily().toString();
+        Key startKey = new Key(row, "d", datatypeAndUid + '\0');
+        Key endKey = new Key(row, "d", datatypeAndUid + '\uffff');
```
apmoriarty: This range will aggregation all children for the provided document key. To aggregate just the child the range should be:
```
Key startKey = new Key(row, "d", datatypeAndUid + '\0');
Key endKey = new Key(row, "d", datatypeAndUid + '\0\0');
```

```
# given a document like
tld
tld.a
tld.a.1
tld.a.2
tld.b
tld.b.1
tld.b.2

# and document keys
tld.a
tld.b.2
```
In the given example I think this code would also pick up `tld.a.1` and `tld.a.2`

warehouse/query-core/src/main/java/datawave/query/function/DocumentMatchContextFunction.java
```diff
@@ -0,0 +1,150 @@
...
+        Range documentColumnRange = new Range(startKey, true, endKey, false);
+        if (log.isDebugEnabled()) {
+            log.debug("Seeking d-column range " + documentColumnRange + " for document key " + documentKey);
+        }
+
+        config.getSource().seek(documentColumnRange, Collections.emptyList(), false);
+
+        while (config.getSource().hasTop() && isDocumentColumn(config.getSource().getTopKey(), documentKey)) {
```
apmoriarty: Personal preference: I would have a `private final SortedKeyValueIterator source` that is set from the config in the constructor call. It would clean up a lot of the "config.getThing()" calls here.

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1075,16 +1105,18 @@ protected JexlEvaluation getJexlEvaluation(String query, NestedQueryIterator<Key
             arithmetic = getArithmetic();
         }
 
+        String rewrittenQuery = rewriteDocumentMatchFunctions(query, arithmetic);
```
apmoriarty: The query shouldn't change once it's in the tablet server/scan server. I would move the visitor call to the DefaultQueryPlanner

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/DocumentMatchFunctionRebuildingVisitor.java
```diff
@@ -0,0 +1,61 @@
...
+import datawave.query.jexl.functions.FunctionJexlNodeVisitor;
+
+/**
+ * Rewrites user-facing {@code document:match(...)} calls into the internal evaluation form that carries the reserved {@code documentMatchContext} argument
+ * explicitly.
+ * <p>
+ * This mirrors the way {@code content:*} functions are evaluated with an explicit {@code termOffsetMap} argument, but preserves the external user syntax for
+ * document matching.
+ */
+public class DocumentMatchFunctionRebuildingVisitor extends RebuildingVisitor {
```
apmoriarty: There's a lot of conventional wisdom that says to to use a rebuilding visitor when modifying the query tree, but in practice this has led to many hard to find bugs. Furthermore, copying every single node in a sufficiently large query can lead to excessive garbage collection.

I have a strong preference to rename this to `DocumentMatchFunctionVisitor extends BaseVisitor/ShortCircuitBaseVisitor`.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/DocumentMatchFunctionRebuildingVisitor.java
```diff
@@ -0,0 +1,61 @@
...
+     *
+     * @param script
+     *            script to inspect
+     * @return {@code true} if any document-match functions are present
+     */
+    public static boolean requiresDocumentMatchContext(ASTJexlScript script) {
+        return JexlASTHelper.getFunctionNodes(script).stream().map(FunctionJexlNodeVisitor::eval)
+                        .anyMatch(function -> DocumentFunctions.DOCUMENT_FUNCTION_NAMESPACE.equals(function.namespace())
+                                        && DocumentFunctions.DOCUMENT_MATCH_FUNCTION_NAME.equals(function.name()));
+    }
```
apmoriarty: While this is technically fine, in the event that the query does contain a document match function the planner will end up visiting the tree twice. Most visits are sub-millisecond so it's not a block but I would just call this visitor once via the static `rewrite` entry point.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/DocumentMatchFunctionRebuildingVisitor.java
apmoriarty: Also private constructor to enforce static access.

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/DocumentMatchFunctionRebuildingVisitor.java
```diff
@@ -0,0 +1,61 @@
...
+                return FunctionJexlNodeVisitor.makeFunctionFrom(visitor.namespace(), visitor.name(),
+                                JexlNodeFactory.buildIdentifier(DocumentFunctions.DOCUMENT_MATCH_CONTEXT_JEXL_VARIABLE_NAME),
+                                RebuildingVisitor.copy(visitor.args().get(0)));
+            } else if (visitor.args().size() == 2) {
+                return FunctionJexlNodeVisitor.makeFunctionFrom(visitor.namespace(), visitor.name(), RebuildingVisitor.copy(visitor.args().get(0)),
+                                JexlNodeFactory.buildIdentifier(DocumentFunctions.DOCUMENT_MATCH_CONTEXT_JEXL_VARIABLE_NAME),
+                                RebuildingVisitor.copy(visitor.args().get(1)));
+            }
+        }
+        return super.visit(node, data);
```
apmoriarty: No need to call `super.visit(node, data)`. Either the function was irrelevant, or it was rewritten. 

I would extract everything after the function namespace check to a separate method to make future refactors easier (i.e., adding a new function like `document:regex`). For example:

```
visit()
  if (namespaceIsDocumentFunction()) {
    handleDocumentFunction()
  }
}

handleDocumentFunction()
  switch(functionName)
    case documentMatch
    case documentRegex
    default
      log.warn("unknown document function:" + functionName
```

warehouse/query-core/src/main/java/datawave/query/jexl/functions/DocumentFunctionsDescriptor.java
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Argument-descriptor factory for the {@code document:*} JEXL namespace.
+ * <p>
+ * {@code document:match(...)} is an evaluation-only function. It does not contribute field normalization rules, event-data filters, index expansion, or
+ * ivarator pushdown. This descriptor exists primarily to validate the namespace/function pairing and to return a descriptor that tells the planner to leave the
+ * function in the evaluation phase.
+ */
+@SuppressWarnings("unused")
+public class DocumentFunctionsDescriptor implements JexlFunctionArgumentDescriptorFactory {
```
apmoriarty, line 27: General comment, not intended to prompt further changes in this merge request.

I may wish to match on multiple terms in the same function, or all terms in the same function. Given the constructor collision with the view-enabled match function, maybe it makes sense to split this out in the future

```
document:match(term)
document:match(termA, termB, termC, ...)
document:matchView(view, term)
document:matchView(view, termA, termB, termC, ...)
```
Need to spend more time thinking about this, but wanted to capture what I had at the moment.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -720,6 +721,7 @@ private void configureIterator(ShardQueryConfiguration config, IteratorSetting c
         addOption(cfg, QueryOptions.HIT_LIST, Boolean.toString(config.isHitList()), false);
         addOption(cfg, QueryOptions.TERM_FREQUENCY_FIELDS, Joiner.on(',').join(config.getQueryTermFrequencyFields()), false);
         addOption(cfg, QueryOptions.TERM_FREQUENCIES_REQUIRED, Boolean.toString(config.isTermFrequenciesRequired()), false);
+        addOption(cfg, QueryOptions.DOCUMENT_MATCH_CONTEXT_REQUIRED, Boolean.toString(config.isDocumentMatchContextRequired()), false);
```
apmoriarty: I would prefer to see a method called `addDocumentMatchOptions` that conditionally adds options. It can be annoying to see unused iterator options in the output of a `listscans` call. 

For example: always add `DOCUMENT_MATCH_CONTEXT_REQUIRED` but only add the max endcoded/decoded size if the first option is true.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/DocumentMatchFunctionRebuildingVisitorTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Verifies that the two-argument form keeps the view selector first and inserts the reserved context variable before the search string.
+     *
+     * @throws Exception
+     *             if parsing fails
+     */
+    @Test
+    public void testRewriteTwoArgumentFunction() throws Exception {
+        ASTJexlScript script = JexlASTHelper.parseAndFlattenJexlQuery("document:match('BODY', 'car')");
```
apmoriarty: Personal preference, extract common operations like parsing the query and rewriting the query to methods. That way the test is calling a method with just two arguments the input and the expected result. It sets up a nice pattern of 

```
void test()
  assertVisit(input, expected)

void assertVisit(input, expected)
  result = parseAndRewrite(input)
  assertEquals(expected, result)
```

warehouse/query-core/src/test/java/datawave/query/DocumentMatchQueryTest.java
```diff
@@ -0,0 +1,278 @@
...
+        expectedDocumentMatches.put("CAPONE", "{\"can\":{\"CONTENT\":[4,61]},\"lawyer\":{\"CONTENT2\":[2]}}");
+        planAndExecuteQuery();
+    }
+
+    /**
+     * Verifies Lucene {@code #DOCUMENT_MATCH(...)} translation and wildcard view-prefix behavior in the full query path.
+     */
+    @Test
+    public void testDocumentMatchLuceneWildcardView() throws Exception {
+        disableQueryPlanAssertion();
```
apmoriarty: Because the DocumentFunction is rewritten to include the `documentMatchContext` the expected final query plan should be specified. There have been bugs related to function rewriting with the content functions and this would be one additional check that reduces the risk of a similar bug being introduced during future refactors.

warehouse/query-core/src/test/java/datawave/query/DocumentMatchQueryTest.java
```diff
@@ -0,0 +1,278 @@
...
+        TimeZone.setDefault(TimeZone.getTimeZone("GMT"));
+        setClientForTest(client);
+
+        URL hadoopConfig = this.getClass().getResource("/testhadoop.config");
+        Preconditions.checkNotNull(hadoopConfig);
+        logic.setHdfsSiteConfigURLs(hadoopConfig.toExternalForm());
+        logic.setIvaratorCacheDirConfigs(Collections.singletonList(new IvaratorCacheDirConfig(folder.toUri().toString())));
+        logic.setMaxFieldIndexRangeSplit(1);
+        logic.setCollapseUids(false);
+        logic.setFullTableScanEnabled(true);
```
apmoriarty: In practice we should never enable full table scans.

warehouse/query-core/src/test/resources/datawave/query/QueryLogicFactory.xml
```diff
@@ -28,6 +28,7 @@
         <bean class="datawave.query.language.functions.jexl.AtomValuesMatchFunction"/>
         <bean class="datawave.query.language.functions.jexl.Options"/>
         <bean class="datawave.query.language.functions.jexl.Rename"/>
+        <bean class="datawave.query.language.functions.jexl.DocumentMatch"/>
```
apmoriarty, line 31: Love seeing default configs propagated into test configs

warehouse/query-core/src/main/java/datawave/query/jexl/functions/DocumentFunctionsDescriptor.java
drewfarris: Ok, yes - I thought about the multi term cases after the initial implementation but hadn't considered that this would result in needing to distinguish between match and matchView due to variable-length argument list. Thanks for the suggestion.

(As an aside, I hadn't done the mental exercise of considering what multiple document:match methods in a single query would mean for performance)

So, a operation like:
```
document:match(cookies cake cola milk)
```
Would require at least one of the words to occur in a document but would return offsets for all words.

There's also a future where we do something like treat the multiple terms as required or optional. This gets us to a similar place we might get with an AND or OR, but gives us a little different control. For example:

```
document:match(+cookies +cake cola milk)
```
Would require that a document has cookies and cake, but also return offsets where cola and milk occur

I think for now I skip the required case, and treat all multi term document:match operations as `at-least one` is required to match..

warehouse/query-core/src/test/java/datawave/query/DocumentMatchQueryTest.java
drewfarris: Wow, great feedback, thanks.

warehouse/query-core/src/main/java/datawave/query/function/DocumentMatchContextFunction.java
drewfarris: I think this means we might do something like:
```java
Key startKey = new Key(row, "d", datatypeAndUid + '\0');
Key endKey = config.isTld() ?
        new Key(row, "d", datatypeAndUid + '\uffff') :
        new Key(row, "d", datatypeAndUid + '.');
```
For the non-tld case, this will stop before we hit any children for the current docKey. "\0\0" didn't do this. Does this approach make sense - or am I missing something?

warehouse/query-core/src/main/java/datawave/query/jexl/visitors/DocumentMatchFunctionRebuildingVisitor.java
drewfarris: I need to look at this a bit more closely - simply subclassing BaseVisitor/ShortCircuitBaseVisitor instead of RebuildingVisitor led to NPE's in `JexlStringBuildingVisitor.buildQueryWithoutParse(JexlStringBuildingVisitor.java:119)`

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1084,7 +1113,7 @@ protected JexlEvaluation getJexlEvaluation(String query, NestedQueryIterator<Key
             } else {
                 jexlEvaluationFunction = nestedQuery.getEvaluation();
                 if (null == jexlEvaluationFunction) {
-                    jexlEvaluationFunction = new JexlEvaluation(query, arithmetic);
+                    jexlEvaluationFunction = new JexlEvaluation(nestedQuery.getQuery(), arithmetic);
```
drewfarris, line 1116: Need to validate that this is correct.

warehouse/query-core/src/test/java/datawave/query/function/JexlEvaluationTest.java
```diff
@@ -192,6 +198,61 @@ public void testContentPhraseFunction() {
...
+            java.io.OutputStream b64s = java.util.Base64.getEncoder().wrap(bos);
+            java.util.zip.GZIPOutputStream gzip = new java.util.zip.GZIPOutputStream(b64s);
+            gzip.write(content.getBytes());
+            gzip.close();
+            b64s.close();
+            bos.close();
+            return bos.toByteArray();
+        } catch (java.io.IOException e) {
+            throw new RuntimeException(e);
+        }
```
apmoriarty, line 313: Do not need fully qualified names here, could be a single try-with-resources

warehouse/query-core/src/test/java/datawave/query/planner/DefaultQueryPlannerTest.java
```diff
@@ -34,6 +37,42 @@
...
+            DefaultQueryPlanner planner = new DefaultQueryPlanner();
+            ShardQueryConfiguration config = new ShardQueryConfiguration();
+            config.setDocumentMatchContextRequired(false);
+            config.setDocumentMatchMaxEncodedSize(111);
+            config.setDocumentMatchMaxDecodedSize(222);
+            IteratorSetting cfg = new IteratorSetting(10, "query", QueryIterator.class);
+
+            planner.configureDocumentMatchOptions(config, cfg);
+
+            Assertions.assertEquals("false", cfg.getOptions().get(QueryOptions.DOCUMENT_MATCH_CONTEXT_REQUIRED));
```
apmoriarty: Consider static import for Assertions

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/DocumentMatchFunctionVisitorTest.java
```diff
@@ -0,0 +1,81 @@
...
+     * @param input
+     *            the input to rewrite
+     * @throws ParseException
+     *             if parsing fails
+     */
+    private static void assertRewrite(String expected, String input) throws ParseException {
+        ASTJexlScript script = JexlASTHelper.parseAndFlattenJexlQuery(input);
+        DocumentMatchFunctionVisitor.rewrite(script);
+        String rewritten = JexlStringBuildingVisitor.buildQueryWithoutParse(script);
+        assertEquals(expected, rewritten);
```
apmoriarty, line 79: Might be worth adding the ASTValidator to the test to verify proper parentage after modifying the tree.
```
private final ASTValidator validator = new ASTValidator();

..

try {
  assertTrue(validator.isValid(negatedScript, RewriteNegationsVisitorTest.class.getSimpleName(), false));
} catch (InvalidQueryTreeException e) {
  fail("Unexpected failure while validating query tree: " + e.getMessage());
}
```

warehouse/query-core/src/main/java/datawave/query/jexl/functions/DocumentFunctionsDescriptor.java
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * Descriptor for {@code document:match(...)}.
+     * <p>
+     * The function is evaluated only after a candidate document has been materialized, so all index-planning hooks intentionally report no fields and no index
+     * query contribution.
+     */
+    public static class DocumentJexlArgumentDescriptor implements JexlArgumentDescriptor {
+        @Override
+        public org.apache.commons.jexl3.parser.JexlNode getIndexQuery(ShardQueryConfiguration config, MetadataHelper helper, DateIndexHelper dateIndexHelper,
```
apmoriarty, line 37: Probably don't need a fully qualified import here

warehouse/query-core/src/main/java/datawave/query/jexl/functions/DocumentFunctionsDescriptor.java
```diff
@@ -0,0 +1,107 @@
...
+                        Set<String> datatypeFilter) {
+            return TRUE_NODE;
```
FineAndDandy, line 39: I see that this is correct by looking at FunctionIndexQueryExpansionVisitor which calls this, but now could be a good time to update the interface docs to make TRUE_NODE explicit as the response if index searching should be skipped for the function. It isn't in the docs and doesn't seem intuitive

warehouse/query-core/src/main/java/datawave/query/function/DocumentMatchContextFunction.java
```diff
@@ -0,0 +1,152 @@
...
+            log.debug("Seeking d-column range " + documentColumnRange + " for document key " + documentKey);
+        }
+
+        source.seek(documentColumnRange, Collections.emptyList(), false);
+
+        while (source.hasTop() && isDocumentColumn(source.getTopKey(), documentKey)) {
+            if (log.isDebugEnabled()) {
+                log.debug("Collected d-column entry " + source.getTopKey() + " for document key " + documentKey);
+            }
+            documentColumns.add(Maps.immutableEntry(source.getTopKey(), source.getTopValue()));
```
FineAndDandy: although later you define and enforce document size limits, these are not enforced here. Its possible we aggregate all the d columns for the document and significantly over-allocate memory here

warehouse/query-core/src/main/java/datawave/query/function/DocumentMatchContextFunction.java
```diff
@@ -0,0 +1,152 @@
...
+    @Override
+    public Tuple3<Key,Document,Map<String,Object>> apply(Tuple3<Key,Document,Map<String,Object>> from) {
+        try {
+            Set<Key> documentKeys = getDocumentKeys(from.first(), from.second());
+            if (log.isDebugEnabled()) {
+                log.debug("Collecting document-match context for tuple key " + from.first() + " using document keys " + documentKeys);
+            }
+
+            List<Entry<Key,Value>> dEntries = collectDocumentColumnAttributes(documentKeys);
+            DocumentMatchContext context = DocumentMatchContext.from(dEntries, config.getTimeFilter(), config.getLimits());
```
FineAndDandy, line 62: at this point the context has full sized d-column keys, with a limit identified, but not applied

warehouse/query-core/src/main/java/datawave/query/function/DocumentMatchContextFunction.java
```diff
@@ -0,0 +1,152 @@
...
+            collectDocumentColumnAttributes(documentKey, documentColumns);
+        }
+        return documentColumns;
+    }
+
+    private void collectDocumentColumnAttributes(Key documentKey, List<Entry<Key,Value>> documentColumns) throws IOException {
+        String row = documentKey.getRow().toString();
+        String datatypeAndUid = documentKey.getColumnFamily().toString();
+        Key startKey = new Key(row, "d", datatypeAndUid + '\0');
+        Key endKey = config.isTld() ? new Key(row, "d", datatypeAndUid + '\uffff') : new Key(row, "d", datatypeAndUid + '.');
```
FineAndDandy: create constant for d-column

warehouse/query-core/src/main/java/datawave/query/function/JexlEvaluation.java
```diff
@@ -97,6 +98,10 @@ public boolean apply(Tuple3<Key,Document,DatawaveJexlContext> input) {
             log.trace("Evaluating " + query + " against document " + input.second().getMetadata() + " with context " + input.third());
         }
 
+        DocumentMatchContext documentMatchContext = (DocumentMatchContext) input.third().get(DocumentFunctions.DOCUMENT_MATCH_CONTEXT_JEXL_VARIABLE_NAME);
+        if (documentMatchContext != null) {
+            documentMatchContext.clearMergedMatches();
```
FineAndDandy: I don't believe this object will be reused so should not need to be cleared here, unless its a shortcut to initialize in which case this should be renamed

warehouse/query-core/src/main/java/datawave/query/jexl/functions/DocumentFunctions.java
```diff
@@ -0,0 +1,259 @@
...
+     */
+    static String decode(byte[] encoded, int maxDecodedValueSize) throws IOException {
+        byte[] decodedBytes;
+        try (ByteArrayInputStream bais = new ByteArrayInputStream(encoded)) {
+            decodedBytes = Base64.getMimeDecoder().decode(bais.readAllBytes());
+        }
+
+        try (ByteArrayInputStream decodedInput = new ByteArrayInputStream(decodedBytes);
+                        GZIPInputStream gzipInputStream = new GZIPInputStream(decodedInput);
+                        ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
```
FineAndDandy: There are a few existing implementations of this already in the code. Consider consolidating and reusing

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
```diff
@@ -128,6 +129,16 @@ public Object visit(ASTFunctionNode node, Object data) {
 
         addHits(result);
 
+        if (isDocumentMatchFunction(nodeString) && result instanceof String) {
+            if (hasSiblings(node)) {
+                resultMap.put(nodeString, result);
+                return result;
+            }
```
FineAndDandy: this is identical to what happens below if it isn't a docuemntMatchFunction

warehouse/query-core/src/main/java/datawave/query/jexl/DatawaveInterpreter.java
```diff
@@ -128,6 +129,16 @@ public Object visit(ASTFunctionNode node, Object data) {
...
+            boolean matched = !((String) result).isEmpty();
+            resultMap.put(nodeString, matched);
+            return matched;
```
FineAndDandy, line 135: this looks like the only special case for a documentMatchFunction

warehouse/query-core/src/main/java/datawave/query/language/functions/jexl/DocumentMatch.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * JEXL-language representation of {@code document:match(...)}.
+ * <p>
+ * This function is produced by the query-language layer after parsing or after Lucene-to-JEXL translation. It validates the supported one-argument and
+ * two-argument forms and renders the canonical JEXL syntax consumed by the runtime query planner.
+ */
+public class DocumentMatch extends JexlQueryFunction {
+    public DocumentMatch() {
+        super("DOCUMENT_MATCH", new ArrayList<>());
```
FineAndDandy: use constant

warehouse/query-core/src/main/java/datawave/query/language/functions/jexl/DocumentMatch.java
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Renders the canonical JEXL form {@code document:match(...)} with escaped arguments.
+     *
+     * @return JEXL representation of this function
+     */
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
+        sb.append("document:match(");
```
FineAndDandy: use constants

warehouse/query-core/src/main/java/datawave/query/language/functions/lucene/DocumentMatch.java
```diff
@@ -0,0 +1,102 @@
...
+import datawave.webservice.query.exception.DatawaveErrorCode;
+
+/**
+ * Lucene-language representation of {@code #DOCUMENT_MATCH(...)}.
+ * <p>
+ * This class exists in the Lucene query-language layer so the parser can recognize the function and carry it through the same fielded-filter machinery used by
+ * other Lucene functions before the query is rendered into JEXL. The runtime semantics are still provided by {@code document:match(...)} in the evaluation
+ * phase.
+ */
+@Deprecated
```
FineAndDandy, line 23: new class @Deprecated?

warehouse/query-core/src/main/java/datawave/query/language/functions/lucene/DocumentMatch.java
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        @Override
+        public String toString() {
+            return renderedQuery;
+        }
+    }
+
+    public DocumentMatch() {
+        super("DOCUMENT_MATCH", new ArrayList<>());
```
FineAndDandy: use constant, shared with jexl class

warehouse/query-core/src/main/java/datawave/query/language/functions/lucene/DocumentMatch.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /**
+     * Builds the parser-layer selector text {@code match(...)} from the raw Lucene arguments.
+     *
+     * @return selector text used by the synthetic fielded filter
+     */
+    private String buildSelector() {
+        StringBuilder sb = new StringBuilder();
+        sb.append("match(");
```
FineAndDandy: use constant shared with jexl class

warehouse/query-core/src/main/java/datawave/query/language/functions/lucene/DocumentMatch.java
```diff
@@ -0,0 +1,102 @@
...
+@Deprecated
+public class DocumentMatch extends LuceneQueryFunction {
+    private static class DocumentMatchFilter extends WildcardFieldedFilter {
+        private final String renderedQuery;
+
+        DocumentMatchFilter(String selector) {
+            super(true, WildcardFieldedFilter.BooleanType.AND);
+            setField("document");
+            setSelector(selector);
+            this.renderedQuery = "document:" + selector;
```
FineAndDandy: use constants shared with jexl class

warehouse/query-core/src/main/java/datawave/query/jexl/functions/DocumentFunctions.java
```diff
@@ -0,0 +1,259 @@
...
+                    if (log.isDebugEnabled()) {
+                        log.debug("document:match found offsets " + offsets + " for search [" + search + "] in view [" + candidateView + "] using key "
+                                        + entry.getKey());
+                    }
+                    if (context.recordMatchingEntry(entry.getKey()) && context.shouldLogVisibilityMismatch()) {
+                        log.info("document:match encountered differing d-column visibilities for document " + context.getFirstMatchingEntry().getRow() + '/'
+                                        + context.getFirstMatchingEntry().getColumnFamily() + "; using visibility from first matched d-column "
+                                        + context.getFirstMatchingEntry() + " and ignoring differing visibility on " + entry.getKey());
+                        context.markVisibilityMismatchLogged();
+                    }
```
FineAndDandy: This feels a bit odd. I'd like to discuss this one in person.

warehouse/query-core/src/main/java/datawave/query/function/KeyToDocumentData.java
```diff
@@ -254,15 +254,52 @@ public List<Entry<Key,Value>> collectDocumentAttributes(final Key documentStartK
...
     }
 
+    /**
+     * extracts the proper column family byte sequence from a key regardless of whether it is an event key or a 'd' column key.
+     *
+     * @param key
+     *            the key to process
+     * @return the column family, consisting of datatype and uid.
+     */
+    private static ByteSequence getDocumentColumnFamily(Key key) {
```
FineAndDandy: I'm a little unclear on how a d-column would come off the IteratorBuildingVisitor to be processed here

warehouse/query-core/src/main/java/datawave/query/iterator/QueryIterator.java
```diff
@@ -1099,6 +1128,40 @@ protected JexlEvaluation getJexlEvaluation(String query, NestedQueryIterator<Key
...
+        NestedQuery<Key> nestedQuery = documentSource.getNestedQuery();
+        if (nestedQuery == null || nestedQuery.getQuery() == null) {
+            return true;
+        }
+
+        ASTJexlScript nestedScript = nestedQuery.getScript();
+        if (nestedScript == null) {
+            nestedScript = ArithmeticJexlEngines.getEngine(getArithmetic()).parse(nestedQuery.getQuery());
+        }
+        return DocumentMatchFunctionVisitor.requiresDocumentMatchContext(nestedScript);
```
FineAndDandy, line 1162: There are methods in JexlASTHelper that can parse the functions out of the JexlNode

warehouse/query-core/src/main/java/datawave/query/predicate/EventDataQueryFieldFilter.java
```diff
@@ -126,6 +134,12 @@ public boolean peek(@Nullable Map.Entry<Key,String> entry) {
      * @return true if the key should be retained
      */
     private boolean apply(Key key, boolean update) {
+        if (retainDocumentColumnFamily && DOCUMENT_COLUMN_FAMILY.equals(key.getColumnFamily().toString())) {
```
FineAndDandy: I commented on this in KeyToDocumentData as well, but I'm unsure how these keys would get here

warehouse/query-core/src/test/java/datawave/query/DocumentMatchQueryTest.java
```diff
@@ -0,0 +1,278 @@
...
+            String uuidValue = getUUID(uuid);
+            String expected = expectedDocumentMatches.get(uuidValue);
+            if (expected != null) {
+                Attribute<?> matches = result.get(DocumentFunctions.DOCUMENT_MATCHES);
+                assertNotNull(matches, "result did not contain DOCUMENT_MATCHES");
+                assertEquals(expected, matches.getData().toString());
+            }
+        }
+    }
+
```
FineAndDandy, line 189: add a test that verifies a document:match() without an anchored index term results in an error

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -339,6 +344,7 @@ public class QueryOptions implements OptionDescriber {
     protected EventDataQueryFilter eventEvaluationFilter;
     // filter specifically for event keys. required when performing a seeking aggregation
     protected EventDataQueryFilter eventFilter;
+    protected boolean retainDocumentColumnFamily = false;
```
FineAndDandy: if we can agree d-colums won't be aggregated into the event I think this can be removed

warehouse/query-core/src/main/java/datawave/query/function/JexlEvaluation.java
```diff
@@ -110,6 +115,16 @@ public boolean apply(Tuple3<Key,Document,DatawaveJexlContext> input) {
             ((DelayedNonEventIndexContext) input.third()).populateDocument(input.second());
         }
 
+        String documentMatches = (documentMatchContext == null) ? "" : DocumentFunctions.toJson(documentMatchContext.getMergedMatches());
```
drewfarris: Based on a conversation with @FineAndDandy, consider whether we want do the serialization to JSON here or move that to a transformer layer, so that the datastructure can be populated through other means (e.g., this is similar to the existing hit term code and thus should utilize a economy of mechanism instead of implementing its own path)

warehouse/query-core/src/main/java/datawave/query/jexl/functions/DocumentFunctions.java
drewfarris: I agree, I was never really happy with this and meant to revisit later. Thanks for paying attention to this.

I think the most basic way to approach this would be to avoid merging matches across multiple document row entries (views). Each view may have its own distinct column visibility, so the match data structure should echo those visibilities. If we avoid merging the results into a single data structure and instead return multiple values, each with its own visibility based on the view from which it was derived that should eliminate the need to 'choose first'.

So, bottom line the change to make here is to stop merging and return individual results each with their own visibilities.

warehouse/query-core/src/main/java/datawave/query/language/functions/lucene/DocumentMatch.java
drewfarris: This was for consistency with the other lucene functions, not entirely certain why they're all marked deprecated.

warehouse/query-core/src/main/java/datawave/query/function/JexlEvaluation.java
drewfarris: Removed this and added a unit test to validate that reuse of the documentMatchContext doesn't happen.

warehouse/query-core/src/main/java/datawave/query/function/DocumentMatchResults.java
```diff
@@ -0,0 +1,98 @@
...
+        String view = getView();
+        if (view == null || matches.isEmpty()) {
+            return payload;
+        }
+        payload.put(VIEW_FIELD, view);
+        Map<String,List<Integer>> jsonMatches = new LinkedHashMap<>();
+        for (Map.Entry<String,SortedSet<Integer>> matchEntry : matches.entrySet()) {
+            jsonMatches.put(matchEntry.getKey(), new ArrayList<>(matchEntry.getValue()));
+        }
+        payload.put(MATCHES_FIELD, jsonMatches);
```
drewfarris, line 82: Consider adding matches to the payload directly instead of converting to a LinkedHashMap

apmoriarty (dismissed):

PR #3478, opened 2026-03-26
Spotless maven plugin

jalphonso:
replace impsort with spotless maven plugin to automatically cleanup unused imports  
formatting is fast and occurs early in the maven lifecycle  
prevents checkstyle from failing later for unused imports


alerman (approved):

.github/workflows/check-formatting.yml
```diff
@@ -42,7 +42,7 @@ jobs:
     - name: Format code
       run: |
         if [[ "${{ github.event_name }}" != "merge_group" ]]; then
-          mvn --show-version --batch-mode --errors --no-transfer-progress "-Dstyle.color=always" clean formatter:format sortpom:sort impsort:sort -Dmaven.build.cache.enabled=false -Pautoformat -Dutils -Dservices -Dstarters
+          mvn --show-version --batch-mode --errors --no-transfer-progress "-Dstyle.color=always" clean formatter:format sortpom:sort -Dmaven.build.cache.enabled=false -Pautoformat -Dutils -Dservices -Dstarters
```
jalphonso, line 45: I will add spotless:apply here once we get all the microservice parent poms released

warehouse/query-core/src/main/java/datawave/query/language/builder/jexl/FunctionQueryNodeBuilder.java
```diff
@@ -29,6 +14,21 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
```
jalphonso, line 16: some of the licenses were in the middle of the imports and were getting removed by the spotless plugin. moving them to the top of the file resolves that.

apmoriarty (approved):

PR #3486, opened 2026-03-31
Fix CI bug for compose jobs

jalphonso:
Here's what changed and why:

Root cause: When a reusable workflow is called via workflow_call, the caller sees the overall workflow result as failure if any job in it fails — even if the gate job (compose-check-jobs) succeeds. So when one compose test fails but the other passes, the coordinator's gate correctly succeeds, but required-checks.yml sees needs.call-compose-tests.result == 'failure'.

Fix (two files):

1. compose-coordinator.yml — Added a workflow_call output (compose-check-result) that exposes the gate job's explicit status. The gate job now sets a status output to success when at least one test passed (before potentially exiting with failure when neither passed). Also replaced the actions/github-script step with a bash conditional for consistency.
2. required-checks.yml — Changed compose-check-jobs to check needs.call-compose-tests.outputs.compose-check-result instead of needs.call-compose-tests.result, so it reads the gate's actual determination rather than the overall workflow result.

### DIFF
--- a/.github/workflows/compose-coordinator.yml
+++ b/.github/workflows/compose-coordinator.yml
@@ -2,6 +2,10 @@ name: Compose Coordinator
 
 on:
   workflow_call:
+    outputs:
+      compose-check-result:
+        description: 'The result of the compose check gate job'
+        value: ${{ jobs.compose-check-jobs.outputs.status }}
 
 jobs:
   call-compose-tests:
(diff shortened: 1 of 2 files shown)


avgAGB (dismissed):

apmoriarty (approved):

dws4 (approved):
👍

PR #3487, opened 2026-03-31
Preserve token context for term frequency entries to avoid false positive hits across multiple field instances

drewfarris:
In some cases, we receive multiple values that get mapped to the same field name, e.g., CONTENT. When these fields are received independently in separate ingest passes and then tokenized, we wind up with tokens in the term frequency portion of the shard table that have offsets that overlap. This will become more frequent as we implement annotation indexing.

For example, one CONTENT field value might be CONTENT:"cat mat" and another value might be "dog bog". In this case, it's possible under certain circumstances that a query CONTENT:"cat bog" would hit on this field and require further filtering to determine that it's a false positive.

This feature adds the ability to employ grouping notation, similar to how it's used on the event fields, to maintain a distinction between the contexts within which a token appears as a part of the term frequency entry. It will involve modifications to ingest and query to preserve this extra context in the form of dot notation.

Replaces #2812 which was owned by @FineAndDandy so they can be a legit reviewer for this feature.


drewfarris:
No downstream effects without modifications to downstream to produce tokens with context.

FineAndDandy (dismissed):

awrigh11 (dismissed):

FineAndDandy (dismissed):

apmoriarty (dismissed):

avgAGB (dismissed):

core/modification/src/main/java/datawave/modification/MutableMetadataHandler.java
```diff
@@ -1159,6 +1163,47 @@ public void close() throws Exception {
         }
     }
 
+    protected TermFrequencyIterable getTermFrequencyKeys(AccumuloClient client, String shardTable, Set<Authorizations> userAuths, String shardId,
+                    String datatype, String eventUid, String fieldValue, String fieldName) throws Exception {
+
+        String qualifierPrefix = datatype + NULL_BYTE + eventUid + NULL_BYTE + fieldValue + NULL_BYTE + fieldName;
+        Key startKey = new Key(shardId, ColumnFamilyConstants.COLF_TF.toString(), qualifierPrefix);
+        Key endKey = new Key(shardId, ColumnFamilyConstants.COLF_TF.toString(), qualifierPrefix + MAX_CHAR);
+        Range range = new Range(startKey, true, endKey, true);
```
FineAndDandy, line 1175: this range could hit on a significant number unintended keys. Two iterators concated, one with the specific field, one with the grouping notation expansion that would be qualifierPrefix + "." + \u0000 to MAX_CHAR might avoid this.

drewfarris: This is a really good catch, thank you. I think one way to fix this would be to modify the end of the range to be `qualifierPrefix + "." + MAX_CHAR". That would work because there are no characters prior to "." that would constitute a legitimate component of a field name and the null character would indicate a complete field name of the field we want to retrieve.

FineAndDandy (approved):

avgAGB (approved):

PR #3510, opened 2026-04-14
Updated to avoid adding hidden fields when expanding model

ivakegg:
* Also added minor fixes to avoid periodically failing tests

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
+++ b/warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
@@ -1714,13 +1714,9 @@ public class ShardQueryLogic extends BaseQueryLogic<Entry<Key,Value>> implements
 
         logQuery(config.getQueryTree(), "Query after flattening");
 
-        // for model expansion, we want to only include fields for the datatypes being queried
-        Set<String> allFields = new HashSet<>(metadataHelper.getAllFields(config.getDatatypeFilter()));
-        // but do not include any of the hidden fields
-        allFields.removeAll(metadataHelper.getHiddenFields(config.getDatatypeFilter()));
-
         // Apply the query model.
-        config.setQueryTree(ShardQueryUtils.applyQueryModel(config.getQueryTree(), config, allFields, this.queryModel));
+        config.setQueryTree(ShardQueryUtils.applyQueryModel(config.getQueryTree(), config, metadataHelper.getModelExpansionFields(config.getDatatypeFilter()),
+                        this.queryModel));
 
         logQuery(config.getQueryTree(), "Query after applying query model");
 
--- a/warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
+++ b/warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
@@ -1714,9 +1714,13 @@ public class ShardQueryLogic extends BaseQueryLogic<Entry<Key,Value>> implements
 
         logQuery(config.getQueryTree(), "Query after flattening");
 
+        // for model expansion, we want to only include fields for the datatypes being queried
+        Set<String> allFields = new HashSet<>(metadataHelper.getAllFields(config.getDatatypeFilter()));
+        // but do not include any of the hidden fields
+        allFields.removeAll(metadataHelper.getHiddenFields(config.getDatatypeFilter()));
+
         // Apply the query model.
-        config.setQueryTree(ShardQueryUtils.applyQueryModel(config.getQueryTree(), config, metadataHelper.getAllFields(config.getDatatypeFilter()),
-                        this.queryModel));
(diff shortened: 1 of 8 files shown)


warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
```diff
@@ -1714,9 +1714,13 @@ public Object validateQuery(AccumuloClient client, Query settings, Set<Authoriza
...
 
+        // for model expansion, we want to only include fields for the datatypes being queried
+        Set<String> allFields = new HashSet<>(metadataHelper.getAllFields(config.getDatatypeFilter()));
+        // but do not include any of the hidden fields
+        allFields.removeAll(metadataHelper.getHiddenFields(config.getDatatypeFilter()));
+
         // Apply the query model.
-        config.setQueryTree(ShardQueryUtils.applyQueryModel(config.getQueryTree(), config, metadataHelper.getAllFields(config.getDatatypeFilter()),
-                        this.queryModel));
+        config.setQueryTree(ShardQueryUtils.applyQueryModel(config.getQueryTree(), config, allFields, this.queryModel));
```
apmoriarty: This feels brittle, perhaps some of this could be extracted to a method called `getFieldsForQueryModelExpansion()`

warehouse/query-core/src/test/java/datawave/query/jexl/lookups/BaseIndexLookupTest.java
```diff
@@ -267,14 +267,14 @@ protected void assertNoResults() {
...
-        assertEquals(expected, result.keySet());
+        assertEquals(new HashSet<>(expected), new HashSet<>(result.keySet()));
     }
 
     protected void assertResultValues(String field, Set<String> values) {
         Preconditions.checkNotNull(result, "result cannot be null");
         assertTrue(result.containsKey(field), "result did not contain field: " + field);
         Set<String> resultValues = new HashSet<>(result.get(field));
-        assertEquals(values, resultValues);
+        assertEquals(new HashSet<>(values), resultValues);
```
apmoriarty, line 277: nice

warehouse/query-core/src/test/java/datawave/query/util/sortedmap/MultiMapBackedKeyValueSortedMapTest.java
```diff
@@ -36,7 +36,7 @@ public Key createKey(byte[] values) {
 
     @Override
     public Value createValue(byte[] values) {
-        return new Value(values);
+        return new Value(values, true);
```
apmoriarty, line 39: If this fixes the random test failures I will be so happy

warehouse/query-core/src/test/java/datawave/query/transformer/QueryModelExpansionTests.java
```diff
@@ -92,6 +92,19 @@ public void testBasicExpansion() throws Exception {
         planAndExecuteQuery();
     }
 
+    /**
+     * Verify that expansion occurs but not for hidden field
```
apmoriarty, line 96: But not for fields marked with `{@link ColumnFamilyConstants#COLF_H}`

We should update the comment for the COLF_H in ColumnFamilyConstants because this column family is now used for more than display in the data dictionary. Now that I think about it perhaps a readme is in order to describe the lifecycle of a field.

warehouse/query-core/src/test/java/datawave/query/transformer/QueryModelExpansionTests.java
```diff
@@ -92,6 +92,19 @@ public void testBasicExpansion() throws Exception {
...
+     */
+    @Test
+    public void testHiddenExpansion() throws Exception {
+        givenQuery("ONE_HIDDEN == 'uuid'");
+        expectPlan("UUID == 'uuid'");
+        planAndExecuteQuery();
+        givenQuery("TWO_HIDDEN == 'uuid'");
+        expectPlan("UUID == 'uuid'");
+        planAndExecuteQuery();
```
apmoriarty, line 105: hot hamn, what a great looking test.

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/tables/ShardQueryLogic.java
ivakegg: not such a bad idea.... I will do it

warehouse/query-core/src/test/java/datawave/query/transformer/QueryModelExpansionTests.java
ivakegg: ok, I will try to update those comments at a minimum.  Julie requested that we fail queries hitting hidden fields which is reasonable.  That is what got us here.

ivakegg: I am updating the comment in the constants file.,  A readme will have to be a separate ticket/commit

billoley (approved):

Issue #3514, opened 2026-04-15
matchesInGroupLeft() does not correctly apply the index


FineAndDandy:
see #3512 

javadoc on the method indicates the correct behavior

```java
    /**
     * <pre>
     * 'args' will be either a matched set of field/regex pairs, or a matched set of field/regex pairs followed by an index integer,
     * in which case the integer is a zero based value to determine where to split off the 'group'.
     * FOO.1 with a index of '0' will split off '1'
     * FOO.BLAH.ZIP.0 with an index of '2' will split off 'BLAH.ZIP.0'
     * If no index is supplied, the default is 0
     * </pre>
     *
     * @param args
     *            matched set of field/regex pairs
     * @return a collection of matches
     */
    public static Collection<?> matchesInGroupLeft(Object... args) {
``` 

Current behavior does not follow this, Update EvaluationPhaseFilterFunctions.getMatchToLeftOfPeriod() to fix

PR #3513, opened 2026-04-15
Add functional tests for matchesInGroup and matchesInGroupLeft (#3512)

FineAndDandy:

PR #3515, opened 2026-04-16
Fixed matchesInGroupLeft to be consistent with documentation and matchesInGroup (#3514)

FineAndDandy:
depends on #3513 for base test expansion exposing inconsistencies


drewfarris (dismissed):

warehouse/query-core/src/main/java/datawave/query/jexl/functions/EvaluationPhaseFilterFunctions.java
```diff
@@ -1653,29 +1653,38 @@ public static ValueTuple getHitTerm(Object valueTuple) {
...
+     * - A value of 2 for pos will result in the substring 'FIRST.SECOND.THIRD'
+     * - A value of 3 for pos will result in the substring 'FIRST.SECOND.THIRD.FOURTH'
+     * - A value of 4+ for pos will result in null being returned
      * </pre>
      *
      * @param input
      *            the input string
      * @param pos
-     *            the Nth position of '.' to end the substring at
+     *            the index of the end group to include in the response. All indices will be included between zero and it
```
avgAGB: for the clause "between zero and it" is the it the end group?

FineAndDandy: I renamed the method and updated the javadoc. Hopefully it feels more clear now.

avgAGB (dismissed):

drewfarris (dismissed):

hlgp:
Looks good!  Much more intuitive.  Can we update the description here 
https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/query-core/src/main/java/datawave/query/language/functions/jexl/MatchesInGroupLeft.java#L19
to align with the changes in the new approach?

hlgp:
One other thing to consider might be to add a -1 option to indicate "I don't know or care how many generations there are, I want them ALL to match" or make 0 = everything, 1 = first group from left, 2 = second group from left, etc.

FineAndDandy:
> Looks good! Much more intuitive. Can we update the description here https://github.com/NationalSecurityAgency/datawave/blob/integration/warehouse/query-core/src/main/java/datawave/query/language/functions/jexl/MatchesInGroupLeft.java#L19 to align with the changes in the new approach?

Yep. Good catch

FineAndDandy:
> One other thing to consider might be to add a -1 option to indicate "I don't know or care how many generations there are, I want them ALL to match" or make 0 = everything, 1 = first group from left, 2 = second group from left, etc.

so that we don't change existing behavior of 0/default I'd prefer -1 for all. I think its reasonable to add this for both #MATCHES_IN_GROUP and #MATCHES_IN_GROUP_LEFT

hlgp (approved):

ivakegg (approved):

PR #3516, opened 2026-04-16
Remove Powermock from CreatedQueryLogicCacheBeanTest

apmoriarty:
Remove Powermock from CreatedQueryLogicCacheBeanTest

Replaces System.currentTimeMillis() call with Clock.millis(), allowing for easier mocking in unit tests.

Refactored CreatedQueryLogicCacheBeanTest such that asserts are more behavior driven and declarative.


avgAGB (dismissed):

web-services/query/src/main/java/datawave/webservice/query/cache/CreatedQueryLogicCacheBean.java
```diff
@@ -86,6 +87,7 @@ public boolean equals(Object o) {
     @Inject
     private AccumuloConnectionFactory connectionFactory;
     private final ConcurrentHashMap<Pair<String,Long>,Triple> cache = new ConcurrentHashMap<>();
+    private Clock clock = Clock.systemDefaultZone();
```
drewfarris: Should we just make this GMT instead of relying on the system'd default zone? If we don't keep this constant regardless of system, we might run into weird test behavior across systems.

apmoriarty: I was going to do this originally, but then it wouldn't be a one-to-one replacement of `System.currentTimeMillis()`, which uses the system clock. 

There are plenty of examples in both production code and test code that set the clock to either GMT or UTC. I think we should pick one or the other, personal preference for UTC

avgAGB (dismissed):

alerman (approved):
LGTM. Love that making this change is cleaning up the tests so well!

ivakegg (approved):

PR #3528, opened 2026-04-29
DocumentIterator stats overhaul

apmoriarty:
DocIdIteratorstats and DocIdQueryIteratorStats now support Kryo serialization.

CandidateResult encapsulates bulk candidates, query stats and iterator stats

Timing is now derived from Clock instead of using System calls

Do not use StatUtil now that time is tracked in milliseconds


warehouse/query-core/src/main/java/datawave/next/scanner/CandidateResultSerializer.java
```diff
@@ -0,0 +1,30 @@
+package datawave.next.scanner;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+
+import com.esotericsoftware.kryo.Kryo;
+import com.esotericsoftware.kryo.io.Input;
+import com.esotericsoftware.kryo.io.Output;
+
```
avgAGB, line 10: I noticed the CandidateResult.java had a javadoc that mentioned it serialized the document keys. So does this piece serialize a level deeper than that?

warehouse/query-core/src/main/java/datawave/next/stats/DocIdQueryIterStats.java
```diff
@@ -0,0 +1,166 @@
...
+import com.esotericsoftware.kryo.KryoSerializable;
+import com.esotericsoftware.kryo.io.Input;
+import com.esotericsoftware.kryo.io.Output;
+import com.google.common.base.Joiner;
+import com.google.common.base.Preconditions;
+
+/**
+ * Holds stats for document iterators by extending {@link DocIterStats} and also tracks total wall clock time and how many results were found
+ */
+public class DocIdQueryIterStats implements KryoSerializable, Serializable {
```
avgAGB, line 19: More general curiosity than anything, but was the previous class "DocIdQueryIteratorStats" and "DocumentIteratorStats" replaced with the renditions in this MR to avoid ambiguity? Or were there potential compatibility issues if those were refactored to include the changes in the newer classes?

apmoriarty: These docs are invalid, they were written during a previous iteration. I will update this.

warehouse/query-core/src/main/java/datawave/next/scanner/CandidateResultSerializer.java
apmoriarty: CandidateResult implements the KryoSerializable interface so the serialization happens natively inside that class. The serializer here is just a nice way to reuse some of the supporting kryo objects.

Rather than serialize the entire key I'm only writing the count, row, and then each column family. Updated serialize method in CandidateResult to explain that algorithm.

avgAGB (dismissed):

warehouse/query-core/src/main/java/datawave/next/scanner/CandidateResult.java
```diff
@@ -0,0 +1,107 @@
...
+import com.esotericsoftware.kryo.io.Input;
+import com.esotericsoftware.kryo.io.Output;
+
+import datawave.next.stats.DocIdQueryIterStats;
+import datawave.next.stats.DocIterStats;
+
+/**
+ * Serializes candidate document keys, {@link DocIdQueryIterStats} and {@link DocIterStats}
+ */
+public class CandidateResult implements KryoSerializable {
```
FineAndDandy, line 21: I'd like to see unit tests on these edges, You cover a few of these in CandidateResultSerializerTest, but ideally just hit these edges in their own test.

warehouse/query-core/src/main/java/datawave/next/DocIdQueryIterator.java
```diff
@@ -56,14 +57,21 @@ public class DocIdQueryIterator implements SortedKeyValueIterator<Key,Value> {
     private long scanTimeout = -1;
     private boolean allowPartialIntersections = true;
 
+    private final Value EMPTY_VALUE = new Value();
```
FineAndDandy: could be static

warehouse/query-core/src/main/java/datawave/next/DocIdQueryIterator.java
```diff
@@ -145,88 +153,81 @@ public boolean hasTop() {
...
-                tk = data.next();
+            if (batchSize == 1) {
+                handleSingleton();
+            } else {
+                handleBatch();
             }
+        } catch (Exception e) {
+            handleException(e);
+        }
+    }
```
FineAndDandy, line 169: Is this a paradigm we want to continue?

warehouse/query-core/src/main/java/datawave/next/DocIdQueryIterator.java
```diff
@@ -145,88 +153,81 @@ public boolean hasTop() {
...
 
-                // close enough
-                timingStats.markRetrievalStop();
+        if (!data.hasNext() && !statsReturned) {
+            statsReturned = true;
+            timingStats.markRetrievalStop(clock.millis());
+            result.setQueryStats(timingStats);
+            result.setIterStats(iteratorStats);
+            tv = serializeValue();
+        }
```
FineAndDandy, line 184: this can all be merged into the else if block above

warehouse/query-core/src/main/java/datawave/next/DocIdQueryIterator.java
```diff
@@ -145,88 +153,81 @@ public boolean hasTop() {
...
+     */
+    private Key createStatsKey() {
+        return new Key(range.getStartKey().getRow(), new Text("STATS"));
     }
 
     @Override
     public void seek(Range range, Collection<ByteSequence> columnFamilies, boolean inclusive) throws IOException {
         try {
-            timingStats.markScanStart();
+            timingStats.markScanStart(clock.millis());
```
FineAndDandy, line 233: if seek is called multiple times on this iterator it will reset the time

warehouse/query-core/src/main/java/datawave/next/scanner/DocumentIdProducer.java
```diff
@@ -138,61 +135,37 @@ private IteratorSetting createIteratorSetting() {
...
-            Set<Key> bulk = new HashSet<>();
-            for (String columnFamily : columnFamilies.split(",")) {
-                bulk.add(new Key(row, columnFamily));
-            }
+    private KeyWithContext parseEntry(Key key, Value value) throws IOException {
+        String cf = key.getColumnFamily().toString();
 
-            if (key.getColumnFamily().toString().equals("STATS")) {
-                // fake key was generated to return stats, return null so the producer skips this key
+        if (cf.equals("STATS")) {
```
FineAndDandy: constant please

warehouse/query-core/src/test/java/datawave/next/stats/StatUtilTest.java
```diff
@@ -0,0 +1,16 @@
...
+import org.junit.jupiter.api.Test;
+
+public class StatUtilTest {
+
+    @Test
+    public void testFormatNanos() {
+        assertEquals("0ns []", StatUtil.formatNanos(123L));
+        assertEquals("1ms [1ms]", StatUtil.formatNanos(1234567L));
+        assertEquals("1s 234ms [1234ms]", StatUtil.formatNanos(1234567890L));
+    }
```
FineAndDandy, line 14: Interesting observations

FineAndDandy (changes requested):
minor asks

warehouse/query-core/src/main/java/datawave/next/DocIdQueryIterator.java
apmoriarty: ...for now

warehouse/query-core/src/main/java/datawave/next/DocIdQueryIterator.java
apmoriarty: That's okay for now

warehouse/query-core/src/main/java/datawave/next/DocIdQueryIterator.java
apmoriarty: This actually cannot be merged. In one case nothing was found, a synthetic stats key was generated, no stats have bene returned yet, so we fall into here.

In the other case say we find a single result. Rather than return the result and then a second stats key this code ensures we can return stats with the last result. 

```
# given List.of(result1).iterator()

if (data.hasNext()) {
    tk = data.next();  # iterator advanced, hasNext is now false
} else if (!statsReturned) {
    tk = createStatsKey();
}

# drop into this
if (!data.hasNext() && !statsReturned) {
```

avgAGB (approved):

warehouse/query-core/src/main/java/datawave/next/DocIdQueryIterator.java
FineAndDandy: Opps. you are correct!

FineAndDandy (approved):

PR #3530, opened 2026-04-30
Added  values  only to discovery count.

hoper-38709:
Add values only option to return from a DiscoveryLogic/ DiscoveryIterator.


warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
```diff
@@ -102,6 +106,46 @@ private Multimap<String,TermEntry> getTermsByDatatype() throws IOException {
...
+        LinkedHashMultimap<String,TermEntry> terms = LinkedHashMultimap.create();
+        Key start = new Key(iterator.getTopKey());
+        Key key;
+        // If we should sum up counts, we want to collect the term entries for each date seen for the current field and term of start.
+        // Otherwise, we only want to collect the term entries for the current field, term, and date of start.
+        BiFunction<Key,Key,Boolean> dateMatchingFunction = sumCounts ? (first, second) -> true : this::datesMatch;
+        // Find all matching entries and parse term entries from them.
+
+        while (iterator.hasTop() && start.equals((key = iterator.getTopKey()), PartialKey.ROW_COLFAM) && dateMatchingFunction.apply(start, key)) {
+            // TermEntry termEntry = new TermEntry(key, iterator.getTopValue());
```
foster33: Can remove the identical commented out code.
```suggestion
```

warehouse/query-core/src/test/java/datawave/query/discovery/DiscoveryLogicTest.java
```diff
@@ -241,8 +241,10 @@ private void assertQueryResults() throws Exception {
         logic.setupQuery(config);
         Iterator<DiscoveredThing> iterator = logic.iterator();
         List<DiscoveredThing> actual = new ArrayList<>();
+        DiscoveredThingValuesOnlyConditionalTransformer dtvoct = new DiscoveredThingValuesOnlyConditionalTransformer(logic.getValuesOnly());
         while (iterator.hasNext()) {
-            actual.add(iterator.next());
+            actual.add(dtvoct.apply(iterator.next()));
+            // actual.add(iterator.next());
```
foster33: Leftover commented out code.
```suggestion
```

warehouse/query-core/src/test/java/datawave/query/discovery/DiscoveryLogicTest.java
```diff
@@ -520,4 +522,25 @@ public void testSumCountsForReverse() throws Exception {
...
+        givenQuery("bbc OR onyx");
+        givenStartDate("20130101");
+        givenEndDate("20130102");
+        givenParameter(DiscoveryLogic.SUM_COUNTS, "true");
+        givenParameter(DiscoveryLogic.VALUES_ONLY, "true");
+
+        // expect(new DiscoveredThing("bbc", "NETWORK", "csv", "", "FOO", 480L, new MapWritable()));
+        // expect(new DiscoveredThing("onyx", "POKEMON", "csv", "", "FOO", 110L, new MapWritable()));
+        // expect(new DiscoveredThing("onyx", "ROCK", "csv", "", "FOO", 4L, new MapWritable()));
+        // expect(new DiscoveredThing("onyx", "ROOSTER", "csv", "", "BAR", 240L, new MapWritable()));
```
foster33: Should these test cases be included?

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
hoper-38709: Done.

warehouse/query-core/src/test/java/datawave/query/discovery/DiscoveryLogicTest.java
hoper-38709: Done.

warehouse/query-core/src/test/java/datawave/query/discovery/DiscoveryLogicTest.java
hoper-38709: Done.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
```diff
@@ -102,6 +106,45 @@ private Multimap<String,TermEntry> getTermsByDatatype() throws IOException {
...
+    private Multimap<String,TermEntry> getTermsOnly() throws IOException {
+        LinkedHashMultimap<String,TermEntry> terms = LinkedHashMultimap.create();
+        Key start = new Key(iterator.getTopKey());
+        Key key;
+        // If we should sum up counts, we want to collect the term entries for each date seen for the current field and term of start.
+        // Otherwise, we only want to collect the term entries for the current field, term, and date of start.
+        BiFunction<Key,Key,Boolean> dateMatchingFunction = sumCounts ? (first, second) -> true : this::datesMatch;
+        // Find all matching entries and parse term entries from them.
+
+        while (iterator.hasTop() && start.equals((key = iterator.getTopKey()), PartialKey.ROW_COLFAM) && dateMatchingFunction.apply(start, key)) {
```
avgAGB: We ran into a case in another iterator through a little bit of a fluke, but would it be worthwhile to add a check to ensure the iterator is not null before checking the top?

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryTransformer.java
```diff
@@ -58,9 +59,12 @@ public EventBase transform(DiscoveredThing thing) {
 
         fields.add(this.makeField("VALUE", markings, "", 0L, thing.getTerm()));
         /**
-         * Added query model to alias FIELD
+         * Added query model to alias FIELD, if DiscoveredThing::field both not NULL and not empty.
```
hlgp, line 72: apply similar logic to DATE and DATATYPE.  We only want to show the term value and the combined vis.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
```diff
@@ -99,6 +132,48 @@ private Multimap<String,TermEntry> getTermsByDatatype() throws IOException {
...
+        // Find all matching entries and parse term entries from them.
+        //@formatter:off
+        while (Optional.ofNullable(iterator).isPresent() &&
+                iterator.hasTop() &&
+                start.equals((key = iterator.getTopKey()), PartialKey.ROW_COLFAM)) {
+
+            TermEntry termEntry = new TermEntry(key, iterator.getTopValue()) {
+                // Only use term and visibility for equality.
+                @Override
+                public boolean equals(Object o) {
```
hlgp: Move these to a TermOnlyEntry?

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
```diff
@@ -84,11 +112,16 @@ private Multimap<String,TermEntry> getTermsByDatatype() throws IOException {
...
+        // If we should sum up counts, we want to collect the term entries for each date seen for the current field and term of start.
+        // Otherwise, we only want to collect the term entries for the current field, term, and date of start.
         BiFunction<Key,Key,Boolean> dateMatchingFunction = sumCounts ? (first, second) -> true : this::datesMatch;
+
         // Find all matching entries and parse term entries from them.
-        while (iterator.hasTop() && start.equals((key = iterator.getTopKey()), PartialKey.ROW_COLFAM) && dateMatchingFunction.apply(start, key)) {
+        //@formatter:off
+        while (Optional.ofNullable(iterator).isPresent() &&
+                iterator.hasTop() &&
+                start.equals((key = iterator.getTopKey()), PartialKey.ROW_COLFAM) &&
```
hlgp: Compare only the ROW_ID, don't include the COLFAM

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
```diff
@@ -74,6 +88,20 @@ public void next() throws IOException {
                 }
             }
         }
+    }
+
+    @Override
+    public void next() throws IOException {
```
hlgp: I liked your previous implementation of this better, where depending on settings you called getTermsByDatatype or getTermsOnly. 
What "tentacling" are you trying to avoid?

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
hoper-38709: Keep as TermEntry so that, downstream, we can re-use aggregate function for both "term only" path and "summation path" . DiscoveredThing aggregate(Collection<TermEntry> termEntries)

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
hoper-38709: Shall do.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
hoper-38709: We getting increased impedance.  e.g. getTermsOnly returns list; getTermsByDataType returns nultimap.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
hlgp: I assumed we would make inheritance appropriate between them such that we could still use the aggregate

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveredThingValuesOnlyConditionalTransformer.java
```diff
@@ -0,0 +1,27 @@
...
+        // @formatter:off
+        return (valuesOnly) ? new DiscoveredThing(dt.getTerm(),
+                "",
+                "",
+                "",
+                dt.getColumnVisibility(),
+                0L,
+                new MapWritable())
+                : dt;
+        // @formatter:on
```
apmoriarty: Few comments here. Do not use a ternary operator and do not disable the formatter to put constructor args on different lines like this.

Use a simple guard statement with a simpler constructor like this:
```
if (valuesOnly ) {
  return new DiscoveredThing(dt.getTerm, dt.getColumnVisibility())
}
return dt;
```

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
```diff
@@ -84,11 +112,16 @@ private Multimap<String,TermEntry> getTermsByDatatype() throws IOException {
...
-        // If we should sum up counts, we want to collect the term entries for each date seen for the current field and term of start. Otherwise, we only want
-        // to collect the term entries for the current field, term, and date of start.
+        // If we should sum up counts, we want to collect the term entries for each date seen for the current field and term of start.
+        // Otherwise, we only want to collect the term entries for the current field, term, and date of start.
         BiFunction<Key,Key,Boolean> dateMatchingFunction = sumCounts ? (first, second) -> true : this::datesMatch;
+
         // Find all matching entries and parse term entries from them.
-        while (iterator.hasTop() && start.equals((key = iterator.getTopKey()), PartialKey.ROW_COLFAM) && dateMatchingFunction.apply(start, key)) {
+        //@formatter:off
+        while (Optional.ofNullable(iterator).isPresent() &&
```
apmoriarty: In the event that the iterator is not set, or is later set to null a NPE is appropriate. This code change will suppress a failure state that we want to know about. Please reset this.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
```diff
@@ -325,5 +405,22 @@ public long getUidListSize() {
...
+                                .append(getUidCount(), other.getUidCount()).append(getUidListSize(), other.getUidListSize()).isEquals();
+            }
+            return false;
+        }
+
+        @Override
+        public int hashCode() {
+            return new HashCodeBuilder().append(getTerm()).append(getField()).append(getVisibility()).append(getDate()).append(getDatatype())
+                            .append(getUidCount()).append(getUidListSize()).toHashCode();
+        }
```
apmoriarty, line 389: equals and hashcode methods are much easier to read when paired with disabling the formatter
```
@formatter:off
builder()
  .append(x)
  .append(y)
  .build();
@formatter:on
```

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveredThingValuesOnlyConditionalTransformer.java
hoper-38709: Good idea.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
hoper-38709: I'll re-set. Another commenter recommended a NULL check.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
hoper-38709: Understood. Agreed.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
hoper-38709: Currently field and date are also needed for a new "Top Entry"  set downstream. These are carried in the Term Entry. I'll add TermOnly for clarity.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
apmoriarty: I think this is a case when we want to surface a NPE up the stack instead of silently failing.

hlgp:
New direction:  Turns out we don't need to aggregate the vis.  We only need to return the first instance of each unique rowid.

So, if valuesOnly, break when you get the first valid TermEntry.  Each next iteration will have just a single DiscoveredThing.  Do the logic to only set the term and vis.

In setTop, precedent is already set to modify the column family to advance to the next column family.  In the values only case we will want to skip to the next row.  Note: appending a null byte to the end of the current row won't work, so recommend the next start key in the skip logic be the row and last possible column family (something like \uffff)

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
```diff
@@ -179,26 +246,43 @@ private DiscoveredThing aggregate(Collection<TermEntry> termEntries) {
...
         String row = (this.reverseIndex ? new StringBuilder().append(thing.getTerm()).reverse().toString() : thing.getTerm());
         Key newKey = new Key(row, thing.getField(), thing.getDate() + "\uffff");
 
         // Create a value from the list of things.
         ArrayWritable thingArray = new ArrayWritable(DiscoveredThing.class, things.toArray(new DiscoveredThing[0]));
         Value newValue = new Value(WritableUtils.toByteArray(thingArray));
+        if (valuesOnly){
+            Key skipKey = new Key(row, "\uffff");
+            if(!columnFamilies.isEmpty()){
+                skipKey = new Key(row, thing.getField(), "\uffff");
```
hlgp, line 209: Would be worth double checking with users to confirm desired behavior if they use valuesOnly with fielded literals.  In that case they may want the discovered thing to be populated with the field.  I could see the asks being:
- unfielded regex -> give me all unique values in the index that match this regex
- fielded regex -> give me all unique values for this field that match this regex
- unfielded literal -> confirm that this value exists at all?
- fielded literal -> confirm this value exists for this/these fields? or for which of these fields do we have this value?  The latter suggesting we should still return the field name in this case

hoper-38709: I contacted PO.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
hoper-38709: OBE. Refactored out.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
hlgp: Update: users want valuesOnly to be firm (meaning no additional info regardless of query syntax) and ideally perform deduping in the Transformer

warehouse/query-core/src/test/java/datawave/query/discovery/DiscoveryLogicTest.java
```diff
@@ -520,4 +530,99 @@ public void testSumCountsForReverse() throws Exception {
...
+        givenParameter(DiscoveryLogic.SUM_COUNTS, "true");
+        givenParameter(DiscoveryLogic.VALUES_ONLY, "true");
+
+        expect(new DiscoveredThing("bbc", "", "", "", "FOO", 0L, new MapWritable()));
+        expect(new DiscoveredThing("onyx", "", "", "", "FOO", 0L, new MapWritable()));
+        assertQueryResults();
+    }
+
+    @Test
+    public void testAccumuloCoreDataRange() {
```
hlgp: Guessing this was for self sanity checking? Remove in final iteration as this just tests accumulo ranges and nothing Discovery specific

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryTransformer.java
```diff
@@ -16,69 +22,105 @@
...
-    private BaseQueryLogic<DiscoveredThing> logic = null;
-    private QueryModel myQueryModel = null;
+    private final BaseQueryLogic<DiscoveredThing> logic;
+    private QueryModel myQueryModel;
     private MarkingFunctions markingFunctions;
-    private ResponseObjectFactory responseObjectFactory;
+    private final ResponseObjectFactory responseObjectFactory;
+    private boolean valuesOnly = false;
 
-    public DiscoveryTransformer(BaseQueryLogic<DiscoveredThing> logic, Query settings, QueryModel qm) {
```
hlgp, line 36: why move this?

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveredThingValuesOnlyConditionalTransformer.java
```diff
@@ -0,0 +1,19 @@
+package datawave.query.discovery;
+
+import java.util.function.UnaryOperator;
+
+public class DiscoveredThingValuesOnlyConditionalTransformer implements UnaryOperator<DiscoveredThing> {
+
+    boolean valuesOnly;
```
apmoriarty: private final

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveredThingValuesOnlyConditionalTransformer.java
```diff
@@ -0,0 +1,19 @@
+package datawave.query.discovery;
+
+import java.util.function.UnaryOperator;
+
+public class DiscoveredThingValuesOnlyConditionalTransformer implements UnaryOperator<DiscoveredThing> {
```
apmoriarty, line 11: class level comment please

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
```diff
@@ -179,26 +197,46 @@ private DiscoveredThing aggregate(Collection<TermEntry> termEntries) {
...
+    private void setTop(List<DiscoveredThing> things) throws IOException {
+
+        // We want the key to be the last possible key for this date. Return the key as it is in the index (reversed if
+        // necessary) to ensure the keys are consistent with the initial seek range.
         DiscoveredThing thing = things.get(0);
         String row = (this.reverseIndex ? new StringBuilder().append(thing.getTerm()).reverse().toString() : thing.getTerm());
         Key newKey = new Key(row, thing.getField(), thing.getDate() + "\uffff");
 
+        // Conditionally trim DiscoveredThing. (Perhaps to term and visibility).
+        // List<DiscoveredThing> thingsConditional = things.stream().map(discoveredThingTransformer).collect(Collectors.toList());
```
apmoriarty: Remove commented out code

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
```diff
@@ -179,26 +197,46 @@ private DiscoveredThing aggregate(Collection<TermEntry> termEntries) {
...
     }
 
     @Override
     public void seek(Range range, Collection<ByteSequence> columnFamilies, boolean inclusive) throws IOException {
+        this.lastRange = range;
+        this.columnFamilies = columnFamilies;
         this.iterator.seek(range, columnFamilies, inclusive);
         if (log.isTraceEnabled()) {
-            log.trace("My source " + (this.iterator.hasTop() ? "does" : "does not") + " have a top.");
+            log.trace("My source " + ((Optional.ofNullable(iterator).isPresent() && this.iterator.hasTop()) ? "does" : "does not") + " have a top.");
```
apmoriarty: Iterator is always set. Remove this use of Optional.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
```diff
@@ -210,13 +248,19 @@ public void init(SortedKeyValueIterator<Key,Value> source, Map<String,String> op
         this.showReferenceCount = Boolean.parseBoolean(options.get(DiscoveryLogic.SHOW_REFERENCE_COUNT));
         this.reverseIndex = Boolean.parseBoolean(options.get(DiscoveryLogic.REVERSE_INDEX));
         this.sumCounts = Boolean.parseBoolean(options.get(DiscoveryLogic.SUM_COUNTS));
+        this.valuesOnly = Boolean.parseBoolean(options.get(DiscoveryLogic.VALUES_ONLY));
```
apmoriarty, line 243: Is VALUES_ONLY always set as an option? Convention in other iterators goes something like this for required options
```
String value = options.get(OPTION_NAME)
if (value == null ){
   throw new Exception("Required option OPTION_NAME not set")
} else {
   bool = Boolean.parseBoolean(value)
}
```

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryTransformer.java
```diff
@@ -16,69 +22,105 @@
...
+    }
+
+    public DiscoveryTransformer(BaseQueryLogic<DiscoveredThing> logic, Query settings, QueryModel qm) {
+        super(new MarkingFunctions.Default());
+        this.markingFunctions = logic.getMarkingFunctions();
+        this.responseObjectFactory = logic.getResponseObjectFactory();
+        this.logic = logic;
+        this.myQueryModel = qm;
+        this.valuesOnly = getOrDefaultBoolean(settings, VALUES_ONLY, false);
+    }
```
apmoriarty: Constructors at the top

warehouse/query-core/src/test/java/datawave/query/discovery/DiscoveryLogicTest.java
```diff
@@ -241,16 +245,22 @@ private void assertQueryResults() throws Exception {
...
         List<DiscoveredThing> actual = new ArrayList<>();
+
+        DiscoveredThingValuesOnlyConditionalTransformer dtvoct = new DiscoveredThingValuesOnlyConditionalTransformer(logic.getValuesOnly());
         while (iterator.hasNext()) {
-            actual.add(iterator.next());
+            DiscoveredThing dtee = iterator.next();
+            actual.add(dtvoct.apply(dtee));
+
+            // actual.add(dtvoct.apply(iterator.next()));
+            // actual.add(iterator.next());
```
apmoriarty: remove commented out code

apmoriarty (changes requested):

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryTransformer.java
hoper-38709: okay.

warehouse/query-core/src/test/java/datawave/query/discovery/DiscoveryLogicTest.java
hoper-38709: okay.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
hoper-38709: Done.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveredThingValuesOnlyConditionalTransformer.java
hoper-38709: Done.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryTransformer.java
hoper-38709: I'll  put constructures before all functions.

warehouse/query-core/src/test/java/datawave/query/discovery/DiscoveryLogicTest.java
hoper-38709: Just sample learning code. Not needed.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveredThingValuesOnlyConditionalTransformer.java
hoper-38709: Shall add.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
hoper-38709: Good catch.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
hoper-38709: It's not required; false if not present. Same pattern as many opthers.

hlgp: This one is not required.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
```diff
@@ -84,15 +90,26 @@ private Multimap<String,TermEntry> getTermsByDatatype() throws IOException {
...
             TermEntry termEntry = new TermEntry(key, iterator.getTopValue());
-            if (termEntry.isValid())
+            if (termEntry.isValid()) {
                 terms.put(termEntry.getDatatype(), termEntry);
-            else {
+                // if 'values only' is selected, then we only need a single TermEntry, the first one. The value of each
+                // term in TermEntry should be the identical as we iterate in this while. Therefore, the first one
+                // encountered will suffice.
+                if (valuesOnly) {
+                    return terms;
```
hlgp, line 106: Since we're short circuiting here, we don't need the ROW vs COLFAM distinction above anymore.  The original COLFAM will do.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveryIterator.java
hoper-38709: Looking into Transformer dedup. Apply Commons:UniqueFilterIterator perhaps?

hoper-38709: Updated DiscoveryLogic adding UniqueFilter.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveredThingTermIsotope.java
```diff
@@ -0,0 +1,40 @@
...
+            DiscoveredThingTermIsotope other = (DiscoveredThingTermIsotope) o;
+            return new EqualsBuilder().append(getTerm(), other.getTerm()).isEquals();
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        // Ignore super results.
+        // super.hashCode();
```
apmoriarty: no dead code please

warehouse/query-core/src/test/java/datawave/query/discovery/DiscoveryLogicTest.java
```diff
@@ -520,4 +537,81 @@ public void testSumCountsForReverse() throws Exception {
...
+        givenQuery("rock:onyx OR pokemon:onyx");
+        givenStartDate("20130101");
+        givenEndDate("20130104");
+
+        givenParameter(DiscoveryLogic.SUM_COUNTS, "true");
+        givenParameter(DiscoveryLogic.VALUES_ONLY, "true");
+
+        expect(new DiscoveredThing("onyx", "", "", "", "FOO", 0L, new MapWritable()));
+        assertQueryResults();
+    }
```
apmoriarty, line 616: Great tests

apmoriarty (changes requested):

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveredThingTermIsotope.java
```diff
@@ -0,0 +1,38 @@
+package datawave.query.discovery;
+
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.apache.hadoop.io.MapWritable;
+
+/**
+ * A variant of DiscoveredThing in which equality if defined as comparing getTerm() only.
```
hlgp: is defined

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveredThingTermIsotope.java
hoper-38709: Fixed.

apmoriarty (dismissed):

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveredThing.java
```diff
@@ -29,6 +29,16 @@ public DiscoveredThing(String term, String field, String type, String date, Stri
         this.countsByColumnVisibility = countsByColumnVisibility;
     }
 
+    public DiscoveredThing(String term, String columnVisibility) {
+        this.term = term;
+        this.field = "";
+        this.type = "";
+        this.date = "";
+        this.columnVisibility = columnVisibility;
```
ivakegg, line 39: Perhaps all of the initial values should be done in the empty constructure, then this contructor can simply call that one followed by setting the term and visibility.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveredThingTermIsotope.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public DiscoveredThingTermIsotope() {
+        super();
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (o instanceof DiscoveredThingTermIsotope) {
+            DiscoveredThingTermIsotope other = (DiscoveredThingTermIsotope) o;
+            return new EqualsBuilder().append(getTerm(), other.getTerm()).isEquals();
```
ivakegg, line 29: should the super equals method or compareTo method be invoked as well?

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveredThingTermIsotope.java
```diff
@@ -0,0 +1,38 @@
...
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        return new HashCodeBuilder().append(getTerm()).toHashCode();
```
ivakegg, line 36: Just the term?

hoper-38709: Correct. The use case is maintaining uniqueness based on the term. Elaboration on justification:  We added to the code to use UniqueFilterIterator to remove duplicates on-the-fly. The value of "term" in  DiscoveredThing solely defines uniqueness. Therefore, equality, using a variant of DiscoveredThing is based on term only.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveredThingTermIsotope.java
hoper-38709: Using UniqueFilterIteraor, we only need hashCode() and equals(). CompareTo() in NOT needed.

warehouse/query-core/src/main/java/datawave/query/discovery/DiscoveredThing.java
hoper-38709: Adding an init block.

avgAGB (approved):

hlgp (approved):

PR #3532, opened 2026-05-01
Support dynamic query limit configuration

### LINKED ISSUE #3511: Query limiter needs to be dynamically configurable
The Query Limiter needs to be dynamically configurable in that we should be able to change limits while the webservers are running.  The thoughts for the implementation are:

1) The configuration will remain in a file form which could be the existing QueryLimiter.xml format or perhaps a json format
2) The location of the file will be stored in a zookeeper node that can be watched for changes
3) The location of the configuration file could be http, hdfs, or local file
4) When the zookeeper node changes, then the file location should be pulled from zookeeper, the file down/uploaded, and configuration updated internally.
    4.1) It would be nice to be able to signal via zookeeper that the file has changed but perhaps the location has not.

After this mechanism is implemented, we will be adding a page to the admin gui (implemented elsewhere) to allow updates to the configuration in a more user friendly manner

lbschanno:
Support the ability to dynamically update concurrent query limits while the webservers are running. 

Add the class `ZkObjectPublisher` and its corresponding implementation `ZkObjectPublisherImpl` that provides the ability to trigger and publish updates of a configured class to all subscribers. See this [README](https://github.com/NationalSecurityAgency/datawave/blob/task/dynamicQueryLimitConfig/web-services/query/src/main/java/datawave/zookeeper/README.md) for full details.

Integrate using a configured `ZkObjectPublisher` (implemented by `QueryLimitConfigPublisher` which extends `ZkObjectPublisherImpl`) within the `QueryLimiter` class to update its internal configured limits whenever an update has been triggered. 

Closes #3511

Depends on changes in PRs #3815 #3816 , #3716.


web-services/query/src/main/java/datawave/webservice/zookeeper/LockedZkClientDispatcher.java
```diff
@@ -0,0 +1,235 @@
...
+import java.util.concurrent.locks.ReentrantLock;
+
+import org.apache.curator.framework.CuratorFramework;
+import org.apache.curator.framework.CuratorFrameworkFactory;
+import org.apache.log4j.Logger;
+
+import com.google.common.base.Preconditions;
+
+/**
+ * A class that provides the ability to access a maintained {@link CuratorFramework} client that is guarded by a lock from access by multiple threads at the
```
ivakegg: Nice class... I like it.  Surprised this functionality is not available somewhere already.

web-services/query/src/main/java/datawave/zookeeper/README.md
```diff
@@ -0,0 +1,36 @@
+# ZkObjectPublisher
+
+The class [ZkObjectPublisher](ZkObjectPublisher.java) provides the ability to trigger and publish updates of a configured class instance to any subscribers using Zookeeper to listen for updates and triggering events. A publisher instance can be configured with the following:
```
ivakegg: wow... awesome writeup.

web-services/query/src/main/java/datawave/zookeeper/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+Upon receiving a trigger event, the publisher will attempt to read and deserialize an instance of the configured class from the filepath stored in the data of the node `/<namespace>/path`. The filepath is expected to be XML, JSON, or YAML, and must conform to one of the following URI schemes:
+* A URL: `http://path/to/file` or `https://path/to/file`
+* An HDFS file: `hdfs://path/to/file`
+* A local file: `file://path/to/file` or `/path/to/file`
+ 
+If an instance of the class is successfully deserialized from the file, it will be validated against any configured object validators. Afterward it will be provided to all subscribers that have subscribed to the publisher via `ZkObjectPublisher.subscribeToUpdates(Consumer)`. The status of any triggered attempt will be recorded under the node `/<namespace>/attempts/<serverIpAddress>`. Upon a success, the children of that node will follow the structure:
+
+```text
+/status # The data will be SUCCESS
```
ivakegg: Loving the feedback mechanism.  Perhaps a little overkill but works

web-services/query/src/main/java/datawave/webservice/query/limit/ActiveQueryTracker.java
```diff
@@ -238,30 +211,29 @@ public int getTotalSystemQueriesForQueryLogic(String system, String queryLogic)
...
      * @param path
      *            the node path
      * @return the total children
      * @throws Exception
      *             if an error occurs while scanning nodes
      */
-    private int getTotalChildrenWithLock(String path) throws Exception {
-        clientLock.lock();
+    private int getTotalChildrenWithLock(CuratorFramework client, String path) throws Exception {
         try {
```
ivakegg, line 238: Why not keep the locking mechanism in here?  Lock is in the name afterall.

web-services/query/src/main/java/datawave/webservice/query/limit/QueryLimitConfigurationValidator.java
```diff
@@ -0,0 +1,14 @@
...
+import com.google.common.base.Preconditions;
+
+import datawave.webservice.zookeeper.ObjectValidator;
+
+public class QueryLimitConfigurationValidator implements ObjectValidator {
+
+    @Override
+    public void validate(Object object) {
+        Preconditions.checkArgument((object instanceof QueryLimitConfiguration), "Object must be an instance of " + QueryLimitConfiguration.class.getName());
+        QueryLimitConfigurationValidationUtils.validate((QueryLimitConfiguration) object);
```
ivakegg: Why have a utils class if this is the only class to use it ?

web-services/query/src/main/java/datawave/webservice/query/limit/ActiveQueryTracker.java
lbschanno: The separate locking mechanism is no longer needed with the use of a client from LockedZkClientDispatcher, but I agree that the method name is now confusing. I will delete this method and directly call `getTotalChildren(CuratorFramework client, String path)` instead.

web-services/query/src/main/java/datawave/webservice/query/limit/QueryLimitConfigurationValidator.java
lbschanno: `QueryLimitConfigurationValidationUtils` is also used in `QueryLimiter.updateConfiguration()` when the argument `validationRequired` is true. This will get used if the initial `QueryLimiter` created by Spring has a configured `QueryLimitConfiguration` that needs to be validated. 

The class `QueryLimitConfigurationValidator` allows us to hook in validation of any updated configurations within the configured `ZkObjectPublisher` and capture any violations in the feedback mechanism.

I was debating between having specific configurable object validators for `ZkObjectPublisher` instances, or perhaps having the validation occur when the configuration is supplied to the QueryLimiter's subscriber. I landed on the former to allow for extra flexibility, but I am open to other approaches.

web-services/query/src/main/java/datawave/zookeeper/README.md
lbschanno: Thanks!

web-services/query/src/main/java/datawave/zookeeper/README.md
lbschanno: Thank you!

web-services/query/src/main/java/datawave/webservice/zookeeper/LockedZkClientDispatcher.java
lbschanno: Glad you like it.

web-services/query/src/main/java/datawave/webservice/query/limit/ActiveQueryTracker.java
lbschanno: This is now complete.

PR #3552, opened 2026-05-11
Remove empty microservice

jalphonso:
Removes `microservices/services/map`


apmoriarty (approved):
Fare thee well, map service

alerman (approved):

PR #3553, opened 2026-05-11
fix: hazelcast cache inspector test

jalphonso:
narrow component scanning and enable caching


apmoriarty (approved):
Always a fan of spring best practices

alerman (approved):

PR #3562, opened 2026-05-19
add sorted tablet location partitioner

hlgp:


warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/partition/SortedTabletLocationPartitioner.java
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.hadoop.io.Text;
+import org.apache.log4j.Logger;
+
+import com.google.common.collect.TreeMultimap;
+
+import datawave.ingest.mapreduce.job.SplitsFile;
+
+public class SortedTabletLocationPartitioner extends MultiTableRangePartitioner {
+
+    private static final Logger log = Logger.getLogger(SortedTabletLocationPartitioner.class);
```
avgAGB, line 19: would it be worthwhile to use the slf4j loggerfactory initialization for the logger here, or is it basically a wash? If so then you can disregard.

avgAGB (approved):

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/partition/SortedTabletLocationPartitioner.java
apmoriarty: +1 to SLF4J

apmoriarty (approved):

PR #3563, opened 2026-05-20
Integrate ScannerBuilder into index lookups

apmoriarty:
Add ExecutionHintHelper to bridge the map of hints in the query logic to scanner specific configs

ScanBuilder is now opinionated about setting null execution hints and consistency levels and will throw an IllegalArgumentException

AsynIndexLookup now tracks ScanBuilder as a protected variable so unit tests can verify execution hints and consistency levels

AsynIndexLookup tests now validate execution hints and consistency level

Fix bug where new index lookups were not setting the correct execution hints


avgAGB (approved):

dws4 (approved):
LGTM

PR #3565, opened 2026-05-20
Fix compose-tests stability: quickstart startup race and missing Hadoop conf path

### LINKED ISSUE #3564: compose-tests: startup race + missing Hadoop conf path in quickstart container
The compose-tests workflow (`.github/workflows/compose-tests.yml`) is unstable due to two structural issues in `docker/docker-compose.yml` and the quickstart Dockerfile. Both reproduce locally and are visible in CI logs.

## 1. Audit (and other ZK/Hadoop-dependent services) race the quickstart container

The compose-tests job intermittently fails with:

```
Container docker-audit-1  Error
dependency failed to start: container docker-audit-1 exited (1)
```

The audit container's Spring `AccumuloAuditor` bean exits 1 with:

```
Caused by: java.lang.RuntimeException: Failed to connect to zookeeper
(quickstart:2181) within 2x zookeeper timeout period 30000
```

**Root cause:** nine services in `docker/docker-compose.yml` use `ZOOKEEPER_HOST` or `HADOOP_HOST` env vars (i.e. depend on the `quickstart` container, which hosts ZK / HDFS / Accumulo) but did not declare a `depends_on: quickstart` relationship. Only `metrics` declared it correctly. When quickstart's cold start (~2-5 min) exceeds the 60s ZK-connect timeout of dependent services, they fail Spring init and exit. docker-compose then aborts with the dependency-failed message above.

**Failing CI runs that show this pattern:**
- PR #3563 — [run 26168060822 / job 76977800154](https://github.com/NationalSecurityAgency/datawave/actions/runs/26168060822/job/76977800154?pr=3563)
- PR #3530 — [run 26176579499 / job 77008277245](https://github.com/NationalSecurityAgency/datawave/actions/runs/26176579499/job/77008277245?pr=3530)
- PR #3532 — [run 26118740680 / job 76814869030](https://github.com/NationalSecurityAgency/datawave/actions/runs/26118740680/job/76814869030?pr=3532) — different symptom (`Query and/or Executor Services never became ready`) but compatible — a downstream service in the chain lost the same race.

The strongest single piece of evidence is PR #3563 itself: the same workflow ran twice in the same job (with and without snapshots). Same code, same time. One job passed, one failed. That is a timing race by definition.

## 2. Accumulo TabletServer scans error with `FileNotFoundException: /etc/hadoop/conf/core-site.xml`

When compose-up succeeds and tests run, `docker/scripts/testAll.sh` consistently fails 8 of 13 tests with "Returned 0 events". The failing tests are the ones that fetch events from the shard table; the passing 5 only scan the field index.

The Accumulo TabletServer log inside `docker-quickstart-1` shows every event-table scan fails server-side:

```
ERROR: error parsing conf file:/etc/hadoop/conf/core-site.xml
java.io.FileNotFoundException: /etc/hadoop/conf/core-site.xml (No such file or directory)
ERROR: Internal error processing startScan
```

**Root cause:** datawave's server-side query iterators load Hadoop config from `/etc/hadoop/conf` when they run inside the Accumulo TabletServer. That path does not exist inside the quickstart container — Hadoop is installed under `/opt/datawave/contrib/datawave-quickstart/hadoop` and `HADOOP_CONF_DIR` points there. Without the path, every event scan errors out and the query service silently returns 0 events.

## Fix

The accompanying PR adds:

1. `contrib/datawave-quickstart/docker/Dockerfile` — symlink `/etc/hadoop/conf -> $HADOOP_CONF_DIR` so iterator scans find the expected config.
2. `docker/docker-compose.yml` — declare `depends_on: quickstart (condition: service_healthy)` for the 9 services that connect to ZK/Hadoop.

Verified locally: `docker/scripts/testAll.sh` goes from 5/13 to 13/13 passing, and compose-up serializes correctly behind the quickstart container.

SethSmucker:
## Summary

- **Dockerfile**: symlink `/etc/hadoop/conf -> $HADOOP_CONF_DIR` in the quickstart image so Accumulo TabletServer-side datawave iterators find the Hadoop config they expect.
- **docker-compose.yml**: declare `depends_on: quickstart (condition: service_healthy)` for the 9 services (annotation, accumulo, audit, dictionary, modification, query, mapreduce-query, executor-pool1, executor-pool2) that connect to ZooKeeper or Hadoop, matching the existing pattern in `metrics`.

Fixes #3564.

See the issue for the full evidence (3 failing CI runs analyzed + local stack-trace captures from the TabletServer and audit containers).

## Test plan

- [x] **Cold local rebuild** with both fixes applied: `docker/scripts/testAll.sh` reports `Tests Passed: 13 / Failed Tests: 0` (was 5/13 prior to either fix).
- [x] **Compose startup ordering** verified end-to-end: `docker-quickstart-1 Healthy` now precedes `docker-audit-1 Started`. Audit's previous ZK-connect retries are gone from the audit container log.
- [x] **`/etc/hadoop/conf` symlink** confirmed inside a freshly-built quickstart container.
- [x] CI run on this PR with this branch as evidence that the workflow no longer hits the `audit exited (1)` flake.


microservices/services/mapreduce-query/jobs/core/pom.xml
```diff
@@ -20,7 +20,7 @@
     <properties>
         <version.datawave>7.33.1</version.datawave>
         <version.datawave.mapreduce-layout-factory>1.0.0</version.datawave.mapreduce-layout-factory>
-        <version.datawave.starter-query>1.0.10</version.datawave.starter-query>
+        <version.datawave.starter-query>1.0.11-SNAPSHOT</version.datawave.starter-query>
```
alerman: We should release starter-query and only merge this after release so we arent merging in dependencies on a snapshot

ivakegg: Are there changes in starter-query that are required for this PR ?

SethSmucker: @ivakegg - changes from 3560 removed eventPerDayThreshold stuff, but the jar wasn't updated so it keeps looking for that property but can never resolve it. 1.0.11-SNAPSHOT is just to include those changes that were made but not reflected in the current jar. So it's more so that the changes were fine, we just need to bump the jar. 

@alerman Agreed, I have the snapshot in here just to make sure that all of the tests actually pass and I diagnosed the issues accurately. So we can remove the snapshot from this PR, but the other changes should still probably be made since that was another chunk of issues that we get when running the CI tests.

PR #3576, opened 2026-06-01
Avoid tserver side unique transformed based on fields selected

### LINKED ISSUE #3568: Avoid server side unique transform based on fields selected
The unique functionality can be done with or without the tserver side unique transform.  It is not uncommon for users to use the unique transform on an ID field to ensure that only one example for each ID is actually returned.  When that is the case, apply the unique transform on the tserver side is not very useful and is simply an extra drain on memory and processing.  We need to be able to configure a set of field names via the QueryLogicFactory.xml for the ShardQueryLogic that will avoid doing the unique transform server side if only those fields are being used.

lbschanno:
The unique functionality can be done with or without the tserver side transform. It is not uncommon for users to use the unique transform on an ID field to ensure only one example for each ID is actuall returned. When this is the case, it is not necessary to apply the unique transform on the tserver side and a drain on memory and processing.

Add the variable `webserverOnlyUniqueFields` as a configurable set to ShardQueryLogic that can be set via the QueryLogicFactory.xml. If a query contains uniquing with fields that are all found in the set `webserverOnlyUniqueFields`, then the unique transform will not be applied on the tserver side.

Closes #3568


warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
```diff
@@ -3437,7 +3445,8 @@ public int hashCode() {
                 isUseTruncatedIndex(),
                 getTruncatedIndexTableName(),
                 getOriginalJexlQuery(),
-                getAllHitsQueryConfig()
+                getAllHitsQueryConfig(),
+                    getWebserverOnlyUniqueFields()
```
avgAGB: a non issue, but it looked like the spacing came out a little funny here

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -710,11 +710,17 @@ private void configureIterator(ShardQueryConfiguration config, IteratorSetting c
...
+                if (config.getWebserverOnlyUniqueFields().isEmpty()
+                                || !config.getWebserverOnlyUniqueFields().containsAll(config.getUniqueFields().getFields())) {
+                    addOption(cfg, QueryOptions.UNIQUE_FIELDS, config.getUniqueFields().toString(), true);
+                    if (config.getUniqueFields().isMostRecent()) {
+                        // this may be redundant with the uniqueFields.toString(), but other code relies on this explicitly being set
+                        addOption(cfg, QueryOptions.MOST_RECENT_UNIQUE, Boolean.toString(true), false);
+                        addOption(cfg, QueryOptions.UNIQUE_CACHE_BUFFER_SIZE, Integer.toString(config.getUniqueCacheBufferSize()), false);
+                    }
+                } else {
+                    log.debug("Query contains unique with webserver-only fields. Unique transform will be skipped on the tserver side.");
```
avgAGB:
```suggestion
if (log.isDebugEnabled()) {
                    log.debug("Query contains unique with webserver-only fields. Unique transform will be skipped on the tserver side.");
                    }
```

warehouse/query-core/src/main/java/datawave/query/config/ShardQueryConfiguration.java
lbschanno: Fixed.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -710,11 +710,19 @@ private void configureIterator(ShardQueryConfiguration config, IteratorSetting c
...
-                    addOption(cfg, QueryOptions.UNIQUE_CACHE_BUFFER_SIZE, Integer.toString(config.getUniqueCacheBufferSize()), false);
+                // If the unique fields contain any fields that are not webserver-only fields, we need to perform the unique transform on the tserver side.
+                if (config.getWebserverOnlyUniqueFields().isEmpty()
+                                || !config.getWebserverOnlyUniqueFields().containsAll(config.getUniqueFields().getFields())) {
+                    addOption(cfg, QueryOptions.UNIQUE_FIELDS, config.getUniqueFields().toString(), true);
+                    if (config.getUniqueFields().isMostRecent()) {
+                        // this may be redundant with the uniqueFields.toString(), but other code relies on this explicitly being set
+                        addOption(cfg, QueryOptions.MOST_RECENT_UNIQUE, Boolean.toString(true), false);
+                        addOption(cfg, QueryOptions.UNIQUE_CACHE_BUFFER_SIZE, Integer.toString(config.getUniqueCacheBufferSize()), false);
+                    }
```
apmoriarty, line 724: I'm trying to think through the case where the unique function is applied to multiple fields, one of which is a high cardinality "skippable" field. 

For example `#UNIQUE(ID, COLOR)` where ID is high cardinality (virtually every instance is unique) and then COLOR is low cardinality (values 'red', 'green', 'blue').

In that case it makes sense still apply the unique transform on the tablet server for just the field COLOR...I think. Thoughts @ivakegg ?

I think what I'm describing is outside the scope of this ticket, so I'll approve and let you all decide if it's worth writing a new issue.

apmoriarty (approved):

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
lbschanno: If `COLOR` is not part of the `webserverOnlyUniqueFields` set, then the unique transform would still get applied on the tserver for `ID` and `COLOR`. Are you thinking of a potential scenario where we would apply a unique transform for a subset of the uniqued fields on the tablet server?

avgAGB (approved):

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
hlgp: this is also the strategy I had in mind.  regardless of the combination of fields, only apply unique on each in their appropriate locations

lbschanno: I think that would be outside the scope of this ticket, and better off as a new issue.

hlgp: I don't think it's outside the scope--I think it's the desired behavior requested, but @ivakegg can confirm/deny

PR #3580, opened 2026-06-03
Update versions due to failed release. Add logic to ignore modifier versions when running

alerman:


apmoriarty (approved):

.github/workflows/microservice-cascade-release.yml
```diff
@@ -137,8 +138,110 @@ jobs:
...
+                  with open(os.path.join(path, 'pom.xml')) as f:
+                      content = f.read()
+                  raw = read_pom_version(content)
+                  if not raw:
+                      raise ValueError('no version')
+              except Exception:
+                  projects.append((path, aid, '???', '???', False, content if 'content' in dir() else ''))
+                  continue
+
+              base = raw.replace('-SNAPSHOT', '')
```
ivakegg, line 201: Isn't there a maven command which will do this type of thing?  It was used in one of the microservice tests that we do on merge/pull requests.  Not saying this is wrong just wondering if you at least looked at that.

ivakegg: I am probably just not getting my head around all of the stuff here but I am perfectly willing to give it the checkmark and we can update later if any changes are needed.

ivakegg (approved):

.github/workflows/microservice-cascade-release.yml
alerman: Yes there is a maven command that does it, and we use that in the actual release process. When running the dry run, i wanted to compute all the things that will change for a quick check without needing to pull dependencies. In order to use the maven command, id have to at least locally go through and run a mvn install on each project in sequence so this was just a way to make the dry run take minutes vice hours.

PR #3582, opened 2026-06-04
Migrate ColumnFamilyConstants to core/tables

apmoriarty:
Copy ColumnFamilyConstants to MetadataColumnFamilyConstants, deprecate old

### DIFF
--- a/core/tables/src/main/java/datawave/table/constants/MetadataColumnFamilyConstants.java
+++ b/core/tables/src/main/java/datawave/table/constants/MetadataColumnFamilyConstants.java
@@ -1,12 +1,11 @@
-package datawave.data;
+package datawave.table.constants;
 
 import org.apache.hadoop.io.Text;
 
 /**
- *
+ * Reserved ColumnFamily names for the DatawaveMetadata table
  */
-@Deprecated(forRemoval = true, since = "7.40.0")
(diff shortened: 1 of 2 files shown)


avgAGB (approved):

drewfarris:
This may not be something you wish to tackle here, but I want to point out that using `Text` objects as constants can be problematic because they are *mutable*. It would be really cool to create a utility subclass of `Text` named something like `ImmutableText` that overrides each of the methods that can be used to mutate `Text` state. It maybe as simple as overriding the `set()` method(s).

An alternate would be to use Strings here, but I think that defeats the purpose of what you're trying to achieve?

apmoriarty:
> This may not be something you wish to tackle here, but I want to point out that using `Text` objects as constants can be problematic because they are _mutable_. It would be really cool to create a utility subclass of `Text` named something like `ImmutableText` that overrides each of the methods that can be used to mutate `Text` state. It maybe as simple as overriding the `set()` method(s).
> 
> An alternate would be to use Strings here, but I think that defeats the purpose of what you're trying to achieve?

Great point. Once everything is migrated I would like to standardize the constants. In some cases we use the Text object, in others the String, and still other places we use Text.toString().

foster33 (approved):

PR #3595, opened 2026-06-09
Refactor Ivarator directory validation

apmoriarty:
- BufferedFileBackedSortedMapTest now tests a Map instead of a Set
- tests now prefer static imports
- duplicate directory validation logic now lives in IvaratorCacheDir
- updated FsStatus constructor to one that accepts a path

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/iterator/ivarator/IvaratorCacheDir.java
+++ b/warehouse/query-core/src/main/java/datawave/query/iterator/ivarator/IvaratorCacheDir.java
@@ -1,8 +1,15 @@
 package datawave.query.iterator.ivarator;
 
+import java.io.IOException;
+
 import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.FsStatus;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class IvaratorCacheDir {
+
+    private static final Logger log = LoggerFactory.getLogger(IvaratorCacheDir.class);
 
     final protected IvaratorCacheDirConfig config;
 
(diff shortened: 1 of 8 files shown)


warehouse/query-core/src/test/java/datawave/query/util/sortedmap/HdfsBackedSortedMapTest.java
```diff
@@ -130,125 +134,125 @@ public void persistCompactReloadTest() throws Exception {
...
         // Ensure that we have 3 entries total
-        Assert.assertEquals(3, compactedSortedSet.size());
+        assertEquals(3, compactedSortedMap.size());
 
         // This is what we expect to be loaded by the set
         results.clear();
-        results.add(someTestString);
-        results.add(anotherTestString);
-        results.add(lastTestString);
+        results.add(new AbstractMap.SimpleEntry<>("key1", someTestString));
```
jschmidt10: A little bit cleaner in java 9 or 11, they added: `Map.entry("key1", someTestString);`

jschmidt10 (dismissed):

warehouse/query-core/src/main/java/datawave/query/iterator/ivarator/IvaratorCacheDir.java
```diff
@@ -26,6 +34,41 @@ public FileSystem getFs() {
         return fs;
     }
 
+    public FsStatus getFsStatus() {
+        try {
+            Path path = new Path(pathURI);
+            return getFs().getStatus(path);
+        } catch (IOException e) {
+            log.warn("Unable to determine status of the filesystem: {}", getFs());
```
dtspence, line 48: Do you think we should also include the exception in the message?

avgAGB (dismissed):

warehouse/query-core/src/main/java/datawave/query/iterator/ivarator/IvaratorCacheDir.java
```diff
@@ -26,6 +40,42 @@ public FileSystem getFs() {
...
+            log.warn("Exception : {}", e.getMessage());
```
dtspence, line 49: Minor question, is stack trace important here?

dtspence (dismissed):

warehouse/query-core/src/main/java/datawave/query/iterator/ivarator/IvaratorCacheDir.java
apmoriarty: I don't think so. We know where in the code this is called, and if this ends up being a benign warning then we don't want to pollute logs with stack traces. Benign meaning several directories were configured, the first one failed validation so the ivarator moved to the second directory.

ivakegg (dismissed):

hlgp (dismissed):

avgAGB (approved):

ivakegg (approved):

PR #3597, opened 2026-06-10
Specify UTF-8 explicitly at ingest String.getBytes() sites

### LINKED ISSUE #3596: Ingest: String.getBytes() uses the platform default charset
Ingest-side code calls `String.getBytes()` (and other default-charset overloads)
without specifying a charset, so it picks up the JVM platform default. In a
container whose default isn't UTF-8, non-ASCII field values are silently mangled
on the encode path.

Scope: production `src/main` in the warehouse ingest modules (`ingest-core`,
`core`, `age-off`, `metrics-core`), encode side only. Decode (`new String(byte[])`)
and non-ingest modules are out of scope.

Fix: make every genuine `String` receiver explicit — `getBytes(StandardCharsets.UTF_8)`.
No behavior change for non-`String` receivers (Hadoop `Text`, Accumulo `Value`,
HyperLogLog, etc.).

SethSmucker:
Closes #3596

Pins the charset on the ingest encode path: `String.getBytes()` ->
`getBytes(StandardCharsets.UTF_8)` across the warehouse ingest modules. Only
`String` receivers are touched; Hadoop `Text` / Accumulo `Value` / HyperLogLog
sites are left as-is.

96 call sites across 49 files. No functional change beyond fixing the charset.

### DIFF
--- a/warehouse/age-off/src/main/java/datawave/iterators/filter/ColumnVisibilityTokenizingFilter.java
+++ b/warehouse/age-off/src/main/java/datawave/iterators/filter/ColumnVisibilityTokenizingFilter.java
@@ -1,12 +1,14 @@
 package datawave.iterators.filter;
 
+import java.nio.charset.StandardCharsets;
+
 import org.apache.accumulo.core.data.Key;
 import org.apache.accumulo.core.data.Value;
 
 import datawave.iterators.filter.ageoff.FilterOptions;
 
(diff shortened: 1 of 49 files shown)


apmoriarty (approved):

avgAGB (approved):

SethSmucker:
Superseded by #3599.

PR #3602, opened 2026-06-12
Remove static mutable shared state across concurrent queries

ghost:
DefaultQueryPlanner: Convert 5 static fields to instance fields so concurrent queries with different datatype filters no longer clobber each other's field-to-type metadata caches. The fields queryFieldsAsDataTypeMap, normalizedFieldAsDataTypeMap, cachedIndexedFields, cachedReverseIndexedFields, and cachedNormalizedFields were declared static and written outside any synchronization block in timedFetchDatatypes, causing cross-query data contamination under concurrent load.

MarkingFunctionsFactory: Make markingFunctions volatile and add synchronized to postContruct to match the synchronization already on createMarkingFunctions, preventing a torn read of the static field.

IteratorThreadPoolManager: Track all ScheduledExecutorService instances in a list instead of discarding their references. Add a shutdown method that shuts down all thread pools and scheduled executors so threads are not leaked across query lifecycle cycles.


apmoriarty:
The MarkingFunctionsFactory and IteratorThreadPoolManager changes could be useful, but those should be separate merge requests. Ideally there would be a test that demonstrates the failure condition and then the commit to fix it.

Regarding the static caches in the DefaultQueryPlanner the code comments indicate it is only enabled for developer testing, and no configs in the project enable that feature. Ideally that feature would be removed altogether because a shared datatype cache is not great design.

Lastly, do not pick up massive commits (i.e., the markings pojo feature) as part of your merge request. I am going to close this merge request not because the work isn't valuable, but because there are too many changes happening at once. Feel free to open separate merge requests for each of the changes.

ghost:
> The MarkingFunctionsFactory and IteratorThreadPoolManager changes could be useful, but those should be separate merge requests. Ideally there would be a test that demonstrates the failure condition and then the commit to fix it.
> 
> Regarding the static caches in the DefaultQueryPlanner the code comments indicate it is only enabled for developer testing, and no configs in the project enable that feature. Ideally that feature would be removed altogether because a shared datatype cache is not great design.
> 
> Lastly, do not pick up massive commits (i.e., the markings pojo feature) as part of your merge request. I am going to close this merge request not because the work isn't valuable, but because there are too many changes happening at once. Feel free to open separate merge requests for each of the changes.

Thank you for the feedback I appreciate you taking the time. I'll admit my contribution etiquette is something I need work on. I'll write up that failing test & fix for both the MarkingFunctionsFactory and the IteratorThreadPoolManager separately, along with ditching the other changes and referencing in new, individual pr's. Thanks again.

PR #3620, opened 2026-06-18
Bugfix/configcheck

billoley:
these test failures were probably missed because configcheck by default doesn't get built and tests run


apmoriarty (approved):
Looks like configs I added but did not add here.

avgAGB (approved):

PR #3627, opened 2026-06-23
Reduce heap size needed by PrintUtility

### LINKED ISSUE #3248: PrintUtility forces us to allocate 2g heaps for unit tests
In the `PrintUtility` that is used to print Accumulo table contents in the context of several tests, there's a method named `getPrintableValue` that will attempt to decode an Accumulo `Key` and `Value` using different strategies that are common to Datawave table structures.

I was attempting to use PrintUtility in a new test related to #3173 and consistently encountered out of memory exceptions from [L146 in getPrintableValue](https://github.com/NationalSecurityAgency/datawave/blob/810570500ad94e7f9021f41b335f2f2ed7e7d2eb/warehouse/query-core/src/test/java/datawave/helpers/PrintUtility.java#L146). 

Upon investigation I observed that the `HyperLogLogPlus` class was falling back to the `legacytDecode` method. This was always allocating a large byte array because it was attempting to interpret a byte[] that wasn't an encoded hyper log log data structure. The size of the array was larger than 1G, but smaller than 2G.

I was able to work around this issue by increasing the heap size for unit tests from 1G to 2G as [seen here](https://github.com/NationalSecurityAgency/datawave/blob/a2f71b68fb14abe896134b0d6bcda9492cd26f02/web-services/pom.xml#L565). Not a showstopper, but could be worth looking into.

lbschanno:
The PrintUtility class forces us to allocate 2g heaps for unit tests due to attempting to decoded non-HyperLogLogPlus cardinality values as such. The HyperLogLogPlus.legacyDecode() method will at times allocate a very large byte array larger than 1g.

Add the class TableWriter to handle writing entries of a table to an output. This class can be configured with custom key, timestamp, and Value formatters. Added implementations of TableWriter for individual table types that customized how timestamps and Values are formatted.

Add methods to PrintUtility that handle writing entries for specific tables, such as shard, metadata, facet, etc. This allows for targeted formatting to avoid accidentally decoding a non-HyperLogLogPlus cardinality value as such.

Update all usages of PrintUtility to call the new methods, and reduce the allocated heap size for unit tests to 1g.

Fixes #3248


apmoriarty (dismissed):

warehouse/query-core/src/test/java/datawave/helpers/Output.java
```diff
@@ -0,0 +1,77 @@
...
+
+    void write(String line);
+
+    default void writeln(String line) {
+        write("\n" + line);
+    }
+
+    void flush();
+
+    class ApacheLog4JOutput implements Output {
```
avgAGB: Curious, would this interface still work if we used the slf4j packages instead? Maybe I've developed a pavlovian response to seeing the stock l4j jars in the repo

lbschanno: Yes, the interface still works. Just pushed an update to use slf4j instead.

avgAGB (dismissed):

apmoriarty (dismissed):

warehouse/query-core/src/test/java/datawave/helpers/PrintUtility.java
```diff
@@ -2,167 +2,424 @@
...
+     * @param auths
+     *            the auths to use when scanning over the table
+     * @throws TableNotFoundException
+     *             if the table could not be found
+     * @see #printTableToDebugLog(AccumuloClient, Authorizations, String, TableWriter)
      */
-    private PrintUtility() {
-        // Nothing to do
+    public static void printShardTable(final AccumuloClient client, final Authorizations auths) throws TableNotFoundException {
+        printShardTable(client, auths, TableName.SHARD_INDEX);
```
apmoriarty: TableName.SHARD

warehouse/query-core/src/test/java/datawave/query/function/HitsAreAlwaysIncludedTest.java
```diff
@@ -115,9 +114,9 @@ public static void beforeAll() throws Exception {
         // internally to derive the other shard index table variants (NO_UID_INDEX, TRUNCATED_INDEX, etc.)
         // that AbstractQueryTest.planAndExecuteQuery() iterates over.
         LimitFieldsTestingIngest.writeItAll(clientForTest, LimitFieldsTestingIngest.WhatKindaRange.DOCUMENT);
-        PrintUtility.printTable(clientForTest, auths, TableName.SHARD);
-        PrintUtility.printTable(clientForTest, auths, TableName.SHARD_INDEX);
-        PrintUtility.printTable(clientForTest, auths, QueryTestTableHelper.MODEL_TABLE_NAME);
+        PrintUtility.printShardTable(clientForTest, auths);
+        PrintUtility.printShardIndexTable(clientForTest, auths);
+        PrintUtility.printMetadataTable(clientForTest, auths);
```
apmoriarty, line 119: Note for other reviewers: the QueryTestTableHelper gets standard table names wrong, this looks like a functional change but is actually an alignment towards standard conventions.

warehouse/query-core/src/test/java/datawave/query/jexl/functions/GroupingRequiredFilterFunctionsIntegrationTest.java
```diff
@@ -107,9 +106,9 @@ public static void beforeAll() throws Exception {
         // ingest with the document range only; GroupingFiltersIngest already calls IndexIngestUtil
         // internally to derive the other shard index table variants that AbstractQueryTest iterates over.
         GroupingFiltersIngest.writeItAll(clientForTest, GroupingFiltersIngest.Range.DOCUMENT);
-        PrintUtility.printTable(clientForTest, auths, TableName.SHARD);
-        PrintUtility.printTable(clientForTest, auths, TableName.SHARD_INDEX);
-        PrintUtility.printTable(clientForTest, auths, QueryTestTableHelper.MODEL_TABLE_NAME);
+        PrintUtility.printShardTable(clientForTest, auths);
+        PrintUtility.printShardIndexTable(clientForTest, auths);
+        PrintUtility.printMetadataTable(clientForTest, auths, QueryTestTableHelper.MODEL_TABLE_NAME);
```
apmoriarty: Change to TableName.METADATA

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/ExecutableExpansionVisitorTest.java
```diff
@@ -111,10 +110,10 @@ public static void beforeAll() throws Exception {
...
 
         WiseGuysIngest.writeItAll(sharedClient, WiseGuysIngest.WhatKindaRange.DOCUMENT);
-        PrintUtility.printTable(sharedClient, auths, TableName.SHARD);
-        PrintUtility.printTable(sharedClient, auths, TableName.SHARD_INDEX);
-        PrintUtility.printTable(sharedClient, auths, QueryTestTableHelper.METADATA_TABLE_NAME);
-        PrintUtility.printTable(sharedClient, auths, QueryTestTableHelper.MODEL_TABLE_NAME);
+        PrintUtility.printShardTable(sharedClient, auths);
+        PrintUtility.printShardIndexTable(sharedClient, auths);
+        PrintUtility.printMetadataTable(sharedClient, auths, QueryTestTableHelper.METADATA_TABLE_NAME);
+        PrintUtility.printMetadataTable(sharedClient, auths, QueryTestTableHelper.MODEL_TABLE_NAME);
```
apmoriarty: Merge into a single metadata table call?

web-services/pom.xml
```diff
@@ -543,7 +543,7 @@
                         <threadCountClasses>0</threadCountClasses>
                         <threadCountMethods>1</threadCountMethods>
                         <threadCountSuites>0</threadCountSuites>
-                        <argLine>@{argLine} -XX:+TieredCompilation -XX:TieredStopAtLevel=1  -Dfile.encoding=UTF8 -Duser.timezone=GMT -Xmx2g -Dapple.awt.UIElement=true -Djava.security.krb5.realm= -Djava.security.krb5.kdc=
```
apmoriarty, line 546: Is this change from 2g to 1g intentional?

warehouse/query-core/src/test/java/datawave/helpers/PrintUtility.java
lbschanno: D'oh. Fixed.

warehouse/query-core/src/test/java/datawave/query/jexl/functions/GroupingRequiredFilterFunctionsIntegrationTest.java
lbschanno: Fixed.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/ExecutableExpansionVisitorTest.java
lbschanno: After looking at WiseGuysIngest.writeItAll(), it looks it only writes metadata/model info to the "DatawaveMetadata" table. I'll collapse these to a single call.

web-services/pom.xml
lbschanno: Yes. We no longer need the heap size of 2g since we are only decoding hyperlog values for facet tables specifically, where it is appropriate to do so. We originally needed 2g when we were potentially decoding  non hyperlog vlaues as such, as noted in #3248.

warehouse/query-core/src/test/java/datawave/query/jexl/visitors/ExecutableExpansionVisitorTest.java
lbschanno: Collapsed.

Issue #2916, opened 2025-05-20
Iterator cleanup in document scheduler


apmoriarty:
Iterators were added over time to the document scheduler and there's some tech debt to pay down

- extract common code to core classes
- improve precondition checks to avoid NPEs
- examine control flow (handling miss and then nexting)
- validate log thresholds
- look at extracting common option parsing to core class
- only add query iterator options when executing the query iterator during a document range scan

PR #3632, opened 2026-06-23
Fix consistency-level and precondition bugs in the document scheduler

SethSmucker:
Part of #2916

Two bug fixes in the document scheduler:
* DocumentRangeScan (retrieval path) checked the search consistency level but set the retrieval one, so a configured level was ignored and a null one could throw. Now it checks and sets retrieval, like the search path in DocumentIdProducer.
* QueryDataConsumer logged the query data before its null check. Moved the null check first.


warehouse/query-core/src/main/java/datawave/next/scanner/DocumentRangeScan.java
```diff
@@ -160,7 +160,7 @@ private Scanner createScanner() throws Exception {
             scanner.setExecutionHints(Map.of("scan_type", config.getRetrievalScanHintPool()));
         }
 
-        if (config.getSearchConsistencyLevel() != null) {
+        if (config.getRetrievalConsistencyLevel() != null) {
```
apmoriarty, line 163: Nice catch

warehouse/query-core/src/main/java/datawave/next/scanner/QueryDataConsumer.java
```diff
@@ -61,24 +61,19 @@ public void run() {
...
-                    log.info("query data settings was null");
-                }
-                if (queryData.getRanges() == null) {
-                    log.info("query data ranges was null");
-                }
                 stats.incrementQueryDataSeen();
 
-                Preconditions.checkArgument(queryData.getSettings().size() == 1);
-                Preconditions.checkArgument(queryData.getRanges().size() == 1);
+                Preconditions.checkArgument(queryData.getSettings() != null && queryData.getSettings().size() == 1, "expected exactly one settings entry");
```
apmoriarty: Looking at this code again with a fresh set of eyes, I think I would prefer something like this: 
```java
Preconditions.checkNotNull(queryData.getSettings(), msg);
Preconditions.checkArgument(queryData.getSettings().size() == 1);

Preconditions.checkNotNull(queryData.getRanges(), msg)
Preconditions.checkArgument(queryData.getRanges().size() == 1);
```

However, that is a personal preference and does not block a review

apmoriarty (dismissed):

foster33 (dismissed):

foster33 (approved):

apmoriarty (approved):

PR #3640, opened 2026-06-25
Fix discrepancy with ambiguously OR'd quoted terms

### LINKED ISSUE #3636: Query Validator discrepancy when using double quotes
When inputting
`(BODY:term1 OR term2 OR term3)`
the query validator correctly  suggests the proper parentheses.
However when inputting
`(BODY:"term1" OR "term2" OR "term3")`
the query validator indicates no validation issues.
These two inputs should be producing the same response

lbschanno:
The AmbiguousOrPhrasesRule will return recommendations for correcting cases where a query is `(BODY:term1 OR term2 OR term3)`. However, the rule does not return the necessary recommendations when the terms are quoted, such as `(BODY:"term1" OR "term2" OR "term3")`.

Update the AmbiguousUnfieldedTermsVisitor to also check quoted terms when evaluating a query.

Fixes #3636


apmoriarty (approved):
Nice cleanup

avgAGB (approved):

hlgp (approved):

PR #3646, opened 2026-06-30
Fix HIT_EXCERPT for grouped term-frequency field names

apmoriarty:
## Problem

`HIT_EXCERPT` does not consistently populate. The root cause is #3487 ("Preserve token context for term frequency entries to avoid false positive hits across multiple field instances"), which introduced grouping/dot notation in the field-name portion of term-frequency keys (e.g. `CONTENT.A1B2C3`) and updated TF scanning, content-function evaluation, offset population, and `isExcerptField` to match grouped fields — but the excerpt generation path was missed and still matched field names exactly:

- `TermFrequencyExcerptIterator.next()` used `fieldName.equals(fieldAndValue[0])`, so grouped TF entries were skipped and the excerpt came back blank.
- `ExcerptTransform.getOffset()` seeked only the exact TF key, so hit-term offsets for grouped TFs could not be resolved.

As a result `HIT_EXCERPT` populated only for documents whose tokens lacked grouping context, and was blank/absent for those that had it — hence the inconsistency.

## Fix

Both spots now accept grouped variants of the requested base field via `JexlASTHelper.isGroupedFieldMatch` (mirroring how #3487 fixed the other matchers):

- `TermFrequencyExcerptIterator.next()` matches the base field or any grouped variant.
- `ExcerptTransform.getOffset()` scans the requested field through all of its grouped variants and considers every matching entry, preserving the original "overlapping position first, else first position" semantics.

## Tests

- `TermFrequencyExcerptIteratorTest#testMatchFoundForGroupedField` — reproduces the blank-excerpt failure for grouped TF entries (fails before the fix, passes after).
- Migrated `ExcerptTransformTest` off EasyMock to a real `TermFrequencyExcerptIterator` backed by a `SortedMapIterator` over real term-frequency data, with real `Document`s. Excerpts are now genuinely generated rather than scripted on a mock, so the tests exercise the real `getOffset` resolution and excerpt generation. Added `testExcerptForGroupedField` to cover the grouped-field path end-to-end through the transform.

All `warehouse/query-core` excerpt unit tests pass (`ExcerptTransformTest`, `TermFrequencyExcerptIteratorTest`, `WordsAndScoresTest`).

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
+++ b/warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
@@ -28,6 +28,7 @@ import com.google.protobuf.InvalidProtocolBufferException;
 
 import datawave.ingest.protobuf.TermWeight;
 import datawave.query.Constants;
+import datawave.query.jexl.JexlASTHelper;
 
 /**
  * This iterator is intended to scan the term frequencies for a specified document, field, and offset range. The result will be excerpts for the field specified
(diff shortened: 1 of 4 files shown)


apmoriarty:
Closing in favor of a branch pushed directly to this repository.

PR #3647, opened 2026-06-30
Fix HIT_EXCERPT for grouped term-frequency field names

apmoriarty:
## Problem

`HIT_EXCERPT` does not consistently populate. The root cause is #3487 ("Preserve token context for term frequency entries to avoid false positive hits across multiple field instances"), which introduced grouping/dot notation in the field-name portion of term-frequency keys (e.g. `CONTENT.A1B2C3`) and updated TF scanning, content-function evaluation, offset population, and `isExcerptField` to match grouped fields — but the excerpt generation path was missed and still matched field names exactly:

- `TermFrequencyExcerptIterator.next()` used `fieldName.equals(fieldAndValue[0])`, so grouped TF entries were skipped and the excerpt came back blank.
- `ExcerptTransform.getOffset()` seeked only the exact TF key, so hit-term offsets for grouped TFs could not be resolved.

As a result `HIT_EXCERPT` populated only for documents whose tokens lacked grouping context, and was blank/absent for those that had it — hence the inconsistency.

## Fix

Both spots now accept grouped variants of the requested base field via `JexlASTHelper.isGroupedFieldMatch` (mirroring how #3487 fixed the other matchers):

- `TermFrequencyExcerptIterator.next()` matches the base field or any grouped variant.
- `ExcerptTransform.getOffset()` scans the requested field through all of its grouped variants and considers every matching entry, preserving the original "overlapping position first, else first position" semantics.

## Tests

- `TermFrequencyExcerptIteratorTest#testMatchFoundForGroupedField` — reproduces the blank-excerpt failure for grouped TF entries (fails before the fix, passes after).
- Migrated `ExcerptTransformTest` off EasyMock to a real `TermFrequencyExcerptIterator` backed by a `SortedMapIterator` over real term-frequency data, with real `Document`s. Excerpts are now genuinely generated rather than scripted on a mock, so the tests exercise the real `getOffset` resolution and excerpt generation. Added `testExcerptForGroupedField` to cover the grouped-field path end-to-end through the transform.

All `warehouse/query-core` excerpt unit tests pass (`ExcerptTransformTest`, `TermFrequencyExcerptIteratorTest`, `WordsAndScoresTest`).


apmoriarty:
A few things still to do here.

warehouse/query-core/src/main/java/datawave/query/iterator/logic/TermFrequencyExcerptIterator.java
```diff
@@ -894,6 +896,21 @@ private static String[] getFieldAndValue(Key tfKey) {
...
+     *
+     * @param requestedField
+     *            the base field name the excerpt was requested for
+     * @param candidateField
+     *            the field name found in the term-frequency key
+     * @return true if the candidate field is the requested field or a grouped variant of it
+     */
+    private static boolean fieldMatches(String requestedField, String candidateField) {
+        return requestedField.equals(candidateField) || JexlASTHelper.isGroupedFieldMatch(requestedField, candidateField);
+    }
```
drewfarris, line 912: Consider extracting `fieldMatches` to a utility class because it is also implemented in `ExcerptTransform`. There is a similar operation in `TermFrequencyIterator`, added in #3487, and another different variant in `TLDTermFrequencyEventDataQueryFilter` too, but these might be out of scope to rectify in this PR, which should focus on the minimal changes required to fix the issue.

warehouse/query-core/src/test/java/datawave/query/iterator/logic/TermFrequencyExcerptIteratorTest.java
```diff
@@ -169,6 +198,31 @@ public void testMatchFound() throws IOException {
...
+        iterator.setHitTermsList(new ArrayList<>(List.of("")));
+        iterator.seek(range, Collections.emptyList(), false);
+
+        assertTrue(iterator.hasTop());
+
+        Key topKey = iterator.getTopKey();
+        assertEquals(row, topKey.getRow());
+        assertEquals(new Text("email" + Constants.NULL + "444.555.666"), topKey.getColumnFamily());
+        assertEquals(new Text("CONTENT" + Constants.NULL + "XXXNOTSCOREDXXX" + Constants.NULL + "annotation indexing makes content grouped" + Constants.NULL
+                        + "XXXNOTSCOREDXXX"), topKey.getColumnQualifier());
```
drewfarris, line 223: Consider using the constant `NOT_SCORED_MARKER` here instead of the string `XXXNOTSCOREDXXX`?

warehouse/query-core/src/test/java/datawave/query/transformer/ExcerptTransformTest.java
```diff
@@ -175,33 +140,26 @@ public void testExcerptOverlapped() throws IOException {
...
-        verifyAll();
 
-        Attributes arg = capturedArg.getValue();
-        assertEquals(1, arg.size());
-        String excerpt = arg.getAttributes().iterator().next().getData().toString();
-        // only one excerpt should return
-        assertEquals("and the [quick] [brown] fox jumped over the lazy dog", excerpt);
+        Set<String> excerpts = getExcerpts();
+        // the hit term's offset (9-10) overlaps the phrase index [10,14], so they merge into a single excerpt. The
+        // longest token at the overlapping offset ("quick brown") is bracketed by the iterator's longest-word rule.
```
drewfarris, line 153: It would be nice if this comment would explain why we get both `[quick]` and `[quick brown]` as output in the excerpt.

apmoriarty:
Closing in favor of https://github.com/NationalSecurityAgency/datawave/pull/3651

PR #3650, opened 2026-07-01
Add tool to identify tablets that need compaction

### LINKED ISSUE #3581: Need to identify tablets requiring forced compactions
We have a situation where we have way too many tablets and we need to merge those down in a datawave shard table.  The problem is that an accumulo merge operation requires a full table lock in accumulo 2.1 (we are currently at 2.1.5).  Moreover within the full table lock the tablets are flushed and compacted which takes time (potentially a long time).  We cannot afford to do this while the system is running (i.e. currently ingesting data).

So the plan is to identify all of the tablets that require compaction, and compact those while the system is active.  Then once all of the tablets of interest are compacted sufficiently such that all data in the RFiles are completely contained within the tablet's extent, we can temporarily stop the system to apply the desired merge operations.

We need a utility that will, for a specified range of tablets in a specified table, determine if any of the files within the tablet contain data outside the tablet's extent/bounds.

Once we have a utility that can do that efficiently then perhaps we can add the ability to fire off a compaction request for the identified tablets.

tom-vt4:
Add a utility tool that is able to identify tablets that need compaction. This tool is hosted in the project check-tablet-extents and can be built as an executable JAR.

Closes #3581


contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
```diff
@@ -0,0 +1,302 @@
...
+        // Examine each file and determine whether any of them would be cleaned up/optimized by a compaction.
+        for (StoredTabletFile file : allFiles) {
+            FileSystem ns = FileSystem.get(file.getPath().toUri(), context.getHadoopConf());
+            try (FileSKVIterator openReader = fileFactory.newReaderBuilder().forFile(file.getPathStr(), ns, ns.getConf(), NoCryptoServiceFactory.NONE)
+                            .withTableConfiguration(tableConf).seekToBeginning().build()) {
+                Key first = openReader.getFirstKey();
+                Key last = openReader.getLastKey();
+
+                // A tablet requires a compaction if any of the following are true:
+                // - The file is empty (first and last are null)
```
ivakegg: Why would an empty file require compaction ?

ivakegg:
In general this looks really good.  I only had one comment in there.  For extra credit I would like for this tool to optionally (based on a parameter) actually kick off the compactions.

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
tom-vt4: I included them in the compaction list because they would be removed after the compaction completes.

This logic references findChopFiles() from the accumulo 2.1 branch class CompactableUtils.java :
accumulo/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableUtils.java

Should the tool ignore empty files and tablets that only need compaction based on having empty RFiles?

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
```diff
@@ -0,0 +1,302 @@
...
+            return s == null ? null : new Text(s);
+        }
+    }
+
+    /**
+     * Represents a set of options that can parsed and used via {@link TabletExtentChecker#main(String[])}.
+     */
+    private static class Opts {
+
+        @Parameter(names = {"-t", "--table"}, description = "The table name", required = true)
```
ivakegg, line 61: I need to be able to supply the username and password via the program arguments, and the password I need to be able to supply via an environment variable.  So something like this: java -cp path/to/jar datawave.TabletExtentChecker -u datawave -p env:PASS -t shard ....

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
tom-vt4: Removed the check for an empty file. Compaction will only trigger on tablets with data outside of their extents.

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
tom-vt4: Added the requested parameters in addition to parameters for specifying the accumulo instance and zookeeper connection string.

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
```diff
@@ -0,0 +1,305 @@
...
+
+import com.beust.jcommander.IStringConverter;
+import com.beust.jcommander.JCommander;
+import com.beust.jcommander.Parameter;
+import com.beust.jcommander.ParameterException;
+
+/**
+ * A utility that will identify tablets that require compaction within a given table for a given range and lists recommended compaction commands.
+ */
+public final class TabletExtentChecker {
```
dlmarion: FYI that if:

  1. you implement KeywordExecutable
  2. create a service file
  3. put this jar in the Accumulo classpath

then it will show up when you run `accumulo -?` and can be run via the `accumulo` command using the correct classpath. An example of how to do this outside of the `accumulo` repo is in the `accumulo-classloaders` repo:

https://github.com/apache/accumulo-classloaders/blob/d12b5456e56b74ca6a48380e5587d93f31886fb8/modules/caching-classloader/src/main/java/org/apache/accumulo/classloader/ccl/cli/CreateManifest.java#L34-L35

tom-vt4: Thank you for the suggestion, I have now implemented this feature.

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
```diff
@@ -0,0 +1,320 @@
...
+            try (FileSKVIterator openReader = fileFactory.newReaderBuilder().forFile(file.getPathStr(), ns, ns.getConf(), NoCryptoServiceFactory.NONE)
+                            .withTableConfiguration(tableConf).seekToBeginning().build()) {
+                Key first = openReader.getFirstKey();
+                Key last = openReader.getLastKey();
+
+                // A tablet requires a compaction if any of the following are true:
+                // - The first key is outside the tablet's extent.
+                // - The last key is outside the tablet's extent.
+                if ((first != null && !extent.contains(first.getRow())) || (last != null && !extent.contains(last.getRow()))) {
+                    return true;
```
ivakegg, line 333: Can we get some logging in here showing the relevant info?  I would love to be able to see what is going on.

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
```diff
@@ -0,0 +1,320 @@
...
+         * @param args
+         *            the arguments to parse.
+         */
+        public void parseArgs(String[] args) {
+            JCommander commander = JCommander.newBuilder().addObject(this).programName(TabletExtentChecker.class.getSimpleName()).build();
+
+            try {
+                commander.parse(args);
+            } catch (ParameterException ex) {
+                commander.usage();
```
ivakegg, line 93: Could we print the error message first before the help.

ivakegg: perhaps with a clear "ERROR: " before the message so that it pops out.  Thanks.

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
```diff
@@ -0,0 +1,320 @@
...
+        @Override
+        public Text convert(String s) {
+            return s == null ? null : new Text(s);
+        }
+    }
+
+    /**
+     * Represents a set of options that can parsed and used via {@link TabletExtentChecker#execute(String[])}.
+     */
+    private static class Opts {
```
dlmarion: With a couple more steps, you can turn this into a server-side utility, meaning that you won't need the args for instance, zookeepers, and username/password.

I would suggest the following:

  1. Extend `ConfigOpts`, then you get `Help` and a `parseArgs` implementation.
  2. Use the `ConfigOpts.getSiteConfiguration()` to create a ServerContext.

To see an example of 1 & 2, look at https://github.com/apache/accumulo/blob/3d34c27538d87c8e6e11879f5f6e9a11d43a159e/server/base/src/main/java/org/apache/accumulo/server/conf/util/ZooInfoViewer.java#L108-L125

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
```diff
@@ -0,0 +1,320 @@
...
+        Set<StoredTabletFile> allFiles = new HashSet<>(tablet.getFiles());
+        final FileOperations fileFactory = FileOperations.getInstance();
+
+        // Examine each file and determine whether any of them would be cleaned up/optimized by a compaction.
+        for (StoredTabletFile file : allFiles) {
+            FileSystem ns = FileSystem.get(file.getPath().toUri(), context.getHadoopConf());
+            try (FileSKVIterator openReader = fileFactory.newReaderBuilder().forFile(file.getPathStr(), ns, ns.getConf(), NoCryptoServiceFactory.NONE)
+                            .withTableConfiguration(tableConf).seekToBeginning().build()) {
+                Key first = openReader.getFirstKey();
+                Key last = openReader.getLastKey();
```
dlmarion, line 320: Suggest looking at `FIleUtil.tryToGetFirstAndLastRows`. You might be able to just use it considering this is becoming a server-side utility and you will have access to it.

https://github.com/apache/accumulo/blob/2.1/server/base/src/main/java/org/apache/accumulo/server/util/FileUtil.java#L484

contrib/datawave-utils/check-tablet-extents/README.md
```diff
@@ -0,0 +1,136 @@
...
+To use the tool with an Accumulo command, copy the shaded jar to the Accumulo lib folder:
+
+```bash
+$ cp path/to/check-tablet-extents-shaded.jar $ACCUMULO_HOME/lib
+```
+
+Then use the following command:
+
+```bash
+$ accumulo check-tablets <options>
```
lbschanno: I would like to see the README go into more detail about how to set this tool up as an Accumulo utility.

lbschanno (changes requested):

contrib/datawave-utils/check-tablet-extents/README.md
lbschanno: Additionally, is using the shaded jar required for the utility tool? Or can we get by with the unshaded jar when setting things up for `accumulo check-tablets`?

dlmarion: My guess is that this project does not need to be shaded and can just depend on the correct accumulo jars. My comments / suggestions above about turning it into a server-side utility may have pointed you all in the wrong direction if that's not what you are trying to achieve. I just assumed.

lbschanno: It depends on how @ivakegg would prefer to use this.

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
tom-vt4: Can you provide some clarification on how the tool would be ran as a server-side utility tool?

contrib/datawave-utils/check-tablet-extents/README.md
lbschanno: I verified that the shaded jar is not needed when using this as an accumulo command.

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
tom-vt4: Resolved in commit 83ececa.

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
tom-vt4: Added extent logging in commit 83ececa.

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
tom-vt4: Reworked the tool to be a server-side utility in commit 83ececa. However, I had to use ClientContext as there were conflicts refactoring the test class to use ServerContext since the test class depends on MiniAccumuloCluster.

contrib/datawave-utils/check-tablet-extents/README.md
tom-vt4: Updated documentation and removed shaded jar dependence in commit 83ececa.

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    @Override
+    public String keyword() {
+        return "check-tablets";
+    }
+
+    @Override
+    public String description() {
+        return "A utility that will identify tablets that require compaction within a given table for a given range and lists recommended compaction commands.";
```
lbschanno: This description will be shown in the list of accumulo command options, so brevity will be a plus here. Recommended update:
```
Identifies tablets with data outside their extents and if specified, compacts them.
```

contrib/datawave-utils/check-tablet-extents/pom.xml
```diff
@@ -0,0 +1,142 @@
...
+            <groupId>com.beust</groupId>
+            <artifactId>jcommander</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.google.auto.service</groupId>
+            <artifactId>auto-service-annotations</artifactId>
+            <version>1.1.1</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
```
lbschanno: This dependency is not used and can be removed.

contrib/datawave-utils/check-tablet-extents/pom.xml
```diff
@@ -0,0 +1,142 @@
...
+            <artifactId>guava</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
```
lbschanno: This dependency is not used and can be removed.

contrib/datawave-utils/check-tablet-extents/pom.xml
```diff
@@ -0,0 +1,142 @@
...
+            <artifactId>opentelemetry-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
```
lbschanno: This dependency is not used and can be removed.

contrib/datawave-utils/check-tablet-extents/pom.xml
```diff
@@ -0,0 +1,142 @@
...
+            <artifactId>accumulo-core</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.apache.logging.log4j</groupId>
+                    <artifactId>log4j-1.2-api</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.accumulo</groupId>
```
lbschanno: This dependency is not used and can be removed.

contrib/datawave-utils/check-tablet-extents/pom.xml
```diff
@@ -0,0 +1,142 @@
...
+            <artifactId>accumulo-hadoop-mapreduce</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.accumulo</groupId>
```
lbschanno, line 45: This dependency is not used and can be removed.

contrib/datawave-utils/check-tablet-extents/pom.xml
```diff
@@ -0,0 +1,142 @@
...
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.20.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-client-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.zookeeper</groupId>
```
lbschanno, line 49: This dependency is not used and can be removed.

contrib/datawave-utils/check-tablet-extents/pom.xml
```diff
@@ -0,0 +1,142 @@
...
+            <artifactId>zookeeper</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.dropwizard.metrics</groupId>
```
lbschanno, line 54: This dependency is not used and can be removed.

contrib/datawave-utils/check-tablet-extents/src/test/java/datawave/TabletExtentCheckerTest.java
```diff
@@ -0,0 +1,637 @@
...
+            expectExtent("1000", "1500");
+            assertResult();
+        }
+
+        /**
+         * Verify that given a table with compactable tablets, a key extent range, starting row and ending row, mergeExtents = true, and compactTablets = true,
+         * {@link TabletExtentChecker#findCompactableTablets(ClientContext, String, Text, Text, boolean )} will return an empty list after compacting the
+         * tablets automatically in {@link TabletExtentChecker#checkTablets(ClientContext, String, Text, Text, boolean, boolean)}.
+         *
+         * @throws TableNotFoundException
```
lbschanno: These @throws tags can be deleted.

contrib/datawave-utils/check-tablet-extents/src/test/java/datawave/TabletExtentCheckerTest.java
```diff
@@ -0,0 +1,637 @@
...
+            expectExtent("1000", "1500");
+            assertResult();
+        }
+
+        /**
+         * Verify that given a table with compactable tablets, a key extent range, starting row and ending row, mergeExtents = false, and compactTablets = true,
+         * {@link TabletExtentChecker#findCompactableTablets(ClientContext, String, Text, Text, boolean )} will return an empty list after compacting the
+         * tablets automatically in {@link TabletExtentChecker#checkTablets(ClientContext, String, Text, Text, boolean, boolean)}.
+         *
+         * @throws TableNotFoundException
```
lbschanno: These @throws tags can be deleted.

contrib/datawave-utils/check-tablet-extents/src/test/java/datawave/TabletExtentCheckerTest.java
```diff
@@ -0,0 +1,637 @@
...
+        this.end = end;
+    }
+
+    private void assertResult() throws AccumuloException, TableNotFoundException, IOException {
+        List<Pair<Text,Text>> result = TabletExtentChecker.findCompactableTablets((ClientContext) client, TABLE_NAME, getText(begin), getText(end),
+                        mergeExtents);
+        assertEquals(expectedExtents, result);
+    }
+
+    private void assertCompactedResult() throws AccumuloException, TableNotFoundException, AccumuloSecurityException, IOException {
```
lbschanno, line 617: This method can be simplified to:
```java
private void assertCompactedResult() throws AccumuloException, TableNotFoundException, AccumuloSecurityException, IOException {
    // Use the tool to compact tablets.
    TabletExtentChecker.checkTablets((ClientContext) client, TABLE_NAME, getText(begin), getText(end), mergeExtents, true);
    
    // Fetch the list of compactable tablets within the same range.
    List<Pair<Text,Text>> extents = TabletExtentChecker.findCompactableTablets((ClientContext) client, TABLE_NAME, getText(begin), getText(end), mergeExtents);
    
    // Verify the list is empty, indicating success.
    assertTrue(extents.isEmpty());
}
```

contrib/datawave-utils/check-tablet-extents/pom.xml
```diff
@@ -0,0 +1,142 @@
...
+                <version>${version.accumulo}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.beust</groupId>
```
lbschanno: This dependency will be on the Accumulo classpath, and can be used here with the provided scope.

contrib/datawave-utils/check-tablet-extents/pom.xml
```diff
@@ -0,0 +1,142 @@
...
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-context</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.accumulo</groupId>
```
lbschanno, line 35: This dependency will be on the Accumulo classpath, and can be used here with the provided scope.

contrib/datawave-utils/check-tablet-extents/pom.xml
```diff
@@ -0,0 +1,142 @@
...
+            <artifactId>accumulo-hadoop-mapreduce</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.accumulo</groupId>
+            <artifactId>accumulo-tserver</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
```
lbschanno: If we change the version here to 3.18.0, this dependency will be on the Accumulo classpath, and can be used here with the provided scope.

contrib/datawave-utils/check-tablet-extents/pom.xml
```diff
@@ -0,0 +1,142 @@
...
+            <artifactId>commons-lang3</artifactId>
+            <version>3.20.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
```
lbschanno: This dependency will be on the Accumulo classpath, and can be used here with the provided scope.

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
```diff
@@ -0,0 +1,319 @@
...
+     *             if an error occurs while reading the client properties file
+     */
+    @Override
+    public void execute(String[] args) throws AccumuloException, TableNotFoundException, IOException, AccumuloSecurityException {
+        // Parse the arguments.
+        Opts opts = new Opts();
+
+        opts.parseArgs(TabletExtentChecker.class.getName(), args);
+
+        if (opts.getPropertiesPath() == null) {
```
lbschanno: I'm not sure that this check here is needed. After removing it and redeploying, I was still able to execute the tool via `accumulo check-tablets` and `accumulo datawave.TabletExtentChecker` without passing in a properties path.

lbschanno (changes requested):

lbschanno:
Overall looks pretty good. Made a couple suggestions.

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
lbschanno: Note: The above applies only if we can expect the properties file to be present in the accumulo classpath after setup.

lbschanno: Follow up: I still don't think we should throw an error here if the properties path is null. This is not an issue if the accumulo classpath contains the properties file. In the scenario where we need to, the properties path can be passed in via `--props`.

tom-vt4: Will apply this fix, thank you for pointing that out.

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
tom-vt4: Fixed, will be in the next commit.

contrib/datawave-utils/check-tablet-extents/pom.xml
tom-vt4: Fixed

contrib/datawave-utils/check-tablet-extents/pom.xml
tom-vt4: Fixed

contrib/datawave-utils/check-tablet-extents/pom.xml
tom-vt4: Fixed

contrib/datawave-utils/check-tablet-extents/pom.xml
tom-vt4: Fixed

contrib/datawave-utils/check-tablet-extents/pom.xml
tom-vt4: Fixed

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
```diff
@@ -0,0 +1,315 @@
...
+            FileSystem ns = FileSystem.get(file.getPath().toUri(), context.getHadoopConf());
+            try (FileSKVIterator openReader = fileFactory.newReaderBuilder().forFile(file.getPathStr(), ns, ns.getConf(), NoCryptoServiceFactory.NONE)
+                            .withTableConfiguration(tableConf).seekToBeginning().build()) {
+                Key first = openReader.getFirstKey();
+                Key last = openReader.getLastKey();
+
+                // A tablet requires a compaction if any of the following are true:
+                // - The first key is outside the tablet's extent.
+                // - The last key is outside the tablet's extent.
+                System.out.println("Extent: " + extent);
```
ivakegg: Lets add a parameter to denote whether the extra debug info is displayed or not.  By default we should probably avoid the extra output.  Other than that I think this tool does what I need.  Thank you!

tom-vt4: Added debug parameter in latest commit.

contrib/datawave-utils/check-tablet-extents/pom.xml
tom-vt4: Removing accumulo-tsever breaks the build on my end.

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
```diff
@@ -0,0 +1,321 @@
...
+
+                if ((first != null && !extent.contains(first.getRow())) || (last != null && !extent.contains(last.getRow()))) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+    public static void checkTablets(ClientContext context, String tableName, Text begin, Text end, boolean mergeExtents, boolean compactTablets, boolean debug)
```
lbschanno: Similar not here about making the argument list simpler by passing in the Opts instance.

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
```diff
@@ -0,0 +1,321 @@
...
+     *            will merge recommended ranges of neighboring tablets requiring compaction
+     * @return the list of tablet boundaries recommended for compaction
+     * @throws AccumuloException
+     *             if an error occurs while connecting to Accumulo
+     * @throws TableNotFoundException
+     *             if the specified table does not exist
+     * @throws IOException
+     *             if an error occurs while reading the client properties file
+     */
+    static List<Pair<Text,Text>> findCompactableTablets(ClientContext context, String tableName, Text begin, Text end, boolean mergeExtents, boolean debug)
```
lbschanno: We're starting to pass a fair number of arguments to this method now. It would be simpler to just pass in the Opts instance as an argument, and fetch the appropriate information from it.

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
```diff
@@ -0,0 +1,321 @@
...
+     *            the arguments
+     * @throws AccumuloException
+     *             if an error occurs while connecting to Accumulo
+     * @throws TableNotFoundException
+     *             if the specified table does not exist
+     * @throws IOException
+     *             if an error occurs while reading the client properties file
+     */
+    @Override
+    public void execute(String[] args) throws AccumuloException, TableNotFoundException, IOException, AccumuloSecurityException {
```
lbschanno, line 138: I would like to see the functions ordered a bit differently here. I'd like to see them in the following order:
- main
- execute
- checkTablets
- formatArg

This allows for a more natural progression of understanding the logic flow when reading over the code.

lbschanno (changes requested):

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
tom-vt4: Fixed

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
tom-vt4: Fixed

contrib/datawave-utils/check-tablet-extents/src/main/java/datawave/TabletExtentChecker.java
tom-vt4: Fixed

contrib/datawave-utils/check-tablet-extents/README.md
```diff
@@ -0,0 +1,248 @@
...
+
+or
+
+```bash
+$ accumulo datawave.TabletExtentChecker -h
+```
+
+Output:
+
+```bash
```
lbschanno, line 50: The README here needs to be updated to reflect the new `--debug` option.

lbschanno (changes requested):

lbschanno (approved):

ivakegg (approved):

PR #3651, opened 2026-07-01
Alternate fix for hit excerpts on grouped term-frequency field names

drewfarris:
#3647 proposes some changes to the hit excerpt code to account for omitted hit excerpts when content contexts are in play. We ran into some issues with that fix pulling in term positions from outside contexts. After some analysis, it was discovered that we weren't properly passing the specific hit context for matched terms. This change addresses that issue and omits some of the changes in #3647 that exacerbated the issue.

This fix also omits #3647's migration from EasyMock, which should be potentially considered for a future PR.

### DIFF
--- a/warehouse/query-core/src/test/java/datawave/query/transformer/ExcerptTransformTest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/transformer/ExcerptTransformTest.java
@@ -290,6 +290,44 @@ public class ExcerptTransformTest extends EasyMockSupport {
     }
 
     /**
+     * Verify that hit terms collected for one document do not leak into excerpts generated for a later document.
+     */
+    @Test
+    @SuppressWarnings("rawtypes")
+    public void testHitTermsClearedBetweenDocuments() throws IOException {
+        givenExcerptField("BODY", 2);
+
+        Key metadata = new Key("shard", "dt\u0000uid");
+        Document firstDocument = mock(Document.class);
+        expect(firstDocument.isToKeep()).andReturn(true);
+        expect(firstDocument.containsKey(ExcerptTransform.PHRASE_INDEXES_ATTRIBUTE)).andReturn(false);
+        expect(firstDocument.containsKey(JexlEvaluation.HIT_TERM_FIELD)).andReturn(true);
+        Attribute firstHitTerms = new Attributes(Collections.singletonList(new Content("BODY:first", metadata, true)), true);
+        expect(firstDocument.get(JexlEvaluation.HIT_TERM_FIELD)).andReturn(firstHitTerms);
+        givenMatchingTermFrequencies("BODY", new int[][] {{2, 2}}, "first");
+        givenMatchingPhrase("BODY", 0, 4, "the [first] document excerpt", List.of("first"));
+        firstDocument.put(eq(ExcerptTransform.HIT_EXCERPT), isA(Attributes.class));
+
+        PhraseIndexes secondPhraseIndexes = new PhraseIndexes();
+        secondPhraseIndexes.addIndexTriplet("BODY", EVENT_ID, 4, 4);
+        Document secondDocument = mock(Document.class);
+        expect(secondDocument.isToKeep()).andReturn(true);
+        expect(secondDocument.containsKey(ExcerptTransform.PHRASE_INDEXES_ATTRIBUTE)).andReturn(true);
+        Attribute phraseIndexAttribute = new Content(secondPhraseIndexes.toString(), metadata, false);
+        expect(secondDocument.get(ExcerptTransform.PHRASE_INDEXES_ATTRIBUTE)).andReturn(phraseIndexAttribute);
+        expect(secondDocument.containsKey(JexlEvaluation.HIT_TERM_FIELD)).andReturn(false);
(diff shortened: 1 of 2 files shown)


warehouse/query-core/src/test/java/datawave/query/transformer/ExcerptTransformTest.java
```diff
@@ -237,6 +237,96 @@ public void testExcerptOverlappedAndPhraseOverlapped() throws IOException {
...
+        initTransform();
+        replayAll();
+
+        applyTransform();
+        verifyAll();
+
+        Attributes arg = capturedArg.getValue();
+        assertEquals(1, arg.size());
+        String excerpt = arg.getAttributes().iterator().next().getData().toString();
+        assertEquals("the quick brown fox jumped over the lazy dog", excerpt);
```
apmoriarty, line 262: No excerpt brackets?

apmoriarty (dismissed):

warehouse/query-core/src/test/java/datawave/query/transformer/ExcerptTransformTest.java
austin007008: the "collections.emptylist()" is us giving no hit terms:

givenMatchingPhrase("BODY.hash1", 8, 16, "the quick brown fox jumped over the lazy dog", **Collections.emptyList()**);

warehouse/query-core/src/test/java/datawave/query/transformer/ExcerptTransformTest.java
```diff
@@ -237,6 +237,96 @@ public void testExcerptOverlappedAndPhraseOverlapped() throws IOException {
...
+        verifyAll();
+
+        Attributes arg = capturedArg.getValue();
+        assertEquals(1, arg.size());
+        String excerpt = arg.getAttributes().iterator().next().getData().toString();
+        assertEquals("and the [word] from bird", excerpt);
+    }
+
+    /**
+     * Verify that hit terms collected for one document do not leak into excerpts generated for a later document.
```
austin007008, line 292: maybe I'm understanding this test wrong but to verify this, wouldn't we want the hit term from the first document to be something that exists in the excerpt for the second document and we want to see that it doesn't get bracketed?

warehouse/query-core/src/test/java/datawave/query/transformer/ExcerptTransformTest.java
apmoriarty: Okay, so when there are no hit terms no excerpt is generated. Maybe rename the test to reflect that this is an exceptional case, not an expected case.

warehouse/query-core/src/test/java/datawave/query/transformer/ExcerptTransformTest.java
```diff
@@ -237,6 +237,96 @@ public void testExcerptOverlappedAndPhraseOverlapped() throws IOException {
...
+        Attribute phraseIndexAttribute = new Content(secondPhraseIndexes.toString(), metadata, false);
+        expect(secondDocument.get(ExcerptTransform.PHRASE_INDEXES_ATTRIBUTE)).andReturn(phraseIndexAttribute);
+        expect(secondDocument.containsKey(JexlEvaluation.HIT_TERM_FIELD)).andReturn(false);
+        givenMatchingPhrase("BODY", 2, 6, "the second document excerpt", Collections.emptyList());
+        secondDocument.put(eq(ExcerptTransform.HIT_EXCERPT), isA(Attributes.class));
+
+        initTransform();
+        replayAll();
+
+        excerptTransform.apply(new AbstractMap.SimpleEntry<>(new Key(), firstDocument));
```
apmoriarty: `Map.entry(key, document)` ?

warehouse/query-core/src/test/java/datawave/query/transformer/ExcerptTransformTest.java
austin007008: yeah to piggyback off of what he is saying, the underlying iterator will actually work with phrase offsets and without the hit terms list populated (you just wont get hit term highlighting), but in reality, that will never happen because the HitListArithmetic relies on the hit list to create the phrase offsets and without those, we can't generate an excerpt.

warehouse/query-core/src/test/java/datawave/query/transformer/ExcerptTransformTest.java
drewfarris: This test essentially verifies that `TermFrequencyExcerptIterator.setHitTermsList([first])` is only called once (for the first document). If the `hitTermValues.clear()` call is removed from `ExcerptTransform` that same method actually gets called twice - once for each document. It's not a great black-box test, which is why it's not as readable as I'd like.

This test will definitely be improved once we replace EasyMock.

warehouse/query-core/src/test/java/datawave/query/transformer/ExcerptTransformTest.java
drewfarris: Done, will push.

apmoriarty (approved):

austin007008 (approved):

PR #3654, opened 2026-07-07
Add query test framework

apmoriarty:
Add a query test framework with generator-driven statistical query coverage

Introduces datawave-core-test, a new module providing composable generators for field names, event ids, and query terms (equality, negation, null checks, phrases) that combine into QueryGenerators capable of producing large, self-verifying sets of queries, expected plans, and expected result ids for a given ingested dataset. StatisticalQueryTest exercises this framework against ShardQueryLogic across indexed, index-only, and event-only fields, numeric types, content:phrase queries, and dynamic multi-shard data, asserting query plan, and result counts for every generated query.

Also discovered some state bugs with the DocumentScheduler when the logic is reused in the same test. Those fixes will follow separately.


warehouse/pom.xml
```diff
@@ -238,7 +238,7 @@
             <dependency>
                 <groupId>org.apache.commons</groupId>
                 <artifactId>commons-lang3</artifactId>
-                <version>3.14.0</version>
```
alerman, line 241: Commons tends to be pretty good about following semantic versioning, but i dont know that id concor this is no-downstream-effects. Maybe bump this in another pr to isolate it and to int testing.

warehouse/query-core/src/test/java/datawave/query/StatisticalQueryTest.java
```diff
@@ -0,0 +1,320 @@
...
+    @Override
+    protected List<String> getIndexTableNames() {
+        return List.of(TableName.SHARD_INDEX);
+    }
+
+    @BeforeEach
+    public void beforeEach() throws IOException {
+        TimeZone.setDefault(TimeZone.getTimeZone("GMT"));
+        setClientForTest(client);
+        givenDate("20260202");
```
alerman, line 138: This may be normal in query testing as its not something i have looked at often, but i don't love the hard coded date

alerman: At least make it a constant since you use it twice in this file.

apmoriarty: It's normal to have the date as a constant. The test framework should have a default date that the operator can override. That can be done here. 

Future work should add support for date ranges so we can cover the multi-day, multi-shard use cases.

warehouse/query-core/src/test/java/datawave/query/StatisticalQueryTest.java
```diff
@@ -0,0 +1,320 @@
...
+        executeQueryMetadata(queryMetadata);
+    }
+
+    protected void executeQueryMetadata(List<QueryMetadata> queryMetadata) throws Exception {
+        for (QueryMetadata metadata : queryMetadata) {
+            givenDate("20260202");
+            givenQuery(metadata.getQuery());
+            expectPlan(metadata.getPlan());
+            expectResultCount(metadata.getIds().size());
+            // log.info("query: {}", metadata.getQuery());
```
alerman: Either raise the log level if you dont want to see the log, or remove the lines. Dont leave in but commented

warehouse/query-core/src/main/java/datawave/next/scanner/DocumentScheduler.java
```diff
@@ -45,7 +49,21 @@ public DocumentScheduler(ShardQueryConfiguration config) {
...
         this.config.setQueryId(config.getQuery().getId().toString());
 
-        this.queryDataIterator = config.getQueriesIter();
+        // the DocumentScannerConfig instance is shared/cloned across every query run through this query logic, but
+        // numSearchScans/numRetrievalScans/stats are per-query runtime state. Reusing them would let a still-unwinding
+        // scan from a prior query (after this.close() calls shutdownNow(), which interrupts but does not wait for
+        // termination) race with and corrupt the counts this query relies on to know when it is done, causing results
+        // to be dropped. Give each query a fresh instance so straggling threads from the previous query only touch
+        // objects nobody is reading anymore.
+        this.config.setNumSearchScans(new AtomicInteger(0));
```
alerman: I think this would also invalidate the no-downstream-effects label

core/tables/src/main/java/datawave/ingest/protobuf/TermWeight.java
alerman, line 1: What caused these protobuf files to get regenerated?

apmoriarty: They were not regenerated, this is a direct lift-and-shift that allows the new test module to avoid a direct dependency on ingest. One of the design goals was to have as few dependencies as possible on query-core and ingest-core

warehouse/query-core/src/test/java/datawave/query/StatisticalQueryTest.java
apmoriarty: Done in https://github.com/NationalSecurityAgency/datawave/pull/3654/changes/a2ae7736e1c3bb874d301677465564636b2bd15a

warehouse/query-core/src/test/java/datawave/query/FizzBuzzQueryTest.java
```diff
@@ -0,0 +1,219 @@
...
+            givenQuery(metadata.getQuery());
+            expectPlan(metadata.getPlan());
+            expectResultCount(metadata.getIds().size());
+            if (log.isDebugEnabled()) {
+                log.debug("query: {}", metadata.getQuery());
+                log.debug(" plan: {}", metadata.getPlan());
+            }
+            planAndExecuteQuery();
+            // clean after each query
+            afterEach();
```
jschmidt10, line 217: It seems like Junit will invoke this for us, given its annotated with @AfterEach.

warehouse/query-core/src/test/java/datawave/query/StatisticalQueryTest.java
```diff
@@ -0,0 +1,322 @@
...
+        executeQueryMetadata(queryMetadata);
+    }
+
+    // ---- Single ----
+
+    // INDEXED == 'value'
+    @Test
+    public void testSingleOfIndexedEq() throws Exception {
+        List<QueryMetadata> queryMetadata = queryGenerator.singleTermIndexed(new EqTerm()).getQueries();
+        assertEquals(48, queryMetadata.size(), ASSERTION_MESSAGE);
```
jschmidt10, line 155: These query numbers we are asserting are in the realm of magic numbers right now. I'm assuming they are relative to the FieldMetadata or event count. Can we update these to be computed? Otherwise, someone might add a new field and be presented with every single unit test failing.

warehouse/query-core/src/test/java/datawave/query/StatisticalQueryTest.java
```diff
@@ -0,0 +1,322 @@
...
+        //  @formatter:off
+        metadata = IngestMetadataBuilder.builder()
+                .setMetadataColumns(List.of(I, E, TF))
+                .addNormalizers(List.of(new LcNoDiacriticsType(), new NumberType()))
+                .enableAlphabeticFields()
+                .enableNumericFields()
+                .build();
+        //  @formatter:on
+
+        metadata.create(75);
```
jschmidt10: Nit: Maybe `createEvents()` would be a little more readable.

core/datawave-core-test/src/main/java/datawave/test/framework/IngestMetadata.java
```diff
@@ -0,0 +1,353 @@
...
+     * First generate all combinations of metadata columns {i, ri, e, tf}
+     * Then, generate all combinations of normalizers
+     * Then, for each metadata column roll through each normalizer.
+     * Then, for each field (alpha, numeric) apply appropriate data (fixed, random)
+     * </pre>
+     *
+     * @param eventCount
+     *            the number of events to generate
+     */
+    public void create(int eventCount, int valuesPerField) {
```
jschmidt10: IntelliJ is reporting that this parameter version of `valuesPerField` is never used. Looks like it's picking up the class member variable in `createMetadataForFields()`. Might be worth a quick code quality run-thru to see if there's any other similar cleanup.

core/datawave-core-test/src/main/java/datawave/test/framework/writers/ShardTableWriter.java
```diff
@@ -0,0 +1,162 @@
...
+                Mutation m = mutationsByRow.computeIfAbsent(row, Mutation::new);
+
+                String value = field.getValueForEventId(eventId);
+                String uid = UidGenerator.uid(String.valueOf(eventId));
+                for (Type<?> normalizer : field.getNormalizers()) {
+                    String normalizedValue = normalizer.normalize(value);
+
+                    // TODO: viz
+                    String cq = normalizedValue + "\0" + datatype + "\0" + uid;
+                    Key key = new Key(row, cf, cq, "ALL", TIMESTAMP);
```
jschmidt10: We're making a Key and then immediately decomposing it. You can just plug `row`, `cf`, etc. into the `m.put()`.

jschmidt10: Same pattern in the other createXXXColumn methods. Not a big deal just a nit.

core/datawave-core-test/src/main/java/datawave/test/framework/writers/ShardTableWriter.java
```diff
@@ -0,0 +1,162 @@
...
+            for (int eventId : field.getEventIds()) {
+                String row = ShardKeyUtil.buildRow(eventId, numShards);
+                Mutation m = mutationsByRow.computeIfAbsent(row, Mutation::new);
+
+                String value = field.getValueForEventId(eventId);
+                String uid = UidGenerator.uid(String.valueOf(eventId));
+                for (Type<?> normalizer : field.getNormalizers()) {
+                    String normalizedValue = normalizer.normalize(value);
+
+                    // TODO: viz
```
jschmidt10: Honestly, I'd be fine to let this framework ignore viz for the foreseeable future. It probably is worth having a constant somewhere that is shared rather than the hard-coded `"ALL"` in both the test framework and the unit tests.

core/datawave-core-test/src/main/java/datawave/test/framework/writers/TableWriter.java
```diff
@@ -0,0 +1,37 @@
...
+     *
+     * @param client
+     *            the accumulo client
+     * @param metadata
+     *            the list of field metadata
+     * @param numShards
+     *            the number of shards to distribute events across
+     */
+    public static void write(AccumuloClient client, List<FieldMetadata> metadata, int numShards) {
+        MetadataTableWriter.write(client, MetadataTableConverter.convert(metadata));
```
jschmidt10: The other XXXTableWriters accept the metadata object and do their thing (such as a switch statement on the columns). Here we packed that switch statement on the columns into a Converter and then we feed Keys into the Writer rather than the metadata object. 

To maintain the established pattern, I'd feed in just the metadata object and get rid of the Converter - put that method directly into MetadataTableWriter and then they all behave the same.

core/datawave-core-test/src/main/java/datawave/test/framework/TableCreator.java
```diff
@@ -0,0 +1,45 @@
...
+        createShardIndex(tops);
+        tops.create(TableName.METADATA);
+    }
+
+    public static void createShardIndex(TableOperations tops) throws Exception {
+        tops.create(TableName.SHARD_INDEX);
+
+        Map<String,String> additions = new HashMap<>();
+        IteratorUtil.IteratorScope[] scopes = IteratorUtil.IteratorScope.values();
+        for (IteratorUtil.IteratorScope scope : scopes) {
```
jschmidt10, line 36: Not a problem to solve today but in the future, we could leverage the TableConfigHelpers to ensure that our iterator stack matches what Datawave queries expect.

core/datawave-core-test/src/main/java/datawave/test/framework/generators/query/IntersectionFactory.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Override
+    protected String operator() {
+        return "&&";
+    }
+
+    @Override
+    protected Set<Integer> combine(Set<Integer> leftIds, boolean leftNegated, Set<Integer> rightIds, boolean rightNegated) {
+        if (leftNegated) {
```
jschmidt10, line 28: This API is a bit tricky as it may edit leftIds or rightIds. A caller might not expect that behavior.

lbschanno:
I worry this may be rewriting the wheel a bit in terms of creating generators for the random data. Have you looked at all into employing libraries such as [datafaker](https://github.com/datafaker-net/datafaker) to handle the random data generation?

apmoriarty:
> I worry this may be rewriting the wheel a bit in terms of creating generators for the random data. Have you looked at all into employing libraries such as [datafaker](https://github.com/datafaker-net/datafaker) to handle the random data generation?

I looked at external data libraries but did not want to include any external dependencies for a first iteration. One of the design principles was an isolated module with the fewest dependencies possible, which is important for downstream consumers.

warehouse/query-core/src/test/java/datawave/query/FizzBuzzQueryTest.java
apmoriarty: This is invoked after each query run in other tests. Added a comment to indicate this is expected. The AbstractQueryTest framework should really have a `resetState()` method that is called in the `afterEach()` method to avoid confusion.

warehouse/query-core/src/test/java/datawave/query/StatisticalQueryTest.java
apmoriarty: I looked at calculating the number of queries generated but didn't like the approach. Might revisit in the future, leaving the magic numbers for now so the operator has an idea of how many queries they are running. There is a scenario where a new test is added that reaches an unsustainable number of test queries, these assertions are meant to guard against that.

warehouse/query-core/src/test/java/datawave/query/StatisticalQueryTest.java
apmoriarty: Easy day. Done.

core/datawave-core-test/src/main/java/datawave/test/framework/IngestMetadata.java
apmoriarty: The variable is now pulled through.

core/datawave-core-test/src/main/java/datawave/test/framework/writers/ShardTableWriter.java
apmoriarty: Good catch, fixed throughout.

core/datawave-core-test/src/main/java/datawave/test/framework/writers/ShardTableWriter.java
apmoriarty: Constant added. Visibility is noted as a roadmap item because there are some test cases where we want to exercise that logic...but most cases should not have to do that.

core/datawave-core-test/src/main/java/datawave/test/framework/writers/TableWriter.java
apmoriarty: Merged per your suggestion. I don't know why that wasn't cleaned up as part of the initial round of local reviews.

core/datawave-core-test/src/main/java/datawave/test/framework/generators/query/IntersectionFactory.java
apmoriarty: Local copy is made to preserve inputs, great catch.

apmoriarty:
Closing in favor of a series of smaller, easier to digest PRs

PR #3664, opened 2026-07-09
Update query-core UID class references to core/tables

apmoriarty:
Points query-core imports at the migrated datawave.table.hash.* classes (UID, HashUID, SnowflakeUID, UIDConstants, UIDBuilder) instead of the deprecated datawave.data.hash.* originals in warehouse/core. Deprecated classes and their dedicated tests are left in place.


apmoriarty:
close for now

PR #3670, opened 2026-07-09
Migrate KeywordQueryLogicFunctionalTest from Arquillian to JUnit5/Spring

apmoriarty:
Drops the Arquillian container deployment in favor of SpringExtension with the same context config used by the AbstractQueryTest-based tests. KeywordQueryLogic does not extend ShardQueryLogic, so this test cannot extend AbstractQueryTest itself, but table setup/ingest and test lifecycle now match the pattern used elsewhere (QueryTestTableHelper + WiseGuysIngest under @BeforeAll, JUnit5 assertions).

### DIFF
--- a/warehouse/query-core/src/test/java/datawave/query/tables/keyword/KeywordQueryLogicFunctionalTest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/tables/keyword/KeywordQueryLogicFunctionalTest.java
@@ -1,11 +1,10 @@
 package datawave.query.tables.keyword;
 
 import static datawave.query.tables.keyword.TagCloudTestUtil.CAPONE_SOURCE;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNull;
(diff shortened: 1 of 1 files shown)


avgAGB:
not to harp on it, but while we're at it with these migrations and refactors maybe could check out if its worth switching the log4j library for slf4j? Would be curious to hear what you think about it

apmoriarty:
> not to harp on it, but while we're at it with these migrations and refactors maybe could check out if its worth switching the log4j library for slf4j? Would be curious to hear what you think about it

Do you mind if I do that a series of separate PRs?

avgAGB:
> > not to harp on it, but while we're at it with these migrations and refactors maybe could check out if its worth switching the log4j library for slf4j? Would be curious to hear what you think about it
> 
> Do you mind if I do that a series of separate PRs?

that's fine with me, LGTM

lbschanno (approved):

hoper-38709 (approved):

PR #3673, opened 2026-07-10
Migrate ShardQueryLogicQueryValidationTest from Arquillian to JUnit5/Spring

apmoriarty:
Drops the Arquillian container deployment in favor of SpringExtension with the same context config used by the AbstractQueryTest-based tests. This test calls ShardQueryLogic.validateQuery() directly and asserts on QueryRuleResult/exception state, never planning or executing a query, so it doesn't fit AbstractQueryTest's plan/execute/assert-results model at all; it's modernized to plain JUnit5 instead, same as the other non-ShardQueryLogic-flow test files. Also swaps the QueryTestTableHelper instance-name label from the still-unmigrated ShardQueryLogicTest.ShardRange (borrowed only for a unique string, no functional dependency) to this class's own name, since that reference would otherwise break once ShardQueryLogicTest.ShardRange is removed by a later migration.

### DIFF
--- a/warehouse/query-core/src/test/java/datawave/query/tables/ShardQueryLogicQueryValidationTest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/tables/ShardQueryLogicQueryValidationTest.java
@@ -1,5 +1,9 @@
 package datawave.query.tables;
 
+import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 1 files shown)


lbschanno (approved):

hoper-38709 (approved):
Nice.

PR #3675, opened 2026-07-10
Migrate UniqueTest from Arquillian to AbstractQueryTest

apmoriarty:
Drops the Arquillian container deployment and the ShardRange/DocumentRange nested subclasses in favor of extending AbstractQueryTest directly, with the SHARD/DOCUMENT ingest range and each test method's per-case loops converted to @ParameterizedTest + @MethodSource, following the pattern used for QueryFunctionQueryTest/FunctionalSetTest. Ingestion for both ranges happens once in @BeforeAll.

This test's assertion shape doesn't fit AbstractQueryTest's default Document/UUID model at all - it needs DocumentTransformer-produced EventBase objects grouped into "unique" sets, matched against sets of acceptable internal IDs. executeQuery() is overridden to build that event list directly (using logic.getConfig().getQuery(), which holds the same settings object passed to initialize()) instead of relying on the base class's Document deserialization, and extraAssertions() replicates the original set-matching logic, using a local copy of expectedGroups per the same destructive-iteration gotcha already found in ExcerptTest/SummaryTest.

Also fixes a stale expected-grouping in testUniquenessWithHitTermField: the three WiseGuys UIDs were expected to collapse into a single unique result, but HIT_TERM legitimately differs per event (the matched term text includes the UUID value itself), so they should not collapse the way the sibling BOTH_NULL-based tests do. Confirmed via a temporary debug print showing all three IDs correctly returned as three distinct events before correcting the expectation to three separate singleton groups.

### DIFF
--- a/warehouse/query-core/src/test/java/datawave/query/UniqueTest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/UniqueTest.java
@@ -1,54 +1,45 @@
 package datawave.query;
 
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
 import java.net.URL;
-import java.text.DateFormat;
-import java.text.SimpleDateFormat;
 import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.TimeZone;
-import java.util.UUID;
-
-import javax.inject.Inject;
+import java.util.stream.Stream;
 
 import org.apache.accumulo.core.client.AccumuloClient;
 import org.apache.accumulo.core.security.Authorizations;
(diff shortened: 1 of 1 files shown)


warehouse/query-core/src/test/java/datawave/query/UniqueTest.java
```diff
@@ -57,303 +48,218 @@
...
+        }
+        assertTrue(remaining.isEmpty(), "Failed to find all expected results.  Missing " + remaining);
     }
 
-    @AfterClass
-    public static void teardown() {
-        TypeRegistry.reset();
+    @BeforeAll
+    public static void beforeAll() throws Exception {
+        // testing tear downs but without consistency, because when we tear it down then we loose the ongoing bloom filter and subsequently the rebuild
```
lbschanno:
```suggestion
        // testing tear downs but without consistency, because when we tear it down then we lose the ongoing bloom filter and subsequently the rebuild
```
Typo

lbschanno (changes requested):

lbschanno (approved):

avgAGB (approved):

PR #3685, opened 2026-07-13
Dynamic day, shard, and timestamp

apmoriarty:
Test framework now generates dynamic row, shard offset, and timestamp. This is required for time-based filter functions, the most recent unique function, and queries whose date range is not midnight-to-midnight (sub or partial date ranges)


avgAGB (approved):

apmoriarty:
Will be reworked as part of a later PR

PR #3693, opened 2026-07-16
Bugfix/excerpt transformer range

FineAndDandy:


warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -163,24 +165,35 @@ private HitOffset getOffset(ValueTuple hitTuple, PhraseIndexes phraseIndexes) {
             return null;
         }
         String fieldName = hitTuple.getFieldName();
+        String baseFieldName = JexlASTHelper.removeGroupingContext(fieldName);
+
         String eventId = keyToEventId(docKey);
 
-        // get the key at which we would find the term frequencies
-        Key tfKey = new Key(docKey.getRow().toString(), Constants.TERM_FREQUENCY_COLUMN_FAMILY.toString(),
+        // Term Frequency ranges have to include the baseFieldName through the <baseFieldName>.\uFFFF
```
avgAGB, line 172: the comment here mentions the range is between baseFieldName and baseFieldName.FFFF , but the first key is assigned as hitTuple.getFieldName() - is it a notable change or no?

warehouse/query-core/src/main/java/datawave/query/transformer/ExcerptTransform.java
```diff
@@ -163,24 +165,35 @@ private HitOffset getOffset(ValueTuple hitTuple, PhraseIndexes phraseIndexes) {
...
+                        docKey.getColumnFamily().toString() + '\u0000' + hitTuple.getValue() + '\u0000' + baseFieldName + "." + '\uFFFF');
+        Range range = new Range(beginTfKey, endTfKey);
+
         try {
             // seek directly to that key
             source.seek(range, Collections.emptyList(), false);
             if (source.hasTop()) {
+                // parse the actual field name of the TF key to use for excerpt generation
+                final String[] splits = source.getTopKey().getColumnQualifier().toString().split(NULL_BYTE);
+                String parsedFieldName = splits[splits.length - 1];
```
avgAGB, line 187: unlikely case, but if the splits comes back empty from a malformed column qualifier you could put a check here before setting up the parsedFieldName.

avgAGB (approved):

apmoriarty (approved):

PR #3696, opened 2026-07-17
Fix TLDTermFrequencyEventDataQueryFilter to recognize grouped TF fields

apmoriarty:
keep() and fieldMatches() compared the full grouped/content-context term-frequency field name (e.g. QUOTE.hash1) against un-grouped requested-field sets with an exact match, so any field written with grouping notation was always dropped from TLD query results even when explicitly requested.


hlgp (approved):

hlgp:
Seems like an opportunity to condense the functionality fieldMatches() and this new method in a common util like JexlASTHelper so we don't miss it again.  isRelevantField looks similar, too

PR #3699, opened 2026-07-17
Allow snapshot compose job to fail without red X or failure email

apmoriarty:
continue-on-error suppresses the job's failure status while keeping the existing compose-check-jobs gate logic (which already treats snapshot failure as tolerable) intact.

### DIFF
--- a/.github/workflows/compose-coordinator.yml
+++ b/.github/workflows/compose-coordinator.yml
@@ -14,6 +14,7 @@ jobs:
       allow-snapshots: false
     secrets: inherit
   call-compose-tests-with-snapshots:
+    continue-on-error: true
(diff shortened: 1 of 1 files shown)


.github/workflows/compose-coordinator.yml
```diff
@@ -14,6 +14,7 @@ jobs:
       allow-snapshots: false
     secrets: inherit
   call-compose-tests-with-snapshots:
+    continue-on-error: true
```
alerman, line 17: if adding here should probably add above to call-compose-tests too. im fine with this as long as it doesnt change the behavior of the check job

apmoriarty:
deferring

PR #3706, opened 2026-07-21
Migrate ColumnFamilyConstants usages to MetadataColumnFamilyConstants

apmoriarty:
datawave.data.ColumnFamilyConstants is deprecated for removal; switch all consumers to datawave.table.constants.MetadataColumnFamilyConstants and add datawave-core-tables as a dependency where it wasn't already on the classpath.


avgAGB (dismissed):

drewfarris:
It would be awesome if the constant class exposed both a Text and a String version of each constant so that we didn't have to call toString() everywhere.

lbschanno (dismissed):

lbschanno:
Not required for my approval, but I agree with @drewfarris about having Text and String versions of each cf constant. Perhaps in a separate PR if it's easier.

apmoriarty:
Wrote https://github.com/NationalSecurityAgency/datawave/issues/3712 to track suggestion.

lbschanno (dismissed):

lbschanno (approved):

avgAGB (approved):

PR #3708, opened 2026-07-21
Update AnnotationManagerBean to read from both annotation and truthmark tables

drewfarris:
- Implements FederatedAnnotationReader that can read from multiple sources and reconcile results.
- Removes unfinished write implementations from AnnotationManagerBean due to new write path design via microservices.
- Removes separate AnnotationSourceReader/Writer interfaces and moved source specific methods to AnnotationReader/Writer to eliminate complexity.
- Added truthmark cases to AnnotationManagerBeanFunctionalTest.
- Added assertion to compare two annotation lists to ensure equivalence.


core/annotation/src/main/java/datawave/annotation/data/v1/FederatedAnnotationReader.java
```diff
@@ -0,0 +1,224 @@
...
+
+    public FederatedAnnotationReader(Map<String,AnnotationReader> AnnotationReaderes) {
+        this(AnnotationReaderes, Executors.newCachedThreadPool(), DEFAULT_DAO_TIMEOUT_MILLIS);
+    }
+
+    public FederatedAnnotationReader(Map<String,AnnotationReader> AnnotationReaderes, ExecutorService executorService) {
+        this(AnnotationReaderes, executorService, DEFAULT_DAO_TIMEOUT_MILLIS);
+    }
+
+    public FederatedAnnotationReader(Map<String,AnnotationReader> AnnotationReaderes, ExecutorService executorService, long daoTimeoutMillis) {
```
lbschanno: Typo in the argument name `AnnotationReaderes`, and it should be camelcase.

core/annotation/src/main/java/datawave/annotation/data/v1/FederatedAnnotationReader.java
```diff
@@ -0,0 +1,224 @@
...
+
+    private final Map<String,AnnotationReader> annotationReaders;
+    private final ExecutorService executorService;
+    private final long daoTimeoutMillis;
+
+    public FederatedAnnotationReader(Map<String,AnnotationReader> AnnotationReaderes) {
+        this(AnnotationReaderes, Executors.newCachedThreadPool(), DEFAULT_DAO_TIMEOUT_MILLIS);
+    }
+
+    public FederatedAnnotationReader(Map<String,AnnotationReader> AnnotationReaderes, ExecutorService executorService) {
```
lbschanno: Typo in the argument name `AnnotationReaderes`, and it should be camelcase.

core/annotation/src/main/java/datawave/annotation/data/v1/FederatedAnnotationReader.java
```diff
@@ -0,0 +1,224 @@
...
+public class FederatedAnnotationReader implements AnnotationReader {
+
+    protected static final Logger log = LoggerFactory.getLogger(FederatedAnnotationReader.class);
+    private static final long DEFAULT_DAO_TIMEOUT_MILLIS = 5000L;
+
+    private final Map<String,AnnotationReader> annotationReaders;
+    private final ExecutorService executorService;
+    private final long daoTimeoutMillis;
+
+    public FederatedAnnotationReader(Map<String,AnnotationReader> AnnotationReaderes) {
```
lbschanno: Typo in the argument name `AnnotationReaderes`, and it should be camelcase.

core/annotation/src/main/java/datawave/annotation/data/v1/FederatedAnnotationReader.java
```diff
@@ -0,0 +1,224 @@
...
+    public Collection<String> getAnnotationTypes(String shard, String datatype, String uid) {
+        final Set<String> result = new TreeSet<>();
+        List<Collection<String>> daoResults = callDaosWithTimeout(dao -> dao.getAnnotationTypes(shard, datatype, uid), "annotation types for %s/%s/%s", shard,
+                        datatype, uid);
+        daoResults.forEach(result::addAll);
+        return result;
+    }
+
+    @Override
+    public List<Annotation> getAnnotations(String shard, String datatype, String uid) {
```
lbschanno: Rather than creating a LinkedHashSet and then an ArrayList from that set, we can accomplish the same goal with streaming:
```java
// @formatter:off
return callDaosWithTimeout(dao -> dao.getAnnotations(shard, datatype, uid), "annotations for %s%s%s", shard, datatype, uid).stream()
                .flatMap(Collection::stream)
                .distinct()
                .collect(Collectors.toCollection(ArrayList::new));
// @formatter:on
```

core/annotation/src/main/java/datawave/annotation/data/v1/FederatedAnnotationReader.java
```diff
@@ -0,0 +1,224 @@
...
+        final Set<Annotation> result = new LinkedHashSet<>();
+        List<List<Annotation>> daoResults = callDaosWithTimeout(dao -> dao.getAnnotations(shard, datatype, uid), "annotations for %s/%s/%s", shard, datatype,
+                        uid);
+        daoResults.forEach(result::addAll);
+        return new ArrayList<>(result);
+    }
+
+    @Override
+    public List<Annotation> getAnnotationsForType(String shard, String datatype, String uid, String annotationType) {
```
lbschanno: Similar note here about being able to create a list of distinct annotations with streaming.

core/annotation/src/main/java/datawave/annotation/data/v1/FederatedAnnotationReader.java
```diff
@@ -0,0 +1,224 @@
...
+    /**
+     * Given a list of results and the desire to populate a singular Optional result, produce the best Optional we can under the circumstances.
+     *
+     * @param results
+     *            the results to assess
+     * @return an optional of the same type as the collection of results.
+     * @param <T>
+     *            the type of object in the results to return as the optional.
+     */
+    protected static <T> Optional<T> getBest(List<T> results) {
```
lbschanno: We can avoid some of the overhead of first creating a list of all non-null instances of T by instead directly passing in the result of the dao call to getBest, and determining the best on the fly:

```java
protected static <T> Optional<T> getBest(List<Optional<T>> results) {
    T first = null;
    for (Optional<T> optional : results) {
        if(optional.isPresent()) {
            if(first == null) {
                first = optional.get();
            } else {
                if(!Objects.equals(first, optional.get())) {
                    throw new RuntimeException("Conflicting federated results returned from multiple data sources");
                }
            }
        }
    }
    return Optional.ofNullable(first);
}
```

core/annotation/src/main/java/datawave/annotation/data/v1/FederatedAnnotationReader.java
```diff
@@ -0,0 +1,224 @@
...
+                    results.add(value);
+                }
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                log.warn("Interrupted while collecting {} from {}", formatOperationDescription(messageTemplate, templateArgs), daoName);
+                break;
+            } catch (ExecutionException e) {
+                // Individual DAO failures should not fail the entire federated read; log and continue.
+                Throwable cause = e.getCause() == null ? e : e.getCause();
+                log.debug("Exception retrieving {} from {}: {}", formatOperationDescription(messageTemplate, templateArgs), daoName, cause.getMessage());
```
lbschanno: Recommend checking if `log.isDebugEnabled()`.

core/annotation/src/main/java/datawave/annotation/data/v1/FederatedAnnotationReader.java
```diff
@@ -0,0 +1,224 @@
...
+        }
+
+        // Walk the futures in DAO order so we can log which backend timed out or failed.
+        for (int i = 0; i < futures.size(); i++) {
+            Future<T> future = futures.get(i);
+            String daoName = daoEntries.get(i).getKey();
+
+            // A canceled future indicates the DAO did not complete before the timeout elapsed.
+            if (future.isCancelled()) {
+                log.debug("Timeout retrieving {} from {} after {}ms", formatOperationDescription(messageTemplate, templateArgs), daoName, daoTimeoutMillis);
```
lbschanno: Recommend checking if `log.isDebugEnabled()`.

core/annotation/src/test/java/datawave/annotation/test/v1/AnnotationAssertions.java
```diff
@@ -46,6 +47,35 @@ public static void assertMetadataEqual(Map<String,String> expectedMetadata, List
         }
     }
 
+    public static void assertAnnotationListsEqual(List<Annotation> expectedAnnotationsTemplate, List<Annotation> actualAnnotations) {
```
lbschanno: Not a requirement for my approval, but the implementation here is a good case for using the AssertJ `Assertions` class, which provides functionality to handle list comparisons. I think the equivalent of this function would be:

```java
// @formatter:off
Assertions.assertThat(expectedAnnotationsTemplate)
                .usingElementComparator(Comparator.comparing(Annotation::getAnnotationId))
                .containsExactlyInAnyOrderElementsOf(actualAnnotations);
// @formatter:on
```
If the assertion fails, it should print out which elements are missing/unexpected.

lbschanno (changes requested):

core/annotation/src/main/java/datawave/annotation/data/v1/AnnotationReader.java
tharke13: Use `annotationId` instead of `annotationUid` like the other getAnnotation() method as to not confuse with the `uid` parameter?

core/annotation/src/main/java/datawave/annotation/data/v1/FederatedAnnotationReader.java
```diff
@@ -0,0 +1,224 @@
...
+    @Override
+    public Optional<AnnotationSource> getAnnotationSource(String analyticHash) {
+        List<Optional<AnnotationSource>> daoResults = callDaosWithTimeout(dao -> dao.getAnnotationSource(analyticHash), "analyticHash %s", analyticHash);
+        final List<AnnotationSource> results = new ArrayList<>();
+        daoResults.forEach(optional -> optional.ifPresent(results::add));
+        return getBest(results);
+    }
+
+    @Override
+    public Optional<Annotation> getAnnotation(String shard, String datatype, String uid, String annotationType, String annotationUid) {
```
tharke13: Similar to other comment use `annotationId` instead of `annotationUid` for parameter name?

core/annotation/src/main/java/datawave/annotation/data/v1/FederatedAnnotationReader.java
drewfarris: Thanks for the suggestion. I took this concept to it's logical conclusion - let me know if you think my approach is sufficiently clear/readable in the latest iteration

core/annotation/src/main/java/datawave/annotation/data/v1/FederatedAnnotationReader.java
drewfarris: Oof, embarassing, I should have caught this.

core/annotation/src/test/java/datawave/annotation/test/v1/AnnotationAssertions.java
drewfarris: Thanks, will look into this.

core/annotation/src/main/java/datawave/annotation/data/v1/FederatedAnnotationReader.java
drewfarris: I think this is addressed by using `Optional::stream` in the new `callDaosThenGetBest` method.

core/annotation/src/main/java/datawave/annotation/data/v1/FederatedAnnotationReader.java
```diff
@@ -0,0 +1,285 @@
...
+     * @param messageTemplate
+     *            logging template used when DAO calls timeout, fail, or are interrupted
+     * @param templateArgs
+     *            arguments used to populate the logging template when needed
+     * @param <T>
+     *            value type contained in each DAO's optional response
+     * @return the resolved federated value, or {@link Optional#empty()} when no DAO produced one
+     */
+    protected <T> Optional<T> callDaosThenGetBest(DaoCall<Optional<T>> daoCall, String messageTemplate, String... templateArgs) {
+        return callDaos(daoCall, messageTemplate, templateArgs).stream().flatMap(Optional::stream)
```
lbschanno: I think we're still introducing an unnecessary overhead of creating a new list of non-null optionals that is then passed to `getBestDaoResponse()`. My suggestion in [this comment](https://github.com/NationalSecurityAgency/datawave/pull/3708#discussion_r3623422326) for changes to `getBestDaoResponse()` would remove the need to create a new list entirely and allow us to directly pass in the `callDaos()` result.

Alternatively, a way to make this stream a bit more efficient would be use `distinct()` paired with `limit()` to short-circuit the streaming as soon as we see two different elements:

```java
List<T> distinct = callDaos(daoCall, messageTemplate, templateArgs).stream()
                .flatMap(Optional::stream)
                .distinct()
                .limit(2)
                .collect(Collectors.toList());

if (distinct.isEmpty()) {
    return Optional.empty();
} else if (distinct.size() == 1) {
    return Optional.of(distinct.get(0));
} else {
    throw new RuntimeException("Conflicting federated results returned from multiple data sources");
}
```

lbschanno (changes requested):

core/annotation/src/main/java/datawave/annotation/data/v1/FederatedAnnotationReader.java
drewfarris: Thanks for being persistent about this and catching that I'd missed this. Will address. I like the limit approach.

core/annotation/src/test/java/datawave/annotation/test/v1/AnnotationAssertions.java
drewfarris: Thanks for the suggestion @lbschanno. Take a look at #3761 for these improvements. There were a couple other cases in AnnotationAssertions that benefited from a similar approach.

core/annotation/src/main/java/datawave/annotation/data/v1/FederatedAnnotationReader.java
drewfarris: I tried an approach that eliminates the collect entirely, let me know what you think of the latest revision.

core/annotation/src/main/java/datawave/annotation/data/v1/FederatedAnnotationReader.java
```diff
@@ -0,0 +1,285 @@
...
+     *            the type of object in the results to return as the optional.
+     * @exception RuntimeException
+     *                if there is more than one result that can't be reconciled.
+     */
+    protected static <T> Optional<T> getBestDaoResponse(Stream<T> results) {
+        // Multiple DAOs can return equivalent copies for the same lookup key.
+        // For annotationId retrievals, the annotation identifier is derived from the content,
+        // so all instances that have the same id should be equal and returning one copy is sufficient.
+        // If this is not the case, we have a more significant runtime problem and when reduce is called
+        // a exception will be thrown.
```
lbschanno:
```suggestion
        // an exception will be thrown.
```

lbschanno (changes requested):

core/annotation/src/main/java/datawave/annotation/data/v1/FederatedAnnotationReader.java
lbschanno: I like it! Very clean.

lbschanno (dismissed):

tharke13 (dismissed):

drewfarris:
@lbschanno @tharke13 I manually rebased to resolve conflicts and sqashed to a single commit. It would be awesome if you both could re-approve. Thank you!

lbschanno (dismissed):

tharke13 (approved):

drewfarris:
Hi @lbschanno I had to tweak something as a result of a failed rebase via the GitLab UI - and lost your approval. If you could re-approve, I'd appreciate it!

lbschanno (approved):

PR #3710, opened 2026-07-21
Integrate annotation service into Datawave codebase

drewfarris:
The annotation service is a microservice for reading annotations and truthmarks, it is designed as a replacement/alternative for the AnnotationManagerBean, currently deployed via wildfly. 

A key tennent of the annotation service is durable annotation writes. It follows the patterns established in the audit service to achieve this and implements write targets for Accumulo, files and logs.


microservices/services/annotation/service/src/main/java/datawave/microservice/annotation/service/AnnotationControllerV1.java
```diff
@@ -0,0 +1,905 @@
...
+    @PutMapping(path = "/{idType}/{id}/annotation/{annotationId}", consumes = MediaType.APPLICATION_JSON_VALUE)
+    @RolesAllowed({"AnnotationWriter"})
+    public ResponseEntity<?> updateAnnotation(@PathVariable String idType, @PathVariable String id, @PathVariable String annotationId, @RequestBody String body,
+                    @RequestParam MultiValueMap<String,String> queryParameters, @AuthenticationPrincipal DatawaveUserDetails currentUser) {
+        final RequestContext context = new RequestContext(queryParameters, currentUser, annotationProperties, connectionFactory, accumuloConnectionRequestMap,
+                        visibilityTransformer, timestampTransformer);
+
+        try {
+            final Annotation rawAnnotation = AnnotationJsonUtils.annotationFromJson(body);
+            final Validator.ValidationState<Annotation> validationState = AnnotationValidators.checkAnnotation(rawAnnotation);
```
tharke13: Should this call `AnnotationValidators.checkAnnotationUpdate(rawAnnotation)` ?

microservices/services/annotation/service/src/main/java/datawave/microservice/annotation/service/AnnotationControllerV1.java
```diff
@@ -0,0 +1,905 @@
...
+        return result;
+    }
+
+    private Optional<Annotation> sendMessage(AnnotationMessage annotationMessage) {
+        String annotationId = annotationMessage.getAnnotation().getAnnotationId();
+
+        boolean success;
+        if (annotationProperties.isAnnotationAckEnabled()) {
+            final CountDownLatch latch = new CountDownLatch(1);
+            correlationLatchMap.put(annotationId, latch);
```
tharke13: This is probably unlikely but do we need to protect against a client trying to write the same annotation twice? Otherwise could the put overwrite an existing latch preventing acknowledgment? Maybe just reuse the existing latch and don't send again if it's already in the map?

microservices/services/annotation/service/src/main/java/datawave/microservice/annotation/health/rabbit/RabbitHealthChecker.java
FineAndDandy, line 1: I like this, but I'm not entirely sold on it belonging here. This feels like it should be running independently.

microservices/services/annotation/service/src/main/java/datawave/microservice/annotation/health/OutagesEndpoint.java
FineAndDandy, line 1: This is great, even if the health checker maybe should live somewhere else

microservices/services/annotation/service/src/main/java/datawave/microservice/annotation/util/lookup/common/LookupRequest.java
FineAndDandy, line 1: Longer term may want to migrate this to a RemoteHttpService

microservices/services/annotation/service/src/main/java/datawave/microservice/annotation/writers/accumulo/config/AccumuloAnnotationWriterProperties.java
```diff
@@ -0,0 +1,47 @@
...
+    private Accumulo accumuloConfig = new Accumulo();
+    private Health health = new Health();
+
+    @Getter
+    @Setter
+    public static class Accumulo {
+        private String zookeepers;
+        private String instanceName;
+        private String username;
+        private String password;
```
FineAndDandy, line 30: just want to flag this to make sure its consistent with how we might be handling it elsewhere

microservices/services/annotation/service/src/main/java/datawave/microservice/annotation/writers/accumulo/health/AccumuloHealthChecker.java
FineAndDandy, line 1: love this too

microservices/services/annotation/api/src/main/java/datawave/microservice/annotation/common/AnnotationMessage.java
FineAndDandy, line 1: In the ideal world this is protobuf as well. Otherwise serializing it along with the embedded Annotation is sorta nasty

microservices/services/annotation/dependency-tree.txt
FineAndDandy, line 1: did you intend to add this?

microservices/services/annotation/pom-original.xml
FineAndDandy, line 1: was this intended to stay?

FineAndDandy (changes requested):

microservices/services/annotation/pom-original.xml
drewfarris: No, I should remove this.

microservices/services/annotation/dependency-tree.txt
drewfarris: I'm feeling a bit mixed about this one. It could probably be removed, but I've found it useful when upgrading dependencies to see what's changed between the current version and the baseline version using `git diff`. That said, it should be trivial enough to capture a baseline before making changes.

drewfarris:
Broken into two separate PR's #3893 #3894

PR #3719, opened 2026-07-22
Fix CSVHelper field filtering, separators, and setup reuse

### LINKED ISSUE #3717: Bug (CSVHelper) - populates field disallowlist from allowlist configuration
## Issue
Dissallowlist contains Allowlist, due to CSVHelper reference issue

## Problem

`CSVHelper.setup(Configuration)` reads the field disallowlist into `cb`. However, it builds `fieldDisallowlist` from `cw`, which contains the allowlist:

https://github.com/NationalSecurityAgency/datawave/blob/02f0ed198f224d91e02c95ab9f6c1ef15fd2ef77/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CSVHelper.java#L182-L192

```java
Collection<String> cw =
        config.getStringCollection(this.getType().typeName() + FIELD_ALLOWLIST);
if (cw != null && !cw.isEmpty()) {
    this.fieldAllowlist = new HashSet<>(cw);
}

Collection<String> cb =
        config.getStringCollection(this.getType().typeName() + FIELD_DISALLOWLIST);
if (cb != null && !cb.isEmpty()) {
    this.fieldDisallowlist = new HashSet<>(cw);
}
```

The last line should likely use `cb`:

```java
this.fieldDisallowlist = new HashSet<>(cb);
```

## Expected behavior

Fields listed in `<datatype>.data.field.disallowlist` should be removed during ingest.

## Actual behavior

The code uses the allowlist values as the disallowlist.

This can cause two problems:

1. If only a disallowlist is set, the listed fields may not be removed.
2. If both lists are set, fields in the allowlist may be removed. The disallowlist takes priority in `keepField()`.

The filtering code is here:

https://github.com/NationalSecurityAgency/datawave/blob/02f0ed198f224d91e02c95ab9f6c1ef15fd2ef77/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/CSVIngestHelper.java#L258-L280

## Affected helpers

Because these classes extend `CSVHelper`, they may also be affected:

- `ExtendedCSVHelper`
- `JsonDataTypeHelper`
- `NYCTLCHelper`

This bug can cause fields to be kept or removed by mistake. It may also affect systems that use the disallowlist to prevent some data from being stored. The security impact depends on each system’s setup and access rules.

## Suggested fix

Change:

```java
this.fieldDisallowlist = new HashSet<>(cw);
```

to:

```java
this.fieldDisallowlist = new HashSet<>(cb);
```

## Test

The current `CSVIngestHelperTest` does not test allowlist or disallowlist behavior:

https://github.com/NationalSecurityAgency/datawave/blob/02f0ed198f224d91e02c95ab9f6c1ef15fd2ef77/warehouse/ingest-core/src/test/java/datawave/ingest/data/config/ingest/CSVIngestHelperTest.java

### LINKED ISSUE #3720: Bug (CSVHelper) - multivalue separator used as unescaped regex
## Issue

`CSVHelper` uses the configured multi-value separator as regex text.

## Problem

`CSVHelper.getEscapeSafeMultiValueSeparatorPattern()` appends the configured separator directly to a regex pattern:

https://github.com/NationalSecurityAgency/datawave/blob/4d52b3b16bef79ebaccd6fccd8eec8a18e0b04a5/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CSVHelper.java#L315-L316

`CSVHelper.cleanEscapedMultivalueSeparators()` also passes the configured separator directly into `replaceAll()`:

https://github.com/NationalSecurityAgency/datawave/blob/4d52b3b16bef79ebaccd6fccd8eec8a18e0b04a5/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CSVHelper.java#L374-L377

These methods are used by `CSVIngestHelper` when splitting and cleaning multi-value fields:

https://github.com/NationalSecurityAgency/datawave/blob/4d52b3b16bef79ebaccd6fccd8eec8a18e0b04a5/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/CSVIngestHelper.java#L181

https://github.com/NationalSecurityAgency/datawave/blob/4d52b3b16bef79ebaccd6fccd8eec8a18e0b04a5/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/CSVIngestHelper.java#L253

If `<datatype>.data.multivalued.separator` is set to a regex character such as `.`, `|`, `+`, `*`, `[`, or `\`, the separator may not be handled as plain text.

## Expected behavior

A configured multi-value separator should be treated as a literal separator.

## Actual behavior

The separator is used as regex text.

This can cause fields to split incorrectly, escaped separators to clean incorrectly, or regex errors for some separator values.

## Affected helpers

Because these classes extend `CSVHelper`, they may also be affected:

- `ExtendedCSVHelper`
- `JsonDataTypeHelper`
- `NYCTLCHelper`

## Suggested fix

Quote the configured separator before using it in regex operations, or avoid regex-based replacement for escaped separator cleanup.

## Test

The current upstream test tree does not appear to test this behavior.

There is no `CSVHelperTest.java` in upstream `integration`. The current `CSVIngestHelperTest` only tests `processExtraField()` with a null field:

https://github.com/NationalSecurityAgency/datawave/blob/4d52b3b16bef79ebaccd6fccd8eec8a18e0b04a5/warehouse/ingest-core/src/test/java/datawave/ingest/data/config/ingest/CSVIngestHelperTest.java#L8-L14

### LINKED ISSUE #3725: Bug(CSVHelper.setup()) - not idempotent on reused helper instance
## Issue

`CSVHelper.setup(Configuration)` is not idempotent when the same helper instance is reused.

## Problem

`CSVHelper` keeps setup state in instance fields. Some values are only replaced when the next configuration contains that setting, while multi-value field maps keep old entries unless they are cleared:

https://github.com/NationalSecurityAgency/datawave/blob/02d686786e11e2ba1f6e21a8dc19d7575c122c19/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CSVHelper.java#L136-L247

The setup state also includes values outside `CSVHelper` itself.

`DataTypeHelperImpl.fieldsToDowncase` is populated while the type is initialized. Once the type and `TypeRegistry` exist, type initialization returns early and the downcase fields are not rebuilt for the next configuration:

https://github.com/NationalSecurityAgency/datawave/blob/02d686786e11e2ba1f6e21a8dc19d7575c122c19/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/DataTypeHelperImpl.java#L29-L31

https://github.com/NationalSecurityAgency/datawave/blob/02d686786e11e2ba1f6e21a8dc19d7575c122c19/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/DataTypeHelperImpl.java#L62-L73

`FieldNameAliaserNormalizer.setup()` rebuilds the raw alias maps, but it does not clear the cached exact and wildcard aliases. If those caches were used before the second setup, aliases from the first configuration can remain active:

https://github.com/NationalSecurityAgency/datawave/blob/02d686786e11e2ba1f6e21a8dc19d7575c122c19/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/FieldNameAliaserNormalizer.java#L34-L36

https://github.com/NationalSecurityAgency/datawave/blob/02d686786e11e2ba1f6e21a8dc19d7575c122c19/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/FieldNameAliaserNormalizer.java#L59-L149

`ExtendedCSVHelper.setup()` also retains subclass-owned state. The old ID downcase value and ignored fields are used as defaults, while security marking mappings and ID parsers are added to their existing collections:

https://github.com/NationalSecurityAgency/datawave/blob/02d686786e11e2ba1f6e21a8dc19d7575c122c19/warehouse/ingest-csv/src/main/java/datawave/ingest/csv/config/helper/ExtendedCSVHelper.java#L70-L106

https://github.com/NationalSecurityAgency/datawave/blob/02d686786e11e2ba1f6e21a8dc19d7575c122c19/warehouse/ingest-csv/src/main/java/datawave/ingest/csv/config/helper/ExtendedCSVHelper.java#L135-L141

## Expected behavior

Calling `setup(Configuration)` should make the helper match the supplied configuration.

After setup with configuration B, a reused helper should have the same state and behavior as a new helper configured only with B. This includes state owned by `CSVHelper`, inherited downcase and alias settings, and state owned by `ExtendedCSVHelper`.

## Actual behavior

Old setup state can remain on a reused helper.

This can cause fields to be kept, removed, split, renamed, downcased, aliased, ignored, or thresholded using stale configuration. An `ExtendedCSVHelper` can also retain an earlier ID downcase setting, security marking mapping, or ID parser.

## Affected helpers

Because these classes use or extend `CSVHelper`, they may also be affected:

- `ExtendedCSVHelper`
- `JsonDataTypeHelper`
- `NYCTLCHelper`

## Suggested fix

Reset mutable `CSVHelper` state before reading each configuration.

Rebuild inherited `fieldsToDowncase` from the current configuration on every setup while keeping the helper bound to its existing datatype. Clear the exact and wildcard alias caches when alias setup runs again.

`ExtendedCSVHelper.setup()` should also reset its ID downcase flag, security marking map, ignored fields, and ID parser map before reading the new configuration.

## Test

Use focused tests that call `setup()` twice on the same objects without resetting `TypeRegistry` between calls.

- Verify inherited downcase fields are replaced by the second configuration.
- Populate exact and wildcard alias caches, run setup again, and verify both caches use the second configuration.
- Configure `ExtendedCSVHelper` with ID downcasing, a security marking, ignored fields, and an ID parser. Run setup again without those settings and verify none of the old values remain.
- Keep the existing `CSVHelper` reuse test for allowlists, disallowlists, required fields, multi-value settings, thresholds, and threshold actions.

### LINKED ISSUE #3768: Bug (CSVHelper) - empty multivalue separator corrupts field processing
## Issue

`CSVHelper` accepts an empty multi-value separator.

## Problem

`CSVHelper.setup(Configuration)` reads `<datatype>.data.multivalued.separator` without checking whether it is empty:

https://github.com/NationalSecurityAgency/datawave/blob/02d686786e11e2ba1f6e21a8dc19d7575c122c19/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CSVHelper.java#L232

```java
this.multiValueSeparator = config.get(this.getType().typeName() + MULTI_VALUED_SEPARATOR, ";");
```

The separator is later used to build the split pattern and remove escape characters:

https://github.com/NationalSecurityAgency/datawave/blob/02d686786e11e2ba1f6e21a8dc19d7575c122c19/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CSVHelper.java#L311-L317

https://github.com/NationalSecurityAgency/datawave/blob/02d686786e11e2ba1f6e21a8dc19d7575c122c19/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CSVHelper.java#L367-L379

With an empty separator, the split pattern matches between characters. A value such as `abc` can be split into `a`, `b`, and `c`. Escape cleanup can also remove backslashes from field values.

## Expected behavior

`CSVHelper.setup(Configuration)` should reject an empty multi-value separator.

The default `;` should still be used when the setting is absent.

## Actual behavior

The empty value is accepted and can split fields at every character or remove backslashes.

## Affected helpers

Ingest helpers that use `CSVHelper` multi-value processing may also be affected, including:

- `CSVIngestHelper`
- `ExtendedCSVIngestHelper`

## Suggested fix

Check the separator during `CSVHelper.setup(Configuration)` and throw `IllegalArgumentException` when it is empty.

The error should name the invalid configuration key.

## Test

The current upstream tests do not cover an empty `<datatype>.data.multivalued.separator` value.

Add a setup test that confirms an empty separator is rejected.

### LINKED ISSUE #3769: Bug (CSVHelper) - negative thresholds cause invalid ingest behavior
## Issue

`CSVHelper` accepts negative field and multi-value thresholds.

## Problem

`CSVHelper.setup(Configuration)` reads both threshold values without checking whether they are negative:

https://github.com/NationalSecurityAgency/datawave/blob/02d686786e11e2ba1f6e21a8dc19d7575c122c19/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CSVHelper.java#L234-L240

```java
this.fieldSizeThreshold =
        config.getInt(this.getType().typeName() + FIELD_SIZE_THRESHOLD, this.fieldSizeThreshold);

this.multiFieldSizeThreshold =
        config.getInt(this.getType().typeName() + MULTI_VALUED_THRESHOLD, this.multiFieldSizeThreshold);
```

A negative field-size threshold makes every field exceed the limit. With the `TRUNCATE` action, the code can call `substring(0, -1)` and throw `StringIndexOutOfBoundsException`:

https://github.com/NationalSecurityAgency/datawave/blob/02d686786e11e2ba1f6e21a8dc19d7575c122c19/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/CSVIngestHelper.java#L199-L213

A negative multi-value threshold is never reached because the value counter starts at zero and is only compared for equality. This silently disables the limit:

https://github.com/NationalSecurityAgency/datawave/blob/02d686786e11e2ba1f6e21a8dc19d7575c122c19/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/CSVIngestHelper.java#L174-L190

## Expected behavior

`CSVHelper.setup(Configuration)` should reject threshold values below zero.

A value of zero should remain valid.

## Actual behavior

Negative values are accepted.

This can disable the multi-value limit or cause field processing to fail with an unrelated index error.

## Affected helpers

Ingest helpers that use `CSVHelper` threshold processing may also be affected, including:

- `CSVIngestHelper`
- `ExtendedCSVIngestHelper`

## Suggested fix

Validate both threshold values during `CSVHelper.setup(Configuration)`.

Throw `IllegalArgumentException` when either value is below zero, and include the invalid key and value in the message.

## Test

The current upstream tests do not cover negative values for:

- `<datatype>.data.field.length.threshold`
- `<datatype>.data.multivalued.threshold`

Add setup tests that confirm negative values are rejected and zero remains valid.

### LINKED ISSUE #3770: Bug (CSVHelper) - threshold action parsing depends on JVM locale
## Issue

`CSVHelper.setup(Configuration)` can reject a valid lowercase threshold action when the JVM uses the Turkish locale.

## Problem

`CSVHelper` changes both configured threshold actions to uppercase before calling `ThresholdAction.valueOf(...)`:

https://github.com/NationalSecurityAgency/datawave/blob/02d686786e11e2ba1f6e21a8dc19d7575c122c19/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/CSVHelper.java#L234-L242

```java
this.thresholdAction = ThresholdAction.valueOf(
        config.get(this.getType().typeName() + THRESHOLD_ACTION,
                        this.thresholdAction.name()).toUpperCase());

this.multiValuedThresholdAction = ThresholdAction.valueOf(
        config.get(this.getType().typeName() + MULTI_VALUED_THRESHOLD_ACTION,
                        this.multiValuedThresholdAction.name()).toUpperCase());
```

`String.toUpperCase()` uses the JVM default locale when no locale is supplied.

Under the Turkish locale, `fail` becomes `FA?L` with a dotted capital `?`. This value does not match the `FAIL` enum value, so setup throws `IllegalArgumentException`.

This affects both settings:

- `<datatype>.data.threshold.action`
- `<datatype>.data.multivalued.threshold.action`

## Expected behavior

A valid threshold action should work the same way under every JVM language and region setting.

For example, lowercase `fail` should select `ThresholdAction.FAIL`.

## Actual behavior

Under the Turkish locale, lowercase `fail` is changed to `FA?L`.

`ThresholdAction.valueOf(...)` cannot find that enum value, and helper setup fails before ingest starts.

## Affected helpers

Because these classes extend `CSVHelper`, they may also be affected:

- `ExtendedCSVHelper`
- `JsonDataTypeHelper`
- `NYCTLCHelper`

## Suggested fix

Use `Locale.ROOT` when changing both threshold action values to uppercase:

```java
value.toUpperCase(Locale.ROOT)
```

## Test

The current upstream test tree does not test threshold action parsing under a non-English JVM locale.

There is no `CSVHelperTest.java` in upstream `integration`. The current `CSVIngestHelperTest` only tests a null extra field:

https://github.com/NationalSecurityAgency/datawave/blob/02d686786e11e2ba1f6e21a8dc19d7575c122c19/warehouse/ingest-core/src/test/java/datawave/ingest/data/config/ingest/CSVIngestHelperTest.java#L8-L14

arthurianresolve:
## Summary

- populate `CSVHelper.fieldDisallowlist` from the configured disallowlist instead of the allowlist
- treat configured multi-value separators as literal text, including backslash when it is both the separator and escape character
- reject an explicit empty multi-value separator while preserving the default `;` when unset
- clean escaped multi-value separators with literal replacement instead of regex replacement
- reset `CSVHelper.setup()` state before reading a new configuration
- rebuild inherited downcase fields and exact or wildcard alias caches during repeated setup
- reset `ExtendedCSVHelper` state so reused instances do not retain ID downcasing, security mappings, ignored fields, or ID parsers
- reject negative CSV field and multi-value thresholds during setup
- parse CSV threshold actions with `Locale.ROOT` so valid lowercase actions do not depend on the JVM default locale
- add focused owner-level regression coverage

Fixes [Bug (CSVHelper) - populates field disallowlist from allowlist configuration #3717](https://github.com/NationalSecurityAgency/datawave/issues/3717).
Fixes [Bug (CSVHelper) - multivalue separator used as unescaped regex #3720](https://github.com/NationalSecurityAgency/datawave/issues/3720).
Fixes [Bug(CSVHelper.setup()) - not idempotent on reused helper instance #3725](https://github.com/NationalSecurityAgency/datawave/issues/3725).
Fixes [Bug (CSVHelper) - empty multivalue separator corrupts field processing #3768](https://github.com/NationalSecurityAgency/datawave/issues/3768).
Fixes [Bug (CSVHelper) - negative thresholds cause invalid ingest behavior #3769](https://github.com/NationalSecurityAgency/datawave/issues/3769).
Fixes [Bug (CSVHelper) - threshold action parsing depends on JVM locale #3770](https://github.com/NationalSecurityAgency/datawave/issues/3770).

## Implementation commits

- [`65d923e49` - Fix CSVHelper field filtering and separators](https://github.com/arthurianresolve/datawave/commit/65d923e49eef6751d8e97d8c7280c14ed947e298)
- [`0d5e04631` - Fix CSVHelper setup idempotency](https://github.com/arthurianresolve/datawave/commit/0d5e046319d4e40c3664ae209cc00efd8ac73155)
- [`187f7a15e` - Fix backslash CSV multivalue separators (#3720)](https://github.com/arthurianresolve/datawave/commit/187f7a15ee9cfbeb97f5dc4900dedc4e36c2dad2)
- [`e4297d20b` - Complete CSV helper setup idempotency (#3725)](https://github.com/arthurianresolve/datawave/commit/e4297d20b2103e98e216652340487d8bb990996e)
- [`d59e1b6ec` - Validate CSVHelper threshold settings (#3769 #3770)](https://github.com/arthurianresolve/datawave/commit/d59e1b6ec89c5b6c5d088993a38f52bcc995081b)
- [`486be700b` - Reject empty CSVHelper multivalue separator (#3768)](https://github.com/arthurianresolve/datawave/commit/486be700b110c5cc6c4d2b52e7a3e7dea020a6f9)

## Root cause and impact

`CSVHelper.setup()` read the disallowlist into `cb` but copied `cw` into `fieldDisallowlist`. A disallowlist configured without an allowlist could therefore fail to remove matching fields. When both lists were set, allowlisted fields could be removed because disallowlist checks take priority.

`CSVHelper.getEscapeSafeMultiValueSeparatorPattern()` and `CSVHelper.cleanEscapedMultivalueSeparators()` also used the configured separator as regex text. Regex metacharacters could split or clean fields incorrectly or throw regex errors. Backslash needed separate handling because it is both the separator and the escape character: odd-length runs contain a delimiter, while even-length runs contain only escaped separators.

An explicitly empty multi-value separator was also accepted. That can create a split pattern that matches between characters and can make escape cleanup remove backslashes from field values.

`CSVHelper.setup()` retained optional state on a reused helper instance when a later configuration omitted those options. The original reset covered fields owned by `CSVHelper`, but inherited `DataTypeHelperImpl.fieldsToDowncase` entries could still remain.

The inherited alias normalizer had the same problem. It rebuilt raw aliases but kept its lazy exact and wildcard caches. `ExtendedCSVHelper` also reused its current ID downcase setting and ignored fields as defaults, and it added security mappings and ID parsers without clearing the old collections.

`CSVHelper.setup()` also accepted negative threshold values. Negative field-size limits can make every field exceed the limit and can make truncation fail with an unrelated substring index error. Negative multi-value limits are never reached by the equality check, which disables the configured limit.

Finally, threshold action parsing used `String.toUpperCase()` without a locale. Under locales such as Turkish, valid lowercase values like `fail` can uppercase to a string that no longer matches the enum name.

The completed fix rebuilds inherited downcase fields on every setup, clears both alias caches, resets subclass-owned `ExtendedCSVHelper` state before reading the next configuration, validates non-empty multi-value separators, rejects negative thresholds with the offending key and value, and parses threshold actions with `Locale.ROOT`. Cached datatype binding is unchanged.

## Test

```shell
mvn --batch-mode --errors --no-transfer-progress -Dmaven.build.cache.enabled=false -pl warehouse/ingest-core,warehouse/ingest-csv -am -Dtest=DataTypeHelperImplTest,FieldNameAliaserNormalizerTest,CSVHelperTest,ExtendedCSVHelperTest -Dsurefire.failIfNoSpecifiedTests=false -DskipITs -DskipFormat test
```

Result: BUILD SUCCESS across 25 reactor projects in 4:46.

- `DataTypeHelperImplTest`: 3 tests, 0 failures, 0 errors, 0 skipped
- `FieldNameAliaserNormalizerTest`: 3 tests, 0 failures, 0 errors, 0 skipped
- `CSVHelperTest`: 6 tests, 0 failures, 0 errors, 0 skipped
- `ExtendedCSVHelperTest`: 1 test, 0 failures, 0 errors, 0 skipped
- Total selected tests: 13 tests, 0 failures, 0 errors, 0 skipped
- Checkstyle: 0 violations in changed modules
- `git diff --check`: passes

The setup reuse coverage verifies `CSVHelper`-owned state, inherited downcase fields, exact and wildcard alias caches, and `ExtendedCSVHelper` state. The separator regression covers delimiter and escaped-separator backslash runs of lengths one through four, and now also rejects an explicit empty multi-value separator while preserving the default separator when unset. The threshold coverage verifies zero thresholds remain valid, negative field and multi-value thresholds are rejected, and lowercase threshold actions parse under a Turkish default JVM locale.

The optional build cache was disabled only for the Windows verification run because its `**/*.log` exclusion is rejected as a Windows path; no build-cache configuration was changed.


warehouse/ingest-core/src/test/java/datawave/ingest/data/config/CSVHelperTest.java
arthurianresolve, line 1: CSVHelperTest verifies allowlist and disallowlist configuration populate separate CSVHelper fields to avoid regression to the current critical issue where values get populated into both causing disfunction of disallowlist

arthurianresolve:
@apmoriarty @alerman @avgAGB I would appreciate if if you could have a look at these. I have 3 larger PRs with over 20 bug fixes ready as PR drafts and cannot push them for review at the moment due to 2 PR restriction

arthurianresolve:

PR #3733, opened 2026-07-23
migrate scanner to scannerbuilder

apmoriarty:
Migrate most instances of Scanner to using a ScannerBuilder. Another brick in the road to native scan server support.


warehouse/query-core/src/main/java/datawave/query/index/lookup/RangeStream.java
```diff
@@ -937,27 +950,47 @@ private void populateCache() {
...
-                        continue;
-                    }
 
-                    scannedKeys++;
-                    String[] parts = cq.split("_");
-                    cache.put(parts[0], Integer.parseInt(parts[1]));
+            try {
+                Authorizations auths = client.securityOperations().getUserAuthorizations(client.whoami());
+
+                // @foramtter:off
```
lbschanno:
```suggestion
                // @formatter:off
```

lbschanno (changes requested):

lbschanno:
Food for thought for perhaps a new issue: I'm seeing Authorizations.EMPTY used in a number of places when creating a scanner. Would it make sense at all to make `datawave.scan.ScanBuilder` default to a value of a singleton list consisting of Authorizations.EMPTY for `ScanBuilder.authorizations`?

lbschanno:
> Food for thought for perhaps a new issue: I'm seeing Authorizations.EMPTY used in a number of places when creating a scanner. Would it make sense at all to make `datawave.scan.ScanBuilder` default to a value of a singleton list consisting of Authorizations.EMPTY for `ScanBuilder.authorizations`?

Alternatively, perhaps if authorizations is null when `ScannerBuilder.build()` is called, but client is not, the authorizations could default to `client.securityOperations().getUserAuthorizations(client.whoami())`.

warehouse/query-core/src/main/java/datawave/next/scanner/DocumentIdProducer.java
```diff
@@ -106,18 +106,20 @@ private Scanner createScanner() throws TableNotFoundException {
...
+        //  @formatter:off
+        ScannerBuilder builder = ScannerBuilder.create(config.getClient())
+                .setTableName(context.getTableName())
+                .setAuthorizations(config.getAuthorizations())
+                .setConsistencyLevel(config.getSearchConsistencyLevel())
+                .setScanType(config.getSearchScanHintPool())
+                .setScanPriority(1);
+        //  @formatter:on
 
-        if (config.getSearchConsistencyLevel() != null) {
```
lbschanno, line 118: Are we certain that `config.getSearchConsistencyLevel()` won't return null? The ScanBuilder will throw an IllegalArgumentException if null is provided to `setConsistencyLevel()`. Perhaps the ScanBuilder should allow null to be set for optional variables. We could then avoid having to do all of these 
```java
if (optionalVar != null) {
    builder.setOptionalVar(optionalVar);
}
```
checks.

apmoriarty: This is by design, I want the DocumentScheduler to be opinionated about configuring the consistency level. It's also a preview for where the project is going as a whole WRT scanners

apmoriarty:
> > Food for thought for perhaps a new issue: I'm seeing Authorizations.EMPTY used in a number of places when creating a scanner. Would it make sense at all to make `datawave.scan.ScanBuilder` default to a value of a singleton list consisting of Authorizations.EMPTY for `ScanBuilder.authorizations`?
> 
> Alternatively, perhaps if authorizations is null when `ScannerBuilder.build()` is called, but client is not, the authorizations could default to `client.securityOperations().getUserAuthorizations(client.whoami())`.

Replying to both comments:

In practice nearly every key has a visibility applied, so I'm not sure what an empty authorizations is doing besides generating spurious scans. I'm purposely not addressing this as part of the migration work.

The datawave user has the full set of scan auths, but is almost certainly performing a scan on behalf of a user, who almost certainly has a subset of the auths. Opening up the scan auths would violate a core principle of a multi-tenant architecture. A user who is only authorized to see VIZ-A is now seeing VIZ-B in their results.

avgAGB (approved):

alerman:
@gemini-cli /review

lbschanno (approved):

PR #3739, opened 2026-07-23
Fix intermediate-result page flood in GroupingTransform/UniqueTransform

apmoriarty:
## Summary
- `GroupingTransform` and `UniqueTransform` never reset their per-page timer after emitting an intermediate PARTIAL page, so once the page timeout elapsed every subsequent `apply()` call emitted another intermediate result, flooding `RunningQuery`'s async results thread instead of pacing to one per timeout window.
- Page timing now goes through an injectable `java.time.Clock` rather than calling `System.currentTimeMillis()` directly.

## Test plan
- [x] Added a regression test to `UniqueTransformTest` asserting intermediate results are paced (one per timeout window) rather than flooding
- [x] Added `GroupingTransformTest` with the equivalent regression test for `GroupingTransform` (no prior test class existed)
- [x] `mvn -pl warehouse/query-core -Dtest="datawave.query.transformer.**,datawave.query.common.grouping.**" test` passes

### DIFF
--- a/warehouse/query-core/src/test/java/datawave/query/transformer/UniqueTransformMostRecentTest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/transformer/UniqueTransformMostRecentTest.java
@@ -68,8 +68,8 @@ public class UniqueTransformMostRecentTest extends UniqueTransformTest {
     }
 
     /**
-     * The mostRecent path never returns a "real" unique result and currently never emits an intermediate result at all, so the base class's
-     * page-timer-reset assertions do not apply here.
+     * The mostRecent path never returns a "real" unique result and currently never emits an intermediate result at all, so the base class's page-timer-reset
+     * assertions do not apply here.
      */
     @Override
     @Ignore
--- a/warehouse/query-core/src/test/java/datawave/query/transformer/UniqueTransformMostRecentTest.java
+++ b/warehouse/query-core/src/test/java/datawave/query/transformer/UniqueTransformMostRecentTest.java
@@ -72,8 +72,22 @@ public class UniqueTransformMostRecentTest extends UniqueTransformTest {
      * assertions do not apply here.
      */
     @Override
+    @Test
     @Ignore
-    public void testIntermediateResultsArePaced_afterPageTimerReset() {}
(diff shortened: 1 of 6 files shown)


warehouse/query-core/src/test/java/datawave/query/transformer/UniqueTransformMostRecentTest.java
```diff
@@ -66,6 +67,14 @@ protected UniqueTransform getUniqueTransform() {
...
     }
 
+    /**
+     * The mostRecent path never returns a "real" unique result and currently never emits an intermediate result at all, so the base class's page-timer-reset
+     * assertions do not apply here.
+     */
+    @Override
+    @Ignore
+    public void testIntermediateResultsArePaced_afterPageTimerReset() {}
+
```
FineAndDandy, line 91: please add reference to issue that tracks this here as a comment so we can make sure the test gets implemented with it

warehouse/query-core/src/test/java/datawave/query/transformer/UniqueTransformTest.java
```diff
@@ -91,6 +92,41 @@ public void testTransformingNullReturnsNull() {
         assertNull(uniqueTransform.apply(null));
     }
 
+    /**
+     * Once the per-page timeout is exceeded, apply() must reset the page timer when it emits an intermediate result, so intermediate results are paced (one per
+     * timeout window) instead of being emitted on every subsequent call.
+     */
+    @Test
+    public void testIntermediateResultsArePaced_afterPageTimerReset() throws IOException {
```
FineAndDandy, line 107: can you add one more test that shows the recovery after a timeout that emits a real page again later?

FineAndDandy (changes requested):

warehouse/query-core/src/test/java/datawave/query/transformer/UniqueTransformTest.java
apmoriarty: Added mirror test for grouping

hlgp (approved):

FineAndDandy (approved):

PR #3740, opened 2026-07-23
Emit intermediate results on the mostRecent-unique path

apmoriarty:
## Summary
- `UniqueTransform.apply()` returned `null` immediately after updating the mostRecent-unique backing map, short-circuiting before the timeout/intermediate-result check. A long-running mostRecent-unique accumulation therefore never emitted a keep-alive intermediate page.
- Falls through to the existing timeout check instead of returning early, so the mostRecent path can emit intermediate results just like the standard unique path.

## Test plan
- [x] Added a regression test to the existing `UniqueTransformMostRecentTest` asserting an intermediate result is emitted once the per-page timeout is exceeded
- [x] `mvn -pl warehouse/query-core -Dtest="datawave.query.transformer.**,datawave.query.common.grouping.**" test` passes

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
+++ b/warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
@@ -112,18 +112,21 @@ public class UniqueTransform extends DocumentTransform.DefaultDocumentTransform 
     }
 
     /**
-     * Add phrase excerpts to the documents from the given iterator.
+     * Apply uniqueness to the documents from the given iterator. This is the tserver-side entry point, driven by the query iterator.
      *
      * @param in
      *            the iterator source
-     * @return an iterator that will supply the enriched documents
+     * @param yieldCallback
+     *            the yield callback
+     * @return an iterator that will supply the unique documents
      */
     public Iterator<Entry<Key,Document>> getIterator(final Iterator<Entry<Key,Document>> in, final YieldCallback yieldCallback) {
         return new UniqueTransformIterator(in, yieldCallback);
     }
 
     /**
-     * Apply uniqueness to a document.
+     * Apply uniqueness to a document. Called on the web server, as part of the {@link DocumentTransformer} chain, where an intermediate result is a meaningful
+     * keep-alive for the client waiting on the current page.
      *
      * @param keyDocumentEntry
      *            document entry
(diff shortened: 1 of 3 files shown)


warehouse/query-core/src/test/java/datawave/query/transformer/UniqueTransformMostRecentTest.java
```diff
@@ -86,4 +100,30 @@ public void testMostRecentUniqueness() {
...
+        uniqueTransform.setQueryExecutionForPageStartTime(clock.millis() - 10_000L);
+
+        int intermediate = 0;
+        for (Document document : inputDocuments) {
+            Map.Entry<Key,Document> result = uniqueTransform.apply(Maps.immutableEntry(document.getMetadata(), document));
+            if (result != null && result.getValue().isIntermediateResult()) {
+                intermediate++;
+            }
+        }
+        assertTrue("expected at least one intermediate result on the mostRecent path", intermediate >= 1);
```
avgAGB: small nit, but is it valuable to check that the document contents in the path match what is expected as well, in addition to the number of results being more than one?

avgAGB: @apmoriarty just checking what's your take on it

apmoriarty: That seems reasonable.

avgAGB (dismissed):

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -132,6 +135,22 @@ public Iterator<Entry<Key,Document>> getIterator(final Iterator<Entry<Key,Docume
...
+    }
+
+    /**
+     * Apply uniqueness to a document.
+     *
+     * @param keyDocumentEntry
+     *            document entry
+     * @param emitIntermediateResults
+     *            whether an intermediate result may be returned once the page timeout has elapsed. Only the web server may do so. The tserver has no page to
+     *            keep alive, {@link #queryExecutionForPageStartTime} is never set there, and {@link Document#isIntermediateResult()} is not carried across
```
ivakegg, line 148: I think I would rather fix the document serialization issue, and allow the tservers to return intermediate results.  Granted that this case should probably never be hit given we use yielding.  However yielding can basically be removed (setting the timeout really high) and hence we still need the intermediate result mechanism in place to avoid hanging scans.

ivakegg (changes requested):

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
```diff
@@ -154,19 +173,22 @@ public Entry<Key,Document> apply(@Nullable Entry<Key,Document> keyDocumentEntry)
...
             } catch (IOException ioe) {
                 log.error("Failed to convert document to bytes.  Returning document as unique.", ioe);
             }
 
-            long elapsedExecutionTimeForCurrentPage = System.currentTimeMillis() - this.queryExecutionForPageStartTime;
-            if (elapsedExecutionTimeForCurrentPage > this.queryExecutionForPageTimeout) {
-                Document intermediateResult = new Document();
-                intermediateResult.setIntermediateResult(true);
-                return Maps.immutableEntry(keyDocumentEntry.getKey(), intermediateResult);
+            if (emitIntermediateResults) {
```
ivakegg, line 185: I think this should always be in place for both the tserver and webserver sides.  Remove the emitIntermediateResults flag.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
apmoriarty: Wrote https://github.com/NationalSecurityAgency/datawave/issues/3796. The change is simple, but will likely trigger lots of test assertion updates which produces a sizeable diff.

warehouse/query-core/src/main/java/datawave/query/transformer/UniqueTransform.java
apmoriarty: This will be superseded by a unit of work that addresses the problem for grouping, count, and unique functions.

avgAGB (approved):

avgAGB:
not a blocker, but some feedback for a potential follow-up:
`not a blocker: queryExecutionForPageStartTime is read/written without synchronization in apply()
  while map access is explicitly synchronized — pre-existing pattern, not introduced by this change, but if DocumentTransformer.apply() can ever be
  invoked concurrently for one page you could see a stale timeout read. Not likely given the surrounding single-threaded page-processing design,
  but flagging since this branch is exactly what touches that timing logic. Could consider making the field volatile, or an AtomicLong.`

ivakegg (approved):

PR #3742, opened 2026-07-23
Fix TcpClient encoding and reconnect handling

### LINKED ISSUE #3730: Bug (TcpClient) - uses the platform default charset for Timely metrics
## Issue

`TcpClient` uses the JVM default charset when it sends metrics to Timely.

## Problem

`TcpClient.connect()` creates a `PrintWriter` without setting a charset:

https://github.com/NationalSecurityAgency/datawave/blob/436fc484454e2d9db86bd11c01a01c470deae56f/warehouse/core/src/main/java/datawave/util/timely/TcpClient.java#L82-L93

```java
out = new PrintWriter(sock.getOutputStream(), false);
```

This uses the platform default charset. On a system that does not use UTF-8, metric names or tag values containing non-ASCII characters can be sent with different bytes.

The other Timely writers already use UTF-8:

- https://github.com/NationalSecurityAgency/datawave/blob/436fc484454e2d9db86bd11c01a01c470deae56f/warehouse/core/src/main/java/datawave/util/timely/UdpClient.java#L27-L32
- https://github.com/NationalSecurityAgency/datawave/blob/436fc484454e2d9db86bd11c01a01c470deae56f/core/metrics-reporter/src/main/java/datawave/metrics/TimelyMetricsReporter.java#L198-L209

Issue #3596 and PR #3599 are related, but they only cover default charset use in ingest code. They do not change `TcpClient`.

https://github.com/NationalSecurityAgency/datawave/issues/3596

https://github.com/NationalSecurityAgency/datawave/pull/3599

## Expected behavior

TCP metrics sent to Timely should always use UTF-8, regardless of the JVM or operating system default charset.

## Actual behavior

`TcpClient` uses the platform default charset.

The query-metric service uses this client when Timely is configured for TCP:

https://github.com/NationalSecurityAgency/datawave/blob/436fc484454e2d9db86bd11c01a01c470deae56f/microservices/services/query-metric/service/src/main/java/datawave/microservice/querymetric/QueryMetricOperationsStats.java#L98-L105

## Suggested fix

Create the writer with UTF-8:

```java
out = new PrintWriter(new OutputStreamWriter(sock.getOutputStream(), UTF_8), false);
```

This matches `TimelyMetricsReporter`.

## Test

There is no focused `TcpClient` test.

Add a loopback TCP test that runs with a non-UTF-8 default charset.

Send a metric containing a non-ASCII value, call `flush()`, and confirm that the server receives the UTF-8 bytes.

### LINKED ISSUE #3731: Bug (TcpClient) - reconnect delay grows without a maximum
## Issue

`TcpClient` can wait for a very long time before reconnecting to Timely after repeated connection failures.

## Problem

`TcpClient.connect()` doubles its reconnect delay after every failed connection:

https://github.com/NationalSecurityAgency/datawave/blob/436fc484454e2d9db86bd11c01a01c470deae56f/warehouse/core/src/main/java/datawave/util/timely/TcpClient.java#L82-L96

```java
backoff = backoff * 2;
```

There is no maximum value.

After 12 failed connection attempts, the delay reaches `8,192,000` milliseconds, which is more than two hours. A temporary Timely outage can therefore delay metric reporting long after Timely becomes available again.

`TimelyMetricsReporter` already limits the reconnect delay to 120 seconds:

https://github.com/NationalSecurityAgency/datawave/blob/436fc484454e2d9db86bd11c01a01c470deae56f/core/metrics-reporter/src/main/java/datawave/metrics/TimelyMetricsReporter.java#L198-L210

```java
backoff = Math.min(backoff * 2, MAX_BACKOFF);
```

## Expected behavior

The reconnect delay should increase after failures but remain below a reasonable maximum.

A successful connection should reset the delay.

## Actual behavior

The reconnect delay doubles without a limit and can grow to minutes or hours.

## Affected components

The query-metric service uses `TcpClient` when Timely is configured for TCP.

## Suggested fix

Apply the same 120-second maximum used by `TimelyMetricsReporter`.

For example:

```java
private static final long MAX_BACKOFF = TimeUnit.SECONDS.toMillis(120);
```

```java
backoff = Math.min(backoff * 2, MAX_BACKOFF);
```

## Test

There is no focused test for `TcpClient` reconnect behavior.

Add a test with a controlled clock or extracted backoff calculation.

Check that:

- the delay increases after connection failures
- the delay never exceeds 120 seconds
- a successful connection resets the delay to 2 seconds

arthurianresolve:
## Summary

- Send Timely TCP metrics using UTF-8 instead of the platform default charset.
- Cap reconnect backoff at 120 seconds.
- Correct the malformed connection-error log message.

## Root cause

`TcpClient` created its writer without an explicit charset and doubled its reconnect delay without a maximum. Its error log also mixed SLF4J placeholders with string concatenation.

A successful connection already reset the delay to two seconds. This change retains that behavior and replaces the literal value with a named constant.

## Impact

TCP metric bytes are consistent across platforms, repeated failures cannot increase the reconnect delay beyond two minutes, and connection errors report the correct host, port, and message.

## Integration

This PR overlaps #3743 in `TcpClient`. Merge this PR first, then rebase #3743 and retain both this PR's UTF-8, capped-backoff, and logging changes and #3743's socket-lifecycle, connection-timeout, and write/flush-error changes.

## Test

```shell
mvn --batch-mode --errors --no-transfer-progress -Dmaven.build.cache.enabled=false -DskipFormat -DskipITs -pl warehouse/core -am -Dtest=TcpClientTest -Dsurefire.failIfNoSpecifiedTests=false test
```

Result: `Tests run: 2, Failures: 0, Errors: 0, Skipped: 0`.

The focused tests cover UTF-8 writer bytes and the maximum reconnect delay. They do not independently test progressive delay growth, the existing successful-connection reset, or log output.

The branch was then synced with current `integration`; the reviewed source and test blobs were unchanged by that sync.

Fixes #3730
Fixes #3731


arthurianresolve:

PR #3743, opened 2026-07-23
Harden Timely formatting, transport, and client lifecycle

### LINKED ISSUE #3729: Bug (Timely) - numeric metric values depend on JVM locale
## Issue

Timely metric values can change with the JVM default locale.

## Problem

`TimelyMetricsReporter` formats decimal values without setting a locale:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/core/metrics-reporter/src/main/java/datawave/metrics/TimelyMetricsReporter.java#L164-L175

```java
reportMetric(metricName, sampleName, String.format("%f", value), sampleType, units, time);
```

The same class formats integer values and complete Timely commands with the default locale:

```java
String.format("%d", value)
String.format("put %s %d %s ...", metricName, time, value, ...)
```

Some FORMAT locales use non-ASCII digits for `%d`. A timestamp or counter value can then contain Arabic or Persian digits.

`QueryMetricOperationsStats` also creates `DecimalFormat` instances without fixed decimal symbols:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/microservices/services/query-metric/service/src/main/java/datawave/microservice/querymetric/QueryMetricOperationsStats.java#L232-L243

`QueryMetricsWriter` has another locale-sensitive decimal formatter:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsWriter.java#L78-L80

It also uses default-locale `%d` formatting for integer values and timestamps:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsWriter.java#L575-L606

For example, a value of `1.5` can become `1,500000` or `1,50` under `Locale.GERMANY`.

Timely expects numeric values in its metric protocol:

https://code.nsa.gov/timely/docs/

## Expected behavior

Values sent to Timely should use a dot as the decimal separator.

Integer values and timestamps should use ASCII digits.

Both rules should hold regardless of the JVM locale.

## Actual behavior

The JVM default locale can control the decimal separator and digits used on the Timely wire.

This can produce a metric that Timely cannot parse correctly.

## Affected components

- `TimelyMetricsReporter`
- `NSQMetricsReporter`, which extends `TimelyMetricsReporter`
- query-metric service statistics sent to Timely over TCP or UDP
- `QueryMetricsWriter` page metrics sent to Timely over UDP

## Suggested fix

Use `Locale.ROOT` or locale-neutral numeric conversion for every value and timestamp sent to Timely.

Use locale-neutral `DecimalFormatSymbols` where `DecimalFormat` is still needed.

Keep localized display or log formatting separate from the Timely wire format.

## Test

Add focused tests for all three Timely formatting paths.

Use `Locale.GERMANY` with a decimal value such as `1.5` and check that the output uses a dot.

Set `Locale.Category.FORMAT` to an Arabic or Persian locale. Report an integer value with a fixed timestamp and check that both contain only ASCII digits.

Restore the original locale after each test.

### LINKED ISSUE #3745: Bug (TcpClient and TimelyMetricsReporter) - failed TCP writes can look successful
## Issue

The Timely TCP writers can treat a failed write or flush as successful.

## Problem

`TcpClient` and `TimelyMetricsReporter` use `PrintWriter` to write and flush metrics:

- https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/warehouse/core/src/main/java/datawave/util/timely/TcpClient.java#L38-L56
- https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/core/metrics-reporter/src/main/java/datawave/metrics/TimelyMetricsReporter.java#L184-L191

`PrintWriter` does not throw I/O exceptions from `write()` or `flush()`. It stores an error flag that callers must check:

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/PrintWriter.html

Neither class checks that flag after sending a batch. They only check it when the next connection attempt starts.

`TcpClient.write()` declares `IOException`, but a failure inside `PrintWriter.write()` is not reported through that exception.

The query-metric service removes metrics from its retry list after `write()` and flushes after the loop:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/microservices/services/query-metric/service/src/main/java/datawave/microservice/querymetric/QueryMetricOperationsStats.java#L170-L202

A failed flush can therefore leave those metrics marked as sent.

## Expected behavior

A local TCP write or flush error should be detected during the same reporting attempt.

The caller should be able to keep or retry metrics that were not sent.

## Actual behavior

`PrintWriter` stores the failure, but the current reporting attempt appears successful.

The failure may only be found before the next batch, after the earlier metrics have been removed from retry handling.

## Affected components

- `TcpClient`
- `TimelyMetricsReporter`
- the query-metric service when Timely uses TCP

## Suggested fix

Check the writer error state when each batch is flushed.

For `TcpClient`, report the failure to the caller and mark the connection as unusable.

For `TimelyMetricsReporter`, log the failure and mark the connection as unusable so the next report reconnects.

The query-metric service should not remove a batch from retry handling until its TCP flush succeeds.

## Test

Use a writer or output stream that fails during flush.

Check that:

- `TcpClient` reports the failure
- query metrics are not treated as successfully sent
- `TimelyMetricsReporter` detects the failure and invalidates the connection

### LINKED ISSUE #3746: Bug (Timely TCP) - reconnect can leave old sockets open
## Issue

The Timely TCP reconnect paths can replace a writer and socket without closing the old ones.

## Problem

Both reconnect methods detect a failed connection and then assign a new socket and writer:

- https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/warehouse/core/src/main/java/datawave/util/timely/TcpClient.java#L82-L103
- https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/core/metrics-reporter/src/main/java/datawave/metrics/TimelyMetricsReporter.java#L194-L217

The existing writer and socket are not closed before their fields are replaced.

The same problem can occur if a new socket is assigned but writer creation then fails. The catch blocks clear the fields without closing that socket.

The normal shutdown methods only close the resources still stored in the fields:

- https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/warehouse/core/src/main/java/datawave/util/timely/TcpClient.java#L65-L79
- https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/core/metrics-reporter/src/main/java/datawave/metrics/TimelyMetricsReporter.java#L220-L235

Repeated connection failures can leave old resources open until they are reclaimed.

## Expected behavior

A writer and socket should be closed before they are replaced.

A partly created connection should also be closed when setup fails.

## Actual behavior

The reconnect paths overwrite or clear the resource fields without first closing the old resources.

## Affected components

- `TcpClient`
- `TimelyMetricsReporter`

## Suggested fix

Add one connection cleanup method to each class.

Use it:

- before replacing a failed connection
- when connection setup fails
- from `close()` or `stop()`

The cleanup should try to close both the writer and socket, then clear both fields.

## Test

Use tracking writer and socket test doubles.

Check that:

- an old writer and socket are closed before reconnect replaces them
- a socket is closed if writer creation fails
- repeated cleanup remains safe

### LINKED ISSUE #3747: Bug (NSQMetricsReporter) - corrupts non-ASCII Timely metrics
## Issue

`NSQMetricsReporter` can change non-ASCII characters before it sends a Timely metric to NSQ.

## Problem

`NSQMetricsReporter.reportMetric(String)` measures the Java character count and writes the metric with `DataOutputStream.writeBytes(String)`:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/core/metrics-reporter/src/main/java/datawave/metrics/NSQMetricsReporter.java#L52-L63

```java
if (dos.size() + timelyMetric.length() > MAX_MESSAGE_SIZE) {
    flush();
    connect();
}

dos.writeBytes(timelyMetric);
```

`writeBytes(String)` discards the high eight bits of each Java character. For example, `é` is written as one byte with value `E9`, instead of its UTF-8 bytes `C3 A9`.

The direct Timely reporter already writes UTF-8:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/core/metrics-reporter/src/main/java/datawave/metrics/TimelyMetricsReporter.java#L194-L204

Timely also decodes TCP metric text as UTF-8:

https://github.com/NationalSecurityAgency/timely/blob/bb3386925c26e6c70bad0739d60cfe505d22aeee/common/src/main/java/timely/netty/tcp/TcpDecoder.java#L23-L35

## Expected behavior

Metrics sent through NSQ should use UTF-8, like metrics sent directly to Timely.

The 7 KiB batch check should use the number of encoded bytes.

## Actual behavior

Non-ASCII metric names or tag values can be changed before NSQ receives them.

The batch check measures Java characters instead of the UTF-8 bytes that should be sent.

## Related work

Issue #3730 and PR #3742 fix UTF-8 handling in `TcpClient`. They do not change `NSQMetricsReporter`.

Issue #3732 and PR #3744 fix empty NSQ batches. They leave `writeBytes()` unchanged, and the proposed tests use ASCII-only data.

## Suggested fix

Encode the metric once as UTF-8, use its byte length for the batch check, and write those same bytes.

For example:

```java
byte[] metricBytes = timelyMetric.getBytes(UTF_8);

if (dos.size() + metricBytes.length > MAX_MESSAGE_SIZE) {
    flush();
    connect();
}

dos.write(metricBytes);
```

## Test

Add focused `NSQMetricsReporter` tests that:

- send a metric containing a non-ASCII name or tag value
- verify that the HTTP request body contains the exact UTF-8 bytes
- use two multibyte metrics whose combined character count is below 7 KiB but whose combined UTF-8 byte count is above 7 KiB
- verify that the metrics are sent in two non-empty batches without data loss

### LINKED ISSUE #3748: Bug (TimelyMetricsReporter) - sends non-numeric gauge values to Timely
## Issue

`TimelyMetricsReporter` can send a non-numeric gauge value even though Timely requires a numeric metric value.

## Problem

`TimelyMetricsReporter.reportGauge(...)` handles numeric gauges as numbers. For every other non-null value, it calls `String.valueOf(value)` and sends the result:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/core/metrics-reporter/src/main/java/datawave/metrics/TimelyMetricsReporter.java#L103-L115

```java
if (value instanceof Float || value instanceof Double) {
    reportMetric(name, "value", ((Number) value).doubleValue(), "GAUGE", null, time);
} else if (value instanceof Number) {
    reportMetric(name, "value", ((Number) value).longValue(), "GAUGE", null, time);
} else {
    reportMetric(name, "value", String.valueOf(value), "GAUGE", null, time);
}
```

The resulting string is inserted into the Timely `put` command without numeric validation:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/core/metrics-reporter/src/main/java/datawave/metrics/TimelyMetricsReporter.java#L172-L181

Timely defines a metric value as a `double`:

https://code.nsa.gov/timely/docs/#put-operation

Its parser reads that field with `Double.parseDouble(...)`:

https://github.com/NationalSecurityAgency/timely/blob/bb3386925c26e6c70bad0739d60cfe505d22aeee/common/src/main/java/timely/model/parse/MetricParser.java#L15-L29

For example, a gauge whose value is `"UP"` produces an invalid command such as:

```text
put service.status 123 UP host=host1 rack=rack1 sample=value sampleType=GAUGE
```

## Expected behavior

`TimelyMetricsReporter` should only send gauge values that Timely can read as numbers.

A non-numeric gauge should be skipped with a clear warning.

## Actual behavior

Any non-null object can be converted to text and sent as the metric value.

Timely cannot parse a value such as `"UP"` or `"true"` as a `double`.

`NSQMetricsReporter` inherits this behavior.

## Related work

Issue #3729 and PR #3743 fix locale-dependent formatting of numeric values. They do not prevent non-numeric gauge values from being sent.

## Suggested fix

Only report gauge values that implement `Number`.

Log and skip other non-null gauge values.

## Test

Add focused tests covering each gauge-value branch:

- `null`: no metric is sent
- integer `Number`: one numeric metric is sent
- floating-point `Number`: one numeric metric is sent
- `String` or `Boolean`: no metric is sent and a warning is logged
- a valid metric after an invalid gauge is still sent

### LINKED ISSUE #3749: Bug (QueryMetricsWriter) - Timely UDP client refresh can race with metric writes
## Issue

A configuration refresh can replace the Timely UDP client while a metric batch is being sent.

## Problem

`onRefresh()` locks the `QueryMetricsWriter` instance before replacing `timelyClient`:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsWriter.java#L154-L158

```java
synchronized (this) {
    timelyClient = createUdpClient();
}
```

`sendMetricsToTimely()` is a synchronized method on the inner `MetricProcessor`, which is a different object:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsWriter.java#L533-L560

The method opens `timelyClient` and reads the shared field again for each write. A refresh between those steps can replace it with a new client that has not been opened.

The old client is not closed during refresh. It is also not closed during shutdown:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsWriter.java#L104-L130

## Expected behavior

One client should handle the full metric batch.

Replaced clients should be closed after active writes finish.

## Actual behavior

A refresh can cause the batch to use two clients.

A write to the new unopened client fails, and the old UDP socket can remain open.

## Suggested fix

Use the same lock for client refresh and metric sending.

Close the replaced client after it is no longer in use. Close the active client during shutdown.

## Test

Pause a test client after `open()`, run a refresh, and then continue the send.

Check that:

- the full batch uses one client
- no write uses an unopened client
- the old client closes after refresh
- the active client closes during shutdown

### LINKED ISSUE #3750: Bug (Timely) - writes whitespace in metric names and tags
## Issue

Timely metric names, tag keys, and tag values are not protected from all whitespace.

## Problem

`QueryMetricOperationsStats` adds query fields directly to Timely lines:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/microservices/services/query-metric/service/src/main/java/datawave/microservice/querymetric/QueryMetricOperationsStats.java#L264-L288

Configured tags are also added without validation:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/microservices/services/query-metric/service/src/main/java/datawave/microservice/querymetric/QueryMetricOperationsStats.java#L304-L310

`TimelyMetricsReporter` replaces literal spaces in metric names, but it does not handle tabs, newlines, or other whitespace:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/core/metrics-reporter/src/main/java/datawave/metrics/TimelyMetricsReporter.java#L103-L149

The name is then added directly to the Timely command:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/core/metrics-reporter/src/main/java/datawave/metrics/TimelyMetricsReporter.java#L172-L181

`QueryMetricsWriter` only replaces the literal space character in tag values. It does not protect tag field names:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsWriter.java#L542-L550

Timely separates fields on whitespace. A space or tab can split one tag into several fields. A newline can split one metric into several commands.

## Expected behavior

Each metric name and tag should remain one valid Timely field.

One input metric should produce exactly one Timely command.

## Actual behavior

Whitespace can split names, keys, values, or full commands.

Timely may reject the metric or read the wrong fields.

## Affected components

- `QueryMetricOperationsStats`
- `TimelyMetricsReporter`
- `NSQMetricsReporter`, which extends `TimelyMetricsReporter`
- `QueryMetricsWriter`

## Suggested fix

Use one shared Timely field encoder for metric names, tag keys, and tag values.

Normalize or reject all unsupported whitespace, not only the literal space character.

Keep the final command newline separate from encoded input.

## Test

Test:

- query tags containing spaces, tabs, and newlines
- a reporter metric name containing a tab or newline
- configured `QueryMetricsWriter` tag fields and values containing whitespace

Check that each input produces exactly one valid Timely command, or is rejected with a clear warning.

### LINKED ISSUE #3751: Bug (QueryMetricOperationsStats) - oversized Timely UDP metric blocks queued metrics
## Issue

One oversized Timely UDP metric can block valid metrics behind it in the retry queue.

## Problem

`UdpClient` sends the complete UTF-8 metric as one datagram without checking its encoded size:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/warehouse/core/src/main/java/datawave/util/timely/UdpClient.java#L27-L32

```java
this.packet.setData(metric.getBytes(UTF_8));
this.sock.send(packet);
```

`writeQueryStatsToTimely()` catches errors outside the send loop. It requeues the failed metric and every later metric:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/microservices/services/query-metric/service/src/main/java/datawave/microservice/querymetric/QueryMetricOperationsStats.java#L170-L197

An oversized datagram fails every time. It stays first in the queue, so later valid metrics are not sent.

If the queue grows past 10,000 entries, the method clears the full queue.

## Expected behavior

One invalid oversized metric should not block valid metrics that follow it.

## Actual behavior

The same oversized metric fails on each retry and keeps all later metrics queued.

## Suggested fix

Check the UTF-8 byte length before sending.

Handle a permanent size error for that metric only, log it clearly, and continue with later metrics. Keep retries for temporary transport errors.

## Test

Use a test UDP client that rejects the first metric as too large and accepts the second metric.

Check that:

- the valid second metric is sent
- the oversized metric does not remain ahead of the valid metric
- unrelated queued metrics are not cleared

### LINKED ISSUE #3752: Bug (Timely TCP) - connection attempts have no timeout
## Issue

The Timely TCP clients do not set a time limit for connection attempts.

## Problem

`TcpClient` opens its socket with `new Socket(host, port)`:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/warehouse/core/src/main/java/datawave/util/timely/TcpClient.java#L82-L103

`TimelyMetricsReporter` uses the same socket constructor:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/core/metrics-reporter/src/main/java/datawave/metrics/TimelyMetricsReporter.java#L194-L217

This leaves the connection time limit to the operating system. The application cannot apply a shorter limit.

A blackholed endpoint can block the reporting thread until the operating system ends the attempt. This can also delay the final query-metric write during shutdown:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/microservices/services/query-metric/service/src/main/java/datawave/microservice/querymetric/QueryMetricOperations.java#L161-L177

## Expected behavior

Each Timely TCP connection attempt should stop after a configured time.

The normal retry backoff should start after the timeout.

## Actual behavior

A connection attempt can block until the operating system's TCP timeout ends.

## Affected components

- `TcpClient`
- `TimelyMetricsReporter`
- `NSQMetricsReporter`, which extends `TimelyMetricsReporter`
- query-metric service TCP output

## Suggested fix

Create an unconnected socket and call `connect(...)` with a positive, configurable timeout.

Close the new socket if setup fails.

## Test

Use an injected socket connector that reports a `SocketTimeoutException`.

Check both TCP implementations:

- the configured timeout is supplied
- the failed socket is closed
- normal reconnect backoff begins

### LINKED ISSUE #3754: Bug (QueryMetricOperationsStats) - does not retry failed Timely UDP initialization
## Issue

`QueryMetricOperationsStats` does not retry Timely UDP setup after an initial failure.

## Problem

The constructor creates and opens the UDP client once:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/microservices/services/query-metric/service/src/main/java/datawave/microservice/querymetric/QueryMetricOperationsStats.java#L98-L108

If that call fails, later publishing methods call `write(...)` without trying `open()` again:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/microservices/services/query-metric/service/src/main/java/datawave/microservice/querymetric/QueryMetricOperationsStats.java#L139-L202

`UdpClient.write(...)` cannot send while its socket is null.

## Expected behavior

A temporary UDP setup failure should not disable Timely metrics until the service restarts.

A later publishing cycle should retry setup.

## Actual behavior

The first UDP setup failure leaves the client unusable for later publishing cycles.

## Affected components

- `QueryMetricOperationsStats`
- query-metric service Timely UDP output

## Suggested fix

Retry `open()` before a later UDP batch when the client is not open.

Keep the retry safe when setup only completed in part.

## Test

Use an injected UDP client whose first `open()` call fails and whose second call succeeds.

Check that a later publishing cycle:

- retries setup
- sends service metrics
- sends queued query metrics

### LINKED ISSUE #3755: Bug (UdpClient) - does not recover when Timely hostname is initially unresolved
## Issue

`UdpClient` does not recover if the Timely hostname cannot be resolved when the client is created.

## Problem

The constructor resolves the hostname once and stores the result in a reusable packet:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/warehouse/core/src/main/java/datawave/util/timely/UdpClient.java#L16-L18

The same packet is used for every send:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/warehouse/core/src/main/java/datawave/util/timely/UdpClient.java#L27-L32

If the first lookup is unresolved, the packet has no destination address. Later writes do not resolve the hostname again, even if DNS has recovered.

## Expected behavior

A failed DNS lookup should report a clear error.

A later connection or write attempt should be able to resolve the hostname again.

## Actual behavior

The client keeps the unresolved destination for its full lifetime.

Later writes continue to fail until a new client is created.

## Affected components

- `UdpClient`
- query-metric service Timely UDP output
- `QueryMetricsWriter` Timely UDP output

## Suggested fix

Keep the hostname and port as client state.

Resolve the hostname during setup or retry, and return a clear `IOException` when resolution fails.

## Test

Use an injected resolver that fails once and then returns a loopback address.

Check that:

- the first attempt reports a clear error
- the second attempt resolves the hostname again
- a loopback receiver gets the exact UTF-8 metric

### LINKED ISSUE #3756: Bug (Timely) - numeric formatting loses metric precision
## Issue

Timely output can round small decimal values to zero or remove useful precision.

## Problem

`TimelyMetricsReporter` formats every `double` with `%f`, which keeps six digits after the decimal point:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/core/metrics-reporter/src/main/java/datawave/metrics/TimelyMetricsReporter.java#L164-L166

For example, `0.0000004` becomes `0.000000`.

`QueryMetricOperationsStats` uses two decimal places for decimal metric values:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/microservices/services/query-metric/service/src/main/java/datawave/microservice/querymetric/QueryMetricOperationsStats.java#L232-L243

`QueryMetricsWriter` also uses two decimal places for `calltimeperrecord`:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/web-services/query/src/main/java/datawave/webservice/query/metric/QueryMetricsWriter.java#L574-L578

For example, `0.004` becomes `0.00`.

## Expected behavior

Timely wire values should keep enough digits to preserve the metric value.

The format should also stay independent of the JVM locale.

## Actual behavior

Small nonzero values can become zero.

Other values can be rounded before Timely receives them.

## Affected components

- `TimelyMetricsReporter`
- `NSQMetricsReporter`, which extends `TimelyMetricsReporter`
- `QueryMetricOperationsStats`
- `QueryMetricsWriter`

## Suggested fix

Use a locale-neutral, round-trip-safe wire format such as `Double.toString(...)`.

Keep display-only formatting separate from the Timely wire format.

## Test

Report values such as:

- `0.0000004`
- `1.23456789`
- `0.004`

Check that:

- nonzero values remain nonzero
- parsing each generated value returns the expected value
- the output is the same under a locale that uses a comma decimal separator

### LINKED ISSUE #3757: Bug (QueryMetricOperationsStats) - does not close Timely clients
## Issue

`QueryMetricOperationsStats` does not close the Timely clients that it creates.

## Problem

The class stores and opens its own TCP or UDP clients:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/microservices/services/query-metric/service/src/main/java/datawave/microservice/querymetric/QueryMetricOperationsStats.java#L57-L108

It has no destroy or close method.

The application shutdown method sends one final metric batch, but it does not close these clients:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/microservices/services/query-metric/service/src/main/java/datawave/microservice/querymetric/QueryMetricOperations.java#L161-L177

## Expected behavior

Owned Timely clients should close when the Spring context stops.

The final metric batch should be sent before they close.

## Actual behavior

The owned clients are left open until garbage collection or JVM exit.

This can retain sockets during a context restart or redeployment.

## Affected components

- `QueryMetricOperationsStats`
- query-metric service Timely TCP and UDP output

## Suggested fix

Add an idempotent shutdown method, such as `@PreDestroy`, that closes each created client.

Make cleanup safe when setup completed only in part.

## Test

Close a Spring test context with tracking TCP and UDP clients.

Check that:

- the final metric batch is sent before close
- each opened client closes once
- repeated or partial cleanup is safe

### LINKED ISSUE #3758: Bug (TimelyProperties) - enabled endpoint is not validated
## Issue

`TimelyProperties` does not validate the endpoint when Timely output is enabled.

## Problem

The configuration allows a null host and does not validate the port:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/microservices/services/query-metric/service/src/main/java/datawave/microservice/querymetric/config/TimelyProperties.java#L8-L15

With Timely enabled, an incomplete endpoint can reach client setup.

TCP and UDP then handle the missing host in different ways. This can connect to an unintended local endpoint or leave Timely output unusable.

## Expected behavior

When Timely output is enabled:

- the host should not be blank
- the port should be from `1` to `65535`
- the protocol should be set

When Timely output is disabled, the endpoint may be omitted.

## Actual behavior

Invalid enabled settings can pass configuration binding and fail later during client setup or use.

## Affected components

- query-metric service Timely TCP output
- query-metric service Timely UDP output

## Suggested fix

Add conditional configuration validation for the enabled Timely endpoint.

Fail application startup with a clear binding message when enabled settings are invalid.

## Test

Use `ApplicationContextRunner` to check:

- enabled with a missing or blank host fails
- enabled with an invalid port fails
- enabled with no protocol fails
- disabled with no endpoint succeeds
- valid TCP and UDP settings succeed

arthurianresolve:
## Summary

Harden Timely metric output across core, metric reporters, the query-metric service, and `QueryMetricsWriter`.

- use locale-neutral, full-precision wire values
- normalize whitespace in metric names and tags
- validate enabled Timely endpoint settings
- close stale TCP sockets, limit connection time, and report write or flush errors
- preserve UTF-8 NSQ bytes and skip non-numeric gauges
- retry UDP setup and DNS resolution
- skip oversized UDP metrics without blocking later metrics
- serialize client refresh, writes, and shutdown
- close service-owned Timely clients after the final Spring shutdown write
- make the warehouse RFile and MiniAccumulo test setup safe on Windows

Each reported issue is kept in a separate commit. The Windows test portability change is also isolated in its own commit.

## Compatibility

- The standalone query-metric service remains on the latest immutable `datawave-core:7.42.0`.
- Package-private service TCP and UDP transports provide checked batch-flush results, bounded TCP connection attempts, and retryable UDP hostname resolution without requiring a mutable core dependency.
- TCP connection attempts use `datawave.query.metric.timely.connect-timeout-millis`. It defaults to 10,000 milliseconds and must be positive when TCP output is enabled.
- The service-local transport boundary is compatible with 7.42 and 7.43 because it uses JDK networking APIs rather than version-specific core client APIs. It should be removed after an immutable core release containing the same fixes becomes the minimum supported version.
- `7.43.0-SNAPSHOT` was used only as a test override. It is not committed as a service dependency.
- Timely configurations with `enabled=true` must set `protocol` to `TCP` or `UDP`.
- `QueryMetricsWriter` waits for an active Timely batch before closing its client. A blocked network call can therefore extend shutdown past `maxShutdownMs`.
- `BulkIngestMapFileLoaderTest` runs normally on Linux and on Windows with `winutils.exe`. It is explicitly skipped on Windows when that required Hadoop native helper is unavailable.

## Review status

This PR remains a draft. Before it is ready:

- merge #3742, then rebase this branch and preserve its UTF-8 writer, capped backoff, and corrected log message together with this PR's TCP lifecycle changes
- merge #3744, then preserve its empty-NSQ-batch and bounded-HTTP-wait behavior together with this PR's NSQ UTF-8 changes
- review the temporary package-private query-metric transport boundary and its removal plan
- separate the unrelated Windows test-portability changes or obtain explicit agreement to review them in this PR
- obtain Linux CI and a complete unit-reactor result or an upstream waiver for the unrelated Yake baseline failures

## Test

Windows host: Java 11.0.31, Maven 3.9.16, and Hadoop 3.4.1 with `winutils.exe` and `hadoop.dll` available.

Passed:

- `metrics-reporter`: 17 tests, 0 failures, 0 errors, 0 skipped.
- Query-metric service against immutable DataWave `7.42.0`: 70 tests, 0 failures, 0 errors, 2 expected skips.
- DataWave `7.43.0-SNAPSHOT` compatibility reactor: all 38 required modules compiled and installed successfully; the POM was not changed to use the snapshot.
- Query-metric service against those 7.43 artifacts: 70 tests, 0 failures, 0 errors, 2 expected skips.
- `AnnotationDataAccessTest`: 17 tests, 0 failures, 0 errors, 0 skipped.
- Warehouse core RFile/input-format and transport tests: 25 tests, 0 failures, 0 errors, 0 skipped.
- `QueryMetricsWriterTest`: 4 tests, 0 failures, 0 errors, 0 skipped.
- `ShardReindexMapperTest`: 66 tests, 0 failures, 0 errors, 0 skipped.
- `ShardReindexVerificationMapperTest`: 16 tests, 0 failures, 0 errors, 0 skipped.
- `BulkIngestMapFileLoaderTest` integration: 54 tests, 0 failures, 0 errors, 0 skipped.
- Full `warehouse/ingest-core` unit module: 852 tests, 0 failures, 0 errors, 1 existing skip.
- Full integration reactor (`verify -DskipUTs=true -DskipITs=false`): 97 tests, 0 failures, 0 errors, 0 skipped; BUILD SUCCESS.

Root unit reactor status:

- Timely and ingest-core modules passed.
- The reactor then stopped in `warehouse/keyword-common` because all four existing `YakeKeywordRegressionTest` golden-output cases differ from their checked-in baselines. This unrelated test data was not changed in this PR.

## Issues

Fixes #3729
Fixes #3745
Fixes #3746
Fixes #3747
Fixes #3748
Fixes #3749
Fixes #3750
Fixes #3751
Fixes #3752
Fixes #3754
Fixes #3755
Fixes #3756
Fixes #3757
Fixes #3758


arthurianresolve:

PR #3744, opened 2026-07-23
Prevent empty NSQ batches and bound HTTP waits

### LINKED ISSUE #3732: Bug (NSQMetricsReporter) - sends empty batches to NSQ
## Issue

`NSQMetricsReporter` can send empty HTTP requests to the NSQ `/mpub` endpoint.

## Problem

The inherited reporting method connects and flushes even when there are no metrics to report:

https://github.com/NationalSecurityAgency/datawave/blob/436fc484454e2d9db86bd11c01a01c470deae56f/core/metrics-reporter/src/main/java/datawave/metrics/TimelyMetricsReporter.java#L64-L96

`NSQMetricsReporter.connect()` creates an empty buffer. Its `flush()` method then sends the buffer without checking whether it contains data:

https://github.com/NationalSecurityAgency/datawave/blob/436fc484454e2d9db86bd11c01a01c470deae56f/core/metrics-reporter/src/main/java/datawave/metrics/NSQMetricsReporter.java#L43-L82

```java
post.setEntity(new ByteArrayEntity(baos.toByteArray(), ContentType.DEFAULT_BINARY));
HttpResponse response = client.execute(post);
```

An empty request can also occur when the first metric is larger than the 7 KiB target batch size:

https://github.com/NationalSecurityAgency/datawave/blob/436fc484454e2d9db86bd11c01a01c470deae56f/core/metrics-reporter/src/main/java/datawave/metrics/NSQMetricsReporter.java#L52-L63

```java
if (dos.size() + timelyMetric.length() > MAX_MESSAGE_SIZE) {
    flush();
    connect();
}
```

NSQ accepts newline-separated messages through `/mpub`:

https://nsq.io/components/nsqd.html#post-mpub

## Expected behavior

`NSQMetricsReporter` should only send a request when the batch contains at least one metric.

A metric larger than the target batch size should be sent once as a non-empty batch.

## Actual behavior

The reporter can send an empty `/mpub` request.

The current NSQ server treats an empty batch as a no-op, so this does not lose the following metric. It does create an unnecessary HTTP request.

## Suggested fix

Return from `flush()` when the buffer is empty.

For example:

```java
if (baos.size() == 0) {
    return;
}
```

The 7 KiB value should be treated as a target batch size because one metric can be larger than that value.

## Test

The `metrics-reporter` module does not contain focused `NSQMetricsReporter` tests.

Add tests that verify:

- an empty reporting cycle makes no HTTP request
- one metric larger than 7 KiB makes one non-empty request
- two metrics whose combined size exceeds 7 KiB make two non-empty requests
- no metric is lost during a batch rollover

### LINKED ISSUE #3753: Bug (NSQMetricsReporter) - HTTP requests have no timeout
## Issue

`NSQMetricsReporter` does not set a time limit for HTTP requests to NSQ.

## Problem

The reporter creates a default HTTP client without a request configuration:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/core/metrics-reporter/src/main/java/datawave/metrics/NSQMetricsReporter.java#L27-L32

`flush()` then waits for the HTTP request to finish:

https://github.com/NationalSecurityAgency/datawave/blob/c3b731dff9aab5fa59b11bc34df0b96a9b24706a/core/metrics-reporter/src/main/java/datawave/metrics/NSQMetricsReporter.java#L67-L82

No connection-request, connection, or response timeout is set.

## Expected behavior

An NSQ request should stop after a configured time if the server does not connect or respond.

Later metric reports should still be able to run.

## Actual behavior

A slow or unresponsive NSQ server can hold the reporter thread based on the HTTP client's network defaults.

This can delay later metric reports.

## Affected components

- `NSQMetricsReporter`
- services configured with `NSQMetricsReporterFactory`

## Suggested fix

Set clear connection-request, connection, and response timeouts.

Make the values configurable if deployments need different limits.

## Test

Use a local HTTP server that accepts a request but does not send a response.

Check that:

- the request ends after the configured timeout
- a later report can still run

arthurianresolve:
## Summary

Improve `NSQMetricsReporter` reliability and make its HTTP timeout configurable through each supported setup path:

- do not post empty metric batches
- set connection-pool, connection, and response-inactivity timeouts
- allow direct Java callers to override the timeout through `NSQMetricsReporterFactory.Builder`
- bind `metrics.reporter.timeout` for Spring services
- bind `dw.metrics.reporter.timeout.millis` for CDI services
- fail clearly when an explicit timeout is used with a reporter that does not support it

## Root cause

`flush()` sent a request even when its buffer was empty. The Apache HTTP client also had no connection-request, connection, or socket timeout, so a slow NSQ endpoint could delay later reporting cycles. Service configuration paths could not pass a timeout to the NSQ builder.

## Impact

NSQ no longer receives empty `/mpub` requests. HTTP waits use a 10-second default limit, and a timed-out request does not prevent a later batch from being attempted. Java, Spring, and CDI deployments can override the shared limit.

## Configuration

The default timeout is 10 seconds for the connection-pool wait, connection attempt, and response wait.

- Direct Java: call `withTimeout(...)` on the NSQ builder.
- Spring: set `metrics.reporter.timeout` with Spring duration syntax, such as `10s`, `500ms`, or a bare number of seconds.
- CDI: set `dw.metrics.reporter.timeout.millis` to a positive millisecond value. The default `-1` leaves the reporter's own default unchanged.

When the property is omitted, existing behavior uses the reporter's 10-second default. Setting it for an unsupported reporter fails with a clear configuration error. The original five-argument CDI producer overload remains available for source and binary compatibility.

## Release sequencing

The new builder capability is in `metrics-reporter:3.0.1-SNAPSHOT`, and the root and starter reactors point to that version so this draft can build the producer and consumers together.

Before this PR is ready to merge, publish immutable `metrics-reporter:3.0.1` and replace both snapshot version pins with `3.0.1`. Clean reactors that do not include the opt-in `metrics-reporter` module must not depend on the snapshot.

## Integration

This PR overlaps #3743 in `NSQMetricsReporter`. Whichever PR merges second should rebase and retain both #3743's UTF-8 byte sizing/writes and this PR's empty-batch and timeout changes, then rerun the `metrics-reporter` tests.

## Test

Windows host: Java 11.0.31 and Maven 3.9.16.

Passed:

- Full `metrics-reporter` module: 5 tests, 0 failures, 0 errors, 0 skipped.
- Full `spring-boot-starter-datawave` module: 43 tests, 0 failures, 0 errors, 0 skipped.
- Full `datawave-ws-common` module: 68 tests, 0 failures, 0 errors, 0 skipped.
- CDI compatibility reactor: all 22 required modules built successfully; `MetricsCdiConfigurationTest` passed.

Coverage includes:

- empty direct flushes
- oversized and rollover batches
- all three Apache HTTP timeout values
- a local server that responds after the configured limit
- a later successful batch after a timeout
- Spring binding for omitted, millisecond-suffixed, and bare-second values
- capable, unsupported, omitted, and invalid timeout configuration branches
- CDI millisecond configuration and preservation of the existing producer overload

Fixes #3732
Fixes #3753


arthurianresolve:

PR #3759, opened 2026-07-24
Add feature-flagged Kryo serialization for TypeMetadata

apmoriarty:
## Summary
- `toString()`/`fromString()` dominated profiler traces (`HashMap.put`/`calculateHash`) from rebuilding index-to-name lookup maps per field/value pair; fixes that hot path.
- Adds native Kryo support: `TypeMetadata` implements `KryoSerializable`, and `TypeMetadataSerializer` wraps it into a Base64 string round trip, reusing its `Kryo`/buffer per instance (matching the `CandidateResultSerializer` convention) since it's called many times per thread.
- The Kryo form uses the same mini-map convention as `toString()`: an ingest type dictionary, a normalizer type dictionary, then one entry per field name holding its `ingestTypeIndex:normalizerTypeIndex` pairs as variable length ints. Each name is written exactly once, and `read` holds both dictionaries as arrays so resolving a pair is an array index rather than a map rebuild, keeping deserialization O(1) per entry.
- Wires it through a new `kryoTypeMetadata` flag on `ShardQueryConfiguration`: `DefaultQueryPlanner` serializes via Kryo when enabled, marking the iterator option so `QueryOptions` decodes it correctly. The flag is skipped during per-shard type metadata reduction, since `VisitorFunction` re-serializes via `toString()` afterward and doesn't understand the Kryo format.
- A parameterized performance test (1/5/20 datatypes x 25/250/2500 fields) shows the Kryo round trip is consistently 1.7-3.0x faster.

## Payload size

The mini-map convention is what keeps the payload smaller than the string form it replaces. An earlier revision of this branch wrote the full ingest type and normalizer class name for every field/normalizer pair, which made it **larger** than `toString()` — measured against the same fixture, with fully qualified normalizer names:

| datatypes x fields | `toString()` | Kryo, repeated names | Kryo, mini-map |
| --- | --- | --- | --- |
| 1 x 25 | 526 | 982 (1.87x) | 401 (0.76x) |
| 1 x 2500 | 41,551 | 101,402 (2.44x) | 31,527 (0.76x) |
| 5 x 250 | 8,099 | 49,526 (6.12x) | 5,063 (0.63x) |
| 5 x 2500 | 81,599 | 507,006 (6.21x) | 51,563 (0.63x) |
| 20 x 250 | 25,799 | 198,111 (7.68x) | 12,708 (0.49x) |
| 20 x 2500 | 256,799 | 2,028,031 (7.90x) | 126,708 (0.49x) |

The Kryo form is now 0.49-0.76x of the string form, i.e. 2.5-16x smaller than writing the names repeatedly.

Two caveats worth noting for review:

- Building the dictionaries and inverting to field-major order is write-side work the repeated-name layout skipped, so the round trip speedup drops from 3.1-5.5x to 1.7-3.0x. Both figures are measured on the same fixture, so this is a real trade: roughly half the speed advantage in exchange for a payload that is 4-16x smaller and no longer larger than what it replaces.
- `TypeMetadataSerializer` Base64-encodes on the way out, which adds a third back. The transported string is therefore at rough parity with `toString()` for a single ingest type (1.00x) and only pulls ahead from two ingest types on (0.66-0.94x). The raw Kryo form is smaller in every combination; Base64 is what consumes the margin at the low end. Both properties are asserted separately rather than averaged together.

## Test plan
- [x] `metadata-utils` module test suite (379 tests) passes, via `install` so the enforcer rules run
- [x] Targeted `query-core` suites (`QueryOptionsTest`, `DefaultQueryPlannerTest`, `DocumentIteratorOptionsTest`, `VisitorFunctionTest`, `TypeMetadataTest`) pass
- [x] `ColorsTest` exercises the Kryo path end to end through the planner and iterator
- [x] `TypeMetadataSerializerTest` covers round trip correctness, instance reuse across differently-sized inputs, and the two payload size properties above
- [x] `TypeMetadataSerializationPerformanceTest` reports throughput and payload size across the matrix

### DIFF
--- a/core/utils/metadata-utils/src/main/java/datawave/query/util/TypeMetadata.java
+++ b/core/utils/metadata-utils/src/main/java/datawave/query/util/TypeMetadata.java
@@ -310,6 +310,11 @@ public class TypeMetadata implements Serializable, KryoSerializable {
         return typeMap;
     }
 
+    /**
+     * Renders this instance in the form described by {@link #fromString(String)}, populating the ingest type and normalizer type mini-maps as a side effect.
+     *
+     * @return a serialized TypeMetadata
+     */
     public String toString() {
         StringBuilder sb = new StringBuilder();
 
@@ -374,17 +379,30 @@ public class TypeMetadata implements Serializable, KryoSerializable {
         return sb.toString();
     }
 
+    /**
+     * Populates this instance from the semicolon delimited form produced by {@link #toString()}: a {@code dts} mini-map assigning an index to each ingest type,
+     * a {@code types} mini-map assigning an index to each normalizer type held against a field, then one entry per field name whose bracketed list holds
+     * {@code ingestTypeIndex:normalizerTypeIndex} pairs.
+     *
+     * <pre>
+     * dts:[0:ingestA,1:ingestB];types:[0:LcNoDiacriticsType,1:NumberType];FIELD1:[0:0,1:0];FIELD2:[0:0,0:1,1:1]
+     * </pre>
+     *
+     * The example above gives FIELD1 an LcNoDiacriticsType in both ingest types, and FIELD2 both types in ingestA and a NumberType in ingestB. Parsing is
+     * lenient: input with fewer than three entries is discarded, empty values are skipped (a field absent from a trailing ingest type emits a trailing comma),
+     * and an index with no mini-map entry yields an empty type name.
+     *
+     * @param data
+     *            a serialized TypeMetadata
+     */
     private void fromString(String data) {
         String[] entries = parse(data, ';');
 
         if (entries.length > 2) {
-            // invert each index -> name mini-map once, as its prefix entry is parsed, so every field/value
-            // entry below is an O(1) lookup instead of rebuilding and linearly scanning an ImmutableMap per
-            // value (see git history for the prior implementation, which dominated profiler traces on large
-            // TypeMetadata instances). These start empty rather than inverting the current mini-maps: those
-            // are unpopulated for a freshly parsed instance, and are still null when called from readObject,
-            // which does not run a constructor. A value entry preceding its prefix entry resolves to "" here
-            // just as it did before.
+            // invert each index -> name mini-map once, as its prefix entry is parsed, so that every field entry below
+            // is an O(1) lookup instead of a linear scan per value. These start empty rather than inverting the
+            // current mini-maps: those are unpopulated for a freshly parsed instance, and are still null when called
+            // from readObject, which does not run a constructor.
             Map<Integer,String> ingestTypesByIndex = Collections.emptyMap();
             Map<Integer,String> dataTypesByIndex = Collections.emptyMap();
 
@@ -396,6 +414,8 @@ public class TypeMetadata implements Serializable, KryoSerializable {
                     setDataTypesMiniMap(parseTypes(entry));
                     dataTypesByIndex = invert(getDataTypesMiniMap());
                 } else {
+                    // a field entry, FIELD_NAME:[ingestTypeIndex:normalizerTypeIndex,...], splitting into the field
+                    // name and the bracketed list of index pairs
                     String[] entrySplits = parse(entry, ':');
 
                     // get rid of the leading and trailing brackets:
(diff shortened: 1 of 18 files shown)


core/utils/metadata-utils/src/main/java/datawave/query/util/TypeMetadata.java
FineAndDandy, line 399: would be nice to take the opportunity of being in this code to update the javadoc to reflect the expected form of data

core/utils/metadata-utils/src/main/java/datawave/query/util/TypeMetadata.java
```diff
@@ -389,27 +404,10 @@ private void fromString(String data) {
...
-                                .entrySet()
-                                .stream()
-                                .filter(e -> e.getValue().equals(Integer.valueOf(vs[1])))
-                                .map(Entry::getKey)
-                                .findFirst().orElse("");
-                        // @formatter:on
+                            String[] vs = Iterables.toArray(Splitter.on(':').omitEmptyStrings().trimResults().split(aValue), String.class);
+
+                            String ingestType = ingestTypesByIndex.getOrDefault(Integer.valueOf(vs[0]), "");
+                            String dataType = dataTypesByIndex.getOrDefault(Integer.valueOf(vs[1]), "");
```
FineAndDandy, line 434: to make this better than what was here before, document what you expect to be parsing

warehouse/query-core/src/main/java/datawave/query/iterator/QueryOptions.java
```diff
@@ -301,6 +303,9 @@ public class QueryOptions implements OptionDescriber {
     protected long sourceLimit = -1;
     protected boolean disableIndexOnlyDocuments = false;
     protected TypeMetadata typeMetadata = new TypeMetadata();
+    // reused across calls to validateTypeMetadata; not thread-safe, so deliberately not copied by deepCopy() -
+    // each QueryOptions (and each deep copy of it) gets its own instance via this field initializer
+    private final TypeMetadataSerializer typeMetadataSerializer = new TypeMetadataSerializer();
```
FineAndDandy: add transient

warehouse/query-core/src/main/java/datawave/next/retrieval/DocumentIteratorOptions.java
```diff
@@ -75,6 +77,8 @@ public class DocumentIteratorOptions implements OptionDescriber {
     // variables set from options
     protected String query;
     protected boolean compressedOptions = false;
+    // not thread-safe; consistent with the rest of this class's per-instance state
+    private final TypeMetadataSerializer typeMetadataSerializer = new TypeMetadataSerializer();
```
FineAndDandy: add transient

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -316,6 +317,8 @@ public class DefaultQueryPlanner extends QueryPlanner implements Cloneable {
 
     protected CompositeMetadata compositeMetadata;
     protected TypeMetadata typeMetadata;
+    // reused across calls to configureTypeMappings; not thread-safe, so each DefaultQueryPlanner (and clone) gets its own instance
+    private final TypeMetadataSerializer typeMetadataSerializer = new TypeMetadataSerializer();
```
FineAndDandy: add transient

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
```diff
@@ -91,6 +92,8 @@ public class VisitorFunction implements Function<ScannerChunk,ScannerChunk> {
     private Cache<String,String> queryCache;
 
     private TypeMetadata cachedTypeMetadata = null;
+    // not thread-safe; consistent with cachedTypeMetadata above, this assumes a VisitorFunction instance is confined to a single thread
+    private final TypeMetadataSerializer typeMetadataSerializer = new TypeMetadataSerializer();
```
FineAndDandy: transient

FineAndDandy (changes requested):
can you add an AbstractQueryTest that exercises the new serialization?

warehouse/query-core/src/main/java/datawave/query/tables/async/event/VisitorFunction.java
apmoriarty: Not strictly required because non of the classes or their ancestors implement Serializable, but I'm find with adding the hint.

apmoriarty:
https://github.com/NationalSecurityAgency/datawave/pull/3759/changes/7700bab8172a116111c6c9c2a21a920c8932b2ba is the result of a contributor catching that the new kryo read/write methods were not using the minimap algorithm.

FineAndDandy (approved):

hlgp (approved):

PR #3764, opened 2026-07-24
Sort segments in hash order prior to calculating annotation hash

tharke13:
Sort the Annotation segment list prior to calculating the annotation hash in order to preserve a consistent Annotation object equals/hashCode throughout the Annotation lifecycle.

Without this change, due to the way Accumulo stores annotations in lexicographic order, the segment list ordering can change between creating a new Annotation, writing to Accumulo, then reading it back. The order of the segments affects the default Annotation.equals method result.


FineAndDandy (dismissed):

sylvesterkaczmarek (commented):
This changes generated annotation IDs for existing multi-segment content, the tests replace many IDs, and Accumulo addresses stored annotations by that ID. I don’t see a migration or alias path for rows written before the upgrade. Re-adding the same logical annotation can therefore create a new ID while stored references keep the old one. How are existing annotation IDs meant to be reconciled or migrated?

core/annotation/src/main/java/datawave/annotation/util/v1/AnnotationUtils.java
```diff
@@ -115,15 +119,42 @@ public static AnnotationSource injectAnnotationSourceHashes(AnnotationSource ann
...
+     * annotation ({@link #injectAnnotationHash(Annotation)}), so segment order is always consistent regardless of caller.
+     *
+     * @param segments
+     *            the segments to sort.
+     * @return a new list containing the segments sorted in ascending segment hash order.
+     */
+    public static List<Segment> sortSegments(List<Segment> segments) {
+        // @formatter:off
+        return segments.stream()
+                .map(s -> s.getSegmentHash().isEmpty() ? injectSegmentHash(s) : s)
```
tharke13, line 160: should we always recompute the hash? In case someone were to add a segment using:
`existingSegmentWithHash.toBuilder().makeChanges().build()`

core/annotation/src/main/java/datawave/annotation/util/v1/AnnotationUtils.java
```diff
@@ -115,15 +119,42 @@ public static AnnotationSource injectAnnotationSourceHashes(AnnotationSource ann
...
      *
      * @param annotation
      *            the annotation to inject.
-     * @return the annotation with boundary type injected.
+     * @return the annotation with its segments sorted in segment hash order and its annotation id injected.
      */
     public static Annotation injectAnnotationHash(Annotation annotation) {
-        final String hash = calculateAnnotationHash(annotation);
-        return annotation.toBuilder().setAnnotationId(hash).build();
+        Annotation sortedAnnotation = annotation.toBuilder().clearSegments().addAllSegments(sortSegments(annotation.getSegmentsList())).build();
```
tharke13: This seems to sort the segments twice. It's probably fine though since the 2nd one should be O(n).

core/annotation/src/main/java/datawave/annotation/util/v1/AnnotationUtils.java
drewfarris: Yes, I wrestled with this one, and you are correct. I think we always want to recompute the hash, even if it means we're potentially recomputing the same hash

tharke13:
@drewfarris Your latest additions to the PR look good to me

core/annotation/src/main/java/datawave/annotation/util/v1/AnnotationUtils.java
drewfarris: After further investigation, if the hash is populated and non-empty, there's a solid chance we received it from upstream. In that case there are reasons we should leave it along.

drewfarris (approved):

apmoriarty (approved):

PR #3360, opened 2026-01-07
Bugfix/sonarq broke us

ivakegg:

PR #3793, opened 2026-07-30
Remove JVM-wide lock from DatawaveArithmetic.matchesFst

apmoriarty:
A fully built Lucene FST is immutable and getBytesReader() allocates a fresh reader per call, so the shared monitor added for SonarQube S2445 bought no safety while serializing every tserver scan thread on the FST list path.


apmoriarty:

avgAGB (approved):

FineAndDandy (approved):

PR #3794, opened 2026-07-30
Confine date filter formats to each thread instead of locking

apmoriarty:
Every date filter evaluation in the JVM serialized on a single static lock, so concurrent tserver scans contended on it. Giving each thread its own SimpleDateFormat instances removes the sharing that the lock was guarding.


apmoriarty:

dtspence (approved):

avgAGB (approved):

PR #3795, opened 2026-07-30
Fix query span locking in LogTiming and QuerySpanCollector

apmoriarty:
Drop the JVM-wide mutex in LogTiming, whose callers all pass a thread-confined span, and return a defensive copy from QuerySpanCollector.getStageTimers so callers cannot observe the backing map being mutated or cleared.


apmoriarty:

avgAGB (approved):

billoley (approved):

PR #3803, opened 2026-08-04
Stop InMemoryScanner filtering user iterator output by range

apmoriarty:
A tablet server returns whatever the topmost user iterator emits; nothing between the stack and the client tests keys against the range. Filtering here hid iterators that emit outside their seek range, and disagreed with InMemoryBatchScanner, which never filtered.


apmoriarty:
This must be integrated behind a tag to take effect

alerman (approved):

jschmidt10 (approved):

Issue #3656, opened 2026-07-07
Protect against queries that pull problematically large documents


ivakegg:
We have a case where extremely large documents can cause tservers to run out of memory when being aggregated for queries.  We need to protect against this.  The KeyToDocumentData class is the one that aggregates the document.  We could create an EventDataQueryFilter which is aware of a size threshold.

Option 1: simply stop the aggregation of a document once it gets to a specified threshold.  If the document had enough to evaluate then it will get passed back as is.  Otherwise it will silently get dropped.  I would think the aggregation should add in a special field to denote that the document is incomplete in this case.

Option 2 is something a little smarter where at some lower threshold it will stop aggregating non-query fields only.  If we hit some higher threshold then revert to Option 1.  Again we need a field to denote that the document is incomplete.

Option 3 is something even smarter.  At some lower threshold it will stop aggregating non-query fields only for those that already have some values.  If we hit some higher threshold then we actively prune all non-query fields down to a limit (maybe 100).  at a slightly higher threshold we actively prune all non-query fields down to a smaller limit (maybe 10).  That at the top threshold we revert to option 2 or 1.

See LimitFields which may already have a mechanism to denote how many of what fields we collapsed.

apmoriarty:
Given the problems we have had with special fields or special attributes in the Document class (see, every unit test we've ever written that has to ignore "special" attributes), perhaps we need to formalize the concept of "reserved fields". 

We could also add an 'incomplete' flag to the Document class proper that denotes when an aggregation failed. I actually like this better because there is less risk of losing a specialized field due to limit fields or return fields, or some other optimistic filtering mechanism.

PR #3804, opened 2026-08-04
Protect against aggregating problematically large documents (part 1)

SethSmucker:
Part of #3656

Document aggregation in KeyToDocumentData.collectDocumentAttributes is unbounded. An exceptionally large document gets buffered in its entirety and can exhaust tserver memory, and the existing doc.agg.threshold stopwatch only logs a warning after the fact. I added a doc.agg.max.entries option (default -1, disabled) and an EventDataQueryEntryLimitFilter that wraps whatever event evaluation filter the query already uses. Once a document hits the limit, the filter rejects the remaining entries, adds a synthetic INCOMPLETE_DOCUMENT marker field so the truncated document is identifiable in results, and seeks past the rest of the document so the excess entries are not read at all.

This is the first tier of the issue and covers standard event queries only. The TLD, ancestor, and parent iterators supply their own evaluation filters and are left for later tiers, along with the proposed non-query-field reductions. New tests cover the filter behavior and the config round-trip.


warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -2551,6 +2551,7 @@ protected void configureSeekingOptions(IteratorSetting cfg, ShardQueryConfigurat
 
         if (config.isSeekingEventAggregation()) {
             addOption(cfg, QueryOptions.SEEKING_EVENT_AGGREGATION, String.valueOf(config.isSeekingEventAggregation()), false);
+            addOption(cfg, QueryOptions.DOC_AGGREGATION_MAX_ENTRIES, String.valueOf(config.getDocAggregationMaxEntries()), false);
```
apmoriarty: This option should be independent of a seeking aggregation.

apmoriarty:
I see this is marked as part one. Please include an integration test that uses AbstractQueryTest and AbstractIngest to demonstrate the event short circuit behavior. 

It's fine if the same large event is ingested repeatedly with different event ids. 

Tests should demonstrate queries that hit entirely within the aggregation window, and also include queries with terms that fall outside the aggregation window (and then document results not coming back, or document an exception to the max size threshold).

warehouse/query-core/src/test/java/datawave/query/LargeDocumentProtectionIT.java
```diff
@@ -0,0 +1,256 @@
...
+     */
+    @Test
+    public void testIndexedTermOutsideWindowDropsDocument() throws Exception {
+        maxEntries = LIMIT;
+        givenDate("20260708");
+        givenQuery("ZEBRA == 'stripes'");
+        expectPlan("ZEBRA == 'stripes'");
+        expectResultCount(0);
+        planAndExecuteQuery();
+    }
```
apmoriarty, line 224: This is perfect, exactly what the edge case I wanted to see demonstrated in a test

PR #3809, opened 2026-08-05
Move tf aggregrator right before versioning

jschmidt10:
The tf aggregator is clobbering the visibilityFilter being added by ScannerHelper (and a few other spots). We want visibility filtering to occur first so I moved tf aggregation immediately before VersionIter.


apmoriarty (approved):
Verified priority 19 is free for the shard table, changes look good.

avgAGB (approved):

PR #3810, opened 2026-08-07
Upgrade in-memory-accumulo to Accumulo 2.1.5

apmoriarty:
2.1.5 removed IteratorConfigUtil.generateInitialTableProperties(boolean) in favor of separate accessors for the default iterators and the key size constraint, so InMemoryTable now assembles the initial table properties itself, applying the versioning iterators only when limitVersion is set.


apmoriarty:
No effects until tag generated and integrated into main repo

FineAndDandy (approved):

jschmidt10 (approved):

PR #3811, opened 2026-08-07
Upgrade Accumulo to 2.1.5

apmoriarty:
Pick up in-memory-accumulo 4.0.6, which is the first release built against Accumulo 2.1.5.

2.1.5 registers an UpgradeUtil KeywordExecutable that every Accumulo process service-loads at startup. It uses JCommander's IUsageFormatter, added in 1.78, so with jcommander 1.72 on the classpath MiniAccumuloCluster fails to launch ZooKeeper and the failure surfaces as a misleading ZooKeeperBindException.


apmoriarty:
Will not build until https://github.com/NationalSecurityAgency/datawave/pull/3810 is merged and tagged

alerman (approved):

FineAndDandy (approved):

PR #3814, opened 2026-08-09
Add usage of String constants for MetadataColumnFamilyConstants

### LINKED ISSUE #3712: MetadataColumnFamilyConstants should provide string options
There are many instances where ColumnFamilyConstants is used to access the Text object and then immediately calls toString(). Just provide string alternatives and integrate.

tom-vt4:
There are multiple occurrences where ColumnFamilyConstants are used to access a Text object with toString() called. String alternatives are provided to simplify this.

- Add String equivalents of each text constant, and make Text constants use the new String constants when initializing. Ex: Text COLF_E = new Text(COLF_E_STR)

- Update all referenced variables to use String constant instead of toString() on a Text constant

- Make MetadataColumnFamilyConstants a final class instead of interface

Closes issue #3712


core/tables/src/main/java/datawave/table/constants/MetadataColumnFamilyConstants.java
```diff
@@ -5,89 +5,123 @@
...
 
     /**
      * a colf of 'count' denotes term counts (cardinality)
      */
-    Text COLF_COUNT = new Text("count");
+    public static final String COLF_COUNT_STR = "count";
+    public static final Text COLF_COUNT = new Text(COLF_COUNT_STR);
 
     /**
      * a colf of 'version' debnotes a version (currently only used for edge_key row)
```
lbschanno:
```suggestion
     * a colf of 'version' denotes a version (currently only used for edge_key row)
```

core/tables/src/main/java/datawave/table/constants/MetadataColumnFamilyConstants.java
```diff
@@ -5,89 +5,123 @@
...
+
+    private MetadataColumnFamilyConstants() {
+        throw new UnsupportedOperationException();
+    }
+
     /**
      * A colf of 'e' denotes event fields (shard table)
      */
-    Text COLF_E = new Text("e");
+    public static final String COLF_E_STR = "e";
```
lbschanno, line 17: We have an opportunity here to make these constants self-documenting about their actual values now that we have the string constants. For example, we can now do

```java
/**
 * A column family of {@value} denotes event fields (shard table).
 */
public static final String COLF_E_STR = "e";

/**
 * The column family {@value COLF_E_STR} in {@link Text} form.
 */
public static final Text COLF_E = new Text(COLF_E_STR);
```
and the value "e" will be rendered in the javadocs for both constants. I'd like to see the javadocs updated in this style for all the constants in this class.

lbschanno (changes requested):

lbschanno (dismissed):

PR #3820, opened 2026-08-10
Do not silently drop system limit configurations

### LINKED ISSUE #3818: Query Limit Feature - System limits can be silently  dropped
SystemLimitProvider.populateLimits() can potentially silently drop `SortableSystemLimit` instances when there are two or more systems with the same Matcher type, equal limits, equal group overrides, and equal countsAgainstUserLimit. Add a tiebreaker on the underlying Matcher equality, and add tests to verify fix.

lbschanno:
SystemLimitProvider.populateLimits() can silently drop SortableSystemLimit instances when there are two or more system configurations that result in the same matcher type, equal limits, equal values for countsAgainstUserLimit, and equal group override maps.

Add a check to SortableSystemLimit.compareTo() to also check against matcher equality to ensure we do not drop matchers of the same type with different underlying patterns.

Fixes #3818


apmoriarty (approved):
I think `QueryLogicGroupLimitProvider.createOverrides()` may have the same comparator fall-through issue, but I don't see any issues with the fix for `SystemLimitProvider`

lbschanno:
`QueryLogicGroupLimitProvider.createOverrides()` is fixed in #3819.

avgAGB (approved):

PR #3834, opened 2026-08-13
Verify Ingest Scripts

apmoriarty:
Specifically the scripts that build classpaths, and propagate critical failures


warehouse/ingest-scripts/src/main/resources/bin/ingest/create-shards-since.sh
```diff
@@ -71,8 +71,8 @@ IFS="$OLD_IFS"
 
 TYPES=${BULK_INGEST_DATA_TYPES},${LIVE_INGEST_DATA_TYPES},${COMPOSITE_DATA_TYPES}
 
-ADDJARS=$THIS_DIR/$DATAWAVE_INGEST_CORE_JAR:$THIS_DIR/$COMMON_UTIL_JAR:$THIS_DIR/$DATAWAVE_CORE_JAR:$THIS_DIR/$DATAWAVE_COMMON_UTILS_JAR:$THIS_DIR/$COMMONS_LANG_JAR
+export CLASSPATH
```
drewfarris, line 74: Consider exporting CLASSPATH once in `ingest-libs.sh`? Since `ingest-libs.sh` is sourced from the caller scripts using the `.` operator, it is running in the same shell context as cal caller script. As a result, calling `export CLASSPATH` from `ingest-libs.sh` would expose the `CLASSPATH` variable to anything called by the original script.

drewfarris (approved):

PR #3835, opened 2026-08-13
Support querying offline tables through scan servers

apmoriarty:
Accumulo 2.1.5 can serve an offline table from a scan server, but only for scans issued at EVENTUAL consistency. Four separate places dropped the configured consistency level and execution hints before the scan was built, so no query logic setting could ever reach a scan server:

- RunningResource guarded the copy with 'baseScanner instanceof SessionOptions', which is never true
- RangeStreamScanner builds its own scanner and never copied either setting
- PushdownFunction looked the maps up by table id instead of table name, which only worked because InMemoryAccumuloClient reports the name as the id
- the index expansion helpers replaced the session options wholesale

ScanServerOfflineTableIT covers this against a MiniAccumuloCluster with a scan server, querying shard and shardIndex while both are offline.

### DIFF
--- a/warehouse/query-core/src/test/java/datawave/query/ScanServerOfflineTableIT.java
+++ b/warehouse/query-core/src/test/java/datawave/query/ScanServerOfflineTableIT.java
@@ -1,0 +1,547 @@
+package datawave.query;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.net.URL;
+import java.nio.file.Path;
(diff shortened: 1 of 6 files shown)


drewfarris (approved):

warehouse/query-core/src/test/java/datawave/query/ScanServerOfflineTableIT.java
FineAndDandy, line 1: great use of the new test framework

FineAndDandy (approved):

dlmarion:
https://github.com/apache/accumulo/pull/6146, included in 2.1.5, adds a server-side property to restrict which tables a scan server group can scan (online or offline). It defaults to all non-accumulo-namespace tables. Just wanted to bring it to your attention in case you wanted to add a test for the failure case (performing a scan using scan servers against a disallowed table).

PR #3837, opened 2026-08-15
BaseIngestHelper: Normalize once

jalphonso:
`BaseIngestHelper` double normalizes  
Default ingest path should not double normalize for the same value  
```
copy.setEventFieldValue(datawaveType.normalize(copy.getIndexedFieldValue))
copy.setIndexedFieldValue(datawaveType.normalize(copy.getIndexedFieldValue)
```
The normalized value should be created once, outside the setters.  
Since the value is a String immutability is inherently preserved.


avgAGB (approved):

jalphonso:
Deferring this PR in favor of #3842

PR #3838, opened 2026-08-16
ShardedDataTypeHandler: overload createKey and createIndexKey and clean up intermediate objects

jalphonso:
## Summary
Reduces per-key object allocation in `ShardedDataTypeHandler`'s hot ingest paths by adding `createKey`/`createIndexKey` overloads and byte-array builder helpers, so callers no longer have to wrap `String`/`byte[]` values in intermediate `Text` objects just to satisfy the existing method signatures. It also removes the last remaining cases where UTF-8 encoding for key construction could throw a checked exception on malformed input.

## Motivation
`ShardedDataTypeHandler` builds one or more Accumulo `Key`s for every field of every ingested event (shard event column, field index, global index, bitset/day/year index, dictionary index). The original `createKey`/`createIndexKey` methods only accepted `Text` for `colf`/`colq`, forcing every caller to allocate `new Text(...)` (and often `TextUtil.textAppend(...)`, which does an encode-then-possibly-grow-then-copy) purely to satisfy the signature, even when the caller already had a `String` or `byte[]`.

## Changes
- **New `createKey`/`createIndexKey` overloads** accepting `byte[]`/`String` combinations for row/colf/colq, all funneling through a single private `buildKey(...)` that constructs the `Key` directly from backing arrays + explicit lengths (avoiding `Text.getBytes()` copies).
- **`utf8(String)`** — a single UTF-8 encoding helper built on the JDK's `String.getBytes(StandardCharsets.UTF_8)`, which never throws: malformed input (e.g. an unpaired surrogate) is silently replaced, byte-for-byte identical to `new Text(s)`. This replaces an earlier `utf8(String, boolean replaceMalformedUTF8)` overload that wrapped `Text.encode(s, replaceMalformedUTF8)` and could throw `CharacterCodingException`/`IllegalArgumentException` when `replaceMalformedUTF8` was `false` — see Behavioral notes below.
- **`joinWithNulls(byte[]... parts)`** — concatenates byte segments with `NUL` separators in one pre-sized array, replacing the `new Text(x)` + repeated `TextUtil.textAppend(...)` pattern used to build multi-part `colf`/`colq` values (e.g. `fieldName\0fieldValue`, `"fi"\0fieldName`, `fieldValue\0dataType\0uid`).
- **`shardPrefixedBytes(byte[] shardId, int prefixLen, byte[] suffix)`** — builds `shardId[0,prefixLen) + '\0' + suffix` directly from bytes.
- Rewrote `writeBitsetIndexKey`, `writeShardDayIndexKey`, and `writeShardYearIndexKeyBitSet` to stop round-tripping `shardId`/`visibility` through `new String(...)` and Accumulo's `Key(String, String, String, String, long)` constructor (which itself allocates 4 `Text` objects internally); they now build keys directly via `buildKey`, reusing the shared colf/colq bytes between the masked and unmasked key writes.
- Updated the field index (`createShardFieldIndexColumn`, both overloads — the highest-volume key type), shard event column (`createShardEventColumn`/`createMaskedShardEventColumn`), and global index (`writeStandardIndexKey`, `createIndexColumn`) builders to construct `colf`/`colq` as `byte[]` via `joinWithNulls`, sharing the common `dataType\0uid` suffix between masked/unmasked variants where applicable.
- `createShardEventColumn`/`createMaskedShardEventColumn` now take `colf` as `byte[]` instead of `Text`. The `colf` built in `createColumns` (`dataType\0uid`) is assembled directly via `joinWithNulls`/`utf8`, which removes the last `TextUtil.textAppend` call — and its `Text` allocation — from `ShardedDataTypeHandler`.
- `ErrorShardedDataTypeHandler` and `ExtendedContentIndexingColumnBasedHandler` updated to use the existing `ColumnFamilyConstants.FULL_CONTENT_TEXT` constant instead of allocating `new Text(ColumnFamilyConstants.FULL_CONTENT)` per call, and to use the new `String`-based overloads where applicable.
- The `replaceMalformedUTF8` flag no longer changes UTF-8 encoding behavior anywhere in key construction: every `ShardUtil.utf8(String)` call (field names, field values, uid/datatype components) now uses the same never-throwing encoder. The flag is still threaded through a couple of method signatures (e.g. `createMaskedShardEventColumn`) for compatibility with overriding subclasses, but it is no longer consulted there.
- `ShardedDataTypeHandler.getValueForBitsetIndex` now delegates to `DateIndexUtil.getValueForDayIndex` instead of duplicating its shard-offset parsing. `DateIndexUtil.getOffsetForYearIndex` gained the same underscore-presence guard `getOffsetForDayIndex` already had. `ShardUtil.joinWithNulls()` with zero arguments now returns an empty array instead of throwing `NegativeArraySizeException`, and `ShardUtil` has a private constructor.

## Behavioral notes
- Index-key rows built from a `String` (e.g. field values used as the row) are now explicitly UTF-8 encoded via the new `utf8()` helper, whereas some previously used `String.getBytes()` (platform default charset) for the row while `Text`-based colf/colq were always UTF-8. This only changes behavior on JVMs where the default charset isn't UTF-8.
- **Key construction can no longer fail a record due to malformed UTF-8 field values.** Previously, `ShardUtil.utf8(s, false)` (the default, since `replaceMalformedUTF8` defaults to `false`) threw on malformed input, matching the old `Text.encode(s, false)`/`TextUtil.textAppendNoNull` behavior it replaced. That throw was **not** governed by `BaseIngestHelper.FailurePolicy` (DROP/LEAVE/FAIL) — that policy only applies to normalization errors in `applyNormalizationAndAddToResults`, which runs before key building — so a malformed field value would fail the whole record: `EventMapper`'s catch rolled back every key already written for the event and routed it to the error table, while `ShardReindexMapper.processDataMap` has no try/catch around `processBulk` at all and would fail the job outright. This PR removes that failure mode entirely by making `ShardUtil.utf8(String)` always replace malformed characters instead of throwing, matching `Text`'s own "replace" semantics.
- All other changes are allocation/copy reductions only — key byte layouts are unchanged, verified via the existing test suite.

## Testing
- `mvn -pl warehouse/ingest-core test -Dtest='datawave.ingest.mapreduce.handler.**'` — 115 tests, all passing.
- `mvn -pl warehouse/ingest-core,warehouse/ingest-wikipedia,warehouse/ingest-csv,warehouse/ingest-json -DskipTests test-compile` — all modules compile successfully.
- `ShardUtilTest` differentially checks `ShardUtil.utf8(String)` against `Text`'s constructor (and `Text.encode(s, true)`) over a corpus including lone/reversed surrogates, asserting it always matches `Text`'s byte output and never throws — this is the permanent regression guard for the non-throwing behavior.

---
[Conversation](https://app.warp.dev/conversation/9a98335f-b732-42b4-a469-2cc768f937c0)

Co-Authored-By: Warp <agent@warp.dev>

### DIFF
--- a/warehouse/core/src/main/java/datawave/ingest/mapreduce/handler/dateindex/DateIndexUtil.java
+++ b/warehouse/core/src/main/java/datawave/ingest/mapreduce/handler/dateindex/DateIndexUtil.java
@@ -144,6 +144,7 @@ public class DateIndexUtil {
      */
     public static int getOffsetForYearIndex(String shard) {
         int index = shard.indexOf('_');
+        Preconditions.checkArgument(index > 0, "shard did not contain an underscore: " + shard);
         String date = shard.substring(0, index);
         Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("GMT"));
         calendar.setTime(DateHelper.parse(date));
--- a/warehouse/core/src/main/java/datawave/ingest/mapreduce/handler/dateindex/DateIndexUtil.java
+++ b/warehouse/core/src/main/java/datawave/ingest/mapreduce/handler/dateindex/DateIndexUtil.java
@@ -105,6 +105,13 @@ public class DateIndexUtil {
         return bits1;
     }
 
+    /**
+     * Constructs a {@link Value} that contains a {@link BitSet}'s backing byte array, with the bit at the day index offset set.
+     *
+     * @param shardId
+     *            the full shard id
+     * @return a Value containing a BitSet with the day index offset bit set
+     */
     public static Value getValueForDayIndex(String shardId) {
         // get the shard number
         int shardNumber = getOffsetForDayIndex(shardId);
@@ -128,6 +135,13 @@ public class DateIndexUtil {
         return Integer.parseInt(bucket);
     }
 
+    /**
+     * Constructs a {@link Value} that contains a {@link BitSet}'s backing byte array, with the bit at the year index offset set.
+     *
+     * @param shard
+     *            the full shard
+     * @return a Value containing a BitSet with the year index offset bit set
+     */
     public static Value getValueForYearIndex(String shard) {
         // get the shard number
         int shardNumber = getOffsetForYearIndex(shard);
@@ -161,6 +175,7 @@ public class DateIndexUtil {
      * trim the event date and ageoff portions of the ts to the beginning of the day
      *
      * @param ts
+     *            the composite timestamp to trim
      * @return the timestamp to be used for index entries
      */
     public static long getIndexTimestamp(long ts) {
@@ -173,6 +188,7 @@ public class DateIndexUtil {
      * Trim ms to the beginning of the day
      *
      * @param date
+     *            the time in milliseconds since the epoch
      * @return the time at the beginning of the day
      */
     public static long trimToBeginningOfDay(long date) {
--- a/warehouse/core/src/main/java/datawave/ingest/mapreduce/handler/dateindex/DateIndexUtil.java
+++ b/warehouse/core/src/main/java/datawave/ingest/mapreduce/handler/dateindex/DateIndexUtil.java
@@ -5,6 +5,16 @@ import java.text.SimpleDateFormat;
 import java.util.BitSet;
 import java.util.Calendar;
 import java.util.Date;
+import java.util.TimeZone;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.accumulo.core.data.Value;
+import org.apache.log4j.Logger;
+
+import com.google.common.base.Preconditions;
+
+import datawave.util.CompositeTimestamp;
+import datawave.util.time.DateHelper;
 
 /**
  *
@@ -15,6 +25,8 @@ public class DateIndexUtil {
     public static final String LOADED_DATE_TYPE = "LOADED";
     public static final String ACTIVITY_DATE_TYPE = "ACTIVITY";
     public static final ThreadLocal<SimpleDateFormat> format = ThreadLocal.withInitial(() -> new SimpleDateFormat("yyyyMMdd"));
+    private static final long MS_PER_DAY = TimeUnit.DAYS.toMillis(1);
+    private static final Logger log = Logger.getLogger(DateIndexUtil.class);
 
     /**
      * Format the date into yyyyMMdd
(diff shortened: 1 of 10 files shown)


warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
```diff
@@ -1148,6 +1145,120 @@ protected byte[] flatten(ColumnVisibility vis) {
...
+     * Encode a String as UTF-8 bytes, honoring the same "replace malformed characters" semantics as {@link TextUtil#textAppend(Text, String, boolean)} /
+     * {@link Text#encode(String, boolean)}, rather than the silent-replace behavior of {@link String#getBytes(java.nio.charset.Charset)}.
+     *
+     * @param s
+     *            the string to encode
+     * @param replaceMalformedUTF8
+     *            whether to replace malformed/unmappable characters instead of throwing
+     * @return the UTF-8 bytes
+     */
+    private static byte[] utf8(String s, boolean replaceMalformedUTF8) {
```
hlgp: These would be better placed in a utility class rather than the datatype handler.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
```diff
@@ -1148,6 +1145,120 @@ protected byte[] flatten(ColumnVisibility vis) {
...
+        try {
+            ByteBuffer buffer = Text.encode(s, replaceMalformedUTF8);
+            byte[] bytes = new byte[buffer.limit()];
+            System.arraycopy(buffer.array(), 0, bytes, 0, bytes.length);
+            return bytes;
+        } catch (CharacterCodingException e) {
+            throw new IllegalArgumentException(e);
```
hlgp: We will want to handle this exception appropriately wherever this method is called.  We do not want to fail an entire job because we had a value we could not encode.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
jalphonso: sounds good, I'll make these and some of the other methods static and move them to a new utility class

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
jalphonso: I considered this, but the code change does not introduce any new behavior. See `TextUtil.textAppend()`
```java
    public static void textAppendNoNull(Text t, String s, boolean replaceBadChar) {
        try {
            ByteBuffer buffer = Text.encode(s, replaceBadChar);
            t.append(buffer.array(), 0, buffer.limit());
        } catch (CharacterCodingException cce) {
            throw new IllegalArgumentException(cce);
        }
    }
```

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
hlgp: Cool thanks.  the shard-specific things can probably stay, but the super generic ones can go in to the util.

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
hlgp: Yep, just a note for us not to lose sight of.

warehouse/core/src/main/java/datawave/ingest/mapreduce/handler/dateindex/DateIndexUtil.java
```diff
@@ -5,6 +5,16 @@
 import java.util.BitSet;
 import java.util.Calendar;
 import java.util.Date;
+import java.util.TimeZone;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.accumulo.core.data.Value;
+import org.apache.log4j.Logger;
```
avgAGB: minor, but would be good to stay consistent with the slf4j library for the Logger.

jalphonso: done

warehouse/ingest-core/src/main/java/datawave/ingest/mapreduce/handler/shard/ShardedDataTypeHandler.java
jalphonso: fixed. replaced several of the Text library calls including using String's built in getBytes method. we no longer throw `CharacterCodingException`s

jalphonso: See `ShardUtil`
```java
    /**
     * Encode a String as UTF-8 bytes. Key components are always UTF-8 encoded.
     * <p>
     * This uses the JDK encoder ({@link String#getBytes(java.nio.charset.Charset)}), which never throws: malformed input (e.g. an unpaired surrogate) is
     * silently replaced, byte for byte identically to {@code new Text(s)} and to {@link Text#encode(String, boolean)} with {@code replace = true}. Key
     * construction is therefore never able to fail a record on account of a malformed field name or value.
     *
     * @param s
     *            the string to encode
     * @return the UTF-8 bytes
     */
    public static byte[] utf8(String s) {
        return s.getBytes(StandardCharsets.UTF_8);
    }
```

apmoriarty (approved):

hlgp (approved):

PR #3842, opened 2026-08-17
Normalize each value once on the normalized-but-not-indexed branch

apmoriarty:
normalizeFieldValue called normalize twice on the same input, and the normalizer is a pure function, so the second call only recomputed the first result. Local benchmark runs measured the normalization phase about 32% faster for a cheap type and 41% for DateType.

### DIFF
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
@@ -761,8 +761,7 @@ public abstract class BaseIngestHelper extends AbstractIngestHelper implements C
         try {
             String normalized = datawaveType.normalize(copy.getIndexedFieldValue());
             copy.setEventFieldValue(normalized);
-            // Separate instance so the two fields never alias.
-            copy.setIndexedFieldValue(new String(normalized));
+            copy.setIndexedFieldValue(normalized);
         } catch (Exception ex) {
             copy.setError(ex);
         }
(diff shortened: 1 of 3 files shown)


warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
```diff
@@ -759,8 +759,10 @@ protected NormalizedContentInterface normalizeFieldValue(NormalizedContentInterf
         // copy it
         NormalizedContentInterface copy = new NormalizedFieldAndValue(normalizedContent);
         try {
-            copy.setEventFieldValue(datawaveType.normalize(copy.getIndexedFieldValue()));
-            copy.setIndexedFieldValue(datawaveType.normalize(copy.getIndexedFieldValue()));
+            String normalized = datawaveType.normalize(copy.getIndexedFieldValue());
+            copy.setEventFieldValue(normalized);
+            // Separate instance so the two fields never alias.
+            copy.setIndexedFieldValue(new String(normalized));
```
jalphonso: Strings are immutable so I do not think this is necessary. Also I had already picked up this ticket but I like your test additions. I just went with existing test suite

jalphonso:
duplicate of https://github.com/NationalSecurityAgency/datawave/pull/3837

warehouse/ingest-core/src/main/java/datawave/ingest/data/config/ingest/BaseIngestHelper.java
apmoriarty: Fixed in https://github.com/NationalSecurityAgency/datawave/pull/3842/commits/1a34e911aafed820450c4b55cbe6bba0da7d667a

avgAGB (approved):

hlgp (approved):

Issue #3846, opened 2026-08-18
AnnotationHitsTransformer may force grouping context or add fields that are never removed


FineAndDandy:
Ensure when grouping notation or extra fields are added due to the AnnotationHitsTransformer that they are removed in all cases as each entry passes through the transformer.

PR #3847, opened 2026-08-18
Ensure AnnotationHitsTransformer removes grouping notation or extra f…

FineAndDandy:
…ields it forced (#3846)

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AnnotationHitsTransformer.java
+++ b/warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AnnotationHitsTransformer.java
@@ -101,7 +101,6 @@ public class AnnotationHitsTransformer extends DocumentTransform.DefaultDocument
     private int contextSize = DEFAULT_CONTEXT_SIZE;
     private float minScore = DEFAULT_MIN_SCORE;
     private TimeUnit timeUnit = DEFAULT_TIMEUNIT;
-    private boolean forcedGroupingNotation = false;
     private List<String> forcedReturnFields = new ArrayList<>();
 
     private Set<Pattern> searchHitTerms;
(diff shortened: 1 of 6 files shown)


alerman:
@gemini-cli /review

alerman:
@gemini-cli /review

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AnnotationHitsTransformer.java
```diff
@@ -267,78 +268,79 @@ public Entry<Key,Document> apply(@Nullable Entry<Key,Document> keyDocumentEntry)
...
+            keyDocumentEntry = stripGroupingNotation(keyDocumentEntry);
+            keyDocumentEntry = removeForcedFields(keyDocumentEntry);
         }
-
-        return keyDocumentEntry;
     }
 
     private Entry<Key,Document> removeForcedFields(Entry<Key,Document> entry) {
-        if (forcedReturnFields.isEmpty()) {
+        if (forcedReturnFields.isEmpty() || entry.getValue() == null) {
```
ivakegg, line 370: include entry == null ?

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AnnotationHitsTransformer.java
```diff
@@ -267,78 +268,79 @@ public Entry<Key,Document> apply(@Nullable Entry<Key,Document> keyDocumentEntry)
...
-                    // strip anything we forced into the Document to complete the query
-                    keyDocumentEntry = stripGroupingNotation(keyDocumentEntry);
-                    keyDocumentEntry = removeForcedFields(keyDocumentEntry);
                 }
             }
+            return keyDocumentEntry;
+        } finally {
+            // strip anything we forced into the Document to complete the query
+            keyDocumentEntry = stripGroupingNotation(keyDocumentEntry);
+            keyDocumentEntry = removeForcedFields(keyDocumentEntry);
```
drewfarris: This one threw me for a bit loop until I read the methods being called. We have a finally block where the variable we're assigning to is effectively ignored because when the finally block is executed, it can't change the object _reference_ that gets returned for `keyToDocumentEntry`

However, both `stripGroupingNotation` and `removeForcedFields` mutate the internal state of the `keyToDocumentEntry` being passed in, so the object referred to in the return method is being mutated. 

It is probably worth rewriting both of these methods to return void so it's clearer that they are mutating what gets passed in vs. the way it's written now that would fool the unfortunate developer into thinking that re-assignment is necessary to propagate the changes each method makes.

warehouse/query-core/src/main/java/datawave/query/transformer/annotation/AnnotationHitsTransformer.java
drewfarris: It's might be worth a second check for tidiness, but it's a private method and the caller has already checked.

FineAndDandy:
I cannot officially approve since I opened the branch, but this now LGTM

drewfarris (approved):

alerman (approved):

PR #3508, opened 2026-04-14
Migrate to Wildfly 26.1.3.Final

lbschanno:
Migrate Datawave to use Wildfly 26.1.3.Final. Additionally, migrate Datawave's security APIs to leverage Wildfly's Elytron security framework instead of Picketbox. Wildfly 26 was chosen due to being Java 17 compatible, but also compatible with Jakarta EE 8 to minimize migration concerns.

## Structural Changes

- A number of commonly-used security classes were relocated to a new project: /commons/security (datawave-commons-security). This project is used by the Datawave webservices and microservices. An overview can be found in the [datawave-commons-security README](https://github.com/NationalSecurityAgency/datawave/blob/task/wildfly-update/commons/security/README.md).

- The project /webservices/security was converted to a parent project and now has three sub-projects. An overview of the projects can be found in the [datawave-ws-security-parent README](https://github.com/NationalSecurityAgency/datawave/blob/task/wildfly-update/web-services/security-parent/README.md).
    - datawave-ws-security
    - datawave-ws-security-elytron
    - datawave-ws-security-elytron-module

## Migration from Picketbox to Elytron

Wildfly dropped support for Picketbox as of Wildfly 25. As a result, we are required to migrate our security solutions to leverage Elytron. The bulk of the new classes that replace the old Picketbox APIs live in webservices/security/datawave-ws-security-elytron-module, and the [README](https://github.com/NationalSecurityAgency/datawave/blob/task/wildfly-update/web-services/security-parent/security-elytron-module/README.md) there goes into detail about the custom Elytron components implemented there.

## JBOSS/Wildfly Modules

Custom Elytron components are required to be deployed within separate Wildfly modules. This means they cannot be packaged with the Datawave EAR deployment. As a result, a number of libraries commonly used between the Datawave Elytron module and the Datawave EAR deployment also had to be moved to their own JBOSS modules to manage classloader conflicts. More details in the [datawave-ws-deploy-application README](https://github.com/NationalSecurityAgency/datawave/blob/task/wildfly-update/web-services/deploy/application/README.md).

## Jackson Downgrade to 2.12.7 

[...]

PR #3850, opened 2026-08-18
Derive the security EJB JNDI name from build.env

SethSmucker:
The Wildfly CLI script hardcodes the "compose" application name in the security EJB's JNDI lookup, so a quickstart deployment fails with a NameNotFoundException naming the compose application. The surrounding pom already parameterises this with ${build.env}, so I used the same property here.


lbschanno:
I suggest pointing this PR at #3508 instead so that we have this change in integration too when that PR is merged. I can merge updates from the wildfly branch into accumulo4-2026-quickstart afterwards.

lbschanno:
Applied this fix directly to the `/tasks/wildfly-update` branch, I'll merge that branch into `accumulo4-2026-quickstart`.

Issue #3766, opened 2026-07-24
Test accumulo 4 with hadoop 3.5, java 17, and datawave


ivakegg:
We need to work through the things we are missing to move us to accumulo 4 and java 17.  accumulo4-2026 is the branch started by our illustrious accumulo team.

lbschanno:
Created branch [accumulo4-2026-quickstart](https://github.com/NationalSecurityAgency/datawave/tree/accumulo4-2026-quickstart). This is where I will track changes needed to get Datawave's docker quickstart to start up.

lbschanno:
Deployed [Zookeeper 3.9.5](https://github.com/NationalSecurityAgency/datawave/packages/2188230) and [Hadoop 3.5.0](https://github.com/NationalSecurityAgency/datawave/packages/2188233) to Datawave's packages in order to be able to build the quickstart.

Will likely need to deploy RC versions of Accumulo 4.0.0-SNAPSHOT to get the quickstart to build. @ivakegg let me know if you have any objections.

lbschanno:
Deployed gov.nsa.datawave.accumulo-4.0.0-3788-RC1 which builds accumulo-4.0.0-SNAPSHOT at commit 6ce09f9b2f. Sha512:
```
1a173b604cf0b1105ecba6892c600df40d5b305ed0f2790c252e908e561c37393a59127afcdee40bd50ca26d557a25729eb5523b650efa97f68bf48162eae96f  accumulo-4.0.0-3788-RC1.tar.gz
```

lbschanno:
Found an Accumulo bug where the Accumulo client does not always display the name of supplied, forbidden legacy options in its error message. This was an issue when figuring out that our accumulo shell commands that used `-p` were outdated.  Created issue [#6501](https://github.com/apache/accumulo/issues/6501) and fix [#6502](https://github.com/apache/accumulo/pull/6502).

lbschanno:
End of week update: the accumulo4-2026-quickstart branch now successfully completes the docker build, but fails to start up in docker compose. Logs showed:
```
quickstart-1  | sed: can't read /opt/datawave/contrib/datawave-quickstart/accumulo/conf/cluster.yaml: No such file or directory
quickstart-1  | cp: cannot stat '/opt/datawave/contrib/datawave-quickstart/hadoop/etc/hadoop/*-site.xml': No such file or directory
[...]

PR #3852, opened 2026-08-18
Wait for a listening tserver client port before starting the web tier

SethSmucker:
datawaveWebReadyToStart polled for a listener on the fixed port 9997, but Accumulo 4 moved tserver.port.client to the 9800-9899 range, so the check never passed and Wildfly never started. Per review, the check now asks accumulo-service which ports the running tservers hold and confirms one of them is being listened on via ss.

Work for #3766

### DIFF
--- a/contrib/datawave-quickstart/bin/services/datawave/bootstrap-web.sh
+++ b/contrib/datawave-quickstart/bin/services/datawave/bootstrap-web.sh
@@ -140,7 +140,9 @@ function datawaveWebIsDeployed() {
 }
 
 function datawaveWebReadyToStart() {
-    ss -ln | grep 8020 && ss -ln | grep 2181 && ss -ln | grep 9997 && return 0
+    # Accumulo 4 tservers no longer bind the fixed 9997 client port (the default
+    # is now the 9800-9899 range), so check for a live tserver process instead
+    ss -ln | grep 8020 && ss -ln | grep 2181 && pgrep -f 'o.start.Main proc tserver' > /dev/null && return 0
     return 1
 }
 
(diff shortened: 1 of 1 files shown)


lbschanno:
Rather than just grepping for the tserver PID, it may be better to first obtain the tserver port via accumulo-service, e.g.:
```
$ accumulo-service tserver list
Currently running tserver processes (fields: process pid port):
tserver_default_1 1605237 9800
```
and then make sure we are listening on that port via `ss -ln`. The accumulo-service script is available within Accumulo's `/bin` folder.

contrib/datawave-quickstart/bin/services/datawave/bootstrap-web.sh
```diff
@@ -139,8 +139,21 @@ function datawaveWebIsDeployed() {
...
 }
 
+# Accumulo 4 tservers no longer bind the fixed 9997 client port (tserver.port.client
+# now defaults to the 9800-9899 range). Ask accumulo-service which ports the running
+# tservers hold, and confirm one of them is actually being listened on. Note that
+# 'list' prints only its header and still exits 0 when nothing is running, so the
+# port lines are what we test, not the exit status.
+function tserverIsListening() {
+    local port
+    for port in $( "${ACCUMULO_HOME}/bin/accumulo-service" tserver list 2>/dev/null | awk 'NR > 1 { print $NF }' | tr ',' ' ' ) ; do
```
lbschanno: Do we need to only confirm that one is listening, or that all of them are listening?

contrib/datawave-quickstart/bin/services/datawave/bootstrap-web.sh
```diff
@@ -139,8 +139,21 @@ function datawaveWebIsDeployed() {
...
+        ss -ln | grep -qE "[:.]${port}([[:space:]]|\$)" && return 0
+    done
+    return 1
```
lbschanno, line 164: Would be nice to have a echo error statement here stating that we're not listening on any tserver ports for easier debugging. Maybe include a list of the ports if present?

lbschanno (changes requested):

PR #3854, opened 2026-08-18
Remove the commons-vfs2 reference from JexlFunctionNamespaceRegistry

SethSmucker:
Edge queries hang with no client error and the tserver halts its VM mid-scan. JexlFunctionNamespaceRegistry does an "instanceof VFSClassLoader" test purely to decide what to put in a debug log line. Accumulo 4 removed VFS classloading and no longer ships commons-vfs2, so EdgeFilterIterator init throws NoClassDefFoundError on a scan thread and the server halts. The reference was dead code, so I removed it along with the now-unused imports.


lbschanno:
Closing this because we're going to replace the VFS classloader in #3875, and the vsf2 jars will become a thing of the past.

PR #3860, opened 2026-08-19
Build the utils, services and starters modules in the quickstart build

SethSmucker:
The quickstart build fails resolving 8.0.0-SNAPSHOT artifacts it cannot get from anywhere: core/in-memory-accumulo (needed by common-test and core/connection-pool), the microservice APIs (core/connection-pool and web-services need accumulo-api) and the spring-boot starters those services depend on. All three are behind property-activated profiles, so I added -Dutils -Dservices -Dstarters to the quickstart's maven command.

That gets the build through, but it does not answer why the default reactor depends on modules its own profiles gate off.


lbschanno:
Typically in-memory-accumulo and the microservices follow their own versioning scheme, and do not stay synchronized with the version defined in the base datawave pom.xml. As such they generally don't need to be included in the quickstart build command. 

Their versions are synchronized in accumulo4-2026-quickstart purely to make it easier to manage conflicts with version changes that appear when merging integration into the branch.

PR #3862, opened 2026-08-19
Load dnutils.properties via the context classloader

SethSmucker:
Authentication fails with "Failed to load properties file dnutils.properties: NullPointerException: inStream parameter is null". DnProperties looks the file up on its own class's classloader, which does not see it in the deployed EAR layout. Added a fall back to the thread context classloader.

This works, but the real question is probably where dnutils.properties should be packaged, so treat it as a stopgap if you would rather fix the packaging.

### DIFF
--- a/commons/security/src/main/java/datawave/security/util/DnProperties.java
+++ b/commons/security/src/main/java/datawave/security/util/DnProperties.java
@@ -58,6 +58,17 @@ public class DnProperties {
      *
      * @return a new {@link DnProperties}
      */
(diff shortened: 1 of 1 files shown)


lbschanno:
This was fixed in [#3508](https://github.com/NationalSecurityAgency/datawave/pull/3508/changes/5f7cb48a47af613d589df9ca48aed15c62c98ea1) and will be merged into accumulo4-2026-quickstart.

PR #3864, opened 2026-08-19
Create DataWave tables with HOSTED tablet availability

SethSmucker:
Accumulo 4 defaults new tablets to ONDEMAND. TableOperations.locate() throws for anything that is not HOSTED, and GenerateShardSplits calls it, so ingest fails. This creates DataWave's tables HOSTED, both in TableConfigurationUtil and in the quickstart's shell init script.

This is the blunt version and I am not sure it is the one you want: forcing HOSTED opts DataWave out of Accumulo 4's elasticity model wholesale. The narrower question is whether GenerateShardSplits should be using locate() at all. I would rather have your call before this goes anywhere.

### DIFF
--- a/contrib/datawave-quickstart/bin/services/datawave/bootstrap.sh
+++ b/contrib/datawave-quickstart/bin/services/datawave/bootstrap.sh
@@ -71,7 +71,9 @@ function createAccumuloShellInitScript() {
    DW_ACCUMULO_SHELL_INIT_SCRIPT="
    createnamespace datawave
    createtable datawave.queryMetrics_m
+   setavailability -t datawave.queryMetrics_m -a HOSTED
    createtable datawave.queryMetrics_s
+   setavailability -t datawave.queryMetrics_s -a HOSTED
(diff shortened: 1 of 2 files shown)


dlmarion:
> The narrower question is whether GenerateShardSplits should be using locate() at all.

I think fixing GenerateShardSplits might be the better solution because  tablets could be made ONDEMAND or NEVER at some point in the future.

Looking at https://github.com/NationalSecurityAgency/datawave/blob/accumulo4-2026-quickstart/warehouse/ingest-core/src/main/java/datawave/ingest/util/GenerateShardSplits.java#L298:

  1. You would only need to countt the tablet locations if the original tablet's availability is set to HOSTED
  2. You can count the tablet locations by using Ample (which will scan the metadata table), like so: https://github.com/apache/accumulo/blob/main/test/src/main/java/org/apache/accumulo/test/functional/ManagerAssignmentIT.java#L453

lbschanno:
@dlmarion could this be done in Accumulo 2.1.x, or would this be Accumulo 4.x only?

lbschanno:
Closing this PR to be worked in issue #3876.

PR #3865, opened 2026-08-19
Warn instead of failing when test user auths cannot be trimmed

SethSmucker:
TestDatawaveUserService reads the Accumulo user's authorizations so it can trim what the test user file supplies. The bean is built during authentication, in the Elytron evidence decoder, before any caller identity exists, so the secured connection factory EJB refuses the call and every login fails with it. This catches the failure, warns, and skips the trimming.

Trimming only affects test users so skipping it is survivable, but this is a symptom fix. The real question is what identity the security layer should be using for its own callouts, and that is yours to answer.

### DIFF
--- a/web-services/security-parent/security/src/main/java/datawave/security/authorization/test/TestDatawaveUserService.java
+++ b/web-services/security-parent/security/src/main/java/datawave/security/authorization/test/TestDatawaveUserService.java
@@ -25,6 +25,8 @@ import javax.interceptor.Interceptor;
 import org.apache.accumulo.core.client.AccumuloClient;
 import org.apache.accumulo.core.security.Authorizations;
 import org.apache.deltaspike.core.api.exclude.Exclude;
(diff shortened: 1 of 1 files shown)


avgAGB (approved):

lbschanno:
This is likely another change that needs to go into #3508, will investigate.

hlgp (approved):

lbschanno:
Please point this PR against the branch `task/wildfly-update`, and leave a comment on PR #3508 requesting review of it so that we can verify if this works for the high-side.

lbschanno:
After a discussion with Ivan, we want to prioritize getting #3508 merged in. This PR should instead be pointed against integration after that happens.

lbschanno:
Additionally we will need to test this fix. I believe the TestDatawaveUserService is not used when we do the standard docker compose quickstart. We will need to test this with the hybrid setup.

PR #3866, opened 2026-08-20
Reorganize datawave-core-tables packages

lbschanno:
The datawave-core-tables module contains classes that are copies of their original counterparts (e.g. datawave.util.CompositeTimestamp). It is the intent to eventually deprecate the originals and use the classes in datawave-core-tables.

The issue is that these classes were copied over while retaining their original packages. This makes it impossible to simultaneously deprecate the originals and update references to use the replacements.

Move the replacement classes into new packages that reflect a consistent naming scheme of `datawave.table.x`. This will make it possible to update references later without deleting the originals, and allow for deprecation.


avgAGB (dismissed):

hlgp (dismissed):

core/tables/src/main/protobuf/TermWeight.proto
```diff
@@ -2,7 +2,7 @@
 //      classpath for compile command should include ../../../target/classes and protobuf-java-2.2.0.jar
 syntax = "proto2";
 
-package datawave.ingest.protobuf;
+package datawave.table.protobuf;
 
 option java_package = "datawave.ingest.protobuf";
```
apmoriarty, line 7: The `option java_package` should change to `datawave.table.protobuf`. If the current edit is accepted today and these files are regenerated new files will get written to `datawave.ingest.protobuf`

Same comment applies to all other proto changes.

lbschanno: Fixed.

apmoriarty (approved):

avgAGB (approved):

PR #3868, opened 2026-08-20
speed up sorted set tests

apmoriarty:
Two changes
- reuse hadoop conf object
- reduce rounds in buffered file map tests

### DIFF
--- a/warehouse/query-core/src/main/java/datawave/query/util/sortedmap/SortedMapTempFileHandler.java
+++ b/warehouse/query-core/src/main/java/datawave/query/util/sortedmap/SortedMapTempFileHandler.java
@@ -15,6 +15,10 @@ import datawave.query.util.sortedset.FileSortedSet;
  * A sorted set file handler factory that uses temporary local based files.
  */
 public class SortedMapTempFileHandler implements FileSortedMap.SortedMapFileHandler {
+    // Configuration re-parses the Hadoop XML resources every time one is constructed, which is far too expensive to repeat for each
+    // persisted file. Nothing here mutates it, so a single shared instance serves every handler.
+    private static final Configuration CONF = new Configuration();
+
     private final FileSystem fs;
     private final File file;
     private final Path path;
@@ -23,8 +27,7 @@ public class SortedMapTempFileHandler implements FileSortedMap.SortedMapFileHand
         this.file = File.createTempFile("SortedSet", ".bin");
         this.file.deleteOnExit();
         this.path = new Path(file.toURI());
-        Configuration conf = new Configuration();
-        this.fs = path.getFileSystem(conf);
+        this.fs = path.getFileSystem(CONF);
     }
 
     public File getFile() {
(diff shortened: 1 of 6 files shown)


warehouse/query-core/src/test/java/datawave/query/util/sortedmap/BufferedFileBackedSortedMapTest.java
```diff
@@ -113,7 +113,7 @@ public boolean isValid() {
 
         // adding in the data map multiple times to create underlying files with duplicate values making the
         // MergeSortIterator's job a little tougher...
-        for (int d = 0; d < 11; d++) {
+        for (int d = 0; d < 4; d++) {
```
alerman, line 116: Without looking at this test further, why the reduction? Any risk here?

apmoriarty: The goal of the test is to introduce duplicates, and we get duplicates with four rounds. The additional 7 rounds are just noise.

warehouse/query-core/src/main/java/datawave/query/util/sortedmap/SortedMapTempFileHandler.java
```diff
@@ -15,6 +15,10 @@
  * A sorted set file handler factory that uses temporary local based files.
  */
 public class SortedMapTempFileHandler implements FileSortedMap.SortedMapFileHandler {
+    // Configuration re-parses the Hadoop XML resources every time one is constructed, which is far too expensive to repeat for each
+    // persisted file. Nothing here mutates it, so a single shared instance serves every handler.
+    private static final Configuration CONF = new Configuration();
```
apmoriarty, line 20: This is a production code change that requires testing

avgAGB:
from running the repo through there looks like there may be a few more spots where we could shave the loop down, would these qualify or do we want to keep the length of these the same?

` 
  - sortedset/BufferedFileBackedSortedSetTest.java:68
  - sortedset/BufferedFileBackedKeySortedSetTest.java:67
  - sortedmap/BufferedFileBackedRewritableSortedMapTest.java:84 (its own testRewrite, separate from the base class's
    @Before)
  - sortedmap/MultiMapBackedSortedMapTest.java:97
  - sortedmap/MultiMapBackedRewritableSortedMapTest.java:75`

apmoriarty:
> from running the repo through there looks like there may be a few more spots where we could shave the loop down, would these qualify or do we want to keep the length of these the same?
> 
> `
> 
> * sortedset/BufferedFileBackedSortedSetTest.java:68
> * sortedset/BufferedFileBackedKeySortedSetTest.java:67
> * sortedmap/BufferedFileBackedRewritableSortedMapTest.java:84 (its own testRewrite, separate from the base class's
>   @before)
> * sortedmap/MultiMapBackedSortedMapTest.java:97
> * sortedmap/MultiMapBackedRewritableSortedMapTest.java:75`

Added in https://github.com/NationalSecurityAgency/datawave/pull/3868/changes/d18bce2265ac7d9d1b90d8615dd208c34b498c17. 
Deferred two instances that are operating on in memory TreeMaps, those tests were executed in less than a second locally.

ivakegg (approved):

avgAGB (approved):

PR #3870, opened 2026-08-20
Fix TypeMetadata serialization race and failure handling

apmoriarty:
toString() built its mini-maps into instance fields, but the metadata helper cache hands the same TypeMetadata to every concurrent query with matching auths, so the first serializations after a cache refresh raced on unsynchronized TreeMaps. They are now built locally and published once.

Where type metadata is missing or malformed, fail instead of degrading: folding away an absent ingest type, or parsing a corrupt mini-map, would normalize values against the wrong types and silently drop valid results. An empty instance also serialized to an unterminated string the tablet servers could not parse.

Planning waits could only iterate on a null result and then spun on a completed future forever, so they now fail fast, and the iterator setting poll blocks on its future rather than busy-waiting on isDone().

### DIFF
--- a/core/utils/metadata-utils/src/main/java/datawave/query/util/TypeMetadata.java
+++ b/core/utils/metadata-utils/src/main/java/datawave/query/util/TypeMetadata.java
@@ -228,6 +228,8 @@ public class TypeMetadata implements Serializable, KryoSerializable {
      *
      * @param ingestTypeFilter
      * @return
+     * @throws IllegalStateException
+     *             if an ingest type is absent, rather than fold it away and normalize its values wrong
      */
     public Multimap<String,String> fold(Set<String> ingestTypeFilter) {
         if (ingestTypeFilter == null || ingestTypeFilter.isEmpty()) {
@@ -236,8 +238,13 @@ public class TypeMetadata implements Serializable, KryoSerializable {
         Multimap<String,String> map = HashMultimap.create();
 
         for (String type : ingestTypeFilter) {
+            Multimap<String,String> types = this.typeMetadata.get(type);
+            if (types == null) {
+                throw new IllegalStateException("No TypeMetadata for ingest type '" + type + "'. Known ingest types: " + this.typeMetadata.keySet()
+                                + ". The metadata table cache is likely incomplete; proceeding would normalize values against the wrong types and silently drop valid results.");
+            }
             // defensive copy
-            map.putAll(HashMultimap.create(this.typeMetadata.get(type)));
+            map.putAll(HashMultimap.create(types));
         }
         return map;
     }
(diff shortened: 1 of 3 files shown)


warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -3510,8 +3530,9 @@ protected CompositeMetadata getCompositeMetadata() {
         if (compositeMetadata == null && compositeMetadataCallable != null) {
             TraceStopwatch stopwatch = stageStopWatch.newStartedStopwatch(compositeMetadataCallable.stageName());
             try {
-                while (compositeMetadata == null) {
-                    compositeMetadata = compositeMetadataFuture.get(concurrentTimeoutMillis, TimeUnit.MILLISECONDS);
+                compositeMetadata = compositeMetadataFuture.get(concurrentTimeoutMillis, TimeUnit.MILLISECONDS);
+                if (compositeMetadata == null) {
+                    throw new ExecutionException(new IllegalStateException("CompositeMetadata was null"));
```
apmoriarty, line 3536: Worth pointing out that this no longer retries, so the concurrentTimeoutMillis may need to be increased. It is now the max wait time before a query fails outright.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -592,7 +592,9 @@ protected CloseableIterable<QueryData> startRangeProcessing(ScannerFactory scann
         stopwatch = timers.newStartedStopwatch("DefaultQueryPlanner - Construct IteratorSettings");
 
         if (!config.isGeneratePlanOnly()) {
+            cfg = getQueryIterator(metadataHelper, config, "", false, false);
```
avgAGB: this might only need to be done when the cfg is not already resolved, so it could be wrapped around a check to save some potential cycles. Not urgent of a change though.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
```diff
@@ -2394,6 +2396,24 @@ protected void configureAdditionalOptions(ShardQueryConfiguration config, Iterat
...
+    /**
+     * Blocks until the iterator setting future completes, so that polling {@link #getQueryIterator} is not a hot spin. A failed future is left for the next
+     * getQueryIterator call to surface.
+     */
+    private void awaitSettingFuture() {
+        if (null == settingFuture) {
+            return;
+        }
+        try {
+            settingFuture.get();
```
avgAGB, line 2409: small note, would it be worthwhile to call the get here similar to the other futures in the class? similar to:
`compositeMetadataFuture.get(concurrentTimeoutMillis, TimeUnit.MILLISECONDS);`

apmoriarty: Tracing the calls from here, this is fine. The future has already completed.

warehouse/query-core/src/main/java/datawave/query/planner/DefaultQueryPlanner.java
apmoriarty: Will do this shortly

core/utils/metadata-utils/src/main/java/datawave/query/util/TypeMetadata.java
```diff
@@ -293,31 +300,55 @@ private static String[] parse(String in, char c) {
...
         // types:[0:DateType,1:IntegerType,2:LcType]
 
         // remove type designation and leading/trailing brackets
-        String types = typeEntry.split(":\\[")[1];
+        String[] split = typeEntry.split(":\\[");
+        if (split.length < 2) {
+            throw new IllegalStateException("Malformed TypeMetadata mini-map, expected a bracketed list: '" + typeEntry + "'");
+        }
+
+        String types = split[1];
```
alerman, line 322: This change ensures that the mini-map starts with [. Should we also confirm it ends with ] before substringing it and accepting the value?

alerman (changes requested):

core/utils/metadata-utils/src/main/java/datawave/query/util/TypeMetadata.java
apmoriarty: Fixed, along with several other instances.

avgAGB (approved):

alerman (approved):

PR #3872, opened 2026-08-20
Add rendezvous sticky balancing option to tiered volume storage

alerman:
CODEX assisted

### DIFF
--- a/warehouse/accumulo-extensions/src/test/java/datawave/ingest/table/volumeChoosers/ShardedTableDateBasedTieredVolumeChooserTest.java
+++ b/warehouse/accumulo-extensions/src/test/java/datawave/ingest/table/volumeChoosers/ShardedTableDateBasedTieredVolumeChooserTest.java
@@ -1,13 +1,18 @@
 package datawave.ingest.table.volumeChoosers;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
-import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


warehouse/accumulo-extensions/src/test/java/datawave/ingest/table/volumeChoosers/ShardedTableDateBasedTieredVolumeChooserTest.java
```diff
@@ -22,309 +27,215 @@
...
+    public void testRendezvousDistribution() {
+        Set<String> volumes = Set.of("volumeA", "volumeB");
+        Map<String,Integer> counts = new LinkedHashMap<>();
+        counts.put("volumeA", 0);
+        counts.put("volumeB", 0);
+        for (int i = 0; i < 10_000; i++) {
+            String choice = ShardedTableDateBasedTieredVolumeChooser.chooseRendezvous(TABLE_ID, new Text("tablet-" + i), volumes);
+            counts.put(choice, counts.get(choice) + 1);
+        }
+        counts.forEach((volume, count) -> assertTrue(volume + " received " + count + " assignments", count >= 4_000 && count <= 6_000));
```
jschmidt10, line 171: I dig this unit test but this is allowing for a 20% imbalance (allows +/- 1000 when the desired goal is 5000). Does the test pass with like a 5% for instance or maybe even less? I was curious about how balanced these would be and it could cause headaches if its actually a 20% difference in prod.

PR #3880, opened 2026-08-24
Add the query test framework leaf generators

apmoriarty:
Field name, event id and value generators for the test framework. These depend only on the util package, so they carry no metadata coupling and can be reviewed on their own.


alerman (dismissed):
First-pass review — three issues to address:

1. `generateCount(0)` does not honor the “first N event ids” contract. In `SequentialEventIdGenerator`, it adds forever because the stop condition is `frequencies.size() == count` after the first element has already been added; in practice this can run until OOM. The modulo, squares, and Fibonacci implementations instead return one element for a zero count. Please either reject non-positive counts centrally or return an empty list consistently, and add shared zero/negative-count coverage.

2. Repeated calls to `NumericFieldNameGenerator.generate(n)` append `1_1...n_1` again, producing duplicates, while `AlphabeticFieldNameGenerator.generate(n)` deliberately continues from the current list size. Please make the repeat-call semantics consistent (or reset explicitly) and test a second call.

3. `NumericFieldNameGeneratorTest.testGetRandomFields` is flaky: `Collections.shuffle` can legally preserve the original order, so `assertNotEquals(expected, fieldNames)` fails with probability 1/7!. The test should verify membership/size without requiring a different permutation, or the API should accept the seeded `Random` used by the framework if deterministic shuffling is part of the contract.

The generator shapes and overflow guards otherwise look sensible on this pass, and all required checks are currently green.

core/datawave-core-test/src/main/java/datawave/test/framework/generators/id/SequentialEventIdGenerator.java
```diff
@@ -0,0 +1,37 @@
...
+        for (long i = 1; i <= bound; i++) {
+            long eventId = i + offset;
+            if (!fitsInInt(eventId)) {
+                break;
+            }
+
+            if (eventId <= bound) {
+                frequencies.add((int) eventId);
+            }
+            if (frequencies.size() == count) {
```
alerman: generateCount(0) can cause a OOM here as add is called before this check and the default bound is Integer.MAX_VALUE

alerman (changes requested):

core/datawave-core-test/src/main/java/datawave/test/framework/generators/id/SequentialEventIdGenerator.java
apmoriarty: Forthcoming

avgAGB (approved):

alerman (approved):

PR #3884, opened 2026-08-24
Use tablet information instead of locate() in GenerateShardSplits

### LINKED ISSUE #3876: Make GenerateShardSplits handle non-HOSTED tablets
`GenerateShardSplits.getTabletLocations()` obtains the locations of tablets via `TableOperations.locate()` which will throw an exception for a tablet that is not HOSTED. Accumulo 4 defaults tablets to ONDEMAND. Update this method to avoid throwing unnecessary exceptions based on this [comment](https://github.com/NationalSecurityAgency/datawave/pull/3864#issuecomment-5355536594).

If this fix will not work in Accumulo 2.1.X, the PR should point to branch `accumulo-2026-quickstart`.

SethSmucker:
Fixes #3876

GenerateShardSplits waited for new tablets to balance by polling TableOperations.locate(), which throws for tablets that are not HOSTED — and new tablets default to ONDEMAND in Accumulo 4. Following Dave Marion's suggestion, tablet locations now come from getTabletInformation(), and the balance wait only happens when the new tablets actually have HOSTED availability, since tablets with other availabilities never get a location to wait for.

Two notes on the details:

* I used the public getTabletInformation() API rather than Ample directly — same metadata, but it avoids adding another non-public-API usage while we're removing them elsewhere.
* Only CURRENT locations count toward the balance check; a FUTURE location is an assignment still in flight.

Compile-verified against Accumulo 4.0.0-SNAPSHOT built from main at 6ce09f9b2f (the RC commit); not run against a live cluster yet. Since getTabletInformation() is @since 4.0.0, this does not backport to 2.1.x as-is, which matches targeting this branch.

Heads up: this branch won't build at all until the datawave-core-test parent version fix goes in (separate PR) — the integration merge left it at 7.45.0-SNAPSHOT.

### DIFF
--- a/warehouse/ingest-core/src/main/java/datawave/ingest/util/GenerateShardSplits.java
+++ b/warehouse/ingest-core/src/main/java/datawave/ingest/util/GenerateShardSplits.java
@@ -3,26 +3,19 @@ package datawave.ingest.util;
 import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
 import java.util.Date;
-import java.util.HashSet;
 import java.util.List;
-import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.accumulo.core.client.Accumulo;
 import org.apache.accumulo.core.client.AccumuloClient;
-import org.apache.accumulo.core.client.AccumuloException;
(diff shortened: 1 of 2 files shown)


dlmarion:
@SethSmucker - reviewing this I'm left wondering *why* the script cares about the tablet location, and why it's waiting for balancing. Do you know?

PR #3893, opened 2026-08-28
Add core annotation module changes from PR #3710

drewfarris:
Extracts the core/annotation changes (AnnotationV1 protobuf definitions and AnnotationUtils) from the feature/annotation-service branch so the core changes can be merged and released independently of the microservices annotation service, which will depend on this released version. The follow on is #3894 which needs to be separate based on exiting CI/CD limitations


tharke13 (approved):
looks good to me

avgAGB (approved):

PR #3899, opened 2026-08-31
Add the query test framework metadata core

apmoriarty:
Describes the fields an ingest will generate before any data is written, so both the writers and the query generator can work from one plan.

### DIFF
--- a/core/datawave-core-test/src/main/java/datawave/test/framework/FieldMetadata.java
+++ b/core/datawave-core-test/src/main/java/datawave/test/framework/FieldMetadata.java
@@ -1,0 +1,334 @@
+package datawave.test.framework;
+
+import static datawave.test.framework.util.MetadataColumn.T;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Preconditions;
+
+import datawave.data.type.Type;
+import datawave.test.framework.generators.id.EventIdGenerator;
+import datawave.test.framework.generators.value.ValueGenerator;
+import datawave.test.framework.util.MetadataColumn;
+
+/**
+ * Contains all information required to generate a field and its associated values
+ * <p>
+ * The {@link ValueGenerator} determines what kind of values are generated
+ * <p>
+ * The {@link EventIdGenerator} determines the distribution of values.
+ * <p>
+ * Note: if the value generator is a singleton the values will be repeated across events
+ * <p>
+ * If the event id count is greater than the number of values then repeats are possible.
+ */
+public class FieldMetadata {
+
+    private static final Logger log = LoggerFactory.getLogger(FieldMetadata.class);
+
+    private final String fieldName;
+    private List<MetadataColumn> metadataColumns;
+    private List<Type<?>> normalizers;
+    private List<String> datatypes;
+
+    private int offset;
+    private List<Integer> eventIds = null;
+    private Map<Integer,Integer> eventIdIndex = null;
+    private List<String> values = null;
+
+    // built on first use and discarded whenever the values or event ids change, so repeated query generation does not rescan the event ids per value
+    private Map<String,List<Integer>> eventIdsByValue = null;
+
+    private boolean isFieldNumeric = false;
+    private ValueGenerator<?> valueGenerator;
+    private EventIdGenerator eventIdGenerator;
+
+    public FieldMetadata(String fieldName) {
+        Preconditions.checkNotNull(fieldName, "FieldName cannot be null");
+        this.fieldName = fieldName;
+    }
+
+    /**
+     * Set the normalizers applied to this field's values.
+     * <p>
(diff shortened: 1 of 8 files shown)


avgAGB (dismissed):

sylvesterkaczmarek (commented):
build() passes the builder-owned metadataColumns and normalizers lists directly into IngestMetadata, and the built object keeps those same mutable instances. Reusing the builder can therefore mutate an already-built metadata plan and change its plan()/createEvents() behavior. Could the constructor defensively copy these lists (and expose immutable views), with a builder-reuse regression?

avgAGB (approved):

sylvesterkaczmarek (commented):
Rechecked current `a86dddca`. My builder-reuse concern is resolved: `IngestMetadata` snapshots both base lists with `List.copyOf`, so later builder/list mutations cannot alter an already-built plan. No remaining blocker from my earlier review.

core/datawave-core-test/src/main/java/datawave/test/framework/FieldMetadata.java
```diff
@@ -0,0 +1,334 @@
...
+        return metadataColumns;
+    }
+
+    public List<String> getDatatypes() {
+        return datatypes;
+    }
+
+    public List<String> getValues() {
+        Preconditions.checkNotNull(values, "FieldMetadata values were not created");
+        return values;
```
drewfarris, line 145: It looks like this may be mutable array reference that's returned, so any changes to this list won't be reflected in getEventIdsForValue. Not certain whether this is a true concern, but worth considering.

core/datawave-core-test/src/main/java/datawave/test/framework/FieldMetadata.java
```diff
@@ -0,0 +1,334 @@
...
+        // valuesPerField is a fixed property of the field's configuration and must not vary with the event distribution, otherwise the number of queries
+        // generated from these values (one per distinct value) would change whenever the event count changes. A value with no backing events is a valid
+        // "matches nothing" test case; writers must skip persisting such a value rather than this method dropping it from the list.
+        for (int i = 0; i < valuesPerField; i++) {
+            values.add(String.valueOf(valueGenerator.next()));
+        }
+    }
+
+    public List<Integer> getEventIds() {
+        return eventIds;
```
drewfarris, line 221: Potentially a similar issue to line 144 where mutable array reference that's returned, so any changes to this list won't be reflected in getEventIdsForValue. Not certain whether this is a true concern, but worth considering.

core/datawave-core-test/src/main/java/datawave/test/framework/FieldMetadata.java
```diff
@@ -0,0 +1,334 @@
...
+
+    private final String fieldName;
+    private List<MetadataColumn> metadataColumns;
+    private List<Type<?>> normalizers;
+    private List<String> datatypes;
+
+    private int offset;
+    private List<Integer> eventIds = null;
+    private Map<Integer,Integer> eventIdIndex = null;
+    private List<String> values = null;
```
drewfarris, line 45: It might be helpful to include a little javadoc here.

core/datawave-core-test/src/main/java/datawave/test/framework/IngestMetadata.java
```diff
@@ -0,0 +1,375 @@
...
+
+                int modCount = modCountIterator.next();
+                for (int offset = 0; offset < FIELDS_PER_PAIRING; offset++) {
+                    FieldMetadata field = new FieldMetadata(fieldNameIterator.next());
+                    field.setFieldNumeric(fieldNameGenerator.isNumeric());
+                    field.setMetadataColumns(metadataColumn);
+                    field.setDatatypes(baseDatatypes);
+                    field.setNormalizers(List.of(normalizer));
+
+                    ValueGenerator<?> valueGenerator = isContentCombo ? PhraseGenerator.create(random) : getValueGeneratorForType(normalizer);
```
drewfarris, line 284: Do we need to check whether the generated values are unique after normalization? (e.g., 'Tt' vs 'tt'). I'm thinking we may have an issue where we get a single query result, bu expect multiple keys.

core/datawave-core-test/src/main/java/datawave/test/framework/IngestMetadata.java
```diff
@@ -0,0 +1,375 @@
...
+        modCounts = modCounts.subList(1, modCounts.size());
+        Iterator<Integer> modCountIterator = new InfiniteIterator<>(modCounts);
+
+        List<FieldMetadata> metadata = new ArrayList<>();
+        for (List<MetadataColumn> metadataColumn : metadataColumns) {
+            for (Type<?> normalizer : baseNormalizers) {
+                boolean isContentCombo = metadataColumn.contains(TF);
+                if (isContentCombo && !(normalizer instanceof LcNoDiacriticsType)) {
+                    // phrases are text-only; skip pairings a phrase generator cannot service
+                    continue;
```
drewfarris, line 273: In addition to skipping unsupported TF/normalizer pairings, could we add something to the builder that asserts that if someone adds a TF metadataColumn that a LcNoDiacriticsType normalizer is added as well? Otherwise we could end up with some weird combos. Do we want to force it to be only LcNoDiacriticsType?

apmoriarty: There's nothing preventing someone from applying a different normalizer to a TF field, but I like enforcing this as a default. The exceptional cases can come later.

avgAGB (approved):

PR #3900, opened 2026-08-31
Execute unions containing negated terms in DocIdIteratorVisitor

apmoriarty:
A union whose children were all non-executable echoed back the caller's own ScanResult, so the enclosing intersection's removeAll wiped every true positive and "A && !(B || C)" returned nothing. Such unions are now evaluated against the intersection's candidate set via De Morgan's law rather than skipped outright.


avgAGB (dismissed):

sylvesterkaczmarek (commented):
In a mixed union, treating a non-executable positive disjunct as “never matches” can create false negatives. For `A && (NON_INDEXED == "z" || !C)`, a document with `NON_INDEXED=z` and `C` satisfies the query, but this path skips the positive term and removes it because of `C`. Non-executable means “cannot evaluate here”, not false. Could this fall back when any positive disjunct is non-executable, with a matching regression?

sylvesterkaczmarek (commented):
Rechecked the current head. The false-negative case I raised is addressed by treating an undecidable union disjunct as unbounded rather than false; the new ScanResult algebra propagates that conservatively and refuses an unbounded top-level result instead of dropping matches. My earlier concern is resolved.

avgAGB (approved):

PR #3902, opened 2026-09-01
Adds the ability to rewrite responses from the AnnotationManagerBean

drewfarris:
In some cases, we'll need to intercept the response from the AnnotationManagerBean and rewrite it to modify the contents in caller or user-specific ways. This PR adds the interface for a JAX-RS response rewriter that can be used for any JAX-RS implementation, updates AnnotationManagerBean to support a pluggable response rewriter and adds tests using a Mock rewriter.

This commit refactors AnnotationManagerRequestContext to be a top level class instead of a static inner class of AnnotationManagerBean.

### DIFF
--- a/web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
+++ b/web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
@@ -10,6 +10,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
+import javax.annotation.PostConstruct;
 import javax.annotation.Resource;
 import javax.annotation.security.DeclareRoles;
 import javax.annotation.security.RolesAllowed;
@@ -41,7 +42,7 @@ import datawave.annotation.protobuf.v1.AnnotationSource;
 import datawave.annotation.protobuf.v1.Segment;
 import datawave.configuration.spring.SpringBean;
 import datawave.core.common.connection.AccumuloConnectionFactory;
-import datawave.core.query.logic.ResponseRewriter;
+import datawave.webservice.common.rest.ResponseRewriter;
 import datawave.webservice.query.exception.QueryException;
 import datawave.webservice.query.result.event.ResponseObjectFactory;
 import datawave.webservice.query.runner.AccumuloConnectionRequestBean;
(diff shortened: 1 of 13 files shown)


apmoriarty (approved):

web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
```diff
@@ -85,6 +74,10 @@ public class AnnotationManagerBean implements AnnotationManager {
     @SpringBean(name = "AnnotationManagerConfig")
     private AnnotationManagerConfig config;
 
+    @Inject
+    @SpringBean(name = "AnnotationResponseRewriter", required = false)
+    private ResponseRewriter responseRewriter;
```
apmoriarty, line 80: Consider a `@PostConstruct` that logs whether the response rewriter is null/non-null

web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
```diff
@@ -98,14 +91,14 @@ public void setEJBContext(EJBContext ctx) {
...
-        final RequestContext context = newRequestContext();
+        final AnnotationManagerRequestContext context = newRequestContext();
         try {
             context.initializeAnnotationService();
             Optional<AnnotationSource> results = context.getAnnotationSource(analyticHash);
             if (results.isEmpty()) {
                 return jsonNotFound("No annotation source found for analyticHash: " + analyticHash);
             }
-            return jsonOk(results.get());
+            return jsonOk(results.get(), context);
```
apmoriarty: Unless the Accumulo client is required for the rewrites, consider moving the rewrite operation outside the try-finally. Otherwise a thread is hanging onto the client well past when the client is in use. 

Here, and throughout the class.

apmoriarty (changes requested):

web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
drewfarris: Done.

web-services/annotations/src/main/java/datawave/webservice/annotation/AnnotationManagerBean.java
drewfarris: Done.

drewfarris:
Replaced by #3908

PR #3905, opened 2026-09-02
Evict metadata helper caches between query-core test classes

apmoriarty:
Groundwork for reusing surefire forks in this module, and inert until that is enabled. The @Cacheable metadata lookups are keyed by {auths, metadataTableName} rather than by Accumulo instance, so with forks reused a test class reads the previous class's field metadata and rejects its own fields as non-existent.

### DIFF
--- a/warehouse/query-core/src/test/java/datawave/query/util/MetadataHelperCacheEvictor.java
+++ b/warehouse/query-core/src/test/java/datawave/query/util/MetadataHelperCacheEvictor.java
@@ -1,0 +1,46 @@
+package datawave.query.util;
+
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.springframework.cache.Cache;
+import org.springframework.cache.CacheManager;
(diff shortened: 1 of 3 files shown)


ivakegg (approved):

avgAGB (approved):

sylvesterkaczmarek (commented):
`query-core` publishes its test output as a `test-jar`, and downstream modules consume that classifier. The new `junit-platform.properties` and ServiceLoader entry therefore export global JUnit extension autodetection into those consumers, potentially running this extension (and other service-loaded extensions) outside query-core. Register it explicitly on the relevant query-core tests/base class, or exclude these resources from the test JAR.

apmoriarty:
> `query-core` publishes its test output as a `test-jar`, and downstream modules consume that classifier. The new `junit-platform.properties` and ServiceLoader entry therefore export global JUnit extension autodetection into those consumers, potentially running this extension (and other service-loaded extensions) outside query-core. Register it explicitly on the relevant query-core tests/base class, or exclude these resources from the test JAR.

Fair enough, will go the exclusion route.

PR #3906, opened 2026-09-02
Keep JUnit extension autodetection out of the query-core test-jar

apmoriarty:
The evictor's junit-platform.properties and ServiceLoader entry were packaged into the test-jar, turning on extension autodetection in the three modules that consume it.

### DIFF
--- a/warehouse/query-core/pom.xml
+++ b/warehouse/query-core/pom.xml
@@ -449,6 +449,15 @@
                         <goals>
                             <goal>test-jar</goal>
                         </goals>
+                        <configuration>
+                            <!-- These two register MetadataHelperCacheEvictor for every test class in this module. Modules that
+                                 consume the test-jar have no junit-platform.properties of their own, so shipping them would turn
+                                 on extension autodetection in those modules as well. -->
+                            <excludes>
+                                <exclude>junit-platform.properties</exclude>
+                                <exclude>META-INF/services/org.junit.jupiter.api.extension.Extension</exclude>
+                            </excludes>
+                        </configuration>
                     </execution>
                 </executions>
             </plugin>


avgAGB (approved):

drewfarris (approved):

drewfarris:
What this changes: In warehouse/query-core/pom.xml, it configures the  maven-jar-plugin's  test-jar  goal to exclude two files from the packaged test-jar artifact:

•  junit-platform.properties 
•  META-INF/services/org.junit.jupiter.api.extension.Extension 

Why (per the PR's own comment): These files register  MetadataHelperCacheEvictor  as a JUnit 5 extension via auto-detection for every test class in this module. That's fine locally. But when other Maven modules depend on  query-core 's test-jar artifact (a common pattern for reusing test utilities/fixtures), those files get bundled into the consuming module's classpath too. Since consumer modules don't have their own  junit-platform.properties  to override this, JUnit's extension auto-detection gets silently turned on for their test runs as a side effect — a leaky packaging bug.

Implications:

• Fixes unintended cross-module behavior: consuming modules will no longer have  MetadataHelperCacheEvictor  (or JUnit extension autodetection generally) silently enabled just by depending on  query-core 's test-jar.
• Scope is narrow and low-risk: it's a build/packaging-only change (Maven plugin config), no source/logic changes — shouldn't affect production behavior or  query-core 's own test runs.
• Possible behavior change for downstream test suites: any module that depended on  query-core 's test-jar and was unknowingly relying on that extension being active (e.g., automatic Accumulo metadata-helper cache eviction between tests) could see different behavior (e.g., stale caches, test isolation issues) after this fix — worth checking CI for those modules.
• Good practice enforced: encapsulates test-only cross-cutting concerns to the module that defines them, preventing accidental "action at a distance" via test-jar dependencies — a reasonable, generally applicable Maven hygiene fix.

PR #3907, opened 2026-09-02
Task/java17 build baseline

alerman:
Codex Assisted

Enable building using java 17, but dont break java 11 yet.

### DIFF
--- a/BUILDME.md
+++ b/BUILDME.md
@@ -26,6 +26,10 @@ The id of the server matters, and should match what is used in the datawave pare
 The root build remains Java 11 by default. Builds run on Java 17 are supported through the
 `java17` profile and remain backward compatible by compiling with `--release 11`.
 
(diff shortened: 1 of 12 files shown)


sylvesterkaczmarek (commented):
This adds Java 17 as a supported build mode, but I don't see a Java 17 CI leg in the change. The existing checks can all stay green on Java 11 even if `-Pjava17` or the microservices Java 17 profile regresses. Could one CI matrix leg run the root and microservices reactors on JDK 17 with `-Pjava17`, while retaining the Java 11 baseline?

apmoriarty (approved):

avgAGB (approved):

avgAGB (approved):

sylvesterkaczmarek (commented):
Rechecked current `97444148`. My earlier CI concern remains blocking: the PR adds and documents the Java 17 build mode, but still adds no CI workflow that runs Maven on JDK 17 with `-Pjava17`. Existing Java 11 checks cannot validate this new mode.

PR #3908, opened 2026-09-02
Adds the ability to rewrite responses from the AnnotationManagerBean

drewfarris:
Supercedes #3902 which was opened from a fork inadvertently.

In some cases, we'll need to intercept the response from the AnnotationManagerBean and rewrite it to modify the contents in caller or user-specific ways. This PR adds the interface for a JAX-RS response rewriter that can be used for any JAX-RS implementation, updates AnnotationManagerBean to support a pluggable response rewriter and adds tests using a Mock rewriter.

This commit refactors AnnotationManagerRequestContext to be a top level class instead of a static inner class of AnnotationManagerBean.

This updates includes changes based on review feedback:
    - Moved ResponseRewriter and friends to datawave-ws-common, removed added dependency from query-core
    - Made AnnotationResponseRewriter bean required in AnnotationManagerBean, provided a no-op implementation to use by default
    - Refactored AnnotationManagerBean to return accumulo client prior to potential costly rewrite operation
    - Added null check in rewriteResponse to prevent null returns. Now properly returns a 500, added test.
    - Identifiers in 404 and 500 responses are OK and added tests to show this.
    - Added tests to use rewriteResponse directly, added tests for null rewriter, NoOp, null return and error/404 responses
    - Removed redundant dependency entry in root pom.
    - Fixed visibility in AnnotationManagerRequest context, remove fully qualified class references, fixed typos and added VisibleForTesting annotation.

### DIFF
--- a/web-services/annotations/src/test/java/datawave/webservice/annotation/AnnotationManagerBeanResponseRewriterTest.java
+++ b/web-services/annotations/src/test/java/datawave/webservice/annotation/AnnotationManagerBeanResponseRewriterTest.java
@@ -1,0 +1,117 @@
+package datawave.webservice.annotation;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.lang.reflect.Field;
+
+import javax.ws.rs.core.Response;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import datawave.core.query.logic.ResponseRewriter;
+import datawave.core.query.logic.ResponseRewriterContext;
+
+/**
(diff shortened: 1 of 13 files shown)


drewfarris:
Added `linked` label because downstream `AnnotationManagerFactory.xml` will need to add a NoOp or some other response rewriter.

apmoriarty (approved):

web-services/annotations/src/test/java/datawave/webservice/annotation/AnnotationManagerBeanResponseRewriterTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+    @BeforeEach
+    public void setup() {
+        bean = new AnnotationManagerBean();
+    }
+
+    @Test
+    public void testRewriterAppliedToSuccessResponse() throws Exception {
+        // Inject the mock rewriter
+        setField(bean, "responseRewriter", new MockResponseRewriter());
```
FineAndDandy, line 36: instead of reflection another option is mockito

FineAndDandy (approved):

avgAGB (approved):

PR #3894, opened 2026-08-28
Add annotation microservice from PR #3710

drewfarris:
Extracts the microservices/ changes (annotation service module) from the feature/annotation-service branch. This is a follow-on to feature/annotation-service-core; once the core annotation changes are merged to integration and included in a tagged release, this branch's dependency on datawave-core-annotation should be updated to reference that released version before this PR is opened.

PR #3909, opened 2026-09-02
Adds annotation microservice to main Datawave project.

drewfarris:
Extracts the changes from the feature/annotation-service branch in #3701 which contained changes to both core/annotation and microservices/services/annotation

This is a follow-on to #3893 updated to depend on release 8.0.0 which includes the required change from core/annotation in that PR. It supercedes #3894 which is now closed.


sylvesterkaczmarek (commented):
`processConfirmAck()` does a separate `containsKey()` and `get()` while the sender removes the latch in its `finally` block. A late acknowledgement can pass `containsKey()`, race with that removal, then dereference `null` in the service activator. Fetch the latch once and null-check it (or use `computeIfPresent`) and add an ack-vs-timeout race test.

avgAGB (dismissed):

FineAndDandy (dismissed):

sylvesterkaczmarek (commented):
Re-reviewed the current head (`3bbe537ca8`) because I am explicitly requested as a reviewer. The acknowledgement race I raised earlier is still present in `AnnotationControllerV1.processConfirmAck()`: it does `containsKey(correlationId)` and then a separate `get(correlationId).countDown()`, while the sending path removes that same latch in its `finally` block. A late ack can therefore pass `containsKey`, race with removal, and dereference `null`. Please fetch the latch once and null-check it (or use an atomic map operation), and add an ack-vs-timeout/removal regression. I do not have another blocker beyond that finding from this re-review.

avgAGB (dismissed):

sylvesterkaczmarek (commented):
Rechecked current `3bbe537c`. My acknowledgement-race concern is resolved: `processConfirmAck()` now uses the tracker’s atomic `countdown(correlationId)` operation, and the late-ack regression covers the post-timeout/removal path without leaking a latch. No remaining blocker from my earlier review.

sylvesterkaczmarek (commented):
Rechecked current `b878ba78`. The per-invocation timer key fixes duplicate/concurrent annotation writes overwriting one another, and `AccumuloHealthChecker` only consumes the map entries/timestamps, not annotation-id key semantics, so the composite key does not change the health calculation. Main and microservice tests are green; the lone compose failure is in unrelated service startup/introspection, while the snapshot compose run passes. No blocker from this latest change.

drewfarris:
This is a large PR. I'm working on breaking it down stacked set of smaller PRs to ease review.

drewfarris:
Replaced by #3936, #3937, #3938, #3939

sylvesterkaczmarek:
Thanks for splitting this into the smaller stack. I’ll follow the replacement PRs; the acknowledgement race I raised here was resolved on the final head, so nothing from my review needs to hold up that carry-forward.

PR #3910, opened 2026-09-03
Update references to Hadoop com.sun.jersey dependencies

lbschanno:
Hadoop 3.5.0 uses Jersey 2.x (org.glassfish.jersey) instead of the legacy Jersey 1.x (com.sun.jersey). Update references to it throughout relevant pom files.

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -10,9 +10,9 @@
     <!-- Declare all datawave modules as dependencies in order to have the javadoc include all the datawave source. -->
     <dependencies>
         <dependency>
-            <groupId>com.sun.jersey</groupId>
+            <groupId>org.glassfish.jersey.core</groupId>
             <artifactId>jersey-server</artifactId>
-            <version>1.9</version>
+            <version>2.46</version>
             <scope>compile</scope>
             <optional>true</optional>
         </dependency>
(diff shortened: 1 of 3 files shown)


sylvesterkaczmarek (commented):
Hi, I think the Jersey exclusion update is incomplete because Jersey 2 splits its artifacts across multiple group IDs. Hadoop 3.5.0 `hadoop-common` depends on `org.glassfish.jersey.core`, `org.glassfish.jersey.containers`, and `org.glassfish.jersey.inject`; excluding only `.core` leaves the servlet/HK2 pieces transitive. If the intent is still to exclude Hadoop's Jersey stack, could those groups be excluded too and verified with `dependency:tree`?

avgAGB (approved):

sylvesterkaczmarek (commented):
Re-reviewed current `d6da9c7`. My earlier Jersey exclusion concern is resolved: the Hadoop dependencies now exclude `org.glassfish.jersey.core`, `.inject`, and `.containers`, covering the split Jersey 2 artifacts. The currently red main build is failing earlier on the unavailable `accumulo-server-base:4.0.0-SNAPSHOT` dependency rather than this exclusion change. No remaining blocker from my review.

PR #3913, opened 2026-09-08
Fixes issues with annotaton validation context, error messages and leaked threads

drewfarris:
- Addressed issue where `Validator.addMemberValidator` discarded its `messageContext` argument, hardcoding an empty string when delegating to the overloaded method. Every `AnnotationValidators` call site using this overloaded method (segment identity, base segment, annotation identity, annotation, annotation-update validation) silently lost its descriptive error-context suffix.
- Fixed incorrect span validation messages: `AnnotationValidators.spanBoundaryValidator` i paired its two checks with the wrong messages: a missing `end` value reported "must have a start value", and a missing `start` value reported "must have an end value". The validation logic was correct, but the 400 response text describing which field was actually missing was reversed. Swapped the two message strings to match the field being tested
- Fixed issue where DAO threads wouldn't be cleaned up on error: `FederatedAnnotationReader.callDaos` logged still-pending DAO `Future`s as timed out but never cancelled them. Since the caller-provided executor is typically an unbounded cached thread pool  a DAO call that blocks indefinitely would permanently occupy a pooled thread that is never reclaimed, since it never returns to idle. Now calls `future.cancel(true)` on each remaining pending DAO after the timeout elapses.
- Added tests for these fixes

### DIFF
--- a/core/annotation/src/test/java/datawave/annotation/util/v1/AnnotationValidatorsTest.java
+++ b/core/annotation/src/test/java/datawave/annotation/util/v1/AnnotationValidatorsTest.java
@@ -10,9 +10,7 @@ import org.junit.jupiter.api.Test;
 
 import datawave.annotation.protobuf.v1.Annotation;
 import datawave.annotation.protobuf.v1.AnnotationSource;
-import datawave.annotation.protobuf.v1.BoundaryType;
 import datawave.annotation.protobuf.v1.Segment;
-import datawave.annotation.protobuf.v1.SegmentBoundary;
 import datawave.annotation.test.v1.AnnotationTestDataUtil;
 import datawave.annotation.util.Validator;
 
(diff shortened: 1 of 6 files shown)


apmoriarty (dismissed):

core/annotation/src/test/java/datawave/annotation/util/v1/AnnotationValidatorsTest.java
```diff
@@ -73,6 +75,33 @@ public void testSegmentValidators() {
...
+    public void testSpanBoundaryMissingStartReportsStartSpecificMessage() {
+        SegmentBoundary missingStart = SegmentBoundary.newBuilder().setBoundaryType(BoundaryType.TIME_MILLI).setEnd(100).build();
+        Segment segment = AnnotationTestDataUtil.generateTestSegment().toBuilder().setBoundary(missingStart).build();
+
+        Validator.ValidationState<Segment> validationState = AnnotationValidators.checkSegment(segment);
+        assertInvalid(validationState);
+        assertTrue(validationState.getErrors().stream().anyMatch(e -> e.contains("must have a start value")), getErrorMessage(validationState));
+        assertFalse(validationState.getErrors().stream().anyMatch(e -> e.contains("must have an end value")), getErrorMessage(validationState));
+    }
+
```
tyjedz, line 104: Are these tests necessary? It shouldn't ever fail now that the message values are swapped right?

drewfarris: Yep, could totally drop these.

tyjedz (approved):

PR #3923, opened 2026-09-09
Run the compose ZooKeeper service from the 3.9.5 distribution

### LINKED ISSUE #3922: Accumulo 4 cannot initialize: the compose ZooKeeper server is 3.5.9 and rejects addWatch
**What happens**

With HDFS, YARN and ZooKeeper all reporting healthy, accumulo-manager comes up unhealthy and never leaves initialization; accumulo-gc, accumulo-monitor and accumulo-tserver therefore never start. The manager log repeats 354 identical cycles of "init ERROR Error setting persistent watcher in ZooKeeper, retrying... KeeperException$ConnectionLossException: KeeperErrorCode = ConnectionLoss for /namespaces" at ZooSession.addPersistentRecursiveWatchers(ZooSession.java:331) / Initialize.doInit(Initialize.java:187). Compose still shows `zookeeper ... (healthy)`.

**Why**

The running ZooKeeper server is bigtop's 3.5.9, which does not implement addWatch (opcode 106, added in ZooKeeper 3.6.0) -- its log says "Received packet at server of unknown type 106" and it drops the connection, so the client reports ConnectionLoss. The stack image contains two ZooKeepers: /opt/zookeeper/lib/zookeeper-3.9.5.jar from datawave-accumulo/Dockerfile (ZOOKEEPER_VERSION=3.9.5, ARG at :10) and /usr/lib/zookeeper/lib/zookeeper-3.5.9.jar from the bigtop yum install. docker-compose.yml's zookeeper service hardcoded entrypoint /usr/lib/zookeeper/bin/zkServer.sh with ZOOBINDIR=/usr/lib/zookeeper/bin, and /opt/zookeeper/bin/zkServer.sh did not exist because the Dockerfile extracted only --wildcards "apache-zookeeper-${ZOOKEEPER_VERSION}-bin/lib/zookeeper*.jar". So 3.9.5 was present for Accumulo's client classpath and there was no 3.9.5 server to run. The healthcheck hid it: `echo ruok` is a 3.4-era four-letter word that 3.5.9 answers happily.

**How this was established**

`echo srvr | nc localhost 2181` against the running server returned "Zookeeper version: 3.5.9-5540c493e389d4405ac969d48bff9d2459893f6e, built on 07/27/2023 05:21 GMT"; the ZooKeeper server log carried "WARN [NIOWorkerThread-30:ZooKeeperServer@837] - Received packet at server of unknown type 106". After the fix: `echo srvr | nc zookeeper 2181` returned "Zookeeper version: 3.9.5-293c895a8d966a3ecb92872be4a1daf87d725da2, built on 2026-02-11" and `docker compose logs zookeeper | grep -c 'unknown type 106'` returned 0; Accumulo 4 initialized, reaching the ordinary "instance secret is still set to the default" warning. Re-verified today: `grep -rn ZOOKEEPER_VERSION datawave-accumulo/Dockerfile` shows ARG ZOOKEEPER_VERSION=3.9.5 at :10 and the tarball extraction at :54, and `git show 1fcf863a5f` shows the wildcard removal plus the generated conf/zoo.cfg.

Found while standing the compose stack up on `accumulo4-2026-quickstart`. Work for #3766.

SethSmucker:
Accumulo 4 sets persistent recursive watchers during init (ZooKeeper addWatch,
opcode 106, added in 3.6). The service ran bigtop's 3.5.9 from /usr/lib/zookeeper,
which answers ruok but logs 'Received packet at server of unknown type 106' and
drops the connection, so Initialize loops on ConnectionLoss for /namespaces and
no tablet server ever starts. Verified with echo srvr: 3.5.9 before, 3.9.5 after,
and 'unknown type 106' goes from present to zero.

Needs the matching datawave-stack-docker-images change that actually puts a
3.9.5 server at /opt/zookeeper - today the image extracts only its jars.

Work for #3766

Fixes #3922

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -118,10 +118,14 @@ services:
     profiles: [datawave-stack]
     image: *stack-image
     user: zookeeper
-    entrypoint: ["/usr/lib/zookeeper/bin/zkServer.sh"]
-    command: ["--config", "/etc/zookeeper/conf.dist", "start-foreground"]
+    # Accumulo 4 needs a ZooKeeper server that understands addWatch (opcode 106,
+    # ZooKeeper 3.6+). /usr/lib/zookeeper is bigtop's 3.5.9, which answers ruok
+    # but rejects addWatch, so Accumulo's Initialize loops on ConnectionLoss.
+    # /opt/zookeeper is the 3.9.5 distribution the stack image downloads.
+    entrypoint: ["/opt/zookeeper/bin/zkServer.sh"]
+    command: ["--config", "/opt/zookeeper/conf", "start-foreground"]
     environment:
-      - ZOOBINDIR=/usr/lib/zookeeper/bin
+      - ZOOBINDIR=/opt/zookeeper/bin
       - ZOO_LOG_DIR=/var/log/zookeeper
     ports:
       - "2181:2181"


SethSmucker:
Closing — superseded by 8b94f9e6c2 on `accumulo4-2026-quickstart`, which makes
the identical change (entrypoint `/opt/zookeeper/bin/zkServer.sh`, `ZOOBINDIR=/opt/zookeeper/bin`).

I hit the same problem independently and had this branch open before spotting
yours; no need for both. The diagnosis behind it is in #3922 if it's useful — the
short version is that bigtop's 3.5.9 answers `ruok` but rejects `addWatch`
(opcode 106, ZooKeeper 3.6+), so Accumulo 4's `Initialize` loops on
`ConnectionLoss for /namespaces` and no tablet server ever starts.

One thing your change may still need, flagged in #3922: the service still passes
`--config /etc/zookeeper/conf.dist`, and that file has no
`4lw.commands.whitelist`. ZooKeeper 3.9 disables four-letter words by default, so
the `ruok` healthcheck can refuse once the server is genuinely 3.9.5. Worth a
quick check on your side — it's one line if it bites.

PR #3929, opened 2026-09-09
Let ingest and verifySplits talk to Accumulo 4

### LINKED ISSUE #3928: ingest cannot reach Accumulo 4: client -p removed, auth.type missing, netty-handler off the shell classpath
**What happens**

The ingest service sits on "Waiting for Accumulo..." for 38 failed healthchecks and never gets past it, blocking query, webservice, metrics and executor. The loop is docker/stack/initialize-datawave.sh:10.

**Why**

Three stacked breaks, each hidden behind the last. (a) Accumulo 4 dropped the client -p option -- `accumulo shell -u root -p secret -e info` fails with "IllegalArgumentException: The Client options -p have been dropped. Use accumulo-client.properties for any connection or token options." -- and initialize-datawave.sh still used it in 7 places (the file is byte-identical between integration and accumulo4-2026-quickstart). (b) Switching to -c surfaced "IllegalArgumentException: auth.type is not set" at ClientProperty.validateProperty(ClientProperty.java:329): docker/stack/accumulo-client.properties sets instance.name, instance.zookeepers, auth.principal and auth.token but not auth.type, which Accumulo 4 validates. (c) ZooKeeper 3.9's client loads io.netty.handler.ssl.SslContext in ZKConfig.handleBackwardCompatibility(ZKConfig.java:133), reached from ZooSession.reconnect(ZooSession.java:209), and the shell threw NoClassDefFoundError for it: lbschanno's a4 compose adds netty-handler to the datawave-extlib init container (commit cf4085c201, the jaxb-api -> netty-handler swap), but that only reaches the Accumulo servers -- the ingest service did not mount datawave_accumulo_ext, and accumulo-env.sh:13 built CLASSPATH="${conf}:${lib}/*:${HADOOP_CONF_DIR}:${ZOOKEEPER_HOME}/*:${ZK_JARS}:/usr/lib/hadoop/client/*" with lib/ext absent, so accumulo shell never saw those jars.

**How this was established**

Each break was surfaced by running the failing command by hand in the container and reading the exception, then re-running: `accumulo shell -u root -p secret -e info` gave the dropped-options error; the -c form gave "auth.type is not set"; adding auth.type gave the SslContext NoClassDefFoundError. Break (c) was proved by copying netty-handler into /opt/accumulo/lib in the running container, after which the shell reported "Shell - Apache Accumulo Interactive Shell / - version: 4.0.0-SNAPSHOT / - instance name: my-instance-01 / - instance id: 5dea9dfe-dcec-4510-82f4-4d1ee8543a25". After the committed fix the "Waiting for Accumulo" count went 38 -> 0 and ingest proceeded to TableCreator and began building the DataWave tables. Re-verified today: `git show a9ebf46909^:docker/stack/initialize-datawave.sh | grep -c '\-p secret'` returns 7 and the current file returns 0; `git show a9ebf46909^:docker/stack/accumulo-env.sh` shows line 13 as the CLASSPATH assignment without ${lib}/ext.

Found while standing the compose stack up on `accumulo4-2026-quickstart`. Work for #3766.

SethSmucker:
`initialize-datawave.sh` hung on "Waiting for Accumulo" because its `accumulo
shell` calls could not connect.

**Accumulo 4 dropped the client `-p` option.** From the release notes: "The
Client options `-p` have been dropped. Use `accumulo-client.properties` for any
connection or token options. See `-c, --config-file` option." All seven shell
calls in `initialize-datawave.sh`, and the one in `scripts/verifySplits.sh`,
still passed `-u`/`-p`, so every one of them failed. They now pass
`-c /opt/accumulo/conf/accumulo-client.properties` - the file compose already
mounts into these containers.

That file needs `auth.type`, which Accumulo 4 validates. #3932 added it.

**One follow-on fix.** #3932 replaced `accumulo-client.properties` with the
stock Accumulo 4 template, which ships `instance.zookeepers=localhost:2181`.
Compose mounts this file into `accumulo-manager`, `accumulo-tserver`,
`accumulo-gc`, `accumulo-monitor` and `ingest`; ZooKeeper runs in its own
`zookeeper` service, so `localhost:2181` resolves to nothing in any of them.
Restored `instance.zookeepers=zookeeper:2181`, the value the file carried
before #3932. `accumulo.properties` still has
`instance.zookeeper.host=zookeeper:2181`, so the server processes were never
affected - only client calls.

Verified on a running a4 stack: the "Waiting for Accumulo" count went 38 to 0,
ingest reached `TableCreator` and completed, and `verifySplits.sh` reported
51/51 splits - the same number integration gives.

Work for #3766

Fixes #3928

### DIFF
--- a/docker/scripts/verifySplits.sh
+++ b/docker/scripts/verifySplits.sh
@@ -80,7 +80,7 @@ fi
 # reader to the init script when the stack is simply down or the table is missing.
 GETSPLITS_ERR=$(mktemp)
 GETSPLITS_OUT=$(compose exec -T "${ACCUMULO_SERVICE}" \
-    accumulo shell -u "${ACCUMULO_USER}" -p "${ACCUMULO_PASSWORD}" -e "getsplits -t ${SHARD_TABLE}" 2>"${GETSPLITS_ERR}")
+    accumulo shell -c /opt/accumulo/conf/accumulo-client.properties -e "getsplits -t ${SHARD_TABLE}" 2>"${GETSPLITS_ERR}")
 GETSPLITS_STATUS=$?
 
 if [ "${GETSPLITS_STATUS}" -ne 0 ] ; then
--- a/docker/stack/accumulo-client.properties
+++ b/docker/stack/accumulo-client.properties
@@ -11,7 +11,7 @@
 instance.name=my-instance-01
 
 ## Zookeeper connection information for Accumulo instance
-instance.zookeepers=localhost:2181
+instance.zookeepers=zookeeper:2181
 
 ## Zookeeper session timeout
 #instance.zookeepers.timeout=30s
--- a/docker/stack/initialize-datawave.sh
+++ b/docker/stack/initialize-datawave.sh
@@ -6,8 +6,9 @@ readonly HDFS_INGEST_DIR=/datawave/ingest
 readonly FIXTURES=/opt/datawave-test-data
 readonly SHARD_SPLITS=/stack/shard-splits.txt
 readonly NUM_SHARDS_FLOOR_DATE=19000101
+readonly ACCUMULO_CLIENT_PROPS=/opt/accumulo/conf/accumulo-client.properties
 
-until accumulo shell -u root -p secret -e info >/dev/null 2>&1; do
+until accumulo shell -c "${ACCUMULO_CLIENT_PROPS}" -e info >/dev/null 2>&1; do
     echo "Waiting for Accumulo..."
     sleep 2
 done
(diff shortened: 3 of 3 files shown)


lbschanno:
The Zookeeper netty issues may be already be fixed by a recent commit to the accumulo4-2026-quickstart command: 8b94f9e6c2. This, in conjunction with the latest changes to the same branch on datawave-stack-docker-images, unpacks the Zookeeper lib and script jars, allowing Accumulo to properly add the Zookeeper jars to the CLASSPATH.

SethSmucker:
You're right for the Accumulo server containers, and your version is better than
mine there — I checked:

- ZooKeeper 3.9.5's own distribution ships netty (`netty-handler-4.1.130.Final`
  and 15 friends in `lib/`).
- `docker/stack/accumulo-env.sh:8-9` builds `ZK_JARS` from
  `${ZOOKEEPER_HOME}/lib/*.jar` and puts it on `CLASSPATH`.
- So once the stack image unpacks the whole `lib/` rather than just
  `zookeeper*.jar`, netty arrives on the Accumulo CLASSPATH for free, at the
  version ZooKeeper itself ships. Confirmed in the stack image: **16** netty jars
  under `/opt/zookeeper/lib`.

That's cleaner than what I did, which pulled netty 4.1.79 out of DataWave's
ingest lib via the extlib volume.

One thing I don't think it covers, and this is the case that bit me: the failure
was in the **ingest** container, which runs `ingest-compose` — DataWave's image,
not the stack image. It has its own `/opt/zookeeper -> /opt/zookeeper-3.9.5`, and
that one still contains only three jars:

```
$ docker compose exec ingest ls /opt/zookeeper/lib/
zookeeper-3.9.5.jar
zookeeper-jute-3.9.5.jar
zookeeper-prometheus-metrics-3.9.5.jar
```

No netty. So `ZK_JARS` is empty of it there, and `accumulo shell` inside ingest
still hits `NoClassDefFoundError: io/netty/handler/ssl/SslContext` from
`ZKConfig.handleBackwardCompatibility` — which is what makes
`initialize-datawave.sh` sit on "Waiting for Accumulo..." forever.

So I think this splits into:

- **the `-p` removal and the missing `auth.type`** — needed regardless, and the
  larger part of this PR
- **the netty half** — needed only because the *ingest* image packages ZooKeeper
  the old way

If the ingest image's ZooKeeper packaging is getting the same treatment, I'm glad
to drop the netty half and rebase on that instead. Happy either way — just say
which and I'll rework it.

